PR #1, opened 2022-08-17
Update CONTRIBUTORS

SwayStar123:
Did as instructed 👍


SwayStar123:
tfw jump trading themselves calls you invalid

PR #42, opened 2023-01-16
util: fd_shmem_cfg needs a guard if command not available

CantelopePeel:
If the specified command is not installed, tell the user to install it. I think there are other places where this is needed, but I just wanted to do one before I do it everywhere else.

TESTED=With and without the command installed.

### DIFF
--- a/src/util/shmem/fd_shmem_cfg
+++ b/src/util/shmem/fd_shmem_cfg
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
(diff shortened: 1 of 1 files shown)


ripatel-fd:
Thanks for this contribution. Since we do not currently accept GitHub PRs, I've moved this to here: https://git.firedancer.io/c/firedancer/+/419

ripatel-fd:
Merged here: https://github.com/firedancer-io/firedancer/commit/7101ab2ffe22f978424d2f1b134c0577d9294dcd

PR #50, opened 2023-01-23
util, tango: add ASan poisoning

ripatel-fd:
Adds manual ASan poisoning to detect OOB memory accesses within a Firedancer shm workspace.

Depends on https://github.com/firedancer-io/firedancer/pull/49


ripatel-fd:
@kbowers-jump Unfortunately, ASan support for tango is a bit more complex. 

The magic number checks occur before ASan permits access to a shared memory object.
Therefore, the magic number checks themselves require a temporary ASan exception.
To not duplicate too much code, I added some helper inlines for this.

ripatel-fd:
Outdated

PR #51, opened 2023-01-23
sha256: export fd_sha256_core

ripatel-fd:
Exports block function of SHA-256 for use in high-performance hashing code.

Change-Id: I65763275dcc61dcb3f132e83ccf0e53ba043c17e

Imported from https://git.firedancer.io/c/firedancer/+/353

### DIFF
--- a/src/ballet/sha256/fd_sha256.c
+++ b/src/ballet/sha256/fd_sha256.c
@@ -102,9 +102,9 @@ fd_sha256_delete( void * shsha ) {
     https://github.com/openssl/openssl/blob/openssl-3.0.7/crypto/sha/sha256.c */
 
 static void
-fd_sha256_core_ref( uint *        state,
-                    uchar const * block,
-                    ulong         block_cnt ) {
+fd_sha256_core_ref( uint        * FD_RESTRICT state,
+                    uchar const * FD_RESTRICT block,
+                    ulong                     block_cnt ) {
 
   static uint const K[64] = {
     0x428a2f98UL, 0x71374491UL, 0xb5c0fbcfUL, 0xe9b5dba5UL,
@@ -200,7 +200,19 @@ fd_sha256_core_ref( uint *        state,
 
 }
 
-#define fd_sha256_core fd_sha256_core_ref
+/* Name of internal definition of block function.
+   See fd_sha256_core_ref (reference implementation) for the method signature.
+   block_cnt must be greater than zero. */
+#define fd_sha256_core_ fd_sha256_core_ref
+
+/* Export block function */
+void
+fd_sha256_core( uchar       * FD_RESTRICT state,
+                uchar const * FD_RESTRICT block,
+                ulong                     block_cnt ) {
+  if( FD_UNLIKELY( block_cnt==0UL ) ) return;
+  fd_sha256_core_ref( (uint *)state, block, block_cnt );
+}
 
 fd_sha256_t *
 fd_sha256_init( fd_sha256_t * sha ) {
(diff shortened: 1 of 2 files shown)


src/ballet/sha256/fd_sha256.h
```diff
@@ -61,6 +62,9 @@ fd_sha256_align( void );
 FD_FN_CONST ulong
 fd_sha256_footprint( void );
 
+// Synchronization functions
```
ptaffet-jump, line 65: "Synchronization" doesn't strike me as the right adjective, but I don't know what other parts of the codebase use

src/ballet/sha256/fd_sha256.h
```diff
@@ -61,6 +62,9 @@ fd_sha256_align( void );
...
+// =========================
```
ptaffet-jump, line 66: The codebase doesn't use // comments much. I think dividers like https://github.com/firedancer-io/firedancer/blob/b2088f1f3ab7362e48233636e16b162a74c99d91/src/util/tmpl/fd_map.c#L276 are more in the right style

src/ballet/sha256/fd_sha256.h
```diff
@@ -118,6 +125,47 @@ void *
 fd_sha256_fini( fd_sha256_t * sha,
                 void *        hash );
 
+// Advanced interface
+// ==================
+
+/* FD_SHA256_INIT_n are the eight 32-bit integers that comprise the initial hash state.
+   Note that the byte order of each integer is reversed (little-endian). */
```
ptaffet-jump, line 132: Comments typically break at 80. If you're using VIM, try `gq`.

src/ballet/sha256/fd_sha256.h
```diff
@@ -118,6 +125,47 @@ void *
...
+  uint * s = (uint *)state;
+  s[0] = FD_SHA256_INIT_0;
+  s[1] = FD_SHA256_INIT_1;
+  s[2] = FD_SHA256_INIT_2;
+  s[3] = FD_SHA256_INIT_3;
+  s[4] = FD_SHA256_INIT_4;
+  s[5] = FD_SHA256_INIT_5;
+  s[6] = FD_SHA256_INIT_6;
+  s[7] = FD_SHA256_INIT_7;
+}
```
ptaffet-jump, line 155: This code is essentially a duplicate of `fd_sha256_init`. Maybe have that function call this function?

src/ballet/sha256/fd_sha256.h
```diff
@@ -118,6 +125,47 @@ void *
...
+
+/* fd_sha256_core invokes the internal block function of SHA-256 block_cnt times.
+   For each iteration, reads and advances the block ptr by one block and updates state.
```
ptaffet-jump, line 158:
```
/* fd_sha256_core hashes the block_cnt blocks of data starting at block, updating 
the internal hash state stored in state.  The computation is essentially an 
optimized version of fd_sha256_append( .., block, block_cnt * FD_SHA256_BLOCK_SZ) 
since it does not need to consider padding or buffered data and does not keep track 
of how much data has been hashed.  Additionally, since the caller provides the 
buffer, it can eliminate a memcpy of the final hash.  Because this function does 
not keep track of how much data has been hashed, it's the calling code's 
responsibility to write the equivalent of `fd_sha256_fini`.  This can still be 
helpful if the size is known statically because then the trailing padding can also 
be generated statically.
```

src/ballet/sha256/fd_sha256.c
```diff
@@ -200,7 +200,19 @@ fd_sha256_core_ref( uint *        state,
...
 
-#define fd_sha256_core fd_sha256_core_ref
+/* Name of internal definition of block function.
+   See fd_sha256_core_ref (reference implementation) for the method signature.
+   block_cnt must be greater than zero. */
+#define fd_sha256_core_ fd_sha256_core_ref
+
+/* Export block function */
+void
+fd_sha256_core( uchar       * FD_RESTRICT state,
```
ptaffet-jump, line 210: What do you think about renaming this to avoid the name collision that you resolved with the extra `_`? Some ideas: `fd_sha256_chunk`, `fd_sha256_block`, `fd_sha256_hash_blocks`

ptaffet-jump (commented):
Copied comments over from Gerrit

ripatel-fd:
Closing since this is not required at the moment. Might revisit this in the future.

PR #52, opened 2023-01-23
bmtree: add module and bmtree{20,32}_commit2

ripatel-fd:
- Adds new bmtree module to ballet (Binary Merkle tree)
- Adds tree template for 20 and 32 byte node variants
- Adds simple root derive algorithm

Change-Id: Ide93a837e3328c489a3a4c625bf1db0d4955e933

Imported from https://git.firedancer.io/c/firedancer/+/304

### DIFF
--- a/src/ballet/bmtree/fd_bmtree.h
+++ b/src/ballet/bmtree/fd_bmtree.h
@@ -1,0 +1,294 @@
+#ifndef HEADER_fd_src_ballet_bmtree_fd_bmtree_h
+#define HEADER_fd_src_ballet_bmtree_fd_bmtree_h
+
+#include "../sha256/fd_sha256.h"
+
+/* fd_bmtree{20,32} is a binary Merkle tree used in the Solana protocol.
+   It uses the SHA-256 hash function.
+
+   Specification:
+   https://github.com/solana-foundation/specs/blob/main/core/merkle-tree.md
+
+   It is generally used as a vector commitment scheme
+   wherein the root node of the tree commits the vector of leaf nodes.
+
+   All methods provided by this Merkle tree derive from these three
+   basic operations.
+
+     1. Construct leaf node:
+
+        (leaf blob) -> (node)
+
+     2. Construct branch node with two children:
+
+        (node, node) -> (node)
+
+     3. Construct branch node with one child:
+
+        (node) -> (node)
+
+   Example derived methods.
+   (TODO not all of them are provided by this header file yet)
+
+     4. Construct full tree:
+
+        (vector of leaf blobs) -> (tree of nodes)
+
+     5. Create inclusion proof from tree data
+
+        (tree of nodes, node index) -> (inclusion proof)
+
+     6. Verify node inclusion proof
+
+        (node, root node, inclusion proof) -> (bool)
+
+   **Topology**
+
+   Tree topology has the following constraints:
+
+    - All leaf nodes are in the bottom level
+
+    - If a given layer `L`
+      with number of nodes `L_n` ...
+
+      ... has exactly one node,
+          this one node is the root node
+          and forms the uppermost layer.
+
+      ... has more than one node
+          ... and `L_n % 2 == 0`,
+              the layer above contains n/2 nodes
+
+          ... and `L_n % 2 == 1`,
+              the layer above contains (n+1)/2 nodes.
+
+   A simple algorithm to approach such a a tree is as follows:
+   (Note that the code uses here uses an optimized approach)
+
+    - Start with the smallest complete binary tree that has at least
+       `n` leaf nodes.
+    - Label the leaf nodes from left to right `L_0`, `L_1`, ... `L_(n-1)`
+    - Delete any un-labeled leaf nodes, and then recursively delete any
+      nodes with no children.
+    - For any nodes with a single remaining child, duplicate the link to the child.
+    - Each non-leaf node now has exactly two children, counting duplicates.
+
+   Example: Tree with 1 leaf node (root node = leaf node)
+
+    L0
+
+   Example: Tree with 4 leaf nodes
+
+           Iδ
+          /  \
+         /    \
+       Iα      Iβ
+      /  \    /  \
+     L0  L1  L2  L3
+
+   Example: Tree with 5 leaf nodes
+
+              Iζ
+             /  \
+            /    \
+           Iδ     Iε
+          /  \     \\
+         /    \     \\
+       Iα      Iβ    Iγ
+      /  \    /  \   ||
+     L0  L1  L2  L3  L4
+
+   **Construction**
+
+   The input data is a vector of arbitrary-sized binary blobs.
+   First, each blob is converted to a fixed-size leaf node by hashing
+   the blob in the `FD_BMTREE_PREFIX_LEAF` hash domain.
+
+   Then, the hash function is recursively applied over pairs of nodes
+   until there is only one node left (the root).
+
+   `fd_bmtree_32` uses the full SHA-256 digest for each tree node
+   and is thus considered cryptographically secure.
+
+   `fd_bmtree_20` uses SHA-256 digests truncated to 160 bits.
+
+   **Inclusion Proofs**
+
+   Inclusion proofs are used to verify whether a set of leaf nodes is
+   part of a commitment (identified by the root node).
+
+   At a high level, inclusion proofs present a sequence of hash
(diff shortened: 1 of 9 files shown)


jumpsiegel (dismissed):

ripatel-fd:
Please do not merge, there are some review concerns flagged by Philip in Gerrit that have not been addressed yet.

src/ballet/bmtree/fd_bmtree.h
```diff
@@ -0,0 +1,284 @@
...
+
+   Tree topology has the following constraints:
+
+    - All leaf nodes are in the bottom level
+
+    - If a given layer `L`
+      with number of nodes `L_n` ...
+
+      ... has exactly one node,
+          this one node contains the root node
```
ptaffet-jump, line 55: `contains` -> `is`, right?

src/ballet/bmtree/fd_bmtree.h
```diff
@@ -0,0 +1,284 @@
...
+          and is the uppermost layer.
+
+      ... has more than one node
+          ... and `L_n % 2 == 0`,
+              the layer above contains n/2 nodes
+
+          ... and `L_n % 2 == 1`,
+              the layer above contains (n/2)+1 nodes.
```
ptaffet-jump, line 63: n is odd, and n/2 written in a mathematical context like this makes me think of normal (non-integer division). What about `contains (n+1)/2 nodes`.

src/ballet/bmtree/fd_bmtree.h
```diff
@@ -0,0 +1,284 @@
...
+
```
ptaffet-jump, line 64: Can you add something like this:

You can imagine the process for generating the tree with n leaf nodes being the following:
Start with the smallest complete binary tree that has at least n leaf nodes. Label the leaf nodes from left to right L0, L1, ... L(n-1). Delete any un-labeled leaf nodes, and then recursively delete any nodes with no children. For any nodes with a single (un-deleted) child, duplicate the link to the child.
Each non-leaf node now has exactly two children, counting duplicates.

src/ballet/bmtree/fd_bmtree.h
```diff
@@ -0,0 +1,284 @@
...
+          /  \     \\
+         /    \     \\
+       Iα      Iβ    Iγ
+      /  \    /  \   ||
+     L0  L1  L2  L3  L4
+
+   **Construction**
+
+   First, each input in the vector (an arbitrary-sized binary blob)
+   is hashed to leaf node.
```
ptaffet-jump, line 93: "hashed to" doesn't make much sense to me. How about something like:
` First, the value of the i^th leaf node is set to the hash of the i^th input (an arbitrary-sized binary blob) in the vector for each i.`

src/ballet/bmtree/fd_bmtree.h
```diff
@@ -0,0 +1,284 @@
...
+
+   During the accumulation phase, the data structure consumes all
+   tree leaf nodes sequentially while calculating and buffering
+   branch nodes of upper layers along the way.
+
+   In the finalization phase, the buffered branch node data is hashed
+   to derive the final root hash.
+
+   The separation of the accumulation and finalization phases is
+   required for trees with leaf counts that are not in `{x | 2^x}`.
```
ptaffet-jump, line 146: If I'm understanding it properly, I think your set builder notation is backwards, but I think the English "powers of two" is more clear anyways. I read what you have as "x where 2^x," which doesn't make much sense.

src/ballet/bmtree/fd_bmtree.h
```diff
@@ -0,0 +1,284 @@
...
+   - `node_buf` is indexed by layer, with 0 being the leaf layer.
+
+     It buffers the left node of the most recently seen pair of nodes
+     in that layer.
+
+     In other words:
+
+     Given a layer `L` containing a vector of nodes known so far,
+     the subset `L_left` contains all nodes with index `i` within
+     that layer where `i%2==0`.
```
ptaffet-jump, line 186: I initially found this confusing, but then after I understood it, I came up with basically the same explanation you gave.
I tried again and came up with:
"node_buf[L] contains the right-most node in layer L (counting from the bottom) that is a left child of its parent."
Then you can add "More precisely, " and your existing explanation.

src/ballet/bmtree/fd_bmtree.h
```diff
@@ -0,0 +1,284 @@
...
+     - node_buf[0] <- l_0
+     - leaf_idx    <- 1
+
+    Insert leaf `l_1`
+     - b_0         <- hash_branch( node_buf[0], l_1 )
+     - node_buf[1] <- b_0
+     - leaf_idx    <- 2
+
+    Insert leaf `l_2`
+     - node_buf[0] <- l_0
```
ptaffet-jump, line 208: `     - node_buf[0] <- l_2`

src/ballet/bmtree/fd_bmtree.inc
```diff
@@ -0,0 +1,153 @@
...
+   for each layer is copied to `state->node_buf`.
+
+   This serves to allow the algorithm to derive a new parent branch node
+   for any pair of children, once the (previously missing) right node
+   becomes available. */
+static void
+BMTREE_(commit_insert)( BMTREE_COMMIT *     state,
+                        BMTREE_NODE const * leaf ) {
+  BMTREE_NODE tmp;
+  fd_memcpy( &tmp, leaf, sizeof(tmp) );
```
ptaffet-jump, line 56: I prefer using sizeof() on types instead of variables where possible. This occurs in a few places.

src/ballet/bmtree/fd_bmtree.inc
```diff
@@ -0,0 +1,153 @@
...
+     Each iteration, merges that pair of nodes into a new branch node.
+     Terminates if the new branch node is the left node of a pair. */
+
+  ulong layer, cursor;
+  for(
+    /* `layer` starts at 0 (leaf nodes) and increments each iteration.
+       `cursor` is the number of known nodes in the current layer. */
+    layer=0UL, cursor=++state->leaf_idx;
+    /* Continue while the right node in the last pair is available.
+       Loop always terminates as `cursor` eventually reaches `1`. */
```
ptaffet-jump, line 73: I read "eventually reaches 1" as eventually it is true that cursor==1, which I think is false. If cursor starts at 6, it'll end at 3.

``Loop always terminates as `cursor` is always strictly positive and thus has at least one bit set to `1`. */``

src/ballet/bmtree/fd_bmtree.inc
```diff
@@ -0,0 +1,153 @@
...
+  uchar hash[FD_SHA256_HASH_SZ] __attribute__((aligned(FD_SHA256_ALIGN)));
+  fd_sha256_fini( &sha, hash );
+
+  /* Write tree node into out, discarding last few bytes. */
+  fd_memcpy( out, hash, BMTREE_NODE_SZ );
+}
+
+/* fd_bmtree*_commit_insert: Accumulates a single leaf node into the tree.
+
+   Updates `state->leaf_idx++`, U.B. if value >= `state->leaf_cnt`.
```
ptaffet-jump, line 44: `value` means `state->leaf_idx`?

src/ballet/bmtree/fd_bmtree.h
```diff
@@ -0,0 +1,284 @@
...
+/* FD_BMTREE32_NODE_SZ: Size in bytes of a SHA-256 tree node */
+#define FD_BMTREE32_NODE_SZ (32UL)
+
+/* fd_bmtree20_node_t is the hash of a SHA-256-160 tree node (20 bytes) */
+typedef uchar fd_bmtree20_node_t[FD_BMTREE20_NODE_SZ];
+
+/* fd_bmtree20_node_t is the hash of a SHA-256 tree node (32 bytes) */
+typedef uchar fd_bmtree32_node_t[FD_BMTREE32_NODE_SZ];
+
+/* fd_bmtree*_commit_t computes the root of a binary Merkle tree.
```
ptaffet-jump, line 134: I'm confused by a data structure being described by a verb. How about this?

`/* fd_bmtree*_commit_t stores intermediate state used to compute the root of a binary Merkle tree built incrementally.`

ripatel-fd:
@ptaffet-jump Thanks a lot for the thorough review. I think this is soon ready to be merged, could you have a final look?
cc @kbowers-jump

ptaffet-jump (approved):

PR #53, opened 2023-01-23
sha256: add test program and CAVP test vectors

ripatel-fd:
Imported from https://git.firedancer.io/c/firedancer/+/287


ripatel-fd:
Closing, might revisit this in the future. For now, this is low-priority.

PR #54, opened 2023-01-23
sha256: add generated code from sha256-mb-x86_64.pl

ripatel-fd:
Imported from https://git.firedancer.io/c/firedancer/+/302


ripatel-fd:
Closing, might revisit this in the future. At the moment, we don't have the required scheduling to use a multi-block SHA-256 API.

PR #56, opened 2023-01-23
make: gracefully handle missing headers in depfile

ripatel-fd:
When switching branches via Make, stale depfiles sometimes depend on headers that do not exist in the current tree.

This causes Make to error out early, requiring a full rebuild.

Updates the depfile rule to use compiler flag `-MP` to generate dummy targets that allow Make to continue if a header file is missing.

Change-Id: I68eb55feb3ca21c44d9b9b638787f48be72420c7

Imported from https://git.firedancer.io/c/firedancer/+/461


ripatel-fd:
Merging as this was approved in Gerrit.

Issue #37, opened 2022-12-29
macOS support


ripatel-fd:
**Problem**

The workspace / tile system uses various Linux-only APIs. Because of this, Firedancer fails to build on macOS.
There are lots of macOS developers who might want to use Firedancer APIs, or even run a macOS validator (although that is not recommended)

**Suggested fix**

However, the Darwin kernel offers replacements for most of these APIs (e.g. CPU pinning and huge pages).
They are just largely undocumented and annoying to work with.

shuimuliang:
do u need Rust/Solana engineer?
i'm open for Solana project.

smcio:
I can implement macOS support, if you're open to community contributions at this time?

ripatel-fd:
@smcio Thanks, we'd love to merge a contribution for macOS support.

Ideally, changes would be submitted to https://git.firedancer.io but a GitHub PR is fine too.

smcio:
Good stuff - I'll get working on it. Thanks.

ripatel-fd:
@smcio FYI, the NUMA interface is slightly changing: https://git.firedancer.io/c/firedancer/+/413
I also added some stubs for fixing NUMA support on macOS, but haven't tested if that compiles.
Hope that simplifies the integration.

smcio:
Hey @ripatel-jump - should hopefully done with this either today or tomorrow.

ripatel-fd:
Thanks a lot! Also, if you have any questions, you can reach us via the GitHub Discussions tab or via e-mail.

smcio:
Hi @ripatel-jump , I've made the changes necessary to get firedancer built on macOS, including replacing use of Linux APIs with Darwin equivalents, and I'd like to get it running to see whether it's working properly. 

It'd be great to ask a few questions to you and the team. I can't see a "Discussions" tab in my GitHub interface unfortunately. Does the team perhaps have a Discord? I can see a channel for Firedancer on the Solana discord server, but I didn't see much recent activity in there and of course I wouldn't have the right permissions to write in there.

CantelopePeel:
If you could share a Discord handle we can add you. Or message me at CantelopePeel#0343.

CantelopePeel:
_but_ @ripatel-jump would you kindly enable the discussions tab?

ripatel-fd:
[...]

PR #57, opened 2023-01-23
shmem: replace libnuma

ripatel-fd:
Adds //src/util/numa to replace libnuma.
This removes licensing restrictions and reduces dependencies.

- Removes rules_foreign_cc from Bazel
- Removes libnuma dep and `-lnuma` linkage
- Vendor system call interface from `numaif.h`
- Adds wrappers for Linux kernel syscalls
- Replaces libnuma functions via Linux sysfs/sysinfo APIs

Imported from https://git.firedancer.io/c/firedancer/+/413


ripatel-fd:
Comment from kbowers

--------

I love killing of the dependency on libnuma.  There's a lot of nuance from how to do it (especially as this would impact internal stuff ... generally speaking, changes to util imply internal changes too).  Debating if the best approach is to just accept the change and do a bunch of belt sanding or if we should update the patch list.  (The changes are file motion, naming conventions, etc.)

I think the big structural organizational thing is that the libnuma was already wrapped for developer consumption by shmem.  Specifically:

```
/* Numa topology API */

/* fd_shmem_{numa,cpu}_cnt returns the number of numa nodes / logical
   cpus configured in system.  numa nodes are indexed in
   [0,fd_shmem_numa_cnt()) where fd_shmem_numa_cnt() is in
   [1,FD_SHMEM_NUMA_MAX] and simiarly for logical cpus.  This value is
   determined at thread group boot.  cpu_cnt>=numa_cnt. */

FD_FN_PURE ulong fd_shmem_numa_cnt( void );
FD_FN_PURE ulong fd_shmem_cpu_cnt ( void );

/* fd_shmem_numa_idx returns the closest numa node to the given logical
   cpu_idx.  Given a cpu_idx in [0,fd_shmem_cpu_cnt()), returns a value
   in [0,fd_shmem_numa_cnt()).  Returns ULONG_MAX otherwise.  The cpu ->
   numa mapping is determined at thread group boot. */

FD_FN_PURE ulong fd_shmem_numa_idx( ulong cpu_idx );

/* fd_shmem_cpu_idx returns the smallest cpu_idx of a cpu close to
   numa_idx.  Given a numa_idx in [0,fd_shmem_numa_cnt()), returns a
   value in [0,fd_shmem_cpu_cnt()).  Returns ULONG_MAX otherwise.  The
   numa -> cpu mapping is determined at thread group boot. */

FD_FN_PURE ulong fd_shmem_cpu_idx( ulong numa_idx );
```

So having a numa module is redundant with this.  Essentially, feels like numa module is actually just a bunch of things for the private use of shmem to implement the above.

Would prefer to above having multiple (and potentially subtly different) APIs for getting to the same info.

If there was case for having separate numa lib (there might be), some of the functions probably should have their names tweaked and what not to use same conventions as the rest of util.  (There is also some missing docs and PROTOTYPES and what not.)

.inc is also not a file extension convention that is currently used (open to doing it but we should be consistent across the code base so need to decide if this is the way we want to do this).  There are probably some cleaner ways to do it with the build system (see recent tweaks to "tile" for an example of doing the switch in the build system as opposed to in the compile unit).

(There's actually something tangentially related to how with-asan.mk and FD_HAS_ASAN is getting introduced into the build environment.)

(For the with-asan.mk, would have a FD_HAS_ASAN:=1 in the make environment to facilitate build selection, would also have FD_HAS_ASAN be explicit defined in the with-asan.  The testing in ASAN should probably be wrapped with an ifndef FD_HAS_ASAN.  There is possibly an implicit fragment ordering constraint with with-optimiation.mk or with-x86-64.mk implied around the use of -fno-omit-frame-pointer that probably needs to be documented.)

In 413 patch, the no empty LD_FLAGS+= in the fragments isn't doing anything can be blown away (minor nit pick).

The upshot of the above though is the patchset currently shows that it is possible and straightforward to remove the libnuma dependency (and that's pretty awesome).  But large amounts of file motion and what not makes me wonder if the best path is to accept the patch and then do a second patch with all the belt sanding or to try do it in 413.

[...]

One other thing that just occurred to me is that this doesn't completely eliminate libnuma dependency.

IIRC, the bash script fd_shmem_cfg uses some command line stuff to get the topology (hwctl stuff).  It'd be straightforward to fix.

src/util/shmem/fd_shmem_private.h
```diff
@@ -62,6 +62,98 @@
...
+
+   Note that 1 can be returned on systems with only one NUMA node,
+   or even operating systems that do not support NUMA
+   (for compatibility). */
+int
+fd_numa_available( void );
+
+/* fd_shmem_numa_cnt_private: Get number of configured NUMA nodes.
+
+   Returns a libc -errno on failure. (`x<=0`) */
```
ptaffet-jump, line 129: Still a little confusing what x==0 means

ptaffet-jump (approved):
I reviewed an earlier version of this more carefully, but this looks good to me. You should probably wait for Kevin's feedback before merging.

kbowers-jump:
So, was going through various PR and what not and trying to figure out what are the next steps for this draft.

This was something that was in-flight around the transition to github and is pretty tricky change to a pretty subtle area of the code.  And it is now increasingly out of date with main (probably mostly superficially though).  And there are some minor cleanups and what not still pending.

But I also love getting rid of unnecessary dependencies.  So definitely want to see something like this done eventually.

My current inclinations are either:

- Close this PR, create a new branch from current main for libnuma dependency elimination, do a minimalist elimination of libnuma for x86_64 builds builds using the ala the relevant changes from this PR to get all the internal and exposed APIs sorted, documentation updated and what not and get that merged.  Then a second PR to introduce the Darwin shim (building on top of the foundation set by the first PR as it is a somewhat orthogonal set of changes even if the Darwin support was the motivation for the first change).

- Rebase and merge this PR more or less as is and then get out of the belt sander to do the outstanding polishing / file motions / renames in a second PR.  A bit uglier in main's history but would get us to the same place.

Thoughts?

ripatel-fd:
@kbowers-jump I like option 1 because it is the most time-efficient for everyone involved. I'd prefer merging #37 first because it gives us actual compatibility requirements that this PR will have to work around. With that, closing this PR and putting up a new one shortly.

PR #60, opened 2023-01-23
pack: add CU estimation table

ripatel-fd:
Imported from https://git.firedancer.io/c/firedancer/+/308


ripatel-fd:
@ptaffet-jump suggest you reopen this one from your user to simplify tracking authorship

PR #61, opened 2023-01-23
txn: update max instruction count

ripatel-fd:
Imported from https://git.firedancer.io/c/firedancer/+/465


ripatel-fd:
@ptaffet-jump suggest you reopen this one from your user to simplify tracking authorship

PR #68, opened 2023-01-26
ci: add codecov-action

ripatel-fd:
Modifies bazel_test CI pipeline to upload coverage report to CodeCov


jumpsiegel (approved):

ripatel-fd:
@mtjmp Can you look into enabling CodeCov so we can benefit from these reports?

PR #69, opened 2023-01-26
ci: run checks on pull requests

ripatel-fd:
Fixes CI trigger rules to run on all pull requests and all pushes to main.


mtjmp (approved):
LGTM

PR #89, opened 2023-02-01
fd_util_base: add fd_offsetof macro

ripatel-fd:


ripatel-fd:
Closing this, better to just use the `<stddef.h>` version.

PR #91, opened 2023-02-01
Belting sanding pass on the fd_ar API.

ripatel-fd:
Opening PR from @kbowers-jump 

-------------------

This started as a quick review of the code but ended up more extensive polishing:

- Eliminated a number of security issues due to underspecification of AR file parsing for meta data (e.g. all whitespace fields in a header causing strtol to run off the end, using negative values for things like file size to trigger weird behavior like infinite loops in archive file iterators).

- Generalized API to take error prone parsing (due to the above) out of user hands and support extending this to obvious variations on the ar file form without requiring users update their code.

- Generalized API to support non-hosted platforms (no need for stdio.h or stdlib.h or FILE abstractions to be a user of the API).

- Consistent handling of EOF condintions between open and init.

- Extensive documentation of the above.

- Other minor stylistic tweaks and cleanups.

### DIFF
--- a/src/util/archive/fd_ar.c
+++ b/src/util/archive/fd_ar.c
@@ -1,47 +1,152 @@
 #include "fd_ar.h"
 
+#if FD_HAS_HOSTED
+
+#include <stdio.h>
+#include <stdlib.h>
 #include <errno.h>
-#include <string.h>
 
-static const char magic_expected[ 8 ] =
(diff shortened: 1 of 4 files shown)


ptaffet-jump (approved):
I made some minor changes to the comment about how numbers are parsed, but otherwise looks great!

daviesalex (approved):

PR #103, opened 2023-02-03
txn: add fd_txn_get_signatures

ripatel-jump:
- ed25519: add fd_ed25519_sig_t
- txn: add fd_txn_get_signatures

### DIFF
--- a/src/ballet/txn/fd_txn.h
+++ b/src/ballet/txn/fd_txn.h
@@ -29,6 +29,8 @@
 
 #include "../fd_ballet_base.h"
 
+#include "../ed25519/fd_ed25519.h"
+
 /* FD_TXN_VLEGACY: the initial, pre-V0 transaction format. */
 #define FD_TXN_VLEGACY ((uchar)0xFF)
 /* FD_TXN_V0: The second transaction format.  Includes a version number and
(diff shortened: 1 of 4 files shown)


gnewsome-jump (approved):

src/ballet/txn/fd_txn.h
```diff
@@ -351,6 +353,17 @@ fd_txn_get_address_tables( fd_txn_t * txn ) {
   return (fd_txn_acct_addr_lut_t *)(txn->instr + txn->instr_cnt);
 }
 
+/* fd_txn_get_signatures: Returns the array of Ed25519 signatures in
+   `payload`, the serialization of the transaction described by `txn`.
+   The number of signatures is seen in `txn->signature_cnt`.
+   U.B. If `payload` and `txn` were not arguments to a valid
+   `fd_txn_parse` call or if either was modified after the parse call. */
+static inline fd_ed25519_sig_t const *
```
ptaffet-jump, line 361: Can you document that the pointer is within payload so has the same lifetime as payload. Also, it might be badly aligned.

ptaffet-jump (approved):
Looks good. I suggested one comment. I guess approving is useless because if you make the change it'll discard my approval, but I like the change. We should add something similar for acct_addr

Issue #95, opened 2023-02-02
util/log: hex dump routine


ripatel-fd:
**Problem**

Hex dumping an arbitrary-sized blob is a fairly common operation.

But at the moment, we only have a helper macro for `printf`-style formatting of 16-byte chunks: `FD_LOG_HEX16_FMT`.

**Suggested Fix**

Introduce an API for writing `hexdump` / `xxd` style dumps to fd_util log.

smcio:
I can take this one. I'll put up a PR for this shortly.

(I'm assuming a community contribution is OK given the 'good first issue' label 🙂 )

ripatel-fd:
Closing this issue since this has been implemented. Thanks a lot for this very useful contribution, @smcio!

PR #106, opened 2023-02-04
Hex dump log utility API

smcio:
Hex dump logging API as per https://github.com/firedancer-io/firedancer/issues/95 for review.


src/util/log/fd_log.c
```diff
@@ -493,7 +493,8 @@ fd_log_private_1( int          level,
     /* 4 */ "ERR    ",
     /* 5 */ "CRIT   ",
     /* 6 */ "ALERT  ",
-    /* 7 */ "EMERG  "
+    /* 7 */ "EMERG  ",
+    /* 8 */ "HEXDUMP"
```
kbowers-jump, line 497: No object to notion of a HEXDUMP but wouldn't do it as log level.  Log levels mirror the syslog log levels and trigger various behaviors based on severity (like flushing to stderr / logfile, aborting program and what not).  So a HEXDUMP at level 8 would be above an emergency from a logging POV.

Ideally should be possible to do a HEXDUMP at an INFO or NOTICE or what not level.

src/util/log/fd_log.c
```diff
@@ -628,6 +629,68 @@ fd_log_private_2( int          level,
...
+void 
+fd_log_hexdump ( const char * descr,
+                 const void * blob,
+                 const size_t len ) {
+    
+  size_t count, blob_len;
+  int is_truncated = 0, len_written;
+  char line_buf[ FD_LOG_HEXDUMP_BYTES_PER_LINE+1 ];
+  const char * blob_ptr = (const char *)blob;
+
```
kbowers-jump, line 641: There are some notions being kicked around for collecting various discussions and comments throughout the headers into one place as a style guide / common practices / etc but in the meantime, some minor pedantic nit picks (not everything applies to to this code but is closely related and thus might be useful):

- The linguistic machine model guarantees by the development environment guarantees that the target platform will have `ulong` is interchangeable with `size_t` and `size_t` has been a 2nd class citizen in the strict C/C++ specification historically (just a mess in the standards to how it evolved).  So `ulong` is generally preferred over `size_t`.

- Some of this might be aspirational but we try just have one variable in a declaration statement, try to use the suffix `_len` on a variable when it refers the length of c-style string len in a `strlen`-sense, try to use the suffix `_sz` to refer to the actual number of contiguous bytes used in the memory (e.g. `ulong cstr_len = strlen( cstr ); ulong cstr_sz = cstr_len + 1UL;`), try to use the suffix `_cnt` when a variable is a natural number count of something and try to use the suffix `_idx` when a variable is natural number index of something in the half open interval of the corresponding `_cnt` (e.g. `ulong foo_cnt; ulong foo_idx; /* in [0,foo_cnt) */`).  While in a case like this might seem superfluous, it is a bit of a habit to reduce risk of making off-by-one / overflow / underflow / ... errors in more complex cases, increase the chance reviews will catch such errors, make room for commenting domains of individual variables and what not.

- And while not an issue here, note that we also try to use `ulong` for the in-register representation of such.  This is to avoid unpleasant surprises in the language itself around UB/IB behaviors with signed types and narrow types (this area of the C/C++ language specs best described as a superfund site), avoid unpleasant surprises when code gets used in the real world on larger data sets that have been tested in the lab (lost too many weekends debugging some code because there was an `int` iterator hidden somewhere deep in the bowels of 3rd party dependency when a user fired up the code a large data for the first time), to reflect how 64-bit CPUs work under the hood and thus improve quality of compiler code gen, etc.

- The precedence rules in C/C++ for modifiers in variable declarations tend to "just work" naturally if read right-to-left unfortunately (another superfund site in the language).  So another habit we try to do that doesn't make much a difference here but can matter in more nuanced situations with restricted pointers, volatile pointers, etc is to write variable declarations in a way that reads naturally right to left.  E.g. "unmodifiable pointer to a unpredictable pointer to an unpredictable but unmodifiable character" would be a `char const volatile * volatile * const` ... tangentially related, we aspire to avoid writing such types when possible because they aren't exactly the most readable).

- Also not an issue here but the C/C++ language is also a disaster area around the sign of `char` (language leaves it IB and different architectures use different behaviors and the standard library and implementations are a bit of free for all to try to get consistent cross platform behavior around something as elementary as `strcmp` ... look at what the standard says or doesn't and glibc does here for an example).  We try to use `char` for contents of a c-style string and avoid arithmetic on it, `uchar` unsigned 8-bit integral data, `schar` for signed 8-bit integral data and `void *` for pointers to raw binary data.

- We also try to put variable declarations as close to use as possible and then limit the scope of variable declarations to be as narrow as possible.

Combining all this (and the above comment), a possible function call signature might look like:

```
void
fd_log_hexdump( int          level,
                char const * descr,
                void const * blob,
                ulong        sz ) {
  ... move the closer to use and initialize on declaration if possible ...
  ulong blob_sz;
  ...
}
```

src/util/log/fd_log.c
```diff
@@ -628,6 +629,68 @@ fd_log_private_2( int          level,
...
+  /* Limit input blob length to a fairly small size (8K) to prevent stack overflow in 
+     the alloca below. */
+  if( FD_UNLIKELY ( len>FD_LOG_MAX_HEXDUMP_BLOB_SIZE ) ) {
+    blob_len = FD_LOG_MAX_HEXDUMP_BLOB_SIZE;
+    is_truncated = 1;
+  }
+
+  /* Enough space plus a bit extra for hexdump-alising the blob. */
+  char * log_buf = fd_alloca( alignof(char), ((blob_len*8)+strlen(descr)+32)*sizeof(char) );
```
kbowers-jump, line 652: An interesting option here would be use a bounded thread local variable buffer (ala `fd_log_private_1` using of `FD_TLS` to make code portable to targets without `FD_HAS_ALLOCA`).

Suspect it is not strictly necessary here as this is probably used in a context where `FD_HAS_ALLOCA` is already implied.  But though I love `alloca` (with appropriate safe guards about bounded stack utilization and stack guards and with proper support for alignment as a first class citizen needed as that has been needed for performance on very architecture since the dinosaurs roamed the earth and the languages still haven't caught up ... e.g. unfortunately stock `alloca.h` ... it is essentially a garbage collected allocator with fast O(1) assembly ops count for allocation and free garbage collection and great L1D cache and L1 TLB efficiency ... often unbeatable for small quick temporary allocations), it is widely misunderstood, misused, abused and/or poorly supported historically.  So I try to avoid it if there are other comparably performant and/or readable options.

src/util/log/fd_log.c
smcio: Thanks very much for the review.

Yeap, I hear you. 

I can just make changes such that hex dumps are done at the INFO level, if this sounds good to you. Or, alternatively, I can the ability to log hex dumps at arbitrary log levels, either a) via macros like `FD_LOG_HEXDUMP_INFO`, `FD_LOG_HEXDUMP_NOTICE`, `FD_LOG_HEXDUMP_WARNING`, etc, or b) by passing in a log level parameter like `FD_LOG_HEXDUMP(FD_LOG_LEVEL_INFO, ...)`, or `FD_LOG_HEXDUMP(FD_LOG_LEVEL_WARNING, ...)`.

Please let me know which you prefer 👍

src/util/log/fd_log.c
```diff
@@ -628,6 +629,68 @@ fd_log_private_2( int          level,
...
+    /* New line. Print previous line's ascii representation and then print the offset. */
+    if( !( count%FD_LOG_HEXDUMP_BYTES_PER_LINE) ) {
+      if( count!=0 ) FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "  %s\n", line_buf);
+      FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "  %04x ", (unsigned int)count );
+    }
+
+    FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( " %02x", blob_ptr[count]&0xff);
+
+    /* If not a printable ASCII character, output a dot. */
+    if( isprint( blob_ptr[ count ] ) )
```
kbowers-jump, line 678: Would have to double check but I think `isprint` includes spaces like tabs, CR, LF, form feed, vertical tabs, etc and such can muck up log readability and break assumptions made by downstream automated archivers / consolidators / consumers of log files.  (We tend to be sensitive about logging because subtle issues here around getting complete info in logging alerts readably in front of the right people in a timely fashion can be literally be firm killers ... there are very famous examples of this at former firms in our traditional stomping grounds.)

FWIW, `fd_pod_ctl.c` already has a bunch of stuff for this.  Includes:
```
static inline int
issingleprint( int c ) {
  return isalnum( c ) | ispunct( c ) | (c==' ');
}
```
and a similar format pretty printer to this around line 98 (it would have to be cleaned up a little for the logging to deal with the stderr and log file stuff or this would have to be generalized a little if we want to consolidate).

src/util/log/fd_log.c
kbowers-jump: See comments below which address a couple of this with that (along with my tendency to go into ad hoc rants of about issues brokenness of languages and the pain it cause people like us ... it's not directed at your code).

Option a is perfectly cromulent from my POV (make fd_log_hexdump becomes something akin to fd_log_private_1 stuff like mentioned below to deal with the other linguistic unpleasantnesses the logger has to fix up).  Also see the pretty printer in fd_pod_ctl.c mentioned below.  It does a two column hex dump style already:
```  0123: 00 11 22 33 44 55 66 77  88 99 AA BB CC DD EE FF  thequickbrownfox```
So some consolidation / reuse opportunity might be worthwhile.

Comment (kbowers-jump) [reply src/util/log/fd_log.c:497]: Also related note, if it is ends up making sense to have `fd_log_hexdump` meant for direct use by log users, the behaviors for various permutations of NULL descr, NULL buf and/or 0 sz need to be specified, trapped and/or logged as appropriate.

File: src/util/log/fd_log.c (hunk 3 above)
Comment (smcio) [reply src/util/log/fd_log.c:652]: Great suggestion. I'll make changes here to use TLS via `FD_TLS` (`__thread`).

File: src/util/log/fd_log.c (hunk 4 above)
Comment (smcio) [reply src/util/log/fd_log.c:678]: Yeah, good point. Will make changes here w.r.t. your comments.

Comment (smcio) [reply src/util/log/fd_log.c:678]: Just checked the documentation for `isprint` (such as C17 ISO/IEC 9899:2018, and earlier), and this actually depends upon locale (see https://en.cppreference.com/w/c/string/byte/isprint).

Space characters such as tabs get 0 returned by `isprint` (as below), but given that `isprint`'s result per character depends on locale, it probably is just better to avoid this API and rework this instead.

```
(base) Shauns-MBP-5:~ shauncolley$ cat isprint.c 
#include <stdio.h>
#include <ctype.h>

int main() {

int result = isprint('\t');
printf("isprint returned %d\n", result);
}
(base) Shauns-MBP-5:~ shauncolley$ ./isprint
isprint returned 0
```

File: src/util/log/fd_log.c (hunk 2 above)
Comment (smcio) [reply src/util/log/fd_log.c:641]: Thanks very much for these in-depth guidelines - much appreciated. Solid rules and guidelines. I'll rework according to these.

**Comment by smcio** [comment]
@kbowers-jump @ripatel-jump I've fixed all the aspects outlined in the comments above, but whilst making these changes, something occurred to me. 

As per the `fd_log_private_0` function, the maximum size for a log message is 4096 (incl. terminating NULL byte), and the rest is discarded. After hexdump-alising a blob, if you use the regular FD_LOG* macros (and therefore `fd_log_private_0`) for logging, you'll get a maximum of under 900 hexdumped bytes being logged. 

It occurred to me that 900 bytes is quite far below Solana's MTU (1280 incl. headers, IIRC). As a result, then, if you wanted to use the hexdump logging facility to write an entire Solana packet to the log, the entry would be truncated by `fd_log_private_0`, and this would most probably be undesirable if you want to see the contents of the entire packet.

I feel that I should modify the hexdump flow to allow hexdump log entries to be a bit larger than 4096 bytes. Your thoughts would be greatly appreciated. If you agree, any opinions on a sensible log size limit for hexdump log entries would be great. 🙂 

(In addition, it might also be worth considering that someone may also wish to hexdump log the contents of an account's data segment, for instance. Last time I checked the Solana validator code, this limit was 10MB. Creating log entries that are multiples of 10MB is probably way too large to be sensible, but I'd say allowing hexdumps at MTU size seems sensible)

**Comment by ripatel-fd** [comment]
@smcio Regarding the log splitting issue, I'd be in favor of keeping the 4096 chars limitation. Log aggregation systems out there might have an issue with very large log entries.

A possible solution is to split the hex dump over several log messages: Keep each message to N lines.

Given the following code

```c
char blob[ 0x2000 ] = {0};
fd_log_hexdump( FD_LOGLVL_NOTICE, "Blob content", blob, 0x2000 );
```

The log might look something like:

```
NOTICE  02-06 12:24:10.736016 94931  f0   0    test.c(127): Blob content (0x0..0x80)
00000000: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000010: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000020: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000030: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000040: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000050: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000060: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000070: 0000 0000 0000 0000 0000 0000 0000 0000  ................
NOTICE  02-06 12:24:10.736016 94931  f0   0    test.c(127): Blob content (0x80..0x100)
00000080: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000090: 0000 0000 0000 0000 0000 0000 0000 0000  ................
000000a0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
000000b0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
000000c0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
000000d0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
000000e0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
000000f0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
```

The downside of this approach is that the hex dump is going to be interleaved with messages from unrelated concurrently running operations.

kbowers-jump:
Re 4096, the historical motivation was a combination of not being too large when porting to more embedded like hardware (where memory footprint is at a premium), matching to normal page sizes for TLB efficient handling in the logger (practically O(1) fixed normal page TLB entry per core used by the logger), matching a maximum log message with typical terminal windows and font sizes (which in turn tend to be similar to using readable fonts on letter / A4-ish paper sizes ... roughly 4K-7K characters per message worst case).

But there's nothing magical about and it.

And when moving code to different environments, unbeknownst to the developer, the implementation under the hood is often radically different (e.g. instead of a TLS buffer, it might actually be the payload region of networking packet under construction to allow non-hosted platforms to communicate their logging to various remote command-and-control facilities and thus implicitly bounded by MTU of the networking layer between the machine and the command-and-control ).

With this in mind, the argument that we should be able to do a pretty printed hexdump of a Solana MTU worth in this implementation of the logging API is pretty compelling.

So it doesn't bother me if we update the size accordingly.  Some reasonable small integer multiple of 4KiB feels right for this application (maybe with some debatable additional 4KiB alignment for good TLB mojo).

To avoid excess bloat of the TLS segment used under the hood, it'd be nice for the hexdump stuff and fd_log_private_0 to use the same buffer for this (seems doable and maybe this has already been done ... haven't read the updated code itself yet ... just giving my thoughts about tweaking the behind the scenes buffer limits)

kbowers-jump:
Especially compelling in this suggestion is the part about message interleaving.  Fragmenting a logical message over multiple log messages is strongly discouraged.  (And even more strongly discouraged, never let a logger fragment individual log messages under the hood.)

(This is _so_ _frequently_ _botched_.  Developers who haven't ever had the joy / sorrow to work on high performance distributed systems look things like `util/log` and think "it's just some unnecessary wrapping around `fprintf` to `stderr` / `cerr`" and then they just do that instead.  They missed the point and then discover the hard way what happens to your giant supercomputer / firm / etc when a million cores distributed over a large geographic region all go into a tight polled loop logging some error condition due to a code bug ... the messages are all fragmented to hell, reordered, interleaved and missing context needed by remote operators ... nothing makes any sense, worse the networks and archival file systems are melting down under the load so operators can't regain control, and worst still the archival storage is getting overrun which is really scary from a compliance perspective, etc ... such things are literal firm killers.)

smcio:
Cool, thanks very much folks. 

I've ended up going with the option for increasing the log entry size to allow for blobs to be logged above the size of Solana's MTU.

To do this, I've increased the log buffer to 16k, and this buffer is shared between `fd_log_private_0` and the hexdump-alising function `fd_log_private_hexdump_msg`. The log length limit of 4096 in `fd_log_private_0` is unchanged as per the bounds of the `vsnprintf` call, so 'regular' log messages are still bound by a 4096 byte limit (incl. '\0' terminator).

`fd_log_private_hexdump_msg` allows incoming blob sizes of 1500 (more than Solana MTU), so given the space required to hexdump-alise a blob, plus using a multiple of 4k, I chose 16k for this shared TLS buffer size.

I've also added some input validation checks to log upon receiving bad inputs, such as NULL ptrs and zero-len blob description.

The API for hexdump logging is via macros that are written in a similar style to the regular logging macros (`FD_LOG_INFO`, `FD_LOG_WARNING`, etc.), as per the tests in test_log.c, i.e. `FD_LOG_HEXDUMP_INFO`, `FD_LOG_HEXDUMP_WARNING`, etc:

```
  FD_LOG_HEXDUMP_WARNING(( "test_blob", "aaaa", strlen( "aaaa" ) ));
```

Please let me know if you have any opinions on improvement. Thanks!

kbowers-jump:
This is looking good.  Couple of minor typographical nits here and there (e.g. stray backslash in the header) but nothing we should let delay getting support merged with a bunch more ping-pong-ing.

So, in lieu of ping-pong, I'd be fine doing a quick round of polishing in this branch and assuming unexpected crops up, merging into main.  Or merging and then doing a quick round of polishing in separate PR.

(Mildly biased to the former so keep main's history as clean as possible open to either ... whichever has better github mojo, to which I defer to Richie as his git-fu is much stronger than mine.)

P.S. I'd also probably just increase the vsnprintf limits.  A downstream log aggregrator just wants to know what's the worst case for their resource planning.  So, if we've set aside the memory for it and we've already have larger worst case implied by the hex dump variant, might as well simplify things.

smcio:
Cool, thanks. I saw the stray backslash you were talking about, so I removed that - see latest commits. I also bumped the log entry limit to 16k as you said.

I didn't notice any other obvious typographical errors, but, then, I wouldn't, because I wrote it 🙂  Please let me know if you know of any other errors and I'll get 'em fixed quickly.

src/util/log/fd_log.c
```diff
@@ -628,6 +631,82 @@ fd_log_private_2( int          level,
...
+     Just print 'msg: empty' for a zero length buffer and return,
+     and make it clear when truncation happened if that is the case. */
+  if( FD_LIKELY( blob_sz ) && !is_truncated ) {
+    FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "HEXDUMP %s:\n", descr );
+  } else if( FD_UNLIKELY( is_truncated ) ) {
+    FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "HEXDUMP %s (truncated):\n", descr );
+  } else {
+    FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "HEXDUMP %s: empty\n", descr );
+    return log_msg;
+  }
```
kbowers-jump, line 678: Recommend including the size explicitly here for easy of readability (and to help log readers understand how much info was lost in the case of a truncated hexdump).

src/util/log/fd_log.c
```diff
@@ -56,6 +56,10 @@
 #include <syscall.h>
 #include <execinfo.h>
 
+/* Log buffer, used by both fd_log_private_0 and fd_log_private_hexdump_msg */
+
+static FD_TLS char log_msg[ FD_LOG_BUF_SZ ];
```
kbowers-jump, line 61: (Ultra pedantic ... probably would name this with a fd_log_private_log_msg to make symbol table dumps easier to read / grep / etc.)

src/util/log/fd_log.h
```diff
@@ -142,6 +142,8 @@
 
 #include "../env/fd_env.h"
 
+#define FD_LOG_BUF_SZ               4*4096
```
kbowers-jump, line 145: This probably belongs in the log implementation itself (as per previous discussions, other implementations of this API might use different sizes and what not ... so details like this probably shouldn't be exposed or should be exposed in via an explicit API to query the implementation).


Also, to avoid Tom-Foolery with macros and token pasting and silent conversions), this should be something like `#define FD_LOG_BUF_SZ (4UL*4096UL)`.

src/util/log/fd_log.h
```diff
@@ -155,6 +157,55 @@
...
 #define FD_LOG_EMERG(a)   do { long _fd_log_msg_now = fd_log_wallclock(); fd_log_private_2( 7, _fd_log_msg_now, __FILE__, __LINE__, __func__, fd_log_private_0 a ); } while(0)
 
+#define FD_LOG_HEXDUMP_BYTES_PER_LINE             16
+#define FD_LOG_HEXDUMP_BLOB_DESCRIPTION_MAX_LEN   32
+#define FD_LOG_HEXDUMP_MAX_INPUT_BLOB_SZ          1500
+
+#define FD_LOG_HEXDUMP_ADD_TO_LOG_BUF(...)  do {                              \
+    num_bytes_written = sprintf ( log_buf_ptr, __VA_ARGS__ );                 \
+    if( FD_LIKELY( num_bytes_written>=0 )) log_buf_ptr += num_bytes_written;  \
+ } while(0)
```
kbowers-jump, line 167: Same comment for 160-164 as for 145.

src/util/log/fd_log.h
```diff
@@ -155,6 +157,55 @@
...
+                                             
+#define FD_LOG_HEXDUMP_DEBUG(a)   do { long _fd_log_msg_now = fd_log_wallclock();           \
+    char const * msg = fd_log_private_hexdump_msg a;                                        \
+    if( msg ) fd_log_private_1( 0, _fd_log_msg_now, __FILE__, __LINE__, __func__, msg );    \
+    } while(0)
```
kbowers-jump, line 172: So, it looks like the hexdump_msg has the property that calls with edge cases return NULL and a LOG_WARNING is always done in such cases.  This could potentially result in a log message at one level having a different apparent priority than the user expected and this is a potential source of nasty.

On the mild end of the spectrum, some code had a FD_LOG_INFO and was working was fine, LOG implementation was tweaked under the hood to use different limits description limits or what not, FD_LOG_INFO code now actually triggers a FD_LOG_WARNING and suddenly no info present.

Or the uh-oh end of the spectrum, some code does:
```
if( ... detected malicious behavior ... )
  FD_LOG_HEXDUMP_ABORT(( "malicious_packet", pkt_payload, pkt_payload_sz ));
```
Now if there was some innocuous code change that caused HEXDUMP to be unhappy, the ABORT would be demoted to WANRING such that the malicious packet details would not be logged (just a note that HEXDUMP wasn't happy at warning level) and the code would continue processing the malicious packet as though it was not malicious.

Talking this through, pretty clear that the user's log warning level is sacrosanct.

Probably should just have `fd_log_private_hexdump_msg` write the edge case message directly in `log_msg` and return it.  This would have the strict property that it would always return non-NULL and could be compositionally fed directly into `fd_log_private_1` (like the above macros) and the only risk is some loss of detail in the log messages in situations like the uh-oh case above.

Similarly for the macros below.

src/util/log/fd_log.c
```diff
@@ -628,6 +631,82 @@ fd_log_private_2( int          level,
...
+
+  if( FD_UNLIKELY( strlen( descr )==0 ) ) {
+    FD_LOG_WARNING(( "Hexdump empty blob description" ));
+    return NULL;
+  }
+
+  if( FD_UNLIKELY( strlen( descr )>FD_LOG_HEXDUMP_BLOB_DESCRIPTION_MAX_LEN ) ) {
+    FD_LOG_WARNING(( "Hexdump blob description too long" ));
+    return NULL;
+  }
```
kbowers-jump, line 656: See comment around `fd_log.h:172`.  One simple mod would be to do `return strcpy( log_msg, "Hexdump ..." );` in lieu of the log warning.

A more elaborate tweak would be allow for NULL descr and/or truncated descr (pretty printing accordingly) and then logging the case of a NULL blob with a non-zero sz with appropriate error message.

The case of NULL blob with zero sz probably shouldn't be treated as an error case.

kbowers-jump:
Apologies for the ping-pong-ing; overall, this is looking quite good.

(The above nit-picking is very much something I'd normally do in a quick pass as part of the review.  But, as shown by the examples mentioned in earlier comments, we are really sensitive about logging.  Loggers are the computer science equivalent of an airplane black-box: they need to record as best possible all that is going in situations where everything is going wrong and are critical to dev and ops to diagnosing and fixing problems in the wild as rapidly as possible.  So hopefully there is some value in the more detailed rationales.)

smcio:
@kbowers-jump  Yeap, agreed on all your comments - they all make perfect sense. I'll make all the appropriate changes in the morning and push 'em up for a last skim 🙂  Thanks for the detailed review.

src/util/log/fd_log.c
```diff
@@ -464,17 +464,109 @@ void fd_log_level_stderr_set ( int level ) { FD_VOLATILE( fd_log_private_level_s
...
+
+/* Log buffer, used by fd_log_private_0 and fd_log_private_hexdump_msg */
+
+static FD_TLS char fd_log_private_log_msg[ FD_LOG_BUF_SZ ];
+
+/* Add into the log buffer (with *printf() formatting) at the current offset and then increment 
+   the pointer accordingly */
+
+#define FD_LOG_HEXDUMP_ADD_TO_LOG_BUF(...)  do {                              \
+    num_bytes_written = sprintf ( log_buf_ptr, __VA_ARGS__ );                 \
```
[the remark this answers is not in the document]
kbowers-jump: Doesn't bother me too much in that the usage should already bound this.  I would put this macro inside the hexdump function itself though (and undef it at the end) as such is the only place it is used and it implicitly assumes the log_buf_ptr variable that is declared there (with the same scope).  That would then also prevent any use cases where the sprintf use wasn't already bounded.

src/util/log/fd_log.h
```diff
@@ -157,53 +155,40 @@
...
-#define FD_LOG_HEXDUMP_BLOB_DESCRIPTION_MAX_LEN   32
-#define FD_LOG_HEXDUMP_MAX_INPUT_BLOB_SZ          1500
-
-#define FD_LOG_HEXDUMP_ADD_TO_LOG_BUF(...)  do {                              \
-    num_bytes_written = sprintf ( log_buf_ptr, __VA_ARGS__ );                 \
-    if( FD_LIKELY( num_bytes_written>=0 )) log_buf_ptr += num_bytes_written;  \
- } while(0)
+#define FD_LOG_HEXDUMP_BYTES_PER_LINE             16UL
+#define FD_LOG_HEXDUMP_BLOB_DESCRIPTION_MAX_LEN   32UL
+#define FD_LOG_HEXDUMP_MAX_INPUT_BLOB_SZ          1500UL
```
kbowers-jump, line 160: Missed moving these to fd_log_private_hexdump too (and, ultra nit picky, would add parens around the values).

kbowers-jump:
Overall, looking good.  FWIW, a lot of the above is getting into excessively picky territory and things that we'd naturally do in the final polish as part of the review instead of ping-ponging (I'm increasingly feeling like Columbo ... catchphrase: "Just one more thing").  But we are still sorting how to go about this with external PRs and we appreciate your contribution here.

src/util/log/fd_log.c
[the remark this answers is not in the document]
smcio: @ripatel-jump I totally understand your preference for `snprintf` - I'm security paranoid myself, being a security guy as my day job, principally in C and C++ code review and blackbox security review (reverse engineering). The reason I chose to go with the `sprintf` API in this context, however, was because a) we can be sure due to the boundary checks (and overall usage context) in `fd_log_private_hexdump_msg` that we'll never extend beyond the bounds of `fd_log_private_log_msg` and corrupt memory, and b) `snprintf` returns the number of characters that *would've* be written if the size were unlimited, rather than the number of characters actually written as `sprintf` does.

If it's satisfactory to you guys, then, I'll simply limit the scope of usage for `FD_LOG_HEXDUMP_ADD_TO_LOG_BUF` by `#define`'ing and `#undef`'ing within `fd_log_private_hexdump_msg`, to be sure that no one ever uses this macro outside of the intended use case.

src/util/log/fd_log.c
```diff
@@ -493,7 +585,7 @@ fd_log_private_1( int          level,
     /* 4 */ "ERR    ",
     /* 5 */ "CRIT   ",
     /* 6 */ "ALERT  ",
-    /* 7 */ "EMERG  "
+    /* 7 */ "EMERG  ",
```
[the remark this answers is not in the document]
smcio: Done, thanks!

src/util/log/fd_log.h
smcio: Thanks, done. Moved them to be `#define`'d and `#undef`'d within the hexdump msg function.

.gitignore
```diff
@@ -9,5 +9,8 @@ bazel-firedancer
 bazel-out
 bazel-testlogs
 
+# macOS specific
+.DS_Store
```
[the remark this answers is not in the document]
smcio: Done - I removed this change. Thanks very much btw - my github skills are fairly weak compared to some folks, so I appreciate the instructions here.

smcio:
Hopefully this should be OK per your review comments now folks. Thanks very much for all the review here.

No problem with nits at all folks - I completely appreciate the policy to build out and maintain your codebase per high quality standards, hence nits are totally appropriate in my eyes! 🙂  I appreciate the opportunity to contribute, so thanks again.

kbowers-jump:
Checked it out locally, fixed some stuff that was causing it not to compile for me and then did a final polish.  Since this is an external PR, as per previous discussion, I don't seem have the ability to push the polishing back to your branch (don't know if that is a permissions on this side or your side or what ever).  So in lieu of that:

Fixes:

- The checkout from seemed to have two copies of the hexdump_log_msg function ... probably an oversight from some of the code motion (blew away the second one).

- Compiler complained about a comparison between signed and unsigned types in the fd_log_private_0 that was trivially fixable.
```
-  if( len>FD_LOG_BUF_SZ-1 ) len = FD_LOG_BUF_SZ-1; /* cmov */
+  if( len>(int)FD_LOG_BUF_SZ-1 ) len = (int)FD_LOG_BUF_SZ-1; /* cmov */
```

Polishing:

This was mostly a quick pass on the output formatting to minimize information loss and improve the ability for automated systems to be able to understand the messages and recover as bit level accurate as possible the details post mortem (and also have a detailed understanding what was omitted).

This includes allowing NULL descriptions, zero length descriptions, and overlength descriptions to be printed with as much programmatically disambiguated detail as possible.  Specifically NULL description will be `-`, an empty description will be `""`, a normal descr will be `"descr"` and a too long truncated description will be printed as `"too_long_truncated_des"...`.  Size will be printed as is and the value of the pointer at run-time is also printed (such can be especially handy in situations with debugging memory mapped I/O, shared memory, etc).  Likewise, edge cases for descr don't inhibit printing of the actual memory region. 
 The remaining edge case of NULL blob with a non-zero sz were similarly regularized.

More details on the truncation itself were printed as well and moved to the end of the message so it is easier for a reader to see in a long message that they didn't get everything and how much was omitted.  There were a couple of minor tweaks on how continuation lines were printed (these historically have a tab as the first character to indicate visually and programmatically the line is a continuation of a log message from a previous line).

The upshot of all this is the unit tests you added to `test_log.c` print to the ephemeral log something like:
```
... SNIP FOR GITHUB ...
NOTICE  02-10 12:53:08.798424 21251  f0   0    src/util/log/test_log.c(39): fd_log_user          kbowers
NOTICE  02-10 12:53:08.798433 21251  f0   0    src/util/log/test_log.c(42): Testing hexdump logging API: 
WARNING 02-10 12:53:08.798445 21251  f0   0    src/util/log/test_log.c(45): HEXDUMP "very_small_blob" (1 bytes at 0x40d0f7)
        0000:  61                                               a
WARNING 02-10 12:53:08.798835 21251  f0   0    src/util/log/test_log.c(48): HEXDUMP "test_blob" (4 bytes at 0x40d0f4)
        0000:  61 61 61 61                                      aaaa
WARNING 02-10 12:53:08.799190 21251  f0   0    src/util/log/test_log.c(51): HEXDUMP "alphabet_blob" (26 bytes at 0x40d107)
        0000:  61 62 63 64 65 66 67 68 69 6a 6b 6c 6d 6e 6f 70  abcdefghijklmnop
        0010:  71 72 73 74 75 76 77 78 79 7a                    qrstuvwxyz
NOTICE  02-10 12:53:08.799567 21251  f0   0    src/util/log/test_log.c(54): HEXDUMP "empty_blob" (0 bytes at 0x40d107)
WARNING 02-10 12:53:08.799581 21251  f0   0    src/util/log/test_log.c(57): HEXDUMP "" (26 bytes at 0x40d107)
        0000:  61 62 63 64 65 66 67 68 69 6a 6b 6c 6d 6e 6f 70  abcdefghijklmnop
        0010:  71 72 73 74 75 76 77 78 79 7a                    qrstuvwxyz
WARNING 02-10 12:53:08.799966 21251  f0   0    src/util/log/test_log.c(60): HEXDUMP "blob_name" (26 bytes at 0x0)
        ... snip (unreadable memory) ...
WARNING 02-10 12:53:08.800337 21251  f0   0    src/util/log/test_log.c(63): HEXDUMP - (26 bytes at 0x40d107)
        0000:  61 62 63 64 65 66 67 68 69 6a 6b 6c 6d 6e 6f 70  abcdefghijklmnop
        0010:  71 72 73 74 75 76 77 78 79 7a                    qrstuvwxyz
NOTICE  02-10 12:53:08.800743 21251  f0   0    src/util/log/test_log.c(68): HEXDUMP "very_large_blob" (50000 bytes at 0x7fffffff1200)
        0000:  62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62  bbbbbbbbbbbbbbbb
        0010:  62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62  bbbbbbbbbbbbbbbb
... SNIP FOR GITHUB ...
        05c0:  62 62 62 62 62 62 62 62 62 62 62 62 62 62 62 62  bbbbbbbbbbbbbbbb
        05d0:  62 62 62 62 62 62 62 62 62 62 62 62              bbbbbbbbbbbb
        ... snip (printed 1500 bytes, omitted 48500 bytes) ...
WARNING 02-10 12:53:08.801724 21251  f0   0    src/util/log/test_log.c(75): HEXDUMP "hex_unprintable_blob" (64 bytes at 0x7fffffff1120)
        0000:  ff 00 ff 82 90 02 05 09 ff 00 ff 82 90 02 05 09  ................
        0010:  ff 00 ff 82 90 02 05 09 ff 00 ff 82 90 02 05 09  ................
        0020:  ff 00 ff 82 90 02 05 09 ff 00 ff 82 90 02 05 09  ................
        0030:  ff 00 ff 82 90 02 05 09 ff 00 ff 82 90 02 05 09  ................
NOTICE  02-10 12:53:08.802132 21251  f0   0    src/util/log/test_log.c(86): HEXDUMP "mixed_blob" (128 bytes at 0x7fffffff1170)
        0000:  ff 00 ff 82 90 02 05 09 ff 00 ff 82 90 02 05 09  ................
        0010:  ff 00 ff 82 90 02 61 09 ff 00 ff 82 90 02 05 09  ......a.........
        0020:  66 90 69 05 72 ff 65 ff 64 90 61 05 6e 00 63 08  f.i.r.e.d.a.n.c.
        0030:  65 00 72 82 90 02 05 09 ff 78 ff 72 90 02 05 09  e.r......x.r....
        0040:  ff 00 41 82 45 02 05 09 ff 78 ff 72 90 02 05 09  ..A.E....x.r....
        0050:  ff 00 ff 82 90 02 05 09 ff 78 ff 72 90 02 05 09  .........x.r....
        0060:  ff 00 ff 82 90 55 05 09 ff 78 ff 72 90 02 05 09  .....U...x.r....
        0070:  50 00 44 82 90 02 05 09 ff 78 ff 72 90 02 05 09  P.D......x.r....
NOTICE  02-10 12:53:08.802240 21251  f0   0    src/util/log/test_log.c(88): Testing log levels
... SNIP FOR GIT HUB ...
```
Something worth considering (noted in the code but wouldn't bother with it now) is that it is probably worthwhile to either restrict with compiler tricks the user from passing a non-compile-time-const string for descr and/or do additional sanitizing of descr in the message function itself (e.g. limit the risk of a user passing a rando pointer into a packet buffer from an untrustred source as the descr argument such that nasty stuff can unintentionally be passed through to downstream loggers).

Also worth considering (and noted in the code but wouldn't bother with it now) is adding a space between col 7 and 8 of the hexdump to make it visually easier for readers to tell their offset within a line.

If we really wanted to optimize this for performance, code footprint and the like (e.g. make this embedded platform friendly), there are a few additional tweaks that could be done but wouldn't hold up the PR to do them.

An, as always, there are always minor naming tweaks and what not but nothing worth holding up the PR over.  Code is readable and the externally exposed API is clean and properly isolates interface from implementation in a target platform independent way.

Since don't seem to have a good means to commit the changes back to your branch at the the moment, below is the function with the above output formatting polishing:
```
char const * 
fd_log_private_hexdump_msg ( char const * descr,
                             void const * blob,
                             ulong        sz ) {

# define FD_LOG_HEXDUMP_BYTES_PER_LINE             (16UL)
# define FD_LOG_HEXDUMP_BLOB_DESCRIPTION_MAX_LEN   (32UL)
# define FD_LOG_HEXDUMP_MAX_INPUT_BLOB_SZ          (1500UL)

# define FD_LOG_HEXDUMP_ADD_TO_LOG_BUF(...)  do {                              \
    num_bytes_written = sprintf ( log_buf_ptr, __VA_ARGS__ );                  \
    if( FD_LIKELY( num_bytes_written>=0 )) log_buf_ptr += num_bytes_written;   \
    } while(0)

  char * log_buf_ptr = fd_log_private_log_msg;   /* used by FD_LOG_HEXDUMP_ADD_TO_LOG_BUF macro */
  int num_bytes_written = 0;                     /* used by the FD_LOG_HEXDUMP_ADD_TO_LOG_BUF macro. signed because *printf() return 'int'. */

  ulong blob_sz = fd_ulong_min( sz, FD_LOG_HEXDUMP_MAX_INPUT_BLOB_SZ );

  /* Print the blob header */
  /* FIXME: consider additional sanitization of descr or using compiler
     tricks to prevent user from passing a non-const-char string (i.e.
     data they got from somewhere else that might not be sanitized). */

  if( FD_UNLIKELY( !descr ) ) {

    FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "HEXDUMP - (%lu bytes at 0x%lx)", sz, (ulong)blob );

  } else if( FD_UNLIKELY( strlen( descr )>FD_LOG_HEXDUMP_BLOB_DESCRIPTION_MAX_LEN ) ) {

    char tmp[ FD_LOG_HEXDUMP_BLOB_DESCRIPTION_MAX_LEN + 1UL ];
    fd_cstr_fini( fd_cstr_append_text( fd_cstr_init( tmp ), descr, FD_LOG_HEXDUMP_BLOB_DESCRIPTION_MAX_LEN ) );
    FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "HEXDUMP \"%s\"... (%lu bytes at 0x%lx)", tmp, sz, (ulong)blob );

  } else {

    FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "HEXDUMP \"%s\" (%lu bytes at 0x%lx)", descr, sz, (ulong)blob );

  }

  if( FD_UNLIKELY( !sz ) ) return fd_log_private_log_msg;

  FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "\n" );

  if( FD_UNLIKELY( !blob ) ) {
    FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "\t... snip (unreadable memory) ..." );
    return fd_log_private_log_msg;
  }

  char line_buf[ FD_LOG_HEXDUMP_BYTES_PER_LINE+1 ];
  const char * blob_ptr = blob;
  ulong count = 0;

  for( ; count<blob_sz; count++ ) {
    /* New line. Print previous line's ASCII representation and then print the offset. */
    if( !( count%FD_LOG_HEXDUMP_BYTES_PER_LINE) ) {
      if( count!=0 ) FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "  %s\n", line_buf );
      FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "\t%04x: ", (unsigned int)count );
    }
    /* FIXME: consider extra space between col 7 and 8 to make easier
       for visual inspection */

    FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( " %02x", blob_ptr[count]&0xff );

    /* If not a printable ASCII character, output a dot. */
    if( isalnum( blob_ptr[ count ] ) || ispunct( blob_ptr[ count ] || blob_ptr[ count ]==' ' ) )
      line_buf[ count%FD_LOG_HEXDUMP_BYTES_PER_LINE ] = blob_ptr[ count ];
    else
      line_buf[ count%FD_LOG_HEXDUMP_BYTES_PER_LINE ] = '.';
    line_buf[ ( count%FD_LOG_HEXDUMP_BYTES_PER_LINE )+1 ] = '\0';
  }

  while( ( count%FD_LOG_HEXDUMP_BYTES_PER_LINE )!=0 ) {
    FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "   " );
    count++;
  }

  FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "  %s", line_buf );
  if( FD_UNLIKELY( blob_sz < sz ) )
    FD_LOG_HEXDUMP_ADD_TO_LOG_BUF( "\n\t... snip (printed %lu bytes, omitted %lu bytes) ...", blob_sz, sz-blob_sz );

  return fd_log_private_log_msg;

# undef FD_LOG_HEXDUMP_BYTES_PER_LINE
# undef FD_LOG_HEXDUMP_BLOB_DESCRIPTION_MAX_LEN
# undef FD_LOG_HEXDUMP_MAX_INPUT_BLOB_SZ
# undef FD_LOG_HEXDUMP_ADD_TO_LOG_BUF
}
```
And with that, I'm happy to approve.

smcio:
Thanks for catching the oversights (I must've accidentally hit Undo a few times or screwed something up between tests/movement of code), and good changes to `fd_log_private_hexdump_msg` for sure, cheers. I've committed all those changes to this PR.

kbowers-jump (approved):

PR #107, opened 2023-02-06
sbpf: add fd_opcode.h

ripatel-fd:
Adds an opcode table for the sBPF instruction set

### DIFF
--- a/src/ballet/sbpf/fd_opcode.h
+++ b/src/ballet/sbpf/fd_opcode.h
@@ -1,0 +1,226 @@
+#ifndef HEADER_fd_src_ballet_sbpf_fd_opcode_h
+#define HEADER_fd_src_ballet_sbpf_fd_opcode_h
+
+/* This header defines the opcodes used in the sBPF (Solana BPF)
+   instruction set. The opcode is the first byte of each instruction
+   slot. */
+
+/* Opcode encoding helpers ********************************************/
+
+/* BPF_OPC: Opcode number given the operation class, mode, sub-operation. */
+#define BPF_OPC( class, mode, variant ) (uchar)( (BPF_CLASS_##class)|(BPF_##mode)|(BPF_##variant) )
+
+/* BPF_CLASS_{...}: Instruction class */
+#define BPF_CLASS_LD     ((uchar)0x00) /* Load                        */
+#define BPF_CLASS_LDX    ((uchar)0x01) /* Load Indexed                */
+#define BPF_CLASS_ST     ((uchar)0x02) /* Store                       */
+#define BPF_CLASS_STX    ((uchar)0x03) /* Store Indexed               */
+#define BPF_CLASS_ALU32  ((uchar)0x04) /* 32-bit integer arithmetic   */
+#define BPF_CLASS_JMP    ((uchar)0x05) /* Control flow                */
+#define BPF_CLASS_ALU64  ((uchar)0x07) /* 64-bit integer arithmetic   */
+
+/* BPF_SRC_{...}: Source operand mode */
+#define BPF_SRC_IMM      ((uchar)0x00) /* Use immediate field         */
+#define BPF_SRC_REG      ((uchar)0x08) /* Use source register         */
+
+/* BPF_SZ_{...}: Operand size modifier */
+#define BPF_SZ_UCHAR     ((uchar)0x10)
+#define BPF_SZ_USHORT    ((uchar)0x08)
+#define BPF_SZ_UINT      ((uchar)0x00)
+#define BPF_SZ_ULONG     ((uchar)0x18)
+
+/* BPF_ADDR_{...}: Addressing mode */
+#define BPF_ADDR_IMM     ((uchar)0x00) /* Use immediate field         */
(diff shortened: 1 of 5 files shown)


src/disco/BUILD
```diff
@@ -21,6 +21,7 @@ fd_cc_library(
         "//src/disco/dedup",
         "//src/disco/mux",
         "//src/disco/replay",
+        "//src/disco/sbpf",
```
kbowers-jump, line 24: This probably belongs in the ballet library as this is not a tile.

src/disco/sbpf/fd_opcode.h
```diff
@@ -0,0 +1,226 @@
+#ifndef HEADER_fd_src_disco_BPF_fd_opcode_h
+#define HEADER_fd_src_disco_BPF_fd_opcode_h
```
kbowers-jump, line 2: After we settle on a name for this, should be consistent with name.  (E.g. sbpf is the directory and that's fine but this implies the name is "BPF")

src/disco/sbpf/fd_sbpf.h
```diff
@@ -0,0 +1,6 @@
+#ifndef HEADER_fd_src_disco_sbpf_fd_sbpf_h
+#define HEADER_fd_src_disco_sbpf_fd_sbpf_h
+
+#include "opcode.h"
```
ptaffet-jump, line 4: Isn't this `fd_opcode.h`

src/disco/sbpf/fd_opcode.h
```diff
@@ -0,0 +1,226 @@
...
+#define BPF_OPC( class, mode, variant ) (uchar)( (BPF_CLASS_##class)|(BPF_##mode)|(BPF_##variant) )
+
+/* BPF_CLASS_{...}: Instruction class */
+#define BPF_CLASS_LD     ((uchar)0x00) /* Load                        */
+#define BPF_CLASS_LDX    ((uchar)0x01) /* Load Indexed                */
+#define BPF_CLASS_ST     ((uchar)0x02) /* Store                       */
+#define BPF_CLASS_STX    ((uchar)0x03) /* Store Indexed               */
+#define BPF_CLASS_ALU32  ((uchar)0x04) /* 32-bit integer arithmetic   */
+#define BPF_CLASS_JMP    ((uchar)0x05) /* Control flow                */
+#define BPF_CLASS_ALU64  ((uchar)0x07) /* 64-bit integer arithmetic   */
```
kbowers-jump, line 20: In addition to the consistent naming, debating with myself whether or not these macros should have a FD_ prefix.  Default for public headers is yes to avoid collisions in user code (that could well be bringing in rando headers from all over the place).  But I suspect (like some of the stuff in simd), these macros are liable to used in pretty well encapsulated locations that the collision risk is low and then often enough in those locations that the namespace-like prefixing is a hinderance to readability.

If we decide to keep as is and not prefixing though, should add a comment at the top to explain why these aren't prefixed though.

src/ballet/sbpf/test_sbpf_opcode.c
```diff
@@ -0,0 +1,130 @@
...
+
+  FD_TEST( 0x85==BPF_OP_CALL       );
+  FD_TEST( 0x8d==BPF_OP_CALLX      );
+  FD_TEST( 0x95==BPF_OP_EXIT       );
+
+  /* Miscellaneous opcodes */
+
+  FD_TEST( 0x18==BPF_OP_LDDW       );
+  FD_TEST( 0xd4==BPF_OP_LE         );
+  FD_TEST( 0xdc==BPF_OP_BE         );
```
kbowers-jump, line 125: Probably would do all these FD_TESTS as a file scope FD_STATIC_ASSERT and the unit test itself is pretty much a default "pass" if the test compiles at all.

src/ballet/sbpf/fd_opcode.h
```diff
@@ -0,0 +1,226 @@
...
+#define HEADER_fd_src_ballet_sbpf_fd_opcode_h
+
+/* This header defines the opcodes used in the sBPF (Solana BPF)
+   instruction set. The opcode is the first byte of each instruction
+   slot. */
+
+/* Opcode encoding helpers ********************************************/
+
+/* BPF_OPC: Opcode number given the operation class, mode, sub-operation. */
+#define BPF_OPC( class, mode, variant ) (uchar)( (BPF_CLASS_##class)|(BPF_##mode)|(BPF_##variant) )
```
kbowers-jump, line 11: Names and prefixes better but still not fully harmonized (this file suggests the prefixes are SBPF but the prefixes are BPF below).  I'm partial to three letter prefix than a four for readability / speakability and partial for disambiguating this from Berkeley Packet Filter as it is evolving (e.g. SBC for Solana Byte Code, SIS for Solana Instruction Set, etc).

Also, the rationale for not having a FD_ prefix on public exposed macros / symbols should be documented at the top.

ripatel-fd:
Marking this as a draft. As per internal discussion, it does not make sense to tackle the SBPF runtime just yet. Will circle back when the runtime scaffold for async replay is set up (milestone 2.1)

ripatel-fd:
Looks like @lheeger-jump rewrote this, closing

PR #109, opened 2023-02-06
shmem: remove libnuma and hwloc dependencies

ripatel-jump:
Adds a libnuma replacement to //src/util/shmem.

This removes licensing restrictions and reduces dependencies.

- Removes libnuma dep and -lnuma linkage
- Vendor system call interface from numaif.h
- Adds wrappers for Linux kernel syscalls
- Replaces libnuma functions via Linux sysfs/sysinfo APIs

Replaces PR #57


ripatel-fd:
Will mark this PR as "ready for review" when it has undergone the previously discussed polishing.

PR #117, opened 2023-02-09
ci: speed up testing, fix ulimits

ripatel-fd:
Currently, merges are blocked until tests with coverage succeed.
However, running coverage tests often takes over 15 minutes.

This change modifies the Bazel build job to also run tests without
coverage enabled.  The branch protection config can be adjusted to only
require tests without coverage, to allow merging before the (slower)
tests with coverage pass.

### DIFF
--- a/.github/scripts/run-ulimit.sh
+++ b/.github/scripts/run-ulimit.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+
+# Runs the given command with increased capabilities.
+# Expects cwd to be the repository root.
+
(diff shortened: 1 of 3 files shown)


ripatel-fd:
@daviesalex Would love to get your thoughts on the general approach of this. Misconfigured ulimits is a problem we'll only have with GitHub-hosted runners since we can fix this on self-hosted runners in the system configuration.

ripatel-fd:
This is very Bazel and GitHub Actions default runner specific, so is not applicable to current main anymore.

PR #121, opened 2023-02-10
tango: remove test_cache from automatic CI config

ripatel-fd:
Unfortunately, CI rlimits are still too low.
Re-enabling this test will require larger changes with our CI runner
setup, such as using larger GitHub-hosted or self-hosted runners.

### DIFF
--- a/.github/workflows/bazel_test.yml
+++ b/.github/workflows/bazel_test.yml
@@ -35,10 +35,12 @@ jobs:
 
       - name: Upload coverage report to CodeCov
         uses: codecov/codecov-action@v3
+        timeout-minutes: 5
         with:
           files: "bazel-out/_coverage/_coverage_report.dat"
           name: codecov-bazel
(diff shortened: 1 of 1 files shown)


ptaffet-jump (approved):
Note: This undoes #116, which was only merged because the Github rules weren't set properly at that point.

.github/workflows/bazel_test.yml
```diff
@@ -35,10 +35,12 @@ jobs:
 
       - name: Upload coverage report to CodeCov
         uses: codecov/codecov-action@v3
+        timeout-minutes: 5
         with:
           files: "bazel-out/_coverage/_coverage_report.dat"
           name: codecov-bazel
-          fail_ci_if_error: true
+          fail_ci_if_error: false
+          functionalities: search
```
ptaffet-jump, line 43: Do we know what this actually means?

ptaffet-jump (approved):

PR #122, opened 2023-02-10
ballet: add block module and fd_microblock_t

### LINKED ISSUE #27: block: add entries and PoH mixin
- [ ] add entry header data structure
- [ ] add entry tx list data structure
- [ ] derive PoH mixins from entries

ripatel-jump:
Adds data structure for microblocks.

- Depends on #125 
- Closes #27


jumpsiegel (dismissed):

src/ballet/block/fd_microblock.h
```diff
@@ -0,0 +1,180 @@
...
+  void * raw; /* TODO: Make this a gaddr instead of laddr */
+
+  /* Size of txn */
+  ushort txn_sz;
+};
+typedef struct fd_rawtxn_b fd_rawtxn_b_t;
+
+/* fd_microblock: Stores a deserialized microblock.
+
+   References an external buffer containing raw txns.
```
ptaffet-jump, line 68:
```suggestion
   References an external buffer containing transaction payloads.
```

src/ballet/block/fd_microblock.h
```diff
@@ -0,0 +1,180 @@
...
+   Reasons for failure include invalid data or unexpected EOF.
+
+   If the returned ptr is non-NULL, it is guaranteed to be
+     1) greater than the provided `buf` ptr
+     2) less or equal than `(uchar *)buf + buf_sz` */
+int
+fd_microblock_deserialize( fd_microblock_t * block,
+                           void **           buf,
+                           ulong *           buf_sz,
+                           fd_txn_parse_counters_t * counters_opt );
```
ptaffet-jump, line 162: I was confused initially because it doesn't return a pointer. I haven't read the implementation yet, but my guess is `buf` is an in/out parameter to facilitate parsing multiple microblocks from a single buffer.

What about taking `buf` as a `uchar const *` and returning a `ulong` which is just how much data was consumed or 0 on error. I could be persuaded to switch to `void *`. Either way, it should be `const`, and I much prefer returning a value instead of using output parameter unless there's a strong reason for it.

src/ballet/block/fd_microblock.h
```diff
@@ -0,0 +1,180 @@
...
+/* fd_microblock_mixin: Calculates the PoH mixin hash.
+
+   Computes the root of a 32-byte binary Merkle tree of a vector with
+   each element containing the first Ed25519 signature of each txn.
+
+   For microblocks that do not contain transactions, a single append
+   instead of a mixin should be applied to PoH.
+
+   U.B. if this microblock contains a txn with zero signatures (illegal txn).
+   U.B. if this microblock's `hdr.txn_cnt` is zero. */
```
ptaffet-jump, line 173: Would it make sense for this code to handle the 0 transactions case? Is there a reason it can't?

src/ballet/block/test_microblock.c
```diff
@@ -0,0 +1,148 @@
+#include "fd_microblock.h"
+
+#include "../../util/sanitize/fd_sanitize.h"
+
+#include <stddef.h>
```
ptaffet-jump, line 5: What does it need from stddef?

src/ballet/block/test_microblock.c
```diff
@@ -0,0 +1,148 @@
...
+FD_STATIC_ASSERT( offsetof( fd_microblock_hdr_t, hash     )==0x08UL, alignment );
+FD_STATIC_ASSERT( offsetof( fd_microblock_hdr_t, txn_cnt  )==0x28UL, alignment );
+FD_STATIC_ASSERT( sizeof  ( fd_microblock_hdr_t           )==0x30UL, alignment );
+FD_STATIC_ASSERT( alignof ( fd_microblock_hdr_t           )==0x01UL, alignment );
+
+FD_STATIC_ASSERT( alignof(fd_microblock_t)== 64UL, alignment );
+FD_STATIC_ASSERT( sizeof (fd_microblock_t)==128UL, alignment );
+
+void
+test_microblock() {
```
ptaffet-jump, line 19:
```suggestion
test_microblock( void ) {
```

src/ballet/block/test_microblock.c
```diff
@@ -0,0 +1,148 @@
...
+  FD_TEST( fd_microblock_join( NULL        )==NULL ); /* null shblock */
+  FD_TEST( fd_microblock_join( (void *)1UL )==NULL ); /* misaligned shblock */
+
+  /* Test microblock creation */
+
+  ulong txn_max_cnt = 16UL;
+  ulong footprint = fd_microblock_footprint( txn_max_cnt );
+  FD_TEST( footprint );
+
+  void * shmem = fd_alloca( FD_MICROBLOCK_ALIGN, footprint );
```
ptaffet-jump, line 48: If you're going to use `fd_alloca`, you need to guard it with `FD_HAS_ALLOCA`. It's probably easier to stick a sufficiently large chunk of statically allocated memory at the top of the C file like you do for `microblock_buf`.

src/ballet/block/test_microblock.c
```diff
@@ -0,0 +1,148 @@
...
+};
+typedef struct fd_microblock_test_vec fd_microblock_test_vec_t;
+
+void
+test_parse_localnet_batch_0( void ) {
+  FD_TEST( localnet_batch_0_sz==3080UL );
+
+  /* Peek the number of entries, which is the first ulong. */
+  ulong microblock_cnt = *(ulong *)localnet_batch_0;
+  FD_TEST( microblock_cnt==64UL );
```
ptaffet-jump, line 99: If I'm understanding properly, this is basically testing the test. If this fails, it doesn't mean there's something wrong with the code, right? Maybe add a comment explaining if it fails.

src/ballet/block/test_microblock.c
```diff
@@ -0,0 +1,148 @@
...
+
+  /* Move past the first ulong to the entries. */
+  void * batch_buf = (void *)((uchar *)localnet_batch_0    + 8UL);
+  ulong  batch_bufsz =                 localnet_batch_0_sz - 8UL ;
+
+  /* Check whether our .bss buffer fits. */
+  ulong const txn_max_cnt = 10;
+  ulong footprint = fd_microblock_footprint( txn_max_cnt );
+  FD_TEST( sizeof(microblock_buf)>=footprint );
```
ptaffet-jump, line 108: Same here, but I guess the comment ` Check whether our .bss buffer fits` is sufficient.

src/ballet/block/test_microblock.c
```diff
@@ -0,0 +1,148 @@
...
+  /* Test bad magic value.
+     Note that at this point our `block` pointer is dangling. */
+  block->magic++;
+  FD_TEST( fd_microblock_delete( shblock )==NULL );
+  block->magic--;
+
+  FD_TEST( fd_microblock_delete( shblock )==shmem );
+}
+
+/* A serialized batch of entries.
```
ptaffet-jump, line 79: `entries` and `entry` occur a few times in this file. Can you rename them all to microblocks?

src/ballet/block/test_microblock.c
```diff
@@ -0,0 +1,148 @@
...
+test_parse_localnet_batch_0( void ) {
+  FD_TEST( localnet_batch_0_sz==3080UL );
+
+  /* Peek the number of entries, which is the first ulong. */
+  ulong microblock_cnt = *(ulong *)localnet_batch_0;
+  FD_TEST( microblock_cnt==64UL );
+
+  /* Move past the first ulong to the entries. */
+  void * batch_buf = (void *)((uchar *)localnet_batch_0    + 8UL);
+  ulong  batch_bufsz =                 localnet_batch_0_sz - 8UL ;
```
ptaffet-jump, line 103: I was surprised to see you manually skipping some bytes until I realized the fixture is actually a batch containing many microblocks not a single microblock. It might be worth throwing in a sentence along those lines.

src/ballet/block/test_microblock.c
```diff
@@ -0,0 +1,148 @@
...
+}
+
+int
+main( int     argc,
+      char ** argv ) {
+  fd_boot( &argc, &argv );
+
+  test_microblock();
+  test_parse_localnet_batch_0();
+
```
ptaffet-jump, line 144: Would it be possible to get a batch one with some transactions in it? The whole transaction processing loop is not tested at the moment.

src/ballet/block/fd_microblock.c
```diff
@@ -0,0 +1,181 @@
...
+
+  return shblock;
+}
+
+int
+fd_microblock_deserialize( fd_microblock_t * block,
+                           void **           buf,
+                           ulong *           buf_sz,
+                           fd_txn_parse_counters_t * counters_opt ) {
+# define ADVANCE(n) do {                       \
```
ptaffet-jump, line 118: Just a formatting nitpick, but I think the whole block should be indented and the `do` should go on its own line.
```suggestion
# define ADVANCE(n)                            \
  do {                                         \
```

ptaffet-jump (changes requested):
Overall good! The test with transactions and the `fd_microblock_deserialize` API with more indirection than it needs are the biggest things. I think the rest of the comments would be helpful but I'm open if you disagree

src/ballet/block/fd_microblock.h
```diff
@@ -0,0 +1,180 @@
...
+   Each microblock contains a list of transactions, with the constraint
+   that if a txn writes to an account, no other txn in that list can
+   read or write to the same account.
+
+   Each microblock also optionally mixes in the Merkle root of the
+   first signature of each txn, and appends a number of PoH hashes. */
+
+#define FD_MICROBLOCK_ALIGN (64UL)
+
+#define FD_MICROBLOCK_MAGIC (0x176aa423e0372d6aUL) /* random */
```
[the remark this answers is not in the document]
jumpsiegel: These blocks might get passed to executors

PR #123, opened 2023-02-10
shred: add deshredder

ripatel-jump:
Adds logic for deshredding shreds to (serialized) entries.

- Depends on #125 
- Depends on #122

### DIFF
--- a/src/ballet/shred/fd_shred.c
+++ b/src/ballet/shred/fd_shred.c
@@ -1,15 +1,125 @@
 #include "fd_shred.h"
 
 fd_shred_t const *
-fd_shred_parse( uchar const * buf ) {
+fd_shred_parse( uchar const * const buf,
+                ulong         const sz ) {
+  /* Reinterpret size as long: `sz` cannot realistically underflow.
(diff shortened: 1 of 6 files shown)


src/ballet/shred/fd_shred.c
```diff
@@ -1,15 +1,125 @@
...
+  shredder->data_off  = 0U;
+  shredder->buf       = buf;
+  shredder->bufsz     = bufsz;
+  shredder->result    = -FD_SHRED_EPIPE;
+}
+
+long
+fd_deshredder_next( fd_deshredder_t * const shredder ) {
+  /* Terminate early if no shreds are left */
+  if( FD_UNLIKELY( shredder->shred_cnt == 0U ) )
```
jumpsiegel, line 75: this check is effectively duplicated at line 84.  I know they are returning a different result but I would prefer to collapse them somehow.  In my use case, I am calling the deshredder one shred at a time (since that avoids an extra memory copy) and eliminating the check/branch here eliminates most of the cost difference between passing in a batch of shreds and calling it one at a time.

PR #125, opened 2023-02-11
txn: extend fd_txn_parse to allow trailing bytes

ripatel-fd:
Adds a payload_size_opt output parameter to fd_txn_parse that permits
providing buffers with trailing bytes after the txn. This is usually the
case when the size of the txn is not known beforehand. For example, the
wire representation of a block relies on the self-describing properties
of the txn serialization to concatenate txns without length prefixes.


ripatel-fd:
Closing this @ptaffet-jump since you have a better fix

PR #129, opened 2023-02-16
aio: add fd_aio_send API

ripatel-fd:
Picked from `nbridge/quic`

### DIFF
--- a/src/tango/aio/fd_aio.h
+++ b/src/tango/aio/fd_aio.h
@@ -1,13 +1,18 @@
-#ifndef HEADER_fd_src_tango_aio_aio_h
-#define HEADER_fd_src_tango_aio_aio_h
+#ifndef HEADER_fd_src_tango_aio_fd_aio_h
+#define HEADER_fd_src_tango_aio_fd_aio_h
 
-#include "../../util/fd_util_base.h"
+#include "../fd_tango_base.h"
 
-/* fd_aio defines a simple abstraction for asynchronous I/O.
+/* fd_aio defines a simple abstraction for asynchronous sending and
+   receiving packets.  It abstracts away many details so the same code
+   can work performant and transparent with different low level I/O
+   libraries and hardware.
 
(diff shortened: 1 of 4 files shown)


src/tango/aio/fd_aio.h
```diff
@@ -0,0 +1,130 @@
...
+       // possibly do some other process that might free up resources
+       // to avoid deadlock
+     } */
+
+
+/* fd_aio_buffer_t is a contiguous range of I/O data in local memory
+   passed from sender to receiver. The receiver may access up to
+   `data_sz` bytes from the memory area beginning at `data`. */
+
+#define FD_AIO_BUFFER_ALIGN (16UL)
```
kbowers-jump, line 63: Probably would have matching footprint for consistency with other such.

src/tango/aio/fd_aio.h
```diff
@@ -0,0 +1,130 @@
...
+
+struct fd_aio {
+  /* Implementation-defined context object.  Accessed by both sender
+     and receiver. */
+  void * context;
+
+  /* Receiver callback called by sender. */
+  fd_aio_cb_receive_t cb_receive;
+};
+typedef struct fd_aio fd_aio_t;
```
kbowers-jump, line 93: Maybe make this opaque or quasi-opaque (if user access to context and cb we desire to have user access to context and cb_receive).  Probably would use send / recv for naming consistency with other code.

src/tango/aio/fd_aio.h
```diff
@@ -0,0 +1,130 @@
...
+
+FD_PROTOTYPES_BEGIN
+
+/* fd_aio_send: Sends a batch of buffers to the receiver.
```
kbowers-jump, line 97: Probably should have the usual APIs for objection creation and destruction for consistency (even if they aren't needed by the current implementation).

tpointon-jump (approved):

PR #132, opened 2023-02-21
Adding fd_alloc API for interprocess shareable and persistent fine grained allocation

kbowers-jump:
And low level commits needed for the implementation.


ripatel-fd:
Will add Bazel targets so this can be merged

src/util/alloc/test_alloc.c
```diff
@@ -0,0 +1,214 @@
...
+/* This is a torture test for same thread allocation */
+/* FIXME: IDEALLY SHOULD ADD TORTURE TEST FOR MALLOC / FREE PAIRS SPLIT
+   BETWEEN THREADS AND ADD ADD INTERPROCESS TESTING MODES. */
+
+static int    _go;
+static void * _shalloc;
+static ulong  _alloc_cnt;
+static ulong  _align_max;
+static ulong  _sz_max;
+
```
jumpsiegel, line 19: Could main get stuck at the top?  Tests tend to be primary documentation so I want to see how to set up the test first.

src/util/alloc/test_alloc.c
```diff
@@ -0,0 +1,214 @@
...
+main( int     argc,
+      char ** argv ) {
+  fd_boot( &argc, &argv );
+
+  char const * name      = fd_env_strip_cmdline_cstr ( &argc, &argv, "--wksp",      NULL,      NULL );
+  ulong        alloc_cnt = fd_env_strip_cmdline_ulong( &argc, &argv, "--alloc-cnt", NULL, 1048576UL );
+  ulong        align_max = fd_env_strip_cmdline_ulong( &argc, &argv, "--align-max", NULL,     256UL );
+  ulong        sz_max    = fd_env_strip_cmdline_ulong( &argc, &argv, "--sz-max",    NULL,   73728UL );
+  ulong        tile_cnt  = fd_tile_cnt();
+
```
jumpsiegel, line 138: if (1U == tile_cnt)  FD_LOG_NOTICE(("tiles not configured.. go read...")) ?

src/util/alloc/fd_alloc.h
```diff
@@ -0,0 +1,314 @@
...
+   other usage (including uses not through this fd_alloc).
+
+   (It would be possible to implement this less efficiently in space and
+   time such that join didn't need to be passed.  The current design has
+   picked efficiency and consistency with other APIs though.) */
+
+void
+fd_alloc_free( fd_alloc_t * join,
+               void *       laddr );
+
```
jumpsiegel, line 304: Also, would love a f_alloc_dump(...) that just prints the allocation tables with size so that I can look for leaks..

src/util/alloc/fd_alloc.c
```diff
@@ -0,0 +1,942 @@
...
+     superblocks and all inactive superblocks.  We will not be able
+     cleanup any superblocks that are fully allocated (and any
+     outstanding large allocations) as we rely on the application to
+     implicitly track them (because they nominally will eventually call
+     free on them).  So hopefully the application did a good job and
+     cleaned up all their outstanding stuff before calling delete. */
+
+  fd_wksp_t * wksp = fd_alloc_private_wksp( alloc );
+
+  for( ulong sizeclass=0UL; sizeclass<FD_ALLOC_SIZECLASS_CNT; sizeclass++ ) {
```
jumpsiegel, line 542: Is there a reason you explicitly take apart your data structure when you could just memset it to zero or just reset the magic number and it is gone?

kbowers-jump: The taking apart is required to get rid of superblocks still lying around even after the user (being righteous and proper) has cleaned up all their allocations.  If it was just zeroed, there would likely be huge superblocks stranded in the allocator of last resort.

src/util/alloc/test_alloc.c
kbowers-jump: Not necessary ... works fine single threaded (extensive single threaded testing with tile_cnt==1 was already done, as was tested with that and with cores on multi-socket high core count boxes, as was deliberately error injection into every code path the implementation to make sure the tester did in fact find atomic bugs in all code paths).

src/util/alloc/test_alloc.c
kbowers-jump: Not sure what this means.  If it got stuck, that would be bug in either the implementation or in the code.  Minimum test is just to call it without arguments on a box with a shmem sandbox and a free gigantic page on the numa node where the test will be run (much like the other tests).  Can do more extensive tests, performance, etc by passing in different sets of tile cpus, shared workspaces backed by differently backed or different affinity memory, etc.

src/util/alloc/fd_alloc.c
kbowers-jump: Note: could argue that you could limit this to just cleaning up huge superblock sizeclasses to tune it.  But this isn't a critical performance path as it is purely teardown.  And, doing it this way is useful was simpler and also was done to make it easy to do the types of instrumentation / integrity checking discussed above without impact performance.

For example, in a righteous and proper application, all the superblocks at this point found should be completely empty if deleted in the above order during the tear down.  Quite possible to test this and noisily log about it.

This noisy logging is not done currently because a common usage pattern to have a quick-and-dirty application call a bunch of mallocs and then just terminate without cleaning up (see rants again ... they aren't actually using malloc-free semantics).  Another common usage pattern is to terminate anomalously leaving things deliberately unclean to help with core dumps and what not.  In either case, a default policy of noisily logging such is not helpful and possibly self defeating in a bigger picture sense if it trains users to ignore logs.

src/util/alloc/test_alloc.c
jumpsiegel: I guess I was asking that the code get moved up here so that you can read from the top of the file down instead of having to jump to the second to last function to see the setup...

src/util/alloc/fd_alloc.c
jumpsiegel: Ahh, I understand

src/util/alloc/fd_alloc.h
jumpsiegel: Eventually we will have leaks and so would want a way to write tests to validate the memory management at a lower level.   This can wait till a future PR

src/util/alloc/test_alloc.c
kbowers-jump: I thought you meant main getting stuck in loop (like the spinning waiting to go).

No strong feelings about it other than to be consistent (e.g. the existing more complex unit tests currently all have their mains at the end but lots of unit tests are simple enough structurally that main is at both the top and bottom ... the main consideration then in mind, if it is worth doing it here, it is probably worth doing it everywhere).

jumpsiegel (approved):

src/util/alloc/fd_alloc.h
kbowers-jump: With the most recent commit, we are 99% the way there.

The allocator and underlying wksp now have enough mojo to detect any leaks at program termination (e.g. as per the other discussion, a diagnostic can run through the superblocks in the alloc to find stray small allocations as described before and then run through the wksp partition table to look for stray large allocations via the pointer tags), to dump out program memory usage, etc.

As a bonus, since this has no additional costs in time / space / parallel efficiency or what not, the support will just be available all the time.

It can also do party tricks like selective massive free with same algorithmic efficiency single free and classify all wksp allocations efficiently (so can start doing garbage collection likes things too if so inclined ... mixed feelings at best about that).

PR #133, opened 2023-02-23
config: add minimal eBPF build environment

ripatel-fd:
Adds fd_ebpf_base.h mocking fd_util_base.h to provide a minimal C
programming environment for eBPF targets like XDP.

### DIFF
--- a/config/everything.mk
+++ b/config/everything.mk
@@ -37,6 +37,7 @@ help:
 	# Explicit goals are: all bin include lib unit-test help clean distclean asm ppp
 	# "make all" is equivalent to "make bin include lib unit-test"
 	# "make bin" makes all binaries for the current platform
+	# "make ebpf-bin" makes all eBPF binaries
 	# "make include" makes all include files for the current platform
 	# "make lib" makes all libraries for the current platform
 	# "make unit-test" makes all unit-tests for the current platform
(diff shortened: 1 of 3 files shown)


jumpsiegel (approved):

config/everything.mk
```diff
@@ -37,6 +37,7 @@ help:
 	# Explicit goals are: all bin include lib unit-test help clean distclean asm ppp
 	# "make all" is equivalent to "make bin include lib unit-test"
 	# "make bin" makes all binaries for the current platform
+	# "make ebpf-bin" makes all eBPF binaries
 	# "make include" makes all include files for the current platform
```
jumpsiegel, line 41: Amusingly, this target is invalid until you get your first use of ebpf (in the next PR).    (Resolve this at will.. it is just a comment)

src/net/ebpf/fd_ebpf_base.h
```diff
@@ -0,0 +1,41 @@
...
+   Clang/LLVM eBPF little endian target.  This is a temporary measure
+   working around the fact that eBPF does not support libc yet, which is
+   a dependency of fd_util_base.  For documentations, see fd_util_base.h */
+
+/* Minimal libc-like environment **************************************/
+
+/* Not even the pre-processor definition is NULL is available on eBPF.
+   We define the bare minimum to resemble libc. */
+
+#define NULL (void *)0
```
nbridge-jump, line 17: we need to be very careful here, since 0 is a valid location in packet contexts.
It's probably best if NULL is NOT defined, since it will always be misleading in those contexts

nbridge-jump: test

nbridge-jump: I'm not sure how that's relevant. I'm saying a test for, for example, p == (void*)0 isn't useful, because it may be true for a valid dereferenceable pointer.

PR #136, opened 2023-02-23
add libbpf dependency

ripatel-fd:
- add libbpf Bazel dep
- add libbpf make profile
- fixes bug where has-hosted is not imported for linux_gcc_x86_64 nor linux_clang_x86_64

PR #139, opened 2023-02-25
config: remove with-libbpf.mk

ripatel-fd:
Fixes regression from #136 
with-libbpf should not be enabled by default


jumpsiegel (approved):

ripatel-fd:
Added #140 to allow for optional loading of configs.

PR #155, opened 2023-03-01
util: add --netns flag to fd_boot()

ripatel-fd:
Allows entering persistent network namespaces (useful for tests and sandboxing)

### DIFF
--- a/src/util/fd_util.c
+++ b/src/util/fd_util.c
@@ -1,3 +1,7 @@
+#if defined(__linux__)
+#include "fd_util_linux.c"
+#endif
+
 #include "fd_util.h"
 
 void
@@ -6,6 +10,9 @@ fd_boot( int *    pargc,
   /* At this point, we are immediately after the program start, there is
      only one thread of execution and fd has not yet been booted. */
   fd_log_private_boot  ( pargc, pargv );
+# if defined(__linux__)
+  fd_linux_private_boot( pargc, pargv );
+# endif
   fd_shmem_private_boot( pargc, pargv );
   fd_tile_private_boot ( pargc, pargv ); /* The caller is now tile 0 */
 }
--- a/src/util/fd_util_linux.c
+++ b/src/util/fd_util_linux.c
@@ -1,0 +1,34 @@
+#define _GNU_SOURCE
+#include <errno.h>
+#include <fcntl.h>
+#include <sched.h>
+
+#include "fd_util_base.h"
+#include "log/fd_log.h"
+
+/* fd_linux_enter_netns: Replaces the network namespace of the calling
+   thread with the namespace located at the given nsfs mount path. */
+static void
+fd_linux_enter_netns( char const * netns ) {
+  /* These syscalls mirror `ip netns exec` */
+  int ns_fd = open( netns, O_RDONLY|O_CLOEXEC );
+  if( FD_UNLIKELY( ns_fd<0 ) ) {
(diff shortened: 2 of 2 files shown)


jumpsiegel (approved):

ripatel-fd:
An open question was whether we want to keep the `--netns` flag on targets that don't support them (e.g. macOS). The current behavior on non-Linux targets is to just ignore the `--netns` flag which can cause issues on downstream command-line handling. Perhaps it is better to remove the flag from argv and log a warning that the flag is ignored for this target.

kbowers-jump:
Don't have a lot of context to have an informed opinion (depending on context, might want to rubber stamp, tweak, etc).

ripatel-fd:
@kbowers-jump An approval would be appreciated. Had to force push to remove my Bazel change.

ripatel-fd:
ping @marcus-jump since this relates to your sandboxing work

kbowers-jump:
(Finally coming up for air and looking at a backlog of PRs.)

I'm generally fine with this in terms of content.  The main question is where this should live in src.

As it is fd_util_linux.c is kinda wonky as it looks like it wants to be in its own compile unit (e.g. as a set of APIs that are dependent on log functionality).  But it is getting directly included the top level fd_util.c (which already exposes all the APIs) and in a top level directory.

Some options to cosnider:

(1) Add this to an existing set of APIs.  util/log, util/tile, util/sanitize and maybe util/env might be applicable ... none are a particularly great fit (but most are already doing things like providing boot/halt, bringing in sched.h, doing _GNU_SOURCE ... signs that at least there is some overlap with the implementations and functionality of this and the functionality provided by those preexisting APIs).

(2) Put this in its own compile unit in the top level directory.  Cleans up some include-age but given the file feels a little out of place.  (E.g. it isn't some low level functionality provide by fd_util_base.h, etc.)

(3) Do 2 but make a fully fledged set of APIs (its own directory and all that).  Since this is pretty light though (the only real API is the boot / halt), this might be overkill.

(4) Do 3 but consolidate a lot of security related APIs (e.g. stuff in sanitize / asan, the jailer security stuff Marc is doing, etc) in that set of APIs.  Such might be good place how to create a natural long term home for APIs like this.

I'm weakly partial to 4 but could be easily persuaded otherwise.  If we want to go down this route, I'm happy to do the necessary shuffling if nobody else has bandwidth to do it.

marcus-ares:
Hey Richie, I would have preferred to leave this this in my sandboxing change.
I think that the proper way to go about it would be to jail each test. How do you feel about it?

kbowers-jump:
Yep, chatting with Marcus, looks like we can do something akin to option 4 and roll this up as part of his sandboxing work.  So, unless there are any objections, my current thinking is to close this PR and merge the meat of it with Marcus's sandboxing PR.

ripatel-fd:
Sounds good to me 👍  @marcus-jump note that this PR does not implement a security feature, but just a way to enter existing networking namespaces for testing.

ripatel-fd:
QUIC and XDP are now active on main. The XDP tests are not functional with this change. Going to merge this in to fix XDP. Please feel free to modify this in future PRs (without breaking XDP network namespace-testing functionality)

lidatong (approved):

lidatong:
> QUIC and XDP are now active on main. The XDP tests are not functional with this change. Going to merge this in to fix XDP. Please feel free to modify this in future PRs (without breaking XDP network namespace-testing functionality)

discussed with richie in person, going to approve, can help with any rebasing

PR #156, opened 2023-03-01
util: add fd_cstr_to_ip4_addr

ripatel-fd:


src/util/cstr/fd_cstr.h
```diff
@@ -53,6 +53,14 @@ FD_FN_PURE  double       fd_cstr_to_double( char const * s );
 
 FD_FN_PURE ulong fd_cstr_to_ulong_octal( char const * s );
 
+/* fd_cstr_to_ip4_addr parses an IPv4 address matching format
+   %u.%u.%u.%u  On success returns the numerical representation of the
+   address in [0;UINT_MAX). On fail returns ULONG_MAX. */
```
tpointon-jump, line 58: nit: I wonder if we should typedef IPv4 addresses, with a comment explaining how the address is represented, because they should be interpreted differently to typical `ulong`s.

src/util/cstr/fd_cstr.c
```diff
@@ -24,6 +24,27 @@ double       fd_cstr_to_double( char const * cstr ) { return         strtod ( cs
...
+
+  int res = sscanf( s, "%u.%u.%u.%u%n", &x[0], &x[1], &x[2], &x[3], &n );
+
+  if( FD_UNLIKELY( res!=4
+                || n<0 || s[n]!='\0'
+                || x[0]>UCHAR_MAX
+                || x[1]>UCHAR_MAX
+                || x[2]>UCHAR_MAX
+                || x[3]>UCHAR_MAX ) )
+    return ULONG_MAX;
```
tpointon-jump, line 42: Just thinking out loud: wonder if it might be worth defining a `FD_FN_CONST static inline is_valid_ip4_addr` function which just returns false if the parameter is `ULONG_MAX`. That way the user doesn't have to remember the convention used for representing invalid values.

tpointon-jump (approved):

Issue #152, opened 2023-02-28
ci: run make tests


ripatel-fd:
Add a CI job that runs unit tests and test scripts generated by `make`.

PR #160, opened 2023-03-02
ci: basic make-based unit testing

ripatel-fd:
- contrib/unit_tests.txt: Adds simple list of unit tests that can run without shared memory / allocations
- contrib: also build for linux_clang_x86_64
- contrib: run make-built tests via parallel

Closes https://github.com/firedancer-io/firedancer/issues/152


.github/workflows/make_build.yml
```diff
@@ -7,12 +7,18 @@ on:
   workflow_dispatch:
 jobs:
   make-build:
+    strategy:
+      matrix:
+        compiler: [gcc, clang]
```
tpointon-jump, line 12: note for later: we should be explicit about the compiler versions here

tpointon-jump (approved):

PR #167, opened 2023-03-02
git: add vim, vimdiff items to gitignore

lheeger-jump:
Gitignore update

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,10 @@ target/
 
 # Nix
 /result
+
+# Vimdiff
+*.orig
+
+# Vim
+*.swp
+*.swo


ripatel-fd:
No objection to merging this since I don't want to be annoying, but IMHO editor files don't belong in project-specific config

```shell
echo "*.swp" >> "$HOME"/.gitignore
git config --global core.excludesfile "$HOME"/.gitignore
```

lheeger-jump:
> No objection to merging this since I don't want to be annoying, but IMHO editor files don't belong in project-specific config
> 
> ```shell
> echo "*.swp" >> "$HOME"/.gitignore
> git config --global core.excludesfile "$HOME"/.gitignore
> ```

Disagree. I do not want anyone accidentally pushing up editor files that shouldn't be in the repo even on a fresh clone. Examples elsewhere:

https://github.com/mysql/mysql-server/blob/8.0/.gitignore
https://github.com/ethereum/go-ethereum/blob/master/.gitignore
https://github.com/torvalds/linux/blob/a9a01e1238cf5b477ec6aa54855356e518998991/.gitignore#L140
https://github.com/solana-labs/solana/blob/7b1d446001e3445b76a8b80a5ed04d4a4866e2d9/.gitignore#L25

ripatel-fd:
The list of operating systems and editors people will use to write code for the project is endless. Vim gets a pass because it's quite widespread. If you fix your system config, you'll save yourself a few keystrokes and PRs in the future, though!

PR #169, opened 2023-03-03
config: add libFuzzer ASan support

ripatel-fd:


config/everything.mk
```diff
@@ -2,7 +2,7 @@ MAKEFLAGS += --no-builtin-rules
 MAKEFLAGS += --no-builtin-variables
 .SUFFIXES:
 .SUFFIXES: .h .hxx .c .cxx .o .a .d .S .i
-.PHONY: all bin include lib unit-test help clean distclean asm ppp show-deps
+.PHONY: all bin include lib unit-test fuzz-test run-unit-test help clean distclean asm ppp show-deps
```
tpointon-jump, line 5: Could we also add these to the `Explicit goals are` comment on line 38?

config/everything.mk
```diff
@@ -184,9 +188,17 @@ run-$(3): run-$(1)
...
 make-bin       = $(eval $(call _make-exe,$(1),$(2),$(3),bin))
 make-unit-test = $(eval $(call _make-exe,$(1),$(2),$(3),unit-test))
+make-fuzz-test =
 run-unit-test = $(eval $(call _run-unit-test,$(1),$(2),unit-test))
+else
+make-bin =
+make-unit-test =
+make-fuzz-test = $(eval $(call _make-exe,$(1),$(2),$(3),fuzz-test))
+run-unit-test =
+endif
```
tpointon-jump, line 201: Instead of doing this conditional compilation and requiring this flag, it might be cleaner to just have the fuzzing script call `make fuzz-test`. Feels like it should be up to the caller to choose what should be built, not have `make` magically exclude stuff.

PR #171, opened 2023-03-03
config: allow arbitrary deps for objs

ripatel-fd:
Expands add-objs, make-bin, make-unit-test to allow specifying arbitrary
prerequisites for object files.  This is useful for binary includes and
"far-away" dependencies such as eBPF programs.

Example:

    $(call make-bin,fd_xdp_ctl,fd_xdp_ctl,fd_xdp fd_util,$(EBPF_BINDIR)/fd_xdp_redirect_prog.o)

### DIFF
--- a/config/everything.mk
+++ b/config/everything.mk
@@ -76,17 +76,20 @@ endef
 make-lib = $(eval $(call _make-lib,$(1)))
 
 ##############################
-# Usage: $(call add-objs,objs,lib)
+# Usage: $(call add-objs,objs,lib,prerequisites)
 
 define _add-objs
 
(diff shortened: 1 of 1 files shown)


lheeger-jump (commented):
Is this property transitive? Does it do transitive deps correctly?

ripatel-fd:
> Is this property transitive? Does it do transitive deps correctly?

@lheeger-jump Yes, that's why I added it :) This will build the XDP redirect program mentioned here, which will in turn build its dependency tree. There are some requirements however: The dependency path must be "clean" (no double slashes, no dots or double dot path segments) and must be relative to the repository root.

PR #174, opened 2023-03-03
build, ci: add native CC config

ripatel-fd:
Fixes GitHub CI incompatibility problems by providing an -march=native profile.


ripatel-fd:
@mmcgee-jump This has been a long-standing issue. The purpose of this PR is to provide a default native config that enables whatever features the host supports, like GFNI, SHA-NI, AVX, etc. I'm not able to bring this change into our Make configuration at the moment, but I'd encourage you to pick it up if you think it's something we should have. Thanks!

PR #195, opened 2023-03-09
ci: switch to Firecracker runners

marcus-jump:
Does what it says on the tin.

### DIFF
--- a/.github/workflows/make_build.yml
+++ b/.github/workflows/make_build.yml
@@ -10,10 +10,10 @@ jobs:
     strategy:
       matrix:
         compiler: [gcc, clang]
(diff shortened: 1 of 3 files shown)


marcus-ares:
Also, `--run` is more appropriate than `--command` based on nix-shell's man page [0].

[0] https://nixos.org/manual/nix/stable/command-ref/nix-shell.html

marcus-ares:
https://github.com/firedancer-io/firedancer/actions/runs/4371310819/jobs/7647069297

This check ran on the Firecracker runner.

ripatel-fd:
Very cool. @marcus-jump should I nuke ci2 so we can switch over to Firedancer?

PR #184, opened 2023-03-08
Implement gossip protocol parsers

smcio:
Gossip protocol parsing for review. The core deserialization for the gossip protocol is done (AFAICT). I haven't finished all the clean-up and optimisation yet, but it'd be great to get your comments, so I've pushed this up as a draft.

Items that are still WIP:

- need to create rigorous testsuite
- look at optimisation
- pretty-printer for potential debugging
- compression for CompressedSlots (flate2)
- create bitvec implementation for later use of bitvec payloads after deserialization

PR #198, opened 2023-03-10
ballet: add fd_slice

ripatel-fd:
- import fd_slice from @smcio's PR #184 
- ballet: polish fd_slice


jumpsiegel (approved):

smcio:
@ripatel-jump  I took a look into what we talked about on Discord. That is to say, it'd be desirable for the compiler to produce machine code that carries out a single boundary check for several sequential 'slice' accesses rather than (somewhat naively) carrying out separate boundary checks for each of the 'slice' accesses.

More concretely, when producing code for:
```
a = fd_slice_read_uint  ( s )
b = fd_slice_read_ushort( s )
c = fd_slice_read_uint  ( s )
```
it'd be desirable for the compiler to optimise the boundary checks each time carried out by the logic in `fd_slice_advance` such that the output machine code essentially just does:
```
if( end - ptr >= 10 )     // a single boundary check of 10 (4 + 2 + 4) for all three accesses
  // get uint, get ushort, get uint
```
In order to look into whether the compiler really does optimise the code in this PR in that way, I cloned `ripatel/slice` and did some experimenting and disassembling.

Firstly I removed `-g` from the compile flags (in case this turns off certain optimisations), and then I made `fd_slice_advance` (which is called by each of the `fd_slice_read*` APIs) as `static inline` to persuade the compiler to inline the code. I then moved the function definition into `fd_slice.h` so that `fd_slice_read*` primitives were in the same compilation unit and therefore able to call the function.

I then compiled a testcase:
```
ulong test_optimization( fd_slice_t * slice) {
  uint a = fd_slice_read_uint  ( slice );
  ushort b = fd_slice_read_ushort( slice );
  uint c = fd_slice_read_uint  ( slice );
  return a + b + c;
}
```
I built with `gcc` and then reversed the `test_optimization` function to the following assembly code:
```
.text:0000000000002D40                 endbr64
.text:0000000000002D44                 mov     rdx, [rdi]
.text:0000000000002D47                 mov     rsi, [rdi+8]
.text:0000000000002D4B                 cmp     rdx, 0FFFFFFFFFFFFFFFCh
.text:0000000000002D4F                 lea     rcx, [rdx+4]     // **** load ptr + 4
.text:0000000000002D53                 setnb   r8b
.text:0000000000002D57                 cmp     rsi, rcx      // **** boundary check of end versus ptr + 4
.text:0000000000002D5A                 setb    al
.text:0000000000002D5D                 or      r8b, al
.text:0000000000002D60                 jnz     short loc_2DC0
.text:0000000000002D62                 mov     [rdi], rcx
.text:0000000000002D65                 test    rdx, rdx
.text:0000000000002D68                 jz      loc_2E20
.text:0000000000002D6E                 mov     eax, [rdx]
.text:0000000000002D70                 db      66h, 66h, 2Eh
.text:0000000000002D70                 nop     word ptr [rax+rax+00000000h]
.text:0000000000002D7B                 nop     dword ptr [rax+rax+00h]
.text:0000000000002D80
.text:0000000000002D80 loc_2D80:                               ; CODE XREF: test_optimization+E2↓j
.text:0000000000002D80                 lea     r8, [rdx+6]       // **** load ptr + 6
.text:0000000000002D84                 cmp     rcx, r8       // **** boundary check of end versus ptr + 6
.text:0000000000002D87                 setnbe  r10b
.text:0000000000002D8B                 cmp     rsi, r8
.text:0000000000002D8E                 setb    r9b
.text:0000000000002D92                 or      r10b, r9b
.text:0000000000002D95                 jnz     short loc_2DE0
.text:0000000000002D97                 movzx   r9d, word ptr [rcx]
.text:0000000000002D9B                 lea     rcx, [rdx+0Ah]         // **** load ptr + 10
.text:0000000000002D9F                 cmp     rcx, r8          // **** boundary check on end versus ptr + 10
.text:0000000000002DA2                 setb    r10b
.text:0000000000002DA6                 cmp     rcx, rsi
.text:0000000000002DA9                 setnbe  r8b
.text:0000000000002DAD                 or      r10b, r8b
.text:0000000000002DB0                 jnz     short loc_2E00
.text:0000000000002DB2                 add     eax, r9d
.text:0000000000002DB5                 mov     [rdi], rcx
.text:0000000000002DB8                 add     eax, [rdx+6]
.text:0000000000002DBB                 retn
```
Note the boundary checks above, annotated with '****' comments. As you can see, there are three boundary checks in here; one for the first DWORD, one for the WORD, and another for the final DWORD. 

As such then, the compiler apparently did not see fit to optimise the sequence of three boundary checks & data fetches down to having just one boundary check that covers the whole 10 bytes.

I also built with `clang` and disassembled the resulting function here as well, and there were also three boundary checks in there as well.

I did these tests with `-O3` enabled, as per the existing compile flags in `config/with-optimization.mk`.

asiegel-jt:
This is an overly general solution to a specific problem (gossip protocol). We decided that a more efficient solution it to write/generate custom serializers for the gossip messages so we can optimize performance for that case.

PR #211, opened 2023-03-20
Accelerated bulk SHA256 and SHA512 calculations

kbowers-jump:
Supports high performance streaming and batching of large numbers of variable sized messages.  Optimized for the case of hashing same size or temporally clustered size messages.

Reference and single-threaded SIMD implementations provided.  The SIMD implementation requires FD_HAS_AVX platform capability but does not use any recent SHA instruction set extensions (so it will work on a broad range of Intel and AMD targets).

The user facing part of the API is designed to be easy to swap in (with minimal user code modifications) thread parallel implementations, hybrid thread parallel / data parallel implementation, hardware co-processor implementations (e.g. FPGA accelerated) implementations, etc.

The SIMD implementation is based on the new fangled vector uint / ulong SIMD stuff.  Basically doing what an optimized CUDA implementation would turn into on a GPU under the hood (e.g. Nvidia GPUs are essentially 32-wide SIMD processors under the hood with a very clever programming model / tooling to make it much easier for devs to expose data parallelism to the tools ... while we don't have the clever programming model on x86, we can still implement the equivalent vector ops ... annoyingly just have to write it from the POV of the vector instead of a vector lane).  Same techniques here can be applied to accelerate other algos as necessary.

To give some performance ballpark, SIMD implementation on old (2017 era) Intel core at 3.3 Ghz/4.5 Ghz turbo lying around (10 core Skylake W-2155, RHEL8) was asymptotically (large number of MTU-ish messages of similar sizes) ~12-13 Gbps/core ethernet equivalent for SHA-256 and ~10-11 Gbps/core ethernet equivalent.

The small message independent hash rate was ~19 Mhash/core (SHA-256) and ~9 Mhash/core (SHA-512).  Probably some opportunity to tune it more (this was just a quick translation of the OpenSSL reference implementation followed by one round reorg to eliminate branches / "warp"-level branch divergences) but it is already pretty good.  As a reference (that probably is using some heavily tuned scalar implementations but probably not all the tricks to bulk accelerate operations) stock OpenSSL (1.1.1g) on the same box got ~6 Mhash/core (SHA-256) and ~4.5 Mhash/core (SHA-512).  YMMV.

TL;DR  This can be used to speed up PoH verification and Merkle tree construction and the SHA-512 parts of sig-verify in software significantly and/or use less hardware / cheaper hardware to hit a particular performance target.  The same techniques used to implement this can be mechnically applied to accelerate other crypto algos appearing in ballet similarly.

### DIFF
--- a/src/ballet/sha256/fd_sha256.h
+++ b/src/ballet/sha256/fd_sha256.h
@@ -140,4 +140,202 @@ fd_sha256_hash( void const * data,
 
 FD_PROTOTYPES_END
 
+#if 0 /* SHA256 batch API details */
+
+/* FD_SHA256_BATCH_{ALIGN,FOOTPRINT} return the alignment and footprint
+   in bytes required for a region of memory to can hold the state of an
+   in-progress set of SHA-256 calculations.  ALIGN will be an integer
+   power of 2 and FOOTPRINT will be a multiple of ALIGN.  These are to
+   facilitate compile time declartions. */
+
+#define FD_SHA256_BATCH_ALIGN     ...
+#define FD_SHA256_BATCH_FOOTPRINT ...
+
+/* A fd_sha256_batch_t is an opaque handle for a set of SHA-256
+   calculations. */
+
+struct fd_sha256_private_batch;
+typedef struct fd_sha256_private_batch fd_sha256_batch_t;
+
+/* fd_sha256_batch_{align,footprint} return
+   FD_SHA256_BATCH_{ALIGN,FOOTPRINT} respectively. */
+
+ulong fd_sha256_batch_align    ( void );
+ulong fd_sha256_batch_footprint( void );
+
+/* fd_sha256_batch_init starts a new batch of SHA-256 calculations.  The
+   state of the in-progress calculation will be held in the memory
+   region whose first byte in the local address space is pointed to by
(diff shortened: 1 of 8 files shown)


src/ballet/sha256/fd_sha256.h
CantelopePeel, line 1: Any chance this can find it's way to its own header?

src/ballet/sha512/fd_sha512.h
CantelopePeel, line 1: Ditto

src/ballet/sha256/fd_sha256.h
kbowers-jump: Don't really have a strong opinion about it and it isn't tough to do.  Default inclination is to try to keep all things SHA-256 in one header for simplicity (and similarly for SHA-512 below).  That is, though there are some exceptions, most modules try to have a single header to expose all public facing interfaces in one place.  Not obvious to me that this fits into any the previous exceptions but haven't thought too hard about it.  Curious as to the motivation here.

jumpsiegel: I prefer one header, tbh...  especially since this will most likely be the API I use for almost everything

jumpsiegel (approved):

PR #212, opened 2023-03-20
Implement gossip encoding (not yet fully refactored&cleaned up) + testcases…

smcio:
This PR implements a gossip serializer / encoder. I haven't cleaned this up nor refactored as much as I'd like to yet, but I'll do that in a later changeset. I've also added testcases and made some small changes to the deserializer code.


jumpsiegel:
Making this a draft because there is other work we are about to bring into the public that we can leverage here

ripatel-fd:
Closing since gossip decode/encode has been added to Firedancer runtime branch

PR #220, opened 2023-03-26
Milestone 1.1 cleanup

terorie:
- Adds macOS support
- Various QUIC-related cleanup


src/tango/quic/tls/fd_quic_tls.c
```diff
@@ -256,7 +278,7 @@ fd_quic_tls_hs_delete( fd_quic_tls_hs_t * self ) {
   // find index into array
   ulong hs_idx = (ulong)( self - quic_tls->handshakes );
   if( quic_tls->used_handshakes[hs_idx] != 1 ) {
```
nbridge-jump, line 280: This if-block may be removed

PR #241, opened 2023-04-05
Pass the right argc to fd_tile_exec_new

marcus-ares:
I'm under the impression that argc should be `3` and not `0`.


jumpsiegel (approved):

ripatel-fd:
This is actually intended -- argv does not contain any real cstr arguments.

PR #244, opened 2023-04-06
ballet: Add Reed-Solomon module

ptaffet-jump:
Adds high performance Reed-Solomon encoding and recovery, currently configured for up to 67 data shreds and 67 parity shreds (inclusive). Tuned for GFNI (~130 Gbps on the common 32/32 encoding case), but also relatively high performance on AVX2 (~50 Gbps for 32/32 encoding). Uses the `FD_HAS_GFNI` flag. Depends on the SIMD changes in #250, so we probably want to merge that PR first.


cavemanloverboy:
You have an implementation which is generalized to all d, p != 32, 32
```c
void fd_reedsol_encode_fini( fd_reedsol_t * rs ) {
#if FD_HAS_GFNI
  if( FD_LIKELY( (rs->data_shred_cnt==32UL) & (rs->parity_shred_cnt==32UL ) ) )
    fd_reedsol_encode_32_32( rs->shred_sz, (uchar const * *)rs->data_shred, rs->parity_shred, rs->scratch );
  else
    fd_reedsol_encode( rs->shred_sz, (uchar const * *)rs->data_shred, rs->data_shred_cnt, rs->parity_shred, rs->parity_shred_cnt );
#else
  fd_reedsol_encode( rs->shred_sz, (uchar const * *)rs->data_shred, rs->data_shred_cnt, rs->parity_shred, rs->parity_shred_cnt );
#endif

  rs->data_shred_cnt = 0UL;
  rs->parity_shred_cnt = 0UL;
}
```
but your struct only supports up to `d = FD_REEDSOL_DATA_SHREDS_MAX` and `p = FD_REEDSOL_PARITY_SHREDS_MAX`.
```c
struct __attribute__((aligned(FD_REEDSOL_ALIGN))) fd_reedsol_private {
  uchar scratch[ 1024 ]; /* Used for the ultra high performance implementation */

  /* shred_sz: the size of each shred in bytes (all shreds must be the
     same size) */
  ulong shred_sz;

  /* {data, parity}_shred_cnt: the number of data or parity shreds
     (respectively) have been added to the in-process operation */
  ulong data_shred_cnt;
  ulong parity_shred_cnt;

  /* {data, parity}_shred: pointers to the first byte of each shred */
  uchar * data_shred[   FD_REEDSOL_DATA_SHREDS_MAX   ];
  uchar * parity_shred[ FD_REEDSOL_PARITY_SHREDS_MAX ];

  /* {data, parity}_shred_valid: whether the shred at the corresponding
     index contains valid data.  Used only for decoding operations. */
  //fd_reedsol_shred_set_t data_shred_valid;
  //fd_reedsol_shred_set_t parity_shred_valid;
};
```
Are there any plans for this library to support gf16 (or maybe even gf32...) in the future? A high-performance RS encoder generalized to these cases would be used by many I imagine...

src/ballet/reedsol/fd_reedsol.c
```diff
@@ -0,0 +1,88 @@
+#include "fd_reedsol_private.h"
+
+/* Include the constants in one central spot */
+
+#if FD_REEDSOL_ARITH_IMPL==0
+FD_IMPORT_BINARY( fd_reedsol_arith_consts_generic_mul, "src/ballet/reedsol/constants/generic_constants.bin" );
```
mmcgee-jump, line 6: Can you encode these dependencies into the Makefile

mmcgee-jump (approved):

PR #248, opened 2023-04-11
Installed support for persistent IPC relocatable vectors

kbowers-jump:
... part of the funky review.

This started life as Alex's fd_vector.h from the funky PR.  But since this was generically useful and support for a vector has been asked for by others, made it a generic templatized code generator.

This involved tweaking to match the naming conventions of other code generators (e.g. join returns a result that looks like a linear array from the caller's POV).  Then added support for long term persistence (e.g. process can die, come back and another process can start using the vector), IPC usage (e.g. multiple processes can join and toy with the map ...  dangerously close to be lockfree concurrent for all operations ... not too hard to finish the job if there was demand), relocation (can pick up the map and move it in memory, move it to disk and back, move it to another host and it will "just work"), etc.

The most notable change was the semantics were changed to encourage developers to "do the right thing" when it comes to expand and contract operations (e.g. if in a loop of lots of pushes, just do overflow/underflow outside the loop and bulk expansion/contraction there).  Very easy to accidentally make a fast O(N) operation into a slow O(N^2) operation (and one that doesn't handle errors correctly) ala the discussion in the fd_alloc_max_expand via standard vector push/pop semantics.

Similarly the remote_at API was renamed to remove_compact and a plain remove API as added that works by backfill.  That is, in many vector use cases the order of the elements isn't particularly important and if doing lots of random removes backfilling will be more efficient as the overall will be O(N) instead another accidental O(N^2).

(And, if doing a filtering of all the vector element, stream compaction will be faster still.  And then a branchless implementation of stream compaction will be faster than that ... often several fold.  And then if a big enough vector, doing a parallel branchless stream compaction will be yet faster again.  But all this is more out-of-scope of this simple code generator except to note that a lot of the standard containers dramatically inhibit the ability to actual use the computer well because the they poorly virtualize how the hardware actually works.)

Added a reasonably comprehensive unit tester as well.

(Note, this commit is technically independent of map-giant but they both touch the local makefile.)


asiegel-jt:
This PR accidently included giant_map files. They need to be removed.

kbowers-jump:
Yep ... the fact there are trivial changes for both to a common Local makefile made it them superficially dependent on each other from git's POV.  In any case, now that map-giant is merged, it is probably within the realm of my limited git-fu to clean up the dubious change dependency.

kbowers-jump:
Cleaned up and commit updated accordingly (with some typo cleanups too).

asiegel-jt (approved):

PR #252, opened 2023-04-13
Sandbox implementation

marcus-ares:
This PR introduces a sandbox implementation.

The purpose of the sandbox is to reduce the impact of a Firedancer compromise.

When starting, before executing any task code and/or processing user-provided input, a Firedancer process prepares everything it needs in order to function properly. This initialization mostly consists in:

- [Creating new tiles](https://github.com/firedancer-io/firedancer/blob/fc95073e1de66796732e679fad3a7d29ff0003b9/src/util/tile/fd_tile_threads.cxx#L479) which will be used for task execution (clone syscall)
- Opening the relevant workspaces which will be used to perform work and communicate (mmap syscall)

Immediately after performing those operations, Firedancer sandboxes itself. Note that firedancer has to be started as *root* in order to be able to sandbox itself. Here are the mechanisms currently used by Firedancer to achieve sandboxing:

- clearenv
    - **What:** The environment variable are cleared.
    - **Why:** Environment variables are commonly used to hold secrets. If Firedancer is compromised, no secrets living in the operator’s environment will be leaked.
    - Note: Future desired behavior: `FD_` prefixed environment variables should not be cleared
- fd_jail_setup_netns
    - **What (simplified):** The process looses access to network interfaces.
    - **What:** The process joins a previously configured network namespace. Most processes (exceptions are metrics & RPC) join an empty namespace (loopback is present).
    - **Why:** Principle of least privilege: in the event where the process was able to interact with a network interface, it should not be able to perform any communication.
- fd_jail_setup_mountns
    - **What (simplified):** The process gets a restricted view of the filesystem.
    - **What:** The ends up in a mountns with a root of its own. There is a bind mount at /log for storing logs. (log mount is not yet impl.)
    - **Why:** Principle of least privilege: the process should only be able to interact with files that it needs to function. Although we could do away with the filesystem altogether (passing log file handles through uds), we believe that it’s okay to grant component access to a mount for the purpose of logging.
- fd_jail_set_resource_limits
    - **What:** For the time being, the number of concurrently opened files is set to a low number. In the future, more resources types can be limited.
    - **Why:** Firedancer processes have well understood expected behaviors and resource needs. A process should not be able to exceed those limits, potentially leading to availability issues.
- fd_jail_set_user
    - **What:** The process drops all of its {real and effective} {user and group} ids to those of a predefined user (or those of nobody).
    - **Why:** In the case where another control was to fail, the process should be interacting with the system as an unprivileged user.
- fd_jail_close_non_std_fds
    - **What:** All file descriptors above a specified number are forcefully closed.
    - **Why:** Similar to and more impactful than `clearenv`, an operator’s process can have FDs opened that are 1. not relevant to Firedancer 2. references to sensitive resources. Those resources should not be made available to Firedancer.
- fd_jail_seccomp:
    - **What:** Prevent the usage of most syscall only allowing those explicitly needed by the specific Firedancer component.
    - **Why:** Syscalls are used to interact with the operating system. [There exists close to 400 syscalls](https://github.com/torvalds/linux/blob/v4.17/arch/x86/entry/syscalls/syscall_64.tbl). While running, A Firedancer process requires [14 syscalls](https://github.com/firedancer-io/firedancer/blob/696158aecaefa43d689e585667af534bcf76007b/src/util/sandbox/fd_sandbox.c#L212-L233) out of those 400 in order to perform its functions. Firedancer will crash if it attempts to use a syscall that is not expected.
    - Note: It also happens that the syscalls that Firedancer is using are ubiquitous and well understood. They have stood the test of time (not that time is an ultimate metric for greatness). We have the luxury of disallowing all of the syscalls that might have received less scrutiny.


ripatel-fd (commented):
This is awesome! Would it be possible to split into multiple PRs perhaps? (e.g. starting with seccomp, then doing the namespaces next) I only had the chance to give it a superficial look so far, but added a few comments. Will review more in depth tomorrow.

src/util/sandbox/test_sandbox.c
```diff
@@ -0,0 +1,360 @@
...
+  } else {
+    FD_LOG_NOTICE(( "[PASS] test_setup_netns_null" ));
+  }
+  return failed_cnt ? 1 : 0;
+}
+
+#else
+
+int
+main() {
```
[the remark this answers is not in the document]
marcus-jump: I'll fix the function signature.

I will call fd_boot and fd_halt in the test children if that's ok with you.

marcus-ares:
https://github.com/firedancer-io/firedancer/pull/252#pullrequestreview-1396932066

I'd rather not split it unless you think it provides value or strongly feel that it's how it should happen. I'd prefer spending my time elsewhere if possible.

src/app/frank/fd_frank_init
```diff
@@ -78,6 +78,17 @@ for((verify_idx=0;verify_idx<VERIFY_CNT;verify_idx++)); do
...
 done
 
+# Create the network namespace(s)
+ip netns add fd-$APP-netless
+
+# Create the thread group users
+useradd -MN -s /sbin/nologin fd-$APP-monitor
+useradd -MN -s /sbin/nologin fd-$APP-verify
+useradd -MN -s /sbin/nologin fd-$APP-dedup
+useradd -MN -s /sbin/nologin fd-$APP-pack
```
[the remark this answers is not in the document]
marcus-jump: I initially thought that we'd need users to gate access to different workspaces but it turns out we don't need users for that. Since shmem is access through hugetblfs, the process can mmap the needed pages and then sandbox itself.
It won't be able to:
1- use mmap (seccomp)
2- see the different wksps since they aren't mounted in its new mountns

src/util/sandbox/test_sandbox.c
```diff
@@ -0,0 +1,366 @@
...
+
+  } else { // child
+    fd_sandbox_seccomp( NULL );
+
+    // The call to execl is expected to have the kernel send a
+    // SIGSYS to the process group. The branch should not execute.
+    if ( execl( "/bin/true", "" ) ) {
+      // This could change if we ever switch stance on seccomp violations.
+      FD_LOG_WARNING(( "execl: should not have been executed %s", strerror(errno) ));
+      exit(0);
```
SEJeff, line 217: Is this `exit(0)` superfluous due to the `exit(0)` on line 220?

src/util/sandbox/test_sandbox.c
```diff
@@ -0,0 +1,366 @@
...
+
+    exit(0);
+  }
+
+return -1;
+}
+
+/* test_setup_netns_null ensures that, after calling `fd_sandbox_setup_mountns`,
+   the root mount should be empty and belong to root.
+*/
```
SEJeff, line 228: Similar comments for the other tests would be nice!

src/util/sandbox/test_sandbox.c
```diff
@@ -0,0 +1,366 @@
...
+      FD_LOG_WARNING(( "stat: expected to work on `/`: %s", strerror(errno) ));
+      exit(1);
+    }
+
+    if ( FD_UNLIKELY ( sb.st_uid ) ) {
+      FD_LOG_WARNING(( "root directory should be owned by user root but is owned by uid=%d", sb.st_uid ));
+      exit(1);
+    }
+
+    if ( FD_UNLIKELY ( sb.st_gid ) ) {
```
SEJeff, line 287: If I understand this test correctly, it is just checking the truthiness of the `st_uid` and `st_gid` of the statbuf. Does it make sense to be explicit and check for `sb.st_uid == 0` and `sb.st_gid == 0`?

SEJeff (dismissed):
Really awesome work! Just some minor comments (and I meant to hit Comment not Approve).

src/util/sandbox/test_sandbox.c
marcus-jump: https://pubs.opengroup.org/onlinepubs/009695399/basedefs/stdbool.h.html

It's just checking if it's 0 as 0 is the only "false" value. IIRC, there are a lot of other places in FD where comparisons like this are made so I just consider it idiomatic. I'll defer to @ripatel-jump to know if the correction is needed.

SEJeff: That works. I just like the python approach of "explicit is better than implicit", though I'll always defer to whatever is idiomatic for the language and project. Likely no correction at all.

src/util/sandbox/test_sandbox.c
```diff
@@ -0,0 +1,397 @@
...
+    
+    // Pick a test ns name.
+    char nsname[ 20 ];
+    snprintf(nsname, 20, "fdtestsbx-%u", getpid() );
+    printf("newname: %s\n", nsname);
+
+    // Create the new ns.
+    char cmd[ 64 ];
+    snprintf( cmd, 64, "ip netns add %s", nsname );
+    system(cmd);
```
SEJeff, line 157: nasty

I love it :)

src/app/frank/fd_frank_mon.bin.c
```diff
@@ -355,6 +359,12 @@ main( int     argc,
   ulong **          tile_fseq   = fd_alloca( alignof(ulong *         ), sizeof(ulong *         )*tile_cnt );
   if( FD_UNLIKELY( (!tile_name) | (!tile_cnc) | (!tile_mcache) | (!tile_fseq) ) ) FD_LOG_ERR(( "fd_alloca failed" )); /* paranoia */
   
+  #ifdef FD_HAS_SANDBOX
+  fd_sandbox_profile_t profile;
+  fd_sandbox_profile_init( &profile );
+  fd_sandbox(&profile);
+  #endif /* FD_HAS_JAIL */
```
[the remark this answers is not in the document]
marcus-jump: Bad string replace

SEJeff (dismissed):
Please make sure to get a review from someone else on the team, but this seems sensible to me and I've confirmed the tests build and run successfully!

src/util/sandbox/fd_sandbox.h
```diff
@@ -0,0 +1,30 @@
...
+  struct sock_fprog * seccomp_prog;
+  char *              user;
+};
+typedef struct fd_sandbox_profile fd_sandbox_profile_t;
+
+/* fd_sandbox sandboxes the current process. */
+int fd_sandbox              ( const fd_sandbox_profile_t * const sandbox_profile );
+
+/* fd_sandbox_profile_init initializes a fd_sandbox_profile_t* with sane defaults. */
+void fd_sandbox_profile_init ( fd_sandbox_profile_t * profile );
```
kbowers-jump, line 27: Getting further along here, feels like there wants to be a sandbox profile that just exists by default for each application thread group (e.g. Linux PID).  On targets without FD_HAS_SANDBOX, this profile would be initialized during fd_boot (after the log has gone live) to indicate to the thread group there is no sandboxing.

Similarly, on targets with FD_HAS_SANDBOX, this profile would be set to sane defaults if no parameters were specified (e.g. no sandboxing) and admin defaults parameters passed by the command line (stripped) / environment. otherwise.

Then then would be some additional APIs that allow sandboxing aware code to further restrict their permissions after they do initialization.  (E.g. setup all shared memory pages and then drop various syscall permissions so that running code can't change any internal memory region firewalling.)

(The boot and and app requested changes should be extensively logged as well to aid in debugging sandboxing, understanding fine grained permissions, etc.)

src/util/sandbox/fd_sandbox_private.h
```diff
@@ -0,0 +1,17 @@
+#ifdef FD_HAS_SANDBOX
```
kbowers-jump, line 1: Ditto

src/util/sandbox/fd_sandbox_private.h
```diff
@@ -0,0 +1,17 @@
...
+
+#ifndef HEADER_fd_src_util_sandbox_fd_sandbox_private_h
+#define HEADER_fd_src_util_sandbox_fd_sandbox_private_h
+
+#include <linux/filter.h>
```
kbowers-jump, line 6: A private internal header makes more sense here for an include like this.  But this is also redundant with the previous header.  And generally private headers are used for all sorts of implementation specific details that should not be public but are needed to coordinate across multiple compile units.  So, there is a good chance this doesn't need to be here either as per the previous discussion.

src/util/sandbox/fd_sandbox_private.h
```diff
@@ -0,0 +1,17 @@
...
+#include <pwd.h>
+
+void fd_sandbox_change_user         ( struct passwd * u );
+void fd_sandbox_close_fds_beyond    ( uint max_fd );
+void fd_sandbox_set_resource_limits ( uint max_open_fds );
+void fd_sandbox_setup_netns         ( char * ns_name );
+void fd_sandbox_setup_mountns       ( char * chroot_path );
+void fd_sandbox_seccomp             ( struct sock_fprog * prog );
```
kbowers-jump, line 14: Not clear if these APIs should be private given the above.  E.g. could have the defaults set during boot.  And then if an application ways to do additional restrictions, let them (and just have these APIs fail and log attempts to request less restrictive permissions than they currently have).

See note previously about "struct sock_fprog".

fd_sandbox_change_user might have implications on fd_log_user functionality there.  Need some thought there what this implies.

src/util/sandbox/test_sandbox.c
```diff
@@ -0,0 +1,396 @@
+#ifndef FD_HAS_SANDBOX
+#define FD_HAS_SANDBOX
+#endif
```
kbowers-jump, line 3: See notes above about #ifndef, fd_util_base.c, etc.

src/util/sandbox/test_sandbox.c
```diff
@@ -0,0 +1,396 @@
...
+
+#ifdef FD_HAS_SANDBOX
```
kbowers-jump, line 5: Likewise, this is probably unnecessary given the tweaks above and perhap usage of FD_HAS_SANDBOX in the Local.mk (such that the test is only built on targets where it is necessary).

src/util/sandbox/test_sandbox.c
```diff
@@ -0,0 +1,396 @@
...
+
+#define _GNU_SOURCE
+
+#include "../fd_util.h"
+#include "fd_sandbox.h"
+#include "fd_sandbox_private.h"
```
kbowers-jump, line 11: Aspirationally, only the top level include (e.g. fd_util.h) would be necessary for a unit tester and and the APIs to test would be such that all functionality is testable public facing APIs.

Given all the low level OS hacker, this is probably not possible here.  But, we probably can get very close and at a minimum here, we shouldn't have to include fd_sandbox.h and fd_sandbox_private.h  (e.g. would it ever make sense to just include fd_sandbox_private.h standalone ... doubt it ... generally private APIs imply the public API will be available too).

src/util/sandbox/fd_sandbox.c
```diff
@@ -0,0 +1,256 @@
+#ifdef FD_HAS_SANDBOX
+
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+
+#include "fd_sandbox.h"
+#include "fd_sandbox_private.h"
```
kbowers-jump, line 8: See notes about private implies public.

src/util/sandbox/fd_sandbox.c
```diff
@@ -0,0 +1,256 @@
...
+#include <sched.h>        /* CLONE_*, setns, unshare */
+#include <stddef.h>
+#include <stdio.h>        /* snprintf */
+#include <stdlib.h>       /* clearenv, mkdtemp*/
+#include <sys/mount.h>    /* MS_*, MNT_*, mount, umount2 */
+#include <sys/prctl.h>
+#include <sys/resource.h> /* RLIMIT_*, rlimit, setrlimit */
+#include <sys/stat.h>     /* mkdir */
+#include <sys/syscall.h>  /* SYS_* */
+#include <unistd.h>       /* set*id, sysconf, close, chdir, rmdir syscall */
```
kbowers-jump, line 24: Willing to bet there is a bunch of cleanups and compression that can be done here.  (Also, the linux/filter.h is redundant with previous, but this is actually the place where it actually should be.)

src/util/sandbox/fd_sandbox.c
```diff
@@ -0,0 +1,256 @@
...
+
+#include "../log/fd_log.h"
```
kbowers-jump, line 26: Should be done by public facing API as per previous comments.

src/util/sandbox/fd_sandbox.c
```diff
@@ -0,0 +1,256 @@
...
+  // setregid before setreuid otherwise setregid won't be allowed
+
+  if ( FD_UNLIKELY( setregid( u->pw_gid, u->pw_gid ) ) ) {
+    FD_LOG_ERR(( "setregid: %s", strerror( errno ) ));
+  }
+
+  if ( FD_UNLIKELY( setreuid( u->pw_uid, u->pw_uid ) ) ) {
+    FD_LOG_ERR(( "setreuid: %s", strerror( errno ) ));
+  }
+}
```
kbowers-jump, line 85: See notes about fd_log_user.  No objection here but how fd_log_user behaves in conjunction with this API should be sorted.

src/util/sandbox/fd_sandbox.c
```diff
@@ -0,0 +1,256 @@
...
+
+#include "../log/fd_log.h"
+
+#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
+
+void
+fd_sandbox_profile_init( fd_sandbox_profile_t * profile ) {
+  profile->initialized      = true;
+  profile->chroot_path      = NULL;
+  profile->netns            = "fd-frank-netless",
```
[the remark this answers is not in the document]
kbowers-jump: Also, if sandbox is util, the netns name probably should be be frank app specific.  This is probably something that comes from the command line / environment on fd_boot (with maybe an option to set it programmatically post boot if it wasn't already set on fd_boot).

src/app/frank/fd_frank_main.c
```diff
@@ -115,6 +119,13 @@ main( int     argc,
 
   } while(0);
 
+  #ifdef FD_HAS_SANDBOX
+  fd_sandbox_profile_t profile;
+  fd_sandbox_profile_init( &profile );
+
+  fd_sandbox(&profile);
+  #endif /* FD_HAS_SANDBOX */
+
```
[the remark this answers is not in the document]
marcus-jump: I don't think that we want to automatically sandbox everything from the start. The initialization might require more permissions as it will do things such as mapping other workspaces or setting up xdp. We could probably have two sandboxing steps (pre-init, post-init), but as long as we can consider init "safe", we can just live with post-init sandboxing.

config/with-sandbox.mk
```diff
@@ -0,0 +1 @@
+CFLAGS+=-DFD_HAS_SANDBOX=1
```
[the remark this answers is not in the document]
mmcgee-jump: Really don't like sandboxing being optional or having "HAS_SANDBOX". Everything should be sandboxed 100% of the time, it's part of the application.

mmcgee-jump (commented):
Took an initial pass -- some comments inline

mmcgee-jump (changes requested):
Took an initial pass -- some comments inline

src/util/sandbox/fd_sandbox.c
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  // todo: maybe not clearenv but only keep vars prefixed with FD_
+  clearenv();
+  fd_sandbox_setup_netns( profile->netns );
+  fd_sandbox_setup_mountns( profile->chroot_path );
+  fd_sandbox_set_resource_limits( profile->max_open_fds );
+  fd_sandbox_change_user( uinfo );
+  fd_sandbox_close_fds_beyond( profile->close_fds_beyond );
+  fd_sandbox_seccomp( profile->seccomp_prog );
```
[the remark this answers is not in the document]
marcus-jump: I was on the fence about doing it but yes, it costs us nothing.

src/app/frank/fd_frank_init
```diff
@@ -78,6 +79,14 @@ for((verify_idx=0;verify_idx<VERIFY_CNT;verify_idx++)); do
     || exit $?
 done
 
+if [ $INIT_SANDBOX = "1" ];
+then
+  # Create the network namespace(s)
+  ip netns add fd-$APP-netless
+  # Create the thread group users
+  useradd -MN -s /sbin/nologin fd-$APP
```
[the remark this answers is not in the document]
marcus-jump: The logs that end up on the host system need to be owned by a sensible user. That can all go away if / when we change how logs are being emitted. We can cross that bridge once we get there.

config/with-sandbox.mk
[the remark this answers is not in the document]
marcus-jump: Kevin and I also sync on it last week. Some platforms won't support sandboxing. That said, there should be a generic sandbox implementation for unsupported platforms that just makes you pass a flag that makes the operator acknowledge that they are doing something bad.

src/util/sandbox/fd_sandbox.c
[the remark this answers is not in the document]
mmcgee-jump: I realized this will be a small problem at the moment since this is how QUIC variables are passed at the moment.

marcus-ares:
https://app.codecov.io/github/firedancer-io/firedancer/commit/a7c6c86839e56e375ef1b320a7191fe73b3cd13b

Looks like coverage of forked processes is not computed.

src/util/sandbox/fd_sandbox.c
```diff
@@ -0,0 +1,256 @@
...
+
+  FD_LOG_INFO(( "thread group sandboxed" ));
+  return 0;
+}
+
+void
+fd_sandbox_change_user( struct passwd * u ) {
+  errno = 0;
```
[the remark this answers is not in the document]
marcus-jump: artifact of debugging :)

mmcgee-jump:
Can you also add a `security.md` to the docs folder describing the security architecture? Could just be basically the same thing as this change overview.

marcus-ares:
@mmcgee-jump -- This is planned for an upcoming PR. Thanks for mentioning it!

src/util/sandbox/fd_sandbox_linux_private.h
```diff
@@ -0,0 +1,82 @@
+#ifndef HEADER_fd_src_util_sandbox_fd_sandbox_linux_h
+#define HEADER_fd_src_util_sandbox_fd_sandbox_linux_h
+
+/* In order to be able to sandbox itself, Firedancer has to be started
```
mmcgee-jump, line 4: I feel like this is just going to get out of sync, better in a `security-architecture.md` file and then point to that from here.

src/util/sandbox/fd_sandbox.h
```diff
@@ -0,0 +1,43 @@
...
+   even with infinite resources, the number of memory corruption bugs in
+   Firedancer at any given time is non-zero and that an exploiter will in 
+   some future attempt to weaponize such a bug in main net.
+
+   When starting, before executing any task code and/or processing
+   user-provided input, a Firedancer process prepares everything it needs 
+   in order to function properly.
+
+   This initialization mostly consists in:
+     - [Creating new tiles](https://github.com/firedancer-io/firedancer/blob/fc95073e1de66796732e679fad3a7d29ff0003b9/src/util/tile/fd_tile_threads.cxx#L479)
```
mmcgee-jump, line 33: Don't link to github here, just reference the code itself.

src/app/frank/fd_frank_main.c
```diff
@@ -35,6 +37,7 @@ fd_frank_signal_trap( int sig ) {
   if( FD_UNLIKELY( sigaction( sig, act, NULL ) ) ) FD_LOG_ERR(( "unable to override signal %i", sig ));
 }
 
+
```
mmcgee-jump, line 40: nit: extra whitespace

src/util/sandbox/fd_sandbox_linux.c
```diff
@@ -0,0 +1,281 @@
...
+
+void
+fd_sandbox_setup_netns( char * ns_name ) {
+  char netns_path[ PATH_MAX ];
+  int nsfd;
+
+  FD_LOG_INFO(( "setting up network namespace" ));
+
+  // realize the namespace path
+  int reslen = snprintf( netns_path, ARRAY_SIZE(netns_path), "/var/run/netns/%s", ns_name );
```
mmcgee-jump, line 137: Not sure why the network namespace needs to be named and set up / destoyed by the init script? I think you can just unshare a new NETNS here and have it be anonymous?

src/app/frank/fd_frank_init
```diff
@@ -78,6 +78,14 @@ for((verify_idx=0;verify_idx<VERIFY_CNT;verify_idx++)); do
     || exit $?
 done
 
+if command -v ip
+then
+  # Create the network namespace(s) (requires iproute2)
+  ip netns add fd-$APP-netless
```
mmcgee-jump, line 84: See comment where this gets used

src/util/sandbox/fd_sandbox_linux.c
```diff
@@ -0,0 +1,281 @@
...
+
+int  fd_sandbox_linux ( const fd_sandbox_profile_linux_t * profile );
+
+void
+fd_sandbox_profile_init_common( fd_sandbox_profile_linux_t * profile ) {
+  profile->initialized      = true;
+  profile->chroot_path      = NULL;
+  profile->netns            = "fd-frank-netless",
+  profile->seccomp_prog     = NULL;
+  profile->user             = "nobody";
```
mmcgee-jump, line 37: This also should be an anonymous user, not named or set up before hand. See chatgpt response for a question like "Can I create an anonymous user  in C code that isn't named, for use with a user namespace?"

src/util/sandbox/fd_sandbox.h
```diff
@@ -0,0 +1,43 @@
+#ifndef HEADER_fd_src_util_sandbox_fd_sandbox_h
+#define HEADER_fd_src_util_sandbox_fd_sandbox_h
+
+#include "../log/fd_log.h"
+
+enum fd_sandbox_profile {
+  FD_SANDBOX_PROFILE_COMMON,
+  FD_SANDBOX_PROFILE_XDP
```
mmcgee-jump, line 8: This is apparently not used anywhere?

src/util/sandbox/fd_sandbox_linux.c
```diff
@@ -0,0 +1,281 @@
...
+  if ( FD_UNLIKELY( setns( nsfd, CLONE_NEWNET ) == -1 ) ) {
+    FD_LOG_ERR(( "netns setns: %s", strerror( errno ) ));
+  };
+
+  // close the namespace reference
+  close( nsfd );
+}
+
+void fd_sandbox_drop_ambient_caps( void ) {
+  if ( FD_UNLIKELY( prctl( PR_CAP_AMBIENT, PR_CAP_AMBIENT_CLEAR_ALL, 0, 0, 0 ) ) ) {
```
mmcgee-jump, line 157: Don't you also need to remove all effective caps from the current process as well?

src/util/sandbox/fd_sandbox_linux.c
```diff
@@ -0,0 +1,281 @@
...
+    FD_LOG_ERR(( "prctl( PR_CAP_AMBIENT, PR_CAP_AMBIENT_CLEAR_ALL, ... ): %s", strerror( errno ) ));
+  }
+}
+
+void 
+fd_sandbox_setup_mountns( char * chroot_path ) {
```
mmcgee-jump, line 163: It doesn't appear this argument is ever non-null? Can we just remove it?

mmcgee-jump (changes requested):
Few more comments inline

src/util/test_util_base.c
```diff
@@ -6,13 +6,23 @@
...
+FD_STATIC_ASSERT( IMPLIES( FD_HAS_SSE, FD_HAS_X86   ), devenv );
+FD_STATIC_ASSERT( IMPLIES( FD_HAS_AVX, FD_HAS_SSE   ), devenv );
+FD_STATIC_ASSERT( IMPLIES( FD_HAS_SHANI, FD_HAS_AVX ), devenv );
+FD_STATIC_ASSERT( IMPLIES( FD_HAS_GFNI, FD_HAS_AVX  ), devenv );
+
+FD_STATIC_ASSERT( IMPLIES( FD_HAS_THREADS, FD_HAS_ATOMIC ), devenv );
+FD_STATIC_ASSERT( IMPLIES( FD_HAS_THREADS, FD_HAS_HOSTED ), devenv );
+
+FD_STATIC_ASSERT( IMPLIES( FD_HAS_SSE, FD_HAS_X86 ), devenv );
+FD_STATIC_ASSERT( IMPLIES( FD_HAS_AVX, FD_HAS_SSE ), devenv );
```
barnjamin, line 20: duplicate?

src/util/sandbox/fd_sandbox_linux_private.h
marcus-jump: I can remove it once we have security-architecture.md.

src/util/sandbox/fd_sandbox_linux.c
marcus-jump: You could decide to have it live in a preconfigured netns. It's probably what the RPC task is going to require.

src/util/sandbox/fd_sandbox.h
marcus-jump: Oh yes, true! We need to talk about. We might need to have a special mode for the quic tile when the driver operates in skb mode. The current reference is inaccurate and I'll just remove it for now.

src/util/sandbox/fd_sandbox_linux.c
marcus-jump: I'm revisiting this whole thing. Thanks for pointing it out! I think that we need to sort out the tile zero issue out first before we can handle caps during init properly.

src/util/sandbox/fd_sandbox_linux.c
marcus-jump: It is not right now but it was made for making it easy for users to pick where the chroot lives. In hindsight, this is an operator concern.

I have greater plans for how we handle mountns in the future so I'm not against removing this argument.

marcus-jump: Argument removed :)

PR #262, opened 2023-04-14
Installed funk fork layer

kbowers-jump:
Extracted as part of the review of Alex's funk store.

These are the APIs for managing forks.  At this point, this layer is all minimum time algorithmically (fast O(number of transactions impacted)), minimum space algorithmically (small O(1)), no allocation required, no operating system overheads calls, minimal TLB overheads and produces no garbage to collect out of the critical path.  It has extensive logging, user error, and memory corruption protections.  APIs are designed to allow users to be able to just worry about the head of the forks they are tracking at most and it will all "just work".  That is, it can simply, reliably and scalably track arbitrarily complex transactions histories in adverse environments until it runs out of physical resources reserved for it.

More layers to come ...


src/funk/fd_funk_txn.c
```diff
@@ -0,0 +1,723 @@
...
+    } else { /* No younger sib and has parent */
+      ASSERT_IN_PREP( parent_idx );
+      map[ parent_idx ].child_tail_cidx = fd_funk_txn_cidx( sibling_prev_idx );
+    }
+  } else { /* Has younger sib */
+    ASSERT_IN_PREP( sibling_next_idx );
+    map[ sibling_next_idx ].sibling_prev_cidx = fd_funk_txn_cidx( sibling_prev_idx );
+  }
+
+  fd_funk_txn_map_remove( map, &map[ txn_idx ].id );
```
asiegel-jt, line 225: You have this form of remove in several places. Let's consider an alternate remove that takes "map + txn_idx" as the argument instead. Might be possible to optimize the map a bit for this case.

src/funk/fd_funk_txn.c
```diff
@@ -0,0 +1,723 @@
...
+
+  funk->child_head_cidx = fd_funk_txn_cidx( child_head_idx );
+  funk->child_tail_cidx = fd_funk_txn_cidx( child_tail_idx );
+
+  /* Remove the mapping */
+
+  fd_funk_txn_id_copy( funk->last_publish, &map[ txn_idx ].id );
+
+  fd_funk_txn_map_remove( map, &map[ txn_idx ].id );
+
```
asiegel-jt, line 492: Should put a comment in the places where we expect to have to clean up transaction content. Lines 225 and 491?

src/funk/fd_funk_txn.c
```diff
@@ -0,0 +1,723 @@
...
+  /* Cancel all transactions and their descendants on the cancel stack */
+
+  ulong cancel_cnt = 0UL;
+
+  while( !fd_funk_txn_idx_is_null( cancel_stack_idx ) ) { /* TODO: peel first iter to make more predictable? */
+    ulong sibling_idx = cancel_stack_idx;
+    cancel_stack_idx  = fd_funk_txn_idx( map[ sibling_idx ].stack_cidx );
+
+    cancel_cnt += fd_funk_txn_cancel_family( funk, map, txn_max, tag, sibling_idx );
```
asiegel-jt, line 377: Seems like you are using the same stack for this function as well as the interior of cancel_family. Is this correct?

src/funk/fd_funk_txn.h
```diff
@@ -0,0 +1,404 @@
...
+   would publish as much currently uncontested transaction history as
+   possible around txn.
+
+   Assumes map == fd_funk_txn_map( funk, fd_funk_wksp( funk ) ), funk is
+   a current local join and txn points to an in-preparation transaction
+   of funk in the caller's address space.  The lifetime of the returned
+   pointer is as described in fd_funk_txn_query. */
+
+FD_FN_PURE static inline fd_funk_txn_t *
+fd_funk_txn_ancestor( fd_funk_txn_t * txn,
```
asiegel-jt, line 231: Are ancestor and descendant used for anything? Not sure I see the need for these

asiegel-jt (approved):

src/funk/fd_funk_txn.c
```diff
@@ -0,0 +1,723 @@
...
+
+  if( FD_UNLIKELY( fd_funk_txn_map_is_full( map ) ) ) {
+    if( FD_UNLIKELY( verbose ) ) FD_LOG_WARNING(( "too many transactions in preparation" ));
+    return NULL;
+  }
+
+  ulong txn_max = fd_funk_txn_map_key_max( map );
+
+  ulong  parent_idx;
+  uint * _child_head_cidx;
```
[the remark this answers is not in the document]
kbowers-jump: It isn't a child_head_cidx / child_tail_cidx but a pointer to the relevant one to use for the prepare.  Would rather give it a trivially related name than something completely given they are closely related.  No strong preferences how that happens outside of keeping the prefix / suffix to a small O(1) character.

src/funk/fd_funk_txn.c
kbowers-jump: Wrote some verbiage in fd_map_giant.h about that (might have blown it away before check in though).

In short, I agree.

In long, that's how the alternative maps already work and for similar reasons.  So when harmonizing the APIs between the new map_giant and the old maps, took a quick stab at doing this.  It was trickier than expected because the same that gives map_giant its properties that make it very useful here.

There were a bunch of possible implementations to do this.  The main thing is getting a way go from a map + idx to the original map in fast O(1) time with small O(1) memory.  The most straightforward is to have the elements themselves encode their index in the field directly.  Costs 8 bytes extra per element and didn't want to add the bloat given the design goals you described previously.  But the current layout with index compression has saved enough that we probably have room.

There are ways to do without any memory overhead at all in the element but they are a lot tougher to use practically (largely playing games with alignment and shared memory mappings) that are probably more mischievous than useful for a generic code generator.

src/funk/fd_funk_txn.c
kbowers-jump: Yep ... saves space and it is a security measure as the families of two different siblings should never share a member.  So this allows the ASSERT_UNTAGGED stuff to detect if there are any such collisions between siblings and their families.

src/funk/fd_funk_txn.c
kbowers-jump: Sounds like a good idea ... I'll include it in the check in of the next layer so as to save ping-pong-ing here.

src/funk/fd_funk_txn.h
kbowers-jump: Not yet.  These are just realizations that seemed useful of concepts that were talked about in the documentation (e.g. prepare / cancel / publish talk a lot about ancestors, descendants, uncontested histories, etc).  Seemed like it might be useful for generic fork management to give the user the ability to ask questions like "will publishing this cancel any competing transaction histories" or hopefully help user deal with lots of intermediate transactions if there are long unresolved forks.  I have no particular love of these functions though.

src/funk/fd_funk_txn.c
kbowers-jump: (Actually comments are already there are 191 and line 458.)

PR #263, opened 2023-04-14
redblack tree template

asiegel-jt:


lheeger-jump (approved):
I am approving this PR as it is well tested but I object to some of the formatting which will be caught by linting later on.

PR #281, opened 2023-04-21
Installed funk record layer

kbowers-jump:
Extracted as part of the review of Alex's funk store.

This is the APIs and supporting for creating, updating, reverting, destroying records in a funk transaction.  No dynamic allocations, no OS calls, NUMA and TLB optimized, persistent, relocatable, zero copy, all operations are fast O(1) in time and in O(1) in space (with one uncommon user operation that is fast O(ancestor count) in time and O(1) in space), fortified against memory corruption (including cycles), extensive test and verification suites.

More to come.


src/funk/fd_funk_txn.c
```diff
@@ -455,9 +487,105 @@ fd_funk_txn_publish_funk_child( fd_funk_t *     funk,
...
+
+      /* Erase (root,key) */
+
+      /* Free root_rec val resources here */
+
+      ulong prev_idx = root_rec->prev_idx;
+      ulong next_idx = root_rec->next_idx;
 
-  /* At this point, txn has been committed to permanent storage */
+      if( FD_UNLIKELY( fd_funk_rec_idx_is_null( prev_idx ) ) ) funk->rec_head_idx           = next_idx;
```
asiegel-jt, line 538: Is rec_head_idx and rec_tail_idx initialized to null before this in this function?

src/funk/fd_funk_txn.c
```diff
@@ -612,28 +740,31 @@ fd_funk_txn_merge( fd_funk_t *     funk,
   parent->child_head_cidx   = fd_funk_txn_cidx( FD_FUNK_TXN_IDX_NULL );
```
asiegel-jt, line 740: Need to move records from child to parent here

kbowers-jump: That's the value layer to come.  (And actually, they don't need to be moved.  Just the gaddrs to them.)

src/funk/fd_funk.c
```diff
@@ -65,17 +66,47 @@ fd_funk_new( void * shmem,
...
+  void * rec_shmap = fd_funk_rec_map_new( rec_shmem, rec_max, seed );
+  if( FD_UNLIKELY( !rec_shmap ) ) {
+    FD_LOG_WARNING(( "fd_funk_rec_map_new failed" ));
+    fd_wksp_free_laddr( rec_shmem );
+    fd_wksp_free_laddr( fd_funk_txn_map_delete( fd_funk_txn_map_leave( txn_map ) ) );
+    return NULL;
+  }
+
+  fd_funk_rec_t * rec_map = fd_funk_rec_map_join( rec_shmap );
+  if( FD_UNLIKELY( !rec_shmap ) ) {
```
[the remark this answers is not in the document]
kbowers-jump: Yep, will fix.

src/funk/fd_funk_txn.c
kbowers-jump: (It's all zero copy.  There should be annotations where the value layer operations will be inserted and what those operations need to be.)

src/funk/fd_funk_txn.c
kbowers-jump: They are initialized to NULL at funk creation and they are a list of all the records in the last published transaction at this point.  Which could be NULL or could be something.

src/funk/fd_funk_txn.c
kbowers-jump: See line 579 for the relevant annotation to this.

src/funk/fd_funk.c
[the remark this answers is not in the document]
kbowers-jump: Fixed

asiegel-jt (approved):

PR #293, opened 2023-04-26
Backport XDP

ripatel-fd:
Backports XDP implementation from milestone-1.1-demo.

- ebpf: add program loader (remove libbpf)
- net: fix fd_cstr_to_ip4_addr
- xdp: various xsk fixes
- xdp: update redirect prog
- xdp: add fd_xdp_ctl
- xdp: add more tests

### DIFF
--- a/config/base.mk
+++ b/config/base.mk
@@ -26,10 +26,6 @@ LLVM_PROFDATA?=llvm-profdata
 # lcov
 GENHTML=genhtml
 
-EBPF_CC:=clang
-EBPF_CPPFLAGS:=-target bpf -O2 -g
-EBPF_CFLAGS:=-std=c17
-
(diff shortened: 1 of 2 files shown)


ripatel-fd:
@marcus-jump @nbridge-jump I've made a few changes that will make this PR easier to merge:

- Removed build system support for the XDP blob and instead checked it into the repo. The XDP blob can still be rebuilt if needed with the new `build-xdp.sh` script.
- Removed and reverted all modifications to the Make build system
- Removed build dependency on Clang. This PR will now build on any machine that can build the current main.

Since this PR contains important bug fixes for XDP and CI, I would appreciate an approval to merge into main soon. This would also unblock further Turbine and gossip work.

jumpsiegel (approved):

PR #295, opened 2023-04-26
contrib: add ubi9 Dockerfile

ripatel-fd:
adds ubi9-based Dockerfile. Requires https://github.com/firedancer-io/firedancer/pull/293


jumpsiegel (approved):
We probably want to eventually set the hashes for these docker images we are pulling from

marcus-jump (approved):

PR #303, opened 2023-04-28
Add sBPF ELF loader

ripatel-fd:
Adds ELF loader that doesn't yet fully match solana-labs/rbpf, but is able to load most on-chain programs.

- [x] elf: add static assertions
- [x] add Murmur3-32 hash (see #304)
- [x] sbpf: add ELF loader
- [x] rodata segment
- [x] Executable section export
- [x] Fixup relative calls
- [x] Entrypoint


ripatel-fd:
sBPF ELF loader is now able to load SPL Token Program (TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA). Ready for review. @lheeger-jump

ripatel-fd:
@lheeger-jump Synced with `ripatel/vm` and ready for review. Suggest we bring in `fd_murmur3` to main and change whatever needs to change afterwards.

PR #324, opened 2023-05-05
ballet: add ChaCha20Rng

AlberErre:
This PR adds a `ChaCha20Rng` module to generate random number by using ChaCha20 algorithm 

It uses `openssl/evp.h` to implement ChaCha20

A test case (`test_fd_chacha20_rng.c`) has been added to verify its functionality, as well as a `README` explaining how to reproduce this on a MacOS machine

### DIFF
--- a/src/ballet/chacha20/README.md
+++ b/src/ballet/chacha20/README.md
@@ -1,0 +1,15 @@
+# fd_chacha20_rng
+
+## Testing (MacOS)
+
+Build binary by using the following command
+
+```
+gcc -o test_fd_chacha20_rng test_fd_chacha20_rng.c fd_chacha20_rng.c -I/usr/local/opt/openssl/include -L/usr/local/opt/openssl/lib -lcrypto
+```
+
+Then, run test to generate random numbers (displayed on console output)
+
+```
+./test_fd_chacha20_rng
+```
--- a/src/ballet/chacha20/fd_chacha20.h
+++ b/src/ballet/chacha20/fd_chacha20.h
@@ -1,0 +1,14 @@
+#ifndef HEADER_fd_src_ballet_chacha20_fd_chacha20_h
+#define HEADER_fd_src_ballet_chacha20_fd_chacha20_h
+
+#include <openssl/evp.h>
+
+typedef unsigned int fd_chacha20_rng_t;
+
+#define FD_CHACHA20_KEY_SIZE 32   // 32 bytes, 256 bits
+#define FD_CHACHA20_NONCE_SIZE 12 // 12 bytes, 96 bits
+
+int fd_chacha20_rng_init(unsigned char *key, unsigned char *nonce);
+int fd_chacha20_rng_get_uint32(fd_chacha20_rng_t *num);
+
+#endif /* HEADER_fd_src_ballet_chacha20_fd_chacha20_h */
--- a/src/ballet/chacha20/fd_chacha20_rng.c
+++ b/src/ballet/chacha20/fd_chacha20_rng.c
@@ -1,0 +1,33 @@
+#include "fd_chacha20.h"
+
+static EVP_CIPHER_CTX *ctx = NULL;
+
+int fd_chacha20_rng_init(unsigned char *key, unsigned char *nonce)
+{
+  // Initialize the ChaCha20 context with the key and nonce
+  if (!(ctx = EVP_CIPHER_CTX_new()) ||
+      !EVP_EncryptInit_ex(ctx, EVP_chacha20(), NULL, key, nonce))
+  {
+    fprintf(stderr, "Error: EVP_EncryptInit_ex()\n");
+    return 1;
+  }
+
+  return 0;
+}
+
+int fd_chacha20_rng_get_uint32(fd_chacha20_rng_t *num)
+{
+  unsigned char buf[4];
+  int outlen;
+
+  // Generate a random 32-bit number
+  if (!EVP_EncryptUpdate(ctx, buf, &outlen, buf, sizeof(buf)))
+  {
+    fprintf(stderr, "Error: EVP_EncryptUpdate()\n");
+    return 1;
+  }
+
+  *num = *((unsigned int *)buf);
+
+  return 0;
+}
--- a/src/ballet/chacha20/test_fd_chacha20_rng.c
+++ b/src/ballet/chacha20/test_fd_chacha20_rng.c
@@ -1,0 +1,60 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <assert.h>
+#include "fd_chacha20.h"
+
+#define TEST_KEY_SIZE FD_CHACHA20_KEY_SIZE
+#define TEST_NONCE_SIZE FD_CHACHA20_NONCE_SIZE
+
+/* Test fd_chacha20_rng_init function */
+void test_fd_chacha20_rng_init()
+{
+  unsigned char key[TEST_KEY_SIZE];
+  unsigned char nonce[TEST_NONCE_SIZE];
+
+  /* Generate key and nonce using /dev/urandom */
+  FILE *fp = fopen("/dev/urandom", "rb");
+  fread(key, sizeof(key), 1, fp);
+  fread(nonce, sizeof(nonce), 1, fp);
+  fclose(fp);
+
+  /* Initialize the random number generator */
+  assert(fd_chacha20_rng_init(key, nonce) == 0);
+}
+
+/* Test fd_chacha20_rng_get_uint32 function */
+void test_fd_chacha20_rng_get_uint32()
+{
+  unsigned char key[TEST_KEY_SIZE];
+  unsigned char nonce[TEST_NONCE_SIZE];
+  fd_chacha20_rng_t num1, num2;
+
+  /* Generate key and nonce using /dev/urandom */
+  FILE *fp = fopen("/dev/urandom", "rb");
+  fread(key, sizeof(key), 1, fp);
+  fread(nonce, sizeof(nonce), 1, fp);
+  fclose(fp);
+
+  /* Initialize the random number generator */
+  assert(fd_chacha20_rng_init(key, nonce) == 0);
+
+  /* Generate two random numbers and make sure they are different */
+  assert(fd_chacha20_rng_get_uint32(&num1) == 0);
+  assert(fd_chacha20_rng_get_uint32(&num2) == 0);
+  assert(num1 != num2);
+
+  /* Print random numbers */
+  printf("%u\n", num1);
+  printf("%u\n", num2);
+}
+
+int main()
+{
+  test_fd_chacha20_rng_init();
+  test_fd_chacha20_rng_get_uint32();
+
+  printf("All fd_chacha20_rng tests passed!\n");
+
+  return 0;
+}
--- a/src/ballet/chacha20/Local.mk
+++ b/src/ballet/chacha20/Local.mk
@@ -1,0 +1,5 @@
+$(call add-hdrs,fd_chacha20.h)
+$(call add-objs,fd_chacha20_rng)
+
+$(call make-unit-test,test_fd_chacha20_rng,test_fd_chacha20_rng,fd_ballet fd_util)
+$(call run-unit-test,test_fd_chacha20_rng)


src/ballet/chacha20/test_fd_chacha20_rng.c
```diff
@@ -0,0 +1,60 @@
+#include <stdio.h>
```
ptaffet-jump, line 1: If you need file IO, it should be protected with `#if FD_HAS_HOSTED`, but I don't think this test actually needs file IO.

src/ballet/chacha20/test_fd_chacha20_rng.c
```diff
@@ -0,0 +1,60 @@
...
+#include <assert.h>
+#include "fd_chacha20.h"
+
+#define TEST_KEY_SIZE FD_CHACHA20_KEY_SIZE
+#define TEST_NONCE_SIZE FD_CHACHA20_NONCE_SIZE
+
+/* Test fd_chacha20_rng_init function */
+void test_fd_chacha20_rng_init()
+{
+  unsigned char key[TEST_KEY_SIZE];
```
ptaffet-jump, line 13: I don't know if we have a public comprehensive style guide yet, but these should be `uchar`

src/ballet/chacha20/test_fd_chacha20_rng.c
```diff
@@ -0,0 +1,60 @@
...
+  unsigned char nonce[TEST_NONCE_SIZE];
+
+  /* Generate key and nonce using /dev/urandom */
+  FILE *fp = fopen("/dev/urandom", "rb");
+  fread(key, sizeof(key), 1, fp);
+  fread(nonce, sizeof(nonce), 1, fp);
+  fclose(fp);
```
ptaffet-jump, line 20: Why add a dependency to urandom when you don't actually care about the randomness? I'd suggest initializing it to something you know (like all 1s) or using `fd_rng` with a fixed seed.

src/ballet/chacha20/test_fd_chacha20_rng.c
```diff
@@ -0,0 +1,60 @@
...
+  fd_chacha20_rng_t num1, num2;
+
+  /* Generate key and nonce using /dev/urandom */
+  FILE *fp = fopen("/dev/urandom", "rb");
+  fread(key, sizeof(key), 1, fp);
+  fread(nonce, sizeof(nonce), 1, fp);
+  fclose(fp);
+
+  /* Initialize the random number generator */
+  assert(fd_chacha20_rng_init(key, nonce) == 0);
```
ptaffet-jump, line 40: We normally prefer `FD_TEST` over `assert` for a unit test

src/ballet/chacha20/test_fd_chacha20_rng.c
```diff
@@ -0,0 +1,60 @@
...
+
+  /* Generate two random numbers and make sure they are different */
+  assert(fd_chacha20_rng_get_uint32(&num1) == 0);
+  assert(fd_chacha20_rng_get_uint32(&num2) == 0);
+  assert(num1 != num2);
```
ptaffet-jump, line 45: We need to make this test much stronger than just checking that the two numbers are different. Richie suggested adding test vectors. I think that's a good idea. You could also add some light randomness tests

src/ballet/chacha20/fd_chacha20_rng.c
```diff
@@ -0,0 +1,33 @@
+#include "fd_chacha20.h"
+
+static EVP_CIPHER_CTX *ctx = NULL;
```
ptaffet-jump, line 3: Firedancer doesn't have much static state like this. Prefer the footprint/align/alloc/new pattern that the bulk of the code follows. If that doesn't work in this situation, this needs a big comment block explaining why.

src/ballet/chacha20/fd_chacha20.h
```diff
@@ -0,0 +1,14 @@
...
+
+#include <openssl/evp.h>
+
+typedef unsigned int fd_chacha20_rng_t;
+
+#define FD_CHACHA20_KEY_SIZE 32   // 32 bytes, 256 bits
+#define FD_CHACHA20_NONCE_SIZE 12 // 12 bytes, 96 bits
+
+int fd_chacha20_rng_init(unsigned char *key, unsigned char *nonce);
+int fd_chacha20_rng_get_uint32(fd_chacha20_rng_t *num);
```
ptaffet-jump, line 12: These functions in the header need some documentation. See other header files in the project to get a sense of what I mean.

src/ballet/chacha20/README.md
```diff
@@ -0,0 +1,15 @@
+# fd_chacha20_rng
+
+## Testing (MacOS)
+
+Build binary by using the following command
+
+```
+gcc -o test_fd_chacha20_rng test_fd_chacha20_rng.c fd_chacha20_rng.c -I/usr/local/opt/openssl/include -L/usr/local/opt/openssl/lib -lcrypto
+```
```
ptaffet-jump, line 9: Adding a README is nice, but the normal way to build a test is with `make`. Does this build with `make`? If so, you should find the unit test in the unit-tests directory.

ptaffet-jump: The unit test should also probably be called `test_chacha20` without the `_rng` suffix.

src/ballet/chacha20/fd_chacha20_rng.c
ptaffet-jump, line 1: I think this should be renamed to `fd_chacha20.c`

src/ballet/chacha20/test_fd_chacha20_rng.c
```diff
@@ -0,0 +1,60 @@
...
+  printf("%u\n", num1);
+  printf("%u\n", num2);
+}
+
+int main()
+{
+  test_fd_chacha20_rng_init();
+  test_fd_chacha20_rng_get_uint32();
+
+  printf("All fd_chacha20_rng tests passed!\n");
```
ptaffet-jump, line 57: This should be FD_LOG_NOTICE

ptaffet-jump (changes requested):
Hi there, thanks for the contribution! There are a few places where your style doesn't match the firedancer code base that definitely need to get fixed before this is merged. Mostly, I'd echo Richie: we would rather copy a few files of code from openssl (so that we can standardize and optimize it) than call out to the library. The other significant issue is the use of the static pointer. That seems to tie our hands in a way we might regret in the future.

src/ballet/chacha20/test_fd_chacha20_rng.c
AlberErre: Looks like `uchar` type is not directly available In OpenSSL

I'm getting the following error while trying to use it -> `error: unknown type name 'uchar'; did you mean 'char'?`

src/ballet/chacha20/test_fd_chacha20_rng.c
AlberErre: Just removed `stdio.h` as no  IO is needed

src/ballet/chacha20/test_fd_chacha20_rng.c
AlberErre: totally agree, just removed it ✅ using a test vector instead

src/ballet/chacha20/test_fd_chacha20_rng.c
AlberErre: just added a test vector ✅

src/ballet/chacha20/fd_chacha20_rng.c
AlberErre: totally agree, deleted this static state ✅, now we generates this EVP_CIPHER_CTX per call and free it from memory once it has finished

src/ballet/chacha20/fd_chacha20.h
AlberErre: Just  updated this by deleting these methods from header (there are now internal functions). Now we're exposing a single method to generate random numbers

AlberErre: Definitely need to add some comments to document this as well

src/ballet/chacha20/README.md
AlberErre: All renaming will be done once code from `openssl` is imported.

Regarding the `make` file, I need to get my feets wet on that 👀

src/ballet/chacha20/fd_chacha20_rng.c
AlberErre: ok! will rename it once code from openssl is imported

AlberErre:
I've refactored the implementation to expose a single method `fd_chacha20_generate_random_number` as the API.

Also, I've updated some stuff based on your comments ✅

Next step will be to copy source from `openssl` directly, as suggested by @ripatel-jump  

I'll keep you posted on discord in case I have any questions 🙏🏼

src/ballet/chacha20/test_fd_chacha20_rng.c
AlberErre: thanks! just solved this by adding `#include "../fd_ballet_base.h"` , which uses `fd_util.h` under the hood 👍🏻

src/ballet/chacha20/test_fd_chacha20_rng.c
AlberErre: updated ✅

src/ballet/chacha20/test_fd_chacha20_rng.c
AlberErre: updated ✅

src/ballet/chacha20/README.md
AlberErre: just added a Makefile and updated the README docs ✅

src/ballet/chacha20/fd_chacha20_rng.c
AlberErre: renamed ✅

AlberErre:
Hey @ripatel-jump @ptaffet-jump I have updated this PR with a different approach 👀

I had a hard time while trying to deterministically reproduce the same random numbers as the rust crate 🥲

The implementation differs from one another, the biggest approximation I got is available on this [tag](https://github.com/AlberErre/firedancer/releases/tag/chacha20-rng.alpha1). However, the random numbers generated, using the same input (key), are not deterministically the same for both implementations ❌

Therefore, I tried a different approach ✅. In order to secure that random numbers are deterministic, as well as sync with other parts of solana, I just integrated this same rust crate [rand_chacha::ChaCha20Rng](https://docs.rs/rand_chacha/latest/rand_chacha/struct.ChaCha20Rng.html), into our c implementation directly. 

As a result, `rust-lib` implements a local rust library that consumes `rand_chacha::ChaCha20Rng` to generate cryptographically secure random numbers using chacha20 cipher.

Then, we use this binary directly in C, which reproduces the same output as the original rust crate.

Details about how to implement Rust code in C are available [here](https://docs.rust-embedded.org/book/interoperability/rust-with-c.html).

ℹ️ This approach has two main benefits:
1) Avoid mirroring implementation logic from rust crate, as if this rust crate changes in future versions, we must update the c implementation as well, which requires extra work.
2) Skip copying and maintaining `openssl` source code directly, which we should update as well, as long a openssl releases new versions.

I have updated the PR accordingly 🎉, also adding a `Makefile` to compile everything, as well as an updating `README` documentation.

ripatel-fd:
@AlberErre Our current approach is to port Solana's cryptographic algorithms to C and include them in this code base (e.g. as done for Ed25519, SHA-256, Blake3, etc). This is motivated by our goals of bringing client diversity to the Solana network and to stay resilient to supply chain attacks in build tooling or dependencies. Importing a binary blob compiled with a different toolchain would be counterproductive to that. Another benefit of porting these algorithms is documentation: A clean reimplementation would allow us to create a specification of the algorithm so it can be ported further https://github.com/solana-foundation/specs

That said, having FFI for the Rust version is useful for ensuring that our C version matches. Thanks for this contribution 🙏

AlberErre:
@ripatel-jump @ptaffet-jump great news! 🎉 

I finally was able to reproduce the same random number as the rust crate, by using a C implementation ✅ (needed to play with nonce and counter params to achieve it).

I've updated the test cases & `Makefile` accordingly 👍🏻 - Now, test cases directly compare the output from both C implementation and FFI rust crate functions, to secure both implementations produces the same random number deterministically 🎉

Next, I will copy the OpenSSL source code as you pointed out @ripatel-jump. Totally understand your point about porting Solana's cryptographic algorithms to C.

ripatel-fd:
> I finally was able to reproduce the same random number as the rust crate, by using a C implementation

Thanks @AlberErre, that is very cool and helps us a lot. I will reach out in Discord to assist with the OpenSSL import.

ripatel-fd:
@AlberErre I've removed the dependency on OpenSSL and ported this PR in https://github.com/firedancer-io/firedancer/pull/375 and https://github.com/firedancer-io/firedancer/pull/380. Thanks a lot for the contribution 🙏

AlberErre:
Great! 🎉 I'm happy to be able to help with this 🙏🏼 just learned a lot about ciphers

PR #347, opened 2023-05-14
deps.sh: add bzip2, RocksDB

ripatel-jump:
Also speed up checkout_repo function

### DIFF
--- a/deps.sh
+++ b/deps.sh
@@ -4,6 +4,7 @@ set -euo pipefail
 
 # Change into Firedancer root directory
 cd "$(dirname "${BASH_SOURCE[0]}")"
+REPO_ROOT="$(pwd)"
 
 # Fix pkg-config path and environment
 # shellcheck source=./activate-opt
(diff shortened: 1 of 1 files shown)


jumpsiegel (approved):
For rocksdb, i found it sometimes using the compiler in /bin/gcc instead of the one in my path.  The workout for this was to do a 'export CC=`which gcc`' before my compile.  

to be more complete, since I wanted everything to install locally and not require root

```
export prefix=$HOME/local-9.3.0
export PREFIX=$HOME/local-9.3.0
export CC=`which gcc`
export cc=`which gcc`


```Anyhow, food for thought...

ripatel-fd:
Marking this as a draft discussed internally. RocksDB should not be a default dependency.

ripatel-fd:
> to be more complete, since I wanted everything to install locally and not require root

@jumpsiegel `deps.sh` already sets the prefix for you. It is in the `opt` dir of the repo.

PR #352, opened 2023-05-15
ci: remove parallelism in unit tests

lheeger-jump:


ripatel-fd:
Closing since it's fixed on main

PR #353, opened 2023-05-15
Added a HPC heap code generator

kbowers-jump:
By popular demand ... specifically to optimize algorithmic space and time overheads to produce the input ordering of various dirty account hashes (on top this, this is an HPC optimized implementation).  This is more generically useful than this particular use case though.

fd_heap builds a heap out of inputs under an arbitrary element comparison function.  But unlike standard heap algorithms (including the prq code generator), it does not impose any requirements on how the element storage is organized.  Instead, it works zero-copy / in-place with whatever application element storage management already exists.  So this can be a useful building block for doing "sort-ish" things but with more flexibility and less overhead than standard heaps, sorts, trees, etc.

It is useful when dealing with streaming inputs that arrive in no particular order (and with no particular expectations about what should arrive) but that need to be processed in a particular order and then ideally without any auxiliary temporary allocations / memory copies / separate sorting passes / out-of-place sorting algos / etc ...

In such cases, we can do any order-independent processing on an input immediately when it arrives and then insert it into the heap however it got placed in memory on arrival.  When ready to do to the order sensitive processing, we can then spool back the inputs from the heap in the appropriate order via the remove_min API.  (Basically a heap sort without memory layout restrictions and that has also been overlapped with user computations and distributed communications overheads.)

Note that, given the zero-copy / in-place nature, it is possible to have multiple (potentially non-disjoint) heaps for the same input stream with minimal additional costs.


src/util/tmpl/fd_heap.c
```diff
@@ -0,0 +1,647 @@
...
+     // Return will be in [0,ele_max) or myheap_idx_null().  Lifetime is
+     // the element storage lifetime.  myheap_idx_fast is the same but
+     // assumes e is not NULL.  myheap_ele returns a pointer in the
+     // caller's address space to element idx.  Assumes idx is in
+     // [0,ele_max) or is myheap_idx_null().  Return pointer lifetime is
+     // ele's local lifetime.  myheap_ele_fast is the same but assumes
+     // idx is not myheap_idx_null().  myheap_ele[_fast]_const is a
+     // const correct version.
+
+     ulong           myheap_idx           ( myele_t const * e, myele_t const * ele );
```
asiegel-jt, line 73: What is the ele argument here? Some sentinel element? It's not clear what your indexes are relative to.

src/util/tmpl/fd_heap.c
```diff
@@ -0,0 +1,647 @@
...
+FD_FN_CONST static inline HEAP_T *       HEAP_(ele_fast)      ( ulong i,          HEAP_T *       ele ) { return ele + i; }
+FD_FN_CONST static inline HEAP_T const * HEAP_(ele_fast_const)( ulong i,          HEAP_T const * ele ) { return ele + i; }
+
+FD_FN_PURE static inline ulong HEAP_(ele_max)( HEAP_(t) const * heap ) { return heap->ele_max; }
+FD_FN_PURE static inline ulong HEAP_(ele_cnt)( HEAP_(t) const * heap ) { return heap->ele_cnt; }
+
+FD_FN_PURE static inline ulong HEAP_(idx_peek_min)( HEAP_(t) const * heap ) { return (ulong)heap->root; }
+
+FD_FN_PURE static inline HEAP_T *
+HEAP_(ele_peek_min)( HEAP_(t) const * heap,
```
asiegel-jt, line 341: In this API and others, the meaning of the ele argument is somewhat mysterious. Is this a well known sentinel?

src/util/tmpl/fd_heap.c
```diff
@@ -0,0 +1,647 @@
...
+
+FD_FN_PURE static inline HEAP_T const *
+HEAP_(ele_peek_min_const)( HEAP_(t) const * heap,
+                           HEAP_T const *   ele ) {
+  ulong i = (ulong)heap->root;
+  return fd_ptr_if( !HEAP_IDX_IS_NULL( i ), ele + i, NULL );
+}
+
+static inline HEAP_(t) *
+HEAP_(ele_insert)( HEAP_(t) * heap,
```
asiegel-jt, line 355: e and ele? This is very confusing.

src/util/tmpl/fd_treap.c
```diff
@@ -205,12 +205,13 @@
      myele_t const * mytreap_ele_query_const( mytreap_t const * treap, char const * q, myele_t const * ele );
```
asiegel-jt, line 205: As with the heap code, the meaning of the ele argument is unclear. Are you assuming all possible elements come from a linear pool?

src/util/tmpl/fd_treap.c
```diff
@@ -492,18 +493,20 @@ TREAP_(ele_query_const)( TREAP_(t) const * treap,
...
+  return treap;
 }
 
-static inline void
+static inline TREAP_(t) *
 TREAP_(ele_remove)( TREAP_(t) * treap,
                     TREAP_T *   e,
                     TREAP_T *   ele ) {
   TREAP_(idx_remove)( treap, (ulong)(e-ele), ele );
+  return treap;
```
asiegel-jt, line 509: These methods in general and the seed function in particular seem to assume the elements come from a flat array. Is this mean to be used with fd_pool? If so, perhaps this should be explicit.

src/util/tmpl/fd_treap.c
```diff
@@ -625,7 +628,7 @@ TREAP_(idx_query)( TREAP_(t) const * treap,
   return i;
```
asiegel-jt, line 628: TREAP_QUERY_T might be different from TREAP_T. Is the different cmp function meant to deal with this?

src/util/tmpl/fd_heap.c
```diff
@@ -0,0 +1,647 @@
...
+     // removed element can be freely modified on return.
+     //
+     // myheap_ele_remove_min is the same and just for naming
+     // consistency.
+     //
+     // These operations have HPC implementations and are O(lg N)
+     // average with an ultra high probability of having a small
+     // coefficient (i.e. close to algorithmically optimal trees).
+
+     myheap_t * myheap_idx_remove_min( myheap_t * heap, myele_t * ele );
```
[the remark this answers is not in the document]
kbowers-jump: See global comment below.

src/util/tmpl/fd_heap.c
kbowers-jump: See global comment below.

src/util/tmpl/fd_heap.c
kbowers-jump: See global comment below.

src/util/tmpl/fd_treap.c
kbowers-jump: See global comment below.

src/util/tmpl/fd_treap.c
kbowers-jump: See global comment below.

src/util/tmpl/fd_treap.c
kbowers-jump: It is possible (see the types of queries done in the treaps used for workspace) to do queries over a range instead of an exact match.  So, an element of treap doesn't specify one possible query.  It can specify a disjoint range of queries satisfied by that element.  This in turn implies the need for some flexibility on the API presented to the user for queries versus how queries are done for internal operations.

kbowers-jump:
For the above mentions and some follow-up discussions with Alex, there was some confusion over the use of the ele arguments.  So the additional comments and the argument naming tweaks in fd_heap / fd_treap.c hopefully will clarify.

High level:

The only assumption the storage region backing the heap / treap is that it is linearly addressable.  That includes fd_pool but does not require it (hence it wasn't called pool to avoid the implication the APIs could only be used with fd_pool but looks like that was created more confusion than not).  This also allows multiple heaps/treaps/maps/etc to be composed simultaneously on top of the same underlying storage region.

The most optimal use of these structures is to work directly in the storage indexing space.  Such is faster, more secure and more flexible.  Namely, it allows these structures to be persistent, relocatable, more robust against type confusion memory security vulnerabilities while still using fast addressing modes on the processor (it is also is easier for the compiler to optimize but these are not used in a context where I expect that to matter much).

The address conversion and ele based on top of that are then to allow users who don't want to work directly in storage indexing space for whatever reason to work in their local address space (like a conventional local address space based implementation) but without losing the persistence, relocatability, inter-process usability, etc properties.

asiegel-jt (approved):

PR #358, opened 2023-05-21
bmtree: refactor, add proofs

ptaffet-jump:
Refactor the original `fd_bmtree` so it's not a template.  Add Merkle proof generation and validation.  Allow specifying different hash domains.

TODO: Do we want to have two Merkel tree implementations long term? Should we merge this and wbmtree? Also, this will probably get a faster caching proof validator soon.


jumpsiegel (approved):
We have a version of this that attempts to use the vectorized sha256 stuff to improve performance.  This inclusion proof stuff has not been implemented in that.  Maybe we do need to merge all these...

PR #396, opened 2023-05-25
license: standardize formatting to enable automatic detection

mmcgee-jump:


ripatel-fd:
This this is touching legal stuff, I think @kbowers-jump should review

mmcgee-jump:
Sure, @kbowers-jump it's the same copy-pasted text, but I think the header preamble was throwing off licensee so the project didn't get tagged with a LICENSE. I surveyed a bunch of established projects and could not find similar preambles.

ripatel-fd:
Looks like Solana Labs ran into a similar issue: https://github.com/solana-labs/solana/commit/c65605802a2fa58d3f48a655b90dfb0b14c2b62e

Merging this

PR #397, opened 2023-05-25
ledger: add leader schedule

ripatel-fd:
- [ ] Add leader schedule data structure
- [ ] Add leader schedule derive algorithm
- [ ] Test against mainnet data
- [ ] Benchmark for performance
- [ ] Consider improving binary search since the input is highly unbalanced


ripatel-fd:
Closing, there will a separate attempt to backport this

PR #400, opened 2023-05-27
quic: remove duplicate eth/ip4/udp header defs

ripatel-fd:
fd_quic defines duplicate definitions of the Ethernet, IPv4, and UDP header types.

Replaces them with util variants and also fixes extended IHL for IPv4 decoding.


lidatong (approved):

lidatong:
will let @nbridge-jump weigh-in

nbridge-jump (approved):

PR #404, opened 2023-05-29
Codeql

marcus-jump:


ripatel-fd:
I'm not super excited by the idea of using a general-purpose linter as these tend to have a lot of false positives, but let's give it a try :)

marcus-ares:
Thanks Richie!
I'll just make sure it works before rebasing.

PR #120, opened 2023-02-10
shmem: warn if rlimit too low

ripatel-fd:
Logs warning messages during shmem boot when the memlock, nice, rtprio rlimits are too low.

PR #405, opened 2023-05-30
fdctl: add checks for resource limits

mmcgee-jump:
Certain system resource limit changes are encouraged so that Firedancer can run correctly. Add checks for these `fdctl` to help identify if the configuration is not right.


ripatel-fd:
Relates to #120

PR #407, opened 2023-05-30
chore: add .clang-format

lidatong:


ripatel-fd:
There is also an `uncrustify` config

lidatong:
closing given `uncrustify`

PR #413, opened 2023-06-01
ffi: fix broken publish

mmcgee-jump:
There were a few problems here,

 (a) Running make dirtied the src tree which caused `cargo package`
     to complain. We give the Makefile a configurable artifact dir
     and set it OUT_DIR.
 (b) An `include` section had been added to the `Cargo.toml` but did
     not include the right things, so these are added.
 (c) `cargo package` will not traverse into directories that themselves
     have a `Cargo.toml` when it's locating files to include in the
     package. This is a problem because we newly have a rust workspace
     in the root of our repo. To workaround this, I set up a new
     "firedancer" symlink directory that unwraps the first level,
     excluding things we don't need. This also has the benefit of
     exluding this junk from the package we publish online.


ripatel-fd:
Thanks for fixing this

PR #420, opened 2023-06-05
add fd_hex_decode

ripatel-fd:
Adds very basic hex decode function. 

Used in QUIC for config and debugging.

### DIFF
--- a/src/ballet/hex/fd_hex.c
+++ b/src/ballet/hex/fd_hex.c
@@ -1,0 +1,31 @@
+#include "fd_hex.h"
+
+/* FIXME use LUT instead? */
+static inline int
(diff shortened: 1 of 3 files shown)


src/ballet/hex/fd_hex.c
```diff
@@ -0,0 +1,31 @@
+#include "fd_hex.h"
+
+/* FIXME use LUT instead? */
```
mmcgee-jump, line 3: My preference is not to leave these FIXME and TODO that will never get fixed. Can just remove them, will be obvious and get addressed if it's a bottleneck.

mmcgee-jump (approved):

PR #425, opened 2023-06-06
sandbox: belt sanding sketch

mmcgee-jump:
I wanted to highlight what a minimally invasive change looks like here. It's largely just a belt sanding / style change, but has a few notable changes,

 * Explicitly clear environment to prevent read from current process.
 * Pipe through "--no-sandbox" argument for development.
 * Update required capabilities for run binary.
 * The `sandbox_mountns` is switched to do what Firecracker does. I couldn't get the prior code to run.
 * The `drop_capabilities` ordering is also switched around, I couldn't get that to run before either.

### DIFF
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -320,6 +320,12 @@ scratch_directory = "/home/{user}/.firedancer/{name}"
         signature_cache_size = 4194302
 
 [development]
+    # For enhanced security, Firedancer runs itself in a restrictive sandbox in production. The
+    # sandbox prevents most system calls and restricts the capabilities of the process after
+    # initialization to make the attack surface smaller. This is required in production, but might
+    # be too restrictive during development.
+    sandbox = true
+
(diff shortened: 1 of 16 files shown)


ripatel-fd:
Could you tag @marcus-jump as the main author (and yourself as coauthor)? Then I think we can merge this as is 👍

claudijd:
Can this PR be this be pointed at the `marcus/jailer` branch instead of `main` as suggested improvements?

This would help preserve a lot of the prior work and commit attribution that were added based on review feedback in https://github.com/firedancer-io/firedancer/pull/252

mmcgee-jump:
@claudijd I will rebase it, but this is not intended to be merged, just a sketch for discussion. If there are useful changes they can be reincorporated into Marcs PR.

claudijd:
> @claudijd I will rebase it, but this is not intended to be merged, just a sketch for discussion. If there are useful changes they can be reincorporated into Marcs PR.

Super helpful context, thank you!

Issue #317, opened 2023-05-03
quic: add qlog event definitions


ripatel-fd:
https://github.com/quicwg/qlog

qlog is a standard logging format for QUIC implementers. 
We might want to support qlog in Firedancer. 

Step one of doing so is adding headers with struct definitions for qlog generic and QUIC events.

https://github.com/quicwg/qlog/blob/main/draft-ietf-quic-qlog-main-schema.md
https://github.com/quicwg/qlog/blob/main/draft-ietf-quic-qlog-quic-events.md

PR #430, opened 2023-06-07
add TLS keylog parser

ripatel-fd:
From milestone 1.1


src/ballet/tls/fuzz_tls_keylog_parse.c
```diff
@@ -0,0 +1,28 @@
...
+#error "This target requires FD_HAS_HOSTED"
+#endif
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "fd_tls_keylog.h"
+
+int
+LLVMFuzzerInitialize( int  *   argc,
```
mmcgee-jump, line 11: If this is just for test / development, why did we write a fuzzer? Is that comment still accurate?

mmcgee-jump (changes requested):
Can you clarify why this is here or what it's used for? There's not even any way to call the code?

mmcgee-jump:
Discussed with Richie, this is used for decrypting `pcapng` files so we can feed them back into Firedancer as a simulation. It's not for production use in live Firedancer itself.

Suggest sticking it in a different lib than `ballet` (eg a new`testutil` library) before merging.

ripatel-fd:
We won't need this if we implement qlog (#317), which also provides access to decrypted frames

PR #432, opened 2023-06-07
aio: add fd_aio_pcapng

ripatel-fd:
Adds aio middleware for logging to pcapng. From milestone 1.1

### DIFF
--- a/src/tango/aio/fd_aio_pcapng.c
+++ b/src/tango/aio/fd_aio_pcapng.c
@@ -1,0 +1,73 @@
+#include "fd_aio_pcapng.h"
+#include "../../util/net/fd_pcapng_private.h"
+
+#include <errno.h>
+
+/* fd_aio_pcapng_send implements fd_aio_send_t */
+
+static int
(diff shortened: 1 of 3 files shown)


src/tango/aio/fd_aio_pcapng.c
```diff
@@ -0,0 +1,73 @@
...
+fd_aio_pcapng_t *
+fd_aio_pcapng_join( void *           _mitm,
+                    fd_aio_t const * dst,
+                    void *           pcapng ) {
+
+  fd_aio_pcapng_t * mitm = (fd_aio_pcapng_t *)_mitm;
+  mitm->dst    = dst;
+  mitm->pcapng = pcapng;
+
+  FD_TEST( fd_aio_join( fd_aio_new( &mitm->local, mitm, fd_aio_pcapng_send ) ) );
```
mmcgee-jump, line 60: Probably prefer a library to return error rather than bring down the process. Minor nit though, up to you though.

PR #433, opened 2023-06-07
config: add with-openssl.mk

ripatel-jump:


ripatel-fd:
@mmcgee-jump Feel free to edit to your liking. This version is pulled from milestone-1.1-demo and has been working flawlessly.

mmcgee-jump:
Looks good for now, but I think we should get rid of the flags (`FD_HAS_OPENSSL`) and also `pkg-config` in a later change.

PR #437, opened 2023-06-07
.gitignore: add certs, pcaps, logs

ripatel-fd:
Sync from milestone 1.1 demo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,16 @@ opt/
 # Vim
 *.swp
 *.swo
+
+# Certificates
+*.csr
+*.crt
+*.key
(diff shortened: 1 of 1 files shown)


mmcgee-jump (approved):

mmcgee-jump:
Approved because it's trivial. But do these files actually routinely appear in the tree? It is kinda bad form to add random stuff to this.

ripatel-fd:
> Approved because it's trivial. But do these files actually routinely appear in the tree?

Yes, all file types added here routinely appear. Certs and pcaps for QUIC tests and log files for `make run-unit-test`

PR #442, opened 2023-06-08
config: enable OpenSSL by default

ripatel-fd:
In preparation for QUIC

### DIFF
--- a/config/with-openssl.mk
+++ b/config/with-openssl.mk
@@ -1,5 +1,8 @@
 LDFLAGS+=-Wl,--push-state,-Bstatic $(shell pkg-config --libs libssl libcrypto) -Wl,--pop-state
 
+# OpenSSL requires dynsym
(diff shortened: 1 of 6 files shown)


ripatel-fd:
@mmcgee-jump btw would be cool to deduplicate the content in `linux_{clang,gcc}_x86_64` and `linux_{clang,gcc}_icelake`

tpointon-jump (approved):

config/with-openssl.mk
```diff
@@ -1,5 +1,8 @@
 LDFLAGS+=-Wl,--push-state,-Bstatic $(shell pkg-config --libs libssl libcrypto) -Wl,--pop-state
 
+# OpenSSL requires dynsym
```
mmcgee-jump, line 3: Hmm. Why. Can we get it to link completely statically?

PR #445, opened 2023-06-08
util: add fd_cstr_to_mac_addr

ripatel-fd:
Required for QUIC merge. Not the cleanest implementation, but would be nice to get merged regardless to unblock QUIC merge. (It supports constant-time evaluation though, which might make this algorithm useful as a macro)

### DIFF
--- a/src/util/net/fd_eth.h
+++ b/src/util/net/fd_eth.h
@@ -39,7 +39,7 @@
    VLAN tags have an unfortunately far wider range of behaviors in the
    wild due to the rather messy set of protocols that have accumulated
    over the decades.
-   
+
    TAG_CNT can only be determined by parsing the packet headers.  0 (raw
    Ethernet) or 1 (VLAN tagged ethernet) are common but there isn't an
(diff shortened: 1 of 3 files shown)


jumpsiegel (approved):

src/util/net/fd_eth.h
```diff
@@ -283,6 +283,16 @@ fd_vlan_tag( void * _tag,
...
 }
 
+/* fd_cstr_to_mac_addr parses a MAC address matching format
+   FD_ETH_MAC_FMT from the given cstr and stores the result into mac.
+   On success returns mac.  On failure, returns NULL and leaves mac
+   in an undefined state.  Reasons for failure include that strlen
+   of s is not exactly 17 or that s is not formatted correctly. */
+
+uchar *
+fd_cstr_to_mac_addr( char const * s,
```
mmcgee-jump, line 293: It's more test-only code going into a runtime lib. Kinda wish we had a better place to put things like this.

PR #454, opened 2023-06-09
build: use native target

mmcgee-jump:
Here's a kind of junk sketch of this. Not meant for super serious review but just for discussion. It probably goes too far in removing the existing MACHINE support, which we might need for cross compiling (would we ever do this?), but doesn't go far enough in moving a lot of this stuff into the compiler rather that defining it at build time.


mmcgee-jump:
replaced by https://github.com/firedancer-io/firedancer/pull/508

PR #458, opened 2023-06-10
Only support AES-128-GCM, remove SHA-384

ripatel-fd:
There is no need to support any ciphersuites other than AES-128-GCM which is the only suite required by RFC 9001 (QUIC-TLS). AES-256-GCM, ChaCha20-Poly1305 are recommended but not required. This hardens our codebase by reducing the amount of code to audit. SHA-384 is not used anywhere else in our codebase, so it can go.

- sha512: revert sha384 support
- quic: only support TLS_AES_128_GCM_256 for now


lidatong (approved):

ripatel-fd:
Closing stale PR

PR #461, opened 2023-06-10
contrib: add deps-bundle.sh

ripatel-fd:


ripatel-fd:
@marcus-jump @mmcgee-jump This is an experiment to make our CI more portable and allow managing the dependency bundle without direct access to the CI runners. Would appreciate a review of the approach.

marcus-jump (approved):
I think that it's worth noting the obvious: this will only works for opt and not for system packages.

I also think that the building and committing of this bundle should be done by a GHA.

ripatel-fd:
@marcus-jump Thanks for reviewing. I agree obviously re GHA -- We could add a manual trigger to refresh the deps bundle.

PR #469, opened 2023-06-13
test with asan

marcus-jump:


marcus-ares:
@ripatel-jump -- it seems to work as is. I confirm that tests are running with asan.

ripatel-fd:
> @ripatel-jump -- it seems to work as is. I confirm that tests are running with asan.

@marcus-jump Did you modify the `linux_clang_fuzz_asan` profile then? This profile skips any targets with a `main` (currently `make-bin` and `make-unit-test`).

PR #475, opened 2023-06-16
Ptaffet/pcap no headers

ptaffet-jump:


ptaffet-jump:
I merged this directly into the 1.4 branch. If we want it in main, we can deal with it later.

PR #479, opened 2023-06-20
sandbox: sandbox implementation

mmcgee-jump:
The purpose of the sandbox is to reduce the impact of a Firedancer compromise.

When starting, before executing any task code and/or processing user-provided input, a Firedancer process prepares everything it needs in order to function properly. This initialization mostly consists in:

 * Creating new tiles which will be used for task execution (clone syscall)
 * Opening the relevant workspaces which will be used to perform work and communicate (mmap syscall)
 * Immediately after performing those operations, Firedancer sandboxes itself. Note that firedancer has to be started as root or with various capabilities in order to be able to sandbox itself.

Here are the mechanisms currently used by Firedancer to achieve sandboxing:

 * The environment variable are cleared. Environment variables are commonly used to hold secrets. If Firedancer is compromised, no secrets living in the operator's environment will be leaked.
 * The process loses access to network interfaces. The process unshares the network namespace to keep the principle of least privilege: in the event where the process was able to interact with a network interface, it should not be able to perform any communication.
 * The process gets a restricted view of the filesystem. It is jailed into a mount namespace with a root of its own. The process should only be able to interact with files that it needs to function.
 * The process is restricted from opening any new files with restrictive resource limits. In the future, more resources types can be limited. Firedancer processes have well understood expected behaviors and resource needs. A process should not be able to exceed those limits, potentially leading to availability issues.
 * The process enters a new user namespace, and the user it runs as inside the namespace is mapped to the overflow user outside of the namespace. In the case where another control was to fail, the process should be interacting with the system as an unprivileged user.
 * All file descriptors above a specified number are forcefully closed. Similar to and more impactful than clearenv, an operator's process can have FDs opened that are 1. not relevant to Firedancer 2. references to sensitive resources. Those resources should not be made available to Firedancer.
 * We prevent the usage of most syscalls, only allowing those explicitly needed by the specific Firedancer component. Syscalls are used to interact with the operating system. There exists close to 400 syscalls. While running, A Firedancer process requires 14 syscalls out of those 400 in order to perform its functions. Firedancer will crash if it attempts to use a syscall that is not expected. Note: It also happens that the syscalls that Firedancer is using are ubiquitous and well understood. They have stood the test of time (not that time is an ultimate metric for greatness). We have the luxury of disallowing all of the syscalls that might have received less scrutiny.

### DIFF
--- a/src/util/env/fd_env.c
+++ b/src/util/env/fd_env.c
@@ -55,6 +55,20 @@ FD_ENV_STRIP_CMDLINE_IMPL( double,       double )
 
 #undef FD_ENV_STRIP_CMDLINE_IMPL
 
+int
+fd_env_strip_cmdline_contains( int * pargc, char *** pargv, char const * key ) {
+  int new_argc = 0;
+  int found = 0;
+  if( key && pargc && pargv ) {
+    for( int arg=0; arg<(*pargc); arg++ )
+      if( strcmp( key, (*pargv)[arg] ) ) (*pargv)[new_argc++] = (*pargv)[arg];
+      else found = 1;
+    (*pargc)           = new_argc;
+    (*pargv)[new_argc] = NULL; /* ANSI - argv is NULL terminated */
+  }
+  return found;
+}
+
 #else
 #error "Unknown FD_ENV_STYLE"
 #endif
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -320,6 +320,12 @@ scratch_directory = "/home/{user}/.firedancer/{name}"
         signature_cache_size = 4194302
 
 [development]
+    # For enhanced security, Firedancer runs itself in a restrictive sandbox in production. The
+    # sandbox prevents most system calls and restricts the capabilities of the process after
+    # initialization to make the attack surface smaller. This is required in production, but might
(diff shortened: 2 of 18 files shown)


src/util/env/fd_env.c
```diff
@@ -55,6 +55,20 @@ FD_ENV_STRIP_CMDLINE_IMPL( double,       double )
...
 #undef FD_ENV_STRIP_CMDLINE_IMPL
 
+int
+fd_env_strip_cmdline_contains( int * pargc, char *** pargv, char const * key ) {
+  int new_argc = 0;
+  int found = 0;
+  if( key && pargc && pargv ) {
+    for( int arg=0; arg<(*pargc); arg++ )
+      if( strcmp( key, (*pargv)[arg] ) ) (*pargv)[new_argc++] = (*pargv)[arg];
+      else if( (++arg)<(*pargc) ) found = 1;
```
marcus-jump, line 65: If the argument you're looking at is the last one, found is going to be 0.

marcus-jump (approved):

src/util/env/fd_env.c
mmcgee-jump: nice find!

marcus-jump (approved):

PR #519, opened 2023-07-10
fdctl: configure large pages without alloc program

mmcgee-jump:


ptaffet-jump (approved):
Re-stamping Richie's approval

PR #520, opened 2023-07-11
fix: x509 cert gen to make rustls happy

lidatong:


ripatel-fd:
Approving since we have to work around rustls bugs

PR #522, opened 2023-07-13
fixup ld library path

ripatel-fd:


ripatel-fd:
For now, we'll fix this by hardcoding the RocksDB rpath.

PR #527, opened 2023-07-17
Temporarily remove runtime deps

ripatel-fd:
RocksDB is not used in the main branch.

### DIFF
--- a/deps.sh
+++ b/deps.sh
@@ -105,12 +105,12 @@ checkout_repo () {
 fetch () {
   mkdir -pv ./opt/git
 
-  checkout_repo zlib      https://github.com/madler/zlib            "v1.2.13"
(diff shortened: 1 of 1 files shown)


mmcgee-jump:
Is any of this stuff except openssl used on main?

ripatel-fd:
> Is any of this stuff except openssl used on main?

No. bzip2 and zstd are required for p2p networking though. secp256k1 will be required for the runtime (not sure if Liam already integrated it into the VM). We should still get rid of them until we're starting to merge the runtime, IMO.

mmcgee-jump:
Yeah I mean at least lets comment those out on main as well?

ripatel-fd:
@mmcgee-jump Good call, done. This makes the milestone 1.5 build scripts more robust.

PR #532, opened 2023-07-19
ballet/pack: Switch to simple version of pack

ptaffet-jump:
For MS 1.5, we're going to use a simpler version of pack.  Unlike the previous version, this schedules microblocks instead of transactions. This version does not include any fancy CU estimation or the opportunity cost-based packing algorithm.  Overall, it's pretty similar to what Solana Labs uses, but it includes a parameter `gap` that creates inter-microblock parallelism.
Right now, schedule is probably too slow because it's O(pending txns). This version also does not support address lookup tables yet.


src/app/frank/fd_frank_pack.c
```diff
@@ -64,30 +64,11 @@ fd_frank_pack_task( int     argc,
   ulong accum_ovrnp_cnt = 0UL;
```
mmcgee-jump, line 64: Needs `frank.c` changes.

src/app/frank/fd_frank_pack.c
```diff
@@ -96,16 +77,15 @@ fd_frank_pack_task( int     argc,
   FD_LOG_INFO(( "joining firedancer.pack.out-dcache" ));
   uchar * out_dcache = fd_dcache_join( fd_wksp_pod_map( cfg_pod, "pack.out-dcache" ) );
   if( FD_UNLIKELY( !out_dcache ) ) FD_LOG_ERR(( "fd_dcache_join failed" ));
-  void * dcache_base = out_dcache;
+  // void * dcache_base = out_dcache;
```
mmcgee-jump, line 80: Delete

src/app/frank/fd_frank_pack.c
```diff
@@ -193,31 +169,19 @@ fd_frank_pack_task( int     argc,
...
-      schedule_ready = now;
     }
 
     /* Is it time to schedule the next transaction? */
-    /* FIXME: add flow control */
-    if( FD_LIKELY( !!fd_pack_avail_txn_cnt( pack ) ) & FD_LIKELY( (now-schedule_ready)>=0L ) ) {
-      fd_pack_scheduled_txn_t t = fd_pack_schedule_next( pack );
-      if( FD_LIKELY( t.txn ) ) {
+    /* for each banking thread, if it has credits */
+    if( 0 ) {
```
mmcgee-jump, line 178: Does this need to be implemented?

mmcgee-jump: (also for the todo on the line below)

PR #541, opened 2023-07-21
add zip to fuzz_artifacts runner

marcus-jump:


marcus-ares:
Tested on https://github.com/firedancer-io/firedancer/actions/runs/5625911245/job/15245653052

nbridge-jump (approved):

PR #543, opened 2023-07-24
Initial solcap API

ripatel-fd:


ripatel-fd:
Need to recreate this PR.

PR #548, opened 2023-07-28
fdctl: workspace isolation

mmcgee-jump:
For security reasons we want to isolate Firedancer tiles so that they don't share memory except where strictly needed to communicate. To do this we split the current one large workspace into many smaller ones. The specific mechanics are,

 * Each tile gets its own workspace, with a config pod and some local objects that might be needed (a `tcache`, or local static config like `idle_timeout_ms` for quic tiles).

  * Each "kind" of tile <-> tile channel gets one workspace. For example, quic <-> verify is a channel, so there is one workspace shared by all quic and verify tiles which contains `mcache`es, `dcaches`, and other concurrent objects for those channels.

  * We need to query this workspace layout in various places so to make that easier it's defined statically as part of application config in `config.c`. Queries like "what workspaces does this tile need to map before calling seccomp?" are answered using the structures.

  * Because each workspace has a pod, it would be a lot to manage with the sidecar `config.cfg` file, so it's removed. Instead, to locate pods in each workspace we make an assumption: the pod is always the first thing allocated, so it will live at the `gaddr_lo` (defined in the workspace internals).


src/app/frank/fd_frank_dedup.c
```diff
@@ -45,43 +27,41 @@ run( fd_frank_args_t * args ) {
...
-    fd_pod_info_t info = fd_pod_iter_info( iter );
-    if( FD_UNLIKELY( info.val_type!=FD_POD_VAL_TYPE_SUBPOD ) ) continue;
-    char const  * verify_name =                info.key;
-    uchar const * verify_pod  = (uchar const *)info.val;
-
-    FD_LOG_INFO(( "joining firedancer.verify.%s.mcache", verify_name ));
-    in_mcache[ in_idx ] = fd_mcache_join( fd_wksp_pod_map( verify_pod, "mcache" ) );
+  for( ulong i=0; i<in_cnt; i++ ) {
+    char path[ 32 ];
+    snprintf( path, 32, "mcache%lu", i );
```
llamb-jump, line 32: stylistically what do you think of `snprintf( path, sizeof(path), "mcache%lu", i );`?

llamb-jump (approved):

PR #554, opened 2023-08-01
fix fd_frank_quic (bad syscall and FDs being closed)

llamb-jump:
fix solana labs being created in the wrong namespace

### DIFF
--- a/src/app/fdctl/utility.c
+++ b/src/app/fdctl/utility.c
@@ -14,19 +14,44 @@
 #include <linux/netlink.h>
 #include <linux/rtnetlink.h>
 
-static int entered = 0;
+static int namespace_original_fd = 0;
 
 void
-enter_network_namespace( config_t * const config ) {
-  if( FD_LIKELY( !config->development.netns.enabled || entered ) ) return;
-  entered = 1;
-
+enter_network_namespace( const char * interface ) {
   char path[ PATH_MAX ];
-  snprintf1( path, PATH_MAX, "/var/run/netns/%s", config->tiles.quic.interface );
+  snprintf1( path, PATH_MAX, "/var/run/netns/%s", interface );
+
+  if( FD_LIKELY( !namespace_original_fd ) ) {
+    namespace_original_fd = open( "/proc/self/ns/net", O_RDONLY | O_CLOEXEC );
+    FD_LOG_NOTICE(("enter_network_namespace original %d", namespace_original_fd));
+    if( FD_UNLIKELY( namespace_original_fd < 0 ) ) FD_LOG_ERR(( "failed to open /proc/self/ns/net (%i-%s)", errno, strerror( errno ) ));
+  }
 
   int fd = open( path, O_RDONLY | O_CLOEXEC );
   if( FD_UNLIKELY( fd < 0 ) ) FD_LOG_ERR(( "failed to open `%s` (%i-%s)", path, errno, strerror( errno ) ));
   if( FD_UNLIKELY( setns( fd, CLONE_NEWNET ) ) ) FD_LOG_ERR(( "failed to enter network namespace `%s` (%i-%s)", path, errno, strerror( errno ) ));
+  int ret = close( fd );
+  if( FD_UNLIKELY( ret ) ) FD_LOG_ERR(( "enter_network_namespace %d (%i-%s)", ret, errno, strerror( errno ) ));
+}
+
+void
+close_network_namespace_original_fd( void ) {
+  int ret = close( namespace_original_fd );
+  if( FD_UNLIKELY( ret ) ) FD_LOG_ERR(( "leave_network_namespace %d (%i-%s)", ret, errno, strerror( errno ) ));
+}
+
+void
+leave_network_namespace( void ) {
+  if( FD_UNLIKELY( !namespace_original_fd ) ) {
+    FD_LOG_NOTICE(( "leave_network_namespace break early"));
+    return;
+  }
+
+  if( FD_UNLIKELY( setns( namespace_original_fd, CLONE_NEWNET ) ) ) FD_LOG_ERR(( "failed to enter original network namespace `%d` (%i-%s)", namespace_original_fd, errno, strerror( errno ) ));
+
+  int ret = close( namespace_original_fd );
+  if( FD_UNLIKELY( ret ) ) FD_LOG_ERR(( "leave_network_namespace %d (%i-%s)", ret, errno, strerror( errno ) ));
+  namespace_original_fd = 0;
 }
 
 void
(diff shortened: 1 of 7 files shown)


src/app/frank/fd_frank_quic.c
```diff
@@ -133,13 +133,14 @@ static long allow_syscalls[] = {
...
   __NR_getrandom, /* OpenSSL RAND_bytes reads getrandom, temporarily used as part of quic_init to generate a certificate */
+  __NR_sendto,    /* fd_xsk requires sendto */
 };
 
 fd_frank_task_t frank_quic = {
   .name     = "quic",
   .in_wksp  = NULL,
   .out_wksp = "quic_verify",
-  .close_fd_start = 4, /* stdin, stdout, stderr, logfile */
+  .close_fd_start = 6, /* stdin, stdout, stderr, logfile, network namespace, xsk->xsk_fd */
```
mmcgee-jump, line 143: Is the network namespace fd still open here? Or the kernel just doesn't reuse the fd?

src/app/fdctl/utility.c
```diff
@@ -14,19 +14,34 @@
...
-  entered = 1;
+static int namespace_original_fd = 0;
 
+int
+enter_network_namespace( const char * interface ) {
   char path[ PATH_MAX ];
-  snprintf1( path, PATH_MAX, "/var/run/netns/%s", config->tiles.quic.interface );
+  snprintf1( path, PATH_MAX, "/var/run/netns/%s", interface );
+
+  if( !namespace_original_fd ) {
```
mmcgee-jump, line 24: if( FD_LIKELY( ... )

src/app/fdctl/utility.c
```diff
@@ -14,19 +14,34 @@
...
+    namespace_original_fd = open( "/proc/self/ns/net", O_RDONLY | O_CLOEXEC );
+    if( FD_UNLIKELY( namespace_original_fd < 0 ) ) FD_LOG_ERR(( "failed to open /proc/self/ns/net (%i-%s)", errno, strerror( errno ) ));
+  }
 
   int fd = open( path, O_RDONLY | O_CLOEXEC );
   if( FD_UNLIKELY( fd < 0 ) ) FD_LOG_ERR(( "failed to open `%s` (%i-%s)", path, errno, strerror( errno ) ));
   if( FD_UNLIKELY( setns( fd, CLONE_NEWNET ) ) ) FD_LOG_ERR(( "failed to enter network namespace `%s` (%i-%s)", path, errno, strerror( errno ) ));
+  close( fd );
```
mmcgee-jump, line 32: Error checking, FD_UNLIKELY etc.

src/app/fdctl/utility.c
```diff
@@ -14,19 +14,34 @@
...
+
+  return 0;
```
mmcgee-jump, line 34: Doesn't need the return I think

src/app/fdctl/utility.c
```diff
@@ -14,19 +14,34 @@
...
+}
+
+void
+leave_network_namespace() {
+  if( !namespace_original_fd ) return;
```
mmcgee-jump, line 39: FD_UNLIKELY

src/app/frank/fd_frank_quic.c
llamb-jump: Updated to close the network namespace FD.

src/app/fdctl/utility.c
```diff
@@ -14,21 +14,37 @@
 #include <linux/netlink.h>
 #include <linux/rtnetlink.h>
 
-static int entered = 0;
+static int namespace_original_fd = 0;
 
 void
-enter_network_namespace( config_t * const config ) {
-  if( FD_LIKELY( !config->development.netns.enabled || entered ) ) return;
```
mmcgee-jump, line 21: Why did we remove this check? I think it's an error to enter the namespace if netns is not enabled, that will prevent bind from succeeding on a host device.

src/app/frank/fd_frank_quic.c
```diff
@@ -19,6 +20,12 @@ init( fd_frank_args_t * args ) {
   args->xsk = fd_xsk_join( fd_wksp_pod_map( args->tile_pod, "xsk" ) );
   if( FD_UNLIKELY( !args->xsk ) ) FD_LOG_ERR(( "fd_xsk_join failed" ));
 
+  /* We close utility.c:namespace_original_fd which is opened with enter_network_namespace and
+     closed in leave_network_namespace. This tile does not want to leave the namespace so close the FD
+     instead of leaking it. */
+  int ret = close( 4 );
```
mmcgee-jump, line 26: This feels a bit hacky and modifies the prod boot sequence to deal with development code, which I don't think we want. If you launch the Solana process before any Firedancer tiles, you can do the following sequence.

 1. Leave netns if it's entered (due to running config), this only needs to happen in `dev.c` as fdctl will never do configuration as part of `run`, close the fd.
 2. Spawn Solana tile. Not in network namespace. No fds open.
 3. Rejoin netns from path /var/... if needed.
 4. Close after joining.
 5. Spawn Firedancer tiles, no fds open.

src/app/frank/fd_frank_quic.c
```diff
@@ -133,13 +140,14 @@ static long allow_syscalls[] = {
   __NR_nanosleep, /* fd_tempo_tick_per_ns calibration */
   __NR_getpid,    /* OpenSSL RAND_bytes checks pid, temporarily used as part of quic_init to generate a certificate */
   __NR_getrandom, /* OpenSSL RAND_bytes reads getrandom, temporarily used as part of quic_init to generate a certificate */
+  __NR_sendto,    /* fd_xsk requires sendto */
```
mmcgee-jump, line 143: This should be changed to __NR_send to be maximally restrictive. I'm not sure why that wasn't defined for you, but maybe it's not in syscalls.h and you need another header. The syscall definitely exists.

mmcgee-jump:
Everyone on the team seems to have different standards for the commit messages, but I would maybe make this a bit more verbose, something like

> fdctl: allow send() syscall in quic tile, and fixes for running under network namespace
>
> QUIC uses the `send` syscall internally to cause the device to flush pending reads/writes. This should tend to happen automatically as the driver spins its own loop reading/writing our XSKs, but we have found for now it can improve performance.
>
> As well, when using network namespaces for development, we should not run Solana inside it for now since they don't support this.

src/app/fdctl/utility.c
llamb-jump: that was a mistake good catch

src/app/frank/fd_frank_quic.c
llamb-jump: __NR_send isn't available on our current hosts

src/app/fdctl/utility.c
llamb-jump: updated

src/app/frank/fd_frank_quic.c
llamb-jump: updated

mmcgee-jump (dismissed):

mmcgee-jump (approved):

PR #555, opened 2023-08-01
Ll/testing

llamb-jump:


src/ballet/base64/Local.mk
```diff
@@ -0,0 +1,3 @@
+$(call add-hdrs,fd_base64.h)
```
mmcgee-jump, line 1: This is test only code right? Prefer to have this in a test lib than shipping as part of ballet.

src/test/frank-single-transaction.sh
```diff
@@ -0,0 +1,56 @@
...
+# bash strict mode
+set -euo pipefail
+IFS=$'\n\t'
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd "${SCRIPT_DIR}/../../"
+
+# create test configuration for fddev
+TMPDIR=$(mktemp -d)
+cat > ${TMPDIR}/config.toml <<EOM
+[development]
```
mmcgee-jump, line 12: Might be better to have the actual file in the test dir?

src/test/frank-single-transaction.sh
```diff
@@ -0,0 +1,56 @@
...
+    interface = "veth_test_xdp_0"
+[layout]
+    affinity = "0-9"
+    verify_tile_count = 2
+    bank_tile_count = 2
+EOM
+export FIREDANCER_CONFIG_TOML=${TMPDIR}/config.toml
+
+# acquire and initialize everything to build
+echo y | ./deps.sh
```
mmcgee-jump, line 27: Can use richies composite deps.sh step instead of having this in here (I think it's cleaner for a few reasons, another one being you can run this on your checkout without it kinda overwriting things)

src/test/frank-single-transaction.sh
```diff
@@ -0,0 +1,56 @@
...
+git submodule init
+git submodule update
```
mmcgee-jump, line 29: I think the github checkout action is recursive so don't need this?

src/test/frank-single-transaction.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# Make absolutely sure we're working with a clean build directory
+# TODO: make clean does not delete everything in the build directory
+rm -rf build
```
mmcgee-jump, line 33: Shouldn't need this if the firecracker dir will get blown away every test run?

src/test/frank-single-transaction.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# build
+./contrib/make-j
```
mmcgee-jump, line 36: Can we have this also part of the make_test.yml steps instead of in here?

src/test/frank-single-transaction.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# start fddev, send a single transaction, and if everything works return 0
+FDDEV=./build/native/gcc/bin/fddev
+TEST_QUIC_TXN=$(find -name test_quic_txn -type f -executable)
```
mmcgee-jump, line 40: should always be in unit-test dir?

mmcgee-jump (approved):

src/test/frank-single-transaction.sh
llamb-jump: Addressed in https://github.com/firedancer-io/firedancer/pull/586 - In the future I would like to have a script we can run after a clean checkout that handles all the dependencies, building, and does a smoke test.

src/test/frank-single-transaction.sh
llamb-jump: addressed in https://github.com/firedancer-io/firedancer/pull/586

src/test/frank-single-transaction.sh
llamb-jump: addressed in https://github.com/firedancer-io/firedancer/pull/586

src/test/frank-single-transaction.sh
llamb-jump: addressed in https://github.com/firedancer-io/firedancer/pull/586

src/test/frank-single-transaction.sh
```diff
@@ -0,0 +1,56 @@
...
+set +e
+# TODO: this is flakey run give it multiple chances
+for i in {1..10}; do
+    sudo /home/runner/firedancer-opts/build/native/gcc/bin/fd_shmem_cfg fini || true
```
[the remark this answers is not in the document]
llamb-jump: Agreed and address in https://github.com/firedancer-io/firedancer/pull/586.

This line is required because without it we get bad magic errors, once we address the underlying issue we can remove it entirely.

I believe the line is currently required because the firecracker runner service uses an older binary included in the snapshot https://github.com/firedancer-io/ci-builder-fedora/blob/master/root/opt/run.sh#L5.

src/ballet/base64/Local.mk
llamb-jump: Good catch I'll follow up with ya offline on that.

src/test/frank-single-transaction.sh
llamb-jump: The underlying motivation is to explicitly guarantee the configuration at runtime, if we can do that another way let's do it.

PR #561, opened 2023-08-02
sandbox: rework open file descriptor closing

mmcgee-jump:
It makes more sense for tiles to specify what file descriptors they expect to be open, not to just randomly close stuff and hope its secure.

### DIFF
--- a/src/app/fdctl/run.c
+++ b/src/app/fdctl/run.c
@@ -106,9 +106,15 @@ tile_main( void * _args ) {
 
   if( FD_UNLIKELY( args->tile->init ) ) args->tile->init( &frank_args );
 
+  int allow_fds[ 32 ];
+  ulong allow_fds_sz = args->tile->allow_fds( &frank_args,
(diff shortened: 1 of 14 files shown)


src/app/fdctl/run.c
```diff
@@ -211,6 +217,8 @@ solana_labs_main( void * args ) {
   ADD1( "fdctl" );
   ADD( "--log", "-" );
 
+  ADD( "--firedancer-pod", "" );
```
llamb-jump, line 220: This appears to cause `fddev` to crash with:
```
error: Found argument '--firedancer-pod' which wasn't expected, or isn't valid in this context

USAGE:
    fdctl --ledger <DIR> --log <FILE>

For more information try --help
```

mmcgee-jump: Ah :) stray line from another change.

llamb-jump:
With the development configuration `fddev` is crashing with:
```
tile 5 (verify) exited with signal 31 (Bad system call)
```

llamb-jump (approved):

PR #564, opened 2023-08-03
integrate verify tile from milestone-1.1-demo

llamb-jump:
fd_frank_quic requires madvise syscall because of OpenSSL

### DIFF
--- a/src/app/frank/fd_frank_verify.c
+++ b/src/app/frank/fd_frank_verify.c
@@ -1,4 +1,5 @@
 #include "fd_frank.h"
+#include "../../ballet/txn/fd_txn.h"
 
 #include <stdio.h>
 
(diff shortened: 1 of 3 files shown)


src/app/frank/fd_frank_verify.c
```diff
@@ -171,15 +264,158 @@ run( fd_frank_args_t * args ) {
...
+    seq   = fd_seq_inc( seq, 1UL );
+    cr_avail--;
+
+    if( FD_UNLIKELY( !ctl_eom ) ) ctl_som = 0;
+    else {
+      ctl_som = 1;
+    }
   }
+
+  /* Clean up */
```
mmcgee-jump, line 402: This should all be removed, there is no cleanup in the current shutdown model (and these would all get BADSYS anyway).

src/app/frank/fd_frank_verify.c
```diff
@@ -171,15 +264,158 @@ run( fd_frank_args_t * args ) {
...
+    if( FD_UNLIKELY( ha_dup ) ) { /* optimize for the non dup case */
+      accum_ha_filt_cnt++;
+      accum_ha_filt_sz += payload_sz; //WW accum_ha_filt_sz += msg_framing + msg_sz;
+      now = fd_tickcount();
+      continue;
+    }
+
+    /* We appear to have a message to verify.  So verify it.
+
+       When running synthetic load, the synthetic data will not fail
```
mmcgee-jump, line 349: Is this comment true on main?

mmcgee-jump (approved):

src/app/frank/fd_frank_verify.c
llamb-jump: I'm not sure - for this PR I'm focusing on getting the wiring correct so packets flow from quic -> verify -> dedup.

In the next PR I'll make sure we are correctly checking signatures and not artificially inducing verification errors.

mmcgee-jump (approved):

mmcgee-jump (approved):

PR #567, opened 2023-08-08
monitor: improve performance and stability

mmcgee-jump:
Various issues with the monitor are fixed, including,

 * It would write to the terminal unbuffered using printf, which is slow but also would cause screen tearing. Switched to use a full size buffer for output. This also fixed the issues with xterm so it no longer flickers.

 * Fixed Ctrl+C to leave a clean terminal behind with just one buffer of output.

 * Removed cursor while drawing to prevent it blinking around randomly.

 * Increased default refresh rate 10x.

 * Tidied up the code structure to have `tile_t` and `link_t` types for the two output formats. Tiles no longer dump information that really belongs to links (in seq, out seq), since a tile can have multiple in/out links now it doesn't make sense.

 * Reworked the IPC object finding to be much cleaner using the workspace definitions in the config object.

 * Fixed up some of the alignment and formatting issues with the output.

 * Fixed some crashes related the seccomp filter.

### DIFF
--- a/src/app/fdctl/monitor/monitor.c
+++ b/src/app/fdctl/monitor/monitor.c
@@ -13,8 +13,8 @@ void
 monitor_cmd_args( int *    pargc,
                   char *** pargv,
                   args_t * args ) {
-  args->monitor.dt_min     = fd_env_strip_cmdline_long( pargc, pargv, "--dt-min",   NULL,   66666667.          );
-  args->monitor.dt_max     = fd_env_strip_cmdline_long( pargc, pargv, "--dt-max",   NULL, 1333333333.          );
+  args->monitor.dt_min     = fd_env_strip_cmdline_long( pargc, pargv, "--dt-min",   NULL,    6666667.          );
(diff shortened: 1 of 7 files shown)


src/app/fdctl/monitor/monitor.c
```diff
@@ -184,116 +151,157 @@ run_monitor( config_t * const config,
...
+  write_stdout( TEXT_NOCURSOR, sizeof(TEXT_NOCURSOR)-1 );
+
   for(;;) {
     /* Wait a somewhat randomized amount and then make a diagnostic
        snapshot */
     fd_log_wait_until( then + dt_min + (long)fd_rng_ulong_roll( rng, 1UL+(ulong)(dt_max-dt_min) ) );
 
-    snap( tile_cnt, snap_cur, tile_cnc, tile_mcache, tile_fseq );
+    tile_snap( tile_snap_cur, tile_cnt, tiles );
+    link_snap( link_snap_prv, link_cnt, links );
```
ptaffet-jump, line 304: This should be `link_snap_cur`, right?

mmcgee-jump: Good catch! Was wondering why it flickered on boot

PR #573, opened 2023-08-10
initiate logger before sandbox tests

marcus-jump:


src/util/sandbox/test_sandbox.c
```diff
@@ -56,47 +58,62 @@
...
+      exit( EXIT_SUCCESS );                                         \
+    }                                                               \
+} while( 0 )
+
 
 /* check_open_fds ensures that `sandbox_unthreaded` verifies
    the file descriptors we allow are exactly those that are matched. */
 void
 check_open_fds( void ) {
-  TEST_FORK_SIG(
```
mmcgee-jump, line 64: Why was this wrong? Can you remove TEST_FORK_SIG

mmcgee-jump (approved):

PR #576, opened 2023-08-10
fix(quic): retry rtv fixes

lidatong:


src/tango/quic/fd_quic.c
```diff
@@ -1181,6 +1183,11 @@ fd_quic_handle_v1_initial( fd_quic_t *               quic,
   ulong rc = fd_quic_decode_initial( initial, cur_ptr, cur_sz );
   if( FD_UNLIKELY( rc == FD_QUIC_PARSE_FAIL ) ) return FD_QUIC_PARSE_FAIL;
 
+  if ( FD_UNLIKELY( ( quic->config.role == FD_QUIC_ROLE_CLIENT || !quic->config.retry ) &&
```
sskeirik, line 1186: This looks right but there is one other condition where a token might be set --- when a server receives an address validation token from a previous NEW_TOKEN frame. A quick comment describing the two checks done here + mentioning the above third possible reason for a token to appear here might be helpful from a documentation standpoint. Some example text that could be modified:

1.  clients should never receive INITIAL packets with a non-empty token (Section 17.2.2)
2.  when servers receive an INITIAL packet with an invalid RETRY token (which is the case when RETRY is disabled or not implemented), servers should drop the INITIAL packet (Section 8.1.2)
3.  when servers receive an INITIAL packet with an invalid NEW_TOKEN token (which is the case when NEW_TOKEN is disabled or not implemented), servers should process the packet as if it had no token (Section 8.1.3)

Since this is a minor documentation issue, only the adopt the above recommendation if it makes sense with regard to your coding standards.

lidatong: yep thanks for enumerating the cases, and agreed some documentation would be good here. i'll add that

nbridge-jump (approved):

PR #582, opened 2023-08-14
fix fuzz artifact publication

marcus-jump:
Does what it says on the tin


marcus-ares:
Confirmed working in https://github.com/firedancer-io/firedancer-marcus/actions/runs/5859911145/job/15886886482

PR #587, opened 2023-08-15
Fuzz: Build fuzz artifacts on 20.04

marcus-jump:


marcus-ares:
Proved to be working on: https://github.com/firedancer-io/firedancer-marcus/actions/runs/5869524866/job/15917473923

PR #591, opened 2023-08-16
Remove frank-single-transaction from CI

ripatel-fd:
Removing because this check is failing CI on an unrelated PR https://github.com/firedancer-io/firedancer/commit/e8bc56f063e26f186ab3a7409ad1ef327d14f6a6


lheeger-jump (approved):

ripatel-fd:
Will try to fix underlying issue instead

PR #593, opened 2023-08-17
fdctl: add combined run and monitor command

mmcgee-jump:
`fddev dev` gets an argument `--monitor` which combines both running and monitoring in one command. Basically, it spawns them each in a separate process, and wires stdout from Firedancer to a pipe in the monitor so the log messages can be cleanly written between output refreshes.


ptaffet-jump (approved):
I didn't run this, but it looks okay and is a nice feature.

ptaffet-jump (approved):

PR #595, opened 2023-08-17
fdctl: various minor fixes

mmcgee-jump:
* The unused `[development.sudo]` configuration is removed, it's toggled automatically depending on if you're using fddev vs fdctl.
 * Added a `--netns` flag to `fddev` which enables the network namespace in the config.
 * Fixed the configure command to support `netns` and `cluster` stages on `fddev`, rather than having them hidden in the `dev` command.
 * If running with `--no-sandbox`, made it so the `setgid` / `setuid` calls still happen to help with debugger attach permissions.
 * Tidied up some error checking and error handling in a few configure commands.
 * Fixed XDP fini to destroy correctly when you've been running in a network namespace.
 * Remove the `fd_linux_private_boot` code which just optionally entered a network namespace, since it's not really used anymore.
 * A few other minor improvements.


src/app/fdctl/config.h
```diff
@@ -70,9 +70,9 @@ typedef struct {
     int    snapshot_fetch;
     int    genesis_fetch;
     int    poh_speed_test;
-    char   expected_genesis_hash[ 32 ];
+    char   expected_genesis_hash[ 64 ];
```
ptaffet-jump, line 73: Are these base58 encoded or binary? If binary, they should be `uchar`. If base58, the array size should be `FD_BASE58_ENCODED_32_SZ` I think

src/util/fd_util.c
```diff
@@ -10,9 +6,6 @@ fd_boot( int *    pargc,
   /* At this point, we are immediately after the program start, there is
      only one thread of execution and fd has not yet been booted. */
   fd_log_private_boot  ( pargc, pargv );
-# if defined(__linux__)
-  fd_linux_private_boot( pargc, pargv );
-# endif
```
ptaffet-jump, line 15: Are you sure we want to get rid of this? Do we ever run unit tests in a network namespace?

ptaffet-jump (commented):
I just want to double check that removing `--netns` for unit tests is okay before we merge this

ptaffet-jump (dismissed):

ptaffet-jump (approved):

PR #599, opened 2023-08-17
Switch to GitHub-managed runners

ripatel-fd:


ripatel-fd:
Will have a second try at this ... :)

PR #613, opened 2023-08-21
update frank-single-transaction.sh workflow to run on github runners

llamb-jump:
This PR reverts https://github.com/firedancer-io/firedancer/pull/504 because it broke `fddev1`.
We need to add it back now that `fddev` is in a testable, working state.


marcus-jump (approved):

ptaffet-jump:
The file is owned by `root` at the time of creation, so we can't open it. We have `CAP_CHOWN` though, so we're allowed to chown it to something we can open.

PR #616, opened 2023-08-22
util: add fd_io_strsignal like fd_io_strerror

mmcgee-jump:
The glibc implementation of `strsignal` can call malloc (and eventually brk) internally, along with futex, and some other stuff that is not async-signal-safe.  Like `fd_io_strsignal`, just implement a static version of this.


ripatel-fd:
> The glibc implementation of strsignal can call malloc (and eventually brk) internally, along with futex, and some other stuff that is not async-signal-safe. 

WTF

PR #619, opened 2023-08-22
test: free the instructions buffer

marcus-jump:
LeakSanitizer is not happy about this: https://github.com/firedancer-io/firedancer/actions/runs/5931788524/job/16084367081#step:6:94


ripatel-fd:
Should remove usage of malloc here in the future.

PR #627, opened 2023-08-23
Minor log tweaks

kbowers-jump:
- cloned process shared log mmap usage cleaned up with additional documentation
- backtracing config can be manually specified at compile time
- final I/O write buffer on stack instead of thread local buffer
- log_fileno uses -1 to indicate no permanent log file instead of 0 as -1 is an imposible value for a file descriptor (e.g. it is what open returns)
- hash dedup seed tweak so no integer IB overflow possible
- other minor commenting / formatting cleanups


src/util/log/fd_log.c
```diff
@@ -1185,6 +1208,16 @@ fd_log_private_halt( void ) {
   fd_log_private_app[0]         = '\0';
   fd_log_private_app_id         = 0UL;
 
+  if( FD_LIKELY( fd_log_private_shared_lock!=fd_log_private_shared_lock_local ) ) {
+    /* Note: this will not unmap this in any clones that also inherited
```
mmcgee-jump, line 1212: The comment isn't strictly true, if it was cloned with `CLONE_VM` then the unmap is shared, but the contract of `clone()` is that the caller is already expected to understand and handle the free/cleanup semantics.

src/util/log/fd_log.c
```diff
@@ -495,39 +498,39 @@ void fd_log_level_stderr_set ( int level ) { FD_VOLATILE( fd_log_private_level_s
...
+static int fd_log_private_shared_lock_local[1] __attribute__((aligned(128))); /* location of lock if boot mmap fails */
 
-/* Lock to sequence writes between different threads */
+void
+fd_log_private_fprintf_0( int          fd,
+                          char const * fmt, ... ) {
 
-static int * fd_log_private_shared_lock = NULL;
+  /* TODO:
+     - Consider moving to util/io as fd_io_printf or renaming to
```
mmcgee-jump, line 518: Probably a good idea, I already needed to call this "private" from some of the code in run.c that does shutdown logging inside signal handlers.

src/util/log/fd_log.c
```diff
@@ -495,39 +498,39 @@ void fd_log_level_stderr_set ( int level ) { FD_VOLATILE( fd_log_private_level_s
...
+       fd_log_printf?
+     - Is msg better to have on stack or via TLS buffer?
+     - Is msg even necessary given shared lock (probably still useful to
```
mmcgee-jump, line 521: What the be the alternative to having `msg`? Isn't some buffer needed to format the arguments, even if you did a vectored write or something you still need to format the non-strings.

mmcgee-jump (dismissed):

mmcgee-jump:
Also didn't see this fix in there? Did I miss it?

> hash dedup seed tweak so no integer IB overflow possible

src/util/log/fd_log.c
kbowers-jump: Played around with some alternative structures that would do multiple writes into the write buffer instead of staging the message body in one buffer and then copying the body the actual write buffer and/or using prepending techniques.  Not that big a deal practically for targets like x86.  Could be bigger deal on targets where memory footprint is at premium.

src/util/log/fd_log.c
kbowers-jump: Line 684.  More theoretical than anything but easy to cleanup while in there.

src/util/log/fd_log.c
```diff
@@ -495,39 +498,39 @@ void fd_log_level_stderr_set ( int level ) { FD_VOLATILE( fd_log_private_level_s
...
+       keep log message write to be a single system call best effort)
+     - Would vdprintf be useful alternative here?
```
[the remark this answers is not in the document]
kbowers-jump: Suspected as much, thought it worthwhile to bring up since superficially the behavior looks vdprintf ish.

kbowers-jump:
Line 684.  More theoretical than anything but easy to cleanup while in there.

PR #636, opened 2023-08-29
leader schedule

llamb-jump:
fd_mvcc from https://github.com/firedancer-io/firedancer/pull/512

fd_leader_schedule skeleton

update .gitignore for JetBrains

add the leader schedule to forward workspace

### DIFF
--- a/ffi/rust/firedancer-sys/src/tango/mod.rs
+++ b/ffi/rust/firedancer-sys/src/tango/mod.rs
@@ -3,7 +3,9 @@ mod dcache;
 mod fctl;
 mod fseq;
 mod mcache;
+mod mvcc;
 mod tcache;
+mod leader_schedule;
 mod xdp;
 
 pub use cnc::*;
@@ -11,5 +13,7 @@ pub use dcache::*;
 pub use fctl::*;
 pub use fseq::*;
 pub use mcache::*;
+pub use mvcc::*;
 pub use tcache::*;
+pub use leader_schedule::*;
 pub use xdp::*;
--- a/src/app/fdctl/configure/workspace.c
+++ b/src/app/fdctl/configure/workspace.c
@@ -95,6 +95,13 @@ static void xsk_aio( void * pod, char * fmt, ulong tx_depth, ulong batch_count, 
             fd_xsk_aio_new      ( shmem,    tx_depth, batch_count ) );
 }
 
+static void leader_schedule( void * pod, char * fmt, ... ) {
+  INSERTER( fmt,
+            fd_leader_schedule_align    (         ),
+            fd_leader_schedule_footprint(         ),
+            fd_leader_schedule_new      ( shmem ) );
+}
+
 FD_FN_UNUSED static void alloc( void * pod, char * fmt, ulong align, ulong sz, ... ) {
   INSERTER( sz, align, sz, 1 );
 }
@@ -307,7 +314,8 @@ init( config_t * const config ) {
         cnc   ( pod, "cnc" );
         break;
       case wksp_forward:
-        cnc   ( pod, "cnc" );
+        cnc            ( pod, "cnc" );
+        leader_schedule( pod, "leader_schedule" );
         break;
     }
 
--- a/src/app/frank/fd_frank_forward.c
+++ b/src/app/frank/fd_frank_forward.c
@@ -1,4 +1,5 @@
 #include "fd_frank.h"
+#include "../../tango/leader_schedule//fd_leader_schedule.h"
 #include <linux/unistd.h>
 
 static void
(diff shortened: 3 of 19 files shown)


ffi/rust/firedancer-sys/src/tango/mod.rs
```diff
@@ -3,13 +3,17 @@ mod dcache;
...
 mod xdp;
 
 pub use cnc::*;
 pub use dcache::*;
 pub use fctl::*;
 pub use fseq::*;
 pub use mcache::*;
+pub use mvcc::*;
 pub use tcache::*;
+pub use validator::*;
```
mmcgee-jump, line 18: Not sure about "validator", maybe something about leader schedule?

solana
mmcgee-jump, line 1: Can you PR the solana side separately to the Solana repo, then this should just be a git fetch of the submodule.

src/app/fdctl/configure/workspace.c
```diff
@@ -294,7 +301,8 @@ init( config_t * const config ) {
         cnc   ( pod, "cnc" );
         break;
       case wksp_forward:
-        cnc   ( pod, "cnc" );
+        cnc            ( pod, "cnc" );
+        leader_schedule( pod, "leader_schedule" );
```
mmcgee-jump, line 305: we need in both pack and forward right?

src/app/frank/fd_frank_forward.c
```diff
@@ -98,6 +105,18 @@ run( fd_frank_args_t * args ) {
       then = now + (long)fd_tempo_async_reload( rng, async_min );
     }
 
+    ulong begin_version = fd_mvcc_begin_read( &leader_schedule->mvcc);
+    if( begin_version != version && begin_version % 2 == 0 ) {
```
mmcgee-jump, line 109: wouldn't this skip the very first update?

mmcgee-jump: Can use FD_LIKELY here to optimize the hot path.

src/app/frank/fd_frank_forward.c
```diff
@@ -98,6 +105,18 @@ run( fd_frank_args_t * args ) {
...
+      ulong size = leader_schedule->size;
+      ulong end_version = fd_mvcc_end_read( &leader_schedule->mvcc);
+      if( begin_version == end_version ) {
```
mmcgee-jump, line 112: FD_LIKELY

src/tango/validator/fd_leader_schedule.c
```diff
@@ -0,0 +1,32 @@
...
+fd_leader_schedule_new( void * mem )
+{
+  fd_leader_schedule_t * schedule = ( fd_leader_schedule_t * )mem;
+  fd_memset( schedule, 0, fd_leader_schedule_footprint() );
+  return (void *)schedule;
+}
+
+/* TODO: LML same functionality as run.c:workspace_pod_join we should refactor */
+fd_leader_schedule_t *
+fd_leader_schedule_get( char const * app_name ) {
```
mmcgee-jump, line 15: I don't think this belongs in tango, the layout of the pod at `gaddr_lo` is a higher level thing that belongs more to fdctl or the tile, than tango.

ffi/rust/firedancer-sys/src/tango/mod.rs
llamb-jump: I'm not sure about it either - I was thinking it might be good to have a directory for all the solana-labs FFI, e.g. leader schedule, staking, slot timing, etc.

I'll change it and we can revisit once we have the other stuff on main.

src/app/fdctl/configure/workspace.c
llamb-jump: I assume so we'll need to coordinate with @ptaffet-jump on that.

src/app/frank/fd_frank_forward.c
llamb-jump: It doesn't skip the first update, I'll add the `FD_UNLIKELY`.

src/tango/validator/fd_leader_schedule.c
llamb-jump: It doesn't belong but this code won't be around for long - once we land https://github.com/firedancer-io/firedancer/blob/mmcgee/solanabank2/ffi/rust/firedancer/src/pod.rs#L44 I'll use that on the rust side and remove this function entirely.

solana
llamb-jump: https://github.com/firedancer-io/solana/pull/7

ptaffet-jump:
Something feels off with including `fd_leader_schedule` in tango. What do you think about giving mvcc an "app data region" like some of the other data structures have then just using an fd_mvcc as the type that is stored in the workspace?

mmcgee-jump:
Agree with Philip, it's a nice idea

mmcgee-jump:
Also not seeing any logic around using this in pack. Maybe it's better for philip to integrate that? But might be some thorny questions there.

PR #640, opened 2023-08-29
C17

cavemanloverboy:
Firedancer requires C Standard version C17 or later. This is not reflected in `build.rs` in `ffi/rust`.


mmcgee-jump:
Thanks for the change! Out of curiosity what was the build error without this?

cavemanloverboy:
> Thanks for the change! Out of curiosity what was the build error without this?

```
Firedancer requires C Standard version C17 or later.
```

mmcgee-jump (approved):

PR #642, opened 2023-08-30
Use ONESHELL Make mode

ripatel-fd:
Switches Firedancer's build system to [One Shell](https://www.gnu.org/software/make/manual/html_node/One-Shell.html).

This means that only one shell will be spawned per-target, as opposed to the default "one shell per line in target".

For instance, the following recipe will now only spawn one shell process instead of two:

```
.PHONY: echo
echo:
    echo A
    echo B
```

This improves Make performance (much less process spawns), but it also fixes line reordering, such that `--output-sync=target` is no longer necessary.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -34,7 +34,7 @@ jobs:
       - uses: ./.github/actions/deps
 
       - name: Build unit tests
-        run: make -j -Otarget unit-test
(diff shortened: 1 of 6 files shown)


ripatel-fd:
Marking as draft - We still need `-Otarget` but the performance benefit is nice, I guess

ripatel-fd:
Closing stale PR

PR #643, opened 2023-08-31
fdctl: add ready and txn commands

mmcgee-jump:
`fdctl ready` waits until all tiles being booted by the systeme are ready to accept transactions.  Currently ignores the bank tile since it takes a while to start and will get backpressured so it doesn't really matter if it's booted anyway.

The `txn` command is stolen from `test_quic_txn.c` but integrates nicely with `fddev` so that it automatically targets the right endpoint and can be run easily.

### DIFF
--- a/src/app/fdctl/ready.c
+++ b/src/app/fdctl/ready.c
@@ -1,0 +1,49 @@
+#include "fdctl.h"
+
+#include "run.h"
+
+#include "../../tango/fd_tango.h"
+
+void
+ready_cmd_fn( args_t *         args,
+              config_t * const config ) {
+  (void)args;
+
+  for( ulong i=0; i<config->shmem.workspaces_cnt; i++ ) {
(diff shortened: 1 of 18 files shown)


src/app/fdctl/ready.c
```diff
@@ -0,0 +1,49 @@
...
+      case wksp_forward:
+        break;
+      case wksp_quic:
+      case wksp_verify:
+      case wksp_dedup:
+      case wksp_pack: {
+        const uchar * pod = workspace_pod_join( config->name, wksp->name, wksp->kind_idx );
+        fd_cnc_t * cnc = fd_cnc_join( fd_wksp_pod_map( pod, "cnc" ) );
+        int first_iter = 1;
+        do {
```
ptaffet-jump, line 32: I think we normally do `for(;;)` for loops like this, but maybe that's just the main tile loop

ptaffet-jump (approved):

PR #645, opened 2023-09-01
Add fuzz_quic and initial corpus.

felix-asym:


ripatel-fd:
Thank you - We should have added this a long time ago

PR #647, opened 2023-09-03
ED25519 AVX-512 acceleration primitives

kbowers-jump:
`fd_r43x6.h` provides primitives for doing AVX-512 accelerated GF(2^255-19) operations used by cryptographic protocols like X25519 and ED25519.

Under the hood, this uses a 6 long limb radix 2^43 representation held in the lower 6 lanes of an AVX-512 register.  This representation allows use of the AVX-512 8-wide `_mm512_madd52{lo,hi}_epu64` intrinsics.  These intrinsics expose the raw vector 52b x 52b -> 104b unsigned integer multipliers used under the hood for vectorized double precision multiplication.  But, unlike a more obvious 52x5 representation to use these, 43x6 has enough "carry headroom" that only vectorizable approximate modulo reductions are needed in long sequential operations chains (like computing the multiplicative inverse) while still fitting into a single AVX-512 register.  Thus it is much faster in long calculations practically.  Details extensively documented in the code.

An AVX-512 accelerated implementation of `fd_x25519.c` using this is also provided.  This looks just like the IETF RFC 7748 reference code but it turns into very fast AVX-512 accelerated code.  It was benchmarked at ~1.7x faster per core than the corresponding AVX-2 implementation and ~2.7x faster per core than the corresponding unaccelerated implementation on a 2.3 GHz icelake-server core.  (FWIW, gcc-12 currently seems to yield marginally better AVX-512 code gen than clang-15.)

In the not too distant future, plan to augment the rest of the ED25519 with AVX-512 accelerated equivalents and hope to see similar gains in performance.


kbowers-jump:
Replacing this with #716

kbowers-jump:
Closed as per above

PR #327, opened 2023-05-08
CBMC Verification Framework

gtrepta:
See `verification/README.md` for installing dependencies and some details on how to run the tests.

This is a first draft of a framework for writing property tests with CBMC.

`verification/proofs` is the directory where the tests live. Running `run-cbmc-proofs.py` from there will run all of them and generate a report.

PR #648, opened 2023-09-04
Backport verification framework to main

ripatel-fd:
Mirrors https://github.com/firedancer-io/firedancer/pull/327 to add verification framework to main.


marcus-jump:
At first glance, this does not look right. I'll circle back tomorrow :)

PR #666, opened 2023-09-06
Revert "split sanitizer and fuzz publish workflows"

ripatel-fd:
This reverts commit 36e173c583cf8b5a3cc94df15aa3a28864266dea.

This commit has broken CI on main and thus should be reverted.
If you instead want to fix CI, please feel free to close this.


marcus-ares:
https://github.com/firedancer-io/firedancer/pull/667

PR #678, opened 2023-09-06
ci: publish fuzz targets to cf

marcus-jump:


marcus-ares:
The workflow has been tested in my fork.

PR #680, opened 2023-09-07
tango: add multi version concurrency control (mvcc)

mmcgee-jump:
A primitive is added for versioning data for single-producer multi-consumer contexts.  This is designed for use between Solana Labs and Firedancer for passing information like stake weights and the leader schedule which Firedancer cannot yet derive independently.

### DIFF
--- a/src/tango/mvcc/test_mvcc.c
+++ b/src/tango/mvcc/test_mvcc.c
@@ -1,0 +1,58 @@
+#include "../../util/fd_util.h"
+#include "fd_mvcc.h"
+
+FD_STATIC_ASSERT( FD_MVCC_ALIGN          ==128UL, unit_test );
+FD_STATIC_ASSERT( FD_MVCC_FOOTPRINT( 0UL)==128UL, unit_test );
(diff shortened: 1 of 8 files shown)


lidatong (dismissed):

src/tango/mvcc/test_mvcc.c
```diff
@@ -0,0 +1,26 @@
+#include "../../util/fd_util.h"
+#include "fd_mvcc.h"
+
+int
+main( int argc, char ** argv ) {
+  fd_boot( &argc, &argv );
+
+  fd_mvcc_t mvcc = { .version = 0 };
```
ptaffet-jump, line 8: This should use new and join so that those functions actually get tested

ptaffet-jump: And maybe testing the app region would be good too. I think the ideal test would spin up two threads and hammer it, but a single threaded test is fine

ptaffet-jump (approved):
The test could be better, but the API looks great!

ptaffet-jump (approved):

PR #687, opened 2023-09-07
Copy gossip and types from private

asiegel-jt:


ripatel-fd:
@asiegel-jt Could you split out the commits for updating types and adding the gossip implementation?

ripatel-fd:
@mmcgee-jump FYI fd_types does not offer safe deserialization. It will crash on untrusted input, and we don't have fuzz coverage. IMHO types should be reworked before we include gossip in 1.5.

mmcgee-jump:
Yeah, I took a read through and this is probably not yet ready for use in 1.5

asiegel-jt:
> @asiegel-jt Could you split out the commits for updating types and adding the gossip implementation?

This seems unnecessary. It's separated by directory.

ripatel-fd:
> > @asiegel-jt Could you split out the commits for updating types and adding the gossip implementation?
> 
> This seems unnecessary. It's separated by directory.

Squashing everything into a single commit makes it harder to revert individual changes and loses information who originally authored this code.

jumpsiegel:
Lets get the types stuff merged in as its own PR since it looks like people want it to be fuzzed and bullet proof.  Once that is merged/updated... gossip looks like it will be a no brainer

PR #699, opened 2023-09-14
config: Remove broken lint rule

ripatel-fd:
Removes the broken lint rule from Make. 

- Nobody uses it
- The usage of the `-or` flag in the `-find` command is broken
- It doesn't match our code style
- It blows up the entire repo - running it results in over 20000 of changes and tons of `.md5~` and `*.unc-backup~`


mmcgee-jump (dismissed):

mmcgee-jump:
Can you also move the `lint.cfg` into contrib for now?

mmcgee-jump (approved):

PR #707, opened 2023-09-15
Initial metrics tile

llamb-jump:
This PR adds a new metrics tile which aggregates measurements from other tiles and publishes them to https://metrics.solana.com. There is a single metrics being published in this PR, `__test_seq`, in `src/disco/mux/fd_mux.c`.

This functionality can be viewed at https://metrics.solana.com:3000/d/b74ee3d7-cc7e-467d-95f0-c7f1c07a1ea0/new-dashboard?orgId=1&from=1695851580220&to=1695884020369. Credentials to log in and view the dashboard are in the Firedancer/IT shared 1password vault.

We publish using [InfluxDB line protocol](https://docs.influxdata.com/influxdb/v2/reference/syntax/line-protocol/).

Highlights / Interesting Points:
- Metrics uses OpenSSL because `metrics.solana.com` requires `https`. The OpenSSL configuration had to be changed in `deps.sh` to give us required functionality.
- The sandbox had to be relaxed for the metrics tile since it opens and closes sockets, uses OpenSSL, and requires DNS. The sandbox for the metrics tile allows files to be opened, does not use a network namespace, and does not use `setup_mountns()`. In a future PR we will handle DNS ourselves by crafting requests and parsing responses which will allow us to use `setup_mountns()` and greatly reduce the number of syscalls which are whitelisted.
- A basic metrics api available in `src/util/metrics/fd_metrics.h` with a more advanced option available in `src/util/metrics/fd_metrics_impl.h`.
- A minor bug was addressed in `src/app/fddev/txn.c` so the flag `--payload-base64-encoded` works now.

Follow on work:
- Additional metrics and reworking the metrics API depending on how it's used.
- Metrics testing in CI. Will likely involve spinning up a simple HTTPS server locally with Python

### DIFF
--- a/src/app/fdctl/config.c
+++ b/src/app/fdctl/config.c
@@ -31,6 +31,12 @@ workspace_kind_str( workspace_kind_t kind ) {
     case wksp_dedup: return "dedup";
     case wksp_pack: return "pack";
     case wksp_bank: return "bank";
+    case wksp_metrics: return "metrics";
+    case wksp_metrics_quic: return "metrics_quic";
+    case wksp_metrics_verify: return "metrics_verify";
+    case wksp_metrics_dedup: return "metrics_dedup";
+    case wksp_metrics_pack: return "metrics_pack";
+    case wksp_metrics_bank: return "metrics_bank";
   }
   return NULL;
 }
@@ -69,6 +75,11 @@ memlock_max_bytes( config_t * const config ) {
       case wksp_dedup_pack:
       case wksp_pack_bank:
       case wksp_bank_shred:
+      case wksp_metrics_quic:
+      case wksp_metrics_verify:
+      case wksp_metrics_dedup:
+      case wksp_metrics_pack:
+      case wksp_metrics_bank:
         break;
       case wksp_net:
         TILE_MAX( net );
@@ -90,6 +101,9 @@ memlock_max_bytes( config_t * const config ) {
         break;
       case wksp_bank:
         TILE_MAX( bank );
+        break;
+      case wksp_metrics:
+        TILE_MAX( metrics );
         break;
     }
   }
@@ -308,6 +322,13 @@ static void parse_key_value( config_t *   config,
   ENTRY_UINT  ( ., tiles.pack,          max_pending_transactions                                  );
 
   ENTRY_UINT  ( ., tiles.bank,          receive_buffer_size                                       );
+
+  ENTRY_UINT  ( ., tiles.metrics,       depth                                                     );
+  ENTRY_UINT  ( ., tiles.metrics,       buffer_size                                               );
+  ENTRY_STR   ( ., tiles.metrics,       server                                                    );
+  ENTRY_STR   ( ., tiles.metrics,       user_pass                                                 );
+  ENTRY_STR   ( ., tiles.metrics,       db                                                        );
+  ENTRY_UINT  ( ., tiles.metrics,       post_interval_ms                                          );
 
   ENTRY_UINT  ( ., tiles.dedup,         signature_cache_size                                      );
 }
@@ -560,6 +581,42 @@ init_workspaces( config_t * config ) {
 
   config->shmem.workspaces[ idx ].kind      = wksp_bank;
   config->shmem.workspaces[ idx ].name      = "bank";
+  config->shmem.workspaces[ idx ].page_size = FD_SHMEM_HUGE_PAGE_SZ;
+  config->shmem.workspaces[ idx ].num_pages = 1;
+  idx++;
+
+  config->shmem.workspaces[ idx ].kind      = wksp_metrics;
+  config->shmem.workspaces[ idx ].name      = "metrics";
+  config->shmem.workspaces[ idx ].page_size = FD_SHMEM_GIGANTIC_PAGE_SZ;
+  config->shmem.workspaces[ idx ].num_pages = 1;
+  idx++;
+
+  config->shmem.workspaces[ idx ].kind      = wksp_metrics_quic;
+  config->shmem.workspaces[ idx ].name      = "metrics_quic";
+  config->shmem.workspaces[ idx ].page_size = FD_SHMEM_HUGE_PAGE_SZ;
+  config->shmem.workspaces[ idx ].num_pages = 1;
+  idx++;
+
+  config->shmem.workspaces[ idx ].kind      = wksp_metrics_verify;
+  config->shmem.workspaces[ idx ].name      = "metrics_verify";
+  config->shmem.workspaces[ idx ].page_size = FD_SHMEM_HUGE_PAGE_SZ;
+  config->shmem.workspaces[ idx ].num_pages = 1;
+  idx++;
+
+  config->shmem.workspaces[ idx ].kind      = wksp_metrics_dedup;
+  config->shmem.workspaces[ idx ].name      = "metrics_dedup";
+  config->shmem.workspaces[ idx ].page_size = FD_SHMEM_HUGE_PAGE_SZ;
+  config->shmem.workspaces[ idx ].num_pages = 1;
+  idx++;
+
+  config->shmem.workspaces[ idx ].kind      = wksp_metrics_pack;
+  config->shmem.workspaces[ idx ].name      = "metrics_pack";
+  config->shmem.workspaces[ idx ].page_size = FD_SHMEM_HUGE_PAGE_SZ;
+  config->shmem.workspaces[ idx ].num_pages = 1;
+  idx++;
+
+  config->shmem.workspaces[ idx ].kind      = wksp_metrics_bank;
+  config->shmem.workspaces[ idx ].name      = "metrics_bank";
   config->shmem.workspaces[ idx ].page_size = FD_SHMEM_HUGE_PAGE_SZ;
   config->shmem.workspaces[ idx ].num_pages = 1;
   idx++;
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -307,7 +307,7 @@ dynamic_port_range = "8000-10000"
     #
     # It is suggested to use all available CPU cores for Firedancer, so that the
     # Solana network can run as fast as possible.
-    affinity = "0-14"
+    affinity = "0-15"
 
     # How many net tiles to run.  Each networking tile will service exactly one
     # queue from a network device being listened to.  If there are less net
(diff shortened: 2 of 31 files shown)


src/app/fdctl/run/tiles/dedup.c
```diff
@@ -12,6 +13,7 @@ init( fd_tile_args_t * args ) {
   /* calling fd_tempo_tick_per_ns requires nanosleep, it is cached with
      a FD_ONCE */
   fd_tempo_tick_per_ns( NULL );
+  metrics_boot( args->metrics_pod, metrics_verify, args->tile_idx );
```
mmcgee-jump, line 16: Howcome this is happening in init? Might be poorly named but init is for privileged setup, and couldn't see anything privileged in there.

src/app/fdctl/run/tiles/metrics.c
```diff
@@ -0,0 +1,392 @@
...
+  /* calling fd_tempo_tick_per_ns requires nanosleep, it is cached with
+     a FD_ONCE */
+  fd_tempo_tick_per_ns( NULL );
+
+  ssl_init();
+  if( DEBUG ) FD_LOG_NOTICE(( "metrics init" ));
+}
+
+static void
+run( fd_tile_args_t * args ) {
```
mmcgee-jump, line 166: Could we make this tile use mux like the other ones, and align it to the same style / semantics

src/app/fdctl/run/tiles/metrics.c
```diff
@@ -0,0 +1,392 @@
...
+
+  ret = SSL_write( ssl, post_buffer, (int)strlen( post_buffer ) );
+  if( FD_UNLIKELY( ret <= 0 ) ) {
+    FD_LOG_WARNING(( "OpenSSL error: %d", SSL_get_error( ssl, ret ) ));
+    goto cleanup1;
+  }
+
+  /* Receive and print the response */
+  char response[4096];
+  ret = SSL_read( ssl, response, sizeof( response ) );
```
mmcgee-jump, line 136: This is a blocking read? I think we want this to be async so it doesn't block the housekeeping loop. Producers could easily get backpressured or drop packets while waiting for this.

src/app/fdctl/run/tiles/metrics.c
```diff
@@ -0,0 +1,392 @@
...
+  if( FD_UNLIKELY( ret <= 0 ) ) {
+    FD_LOG_WARNING(( "OpenSSL error: %d", SSL_get_error( ssl, ret ) ));
+    goto cleanup1;
+  }
+  if( ret > 0 ) {
+    response[ ret ] = '\0';
+    FD_LOG_NOTICE(( "response %s", response ));
+  }
+
+  /* Clean up */
```
mmcgee-jump, line 146: Probably want to parse the response for a HTTP 200?

mmcgee-jump: (and keep the buffered metrics in that case)

src/app/fdctl/run/tiles/metrics.c
mmcgee-jump: The lifecycle here is a little unclear to me. Do you need to call SSL_shutdown to send close notify? As well, one SSL_read might not return all the bytes? I think you need to keep calling read until you can at least parse the HTTP response for a 200, but you maybe just want to read the whole thing out.

src/app/fdctl/run/tiles/metrics.c
```diff
@@ -0,0 +1,392 @@
...
+  struct sockaddr sockaddr;
+  memcpy( &sockaddr, &server_addr, sizeof( struct sockaddr ) );
+  int ret = connect( ssl_sockfd, &sockaddr, sizeof( server_addr ) );
+  if( FD_UNLIKELY( ret < 0 ) ) {
+    FD_LOG_WARNING(( "connect error %d (%d) %s", ret, errno, strerror( errno ) ));
+    goto cleanup0;
+  }
+
+  /* Create an SSL context */
+  SSL_CTX * ctx = SSL_CTX_new( TLS_client_method() );
```
mmcgee-jump, line 104: These kinds of things can be initialized once at tile boot

src/app/fdctl/run/tiles/metrics.c
```diff
@@ -0,0 +1,392 @@
...
+               const ulong post_buffer_sz,
+               char * const data ) {
+  int ssl_sockfd = socket( AF_INET, SOCK_STREAM, 0 );
+  if( FD_UNLIKELY( ssl_sockfd < 0 ) ) {
+    FD_LOG_WARNING(( "socket error %d (%d) %s", ssl_sockfd, errno, strerror( errno ) ));
+    return;
+  }
+
+  /* Resolve the server address */
+  struct hostent * server = gethostbyname( _server );
```
mmcgee-jump, line 83: Per discussion, kind of hate this glibc function, god knows what it ends up doing. Could we just include the MUSL `gethostbyname` code directly?

mmcgee-jump: I also think gethostbyname is blocking, which has the same issues as below. This could be async / polled as part of the run loop instead.

src/app/fdctl/run/tiles/metrics.c
```diff
@@ -0,0 +1,392 @@
...
+          FD_LOG_WARNING(( "Unexpected signal %s (%lu) received; trying to resume", fd_cnc_signal_cstr( s, buf ), s ));
+        }
+        fd_cnc_signal( cnc, FD_CNC_SIGNAL_RUN);
+      }
+      cnc_service_deadline = now + ( long ) fd_tempo_async_reload( rng, async_min );
+    }
+  }
+}
+
+static long allow_syscalls[] = {
```
mmcgee-jump, line 342: I know it's a pain but we can we document why each of these are needed, same as the other tiles? It's very useful for auditing and code review.

src/app/fdctl/config.c
```diff
@@ -537,6 +558,42 @@ init_workspaces( config_t * config ) {
...
+  config->shmem.workspaces[ idx ].num_pages = 1;
+  idx++;
+
+  config->shmem.workspaces[ idx ].kind      = wksp_metrics_quic;
+  config->shmem.workspaces[ idx ].name      = "metrics_quic";
+  config->shmem.workspaces[ idx ].page_size = FD_SHMEM_HUGE_PAGE_SZ;
+  config->shmem.workspaces[ idx ].num_pages = 1;
+  idx++;
+
+  config->shmem.workspaces[ idx ].kind      = wksp_metrics_verify;
```
mmcgee-jump, line 573: Curious why one workspace per tile -> metrics? My feeling is it might be cleaner with just one <everyone> -> metrics workspace

src/app/fdctl/config/default.toml
```diff
@@ -514,6 +514,15 @@ dynamic_port_range = "8000-10000"
     [tiles.bank]
         receive_buffer_size = 128
 
+    [tiles.metrics]
```
mmcgee-jump, line 517: Please document heavily in line with the rest of the configuration

src/app/fdctl/run/tiles/metrics.c
```diff
@@ -0,0 +1,392 @@
...
+#define __USE_MISC
+#include <netdb.h>
+
+#define DEBUG 0
+
+#define METRICS_TAG 0x8bf72a57ca2947b7UL
+
+static void
+ssl_init( void ) {
+  SSL_library_init();
```
mmcgee-jump, line 23: Are these all necessary during privileged boot? Can you add some comments. Also error checking, same as https://github.com/firedancer-io/firedancer/blob/b0f14a28f9a4f387ab35c481760d31a8c24cf3e1/src/app/fdctl/run/tiles/quic.c#L33

src/util/metrics/fd_metrics.h
```diff
@@ -0,0 +1,79 @@
...
+  char                name[METRICS_NAME_SZ];
+} Measurement;
+
+typedef struct {
+  Measurement            measurement;
+  metrics_value          value;
+} Datapoint;
+
+void
+metrics_boot( uchar const * pod, const metrics_tile_t tile, const ulong idx );
```
mmcgee-jump, line 49: The code in util/ has much more requirements than the rest of the codebase, since it's a library.

Could we follow the style in fd_log.h a lot more closely? Eg, these exported functions and types should be prefixed with `fd_` so importers don't get symbol collisions. Use FD_PROTOTYPES_BEGIN, I don't think fd_metrics_format should be exposed, or the enum etc?

As far as I can tell, fd_metrics.c could have an API that's basically like,

* fd_metrics_boot()`
* #FD_METRICS_ULONG(tag, value) ... (or fd_metrics_push).

src/util/metrics/fd_metrics.h
```diff
@@ -0,0 +1,79 @@
...
+  double f;
+  bool   b;
+  char   s[METRICS_VALUE_STRING_SZ];
+} metrics_value;
+
+#define METRICS_NAME_SZ 40
+
+typedef struct {
+  metrics_datapoint_t type;
+  char                name[METRICS_NAME_SZ];
```
mmcgee-jump, line 40: Don't think these strings should be passing between tiles, since it's a lot of copying. Could be cleaner to have tag be a u16, then you can eliminate the dcache completely and just send mcache lines which are one cache line. Then metrics tile would map the tags back to strings before posting.

src/app/fdctl/run/tiles/dedup.c
llamb-jump: Updated to be in `run()`

src/app/fdctl/run/tiles/metrics.c
llamb-jump: The metrics tile won't use the mux tile for now since it does not require flow control and it's already implemented.

src/app/fdctl/run/tiles/metrics.c
llamb-jump: This has all been reworked to use nonblocking OpenSSL.

src/app/fdctl/run/tiles/metrics.c
llamb-jump: Updated.

src/app/fdctl/config.c
llamb-jump: We discussed this a few weeks ago and opted for one workspace per tile type for security.

src/app/fdctl/run/tiles/metrics.c
llamb-jump: They are - none of them have a return value.

src/util/metrics/fd_metrics.h
llamb-jump: Move out of `util/` and now the prototype now looks like
```
void
fd_metrics_push( uint tag, ulong value )

src/app/fdctl/config/default.toml
llamb-jump: Updated

src/app/fdctl/run/tiles/metrics.c
llamb-jump: Updated to parse the response and log if it is not correct.

src/app/fdctl/run/tiles/metrics.c
llamb-jump: Updated

PR #708, opened 2023-09-18
Disable -Wdangling-pointer in call to fd_scratch_attach

Fire30:
Compilation fails with the following error:

```
src/util/tpool/../scratch/fd_scratch.h:176:32: error: storing the
address of local variable ‘scratch_frame’ in ‘fd_scratch_private_frame’
[-Werror=dangling-pointer=]
  176 |   fd_scratch_private_frame     = (ulong *)fmem;
```

This is due to the fact that we are storing a stack pointer to a global variable. However there is a matching call to `fd_scratch_detach` that ensures the pointer is cleared.

### DIFF
--- a/config/with-brutality.mk
+++ b/config/with-brutality.mk
@@ -18,4 +18,12 @@ ifdef FD_USING_GCC
 
 CPPFLAGS+=-Wimplicit-fallthrough=2
 
+# Check for GCC version and only warn for -Wdangling-pointer if we are using 
+# gcc 13 as it is giving a false positive 
+# Note we have to add -Wdangling-pointer as in `with-gcc.mk` we disable the 
+# warning entirely
+ifeq "$(CC_MAJOR_VERSION)" "13"
+CPPFLAGS+=-Wdangling-pointer -Wno-error=dangling-pointer
+endif
+
 endif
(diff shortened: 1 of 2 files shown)


Fire30:
Updated commit as it would fail to compile on GCC < 13 due to `-Werror=dangling-pointer` not existing until GCC 13

Fire30:
I have added a version check in the brutality config and downgrade the error to a warning now. 
Would it make more sense to find a place to store a variable like `FD_GCC_MAJOR_VERSION` and be able to access it in other files?

ripatel-fd:
> I have added a version check in the brutality config and downgrade the error to a warning now.

Thanks a lot, this looks great. Could you please also ignore this lint entirely for the default config when GCC 13 is detected (in with-gcc.mk)? 

> Would it make more sense to find a place to store a variable like FD_GCC_MAJOR_VERSION and be able to access it in other files?

That's a great idea. Perhaps something like

```
CC_MAJOR_VERSION=$(shell $(CC) -dumpversion | cut -f1 -d.)
```

in `everything.mk`. That way, we get this feature for Clang too.

Fire30:
Ok I added the `CC_MAJOR_VERSION` variable in `base.mk` instead of `everything.mk` as it seems like that was the correct place to put it(no variable definitions seem to happen in `everything.mk`).

PR #712, opened 2023-09-21
disco: refactor tiles into standard layout

### LINKED ISSUE #657: Remove shared transaction workspace beween tiles
For performance it would be nice if the TPU tiles shared a workspace for transactions so they don't have to copy between each other. This isn't a good security design, because a compromised QUIC tile could alter transactions after they have been verified by sigverify.

### LINKED ISSUE #677: Audit and fix flow control issues between tiles
All of the flow control for different tiles is different, and at least one or two of them seem buggy (the QUIC tile cannot read only a partial batch from the XSK if the downstream consumer is backpressuring).

mmcgee-jump:
The tile code had been getting copy pasted and duplicated in interesting ways that were starting to show their age.  This commit refactors all of the tiles onto a standard infrastructure, the mux tile.  The mux tile is used because it is the most general.

 * All tiles are rewritten to be a small wrapper around the mux tile. The mux tile gains some functionality, namely the ability to make callbacks to the user, to enable the tiles to change its behavior. This eliminates a large amount of boilerplate and makes it obvious in what ways each tile deviate from the standard.  Mux also gains the ability to filter fragments.

 * As part of this refactor, many of the deviations from the standard were removed because they were either no longer needed or were not correct. Some of these include,

    - Each flow control imlementation was subtly different, and these differences were dropped.  All flow control is now standard mux logic.  Same goes for deciding which (if any) producer to poll.
    - Redundant metrics written to the cnc diagnostic region by the QUIC tile are removed.
    - Some tiles republished ctl, tspub, and other fields, and some did not. All tiles now republish as appropriate.
    - The pack tile distributed to N consumers with N mcaches, rather than 1 like the mux tile.  It also repurposed flow control credits to be used as a "pack" credit, which wasn't being initialized correctly but also was unclear.
    - Various others.

 * The boundary between fdctl and frank has been removed, and the boundary betwen (what was) frank and disco has been clarified. The tiles in disco are an API, with fdctl as one of the callers.  The API boundary is that disco will not allocate any memory, and expects all objects needed to communicate with the rest of the system to be provided to it with no further configuration needed. Objects that are shared between multiple processes or need to be observed by a supervisor are created by, and then mapped into the tile processes. Objects only needed by an individual tile (for example, the tcache in dedup), are created in scratch space on the stack if possible, or if they are too large, in the tile's workspace. There are no other allocations.

 * The bounds checking on tile fragment readers has been improved (or implemented at all in some cases).

 * The QUIC tile has been refactored more deeply.  The raw UDP transaction handling is moved out to disco, and the context object for that logic is merged with the QUIC context object for simplicity.

 * The shared TPU workspace between tiles has been removed as a security issue, tiles are back to copying data between each stage of the TPU pipeline.

 * The TPU MTU was removed from the config file since it's always static.

 * The forwarding tile skeleton has been removed, since we have decided not to implement this functionality.

 * Moved the frank tile pieces into fdctl and removed all references to "frank" in the codebase. Frankendancer remains only in the README, to describe the current side-by-side model on main.

 * Removed disco from FFI, there's not really any reason anyone would need to call Firedancer tiles from Rust, and it just slowed down the build.

 * The dev1 command had broken at some point due to sandbox changes, and has been restored.

 * The `fd_tempo_tick_per_ns` calibration is moved to the `init` step of each tile, if they need to access fd_tempo_tick_per_ns, rather than being passed in via. fdctl.

closes #677 
closes #657


src/disco/mux/fd_mux.h
```diff
@@ -51,6 +92,152 @@
...
+                                      ulong  sz,
+                                      int *  opt_filter );
+
+/* fd_mux_after_frag_fn is called immediately after the
+   fd_mux_during_frag_fn, along with an additional check that the reader
+   was not overrun while handling the frag.  If the reader was overrun,
+   the frag is abandoned and this function is not called.  This callback
+   is not invoked if the mux is backpressured, as it would not read a
+   frag in the first place.  It is also not invoked if
+   fd_mux_during_frag returns non-zero, indicating to filter the frag.
```
ptaffet-jump, line 175: Do you mean `returns` in a figurative sense, or did you change the function signature and forget to update the comment?

src/disco/mux/fd_mux.h
```diff
@@ -51,6 +92,152 @@
...
+
+   You should not read the frag data directly here, as it might still
+   get overrun, instead it should be copied out of the frag during the
+   read callback if needed later.
```
ptaffet-jump, line 179: The name `read callback` made me start thinking about the names of these callbacks. What do you think about naming `fd_mux_during_frag_fn` something like speculative read. Then `fd_mux_after_frag_fn` -> read confirmed or something like that?

src/disco/mux/fd_mux.h
```diff
@@ -51,6 +92,152 @@
...
+
+   The ctx is a user-provided context object from when the mux tile was
+   initialized.
+
+   opt_sig, opt_chunk, and opt_sz are the respective fields from the
+   mcache fragment that was received.  If the producer is not respecting
+   flow control, these may be corrupt or torn and should not be trusted.
+   The callback can modify these values to change the sig, chunk, and sz
+   of the outgoing frag that is being copied to downstream consumers. */
+
```
ptaffet-jump, line 193: I'd just flip the order of the second and third sentences. I initially didn't understand why the args were e.g. `ulong * opt_sig` instead of `ulong sig` as in the previous callback. This comment is pretty clear though, so it's probably my fault.
If `fd_frag_meta_t` were slightly smaller, I'd just suggest having this function return it.

src/disco/pack/fd_pack.c
```diff
@@ -0,0 +1,225 @@
...
+
+static inline void
+before_poll( void * _ctx ) {
+  fd_pack_ctx_t * ctx = (fd_pack_ctx_t *)_ctx;
+
+  if( FD_UNLIKELY( ctx->cur_slot ) ) {
+    /* overrun while processing a slot, return it to the pool */
+    fd_pack_insert_txn_cancel( ctx->pack, ctx->cur_slot );
+    ctx->cur_slot = NULL;
+  }
```
ptaffet-jump, line 41: It took me a second to understand why this was here. How do you feel about adding an `overrun` callback? The init, fini/cancel pattern is fairly common in FD, so this would kind of echo it

src/app/fdctl/run/tiles/dedup.c
```diff
@@ -0,0 +1,82 @@
...
+                 in_dcache,
+                 fd_tcache_join( fd_tcache_new( fd_wksp_alloc_laddr( fd_wksp_containing( args->tile_pod ), FD_TCACHE_ALIGN, FD_TCACHE_FOOTPRINT( tcache_depth, 0 ), 1UL ), tcache_depth, 0 ) ),
+                 fd_mcache_join( fd_wksp_pod_map( args->out_pod, "mcache" ) ),
+                 fd_dcache_join( fd_wksp_pod_map( args->out_pod, "dcache" ) ),
+                 1,
+                 &(ulong*){ fd_fseq_join( fd_wksp_pod_map( args->out_pod, "fseq" ) ) },
+                 0,
+                 0,
+                 fd_rng_join( fd_rng_new( _rng, 0, 0UL ) ),
+                 fd_alloca( FD_DEDUP_TILE_SCRATCH_ALIGN, FD_DEDUP_TILE_SCRATCH_FOOTPRINT( in_cnt, 1 ) ) );
```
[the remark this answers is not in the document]
ptaffet-jump: Maybe even a function that joins them all, since the naming convention we use in the workspace is consistent

ptaffet-jump (commented):
I didn't look at everything, but overall I like the idea a lot! The biggest things I would change are:

- Can we make the functions not take a dozen arguments?
- Can we name the callbacks more with hints of what you're supposed to do in them rather than their current names? Where they get called is important and I don't want to hide it, but I think other names might be more informative

src/app/fdctl/monitor/monitor.c
```diff
@@ -195,13 +195,24 @@ run_monitor( config_t * const config,
...
+      case wksp_pack_bank:
+      case wksp_bank_shred:
+        break;
+      case wksp_quic:
+      case wksp_verify:
+      case wksp_dedup:
+      case wksp_pack:
+      case wksp_bank:
+        tile_cnt++;
+        break;
```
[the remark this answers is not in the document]
mmcgee-jump: I agree but I want to do a bigger refactor of this structure so we don't have to do any of this stuff. Will be a follow up change.

src/disco/mux/fd_mux.h
```diff
@@ -51,6 +92,152 @@
...
+   to 0 in clear().
+
+   The ctx is a user-provided context object from when the mux tile was
+   initialized. */
+
+typedef void (fd_mux_cnc_diag_write)( void *  ctx,
+                                      ulong * cnc_app );
+
+typedef void (fd_mux_cnc_diag_clear)( void * ctx );
+
```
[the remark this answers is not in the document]
mmcgee-jump: Nope but we can add as needed.

src/disco/mux/fd_mux.h
mmcgee-jump: The latter :) will fix

src/disco/pack/fd_pack.c
mmcgee-jump: It's not a bad idea but I clarified the comment for now, I think we can add these callbacks when they are more strictly needed.

src/app/fdctl/run/tiles/dedup.c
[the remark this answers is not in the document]
mmcgee-jump: I've been thinking about how to factor all this stuff out yet but the right approach hasn't struck me yet, so going to defer this one.  It's also fwiw just copied from the mux tile directly.

src/test/single-transaction.sh
llamb-jump, line 1: The functional test is failing with:

```
[2023-09-22T15:21:41.960726464Z INFO  solana_core::snapshot_packager_service] SnapshotPackagerService has started
[2023-09-22T15:21:41.960757132Z INFO  solana_core::validator] Using: block-verification-method: blockstore-processor, block-production-method: thread-local-multi-iterator
[2023-09-22T15:21:41.960781168Z INFO  solana_core::sample_performance_service] Starting SamplePerformance service
[2023-09-22T15:21:41.960756040Z INFO  solana_core::accounts_hash_verifier] AccountsHashVerifier has started
[2023-09-22T15:21:41.960823588Z INFO  solana_runtime::accounts_background_service] AccountsBackgroundService has started
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', poh/src/poh_recorder.rs:982:113
stack backtrace:
   0: rust_begin_unwind
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/panicking.rs:64:14
   2: core::panicking::panic
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/panicking.rs:114:5
   3: solana_poh::poh_recorder::PohRecorder::new_with_clear_signal
   4: solana_core::validator::Validator::new
   5: solana_validator::main1::main
   6: solana_validator_main
   7: solana_labs_main
             at ./src/app/fdctl/run/run.c:302:3
   8: clone
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
[2023-09-22T15:21:42.227280938Z ERROR solana_metrics::metrics] datapoint: panic program="validator" thread="?" one=1i message="panicked at 'called `Option::unwrap()` on a `None` value', poh/src/poh_recorder.rs:982:113" location="poh/src/poh_recorder.rs:982:113" version="1.16.14 (src:00000000; feat:3949673676, client:SolanaLabs)"
Log at "/data/llamb/log"
tile 0 (solana-labs) exited with code 1
```

src/disco/dedup/fd_dedup.c
```diff
@@ -1,587 +1,154 @@
...
+  (void)opt_filter;
+
+  fd_dedup_ctx_t * ctx = (fd_dedup_ctx_t *)_ctx;
 
-#define SCRATCH_ALLOC( a, s ) (__extension__({                    \
-    ulong _scratch_alloc = fd_ulong_align_up( scratch_top, (a) ); \
-    scratch_top = _scratch_alloc + (s);                           \
-    (void *)_scratch_alloc;                                       \
-  }))
+  if( FD_UNLIKELY( chunk<ctx->in[in_idx].chunk0 || chunk>=ctx->in[in_idx].wmark || sz > FD_TPU_DCACHE_MTU ) )
```
llamb-jump, line 54:
```suggestion
  if( FD_UNLIKELY( chunk<ctx->in[ in_idx ].chunk0 || chunk>=ctx->in[ in_idx ].wmark || sz > FD_TPU_DCACHE_MTU ) )
```

src/disco/dedup/fd_dedup.c
```diff
@@ -1,587 +1,154 @@
...
+    FD_LOG_ERR(( "chunk %lu %lu corrupt, not in range [%lu,%lu)", chunk, sz, ctx->in[in_idx].chunk0, ctx->in[in_idx].wmark ));
```
llamb-jump, line 55:
```suggestion
    FD_LOG_ERR(( "chunk %lu %lu corrupt, not in range [%lu,%lu)", chunk, sz, ctx->in[ in_idx ].chunk0, ctx->in[ in_idx ].wmark ));
```

src/test/single-transaction.sh
mmcgee-jump: This will go away when rebased on top of philips changes.

ptaffet-jump (approved):

PR #714, opened 2023-09-22
IP stack - ARP and routing table

nbridge-jump:
This PR adds functionality to do the following, needed for XDP kernel bypass:
  Fetch ARP and Routing tables from the kernel via Netlink
  Run a routing algo for an IP address
  generate an ARP request
  prime the kernel ARP table to accept an ARP response for a specific IP address
It is only intended to support Ethernet and IPv4

### DIFF
--- a/src/tango/ip/fd_ip.c
+++ b/src/tango/ip/fd_ip.c
@@ -1,0 +1,392 @@
+#include "fd_ip.h"
+
+#include <arpa/inet.h>
+
+ulong
+fd_ip_align( void ) {
+  return FD_IP_ALIGN;
+}
+
+
+ulong
+fd_ip_footprint( ulong arp_entries,
+                 ulong route_entries ) {
+
+  ulong l;
+
+  l = FD_LAYOUT_INIT;
+  l = FD_LAYOUT_APPEND( l, FD_IP_ALIGN, sizeof(fd_ip_t)                             );
+  l = FD_LAYOUT_APPEND( l, FD_IP_ALIGN, sizeof(fd_nl_t)                             );
+  l = FD_LAYOUT_APPEND( l, FD_IP_ALIGN, arp_entries   * sizeof(fd_nl_arp_entry_t)   );
+  l = FD_LAYOUT_APPEND( l, FD_IP_ALIGN, route_entries * sizeof(fd_nl_route_entry_t) );
+
+  return FD_LAYOUT_FINI( l, FD_IP_ALIGN );
+}
+
+
+void *
+fd_ip_new( void * shmem,
+           ulong  arp_entries,
+           ulong  route_entries ) {
+  if( !fd_ulong_is_aligned( (ulong)shmem, FD_IP_ALIGN ) ) {
+    FD_LOG_ERR(( "Attempt to fd_ip_new with unaligned memory" ));
+    return NULL;
+  }
+
+  ulong l;
+  uchar * mem = (uchar*)shmem;
+
+  l = FD_LAYOUT_INIT;
+
+  fd_ip_t * ip = (fd_ip_t*)mem;
+  l = FD_LAYOUT_APPEND( l, FD_IP_ALIGN, sizeof(fd_ip_t)                             );
+
+  ulong ofs_netlink = l;
+  l = FD_LAYOUT_APPEND( l, FD_IP_ALIGN, sizeof(fd_nl_t)                             );
+
+  ulong ofs_arp_table   = FD_ULONG_ALIGN_UP( l, FD_IP_ALIGN );
+  l = FD_LAYOUT_APPEND( l, FD_IP_ALIGN, arp_entries   * sizeof(fd_nl_arp_entry_t)   );
+
+  ulong ofs_route_table = FD_ULONG_ALIGN_UP( l, FD_IP_ALIGN );
+  l = FD_LAYOUT_APPEND( l, FD_IP_ALIGN, route_entries * sizeof(fd_nl_route_entry_t) );
+
+  ulong mem_sz = FD_LAYOUT_FINI( l, FD_IP_ALIGN );
+
+  /* clear all to zero */
+  fd_memset( ip, 0, mem_sz );
+
+  /* set values in ip */
+  ip->num_arp_entries      = arp_entries;
+  ip->num_route_entries    = route_entries;
+  ip->ofs_netlink          = ofs_netlink;
+  ip->ofs_arp_table        = ofs_arp_table;
+  ip->ofs_route_table      = ofs_route_table;
+
+  /* set magic last, after a fence */
+  FD_COMPILER_MFENCE();
+  ip->magic                = FD_IP_MAGIC;
+
+  return (void*)ip;
+}
+
+
+fd_ip_t *
+fd_ip_join( void * mem ) {
+  if( !mem ) {
+    FD_LOG_ERR(( "Attempt to fd_ip_join a NULL" ));
+    return NULL;
+  }
+
+  if( !fd_ulong_is_aligned( (ulong)mem, FD_IP_ALIGN ) ) {
+    FD_LOG_ERR(( "Attempt to fd_ip_join with unaligned memory" ));
+    return NULL;
+  }
+
+  fd_ip_t * ip = (fd_ip_t*)mem;
+
+  if( ip->magic != FD_IP_MAGIC ) {
+    FD_LOG_ERR(( "Failed to fd_ip_join. Possibly memory corrupt" ));
+    return NULL;
+  }
+
+  /* initialize netlink */
+  fd_nl_t * netlink = fd_ip_netlink_get( ip );
+  if( fd_nl_init( netlink, 0 ) ) {
+    FD_LOG_ERR(( "Failed to initialize fd_netlink." ));
+    return NULL;
+  }
+
+  return ip;
+}
+
+
+void *
+fd_ip_leave( fd_ip_t * ip ) {
+  if( !ip ) {
+    FD_LOG_WARNING(( "fd_ip_leave a NULL fd_ip" ));
+    return NULL;
+  }
+
+  /* clear out the magic first */
+  ip->magic = 0;
+
+  /* then fence */
+  FD_COMPILER_MFENCE();
+
+  /* finalize the netlink */
+  fd_nl_t * netlink = fd_ip_netlink_get( ip );
+  fd_nl_fini( netlink );
+
+  fd_memset( ip, 0, sizeof( *ip ) );
+
+  return (void*)ip;
+}
+
+
+/* get pointer to fd_nl_t */
+fd_nl_t *
+fd_ip_netlink_get( fd_ip_t * ip ) {
+  ulong mem = (ulong)ip;
+
+  return (fd_nl_t*)( mem + ip->ofs_netlink );
+}
+
+
+/* get pointer to start of routing table */
+fd_ip_route_entry_t *
+fd_ip_route_table_get( fd_ip_t * ip ) {
+  ulong mem = (ulong)ip;
+
+  return (fd_ip_route_entry_t*)( mem + ip->ofs_route_table );
+}
+
+
+/* get pointer to start of arp table */
+fd_ip_arp_entry_t *
+fd_ip_arp_table_get( fd_ip_t * ip ) {
+  ulong mem = (ulong)ip;
+
+  return (fd_ip_arp_entry_t*)( mem + ip->ofs_arp_table );
+}
+
+
+void
+fd_ip_arp_fetch( fd_ip_t * ip ) {
+  fd_ip_arp_entry_t * arp_table     = fd_ip_arp_table_get( ip );
+  ulong               arp_table_cap = ip->num_arp_entries;
+  fd_nl_t *           netlink       = fd_ip_netlink_get( ip );
+
+  long num_entries = fd_nl_load_arp_table( netlink, arp_table, arp_table_cap );
+
+  if( num_entries < 0L ) {
+    return;
+  }
+
+  ip->cur_num_arp_entries = (ulong)num_entries;
+}
+
+
+/* query an arp entry
+
+   searches for an IP address in the table
+
+   if found, the resulting data is written into the destination and the function
+       returns 0
+
+   otherwise, the function returns 1 */
+
+int
+fd_ip_arp_query( fd_ip_t *            ip,
+                 fd_ip_arp_entry_t ** arp,
+                 uint                 ip_addr ) {
+  fd_ip_arp_entry_t * arp_table     = fd_ip_arp_table_get( ip );
+  ulong               arp_table_cap = ip->num_arp_entries;
+
+  fd_ip_arp_entry_t * entry = fd_nl_arp_query( arp_table, arp_table_cap, ip_addr );
+
+  if( FD_UNLIKELY( !entry ) ) return 1;
+
+  *arp = entry;
+
+  return 0;
+}
+
+
+/* generate a raw ARP packet
+
+   used for caller to generate an ARP packet to send in the event
+     we don't have an existing ARP entry
+
+   writes ARP packet into dest
+
+   if successful, returns 0
+
+   if unable to generate ARP, if the dest capacity (dest_cap) is not enough space
+     then the function returns 1 */
+
+int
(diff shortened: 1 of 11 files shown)


mmcgee-jump:
Could be worth squashing some of these smaller commits, or even squashing everything into 1 commit for saner history

src/tango/ip/fd_ip.c
```diff
@@ -0,0 +1,381 @@
...
+  ulong ofs_route_table = FD_ULONG_ALIGN_UP( l, FD_IP_ALIGN );
+  l = FD_LAYOUT_APPEND( l, FD_IP_ALIGN, route_entries * sizeof(fd_nl_route_entry_t) );
+
+  ulong mem_sz = FD_LAYOUT_FINI( l, FD_IP_ALIGN );
+
+  /* clear all to zero */
+  fd_memset( ip, 0, mem_sz );
+
+  /* set values in ip */
+  ip->magic                = FD_IP_MAGIC;
```
mmcgee-jump, line 59: Can align this to the pattern used elsewhere where magic is set last, so if it matches you know the struct has been initialized. Also with a MFENCE, eg see CNC at

https://github.com/firedancer-io/firedancer/blob/70b0fabcba4efa67c6e3aa4b42ad9368825f66cc/src/tango/cnc/fd_cnc.c#L47

src/tango/ip/fd_ip.c
```diff
@@ -0,0 +1,381 @@
...
+  ip->ofs_arp_table        = ofs_arp_table;
+  ip->ofs_route_table      = ofs_route_table;
+
+  return (void*)ip;
+}
+
+
+fd_ip_t *
+fd_ip_join( void * mem ) {
+  if( !fd_ulong_is_aligned( (ulong)mem, FD_IP_ALIGN ) ) {
```
mmcgee-jump, line 72: Add a NULL check for mem, same as the other join implementations.

src/tango/ip/fd_ip.c
```diff
@@ -0,0 +1,381 @@
...
+  if( fd_nl_init( netlink, 0 ) ) {
+    FD_LOG_ERR(( "Failed to initialize fd_netlink." ));
+    return NULL;
+  }
+
+  return ip;
+}
+
+
+void
```
mmcgee-jump, line 95: Can you align all these semantics to match the existing impls. Eg cnc / mcache / fctl, ...

leave returns a `void *` and will log warning for NULL ip etc. There's a few inconsistencies like this throughout this interface.

src/tango/ip/fd_ip.c
```diff
@@ -0,0 +1,381 @@
...
+fd_ip_arp_fetch( fd_ip_t * ip ) {
+  fd_ip_arp_entry_t * arp_table     = fd_ip_arp_table_get( ip );
+  ulong               arp_table_cap = ip->num_arp_entries;
+  fd_nl_t *           netlink       = fd_ip_netlink_get( ip );
+
+  long num_entries = fd_nl_load_arp_table( netlink, arp_table, arp_table_cap );
+
+  if( num_entries < 0L ) {
+    /* TODO don't switch */
```
mmcgee-jump, line 143: What do these TODOs mean? there's a few of them I haven't understood

src/tango/ip/fd_ip.c
```diff
@@ -0,0 +1,381 @@
...
+
+    /* map to ethernet space */
+    out_dst_mac[0] = 0x01;
+    out_dst_mac[1] = 0x00;
+    out_dst_mac[2] = 0x5e;
+    out_dst_mac[3] = (uchar)( ( ip_addr >> 020 ) & 0x7fU );
+    out_dst_mac[4] = (uchar)( ( ip_addr >> 010 ) & 0xffU );
+    out_dst_mac[5] = (uchar)( ( ip_addr >> 000 ) & 0xffU );
+
+    /* TODO choose an ifindex */
```
mmcgee-jump, line 297: Noting the TODO. What's a good strategy? Send to same ifindex that routes to 1.1.1.1 or index 0 if not found?

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
...
+  nl_request.nlh.nlmsg_flags = NLM_F_REQUEST | NLM_F_ROOT;
+#endif
+  nl_request.nlh.nlmsg_len   = sizeof(nl_request);
+  nl_request.nlh.nlmsg_pid   = 0;
+  nl_request.nlh.nlmsg_seq   = seq;
+
+  /* request only IPv4 routes */
+  nl_request.rtm.rtm_family  = AF_INET;
+
+  long sent = send( fd, &nl_request, sizeof(nl_request), 0 );
```
mmcgee-jump, line 215: In practice you're probably guaranteed that this won't send a partial write on any reasonable platform, but for sanity we should either assert that, or keep calling send here until sent_bytes == sizeof(nl_request)

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
...
+  /* set alignment such that the returned data is aligned */
+  uchar __attribute__(( aligned(16) )) ibuf[4096]   = {0};
+
+  /* Pointer to the messages head */
+  struct nlmsghdr *h = (struct nlmsghdr *)ibuf;
+
+  /* Try to read the message in case of busy or interrupted call */
+  while(1) {
+    do {
+      len = recv( fd, ibuf, sizeof(ibuf), 0 );
```
mmcgee-jump, line 233: As above, either assert that len == expected, or you need to repeatedly call receive until we get all the bytes we want.

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
...
+    } while( len < 0 && (errno == EINTR || errno == EAGAIN) );
+
+    if( len < 0 ) {
```
mmcgee-jump, line 236: Might be useful to sprinkle in some FD_UNLIKELY for these ones that are called in our main loops?

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
...
+      FD_LOG_WARNING(( "netlink recv failed with %d %s", errno, strerror( errno ) ));
```
mmcgee-jump, line 237: I think you basically need to abandon the channel after this? There might be partial data in there that gets read next time so it seems difficult to recover from.

mmcgee-jump: Oh I see, nvm the seq number below protects it

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
...
+    /* current route entry */
+    fd_nl_route_entry_t * entry = route_table + route_entry_idx;
+
+    /* clear current entry */
+    fd_memset( entry, 0, sizeof( *entry ) );
+
+    if( h->nlmsg_flags & NLM_F_DUMP_INTR ) {
+      /* function was interrupted - don't switch to new routing table */
+
+      return -1; /* return failure */
```
mmcgee-jump, line 269: Don't love this error code here. Most of these errors are unrecoverable and we would want to FD_LOG_ERR and crash in response, but this is a benign retryable condition. Can you separate the codes with a good note about this in the header comments?

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
...
+
+  uint net_ip_addr = htonl( ip_addr );
+
+  /* format the request */
+
+  /* Request struct */
+  struct {
+    struct nlmsghdr nlh;      /* Netlink header */
+    struct ndmsg    ndm;      /* Payload - neighbor message */
+    uchar           buf[128]; /* TODO choose approproate length */
```
mmcgee-jump, line 664: looks like this is known below

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
...
+  struct nlmsghdr *h = (struct nlmsghdr *)ibuf;
+
+  /* iterate thru until no more messages looking for error */
+  while(1) {
+    do {
+      len = recv( fd, ibuf, sizeof(ibuf), 0 );
+    } while( len <= 0 && errno == EINTR );
+
+    if( len < 0 ) {
+      if( errno == EAGAIN ) break;
```
mmcgee-jump, line 723: Should keep recving on EAGAIN? Also see notes above on recv'ing the whole buffer, or at least checking len is what we expect

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
+/* TODO pare these down to only necessary */
```
mmcgee-jump, line 1: Noting this todo

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <errno.h>
+#include <unistd.h>
+
+#include "fd_netlink.h"
+
+#if 0
+#define FD_LOG( x ) \
```
mmcgee-jump, line 19: Doesn't seem needed?

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
...
+
+  rqs_rat->rta_type = NDA_DST;
+  rqs_rat->rta_len =  RTA_LENGTH(4);
+
+  memcpy( RTA_DATA(rqs_rat), &net_ip_addr, 4 );
+  nl_request.nlh.nlmsg_len += sizeof(struct rtattr) + 4;
+
+  ulong nl_request_sz = (ulong)nl_request.nlh.nlmsg_len;
+
+  long sent = send( fd, &nl_request, nl_request_sz, 0 );
```
mmcgee-jump, line 701: As above, want to check that sent is the expected size or send() in a loop til the buffer is drained

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
...
+    /* clear current entry */
+    fd_memset( entry, 0, sizeof( *entry ) );
+
+    if( h->nlmsg_flags & NLM_F_DUMP_INTR ) {
+      /* function was interrupted - don't switch to new routing table */
+
+      return -1; /* return failure */
+    }
+
+    /* TODO can you spuriously get data from another process?? */
```
mmcgee-jump, line 512: This is misnamed, but is basically just an ID the kernel echoes back to the process. You can delete the commented code.

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
...
+      return -1; /* return failure */
+    }
+
+    /* current arp entry */
+    fd_nl_arp_entry_t * entry = arp_table + arp_entry_idx;
+
+    /* clear current entry */
+    fd_memset( entry, 0, sizeof( *entry ) );
+
+    if( h->nlmsg_flags & NLM_F_DUMP_INTR ) {
```
mmcgee-jump, line 506: Same as comment above, I think this error case would be handled differently by the caller and should be documented / separated out.

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
...
+  long arp_entry_idx = 0L;
+
+  /* interpret the response */
+
+  long msglen = len;
+
+  /* Iterate through all messages in buffer */
+
+  while( NLMSG_OK(h, msglen) ) {
+    /* are we still within the bounds of the table? */
```
mmcgee-jump, line 494: Can we log warning here, this could be hard to diagnose.

src/tango/ip/fd_netlink.c
```diff
@@ -0,0 +1,760 @@
...
+
+  /* set alignment such that the returned data is aligned */
+  uchar __attribute__(( aligned(16) )) ibuf[4096]   = {0};
+
+  /* Pointer to the messages head */
+  struct nlmsghdr *h = (struct nlmsghdr *)ibuf;
+
+  /* Try to read the message in case of busy or interrupted call */
+  while(1) {
+    //uchar *     p     = ibuf;
```
mmcgee-jump, line 470: nit: commented lines

src/tango/ip/fd_ip.h
```diff
@@ -0,0 +1,295 @@
...
+
+/* Prepares the ARP cache for recieving a new ARP entry
+
+   Should be called prior to sending an ARP request
+
+   The kernel ignores unsolicited ARP responses, so this function
+   allows our ARP requests to update the kernel ARP table
+
+   It only adds an entry in the case one does not already exist
+   The entry added is in state "INCOMPLETE", and gets updated to a
```
mmcgee-jump, line 284: Kind of a nit but sometimes this should be PROBE rather than INCOMPLETE I think? Not sure if there's a difference in the kernel?

nbridge-jump: The state is "incomplete" in netlink

src/tango/ip/fd_ip.c
nbridge-jump: Good catch.
Although FD_VOLATILE is incorrect there - volatile writes only guarantee in-order writes with respect to other volatiles. I'll use a fence

src/tango/ip/fd_ip.c
nbridge-jump: I guess with sockets, broadcast and multicast are either bound to an interface, or will use ALL multicast or broadcast interfaces available respectively. 
So the routing algo really doesn't need to do anything here.

src/tango/ip/fd_netlink.c
nbridge-jump: I thought about this. I'm not at all convinced that netlink can do a partial read at all. But on balance, it's the nature of the interface. Will fix

src/tango/ip/fd_netlink.c
nbridge-jump: We can't know ahead of time how many bytes were going to get back. Examples on the internet don't have code for partial reads. In fact, many use SOCK_DGRAM instead of SOCK_RAW. The docs say these aren't treated differently.
It's not clear how the read code /should/ be.

src/tango/ip/fd_netlink.c
nbridge-jump: recv is a syscall. Micro-optimizations like FD_(UN)LIKELY aren't likely to be observable

src/tango/ip/fd_netlink.c
nbridge-jump: I originally had this returning void. I decided to return the number of entries, and forgot to add named constants. Will fix

src/tango/ip/fd_netlink.c
nbridge-jump: EAGAIN will continue indefinitely

src/tango/ip/fd_netlink.c
nbridge-jump: correct - removing

src/tango/ip/fd_netlink.c
nbridge-jump: I expect caller to retry later in any case

src/tango/ip/fd_netlink.c
nbridge-jump: I kept -1 as the only error. It's assumed to be transient, and that's in the comments in the header. Any known non-transients are handled with FD_LOG_ERR instead

src/tango/ip/fd_netlink.c
nbridge-jump: done

src/tango/ip/fd_ip.c
mmcgee-jump: The standard pattern in CNC is to use two MFENCEs, one on each side, along with the FD_VOLATILE

src/tango/ip/fd_netlink.c
mmcgee-jump: I think you need to read the first few bytes to get nlmsg_len, then keep polling with recv until you have read at most nlmsg_len.

src/tango/ip/fd_netlink.c
mmcgee-jump: It looks here like EAGAIN breaks out of the loop? Am I reading this incorrectly?  Also note above about recv applies here, too checking read_bytes >= nlmsg_len

mmcgee-jump (approved):

mmcgee-jump (approved):

PR #715, opened 2023-09-22
[ci/test/fuzz]: combi tests

marcus-jump:
This change does a few things:

* Adds three new machine targets (`combi_{lowend,modern,highend}`) which aim to increase test/fuzz coverage
  * highend is based on icelake
  * modern is based on x86
  * lowend is based on noarch128
* Gets rid of machine targets `linux_clang_x86_64_asan` and `linux_clang_x86_64_fuzz_asan`
* Adds extras: with-fuzz, with-asan, with-ubsan
* Adds a make target, `combicov-report`, for merging coverage of all of the combi targets
* Tests both combi targets against asan and ubsan
* Temporarily disables coverage generation and ClusterFuzz corpus testing (will be addressed in another PR)


marcus-ares:
~It is expected that the `Unit Tests with Sanitizers` are red. They should be fixed in another PR.~

.github/workflows-disabled/coverage.yml
```diff
@@ -32,22 +32,33 @@ jobs:
           submodules: recursive
 
       - uses: ./.github/actions/deps
+      - uses: ./.github/actions/hugepages
```
marcus-jump, line 35: Note that this file workflow is disabled and will be adjusted in an upcoming PR.

.github/workflows-disabled/test_fuzz_regressions.yml
```diff
@@ -0,0 +1,28 @@
+name: Test Fuzz Regressions
```
marcus-jump, line 1: Note that this workflow is also disabled.

PR #716, opened 2023-09-22
AVX512 accelerated ED25519 and X25519 implementations

kbowers-jump:
This replaces PR #647

Quick and dirty benchmarks on a single icelake-server core at 2.3 GHz under gcc-12 yields:

```
       |         implementation         |    speed up over
op     | portable |     AVX2 |   AVX512 | portable |     AVX2
-------+----------+----------+----------+----------+---------
x25519 | 135.0 us |  84.8 us |  34.7 us |     3.9x |     2.4x
keygen |  46.4 us |  25.0 us |  11.4 us |     4.1x |     2.2x
sign   |  49.5 us |  26.8 us |  13.2 us |     3.8x |     2.0x
verify | 154.4 us |  70.3 us |  34.2 us |     4.5x |     2.1x
```

Roughly, this implementation is ~4x faster than the existing scalar implementation and ~2x faster than the existing AVX2 implementation on the same AVX512 supporting hardware with the same compiler and build flags.  YMMV for other processor architectures.  Wallclock performance is expected for operations like this to scale with clock speed.

(clang-15 does not seem to optimize AVX512 as well as gcc-12 and runs the AVX512 implementation 10-15% slower.)

Details documented extensively in the code and commit comments, including additional potential algorithmic and implementation improvements, areas in which other implementations diverge from the standard and how to tweak this implementation to match an particular implementation behaviors.


kbowers-jump:
Since Marc was asking, the above measurements was with Turbo mode disabled (specifically the "turbo mode disable bit", bit 38 in MSR 0x1a0 was set ... for reference, typically run benchmarks without turbo mode to get stable results, especially when benchmarking an _implementation_ ... real pain to do, say, A/B testing to tune up an implementation when Intel is changing the clock speed between and during runs on a whim).

The exact processor is a Platinum 8380: https://www.intel.com/content/www/us/en/products/sku/212287/intel-xeon-platinum-8380-processor-60m-cache-2-30-ghz/specifications.html

Did a quick and dirty experiment on the same core with Turbo mode enabled (but didn't poke around all the clock frequency ratio / uncore ratio / governor / thermal / voltage MSRs and/or kernel cpufreq governors to try to max out clock frequency and/or stabilize system performance ... Intel tends to not let AVX-512 heavy code hit the top frequencies anyway so have no clue how close or stable the clock got to Turbo max speed of 3.4 GHz on this processor).  This got X25519 to 27.5 us, keygen at 9.1 us, signing at 10.5 us, verify at 27.7 us ... about +25% faster.  YMMV.

kaasaraai-jump:
Javier, Emanuele and I have been going through the code, looks impressive / nice-to-follow so far.  Specifically so far we've covered ED25519 mul-modp at bit-level and sig-verify / signing at data-flow level.

jvarela-jump (approved):

PR #719, opened 2023-09-26
wsample: refactor

ptaffet-jump:
`delete` is already used in the `init, join, leave, delete` pattern, so it's probably better to use a different word.
When using `fd_wsample` in practice, we often had a struct with the weight as a field, so calling the only `fd_wsample_new` function with an array of weights was cumbersome. Now we can just iterate through the array of structs and add each weight individually.


mmcgee-jump (dismissed):

src/ballet/wsample/fd_wsample.h
```diff
@@ -52,81 +52,119 @@ void *            fd_wsample_delete   ( void * shmem  );
...
+   fd_wsample_new_fini format a memory region with the appropriate
+   alignment and footprint to be usable as a weighted sampler.  This
+   multi-function intitialization process prevents needing to construct
+   a flat array of weights, which is often inconvenient.
+
+   The caller must first call fd_wsample_new_init, then new_add_weight
+   ele_cnt times, and finally new_fini.  Only at that point will the
+   region of memory be ready to be joined.
+
+   fd_wsample_new_init begins the formartting a memory region.
```
mmcgee-jump, line 75: nit: typo "formatting"

mmcgee-jump: Another nit this whole comment could be rewrapped, the line breaking is a bit off.

mmcgee-jump (approved):

mmcgee-jump (approved):

PR #723, opened 2023-09-26
tiles: rename quic to serve, improve docs

mmcgee-jump:


src/app/fdctl/config/default.toml
```diff
@@ -1,564 +1,631 @@
...
+    # `--enable-rpc-bigtable-ledger-storage` argument.
     bigtable_storage = false
 
-    # If nonempty, enable an accounts index indexed by the specified field. The
-    # account field must be one of "program-id", "spl-token-owner", or
-    # "spl-token-mint". These options are passed to the Solana Labs client with
-    # the `--account-index` argument.
+    # If nonempty, enable an accounts index indexed by the specified
+    # field.  The account field must be one of "program-id",
+    # "spl-token-owner", or "spl-token-mint".  These options are passed
```
ptaffet-jump, line 91: The `--account-index` flag seems to be able to take multiple of these values

PR #743, opened 2023-10-02
txn-update: fix updates to support test suites

jumpsiegel:


src/ballet/txn/fd_txn_parse.c
```diff
@@ -147,16 +155,18 @@ fd_txn_parse( uchar const             * payload,
...
-    parsed->instr[ j ]._padding_reserved_1 = (uchar)0;
-    parsed->instr[ j ].acct_cnt            = acct_cnt;
-    parsed->instr[ j ].data_sz             = data_sz;
-    /* By our invariant, i<size when it was copied into acct_off and data_off,
-       and size<=USHORT_MAX from above, so this cast is safe */
-    parsed->instr[ j ].acct_off            = (ushort)acct_off;
-    parsed->instr[ j ].data_off            = (ushort)data_off;
+    CHECK( allow_zero_signatures | ((0UL < (ulong)program_id) & ((ulong)program_id < (ulong)acct_addr_cnt) ) );
+
+    if (NULL != parsed) {
```
ptaffet-jump, line 160:
```suggestion
    if( NULL != parsed ){
```

src/ballet/txn/fd_txn_parse.c
```diff
@@ -184,39 +194,120 @@ fd_txn_parse( uchar const             * payload,
...
+  } else {
+    transaction_version = FD_TXN_VLEGACY;
+  }
+  i++;
+  i++;
+
+  ushort acct_addr_cnt = (ushort)0;
+  READ_COMPACT_U16( acct_addr_cnt, i );
+  i += FD_TXN_ACCT_ADDR_SZ*acct_addr_cnt;
+  i += FD_TXN_BLOCKHASH_SZ;
```
ptaffet-jump, line 278: Unless this is designed for only trusted inputs, this function is missing a bunch of bounds checks. If acct_addr_cnt is huge, then `i` will go past payload_sz. `payload_sz-_where` will underflow.  It'll read a bunch of data past the end of the array, and I'm not convinced the final check at the end will necessarily catch it, because it could maybe overflow again.

src/ballet/txn/fd_txn.h
```diff
@@ -632,7 +642,57 @@ static inline ulong FD_FN_CONST fd_txn_acct_iter_end( void ) { return FD_TXN_ACC
...
+
+static inline int
+fd_txn_is_signer( fd_txn_t const * txn, int idx ) {
+  return idx < txn->signature_cnt;
+}
+
+static inline ulong fd_txn_num_writable_accounts( fd_txn_t * txn ) {
+  return (ulong)((txn->signature_cnt - txn->readonly_signed_cnt)
+      + ((txn->acct_addr_cnt - txn->readonly_unsigned_cnt) - txn->signature_cnt));
+}
```
ptaffet-jump, line 695: You can replace this with the new `fd_txn_account_cnt` function which is more general and more precise. E.g. without looking at the implementation, I wouldn't know if this included writable accounts in the address lookup tables or not.

ptaffet-jump (changes requested):
fd_txn_xray needs lots of bounds checks and some documentation. (or maybe if I'm misunderstanding it, it just needs documentation to the effect of "trusted inputs only")
Also, can you fix the spacing around `if` statements to match the rest of the FD codebase. I know it's unusual, but it's the FD format.

src/ballet/txn/fd_txn_parse.c
jumpsiegel: added in bounds checks

ptaffet-jump: I don't think they're quite right. E.g. the first check checks `payload_sz>=0` when it needs to check `payload_sz>=1`. The one on line 261 doesn't protect the read that follows of the header, etc.

jumpsiegel: how is this?

src/ballet/txn/fd_txn.h
jumpsiegel: removed

asiegel-jt (approved):

mmcgee-jump:
Could you please describe the purpose of `fd_txn_xray` a little bit more? It's unclear what standard we should apply to this code review, if this is a parser for untrusted inputs, it needs to be more rigorous, and ideally include a fuzzer and comprehensive bounds checking.

mmcgee-jump:
On the other hand, if it's for test only code it maybe should be moved to a test specific location and be documented that way, in which case I don't particularly care about the parsing.

jumpsiegel:
> On the other hand, if it's for test only code it maybe should be moved to a test specific location and be documented that way, in which case I don't particularly care about the parsing.

It has bounds checking and has been fuzzed.  We don't think it will go beyond the bounds of data supplied.  Its purpose is to just quickly give us an estimate of the end of a txn so that we can walk through a microblock.  It does not tell us if the txn is valid or not but just finds the end.

We could instead call the main parse with something to tell it to copy less data and run faster...

ptaffet-jump (approved):

PR #745, opened 2023-10-02
tiles: add net and netmux tiles

mmcgee-jump:
A networking tile is addded to decouple packet receipt from the various consumers that may wish to handle packets, in the near term the QUIC and shred tiles.  The number of net tiles should exactly match the number of network queues being serviced.

A netmux tile is also added, which zero-copy muxes fragments from all the networking tiles to the interested consumers, and then back in the other direction for the tx path.

### DIFF
--- a/src/app/fdctl/config.c
+++ b/src/app/fdctl/config.c
@@ -15,21 +15,35 @@
 
 FD_IMPORT_CSTR( default_config, "src/app/fdctl/config/default.toml" );
 
-static workspace_config_t *
-find_wksp( config_t * const config,
-           char * name ) {
-  for( ulong i=0; i<config->shmem.workspaces_cnt; i++ ) {
-    workspace_config_t * wksp = &config->shmem.workspaces[ i ];
-    if( FD_UNLIKELY( !strcmp( wksp->name, name ) ) ) return wksp;
+FD_FN_CONST char *
+workspace_kind_str( workspace_kind_t kind ) {
+  switch( kind ) {
+    case wksp_netmux_inout: return "netmux_inout";
+    case wksp_quic_verify: return "quic_verify";
+    case wksp_verify_dedup: return "verify_dedup";
+    case wksp_dedup_pack: return "dedup_pack";
+    case wksp_pack_bank: return "pack_bank";
+    case wksp_bank_shred: return "bank_shred";
+    case wksp_net: return "net";
+    case wksp_netmux: return "netmux";
+    case wksp_quic: return "quic";
+    case wksp_verify: return "verify";
+    case wksp_dedup: return "dedup";
+    case wksp_pack: return "pack";
+    case wksp_bank: return "bank";
   }
-  FD_LOG_ERR(( "no workspace with name `%s` found", name ));
+  return NULL;
 }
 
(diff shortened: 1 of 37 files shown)


src/app/fdctl/run/run.h
```diff
@@ -6,33 +6,35 @@
...
-   uchar const * in_pod;
-   uchar const * out_pod;
-   fd_xsk_t    * xsk;
-   fd_xsk_t    * lo_xsk;
+   int                pid;
+   char *             app_name;
+   char *             tile_name;
+   ulong              tile_idx;
+   ulong              idx;
+   const uchar **     wksp_pod;
```
ptaffet-jump, line 14: How many workspaces are in this array? Would it be easier to make it a static array of 16 or something? I guess that would only save one alloca

src/app/fdctl/run/run.h
```diff
@@ -6,33 +6,35 @@
...
 } fd_tile_args_t;
 
 typedef struct {
-   char *  name;
-   char *  in_wksp;
-   char *  out_wksp;
-   ushort  allow_syscalls_sz;
-   long *  allow_syscalls;
+   char *             name;
+   ushort             allow_workspaces_sz;
```
ptaffet-jump, line 21: I'd prefer to rename this `_cnt`, but I don't feel too strongly about it

src/disco/mux/fd_mux.c
```diff
@@ -544,6 +532,20 @@ fd_mux_tile( fd_cnc_t *              cnc,
       continue;
     }
 
+    ulong sig = fd_frag_meta_sse0_sig( seq_sig );
+    if( FD_UNLIKELY( callbacks->before_frag ) ) {
+      int filter = 0;
+      callbacks->before_frag( ctx, (ulong)this_in->idx, seq, sig, &filter );
```
ptaffet-jump, line 538: Does moving the call to this callback not change it?

ptaffet-jump (dismissed):
I didn't read everything carefully, but it mostly looks good.

src/app/fdctl/config.c
```diff
@@ -15,51 +15,67 @@
...
-                                      out_bytes );                            \
+#define TILE_MAX( tile ) do {                                   \
+    ulong in_bytes = 0, out_bytes = 0;                          \
+    for( ulong i=0; i<tile.allow_workspaces_sz; i++ ) {         \
+      workspace_kind_t kind = tile.allow_workspaces[ i ];       \
+      workspace_config_t * in_wksp = find_wksp( config, kind ); \
+      in_bytes += in_wksp->num_pages * in_wksp->page_size;      \
+    }                                                           \
+    memlock_max_bytes = fd_ulong_max( memlock_max_bytes,        \
+                                      in_bytes + out_bytes );   \
```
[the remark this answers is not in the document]
mmcgee-jump: yup!

src/disco/mux/fd_mux.c
mmcgee-jump: It had to be moved, because otherwise we don't know seq yet (there may not even be a new message)

ptaffet-jump (approved):
Restamp

PR #746, opened 2023-10-03
shred: Add fd_fec_resolver (shreds->FEC sets)

ptaffet-jump:


src/disco/shred/fd_fec_resolver.h
```diff
@@ -0,0 +1,162 @@
...
+   completions.
+
+   It might seems like this system is overkill and one combined longer
+   freelist would suffice, but that's incorrect. Consider the following
+   scenario: Suppose we've just completed FEC set A.  Then we receive
+   lots of shreds for lots of FEC sets, but never enough to complete any
+   one of them.  We churn through the whole singular freelist with
+   spilling until A is at the head of the freelist.  We receive shreds
+   for another FEC set, so we reuse the memory we used for A, and it's
+   the first set that we can complete, so we end up returning the same
```
mmcgee-jump, line 79: I don't understand this part. When we "reuse the memory we used for A", it would be considered dirty for A, so the downstream consumer wouldn't use it anymore, then if the "new A" completes, it goes onto the freelist? What does it mean to "return the same memory twice in a row"

src/disco/shred/fd_fec_resolver.c
```diff
@@ -0,0 +1,527 @@
...
+  void * cmplst      = FD_SCRATCH_ALLOC_APPEND( l, freelist_align(),       freelist_footprint( complete_depth+1UL  )       );
+  void * bmfree      = FD_SCRATCH_ALLOC_APPEND( l, bmtrlist_align(),       bmtrlist_footprint( depth+1UL )                 );
+  FD_SCRATCH_ALLOC_FINI( l, FD_FEC_RESOLVER_ALIGN );
+
+  resolver->done             = fd_tcache_join( done   );
+  resolver->curr_map         = ctx_map_join  ( curr   );
+  resolver->free_list        = freelist_join ( free   );
+  resolver->complete_list    = freelist_join ( cmplst );
+  resolver->bmtree_free_list = bmtrlist_join ( bmfree );
+  fd_sha512_join( resolver->sha512 );
```
mmcgee-jump, line 206: I think the API of join is that you will return NULL here if these inner joins fail?

src/disco/shred/fd_fec_resolver.c
```diff
@@ -0,0 +1,527 @@
...
+  /* Invariants:
+      * no key is in both the done tcache and the current map
+      * each set pointer provided to the new function is in exactly one
+          of curr_map, freelist, or complete_list
+      * bmtree_free_list has exactly partial_depth fewer elements than
+          freelist
+   */
+
+  int dup = 0;
+  (void)dup;
```
mmcgee-jump, line 244: Why (void) ?

src/disco/shred/fd_fec_resolver.c
```diff
@@ -0,0 +1,527 @@
...
+    if( FD_UNLIKELY( freelist_cnt( free_list )<=partial_depth ) ) {
+      /* Packet loss is really high and we have a lot of in-progress FEC
+         sets that we haven't been able to finish.  Take the resources
+         bmtree from the oldest, and send the oldest FEC set to the back
+         of the free list. */
+      ulong oldest_key = resolver->curr_tail_tag;
+      set_ctx_t * victim_ctx = ctx_map_query( curr_map, oldest_key, NULL );
+      if( FD_UNLIKELY( !victim_ctx ) ) FD_LOG_ERR(( "data structures not in sync. Couldn't find %lx in the map", oldest_key ));
+      resolver->curr_tail_tag = victim_ctx->prev_tag;
+      ctx_map_query( curr_map, victim_ctx->prev_tag, NULL )->next_tag = 0UL;
```
mmcgee-jump, line 294: Also check for corruption here if ctx_map_query is NULL ?

src/disco/shred/fd_fec_resolver.c
```diff
@@ -0,0 +1,527 @@
...
+          of curr_map, freelist, or complete_list
+      * bmtree_free_list has exactly partial_depth fewer elements than
+          freelist
+   */
+
+  int dup = 0;
+  (void)dup;
+  /* Note: we identify FEC sets by the first 64 bits of their signature.
+     Given how slow Ed25519 is and how short of a time these are
+     relevant for, this seems safe, but we should research the issue
```
mmcgee-jump, line 247: I really don't like the 64 bit truncation here, we aren't cryptographers and this has consensus implications if we can't replay certain shreds that the rest of the cluster can.

Can we just make this AVX dependent and use all 256 bits?

See https://crypto.stackexchange.com/questions/45291/are-the-first-4-bytes-of-a-ed25519-public-key-random

src/disco/shred/fd_fec_resolver.c
```diff
@@ -0,0 +1,527 @@
...
+     further. */
+
+  /* Are we already done with this FEC set? */
+  /* TODO: xor this with something semi-secret to prevent someone from
+     spending a lot of compute to make a shred with a 0 signature which
```
mmcgee-jump, line 252: This is also resolved by the using all 256 bits. Even though it should be impossible, I'd be more comfortable adding a `FD_TEST( signature )` to validate that assumption, since it breaks the invariants of the map, seems like a vector an attacker could use to get an arbitrary write primitive.

src/disco/shred/fd_fec_resolver.c
```diff
@@ -0,0 +1,527 @@
...
+
+  fd_bmtree_hash_leaf( _leaf, (uchar const *)shred + sizeof(fd_ed25519_sig_t), merkle_protected_sz, FD_BMTREE_LONG_PREFIX_SZ );
+
+  /* code_idx: if this is a Merkle coding shred, the index of it in the
+     Merkle array, otherwise 0. */
+  ulong code_idx   = fd_ulong_if( shred_type==FD_SHRED_TYPE_MERKLE_DATA, 0UL, shred->code.idx );
+  ulong shred_idx  = fd_ulong_if( shred_type==FD_SHRED_TYPE_MERKLE_DATA,
+                                  shred->idx - shred->fec_set_idx,
+                                  shred->code.data_cnt + code_idx );
+  /* TODO: bounds check shred_pos. make sure proof depth has the right
```
mmcgee-jump, line 280: Bounds checks again... this is unverified input right? We need to make this bulletproof.

mmcgee-jump (changes requested):
comments inline

src/disco/shred/fd_fec_resolver.c
ptaffet-jump: It was to prevent a variable read but never written warning. It's gone now with the switch from tcache to map.

src/disco/shred/fd_fec_resolver.c
ptaffet-jump: This is also gone now

src/disco/shred/fd_fec_resolver.c
ptaffet-jump: The signature is 512b (64B), but I switched to using the full thing.

src/disco/shred/fd_fec_resolver.c
ptaffet-jump: I added a check to immediately reject a 0 signature.

src/disco/shred/fd_fec_resolver.c
ptaffet-jump: Fixed

src/disco/shred/fd_fec_resolver.c
ptaffet-jump: Thanks, I rejiggered the indices and added the appropriate checks.

src/disco/shred/fd_fec_resolver.h
ptaffet-jump: Is this more clear?

mmcgee-jump (approved):

PR #747, opened 2023-10-03
fd_deque_dynamic: fix iteration on full deque

ptaffet-jump:
Kevin and I discussed some issues with fd_deque_dynamic and he suggested some changes. Previously iterating over a full deque would return 0 elements.

### DIFF
--- a/src/util/tmpl/fd_deque_dynamic.c
+++ b/src/util/tmpl/fd_deque_dynamic.c
@@ -352,36 +352,41 @@ DEQUE_(remove_all)( DEQUE_T * deque ) {
   return deque;
 }
 
-typedef ulong DEQUE_(iter_t);
+typedef struct { ulong rem; ulong idx; } DEQUE_(iter_t);
 
 static inline DEQUE_(iter_t)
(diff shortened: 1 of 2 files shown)


kbowers-jump (approved):

kbowers-jump:
The underlying issue here was the iterator assumed head==tail implies full.

This is true for the compile time sized deque.  Notwithstanding some of the nuances discussed in the code around doing more than 2^63 operations on a single deque (not going to happen practically) when max is not a power of 2, this has the invariant (tail-head)==cnt.

This is not true for the run time deque.  Because of the difference in implementation required to eliminate slow run time mod calculations that can be optimized at compile time in the compile time sized implementation, it's equivalent invariant is ((tail-head) mod max)==(cnt mod max).  As such, head==tail indicates the queue is either full or empty.  This is why the run time sized deque needs to track cnt explicitly.

(An alternative would be to have the run time deque use max+1 slots under the hood.  That's less memory efficient and tweaking the iterator implementation as done here was less invasive to the code.)

More minor nitpick:  There's probably also a case for a reverse direction iterator for completeness given this is a double ended (and similarly for the compile time sized deque).  No known pressing need though.

PR #757, opened 2023-10-09
fd_tar

ripatel-fd:
Adds tar decompression code from runtime branch

### DIFF
--- a/src/util/archive/fuzz_tar.c
+++ b/src/util/archive/fuzz_tar.c
@@ -1,0 +1,89 @@
+#if !FD_HAS_HOSTED
+#error "This target requires FD_HAS_HOSTED"
+#endif
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "../../util/fd_util.h"
+#include "fd_tar.h"
+
+int
+LLVMFuzzerInitialize( int  *   argc,
+                      char *** argv ) {
+  /* Set up shell without signal handlers */
+  putenv( "FD_LOG_BACKTRACE=0" );
+  fd_boot( argc, argv );
+  atexit( fd_halt );
+  fd_log_level_stderr_set(4);
+  return 0;
+}
+
+static int
+tar_file( void *                cb_arg,
+          fd_tar_meta_t const * meta,
+          ulong FD_FN_UNUSED    sz ) {
+  FD_TEST( (ulong)cb_arg == 0x1234UL );
+
+  /* Read meta to ensure it is accessible */
+  fd_tar_meta_t meta2;
+  meta2 = *meta;
+  void * meta_ptr = &meta2;
+  FD_COMPILER_FORGET( meta_ptr );
+
+  return 0;
+}
+
+static int
+tar_read( void *       cb_arg,
+          void const * buf,
+          ulong        bufsz ) {
+  FD_TEST( (ulong)cb_arg == 0x1234UL );
+
+  /* Read buf to ensure it is accessible */
+  int x = 0;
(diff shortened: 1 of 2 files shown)


src/util/archive/fd_tar.c
```diff
@@ -0,0 +1,198 @@
...
+
+int
+fd_tar_read_fd( fd_tar_reader_t * reader,
+                int               fd ) {
+  for(;;) {
+    uchar buf[ 16384 ];
+    int   terr = 0;
+    ulong rsz; int rerr = fd_io_read( fd, buf, 0UL, sizeof(buf), &rsz );
+    if     ( rerr==0      ) terr = fd_tar_read( reader, buf, rsz );
+    else if( rerr==EAGAIN ) sched_yield();
```
mmcgee-jump, line 191: Do you need EINTR here as well?

src/util/archive/fd_tar.c
```diff
@@ -0,0 +1,198 @@
...
+    buf[ i ] = '0' + (char)( val&7UL );
+    val>>=3;
+  }
+  return val==0UL;
+}
+
+#include <sched.h>
+
+int
+fd_tar_read_fd( fd_tar_reader_t * reader,
```
mmcgee-jump, line 184: I kind of hate this read_fd API. There's no file descriptor that you want to spin loop and read() from, even with the sched_yield, you're just going to lock the core up with read() -> EAGAIN a million times.  Has this API been useful? I feel like if the caller is not driving the read loop and you really want this function, it should just issue blocking reads on FD and just park the OS thread.

src/util/archive/fd_tar.c
```diff
@@ -0,0 +1,198 @@
...
+
+  /* Determine number of bytes to read */
+  long chunk_sz = (long)sizeof(fd_tar_meta_t) - (long)reader->buf_ctr;
+  FD_TEST( chunk_sz>=0L );
+  if( end-cur < chunk_sz ) chunk_sz = end-cur;
+
+  /* Copy to header */
+  fd_memcpy( reader->buf + reader->buf_ctr, cur, (ulong)chunk_sz );
+  cur             +=         chunk_sz;
+  reader->buf_ctr += (ushort)chunk_sz;
```
mmcgee-jump, line 95: FD_TEST this cast for paranoia (if someone changes fd_tar_meta_t in future or the compiler pads with junk)

src/util/archive/fd_tar.c
```diff
@@ -0,0 +1,198 @@
...
+fd_tar_reader_delete( fd_tar_reader_t * reader ) {
+  if( FD_UNLIKELY( !reader ) ) return NULL;
+  fd_memset( reader, 0, sizeof(fd_tar_reader_t) );
+  return reader;
+}
+
+static int
+fd_tar_process_hdr( fd_tar_reader_t * reader ) {
+
+  /* Detect EOF.  A TAR EOF is marked by 1024 bytes of zeros.
```
mmcgee-jump, line 41: Is this needed? Wouldn't it fail the hdr-magic check below?

src/util/archive/fuzz_tar.c
```diff
@@ -0,0 +1,89 @@
+#if !FD_HAS_HOSTED
```
mmcgee-jump, line 1: Looks like there's a good fuzz corpus from go-tar, https://github.com/strongcourage/fuzzing-corpus/tree/master/tar/go-fuzz

Can we use this?

mmcgee-jump (approved):

PR #759, opened 2023-10-09
chacha20: fix potential overflow

ptaffet-jump:
Kevin noticed that the addition to form `lh` could overflow and reminded me about `fd_uwide`. The existing tests weren't exhaustive enough to catch the mistake, so I added one that does catch it.


kbowers-jump (approved):

kbowers-jump:
I think technically given this implementation you could just uwide and it would "do the right thing" on platforms with 128-bit support (use the 128-bit integer support) and without (emulate it as needed).   I'll let you decide to merge or not.

PR #763, opened 2023-10-11
verify - added __NR_sched_yield to allow_syscalls

nbridge-jump:


mmcgee-jump:
Can you post the full callstack that tripped this? I went to some lengths to remove the sched_yield syscall a while back, it shouldn't be needed and we should try and fix the call site rather than adding it.

nbridge-jump:
sched_yield shouldn't be needed from verify

PR #771, opened 2023-10-13
Fix fd_io_{read,write} with zero sz

ripatel-fd:


kbowers-jump (approved):
As discussed with Richie internally and might be more useful to have recorded:

This is software engineering equivalent to the the math question:"what's 0/0?".

Under the grade school rule of thumb "0/x simplifies to 0", it seems like it should be 0.  And, under the grade school rule of thumb "x/x simplifies to 1", it seems like it should be 1.  If viewed it as a limiting process in calculus, you can find a justification to make 0/0 come out however you want ... the two grade school rules of thumbs are just commonly used limiting processes.

Which leads the real world handling of 0/0.  In grade school, when the precocious math student asks about it, the teacher explains "shut up".  And, in computer science, the computer explains "shut up" either via the near universally misunderstood and mishandled "NaN" (floating point 0./0.) or divide-by-zero signal (integer 0/0).

The debate for this API design is similar.  Developers expect `write( fd=garbage, ... )` should always fail.  Developers expect `write( ..., sz=0 )` should always work.  But what should `write( fd=garbage, ..., sz=0 )` do then?

The original implementation (in the interest of speed because this is usually wrapped up by APIs that ensure sz non-zero anyway) forbids talking about it by making positive size a prereq ("shut up" it explains).

Completely fine with changing it to use a reasonable limiting case behavior here (which this does), documenting the choice (which this does) and making sure the same limiting behavior is applied to the read (which this does and thus makes my inner physicist happy with the preservation of time-reversal symmetry).

PR #773, opened 2023-10-13
Copied types, nanopb, and valloc from private

asiegel-jt:


ripatel-fd:
@asiegel-jt I've split off the nanopb and valloc changes into separate PRs: #775 #780

PR #780, opened 2023-10-15
Import nanopb 0.4.7

ripatel-fd:
The Firedancer runtime branch currently uses Nanopb for handling Protobuf messages.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -282,3 +282,29 @@ License-Text:
  v2.2 or v3.x or whatever), unless explicitly otherwise stated.
 
 			Linus Torvalds
+
+=======
+
+'src/flamenco/nanopb' is currently a modified copy of the Nanopb
+library at https://jpa.kapsi.fi/nanopb/ (imported ca 2023-Nov).
+
+Copyright (c) 2011 Petteri Aimonen <jpa at nanopb.mail.kapsi.fi>
+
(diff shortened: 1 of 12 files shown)


mmcgee-jump:
The NOTICE says

> 'src/flamenco/nanopb' is currently a modified copy of the Nanopb library at https://jpa.kapsi.fi/nanopb/ (imported ca 2023-Sep).

But I had a hard time figuring out what the changes were. Is it modified? Can we use a .patch file applied at build time for these modifications? It will make review easier and save some merge and other headaches later.

ripatel-fd:
> The NOTICE says
> 
> > 'src/flamenco/nanopb' is currently a modified copy of the Nanopb library at https://jpa.kapsi.fi/nanopb/ (imported ca 2023-Sep).
> 
> But I had a hard time figuring out what the changes were. Is it modified? Can we use a .patch file applied at build time for these modifications? It will make review easier and save some merge and other headaches later.

Good idea 👍

jumpsiegel:
https://github.com/firedancer-io/firedancer/issues/920

ripatel-fd:
~~@mmcgee-jump I'm going to restore this branch to its original state and move your changes to `mmcgee/nanopb`.~~

mmcgee-jump (approved):

PR #781, opened 2023-10-15
Add SipHash13

ripatel-fd:
In Solana's spirit of "never met a hash function I didn't like", adds the reference implementation of SipHash13.

### DIFF
--- a/src/ballet/siphash13/fd_siphash13.c
+++ b/src/ballet/siphash13/fd_siphash13.c
@@ -1,24 +1,149 @@
 #include "fd_siphash13.h"
 
-#define FD_SIPHASH_ROUND(a,b,c,d)        \
-  do {                                   \
-    (a) += (b);                          \
-    (b) = fd_ulong_rotate_left((b), 13); \
-    (b) ^= (a);                          \
-    (a) = fd_ulong_rotate_left((a), 32); \
-    (c) += (d);                          \
-    (d) = fd_ulong_rotate_left((d), 16); \
-    (d) ^= (c);                          \
-    (a) += (d);                          \
-    (d) = fd_ulong_rotate_left((d), 21); \
-    (d) ^= (a);                          \
-    (c) += (b);                          \
-    (b) = fd_ulong_rotate_left((b), 17); \
-    (b) ^= (c);                          \
-    (c) = fd_ulong_rotate_left((c), 32); \
-  } while (0)
+/* This code is a modified version of https://github.com/antirez/siphash
+   For further license info see NOTICE in the root of this repo.
+
+   Copyright (c) 2012-2016 Jean-Philippe Aumasson
+   <jeanphilippe.aumasson@gmail.com>
+   Copyright (c) 2012-2014 Daniel J. Bernstein <djb@cr.yp.to>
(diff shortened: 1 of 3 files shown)


src/ballet/siphash13/fd_siphash13.h
```diff
@@ -0,0 +1,83 @@
+#ifndef HEADER_fd_src_ballet_siphash13_fd_siphash13_h
```
mmcgee-jump, line 1: It's a nit but can you put an origination notice in the header / impl as well? Someone may not know to look in NOTICE before making changes.

src/ballet/siphash13/fd_siphash13.h
```diff
@@ -0,0 +1,83 @@
...
+#define HEADER_fd_src_ballet_siphash13_fd_siphash13_h
+
+/* fd_siphash13 provides APIs for SipHash1-3.
+   (1 compression round, 3 finalization rounds) */
+
+#include "../fd_ballet_base.h"
+
+#define FD_SIPHASH13_ALIGN     (128UL)
```
mmcgee-jump, line 9: Curious about the ALIGN and FOOTPRINT here. I know we use 128 for concurrent data to prevent false sharing, but this looks very single threaded. Is there a perf/other benefit to aligning 128 rather than `alignof( fd_siphash13_private )` ?

src/ballet/siphash13/fd_siphash13.c
```diff
@@ -0,0 +1,191 @@
...
+fd_siphash13_t *
+fd_siphash13_init( fd_siphash13_t * sip,
+                   ulong            k0,
+                   ulong            k1 ) {
+
+  memset( sip, 0, sizeof(fd_siphash13_t) );
+
+  ulong * v = sip->v;
+
+  memcpy( v, fd_siphash13_initial, 32UL );
```
mmcgee-jump, line 20: nit but you use `fd_memcpy` in other parts of this file

mmcgee-jump: (Also imo this part might be clearer without the memcpy)

src/ballet/siphash13/fd_siphash13.c
```diff
@@ -0,0 +1,191 @@
...
+
+  ulong * v        = sip->v;
+  uchar * buf      = sip->buf;
+  ulong   buf_used = sip->n & 7UL;
+
+  sip->n += sz;
+
+  if( FD_UNLIKELY( buf_used ) ) { /* optimized for well aligned use of append */
+
+    /* If the append isn't large enough to complete the current blcok,
```
mmcgee-jump, line 55:
```suggestion
    /* If the append isn't large enough to complete the current block,
```

src/ballet/siphash13/fd_siphash13.c
```diff
@@ -0,0 +1,191 @@
...
+      buffer these bytes too and return */
```
mmcgee-jump, line 56:
```suggestion
       buffer these bytes too and return */
```

mmcgee-jump (approved):

mmcgee-jump:
Nice! 👍 LGTM

jumpsiegel (approved):

PR #788, opened 2023-10-16
Fuzzers for base58/base64/hex encoders/decoders in ballet

theo25:
This PR adds fuzzers for the encoders/decoders found in `ballet`:
- `base58`
- `base64`
- `hex`

### DIFF
--- a/src/ballet/base64/fuzz_base64.c
+++ b/src/ballet/base64/fuzz_base64.c
@@ -1,0 +1,49 @@
+#if !FD_HAS_HOSTED
+#error "This target requires FD_HAS_HOSTED"
+#endif
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "../../util/fd_util.h"
+#include "fd_base64.h"
+
+int
+LLVMFuzzerInitialize( int  *   argc,
+                      char *** argv ) {
+  /* Set up shell without signal handlers */
+  putenv( "FD_LOG_BACKTRACE=0" );
+  fd_boot( argc, argv );
+  atexit( fd_halt );
+  return 0;
+}
+
+#define get_encoded_len(bytes) ( ( ( ( bytes*4UL + 2UL) / 3UL ) + 3UL ) & ~3UL )
+#define MAX_DATA_SZ 4096UL
+#define MAX_ENCODED_SZ ( get_encoded_len( MAX_DATA_SZ )+1UL )
+
+int
+LLVMFuzzerTestOneInput( uchar const * data,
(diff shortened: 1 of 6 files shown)


src/ballet/hex/fuzz_hex.c
```diff
@@ -0,0 +1,55 @@
...
+                      char *** argv ) {
+  /* Set up shell without signal handlers */
+  putenv( "FD_LOG_BACKTRACE=0" );
+  fd_boot( argc, argv );
+  atexit( fd_halt );
+  return 0;
+}
+
+// checks if given encoding of sz chars is valid, sz must be even
+// returns sz on success, index of first invalid char on failure
```
ptaffet-jump, line 22:
```suggestion
/* checks if given encoding of sz chars is valid, sz must be even
   returns sz on success, index of first invalid char on failure */
```

src/ballet/hex/fuzz_hex.c
```diff
@@ -0,0 +1,55 @@
...
+#define MAX_DATA_SZ 4096UL
+#define MAX_DECODED_SZ ( MAX_DATA_SZ / 2UL )
+
+int
+LLVMFuzzerTestOneInput( uchar const * data,
+                        ulong         size ) {
+  if( FD_UNLIKELY( size > MAX_DATA_SZ ) ) return -1;
+  char const * encoded = ( char  const * ) data;
+
+  size = size & ~1UL; // ignore last char of encoding if size is odd
```
ptaffet-jump, line 45:
```suggestion
  size = size & ~1UL; /* ignore last char of encoding if size is odd */
```

ptaffet-jump (approved):
We normally use `/* */` comments in FD. Otherwise, this looks good to me.

src/ballet/base64/fuzz_base64.c
```diff
@@ -0,0 +1,49 @@
...
+  atexit( fd_halt );
+  return 0;
+}
+
+#define get_encoded_len(bytes) ( ( ( ( bytes*4UL + 2UL) / 3UL ) + 3UL ) & ~3UL )
+#define MAX_DATA_SZ 4096UL
+#define MAX_ENCODED_SZ ( get_encoded_len( MAX_DATA_SZ )+1UL )
+
+int
+LLVMFuzzerTestOneInput( uchar const * data,
```
marcus-jump, line 26: I feel like this fuzz target should be called fuzz_base64_roundtrip and that there should be another one attempting to decode trash.

PR #789, opened 2023-10-16
fdctl: tidy up capability checks

### LINKED ISSUE #661: Print missing capability names rather than numbers
When running `fdctl` without proper permissions an error message will be reported like `security: capability 13 is required to call `bind(2)` to bind a socket with SOCK_RAW`. This is confusing for an operator and we should just print the actual capability string, here `CAP_NET_RAW`.

mmcgee-jump:
This isn't really security related so renamed it to better reflect what it's doing.  Also the error messages are now pretty printing missing capability names for easier operator experience.

Fixes #661


ptaffet-jump (approved):
I didn't execute it or anything, but it seems good to me

src/app/fdctl/main1.c
```diff
@@ -39,13 +39,32 @@ main1( int     argc,
...
+      } else if( FD_LIKELY( !strcmp( action->name, "configure" ) ) ) {
+        FD_LOG_ERR(( "insufficient permissions to execute command `%s`. It is recommended "
+                     "to configure Firedancer as the root user. Firedancer configuration requires "
+                     "root because it does privileged operating system actions like setting up XDP. "
+                     "Configuration is a local action that does not access the network, and the process "
+                     "exits immedaitely once configuration completes. The user that Firedancer runs "
+                     "as is specified in your configuration file, and although configuration runs as root "
+                     "it will permission the relevant resources for the user in your configuration file, "
+                     "which can be an anonymous maximally restrictive account with no privileges." ));
+      } else {
```
ptaffet-jump, line 65:
```suggestion
                     "which can be an anonymous maximally restrictive account with no privileges.", action->name ));
      } else {
```

CI was right

ptaffet-jump (changes requested):

ptaffet-jump (approved):

PR #792, opened 2023-10-16
xdp: correctly notify driver when needed

### LINKED ISSUE #674: Investigate and hopefully remove `sendto` syscall from QUIC tiles
When binding the XSK to queue on network interface, we set

```C
sxdp_flags    = XDP_USE_NEED_WAKEUP | (ushort)xsk->params.zerocopy
```

`XDP_USE_NEED_WAKEUP` triggers a mode where we need to call `sendto(2)` to have the kernel / driver service the queues. It does not feel like this should be necessary. The syscall is a performance and security hit, and the non `NEED_WAKEUP` option should have the driver take a core and spin, constantly servicing the queue without either of these issues.

mmcgee-jump:
Fixed #674

### DIFF
--- a/src/tango/xdp/fd_xsk.c
+++ b/src/tango/xdp/fd_xsk.c
@@ -170,7 +170,7 @@ fd_xsk_unbind( void * shxsk ) {
   /* Reset */
 
   fd_memset( xsk->if_name_cstr, 0, IF_NAMESIZE );
-  xsk->if_queue_id = 0U; /* TODO should reset to -1? */
+  xsk->if_queue_id = UINT_MAX;
 
   return shxsk;
 }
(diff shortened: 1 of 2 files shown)


src/tango/xdp/fd_xsk.c
```diff
@@ -730,12 +789,7 @@ fd_xsk_tx_enqueue( fd_xsk_t *            xsk,
 
   /* sz is min( available, count ) */
   uint sz = cap - ( prod - cons );
-  /* TODO this doesn't work as expected
-     if we early exit here, no wakeup occurs, sendto doesn't get called again
-     and the ring doesn't get serviced
-     This implies we need to call sendto AGAIN even if the ring hasn't changed
   if( sz == 0 )    return 0;
```
nbridge-jump, line 792: This needs to be removed. If a prior call did not have "flush" set, the ring may be full and yet the driver may not know it. Calling with "flush" should invoke wakeup regardless.

nbridge-jump (changes requested):
Just one comment

nbridge-jump (approved):

PR #802, opened 2023-10-18
change toml parsing to fail on unrecognized keys

llamb-jump:
Testing

After adding an invalid key to `default.toml` we get the following output
```
llamb@emny-ossdev-firedancer11> fddev
--log-path not specified; using autogenerated path
Log at "/tmp/fd-0.0.0_577344_llamb_emny-ossdev-firedancer11_2023_10_18_04_22_27_927939606_GMT+00"
ERR     10-18 04:22:27.928847 577344 f0   main src/app/fdctl/config.c(416): Error while parsing the embedded configuration. The configuration had an unrecognized key [tiles.quic.wat].
```

Doing the same test but after copying `default.toml` to `/tmp`
```
llamb@emny-ossdev-firedancer11> fddev --config /tmp/default.toml
--log-path not specified; using autogenerated path
Log at "/tmp/fd-0.0.0_578027_llamb_emny-ossdev-firedancer11_2023_10_18_04_23_58_023990675_GMT+00"
ERR     10-18 04:23:58.024996 578027 f0   main src/app/fdctl/config.c(418): Error while parsing user configuration TOML file at /tmp/default.toml. The configuration had an unrecognized key [tiles.quic.wat].
```

### DIFF
--- a/src/app/fdctl/config.c
+++ b/src/app/fdctl/config.c
@@ -112,22 +112,22 @@ default_user( void ) {
   return name;
 }
 
-static void parse_key_value( config_t *   config,
-                             const char * section,
-                             const char * key,
-                             char * value ) {
+static int parse_key_value( config_t *   config,
+                            const char * section,
+                            const char * key,
+                            char * value ) {
 #define ENTRY_STR(edot, esection, ekey) do {                                         \
     if( FD_UNLIKELY( !strcmp( section, #esection ) && !strcmp( key, #ekey ) ) ) {    \
       ulong len = strlen( value );                                                   \
       if( FD_UNLIKELY( len < 2 || value[ 0 ] != '"' || value[ len - 1 ] != '"' ) ) { \
         FD_LOG_ERR(( "invalid value for %s.%s: `%s`", section, key, value ));        \
-        return;                                                                      \
+        return 1;                                                                    \
       }                                                                              \
       if( FD_UNLIKELY( len >= sizeof( config->esection edot ekey ) + 2 ) )           \
         FD_LOG_ERR(( "value for %s.%s is too long: `%s`", section, key, value ));    \
       strncpy( config->esection edot ekey, value + 1, len - 2 );                     \
       config->esection edot ekey[ len - 2 ] = '\0';                                  \
-      return;                                                                        \
+      return 1;                                                                      \
     }                                                                                \
   } while( 0 )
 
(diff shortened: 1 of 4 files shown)


src/app/fdctl/config/default.toml
```diff
@@ -512,7 +501,7 @@ dynamic_port_range = "8900-9000"
     [tiles.quic]
         # The period in which there are no active requests.
```
mmcgee-jump, line 502: Another nit be can we reword this,

```# The duration after which an idle connection will be timed out, and the connection will be closed.  An idle connection is one where no data is received from the peer.```

Also can you wrap to 72 columns. There's a VS code extension that can help with this.

File: src/app/fdctl/config/default.toml [hunk 2]
```diff
@@ -429,17 +429,6 @@ dynamic_port_range = "8900-9000"
     # already exists it should be writable by the Firedancer user.
     huge_page_mount_path = "/mnt/.fd/.huge"
 
-    # Minimum number of gigantic pages the kernel should pre-allocate
```
mmcgee-jump, line 432: Nice!

src/app/fdctl/configure/workspace.c
```diff
@@ -329,7 +329,7 @@ init( config_t * const config ) {
         buf    ( pod, "src_mac_addr",                 config->tiles.net.mac_addr, 6 );
         ushort1( pod, "transaction_listen_port",      config->tiles.quic.regular_transaction_listen_port, 0 );
         ushort1( pod, "quic_transaction_listen_port", config->tiles.quic.quic_transaction_listen_port, 0 );
-        ulong1 ( pod, "idle_timeout_ms",              1000 );
+        ulong1 ( pod, "idle_timeout_ms",              config->tiles.quic.idle_timeout_millis, 10000 );
```
mmcgee-jump, line 332: I think theres just one arg here, the 10000 is not needed and is probably ignored

src/app/fdctl/config.c
```diff
@@ -354,7 +358,11 @@ config_parse_array( config_t * config,
     char * end = token + strlen( token ) - 1;
     while( FD_UNLIKELY( *end == ' ' ) ) end--;
     *(end+1) = '\0';
-    if( FD_LIKELY( end > token ) ) parse_key_value( config, section, key, token );
+    if( FD_LIKELY( end > token ) ) {
+      if( FD_UNLIKELY( !parse_key_value( config, section, key, token ) ) ) {
+        FD_LOG_ERR(( "unrecognized key %s", key ));
```
mmcgee-jump, line 363: Another nit but since this is a user configuration file, it would be good to have a long descriptive error message for this failure case. Eg,

"Error while parsing user configuration TOML file at `%s`. The configuration had an unrecognized key `[%s.%s]`.", file_path, section, key

(I know we don't do this for some other errors currently)

llamb-jump: Updated and added some test output to the description

mmcgee-jump (dismissed):

src/app/fdctl/config/default.toml
```diff
@@ -510,9 +499,10 @@ dynamic_port_range = "8900-9000"
     # forward well formed (but not necessarily valid) ones to verify
     # tiles.
     [tiles.quic]
-        # The period in which there are no active requests.
-        # When the idle timeout is reached the connection will be closed.
-        idle_timeout_ms = 10000
+        # The duration after which an idle connection will be timed out,
+        # and the connection will be closed. An idle connection is one
```
mmcgee-jump, line 503:
```suggestion
        # and the connection will be closed.  An idle connection is one
```

mmcgee-jump (approved):

PR #806, opened 2023-10-18
Fuzzer for the Reed-Solomon implementation in ballet

theo25:
This PR adds a fuzzer for the Reed-Solomon implementation found in `ballet`.


ptaffet-jump:
I think we want to remove some of the loops over all values and replace it with a single value derived from the fuzzer input. Basically, I think this version does too much work per fuzz input, and I think it'll be more interesting to use the fuzz input to control things like number of shreds, shred size, which shreds get erased/corrupted.

theo25:
I have changed the implementation according to @ptaffet-jump's feedback. Now the fuzzer uses its input for specifying shred size, number of data/parity/erased threads, as well as for which shred to corrupt.

marcus-ares (approved):

PR #813, opened 2023-10-19
Improve build security flags

marcus-jump:
* make executables position independent
* use fortify source
* build with immediate bindings if possible
* enable stack canaries (previously, only the gcc builds would end up having stack canaries)

### DIFF
--- a/config/with-security.mk
+++ b/config/with-security.mk
@@ -1,0 +1,2 @@
+CPPFLAGS+=-D_FORTIFY_SOURCE=2 -fpie -pie -fPIC -Wl,-z,now -fstack-protector-strong
+LDFLAGS+=-fpie -pie -fPIC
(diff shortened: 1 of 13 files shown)


mmcgee-jump:
Can you move the clang change out of this into another PR? The others I can approve but I think we want a longer discussion about that.

config/with-security.mk
```diff
@@ -0,0 +1,2 @@
+CPPFLAGS+=-D_FORTIFY_SOURCE=2 -fpie -pie -fPIC -Wl,-z,now -fstack-protector-strong
+LDFLOGS+=-fpie -pie -fPIC
```
mmcgee-jump, line 2:
```suggestion
LDFLOGS+=-fpie -pie -fPIC

```

src/flamenco/vm/Local.mk
```diff
@@ -6,4 +6,6 @@ $(call make-bin,fd_vm_tool,fd_vm_tool,fd_flamenco fd_ballet fd_util)
 endif
 
 $(call make-unit-test,test_vm_interp,test_vm_interp,fd_flamenco fd_ballet fd_util)
-$(call run-unit-test,test_vm_interp)
+
+# Todo(lheeger-jump): enable this
```
mmcgee-jump, line 10: What's the issue with this?

marcus-jump: This test fails but Liam is going to fix that in the runtime branch.

mmcgee-jump (approved):

PR #815, opened 2023-10-19
fix some port traffic related issues

anwayde:
Range end was not getting set correctly resulting in:
```
ERR     10-19 20:15:38.662281 16     f0   shred src/disco/shred/fd_shred_tile.c(331): chunk 823780 1245 corrupt, not in range [16370,16370]
```

If some port, for e.g. shred port is set to 8003 and it is not in the range of handled_port, we were erroring out:
```
ERR     10-19 20:15:38.656524 6      f0   quic src/disco/quic/fd_quic.c(313): Firedancer received a UDP packet on port 8003 which was not expected. Only ports 9007 and 9001 should be configured to forward packets. Do you need to reload the XDP program?
```

### DIFF
--- a/src/disco/quic/fd_quic.c
+++ b/src/disco/quic/fd_quic.c
@@ -306,11 +306,10 @@ before_frag( void * _ctx,
   int handled_port = dst_port == ctx->legacy_transaction_port ||
                      dst_port == ctx->quic->config.net.listen_udp_port;
 
+  /* Ignore traffic e.g. for shred tile */
   if( FD_UNLIKELY( !handled_port ) ) {
-    FD_LOG_ERR(( "Firedancer received a UDP packet on port %hu which was not expected. "
-                 "Only ports %hu and %hu should be configured to forward packets. Do "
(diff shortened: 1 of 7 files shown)


src/disco/quic/fd_quic.c
```diff
@@ -306,11 +306,10 @@ before_frag( void * _ctx,
   int handled_port = dst_port == ctx->legacy_transaction_port ||
                      dst_port == ctx->quic->config.net.listen_udp_port;
 
+  /* Ignore traffic e.g. for shred tile */
   if( FD_UNLIKELY( !handled_port ) ) {
-    FD_LOG_ERR(( "Firedancer received a UDP packet on port %hu which was not expected. "
```
mmcgee-jump, line 310: We want to keep this check at the moment to verify the XDP traffic is right. Can you plumb shred port through and add it to the check here

ptaffet-jump (approved):

mmcgee-jump (approved):

PR #819, opened 2023-10-20
clang as golden compiler

marcus-ares:
We've talked about this change earlier this year. Here it is!


marcus-ares:
Motivation for this change:

* It's easier to focus security testing around one compiler target. We already use libfuzzer which is part of LLVM.
* [@ripatel-fd] Using Clang in 1.5 reduces ABI compatibility problems

PR #829, opened 2023-10-22
fdctl: map some workspaces as readonly if possible

mmcgee-jump:
We want to map shared memory as PROT_READ only for added security if a tile does not write to the associated workspace.  We automatically determine if this is possible from the topology.


ptaffet-jump (approved):
Wait for CI, but looks fine to me

PR #835, opened 2023-10-24
security: restrict syscall arguments

### LINKED ISSUE #656: Add seccomp restriction on syscall argument values
Currently we use seccomp to create an allow list of syscalls, but then all arguments to those syscalls are permitted.

We should add functionality to filter syscall arguments as well (eg, the write syscall limited only to descriptor 3, stderr), and implement that across all allowed syscalls.

mmcgee-jump:
fd_sandbox is switched over to taking a list of sock_filter rather
than a list of sycall NRs, and then the generated filter programs
and linked to individual tiles and loaded at init time.

Fixes #656


marcus-ares (approved):

src/tango/quic/fd_quic.h
```diff
@@ -478,7 +476,8 @@ fd_quic_footprint( fd_quic_limits_t const * limits );
 
 FD_QUIC_API void *
```
ptaffet-jump, line 477: Can you add a comment here about the ip ownership changing. Something like: `fd_quic_new` retains a read/write interest in `ip` that persists until the memory is unformatted.

src/tango/quic/tests/arp/test_quic_arp_server.c
```diff
@@ -128,8 +128,6 @@ main( int argc, char ** argv ) {
     .stream_cnt       = { 2, 2, 2, 2 },
     .inflight_pkt_cnt = 1024,
     .tx_buf_sz        = 1<<14,
-    .routing_entries  = 32,
-    .arp_entries      = 32
```
ptaffet-jump, line 132: Is there a reason to get rid of these and only use hardcoded 256?

src/util/sandbox/fd_sandbox.c
```diff
@@ -128,41 +128,12 @@ check_fds( ulong allow_fds_cnt,
...
-  /* none of the syscalls approved were matched: die */
-  struct sock_filter kill = BPF_STMT(BPF_RET | BPF_K, SECCOMP_RET_KILL_PROCESS);
-  filter[ 4+(allow_syscalls_cnt*2) ] = kill;
-
-  FD_TESTV( 0 == prctl( PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0 ) );
-
-  struct sock_fprog default_prog = {
-    .len = (ushort)(5+(allow_syscalls_cnt*2)),
-    .filter = filter,
+install_seccomp( ushort allow_syscalls_cnt, struct sock_filter * allow_syscalls ) {
```
ptaffet-jump, line 131: Do you want to change the names of these arguments?

src/util/sandbox/fd_sandbox.h
```diff
@@ -67,14 +70,13 @@
 /* fd_sandbox sandboxes the current process, performing both privileged and
    private steps. */
 void
-fd_sandbox( int    full_sandbox,
```
ptaffet-jump, line 70: Could you update the comment for this function? It has gotten a little stale.

src/util/sandbox/fd_sandbox.h
```diff
@@ -1,10 +1,13 @@
 #ifndef HEADER_fd_src_util_sandbox_fd_sandbox_h
 #define HEADER_fd_src_util_sandbox_fd_sandbox_h
 
+#ifndef FD_HAS_FFI
```
ptaffet-jump, line 4: Does this also need `HOSTED`?

ptaffet-jump (approved):
I did not review any of the seccomppolicies themselves, the header files they generate, or the filter compiler. I reviewed the rest, and it looks good. Only minor comments.

src/tango/quic/tests/arp/test_quic_arp_server.c
mmcgee-jump: The caller configures them now, not QUIC. This caller could set 32 if it wanted.

ptaffet-jump (approved):

PR #839, opened 2023-10-24
redo plumbing of stake weights, contact info into pack, shred tiles

ptaffet-jump:
Rather than sending the full leader schedule, we now send the stake weights. These are usable by the pack tile and the shred tile. Also changed the splice point for contact info.

Around epoch boundaries, this can get a little tricky (especially in the shred tile case), so we keep around an even epoch and an odd epoch, which is normally the current epoch and the next epoch. We get the next epoch as soon as the current one is rooted.

### DIFF
--- a/src/app/fdctl/topology.h
+++ b/src/app/fdctl/topology.h
@@ -34,15 +34,16 @@
 #define FD_TOPO_WKSP_KIND_PACK_BANK    ( 4UL)
 #define FD_TOPO_WKSP_KIND_BANK_SHRED   ( 5UL)
 #define FD_TOPO_WKSP_KIND_SHRED_STORE  ( 6UL)
-#define FD_TOPO_WKSP_KIND_NET          ( 7UL)
-#define FD_TOPO_WKSP_KIND_NETMUX       ( 8UL)
-#define FD_TOPO_WKSP_KIND_QUIC         ( 9UL)
-#define FD_TOPO_WKSP_KIND_VERIFY       (10UL)
-#define FD_TOPO_WKSP_KIND_DEDUP        (11UL)
-#define FD_TOPO_WKSP_KIND_PACK         (12UL)
-#define FD_TOPO_WKSP_KIND_BANK         (13UL)
-#define FD_TOPO_WKSP_KIND_SHRED        (14UL)
-#define FD_TOPO_WKSP_KIND_STORE        (15UL)
+#define FD_TOPO_WKSP_KIND_STAKE_OUT    ( 7UL)
+#define FD_TOPO_WKSP_KIND_NET          ( 8UL)
+#define FD_TOPO_WKSP_KIND_NETMUX       ( 9UL)
+#define FD_TOPO_WKSP_KIND_QUIC         (10UL)
+#define FD_TOPO_WKSP_KIND_VERIFY       (11UL)
+#define FD_TOPO_WKSP_KIND_DEDUP        (12UL)
+#define FD_TOPO_WKSP_KIND_PACK         (13UL)
+#define FD_TOPO_WKSP_KIND_BANK         (14UL)
+#define FD_TOPO_WKSP_KIND_SHRED        (15UL)
+#define FD_TOPO_WKSP_KIND_STORE        (16UL)
 #define FD_TOPO_WKSP_KIND_MAX          ( FD_TOPO_WKSP_KIND_STORE+1 ) /* Keep updated with maximum tile IDX */
 
 /* FD_TOPO_LINK_KIND_* is an identifier for a particular kind of link. A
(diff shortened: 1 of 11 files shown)


src/app/fdctl/topology.h
```diff
@@ -34,15 +34,16 @@
...
-#define FD_TOPO_WKSP_KIND_NET          ( 7UL)
-#define FD_TOPO_WKSP_KIND_NETMUX       ( 8UL)
-#define FD_TOPO_WKSP_KIND_QUIC         ( 9UL)
-#define FD_TOPO_WKSP_KIND_VERIFY       (10UL)
-#define FD_TOPO_WKSP_KIND_DEDUP        (11UL)
-#define FD_TOPO_WKSP_KIND_PACK         (12UL)
-#define FD_TOPO_WKSP_KIND_BANK         (13UL)
-#define FD_TOPO_WKSP_KIND_SHRED        (14UL)
-#define FD_TOPO_WKSP_KIND_STORE        (15UL)
+#define FD_TOPO_WKSP_KIND_STAKES_FD    ( 7UL)
```
mmcgee-jump, line 37: Rename `FD_TOPO_WKSP_KIND_STAKE_OUT`

src/app/fdctl/topology.h
```diff
@@ -350,11 +353,12 @@ fd_topo_link_kind_str( ulong kind ) {
     case FD_TOPO_LINK_KIND_VERIFY_TO_DEDUP: return "verify_dedup";
     case FD_TOPO_LINK_KIND_DEDUP_TO_PACK:   return "dedup_pack";
     case FD_TOPO_LINK_KIND_GOSSIP_TO_PACK:  return "gossip_pack";
-    case FD_TOPO_LINK_KIND_LSCHED_TO_PACK:  return "lsched_pack";
+    case FD_TOPO_LINK_KIND_STAKES_TO_FD:    return "stakes_fd";
```
mmcgee-jump, line 356: STAKES_TO_OUT, same as NETMUX to indicate multiple consumers

src/app/fdctl/topology.h
```diff
@@ -350,11 +353,12 @@ fd_topo_link_kind_str( ulong kind ) {
...
     case FD_TOPO_LINK_KIND_PACK_TO_BANK:    return "pack_bank";
     case FD_TOPO_LINK_KIND_POH_TO_SHRED:    return "poh_shred";
     case FD_TOPO_LINK_KIND_SHRED_TO_NETMUX: return "shred_netmux";
     case FD_TOPO_LINK_KIND_SHRED_TO_STORE:  return "shred_store";
+    case FD_TOPO_LINK_KIND_CONTACT_TO_SHRED:return "contact_shred";
```
mmcgee-jump, line 361: I don't like that this is 7 chars for "contact", I think monitor caps out at 6 and it's nice to keep short for display. Maybe name "cntact_shred", or rename the whole thing? GOSSIP vs. CONTACT is pretty vague here, since they are both from the same place?

src/disco/shred/fd_stake_ci.h
```diff
@@ -0,0 +1,80 @@
...
+
+#include "fd_shred_dest.h"
+#include "../../flamenco/leaders/fd_leaders.h"
+
+#define MAX_SHRED_DESTS 50000UL
+#define MAX_SLOTS_PER_EPOCH 432000UL
+#define MAX_SHRED_DEST_FOOTPRINT 11244032UL /* == fd_shred_dest_footprint( MAX_SHRED_DESTS ), runtime asserted in new */
+
+struct fd_per_epoch_info_private {
+  /* Epoch, and [start_slot, start_slot+slot_cnt) refer to the time
```
mmcgee-jump, line 17: This comment is pretty explicit but just because of the confusion this caused us I would add more detail.

I think specifically call out that you do not need to adjust epoch by +1 or +2, even though the leader schedule is based on stakes at a prior epoch.

src/disco/shred/fd_stake_ci.c
```diff
@@ -0,0 +1,258 @@
...
+  info->scratch->epoch      = epoch;
+  info->scratch->start_slot = start_slot;
+  info->scratch->slot_cnt   = slot_cnt;
+  info->scratch->staked_cnt = staked_cnt;
+
+  fd_memcpy( info->stake_weight, hdr+4UL, sizeof(fd_stake_weight_t)*staked_cnt );
+}
+
+static inline void
+log_summary( char const * msg, fd_stake_ci_t * info ) {
```
mmcgee-jump, line 65: Do we want this on NOTICE in prod?

src/disco/shred/fd_stake_ci.c
```diff
@@ -0,0 +1,258 @@
...
+fd_stake_ci_stake_msg_init( fd_stake_ci_t * info,
+                            uchar const   * new_message ) {
+  ulong const * hdr = fd_type_pun_const( new_message );
+
+  ulong epoch               = hdr[ 0 ];
+  ulong staked_cnt          = hdr[ 1 ];
+  ulong start_slot          = hdr[ 2 ];
+  ulong slot_cnt            = hdr[ 3 ];
+
+  /* TODO: double check dest_cnt */
```
mmcgee-jump, line 54: TODO

mmcgee-jump (approved):

src/disco/shred/fd_stake_ci.c
ptaffet-jump: I removed the one on lookup failure, which could be triggered by someone sending a malicious shred. I'll leave the other ones for now, since they happen only around epoch transitions or contact info changes (capped at 1x per 5 seconds). Solana Labs does something similar, and it has been useful for debugging.

mmcgee-jump (approved):

PR #847, opened 2023-10-26
fctl: give each tile unique adress space layout

mmcgee-jump:
The boot path is rewired to use pipes and SIGHUP for determining when a
process is dead.  This is needed to fix some rare cases where orphaned
processes would be leftover.

Then the boot sequence is switched so that tiles are started via. execve
rather than clone.  This boot sequence is now:

 1. The main process starts.
    - It creates a pipe, main_pipe, with the read side CLOEXEC.
    - It spawns a child process, pidns with clone(CLONE_NEWPID).  This
      process exists solely to be the root of a PID namespace. The read
      side of main_pipe is closed in pidns and it holds onto the write
      side.
    - The main process closes the write side of main_pipe.

Now main and pidns can tell if the other one died by polling for HUP on
the pipe fd.  Main doesn't need to do this since it can call waitpid()
on the pidns pid, but the pidns does need to poll, in-case main is
killed with a SIGKILL.

  2. The pidns process starts.
    - It spawns each tile one by one, by calling fork() and then
      execve(). Before spawning the tile it creates a pipe, tile_pipe,
      with the read side CLOEXEC.  It passes the write side to the child
      (tile) when it is execve'd.
    - Because memory maps are lost on an execve, but it needs to share
      the configuration with child tiles, it creates a memfd with
      memfd_create, writes the config there, passes the fd through the
      execve call, and has the tile map the memfd in and read it.
    - A similar memfd sequence is done with the lock object for
      sequencing writes the log file.
    - The pidns process closes the write side of tile_pipe.

The pidns process can now call poll( main_read, tile_pipe0_read,
tile_pipe1_read, ... ) to wait for any the parent or child processes to
die.  After that happens it will exit itself, and because it's the init
of a PID namespace, all of its children will then be terminated as well
by the kernel.

   3. Each tile process starts.
    - It immediately calls clone() to spawn a new process with a
      stack backed by huge pages.  The tile_pipe write side is
      passed onto the cloned child.
    - It returns so that the non-hugepage process exits.
    - The pidns process reaps the non-hugepage process.

Now we finally have all our tile processes ready, running on huge pages,
with a unique address space layout because they were restarted via.
execve.  The pidns and main processes are linked by a pipe, and the
pidns and child tiles are linked by a pipe as well, meaning that if any
process in the tree exits, everything else will get taken down cleanly.


src/app/fdctl/run/tiles/run.seccomppolicy
```diff
@@ -10,6 +10,3 @@ wait4: (and (eq (arg 0) pid_namespace)
 
 # exit process
 exit_group
-
-# kill the pid namespaced child process
-kill: (eq (arg 1) SIGKILL)
```
mmcgee-jump, line 15: Needs to come back

src/util/sandbox/fd_sandbox.c
```diff
@@ -251,8 +251,8 @@ sandbox_unthreaded( ulong allow_fds_cnt,
                     uint gid ) {
   check_fds( allow_fds_cnt, allow_fds );
   unshare_user( uid, gid );
-  struct rlimit limit = { .rlim_cur = 0, .rlim_max = 0 };
```
mmcgee-jump, line 254: Needs to be fixed. the `poll()` syscall requires rlimit = nfds

src/app/fdctl/run/run.c
```diff
@@ -398,17 +204,26 @@ main_pid_namespace( void * args ) {
   if( FD_UNLIKELY( sched_getaffinity( 0, sizeof(cpu_set_t), floating_cpu_set ) ) )
     FD_LOG_ERR(( "sched_getaffinity failed (%i-%s)", errno, fd_io_strerror( errno ) ));
 
-  pid_t child_pids[ FD_TILE_MAX + 1 ];
-  char  child_names[ FD_TILE_MAX + 1 ][ 32 ];
+  pid_t child_pids[ FD_TOPO_MAX_TILES+1 ];
+  char  child_names[ FD_TOPO_MAX_TILES+1 ][ 32 ];
+  struct pollfd fds[ FD_TOPO_MAX_TILES+2 ];
+
+  fds[ 0 ] = (struct pollfd){ .fd = args->pipefd[ 1 ], .events = 0 };
```
ptaffet-jump, line 211: You can move this element to the end of the array to eliminate all the `1+`s below

src/app/fdctl/run/run.c
```diff
@@ -60,145 +70,56 @@ install_tile_signals( void ) {
...
-              seccomp_filter_cnt,
-              seccomp_filter );
-
-  /* Now we are sandboxed, join all the tango IPC objects in the workspaces */
-  fd_topo_fill_tile( &args->config->topo, tile, FD_TOPO_FILL_MODE_JOIN );
-  FD_TEST( tile->cnc );
+  uchar * bytes = mmap( NULL, sizeof( config_t ), PROT_READ | PROT_WRITE, MAP_SHARED, config_memfd, 0 );
+  if( FD_UNLIKELY( bytes == MAP_FAILED ) ) FD_LOG_ERR(( "mmap() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+  fd_memcpy( bytes, config, sizeof( config_t ) );
+  if( FD_UNLIKELY( munmap( bytes, sizeof( config_t ) ) ) ) FD_LOG_ERR(( "munmap() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
```
ptaffet-jump, line 83: It's probably not necessary since the tiles close the memfd early on in the boot process and map it readonly, but you could seal the memfd at this point to prevent writing for paranoia

src/app/fdctl/run/run.c
```diff
@@ -417,83 +232,143 @@ main_pid_namespace( void * args ) {
...
-     init process, which will cause the kernel to terminate all other children
-     bringing all of our processes down as a group. */
-  int wstatus;
-  pid_t exited_pid = wait4( -1, &wstatus, (int)__WCLONE, NULL );
-  if( FD_UNLIKELY( exited_pid == -1 ) ) {
-    fd_log_private_fprintf_nolock_0( STDERR_FILENO, "wait4() failed (%i-%s)", errno, fd_io_strerror( errno ) );
-    exit_group( 1 );
-  }
+  /* No more access to the log file, write log lines only to STDERR from
+     here out. */
```
ptaffet-jump, line 273: Why not? `logfile` was passed to `allow_fds`

src/app/fdctl/run/run.c
```diff
@@ -417,83 +232,143 @@ main_pid_namespace( void * args ) {
...
 
-  char * name = "unknown";
-  ulong tile_idx = ULONG_MAX;
   for( ulong i=0; i<child_cnt; i++ ) {
```
ptaffet-jump, line 275: The point of this loop is just to print more friendly error messages about failure on boot, right? Otherwise isn't there a race between this code and the tile boot?

src/app/fdctl/run/run.c
```diff
@@ -417,83 +232,143 @@ main_pid_namespace( void * args ) {
...
+    if( FD_UNLIKELY( -1==poll( fds, 1+child_cnt, -1 ) ) ) {
+      fd_log_private_fprintf_nolock_0( STDERR_FILENO, "poll() failed (%i-%s)", errno, fd_io_strerror( errno ) );
+      exit_group( 1 );
+    }
+
+    for( ulong i=0; i<1+child_cnt; i++ ) {
+      if( FD_UNLIKELY( fds[ i ].revents ) ) {
+        if( FD_UNLIKELY( !i ) ) {
+          /* Parent process died, probably SIGINT, exit gracefully. */
+          exit_group( 0 );
```
ptaffet-jump, line 307: Doesn't SIGINT normally cause an ungraceful exit (130?)

src/app/fdctl/run/run.c
```diff
@@ -417,83 +232,143 @@ main_pid_namespace( void * args ) {
...
+     a group.  The parent process will also die if this process dies,
+     due to getting SIGHUP on the pipe. */
+  while( 1 ) {
+    if( FD_UNLIKELY( -1==poll( fds, 1+child_cnt, -1 ) ) ) {
+      fd_log_private_fprintf_nolock_0( STDERR_FILENO, "poll() failed (%i-%s)", errno, fd_io_strerror( errno ) );
+      exit_group( 1 );
+    }
+
+    for( ulong i=0; i<1+child_cnt; i++ ) {
+      if( FD_UNLIKELY( fds[ i ].revents ) ) {
```
ptaffet-jump, line 304: You might want to distinguish between `POLLHUP` and the other ones

src/app/fdctl/run/run.c
```diff
@@ -417,83 +232,143 @@ main_pid_namespace( void * args ) {
...
+
+    (a) pidns is the init process of a PID namespace, so if it dies the
+        kernel will terminate the child processes.
+
+    (b) main is the parent of pidns, so it can issue a waitpid() on the
+        child PID, and when it completes terminate itself.
+
+    (c) pidns is the parent of solana-labs and the tiles, so it could
+        issue a waitpid() of -1 to wait for any of them to terminate,
+        but how would it know if main has dies?
```
ptaffet-jump, line 358:
```suggestion
        but how would it know if main has died?
```

src/app/fdctl/run/run.c
```diff
@@ -417,83 +232,143 @@ main_pid_namespace( void * args ) {
...
+   processes will also die.  This is done as follows,
+
+    (a) pidns is the init process of a PID namespace, so if it dies the
+        kernel will terminate the child processes.
+
+    (b) main is the parent of pidns, so it can issue a waitpid() on the
+        child PID, and when it completes terminate itself.
+
+    (c) pidns is the parent of solana-labs and the tiles, so it could
+        issue a waitpid() of -1 to wait for any of them to terminate,
```
ptaffet-jump, line 357: If it can do this, why do you need the child->pidns pipe?

ptaffet-jump (commented):
Just a bunch of questions. Will finish reviewing tomorrow.

src/app/fdctl/Local.mk
```diff
@@ -5,11 +5,20 @@ ifdef FD_HAS_DOUBLE
...
 $(OBJDIR)/obj/app/fdctl/config.o: src/app/fdctl/config/default.toml
 
+$(OBJDIR)/obj/app/fdctl/run/run.o: src/app/fdctl/run/generated/main_seccomp.h src/app/fdctl/run/generated/pidns_seccomp.h
+$(OBJDIR)/obj/app/fdctl/run/tiles/fd_dedup.o: src/app/fdctl/run/tiles/generated/dedup_seccomp.h
+$(OBJDIR)/obj/app/fdctl/run/tiles/fd_net.o: src/app/fdctl/run/tiles/generated/net_seccomp.h
+$(OBJDIR)/obj/app/fdctl/run/tiles/fd_netmux.o: src/app/fdctl/run/tiles/generated/netmux_seccomp.h
+$(OBJDIR)/obj/app/fdctl/run/tiles/fd_pack.o: src/app/fdctl/run/tiles/generated/pack_seccomp.h
+$(OBJDIR)/obj/app/fdctl/run/tiles/fd_quic.o: src/app/fdctl/run/tiles/generated/quic_seccomp.h
+$(OBJDIR)/obj/app/fdctl/run/tiles/fd_shred.o: src/app/fdctl/run/tiles/generated/shred_seccomp.h
+$(OBJDIR)/obj/app/fdctl/run/tiles/fd_verify.o: src/app/fdctl/run/tiles/generated/verify_seccomp.h
```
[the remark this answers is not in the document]
marcus-ares: I think that Michael does it this way because there is no Local.mk in the generated folder.

marcus-ares (dismissed):

src/util/sandbox/fd_sandbox.h
```diff
@@ -32,6 +32,13 @@
    only step that will be performed is switching UID and GID to the
    provided ones.  All other arguments are ignored.
 
+   The rlimit_nofile argument determines what RLIMIT_NOFILE is set to.
+   This should almost always be zero.  It does not affect already open
+   descriptors, and only is in effect for future syscalls.  It is
+   only needed for cases where poll() or similar syscalls are used which
+   require RLIMIT_NOFILE values to be increased for no particular
+   reason.
```
ptaffet-jump, line 40: It's possible `epoll` doesn't have this issue. I don't see anything about `RLIMIT_NOFILE` on the manpage for it

src/app/fdctl/Local.mk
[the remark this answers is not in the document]
mmcgee-jump: :shrug: I couldn't get it to detect changes so just did this

src/app/fdctl/run/run.c
```diff
@@ -221,163 +142,48 @@ clone_tile( config_t *       config,
...
+    }
+    
+    char kind_id[ 32 ], config_fd[ 32 ], pipe_fd[ 32 ];
+    snprintf1( kind_id, sizeof( kind_id ), "%lu", tile->kind_id );
+    snprintf1( config_fd, sizeof( config_fd ), "%d", config_memfd );
+    snprintf1( pipe_fd, sizeof( pipe_fd ), "%d", pipefd );
+    char * args[ 9 ] = { self_exe_path, "run1", fd_topo_tile_kind_str( tile->kind ), kind_id, "--pipe-fd", pipe_fd, "--boot-memfd", config_fd, NULL };
+    if( FD_UNLIKELY( -1==execve( self_exe_path, args, envp ) ) ) FD_LOG_ERR(( "execve() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+  } else {
+    if( FD_UNLIKELY( -1==fcntl( pipefd, F_SETFD, FD_CLOEXEC ) ) ) FD_LOG_ERR(( "fcntl(F_SETFD,FD_CLOEXEC) failed (%i-%s)", errno, fd_io_strerror( errno ) ));
```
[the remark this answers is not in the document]
mmcgee-jump: It's clearer to me that the creator of the pipe (the caller) closes it, but this function clears CLOEXEC, so it needs to restore it.

src/app/fdctl/run/run.c
mmcgee-jump: Not needed imo

src/app/fdctl/run/run.c
mmcgee-jump: It makes the logic more difficult, you end up with something like

```while(1) { if wait(-1, NONBLOCK) { ..child died } else if (poll(parent_pipe_fd, NONBLOCK)) ... }```

so you have to spin a core and so on. Better just issue one syscall and let the kernel deal with it.

File: src/util/sandbox/fd_sandbox.h (hunk 12 above)
Comment (mmcgee-jump) [reply src/util/sandbox/fd_sandbox.h:40]: Made a mental note to look into this, but leaving for a future change

File: src/util/log/fd_log.h [hunk 14]
```diff
@@ -234,6 +234,20 @@
...
 #define FD_LOG_HEXDUMP_EMERG(a)   do { long _fd_log_msg_now = fd_log_wallclock(); fd_log_private_2( 7, _fd_log_msg_now, __FILE__, __LINE__, __func__, fd_log_private_hexdump_msg a ); } while(0)
 
+/* FD_LOG_RAW_ERR(()) is a special async-signal safe logger that can be called from
+   inside a signal handler.  It also does not require that the log system has been
+   initialized, and will instead only log information to stdout before exiting.
+   
+   This should not be used outside of very early in the boot process, or when tearing
+   down inside a signal handler as it can cause interleaving of log messages. */
+
+#define FD_LOG_RAW_ERR(a) do { fd_log_private_raw_2( __FILE__, __LINE__, __func__, fd_log_private_0 a ); } while(0)
```
ptaffet-jump, line 244: I looked at what functions get called before initializing the logging system. It's a scarily long list. You switched almost all of them to call this one, but I think `getpid1`, `snprintf1`, `fd_topo_validate` (and what it calls), slipped through the cracks. I'm not sure what you want to do about snprintf1, since I think that also gets called after logging is initialized. I did this analysis by hand, so I might have missed something or have false positives.

At a higher level, I like this better than the old way which called `fd_log_private_raw_2`, but I think if it weren't too gross to have normal `FD_LOG_ERR` etc behave like `FD_LOG_RAW_ERR` when not initialized, that would be easier to use.

src/app/fdctl/utility.c
```diff
@@ -270,7 +283,7 @@ snprintf1( char * s,
 }
 
 void
-self_exe( char * path ) {
+current_executable_path( char path[ PATH_MAX ] ) {
```
ptaffet-jump, line 286:
```suggestion
current_executable_path( char path[ static PATH_MAX ] ) {
```
I think that's the syntax. Richie showed me. `static` makes the compiler actually check it.

src/util/sandbox/test_sandbox.c
```diff
@@ -208,12 +208,12 @@ mountns_null( void ) {
...
 void
 seccomp_default_filter( void ) {
-  int fds[ 4 ] = { 0, 1, 2, 3 };
+  int fds[ 5 ] = { 0, 1, 2, 3, 4 };
 
   struct sock_filter seccomp_filter[ 128UL ];
   populate_sock_filter_policy_test_sandbox( 128UL, seccomp_filter );
 
-  TEST_FORK_OK( fd_sandbox( 1, getuid(), getgid(), SIZEOFA( fds ), fds, sock_filter_policy_test_sandbox_instr_cnt, seccomp_filter ); );
+  TEST_FORK_OK( fd_sandbox( 1, getuid(), getgid(), 0, SIZEOFA( fds ), fds, sock_filter_policy_test_sandbox_instr_cnt, seccomp_filter ); );
```
ptaffet-jump, line 216: It's an extremely small detail, but I think it does help to make these `0UL` so it's clear it's a count of something (which then reminds me it's the rlimit_file_cnt) and not a bool

src/app/fddev/dev1.c
```diff
@@ -13,7 +13,7 @@ void
 dev1_cmd_args( int *    pargc,
                char *** pargv,
                args_t * args) {
-  char * usage = "usage: run1 <tile>";
+  char * usage = "usage: dev11 <tile>";
```
ptaffet-jump, line 16:
```suggestion
  char * usage = "usage: dev1 <tile>";
```

src/app/fdctl/config.c
```diff
@@ -712,207 +733,245 @@ config_parse( int *    pargc,
...
       default:
-        FD_LOG_ERR(( "unknown tile kind %lu", tile->kind ));
+        FD_LOG_RAW_ERR(( "unknown tile kind %lu", tile->kind ));
     }
   }
+}
+
+int
+config_write_memfd( config_t * config ) {
+  int config_memfd = memfd_create( "config", 0 );
```
ptaffet-jump, line 967: I know this doesn't have to be unique, but I think making it clear its firedancer's config (as opposed to the Labs part maybe) can't hurt, e.g. `fd_config` or `firedancer_config`

src/app/fdctl/main1.c
```diff
@@ -20,17 +29,105 @@ struct action_alias ALIASES[] = {
...
+  int pid = getpid1(); /* Need to read /proc since we might be in a PID namespace now */;
+
+  log_path = config.log.path;
+  if( FD_LIKELY( config.log.path[ 0 ]=='\0' ) ) log_path = NULL;
+
+  fd_log_private_boot1( log_lock,
+                        0UL,
+                        config.name,
+                        0UL,    /* Thread ID will be initialized later */
+                        "main", /* Thread will be initialized later */
```
ptaffet-jump, line 100: Maybe name this something expressing that it's uninitialized if it will be initialized later? Just like an empty string or "unset" or something

src/app/fdctl/run/run.c
```diff
@@ -61,145 +74,35 @@ install_tile_signals( void ) {
...
-
-  fd_tile_config_t * config = fd_topo_tile_to_config( tile );
-
-  void * scratch_mem   = NULL;
-  if( FD_LIKELY( config->scratch_align ) ) {
-    scratch_mem = (uchar*)args->config->topo.workspaces[ tile->wksp_id ].wksp + tile->user_mem_offset;
-    if( FD_UNLIKELY( !fd_ulong_is_aligned( (ulong)scratch_mem, config->scratch_align() ) ) )
-      FD_LOG_ERR(( "scratch_mem is not aligned to %lu", config->scratch_align() ));
-  }
+extern int * fd_log_private_shared_lock;
```
ptaffet-jump, line 77: I think you can get rid of this line now, right? Seems unused. Same for in main1.c

ptaffet-jump (commented):
It's in pretty good shape, but a few more things

ptaffet-jump (approved):

PR #849, opened 2023-10-26
quic-arp - disable broadcast ips

nbridge-jump:
This code misbehaved. It's unneeded for now, so disable for now.


mmcgee-jump:
Why is it misbehaving? What's the right thing to do here instead?

mmcgee-jump (approved):

PR #850, opened 2023-10-26
ci: reorganize workflows

marcus-ares:


.github/workflows/on_main_push.yml
```diff
@@ -0,0 +1,34 @@
+name: On Main Push
```
mmcgee-jump, line 1: I don't think we should have different sets for push and pull request. How would you know that the push is going to fail?

.github/workflows/on_daily_with_change.yml
```diff
@@ -0,0 +1,28 @@
+# Kick off a daily battery of extended tests.
```
mmcgee-jump, line 1: Isn't this identical to on_daily ?

marctrem: This one just runs if there were changes since yesterday. For instance: we shouldn't run CodeQL everyday if there have been no changes. I'll clarify it in the header.

mmcgee-jump (approved):

PR #855, opened 2023-10-27
app: fix mismatched MTUs

ptaffet-jump:
Gossiped votes -> Pack was wrong in Rust, and two on the C side had their MTUs switched.


mmcgee-jump:
Is the Solana part rebased in properly?

mmcgee-jump (approved):

PR #859, opened 2023-10-27
add two base58 decoder crash cases

marcus-ares:


ripatel-fd:
These are false positives due to broken fuzzer.
https://github.com/firedancer-io/firedancer/pull/865

jumpsiegel:
false positive

PR #869, opened 2023-10-27
quic: do not clear on join

ripatel-fd:
As additional tiles might join a QUIC object (e.g. to read metrics), the join function should not clear memory.

### DIFF
--- a/src/tango/quic/fd_quic.c
+++ b/src/tango/quic/fd_quic.c
@@ -308,13 +308,6 @@ fd_quic_join( void * shquic ) {
     return NULL;
   }
 
-  /* Clear join-lifetime memory regions */
-
-  quic->cert_object     = NULL;
-  quic->cert_key_object = NULL;
-
-  memset( &quic->cb, 0, sizeof( fd_quic_callbacks_t  ) );
-
   return quic;
 }
 


mmcgee-jump:
I mean, but if you joined it and ran fd_quic_service() from another address space it's gonna crash. I think we should prevent that by gating with fd_log_group_id() or something, or not call this join?

llamb-jump (approved):

ripatel-fd:
> I mean, but if you joined it and ran fd_quic_service() from another address space it's gonna crash. I think we should prevent that by gating with fd_log_group_id() or something, or not call this join?

@mmcgee-jump There is an already established pattern of multiple joins. A join is just used to prepare shared memory access to an object, and should never write.

For example, a fd_cnc is shared by (tile, control, and monitoring). Tile writes its status, control writes the shut-down signal, and all tiles can read at any time. Obviously, the application will misbehave if control writes over the fields owned by tile.

Similarly, an fd_quic is shared by (tile, monitoring). The tile drives the fd_quic, monitoring reads the metrics region.

I think we should just document these roles along the lines of "All data and methods not marked as 'shared' may only be accessed through the join that initialized the fd_quic." And then mark the metrics region as 'shared / safe to read by secondary joins'.

To improve process isolation, we could also declare many little metrics workspaces that get flushed on housekeep. Such that the monitor is unable to see any of the other fd_quic memory (which will include temporary symmetric encryption keys, etc)

ripatel-fd:
Regardless, the behavior of overwriting memory on join is invalid and should be fixed.

There may be code relying on this behavior. Any tile that sets up a preallocated fd_quic should ensure that externally-specified parameters of fd_quic are properly initialized after the first join in the instance lifetime.

mmcgee-jump:
:shrug: stamping but can you ask kbowers if this is the correct mental model for this?

mmcgee-jump (approved):

ripatel-fd:
> 🤷 stamping but can you ask kbowers if this is the correct mental model for this?

Sure ... @kbowers-jump, what do you think?

PR #870, opened 2023-10-27
fd_ip - added diagnostics for unusual LL ADDR length

nbridge-jump:
Getting an warning reported from a user.
This adds the needed info from netlink to diagnose the issue.

### DIFF
--- a/src/tango/ip/fd_netlink.c
+++ b/src/tango/ip/fd_netlink.c
@@ -9,6 +9,7 @@
 #include <unistd.h>
 
 #include "fd_netlink.h"
+#include "../../util/cstr/fd_cstr.h"
 
 void
 fd_dump_nla_err( struct nlmsghdr * nlh, uint ip_addr, uint ifindex );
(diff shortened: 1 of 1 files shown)


src/tango/ip/fd_netlink.c
```diff
@@ -112,6 +113,96 @@ fd_nda_type_to_label( uint nda_type ) {
...
+    /* write value in hex */
+    fd_nl_write_hex( p, p_sz, &out_len, rta_data, rta_data_sz );
+
+    p    += out_len;
+    p_sz -= out_len;
+
+    rat = RTA_NEXT( rat, ratmsglen );
+  }
+
+  FD_LOG_WARNING(( "%s", buf ));
```
mmcgee-jump, line 203: Can we log at NOTICE instead, this will dump a pretty long warning to the console otherwise

nbridge-jump: NOTICE is fine with me

mmcgee-jump (approved):

PR #873, opened 2023-10-28
Zen 2 support

ripatel-fd:
- Reverts disabling of AVX-512 on Icelake Sever
- Instead, adds a dedicated Zen 2 target


ptaffet-jump (approved):

ripatel-fd:
@marcus-ares This will likely break fuzzing CI. But the issue here is that the CI environment is incorrectly using an Icelake target when it doesn't seem to support all Icelake capabilities (namely AVX-512F).

PR #881, opened 2023-10-30
quic: support loopback for QUIC connections

mmcgee-jump:
We are trying to send outgoing packets for 127.0.0.1 to the external facing NIC rather than on the loopback device, which won't work.

This is mostly to support development and testing, although it's conceivable someone would send transactions via. loopback in production as well.

### DIFF
--- a/.github/workflows/functional_tests.yml
+++ b/.github/workflows/functional_tests.yml
@@ -6,7 +6,7 @@ on:
       - main
   workflow_dispatch:
 jobs:
-  single-transaction:
+  functional-tests:
     runs-on:
       group: github-v1
     env:
@@ -24,7 +24,10 @@ jobs:
       - uses: ./.github/actions/deps
 
       - name: Build everything
-        run: make -j -Otarget fddev unit-test
+        run: make -j -Otarget fddev
 
-      - name: Run functional tests
+      - name: Run single transaction
         run: ./src/test/single-transaction.sh
+
+      - name: Run single transfer
+        run: ./src/test/single-transfer.sh
--- a/src/app/fdctl/Local.mk
+++ b/src/app/fdctl/Local.mk
@@ -3,7 +3,7 @@ ifdef FD_HAS_ALLOCA
 ifdef FD_HAS_X86
 ifdef FD_HAS_DOUBLE
 
-.PHONY: fdctl cargo
+.PHONY: fdctl cargo rust solana
 
 $(call add-objs,main1 config caps utility topology keygen ready mem spy help run/run run/tiles/tiles run/run1 run/run_solana run/tiles/tiles run/tiles/fd_net run/tiles/fd_netmux run/tiles/fd_dedup run/tiles/fd_pack run/tiles/fd_quic run/tiles/fd_verify run/tiles/fd_bank run/tiles/fd_shred run/tiles/fd_store monitor/monitor monitor/helper configure/configure configure/large_pages configure/sysctl configure/shmem configure/xdp configure/xdp_leftover configure/ethtool configure/workspace_leftover configure/workspace,fd_fdctl)
 $(call make-bin-rust,fdctl,main,fd_fdctl fd_disco fd_flamenco fd_ip fd_reedsol fd_ballet fd_tango fd_util fd_quic solana_validator)
@@ -55,6 +55,10 @@ $(OBJDIR)/bin/solana: solana/target/$(RUST_PROFILE)/solana
 	$(MKDIR) $(dir $@) && cp solana/target/$(RUST_PROFILE)/solana $@
 
 rust: $(OBJDIR)/bin/solana
+
+solana: $(OBJDIR)/bin/solana
+
+else
 
 endif
 endif
--- a/src/app/fdctl/Local.mk
+++ b/src/app/fdctl/Local.mk
@@ -3,9 +3,9 @@ ifdef FD_HAS_ALLOCA
 ifdef FD_HAS_X86
 ifdef FD_HAS_DOUBLE
 
-.PHONY: fdctl cargo rust solana
+.PHONY: fdctl cargo
 
-$(call add-objs,main1 config caps utility topology keygen ready mem spy help run/run run/tiles/tiles run/run1 run/run_solana run/tiles/tiles run/tiles/fd_net run/tiles/fd_netmux run/tiles/fd_dedup run/tiles/fd_pack run/tiles/fd_quic run/tiles/fd_verify run/tiles/fd_bank run/tiles/fd_shred run/tiles/fd_store monitor/monitor monitor/helper configure/configure configure/large_pages configure/sysctl configure/shmem configure/xdp configure/xdp_leftover configure/ethtool configure/workspace_leftover configure/workspace,fd_fdctl)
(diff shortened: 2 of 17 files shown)


src/app/fdctl/run/tiles/fd_net.c
```diff
@@ -197,6 +219,12 @@ during_frag( void * _ctx,
   fd_memcpy( ctx->frame, src, sz ); // TODO: Change xsk_aio interface to eliminate this copy
 }
 
+typedef struct __attribute__((packed)) {
+  fd_eth_hdr_t eth[1];
+  fd_ip4_hdr_t ip4[1];
+  fd_udp_hdr_t udp[1];
+} eth_ip_udp_t;
+
```
ptaffet-jump, line 227: This was just debugging code and isn't needed anymore

src/app/fdctl/run/tiles/fd_net.c
```diff
@@ -6,6 +6,10 @@
 #include "../../../../tango/xdp/fd_xdp.h"
 #include "../../../../tango/xdp/fd_xsk_private.h"
 
+#include "../../../../util/net/fd_eth.h"
+#include "../../../../util/net/fd_ip4.h"
+#include "../../../../util/net/fd_udp.h"
```
ptaffet-jump, line 11: Also unnecessary I think

src/app/fdctl/Local.mk
```diff
@@ -54,6 +54,9 @@ fdctl: $(OBJDIR)/bin/fdctl
 $(OBJDIR)/bin/solana: solana/target/$(RUST_PROFILE)/solana
 	$(MKDIR) $(dir $@) && cp solana/target/$(RUST_PROFILE)/solana $@
 
+$(OBJDIR)/bin/solana-keygen: solana/target/$(RUST_PROFILE)/solana-keygen
```
mmcgee-jump, line 57: Prefer to use `fdctl keygen` maybe add a faucet keygen for it?

src/app/fdctl/configure/sysctl.c
```diff
@@ -21,6 +21,8 @@ static const char * params[] = {
   "/proc/sys/net/core/bpf_jit_enable",
   "/proc/sys/fs/file-max",
   "/proc/sys/fs/nr_open",
+  "/proc/sys/net/ipv4/conf/lo/rp_filter",
```
mmcgee-jump, line 24: Document under `initializing.md`

src/app/fdctl/run/tiles/fd_net.c
```diff
@@ -243,29 +233,8 @@ after_frag( void *             _ctx,
...
     if( FD_UNLIKELY( *opt_sz < sizeof( eth_ip_udp_t ) ) )
-      FD_LOG_ERR(( "loopback packet too small %lu", *opt_sz ));
-
-    fd_memset( hdr->eth->dst, 0, 6UL );
-    fd_memset( hdr->ip4->daddr_c, 0, 4UL );
-    fd_memset( hdr->ip4->saddr_c, 0, 4UL );
-    hdr->ip4->check = 0U;
-    /* TODO: Do we need to update check here? */
-    hdr->ip4->check = fd_ip4_hdr_check( ( fd_ip4_hdr_t const *) FD_ADDRESS_OF_PACKED_MEMBER( hdr->ip4 ) );
+      FD_LOG_WARNING(( "loopback packet too small %lu", *opt_sz ));
```
mmcgee-jump, line 237: Why change this to WARNING? Should never happen.

src/app/fdctl/Local.mk
llamb-jump: I added the subcommand `fddev keygen pubkey <path-to-file>` which is used in the new test.

src/app/fdctl/run/tiles/fd_net.c
llamb-jump: I ended up removing this check - I had changed it to warning so we wouldn't crash at runtime if sending a small packet due to a bug. LMK if you want it added back.

src/app/fdctl/keygen.c
```diff
@@ -84,6 +94,36 @@ generate_keypair( const char * keyfile,
   if( FD_UNLIKELY( setegid( gid ) ) ) FD_LOG_ERR(( "setegid() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
 }
 
+void
+keygen_pubkey( const char * file_path ) {
+  FILE * fp = fopen( file_path, "r" );
```
mmcgee-jump, line 99: You can use `load_key_into_protected_memory( path, 1 )` here instead.

src/app/fdctl/keygen.c
```diff
@@ -7,18 +7,28 @@
...
+  KEYGEN_KEY_VOTE_ACCOUNT,
+  KEYGEN_PUBKEY,
+} keygen_key_type_t;
 
 void
 keygen_cmd_args( int *    pargc,
                  char *** pargv,
                  args_t * args) {
-  char * usage = "usage: keygen <identity|vote>";
+  char * usage = "usage: keygen <identity|vote|pubkey>";
```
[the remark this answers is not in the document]
llamb-jump: This matches the solana invocation `solana-keygen pubkey <file>`.

.github/workflows/functional_tests.yml
```diff
@@ -6,14 +6,14 @@ on:
...
-  single-transaction:
+  functional-tests:
     runs-on:
       group: github-v1
     env:
       CC: gcc
     steps:
       - name: Install ausearch for debugging seccomp/bad syscalls
-        run: sudo apt update && sudo apt install -y auditd
+        run: sudo apt update && sudo apt install -y auditd libudev-dev
```
mmcgee-jump, line 16: Think this dep can be removed now?

src/app/fdctl/keygen.c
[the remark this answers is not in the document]
mmcgee-jump: `solana-keygen` is their top level binary, so they have subcommands underneath that (`new`, `pubkey`, `grind`, ...). We don't have that, fdctl is the top level, so subcommands mostly would go under there.

If you want to have a `keys` category or something which mirrors the structure of `solana-keygen` I don't mind that, but then the args would be `<new|pubkey>` and new would take a subarg of `<identity|vote>`.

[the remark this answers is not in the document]
mmcgee-jump: (This is a public interface that will be frozen for all of time once it has a few external consumers so I think it's worth thinking about this even if it's incidental to the PR)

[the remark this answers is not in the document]
mmcgee-jump: It's fine to do that, but it should be hierarchical. Do you disagree about the hierarchy?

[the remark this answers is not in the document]
llamb-jump: I updated it to be:
```
usage:
  keys new identity
  keys new vote
  keys pubkey <path-to-keyfile>
  ```

.github/workflows/functional_tests.yml
mmcgee-jump: But you're adding it as part of this change? What requires it? I think it was here when you built the `solana-keygen` command but isn't needed now?

llamb-jump: You're right good catch, updated.

ptaffet-jump (approved):
Great!

PR #884, opened 2023-10-30
caps: log the particular resource limit if one is too low

SOELTH:
Hello frens. Was setting up my firedancer test node and a resource limit was missing but didn't know which one. Modified the ERR log to log which one is too low.

Example log: `ERR     10-30 19:09:18.536913 76797  f0   main src/app/fdctl/caps.c(158): setrlimit failed (1-EPERM-operation not permitted) for resource RLIMIT_NOFILE`

happy to modify if any of these are irrelevant / if I missed any that we might need

### DIFF
--- a/src/app/fdctl/caps.c
+++ b/src/app/fdctl/caps.c
@@ -105,17 +105,24 @@ fd_caps_check_capability( fd_caps_ctx_t * ctx,
 FD_FN_CONST static char *
 fd_caps_resource_str( fd_rlimit_res_t resource ) {
   switch( resource ) {
-    case RLIMIT_CPU:     return "RLIMIT_CPU";
-    case RLIMIT_FSIZE:   return "RLIMIT_FSIZE";
-    case RLIMIT_DATA:    return "RLIMIT_DATA";
-    case RLIMIT_STACK:   return "RLIMIT_STACK";
-    case RLIMIT_CORE:    return "RLIMIT_CORE";
-    case RLIMIT_RSS:     return "RLIMIT_RSS";
-    case RLIMIT_NOFILE:  return "RLIMIT_NOFILE";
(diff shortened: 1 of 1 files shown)


src/app/fdctl/caps.c
```diff
@@ -137,6 +154,6 @@ fd_caps_check_resource( fd_caps_ctx_t * ctx,
     rlim.rlim_cur = limit;
     rlim.rlim_max = limit;
     if( FD_UNLIKELY( setrlimit( resource, &rlim ) ) )
-      FD_LOG_ERR(( "setrlimit failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+      FD_LOG_ERR(( "setrlimit failed (%i-%s) for resource %s", errno, fd_io_strerror( errno ), fd_caps_resource_str( resource ) ));
```
mmcgee-jump, line 157: I'm curious how you ended up hitting this? If your process has CAP_SYS_RESOURCE, I don't think you should hit EPERM here?

mmcgee-jump (changes requested):
Thanks for the PR! Couple of comments

src/app/fdctl/caps.c
SOELTH: EPERM can arise from attempting to raise `RLIMIT_NOFILE` beyond `fs.nr_open`. This isn't well known and is probably worth documenting somewhere else for the next user that hits this

SOELTH:
Added the missing resource codes. Could also slim it down to include only the ones that are needed by fd. I would only expect this error to be hit on `RLIMIT_NOFILE` due to the edge case I mentioned [here](https://github.com/firedancer-io/firedancer/pull/884#discussion_r1380087392). I've erroneously made such "this won't be hit" claims a lot when dealing with kernels however...

mmcgee-jump (approved):

PR #890, opened 2023-11-01
book: add hardware requirements

### LINKED ISSUE #886: Operator book should document hardware requirements
Currently, it only documents software requirements.

anwayde:
Fixes #886

### DIFF
--- a/book/guide/getting-started.md
+++ b/book/guide/getting-started.md
@@ -1,5 +1,24 @@
 # Getting Started
 
+## Hardware Requirements
+
+Firedancer currently uses a lot of components from Solana. Hence,
+the hardware requirements are similar to that of [existing Solana
+validators](https://docs.solana.com/running-validator/validator-reqs).
+
+### Minimum Specs
+
+- 12-Core CPU @ >2.5GHz
+- 64GB RAM
+- 512GB SSD
+
+### Recommended Specs
+
+- 32-Core CPU @ >3GHz with AVX512 support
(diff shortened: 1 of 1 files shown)


book/.vitepress/config.mts
```diff
@@ -35,6 +35,7 @@ export default defineConfig({
             { text: 'Getting Started', link: 'getting-started' },
             { text: 'Configuring', link: 'configuring' },
             { text: 'Initializing', link: 'initializing' },
+            { text: 'Hardware Requirements', link: 'hardware-requirements' },
```
mmcgee-jump, line 38: Not the right place I think. Just add it to the top of the getting started guide.

book/guide/hardware-requirements.md
```diff
@@ -0,0 +1,107 @@
+# Hardware Requirements
+
+Currently, the Frankendancer validator uses a lot of components from
+Solana. As such, the hardware requirements are similar to that of [existing
+Solana validators](https://docs.solana.com/running-validator/validator-reqs).
+This page will be updated as and when these requirements change.
+
+## Node Specs
+
+### Validator Node
```
mmcgee-jump, line 10: If you're going to put tables in here, we can format them as markdown. But I think we can just remove all of this and have a few bullets with the minimum requirements, which will just match the Solana Labs ones.

mmcgee-jump (changes requested):
q

Issue #675, opened 2023-09-06
Don't install or use XDP program if kernel is too old


mmcgee-jump:
User should get a nice error message and be told to disable XDP in config file, if their kernel is too old. Project should still build correctly. Probably use `dlsym` with `bpf` syscall.

ripatel-fd:
Is this about checking whether the kernel headers support XDP (compile-time) or whether the actual kernel syscalls function correctly?

I'm not sure we can rely on kernel version checks using the number alone. On the RHEL kernels we use, bpf functionality is backported to very old versions. The original upstream releases of those versions cannot run our BPF program, as far as I know.

For the kernel header problem, we can just ship our own. For at-runtime feature detection, it would be nice to have some sort of test suite that the user can run to generate an appropriate config file.

mmcgee-jump:
Yeah, I don't think we should rely on these headers, and then we need very functional error messages if the bpf() syscalls fail at runtime

ripatel-fd:
I'll try to run fdctl under Ubuntu 18.04. If it succeeds, I'll close this issue.

ripatel-fd:
```
NOTICE  06-23 16:11:25.171015 3409   1    net:0 src/disco/topo/fd_topo_run.c(30): booting tile net:0 pid:3409 tid:3468
WARNING 06-23 16:11:25.171102 3409   1    net:0 src/waltz/xdp/fd_xdp_redirect_user.c(128): Failed to create XSKMAP (22-EINVAL-invalid argument)
ERR     06-23 16:11:25.172968 3409   1    net:0 src/app/fdctl/run/tiles/fd_net.c(547): fd_xdp_link_session_init failed
ripatel@ubuntu-18:~$ uname -a
Linux ubuntu-18 4.15.0-213-generic #224-Ubuntu SMP Mon Jun 19 13:30:12 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

I managed to find an old kernel but that one doesn't even have AF_XDP support.

@mmcgee-jump Almost all of the userbase uses Ubuntu. 18.04 is EOL and AF_XDP works on other versions. What do you suggest we do here? Should we just close this?

PR #893, opened 2023-11-02
Linux uapi

ripatel-fd:
- compat/linux: add sync-uapi script
- compat/linux: import Linux 6.5.9 uapi


mmcgee-jump (approved):

mmcgee-jump:
We need to add some build time kernel feature detection to make sure we fail if it's ENOSYS, but otherwise LGTM. Also would be better to just have a minimal header with the stuff we need. See eg https://github.com/firedancer-io/firedancer/issues/675

PR #896, opened 2023-11-02
deps: add libucontext

ripatel-fd:
Experiment to replace ucontext system dependency, which is no longer part of POSIX nor musl.

(Still need to add licensing information here, do not approve for merge)


ripatel-fd:
Abandoning as we won't support musl in the near time.

PR #898, opened 2023-11-02
tile: work around musl strict aliasing bug

ripatel-fd:
musl's `CPU_SET` macro in `sched.h` has a bug where it casts a struct to a pointer of unsigned longs.

This is clearly a strict aliasing violation and has no place in a general-purpose libc, since it violates ISO C conventions.

This workaround redefines the internal macro causing the problem.

(We should probably drop support for musl because of conflicting philosophy that cause these recurring incompatibilities)


jumpsiegel:
Not supporting musl for now

PR #902, opened 2023-11-03
book: add expected genesis hash

anwayde:


book/guide/getting-started.md
```diff
@@ -115,6 +117,14 @@ on the local machine. The `identity_path` and `vote_account_path` should
 be Solana Labs style keys, which can be generated with the Solana Labs
 CLI.
 
+::: warning WARNING
```
mmcgee-jump, line 120: Not a warning, I think just have the text here normally.

mmcgee-jump (changes requested):
a

mmcgee-jump:
I think adding docs is ok but this should be self documenting in the error message. Can we make the error say how to fix it? Eg,

```
    FD_LOG_EMERG(( "Attempted to start against a live non-testnet cluster. Firedancer is not "
                   "ready for production deployment, has not been tested, and is "
                   "missing consensus critical functionality. Joining a live Solana "
                   "cluster may destabilize the network. Please do not attempt. You can start against the testnet cluster by specifying the testnet [consensus.expected_genesis_hash] in your configuration file." ));
```

The underlying problem here is we are allowing an empty genesis hash for fdctl, which shouldn't be allowed. You can update `config_parse` to take a boolean of whether it's `fddev` or `fdctl`. For `fdctl`, there should be a check before we would hit this EMERG error that `expected_genesis_hash` has been specified, with an error if it has not. Then these users wouldn't have got to this case anyway.

mmcgee-jump (approved):

mmcgee-jump (approved):

PR #909, opened 2023-11-04
fdctl: increase fs.nr_open sysctl

### LINKED ISSUE #908: Add `fs.file-max` to sysctl, better checking of this sysctl when raising RLIMIT_NOFILE
* We can't raise RLIMIT_NOFILE past the `fs.file-max` sysctl, so that sysctl should get raised during the `sysctl.c` step. Make sure to add to book documentation.
* The `caps.c` check for has_capability(CAP_SYS_RESOURCE) isn't enough to raise this, and the check to raise it should also fail with a good error message if fs.file-max isn't high enough.

mmcgee-jump:
Fixes #908


ptaffet-jump (approved):
I'm no expert on this, but it looks fine. Stamped.

PR #910, opened 2023-11-05
config: error if iface address is not public

### LINKED ISSUE #906: Exit with error if IP address is not public
fdctl should prevent this common misconfiguration (eg, localhost, local subnets)

SOELTH:
fixes #906 

lmk if anything is in the wrong place (pretty new to this repo) or if any formatting issues 

note that there's some technically non-public ranges that aren't covered here, but this should catch the common cases...

### DIFF
--- a/src/app/fdctl/config.c
+++ b/src/app/fdctl/config.c
@@ -745,6 +745,11 @@ config_parse( int *    pargc,
       FD_LOG_ERR(( "could not get name of interface with index %u", ifindex ));
   }
 
+  ulong cluster = determine_cluster( result.gossip.entrypoints_cnt,
+                                     result.gossip.entrypoints,
+                                     result.consensus.expected_genesis_hash );
+  result.is_live_cluster = cluster != FD_CONFIG_CLUSTER_UNKNOWN;
+
   if( FD_UNLIKELY( result.development.netns.enabled ) ) {
     if( FD_UNLIKELY( strcmp( result.development.netns.interface0, result.tiles.net.interface ) ) )
       FD_LOG_ERR(( "netns interface and firedancer interface are different. If you are using the "
@@ -759,7 +764,14 @@ config_parse( int *    pargc,
   } else {
     if( FD_UNLIKELY( !if_nametoindex( result.tiles.net.interface ) ) )
       FD_LOG_ERR(( "configuration specifies network interface `%s` which does not exist", result.tiles.net.interface ));
-    result.tiles.net.ip_addr = listen_address( result.tiles.net.interface );
+    uint iface_ip = listen_address( result.tiles.net.interface );
+    if ( FD_UNLIKELY( !fd_ip4_addr_is_public ( iface_ip ) && result.is_live_cluster ) )
+      FD_LOG_ERR(( "Trying to use network interface `%s` for listening to incoming transactions, "
+                   "but it has IPv4 address " FD_IP4_ADDR_FMT " "
+                   "which is part of a private network and will not be routable for other Solana network nodes.", 
+                   result.tiles.net.interface, FD_IP4_ADDR_FMT_ARGS( iface_ip ) ));
+
+    result.tiles.net.ip_addr = iface_ip;
     mac_address( result.tiles.net.interface, result.tiles.net.mac_addr );
   }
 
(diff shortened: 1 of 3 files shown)


src/app/fdctl/config.c
```diff
@@ -759,7 +759,12 @@ config_parse( int *    pargc,
   } else {
     if( FD_UNLIKELY( !if_nametoindex( result.tiles.net.interface ) ) )
       FD_LOG_ERR(( "configuration specifies network interface `%s` which does not exist", result.tiles.net.interface ));
-    result.tiles.net.ip_addr = listen_address( result.tiles.net.interface );
+      
+    uint iface_ip = listen_address( result.tiles.net.interface );
+    if ( FD_UNLIKELY( !fd_ip4_addr_is_public ( iface_ip ) ) )
```
mmcgee-jump, line 764: I think we probably want to allow this for non-public clusters. Eg, if `result.is_live_cluster` below, then it must be public but otherwise private subnets are ok.

src/util/net/fd_ip4.h
```diff
@@ -21,6 +21,16 @@
 #define FD_IP4_OPT_RA  ((uchar)148) /* This option is a router alert option */
 #define FD_IP4_OPT_EOL ((uchar)0)   /* This is the end of the options list */
 
+#define IP4_PRIVATE_RANGE1_START 0x0a000000 // 10.0.0.0
```
mmcgee-jump, line 24: Up to you but could be a little clearer to use `FD_IP4_ADDR(10,0,0,0)` here, will get optimized down anyway.

src/util/net/test_ip4.c
```diff
@@ -38,6 +38,31 @@ test_cstr_to_ip4_addr( void ) {
   FD_TEST( fd_cstr_to_ip4_addr( "36893488147419103232.0.0.0", &ip )==0  );
 }
 
+static void
+test_ip4_addr_is_public( void ) {
+  // Public addresses should return 1 for fd_ip4_addr_is_public
+  FD_TEST( fd_ip4_addr_is_public(0x08080808) == 1 ); // 8.8.8.8
```
mmcgee-jump, line 44: Same comment as above for these addresses.

src/app/fdctl/config.c
```diff
@@ -759,7 +759,12 @@ config_parse( int *    pargc,
...
+      FD_LOG_ERR(( "configuration specifies network interface `%s` which does not have a public IP address", result.tiles.net.interface ));
```
mmcgee-jump, line 765: The comment could be a little wrong, since the configuration can have an empty interface which will get autodetected above. How about

```suggestion
      FD_LOG_ERR(( "Trying to use network interface `%s` for listening to incoming transactions, but it has IPv4 address " FD_IP4_ADDR_FMT " which is part of a private network and will not be routable for other Solana network nodes.", result.tiles.net.interface, FD_IP4_ADDR_FMT_ARGS( iface_ip ) ))
```

mmcgee-jump (changes requested):
Comments inline

mmcgee-jump:
Nice change though! Thank you for the PR :)

SOELTH:
Added suggested changes and fixed some assumptions around endianness of the IPs. 

Also curious what assumptions should be around FD_IP4_ADDR. Considering everything it generates is in little endian, I would assume FD_IP4_ADDR will always be host byte order in the future? This pr operates with that assumption

ptaffet-jump:
Hi, and thanks for the contribution!

First, to answer your question, Firedancer intentionally only supports little endian platforms. It's part of our [machine model](https://github.com/firedancer-io/firedancer/blob/e2df28e949bfbd33aa737348f49543520008ff8b/src/util/fd_util_base.h#L246-L258). We've made a deliberate tradeoff of some portability to simplify code and eliminate the potential for subtle errors.

That said, I think your code is a little mixed up on endianness. In my mind, IP addresses are byte strings like mac addresses, so they don't have an endianness, but if you want to describe them as having endianness, then `FD_IP4_ADDR` returns big endian results. 
`#define FD_IP4_ADDR(x,y,z,w) (((uint)(x)) | (((uint)(y)) << 8) | (((uint)(z)) << 16) | (((uint)(w)) << 24))`

`"1.2.3.4" -> FD_IP4_ADDR(0x01, 0x02, 0x03, 0x04) = 0x04030201`

Since you want to compare in lexicographical order, you do need an endianness conversion, but the appropriate one would be `ntohl`. I'd keep the preprocessor `#define`s in network byte order (meaning no function call in the macro) and only do the conversion in the function itself. Of course, endianness conversion is a self-inverse, so `ntohl` and `htonl` actually do the same thing, so the test passes. All that said, I'd actually suggest using `fd_uint_bswap` from fd_bits instead of `ntohl`. That'll let you get rid of the cast to `uint` as well.

Finally, a few small nitpicks: we put spaces after the opening parenthesis and before the closing parenthesis for any function call.  Unusual, I know, but that's the format. Macros that are in all caps don't need the space. Second, we love internal alignment. Your alignment in `fd_ip4_addr_is_public` is great, but the list of IP addresses in the `#define`s and in the test should be aligned. Just prefix with spaces to make every byte in an IP address take up 3 characters.

SOELTH:
Ah thank you for the clarifications this makes much more sense now! Implemented suggested changes and fixes in a separate comment so it's easy to see the diff. Happy to rebase if needed :)

mmcgee-jump:
Amazing! Nice change. I noticed you caught some stray changes to the `solana` submodule in there. Can you revert those and we are good to go!

SOELTH:
Oops! Squashed and cut out the stray solana changes 🙃

ptaffet-jump (approved):

mmcgee-jump (approved):

PR #916, opened 2023-11-06
Txn size proof

ripatel-fd:
- cbmc: remove cbmc-proof.txt marker file
- cbmc: improve fd_log stub
- txn: add FD_TXN_MIN_SERIALIZED_SZ


ptaffet-jump (approved):
Thanks! This matches [fixture4](https://github.com/firedancer-io/firedancer/blob/main/src/ballet/txn/fixtures/transaction4.bin), which is good confirmation.

PR #937, opened 2023-11-08
gossip protocol implementation and spy

asiegel-jt:


jumpsiegel (approved):
Approving since richie plans on fuzzing/hardening this in a different PR

PR #941, opened 2023-11-10
util: add histogram

mmcgee-jump:


src/util/hist/fd_hist.h
```diff
@@ -0,0 +1,97 @@
...
+             ulong  max_value ) {
+  if( FD_UNLIKELY( bucket_cnt<2 ) ) return NULL;
+  if( FD_UNLIKELY( max_value<bucket_cnt-1 ) ) return NULL;
+  if( FD_UNLIKELY( max_value % (bucket_cnt-1) ) ) return NULL;
+  if( FD_UNLIKELY( !fd_ulong_is_pow2( max_value / (bucket_cnt-1) ) ) ) return NULL;
+
+  fd_hist_t * hist = (fd_hist_t*)mem;
+  hist->max_value = max_value;
+  hist->shift     = (ulong)fd_ulong_find_msb( max_value / (bucket_cnt-1) );
+  return (void*)hist;
```
ptaffet-jump, line 65: You need to memset the buckets to 0, right?

ripatel-fd:
This seems use case specific so I suggest keeping this code in the disco subdirectory of the tile that needs it

mmcgee-jump:
It's going to be used by every tile, and code in disco/tango and probably util as well (eg, wksp_alloc counters). I think this is the right place.

mmcgee-jump:
Replaced by #961

PR #943, opened 2023-11-13
ffi: separate libs into rust modules

mmcgee-jump:
Bindgen adds a preamble to the generated code for each module, so when we concat the modules together, they can conflict if they caused bindgen to generate similar preambles.  We tidy this up by giving each bindgenned source file an actual module.


ptaffet-jump (approved):
I think `gen_ballet` etc. would be more clear, but this is okay.

PR #944, opened 2023-11-13
Add a gossip spy command to fdctl

asiegel-jt:


src/app/fdctl/config.c
```diff
@@ -168,6 +168,7 @@ static int parse_key_value( config_t *   config,
   ENTRY_VSTR  ( ., gossip,              entrypoints                                               );
   ENTRY_BOOL  ( ., gossip,              port_check                                                );
   ENTRY_USHORT( ., gossip,              port                                                      );
+  ENTRY_USHORT( ., gossip,              shred_version                                             );
```
mmcgee-jump, line 171: Don't think this belongs here. We already have `expected_shred_version` in this file.

src/app/fdctl/main1.c
```diff
@@ -16,6 +17,7 @@ action_t ACTIONS[ ACTIONS_CNT ] = {
   { .name = "keygen",     .args = keygen_cmd_args,    .fn = keygen_cmd_fn,     .perm = NULL,                .description = "Generate new keypairs for use with the validator" },
   { .name = "ready",      .args = NULL,               .fn = ready_cmd_fn,      .perm = NULL,                .description = "Wait for all tiles to be running" },
   { .name = "mem",        .args = NULL,               .fn = mem_cmd_fn,        .perm = NULL,                .description = "Print workspace memory and tile topology information" },
+  { .name = "spy",        .args = NULL,               .fn = spy_cmd_fn,        .perm = NULL,                .description = "Run the gossip spy" },
```
mmcgee-jump, line 20:
```suggestion
  { .name = "spy",        .args = NULL,               .fn = spy_cmd_fn,        .perm = NULL,                .description = "Spy on  and print out gossip traffic" },
```

src/app/fdctl/main1.c
```diff
@@ -130,6 +132,7 @@ fdctl_boot( int *        pargc,
   config.log.log_fd = fd_log_private_logfile_fd();
   fd_shmem_private_boot( pargc, pargv );;
   fd_tile_private_boot( 0, NULL );
+  fd_flamenco_boot( pargc, pargv );
```
mmcgee-jump, line 135: Can we remove this for now? Not sure it's needed and it's glibc specific.

src/app/fdctl/spy.c
```diff
@@ -0,0 +1,249 @@
+/*
+ build/native/gcc/bin/fdctl spy --config src/app/fdctl/config/testnetoml
```
mmcgee-jump, line 2:
```suggestion
 build/native/gcc/bin/fdctl spy --config src/app/fdctl/config/testnet.toml
```

mmcgee-jump (changes requested):
see comments

src/app/fdctl/config.c
asiegel-jt: It's in consensus though

src/app/fdctl/main1.c
asiegel-jt: The yaml dump depends on this for printing public keys

mmcgee-jump: Can you make it not depend on this?

asiegel-jt: That would be a @ripatel-fd thing

asiegel-jt: It's harmless in the non-gcc case

src/app/fdctl/config/testnet.toml
```diff
@@ -14,3 +14,4 @@
         "eoKpUABi59aT4rR9HGS3LcMecfut9x7zJyodWWP43YQ",
         "9QxCLckBiJc783jnMvXZubK4wH86Eqqvashtrwvcsgkv",
     ]
+    expected_shred_version = 61807
```
mmcgee-jump, line 17: Shouldn't be here, these change frequently

src/app/fdctl/main1.c
mmcgee-jump: It's not going to dump it properly? This dependency is annoying and I'd prefer to not have it sneak in here when it's not really needed yet.

PR #945, opened 2023-11-13
Add tpu_reasm class

ripatel-fd:
- quic: add tpu_reasm class
- quic: integrate tpu_reasm into tile


src/app/fdctl/config/default.toml
```diff
@@ -687,6 +687,8 @@ dynamic_port_range = "8900-9000"
         # milliseconds
         idle_timeout_millis = 10000
 
+        reasm_cnt = 16384
```
mmcgee-jump, line 690:
```suggestion
        reassembly_buffer_depth = 16384
```

Please add a long descriptive comment, targeted at validator operators who are maybe unfamiliar with the code and not super technical. The rest of the file should be a decent example.

mmcgee-jump (approved):

PR #947, opened 2023-11-14
Ripatel/spellcheck

ripatel-fd:


mmcgee-jump:
Looked at like 30 files and seems good.

mmcgee-jump (approved):

ripatel-fd:
I also double checked all of my changes

PR #955, opened 2023-11-15
can't get child exit code without sandbox (2)

mmcgee-jump:
Thanks to Richie for the suggestion.


ripatel-fd:
There's still a deadlock in process tree destruction. I'm not sure if this is a separate bug: https://github.com/firedancer-io/firedancer/issues/959

PR #966, opened 2023-11-16
shmem: Remove invalid call to `munmap` in `fd_shmem_join`

Fire30:
The call to `munmap` is invalid as it will only be called when `shmem` is set to `MAP_FAILED`. This is not a valid userspace address and `munmap` will fail every time.


ptaffet-jump (approved):

ptaffet-jump:
Sorry, I was waiting for you to click merge and then I realized that you may not have permission to do that. I merged it. Thanks!

Fire30:
No worries! Yeah I do not have merging permissions.

PR #968, opened 2023-11-16
shmem: randomize address of shared memory mappings in tile processes

Fire30:
Due to the behavior of `mmap` in the Linux kernel the shared memory mappings in each tile processes will most likely be adjacent to each other. Relevant output from `/proc/pid/maps` is below:

```
7f64c0000000-7f6500000000 rw-s ... 6736764  /mnt/.fd/.gigantic/fd1_verify_dedup.wksp
7f6500000000-7f6540000000 rw-s ... 6736763  /mnt/.fd/.gigantic/fd1_quic_verify.wksp
7f6577200000-7f6577400000 rw-s ... 6736773  /mnt/.fd/.huge/fd1_verify.wksp
```

This makes it such that out of bounds access vulnerabilities can possibly corrupt other shared memory regions causing further unintended behavior.

To mitigate this possible situation we introduce a function `fd_shmem_private_get_random_mappable_addr` which will return a random mappable address for a passed size that can be used with `MAP_FIXED`. Shared memory regions now will no longer be adjacent to each other.

### DIFF
--- a/src/util/shmem/fd_shmem_user.c
+++ b/src/util/shmem/fd_shmem_user.c
@@ -10,6 +10,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <sys/mman.h>
+#include <sys/random.h>
 
 /* fd_shmem_private_key converts the cstr pointed to by name into a
    valid key and stores it at the location pointed to by key assumed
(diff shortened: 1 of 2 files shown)


Fire30:
So I believe this can be made better. I just want to make sure that this is something that you all would want implemented. 

If this is something that you think makes sense to do I can do the following:
- Right now the mask we use is for gigantic pages, passing a flag in for whether we are trying to map huge/gigantic would allow for more randomization
- Document the functions (and possibly place them somewhere else so they can be used elsewhere)
- Better address generation. I am open to a better strategy for creating random addresses. 

Also finally is there a tool you use for formatting the codebase? I used uncrustify with the lint.cfg and it did not seem to format it completely properly. The following command I used:

```bash
uncrustify -c contrib/lint.cfg --no-backup src/util/shmem/fd_shmem_user.c
```

Thanks!

src/util/shmem/fd_shmem_user.c
```diff
@@ -74,6 +74,46 @@ fd_shmem_private_map_query_by_addr( fd_shmem_join_info_t * map,
...
+  }
+
+  return 1;
+}
+
+static ulong
+fd_shmem_private_get_random_mappable_addr( ulong size ) {
+  ulong ret_addr = 0;
+
+  fd_rng_t _rng[1]; fd_rng_t * rng = fd_rng_join( fd_rng_new( _rng, (uint)fd_tickcount(), 0UL ) );
```
mmcgee-jump, line 99: I think prefer `getrandom(2)` here since it's security related.

src/util/shmem/fd_shmem_user.c
```diff
@@ -74,6 +75,46 @@ fd_shmem_private_map_query_by_addr( fd_shmem_join_info_t * map,
...
+  if( (ulong)mmap_ret != addr ) {
+    munmap( mmap_ret, size );
+    return 0;
+  }
+
+  return 1;
+}
+
+static ulong
+fd_shmem_private_get_random_mappable_addr( ulong size ) {
```
mmcgee-jump, line 97: From the man page for `MAP_FIXED`:

```
If  the memory region specified by addr and len overlaps pages of any existing mapping(s), then the overlapped part of the existing mapping(s) will be discarded. [...] Furthermore, this option is extremely hazardous (when used on its own), because it forcibly removes preexisting mappings, making it easy for a multithreaded process to corrupt its own address space.
```

Is this right? Do you need to keep rolling if you conflict with an existing mapping? (I guess there's basically no chance with a true RNG for 64 bits).

I took a look at the chrome randomizer and it gave a few pointers as well,

https://github.com/chromium/chromium/blob/abe2b94097458ae7e3741706b310aa70c4e7ee07/base/allocator/partition_allocator/src/partition_alloc/address_space_randomization.h#L146

We want to disable for ASAN builds. And I'm not sure if the address space we use matches

```
      // Linux (and macOS) support the full 47-bit user space of x64 processors.
      // Use only 46 to allow the kernel a chance to fulfill the request.
      PA_ALWAYS_INLINE constexpr uintptr_t ASLRMask() {
        return AslrMask(46);
      }
      PA_ALWAYS_INLINE constexpr uintptr_t ASLROffset() {
        return AslrAddress(0);
      }
```

They also just mask off the bits outside bounds rather than repeatedly rolling.

```
  uintptr_t random = static_cast<uintptr_t>(internal::RandomValue());

#if BUILDFLAG(HAS_64_BIT_POINTERS)
  random <<= 32ULL;
  random |= static_cast<uintptr_t>(internal::RandomValue());

  // The ASLRMask() and ASLROffset() constants will be suitable for the
  // OS and build configuration.
  random &= internal::ASLRMask();
  random += internal::ASLROffset();
```

src/util/shmem/fd_shmem_user.c
```diff
@@ -74,6 +75,59 @@ fd_shmem_private_map_query_by_addr( fd_shmem_join_info_t * map,
...
+  return 1;
+}
+
+static ulong
+fd_shmem_private_get_random_mappable_addr( ulong size, 
+                                           ulong page_size ) {
+  ulong ret_addr = 0;
+
+  for(;;){
+    long n = getrandom(&ret_addr, sizeof(ret_addr), 0);
```
mmcgee-jump, line 112: You need to check `n == sizeof(ret_addr)` and loop til you get all the bytes. From `man 2 getrandom`

> On success, getrandom() returns the number of bytes that were copied to the buffer buf.  This may be less than the number of bytes requested via buflen if either GRND_RANDOM was specified in flags and insufficient entropy was present in the random source or the system call was interrupted by a signal.

src/util/shmem/fd_shmem_user.c
```diff
@@ -74,6 +75,59 @@ fd_shmem_private_map_query_by_addr( fd_shmem_join_info_t * map,
...
+
+  return 1;
+}
+
+static ulong
+fd_shmem_private_get_random_mappable_addr( ulong size, 
+                                           ulong page_size ) {
+  ulong ret_addr = 0;
+
+  for(;;){
```
mmcgee-jump, line 111: I would do `for( ulong i=0; i<1000; i++ ) {` and `FD_LOG_ERR(( "unable to find random address for memory map" ))` or whatever if you get past the loop so we can't hang here. Maybe use a different value than 1000, depends how often this loop will fail but you want to make it 0.00000001% or whatever.

src/util/shmem/fd_shmem_user.c
```diff
@@ -74,6 +75,59 @@ fd_shmem_private_map_query_by_addr( fd_shmem_join_info_t * map,
...
+  void *mmap_ret;
+
+  mmap_ret = mmap( (void*)addr, size, PROT_READ, MAP_ANON|MAP_PRIVATE, -1, 0 );
+
+  if( mmap_ret == MAP_FAILED )
+    return 0;
+
+  /* Only call munmap on failure case. On success we want to keep the mapping */
+  if( (ulong)mmap_ret != addr ) {
+    munmap( mmap_ret, size );
```
mmcgee-jump, line 99: > On success, munmap() returns 0.  On failure, it returns -1, and errno is set to indicate the cause of the error (probably to EINVAL).

I think check the return code and error if it fails, so we don't have a stray memory map left in the process.

mmcgee-jump (approved):
lgtm

mmcgee-jump:
Looks good! Filed #989 to harden a little bit in future.

Jefferson111:
Just wondering, what would be the difference in [probing](https://developers.redhat.com/blog/2020/05/22/stack-clash-mitigation-in-gcc-part-3#) vs this current randomization approach? (since guard regions already exist in the current tile's stack)

Like on stack memory allocation like `alloca/scratch`, introduce an additional API that performs probing to catch those OOB access.

Preliminary thought is that randomization is dependent on the density of memory utilization (probability of hitting any memory region is lower as memory utilization decrease) and amount of memory (probability of targeting a chosen memory region is lower as memory increase). But in terms of costs, randomization would be cheaper than probing I guess since it doesn't require those extra checks every time.

PR #974, opened 2023-11-18
shred: batch microblocks up to one full FEC set

ptaffet-jump:
Without batching, we produce many small FEC sets, which make inefficient use of our parity shred budget. This implements the simplest form of batching, where each batch is still limited to one FEC set.

This change enables relaxing some limits in pack. I'll make those changes in another PR.

### DIFF
--- a/src/app/fdctl/run/tiles/fd_shred.c
+++ b/src/app/fdctl/run/tiles/fd_shred.c
@@ -81,9 +81,6 @@
 
 #define FD_SHRED_TILE_SCRATCH_ALIGN 128UL
 
-/* FD_BANK_SHRED_MTU comes from the maximum size payload can fit in one
-   FEC set plus the fd_entry_batch_meta_t header. */
-#define FD_BANK_SHRED_MTU 63719UL
 
 #define NET_IN_IDX      0
 #define POH_IN_IDX      1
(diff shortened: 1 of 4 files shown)


src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -342,35 +357,58 @@ during_frag( void * _ctx,
...
+    ctx->pending_batch.slot = entry_meta->slot;
+    /* Ugh, yet another memcpy */
+    fd_memcpy( ctx->pending_batch.payload + ctx->pending_batch.pos, entry, entry_sz );
+    ctx->pending_batch.pos += entry_sz;
+    ctx->pending_batch.microblock_cnt++;
+
+    int last_in_batch = (entry_meta->tick==entry_meta->bank_max_tick_height) | (ctx->pending_batch.pos > PENDING_BATCH_WMARK);
+
+    if( FD_UNLIKELY( last_in_batch )) {
+      fd_shredder_init_batch( ctx->shredder, ctx->pending_batch.raw, ctx->pending_batch.pos, entry_meta );
```
ptaffet-jump, line 394:
```suggestion
      fd_shredder_init_batch( ctx->shredder, ctx->pending_batch.raw, ctx->pending_batch.pos + sizeof(ctx->pending_batch.microblock_cnt), entry_meta );
```
I think we need this +8 here, but I'm not sure why my tests worked fine.

mmcgee-jump:
Is the solana side properly rebased?

mmcgee-jump (approved):

PR #977, opened 2023-11-19
fdctl: pass accounts_path onto labs client

kaplanmaxe:
Despite `[ledger.accounts_path]` being a valid configuration option, the path never gets passed onto the labs client. This causes both accounts and ledger to be written to the same disk, which can be painful on nodes with smaller disks for operators.

### DIFF
--- a/src/app/fdctl/run/run_solana.c
+++ b/src/app/fdctl/run/run_solana.c
@@ -85,6 +85,7 @@ solana_labs_main( void * args ) {
 
   /* ledger */
   ADD( "--ledger", config->ledger.path );
+  if( strcmp( "", config->ledger.accounts_path ) ) ADD( "--accounts", config->ledger.accounts_path );
   ADDU( "--limit-ledger-size", config->ledger.limit_size );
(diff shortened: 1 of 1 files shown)


src/app/fdctl/run/run_solana.c
```diff
@@ -85,6 +85,7 @@ solana_labs_main( void * args ) {
 
   /* ledger */
   ADD( "--ledger", config->ledger.path );
+  ADD( "--accounts", config->ledger.accounts_path );
```
mmcgee-jump, line 88: Thanks for the fix! I think we should only pass this if `if( !strcmp( "", config->ledger.accounts_path ) )`

mmcgee-jump (changes requested):

src/app/fdctl/run/run_solana.c
kaplanmaxe: done!

mmcgee-jump (approved):

mmcgee-jump:
Thank you for the contribution!!

PR #983, opened 2023-11-21
quic: retry fixes

ripatel-fd:


lidatong (approved):

nbridge-jump:
For what it's worth, looks good to me also

PR #985, opened 2023-11-22
fdctl: preserve ticks_per_ns accross processes

mmcgee-jump:
Now that processes are restarted with execve() to get randomized address layout, they lose the fd_tempo_tick_per_ns value.  This isn't much of a problem normally, they will be similar and just used for things like randomizing wait times.

But duration histograms for metrics are an exception.  These histograms have buckets defined in wallclock time (eg, 0-1s, 1s-2s, 2s+).  But they are measured on the tile in ticks (eg, 0-1s/ticks_per_sec, 1s-2s/ticks_per_sec,, 2s/ticks_per_sec+), to enable faster collection without scaling.  The metrics tile that reports to external viewers will rescale to wallclock time.

So the tiles doing collection and the tile doing reporting need to agree on the ticks_per_ns value so they can create the same histogram buckets.  We achieve this by just passing it through and restoring from the config object.

### DIFF
--- a/src/tango/tempo/fd_tempo.c
+++ b/src/tango/tempo/fd_tempo.c
@@ -100,45 +100,60 @@ fd_tempo_tickcount_model( double * opt_tau ) {
   return t0;
 }
 
+static double mu;
+static double sigma;
+static int explicit_set;
+
+void
+fd_tempo_set_tick_per_ns( double _mu,
+                          double _sigma ) {
+  explicit_set = 1;
+  mu    = _mu;
+  sigma = _sigma;
+}
+
 double
 fd_tempo_tick_per_ns( double * opt_sigma ) {
-  static double mu;
-  static double sigma;
 
   FD_ONCE_BEGIN {
 
-    /* We measure repeatedly how much the tickcount and wallclock change
-       over the same approximately constant time interval.  We do a pair
-       observations to minimize errors in computing the interval (note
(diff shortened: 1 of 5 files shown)


src/tango/tempo/fd_tempo.c
```diff
@@ -100,50 +100,66 @@ fd_tempo_tickcount_model( double * opt_tau ) {
   return t0;
 }
 
+static double mu;
+static double sigma;
+static int explicit_set;
+
+void
+fd_tempo_private_set_tick_per_ns( double _mu,
```
ptaffet-jump, line 108: I don't think this needs to be `private`. I'd be fine with it public, though make sure to document it.

src/tango/tempo/fd_tempo.c
```diff
@@ -100,50 +100,66 @@ fd_tempo_tickcount_model( double * opt_tau ) {
...
+          FD_LOG_WARNING(( "unable to measure tick_per_ns accurately; using fallback and attempting to continue" ));
+          mu = 3.; sigma = 1e-7;
+          break;
+        }
       }
     }
 
   } FD_ONCE_END;
 
+finish:
```
ptaffet-jump, line 162: The compilation issues seem to be due to this unused label

ptaffet-jump (approved):

ptaffet-jump (approved):

PR #987, opened 2023-11-22
verify: validate txn fields and remove double fetches

Fire30:
In `after_frag` the txn is located in shared memory accessible to the dedup tile and the contents are populated the quic tile. We add checks to validate all offset fields and remove double fetches of non-data fields.

### DIFF
--- a/src/app/fdctl/run/tiles/fd_verify.c
+++ b/src/app/fdctl/run/tiles/fd_verify.c
@@ -95,21 +95,46 @@ after_frag( void *             _ctx,
 
   fd_verify_ctx_t * ctx = (fd_verify_ctx_t *)_ctx;
 
+  /* Sanity check that should never fail. We should have atleast
+     FD_TPU_DCACHE_MTU bytes available. */
+  if( FD_UNLIKELY( *opt_sz < sizeof(ushort) ) ) {
+    FD_LOG_ERR( ("invalid opt_sz(%x)", *opt_sz ) );
+  }
+
   uchar * udp_payload = (uchar *)fd_chunk_to_laddr( ctx->out_mem, ctx->out_chunk );
   ushort payload_sz = *(ushort*)(udp_payload + *opt_sz - sizeof(ushort));
+
+  /* Make sure payload_sz is valid */
+  if( FD_UNLIKELY( payload_sz > FD_TPU_DCACHE_MTU ) ) {
+    FD_LOG_ERR( ("invalid payload_sz(%x)", payload_sz) );
+  }
+
+  /* txn contents are located in shared memory accessible to the dedup tile 
+     and the contents are controlled by the quic tile. We must perform
+     validation */
   fd_txn_t * txn = (fd_txn_t*) fd_ulong_align_up( (ulong)(udp_payload) + payload_sz, 2UL );
 
-  ulong const * public_key = (ulong const *)(udp_payload + txn->acct_addr_off);
-  ulong const * sig        = (ulong const *)(udp_payload + txn->signature_off);
-  uchar const * msg        = (uchar const *)(udp_payload + txn->message_off);
-  ulong msg_sz             = (ulong)payload_sz - txn->message_off;
+  /* We do not want to deref any non-data field from the txn struct more than once */
+  ushort message_off    = txn->message_off;
+  ushort acct_addr_off  = txn->acct_addr_off;
+  ushort signature_off  = txn->signature_off;
+  
+  if( FD_UNLIKELY( message_off >= payload_sz  || acct_addr_off + FD_TXN_ACCT_ADDR_SZ > payload_sz || signature_off + FD_TXN_SIGNATURE_SZ > payload_sz ) ) {
+    FD_LOG_ERR( ("txn is invalid: payload_sz = %x, message_off = %x, acct_addr_off = %x, signature_off = %x", payload_sz, message_off, acct_addr_off, signature_off ) );
+  }
+  
+  uchar local_sig[FD_TXN_SIGNATURE_SZ]  __attribute__((aligned(8)));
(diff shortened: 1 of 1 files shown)


Fire30:
I am not quite sure the expected behavior for the signature. It is a 64 byte field from what I understand, however it is sometimes dereferenced as a `ulong`. Please verify that I did not mess anything up.

Fire30:
It appears the build is failing due to newly committed code in `during_frag`. Do you want me to include fix in this pull request or open new one?

```
    Finished release [optimized] target(s) in 0.39s
src/app/fdctl/run/tiles/fd_verify.c: In function ‘during_frag’:
src/app/fdctl/run/tiles/fd_verify.c:69:20: error: unused parameter ‘opt_filter’ [-Werror=unused-parameter]
   69 |              int * opt_filter ) {
      |              ~~~~~~^~~~~~~~~~
```

marctrem:
LGTM

ripatel-fd:
> It appears the build is failing due to newly committed code in `during_frag`. Do you want me to include fix in this pull request or open new one?

@Fire30 Sorry about that. It is fixed on main now, so a rebase should fix it.

ripatel-fd:
> I am not quite sure the expected behavior for the signature. It is a 64 byte field from what I understand, however it is sometimes dereferenced as a `ulong`. Please verify that I did not mess anything up.

@Fire30 If you mean the `fd_frag_meta_t::sig` field, it is a 64-bit user-defined field. Typically, it is used for fast filtering. The naming is unfortunate, and it has nothing to do with the 64 byte Ed25519 signatures.

It looks like coincidentally the first 8 bytes of the signature were used as a "hash" in the deduplication algorithm, via the frag meta sig field. We should fix this, ideally by using a permutation like `fd_ulong_hash` with a randomly generated constant. Such that untrusted clients cannot manipulate the server's dedup/tcache.

Fire30:
> @Fire30 If you mean the `fd_frag_meta_t::sig` field, it is a 64-bit user-defined field. Typically, it is used for fast filtering. The naming is unfortunate, and it has nothing to do with the 64 byte Ed25519 signatures.
> 
> It looks like coincidentally the first 8 bytes of the signature were used as a "hash" in the deduplication algorithm, via the frag meta sig field. We should fix this, ideally by using a permutation like `fd_ulong_hash` with a randomly generated constant. Such that untrusted clients cannot manipulate the server's dedup/tcache.

I am talking about the following variable in the old code:

```c
ulong const * sig        = (ulong const *)(udp_payload + txn->signature_off);
```

This variable is passed as the `sig` field to `fd_ed25519_verify` which appears to act on all 64 bytes. 

```
int
fd_ed25519_verify( void const *  msg,
                   ulong         sz,
                   void const *  sig,
                   void const *  public_key,
                   fd_sha512_t * sha ) {
  uchar const * r = (uchar const *)sig;
  uchar const * s = r + 32;
  ...
    if( FD_UNLIKELY( s[31]> 0x10 ) ) return FD_ED25519_ERR_SIG;
  if( FD_UNLIKELY( s[31]==0x10 ) ) {
``` 

If this field is only 64 bits, then that means we are passing the wrong variable to `fd_ed25519_verify`. Or if this is supposed to be 64 bytes, than I think the fast filtering code is using the wrong sig variable?

Edit: Yeah I think we are on same page actually. We were both talking about the txn signature field.

src/app/fdctl/run/tiles/fd_verify.c
```diff
@@ -95,33 +95,57 @@ after_frag( void *             _ctx,
 
   fd_verify_ctx_t * ctx = (fd_verify_ctx_t *)_ctx;
 
+  /* XXX is this check needed? */
```
mmcgee-jump, line 98: It doesn't really matter. We are guaranteed the txn starts before the high watermark, so there is at least FD_TPU_DCACHE_MTU bytes available to read/write. If the txn is too short or there's garbage, it will fail to verify.

src/app/fdctl/run/tiles/fd_verify.c
```diff
@@ -95,33 +95,57 @@ after_frag( void *             _ctx,
...
+  /* We do not want to deref any non-data field from the txn struct more than once */
+  ushort message_off    = txn->message_off;
+  ushort acct_addr_off  = txn->acct_addr_off;
+  ushort signature_off  = txn->signature_off;
+  
+  if( FD_UNLIKELY( message_off >= payload_sz  || acct_addr_off + FD_TXN_ACCT_ADDR_SZ > payload_sz || signature_off + FD_TXN_SIGNATURE_SZ > payload_sz ) ) {
+    FD_LOG_ERR( ("txn is invalid: payload_sz = %x, message_off = %x, acct_addr_off = %x, signature_off = %x", payload_sz, message_off, acct_addr_off, signature_off ) );
+  }
+  
+  uchar local_sig[FD_TXN_SIGNATURE_SZ];
```
mmcgee-jump, line 125: Should be aligned(8)

mmcgee-jump:
It uses the first 8 bytes of the signature as a key for deduplication. We should probably get a cryptographer to opine on if this is OK to do, or if it makes us weaker to some kind of first-bytes preimage attack.

It's a bit worse than a censorship vehicle, because if a user can create a txn that starts with `0x00000000.....` then it can probably corrupt the TCACHE which doesn't expect the NULL (0) value to be inserted. I filed an issue to investigate, see #990

Fire30:
> It's a bit worse than a censorship vehicle, because if a user can create a txn that starts with `0x00000000.....` then it can probably corrupt the TCACHE which doesn't expect the NULL (0) value to be inserted. I filed an issue to investigate, see #990

I believe we should add a check for the NULL value then(can do in this PR as well) because if the QUIC tile is compromised then the txn can contain arbitrary contents.

mmcgee-jump (approved):

PR #992, opened 2023-11-28
ffi: fix fuzz-asan

ripatel-fd:
Introduces a new native_fuzz_asan target.
Updates build.rs to repair the fuzz-asan functionality.
This required undoing a recent change that forced GCC for Firedancer FFI library builds.


ripatel-fd:
@mmcgee-jump Please feel free to re-add a check that enforces GCC for regular FFI builds.

PR #1021, opened 2023-12-01
correctly calculate tpu slot msg

llamb-jump:
This PR fixes a hard crash [here](https://github.com/firedancer-io/firedancer/blob/main/src/app/fdctl/run/tiles/fd_verify.c#L112) and [here](https://github.com/firedancer-io/firedancer/blob/main/src/app/fdctl/run/tiles/fd_verify.c#L126).


ripatel-fd:
Will make sure we regression test this

PR #1034, opened 2023-12-05
wksp: freeze shared workspaces

mmcgee-jump:


ptaffet-jump (commented):
The code is good, but since wksp is one of the core pieces, the change needs a bit more documentation. Can you add:

- (in the doc comment for fd_wksp_new or something) the wksp is created in unfrozen mode
- (in the doc comment for fd_wksp_freeze) it's not possible to unfreeze a wksp and it affects all existing and future joins
- (in the doc comment for all functions that fail when the wksp is frozen) that they require the wksp to be unfrozen

ripatel-fd:
Closing because this is unmergeable at this point

PR #1035, opened 2023-12-06
fdctl: call setpriority to raise thread priority

mmcgee-jump:


src/app/fdctl/run/run.c
```diff
@@ -119,8 +120,10 @@ execve_tile( fd_topo_tile_t * tile,
       cpu_set_t cpu_set[1];
       CPU_ZERO( cpu_set );
       CPU_SET( cpu_idx, cpu_set );
+      if( FD_UNLIKELY( -1==setpriority( PRIO_PROCESS, 0, -19 ) ) ) FD_LOG_ERR(( "setpriority() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
```
ptaffet-jump, line 123: My vote is that these should be warnings, but if you think errors, I'm happy to be convinced

ptaffet-jump (approved):

src/app/fdctl/run/run.c
mmcgee-jump: I don't think this call should ever fail in this context. If we see spurious / unexpected fails we can downgrade to warning.

PR #1055, opened 2023-12-12
quic: increase default concurrent connections

mmcgee-jump:


ptaffet-jump (approved):
Makes sense as a temporary fix. We should probably switch to stream pools in the future to reduce memory consumption.

PR #1063, opened 2023-12-14
fuzz: add rust fuzz targets to highend fuzz artifacts

marctrem:
If the current fuzz build is `highend`, the `publish_to_clusterfuzz.yml` now builds the rust fuzz targets and places them next to the `highend` fuzz artifacts before the invocation to `firedancer-io/clusterfuzz-action`.

### DIFF
--- a/.github/workflows/_publish_to_clusterfuzz.yml
+++ b/.github/workflows/_publish_to_clusterfuzz.yml
@@ -26,6 +26,12 @@ jobs:
         with:
           command: make -j -Otarget fuzz-test
 
+      - uses: firedancer-io/fuzzbot-builder@main
+        name: Build rust fuzz tests
+        with:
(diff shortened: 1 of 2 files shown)


lheeger-jump (approved):

terorie:
Why did this get merged? @lheeger-jump @marctrem Please don't rush things to main

riptl:
Looks like this broke the ClusterFuzz integration for highend target on main.

riptl:
@marctrem The ffi/rust fuzzers are somewhat flakey and are about to be replaced with a more stable approach. I don't think it's a good idea to build them in CI unless we have someone that can maintain them.

It would be nice to have this fuzzer in ClusterFuzz anyways. We previously discussed manually uploading the target to the cluster. Can we still do this or is it better to wait for the new target?

For now I sent a revert PR above to fix the build failure, so the other fuzz uploads work again.

Thanks for working on this!

PR #1065, opened 2023-12-15
Revert "fuzz: add rust fuzz targets to highend fuzz artifacts"

terorie:
Reverts firedancer-io/firedancer#1063

Fixes a CI failure due to a regression


riptl:
@marctrem @mmcgee-jump Please take a look, thanks🙏

lheeger-jump (approved):

PR #1066, opened 2023-12-15
fuzz: fix rust fuzz

marctrem:


ripatel-fd:
@marctrem Thanks for the fix. I'm going to close this since the legacy Rust-based fuzzer is going to be removed shortly in favor of your new solana-industry solution.

PR #1068, opened 2023-12-15
fd_verify: insert verified txs only in pre-dedup tcache

0x0ece:
This PR fixes a shortcoming of the pre-dedup cache in the verify tile, by:
1) query if a tx is in the pre-dedup cache; if so, discard
2) verify ed25519 signature; if not verified, discard
3) insert in the pre-dedup cache; if already in, discard

The following describes an attack that was possible before this patch.

An attacker could front run any user and DoS them:
- User wants to send tx_a
- Attacker sees tx_a, sends an invalid tx_b with the same signature of tx_a (the signature doesn't have to validate, the - attacker can simply take tx_a and truncate it)
- Assumption: attacker can send tx_b before user's tx_a

Expected behavior:
- tx_b is invalid so it should be dropped
- tx_a is valid so it should be processed

Current behavior (before PR):
- tx_b sig is added to the dedup tcache
- therefore, tx_a will be discarded

Behavior (after PR):
- before sig verification, we limit to query the tcache
- after sig verification, we insert the valid tx into the tcache

Note that an attacker could now fully duplicate valid users' txs, forcing the system to run sigverify multiple times on the same exact data. This is accepted as a risk, because we can't really distinguish a valid vs invalid tx before running sigverify.
It's still important to check for dedup during the insert.

### DIFF
--- a/src/app/fdctl/run/tiles/fd_verify.c
+++ b/src/app/fdctl/run/tiles/fd_verify.c
@@ -137,7 +137,8 @@ after_frag( void *             _ctx,
      quick dedup of ha traffic. */
 
   int ha_dup;
-  FD_TCACHE_INSERT( ha_dup, *ctx->tcache_sync, ctx->tcache_ring, ctx->tcache_depth, ctx->tcache_map, ctx->tcache_map_cnt, *(ulong *)local_sig );
+  ulong tcache_map_idx = 0; /* ignored */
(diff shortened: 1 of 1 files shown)


mmcgee-jump (approved):

ripatel-fd:
@0x0ece @mmcgee-jump Could you add a regression test case for this?

PR #1072, opened 2023-12-17
config: enable security profile for native target

ripatel-fd:
All configs except native include `with-security.mk`. This is presumably an oversight from when the security team first added this profile.


mmcgee-jump (approved):

mmcgee-jump:
:'( nice catch

PR #1078, opened 2023-12-18
make: security options now part of base

marctrem:
It does what it says on the tin.

### DIFF
--- a/config/base.mk
+++ b/config/base.mk
@@ -1,3 +1,7 @@
+ifndef MAY_INCLUDE_BASE
+$(error cannot include "base.mk" directly)
+endif
+
 BASEDIR?=build
 
 SHELL:=bash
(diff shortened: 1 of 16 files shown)


ripatel-fd (changes requested):
I think the general approach here is too inflexible. Currently, the build system is maintained in a way to allow for future support for embedded toolchains (like the sbpf LLVM toolchain). This PR probably breaks that capability on platforms that don't support stack protectors or position-independent code. (Those would fail with an "unrecognized flag" error). The flags added here are broadly specific to some architecture+OS combos (currently, only Linux x86_64).

Here are some other ways to skin this cat:
- Add a `linux_x86_64_base.mk` that extends `base.mk`. Change all x86 targets to extend `linux_x86_64_base.mk`. Wire it up in a way that Make errors out if a target accidentally includes `base.mk` directly.
- `configure` script (IMO, this is long overdue anyways)

If it's confirmed that this change works with all existing targets in the config folder, I'll approve it. We can leave this change temporarily until we have a better way.

ripatel-fd:
Closing since it was agreed internally to not force with-security.mk on all targets. It should be enabled for Linux targets only.

PR #1082, opened 2023-12-21
standardized container build based on rhel8.5

gearlbrace:
- creates a build container with all build dependencies that will be used for subsequent builds. First build takes slightly longer as it creates the container.
- spits out fdctl and solana bins to `~/build/machinetype`
- can optionally pass a tag, branch, or release to buildit.sh as `./buildit.sh --tag somebranch`
- optionally specify a machine type with `./buildit.sh --machine machinetype` but defaults to linux_gcc_x86_64

### DIFF
--- a/contrib/container-build/buildit.sh
+++ b/contrib/container-build/buildit.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+OUTDIR=~/build/out
+SCRIPTSDIR=$PWD/scripts
+
+chmod +x "$SCRIPTSDIR/*.sh"
+mkdir -p $OUTDIR
+
+if which podman >/dev/null 2>&1
+  then DOCKER=podman
+elif which docker >/dev/null 2>&1
+  then DOCKER=docker
+else echo "Please install docker or podman"
+  exit 1
+fi
+
+$DOCKER image exists fdbuilder:latest || $DOCKER build -t fdbuilder:latest .
+
+$DOCKER run --rm \
+        --volume "$OUTDIR:/build/out:Z" \
+        --volume "$SCRIPTSDIR:/build/scripts:ro,Z" \
+        fdbuilder:latest \
+        /build/scripts/fdbuild.sh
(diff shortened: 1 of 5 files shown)


contrib/container-build/scripts/fdbuild.sh
```diff
@@ -0,0 +1,20 @@
...
+cd "$GITDIR"
+
+# Allow for a tag, release, or branch
+[ -n "$1" ] && git checkout "$1"
+
+# Currently needed to build and install openssl build libs that support QUIC
+./deps.sh install
+
+# Do the thing
+MACHINE=linux_gcc_x86_64 make -j fdctl solana
```
anwayde, line 17: Can we make `MACHINE` a parameter, with `linux_gcc_x86_64` as the default?

contrib/container-build/buildit.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+OUTDIR=~/build/out
```
anwayde, line 3: Let's get the binaries out to the same directory that it would've gotten out at, that way we also will retain the architecture and compiler information.

contrib/container-build/scripts/fdbuild.sh
gearlbrace: added... with an additional check whether a specific machine type exists

contrib/container-build/buildit.sh
gearlbrace: Done and done

ripatel-fd:
Looks great 

We should probably remove these older images from the repo: https://github.com/firedancer-io/firedancer/tree/main/contrib/docker
As to not confuse users

gearlbrace:
> Looks great
> 
> We should probably remove these older images from the repo: https://github.com/firedancer-io/firedancer/tree/main/contrib/docker As to not confuse users

Absolutely worth discussing. This just spits out binaries, whereas the other build generates containers with the binaries for execution within the container. Different aims in mind.

anwayde (approved):

PR #1096, opened 2023-12-30
simd: use AVX-512 native rotate

ripatel-fd:
AVX-512F comes with native rotate instructions which are slightly
faster than the existing emulated rotate.

Improves SHA-256 batch throughput by 0.6% on Sapphire Rapids.

### DIFF
--- a/src/util/simd/fd_avx512_wwi.h
+++ b/src/util/simd/fd_avx512_wwi.h
@@ -87,8 +87,17 @@ static inline void  wwi_stu( void * m, wwi_t x ) { _mm512_storeu_epi32( m, x ); 
    wwi_rol_vector(x,y) returns wwi( rotate_left (x0,y0), rotate_left (x1,y1), ... )
    wwi_ror_vector(x,y) returns wwi( rotate_right(x0,y0), rotate_right(x1,y1), ... ) */
 
-static inline wwi_t wwi_rol( wwi_t a, int n ) { return wwi_or( wwi_shl ( a, n & 31 ), wwi_shru( a, (-n) & 31 ) ); }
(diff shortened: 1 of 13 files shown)


ripatel-fd:
TODO needs test

ptaffet-jump (dismissed):

src/util/simd/fd_avx512_wwi.h
```diff
@@ -87,8 +87,17 @@ static inline void  wwi_stu( void * m, wwi_t x ) { _mm512_storeu_epi32( m, x );
    wwi_rol_vector(x,y) returns wwi( rotate_left (x0,y0), rotate_left (x1,y1), ... )
    wwi_ror_vector(x,y) returns wwi( rotate_right(x0,y0), rotate_right(x1,y1), ... ) */
 
-static inline wwi_t wwi_rol( wwi_t a, int n ) { return wwi_or( wwi_shl ( a, n & 31 ), wwi_shru( a, (-n) & 31 ) ); }
-static inline wwi_t wwi_ror( wwi_t a, int n ) { return wwi_or( wwi_shru( a, n & 31 ), wwi_shl ( a, (-n) & 31 ) ); }
+static inline wwi_t wwi_rol( wwi_t a, int n ) {
```
ptaffet-jump, line 90: I don't think `wwi_shl` is compatible with non-constant shift amount, so I don't think this branch is useful

ptaffet-jump: This comment applies to all the similar places

ptaffet-jump: Sounds good!

ptaffet-jump (approved):

PR #1098, opened 2023-12-30
BLAKE3 rewrite (wip)

ripatel-fd:
BLAKE3 rewrite with support for batched hashing

- [x] New parallel scheduler
- [x] Port ref, sse41, avx2, avx512 backends
- [x] Faster AVX2 streamlined hashing
- [x] Faster AVX2 batch hashing
- [ ] Get SSE only mode to work
- [ ] Get noarch only mode to work
- [ ] Test AVX512


mmcgee-jump:
Marking as nice to have for 1.5, it improves bank perf

ripatel-fd:
> Marking as nice to have for 1.5, it improves bank perf

I'll revisit it in a few months for 2, where it becomes more important for runtime performance (specifically lthash)

PR #1100, opened 2024-01-02
chacha20: SSE fixes

ripatel-fd:
- Rename fd_chacha20_avx to fd_chacha20_sse
- Fix 4 byte OOB access when loading nonce field
- Merge block index and nonce into a 16 byte region
- Tune alignment and use aligned memory accesses


src/ballet/chacha20/test_chacha20.c
```diff
@@ -55,23 +55,28 @@ bench_chacha20_block( void ) {
...
 
   for( ulong idx=0U; idx<2UL; idx++ ) {
     FD_LOG_NOTICE(( "Benchmarking fd_chacha20_block, run %lu", idx ));
     key[ 0 ]++;
 
     /* warmup */
-    for( ulong rem=100000UL; rem; rem-- ) fd_chacha20_block( &block, &key, block_idx++, &nonce );
+    for( ulong rem=100000UL; rem; rem-- ) {
+      idx_nonce[0]++;
+      fd_chacha20_block( &block, &key, idx_nonce );
```
ptaffet-jump, line 70:
```suggestion
      fd_chacha20_block( block, &key, idx_nonce );
```
Right?

ptaffet-jump (approved):

PR #1102, opened 2024-01-03
nanopb: fix system header includes

ripatel-fd:
Use code generation to include the Firedancer integration header
instead of adding project-wide compiler flags.


jumpsiegel (approved):

jumpsiegel (approved):

ripatel-fd:
Looks like the rebase garbled the commit message

PR #1133, opened 2024-01-08
Build System Cleanup

ripatel-fd:
- ffi: remove fuzz-asan crate feature
- Merges x86-64-clang-flags.mk, x86-64-flags.mk, and x86-64-gcc-flags.mk into with-x86-64.mk
- Removes native_fuzz_asan machine target
- Removes "combi" machine targets
- Fixes a couple bugs and missing with-security.mk
- Move all test scripts to contrib/test
  - Collapse 'test.sh' into 'run_unit_tests.sh'
  - Fix 'run_script_tests.sh' always using 'native' machine regardless of '$MACHINE'
  - Fix incorrect 'LLVM_PROFILE_FILE' in some of the tests (coverage related)
  - Fix incorrect usages of .PHONY
- Rewrite coverage report script into more generic ci_tests.sh
  - Remove duplicate "combi" code
  - Merge all coverage reports into a single big report (previously, there was a combination of 8 different reports)
  - Remove "index" HTML generator that linked to the set of coverage reports; Rather just have the single big report at the root now
  - Fix incorrect invocations of "llvm-cov export" that produced an incomplete lcov report in some cases
  - Move merging of coverage reports from shell scripts to Makefile
  - Collapse coverage.mk into everything.mk
  - Don't include ClusterFuzz corpus in coverage (allows running the script outside of CI)
- Fix design flaw where bin and unit-test targets are not built if "fuzz" feature is enabled
  - Remove "FD_HAS_MAIN" flag
- Rename "fuzz-test" directive to "make-fuzz-test"
- Add target for running integration tests
- Fix hardcoded build dir in integration tests
- Remove unused with-static.mk
- Permit with-llvm-cov.mk for GCC builds with warning
- deps.sh: install LLVM and lcov on Debian
- Run all tests for both pull requests and main commits
- Skip daily coverage report if there are no code changes
- Upload coverage report directly to destination instead of HTTP redirecting;
  Also, delete old coverage report instead of keeping historical versions
- Switch to ci_test.sh script, which allows merging a few workflows
- Remove CodeQL workflow
- Re-introduce exporting to CodeCov

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@ target/
 *.lcov
 *.profraw
 
+# Clang
+.clangd
+
 # Rust
 **/Cargo.lock
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,17 +12,9 @@ behaviors of various common development environments like emacs.
 
 ### Organization
 
-- Please avoid cluttering the repository root
-- `.github`: CI- and repo-related config
-- `config`: GNU Make build system config
-- `contrib`: Miscellaneous code
-- `ffi/rust`: Rust FFI bindings
-- `src`: Main source tree
-  - `app`: Main programs
-  - `ballet`: Protocol-related code (parsers, serializers, cryptography)
-  - `disco`: Tiles
-  - `tango`: Concurrency-related code (message queues, I/O)
-  - `util`: Firedancer standard library
+See [organization.txt](./doc/organization.txt).
+
+Please avoid cluttering the repository root.
 
 ### File Extensions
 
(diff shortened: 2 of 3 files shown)


jumpsiegel (approved):
I am concerned about the fact that the moving of the files into the machines sub-directory are showing up as a delete/add instead of as a move.  This will lead to some annoying merge conflicts when moved around.  Normally git is better at this..

mmcgee-jump:
I don't see where FD_USING_GCC gets set for the non-native targets? Otherwise LGTM

mmcgee-jump (approved):

Issue #150, opened 2023-02-28
pack: add fuzz target for fd_compute_budget_program_parse


ripatel-fd:
The `fd_compute_budget_program_parse` in //src/ballet/txn function takes untrusted binary blobs as an input.

Add a fuzz target suitable for libFuzzer. 

Example fuzz target: https://github.com/firedancer-io/firedancer/blob/main/src/ballet/txn/fuzz_txn_parse.c

Bazel guide: https://github.com/firedancer-io/firedancer/blob/main/bazel/README.md

PR #1134, opened 2024-01-08
Fuzz fd_compute_budget_program_parse

ripatel-fd:
Fixes https://github.com/firedancer-io/firedancer/issues/150


ptaffet-jump (approved):
This is better than nothing, but `fd_compute_budget_program_parse` is stateful, and this fuzzer doesn't exercise any of that statefulness. I couldn't think of an easy, good way to do so, though.

ripatel-fd:
@ptaffet-jump IMHO The fuzzer is useless in its current form. (It correctly pointed out some unaligned access problems though). I added it just as an additional layer of safety in case someone changes the parser in the future and adds some silly bug. Stateful checks are probably best done using CBMC and proof by induction (this approach worked great for things like fd_dcache).

@marctrem The fuzzer can explore the whole target in about a second. Will ClusterFuzz be smart enough to stop exploring this target or will it waste resources on this?

PR #1145, opened 2024-01-10
runtime: switch sysvar checking to use perfect hash table

ptaffet-jump:
Also adds a small snippet of code to the perfect hash table test to help find constants


ripatel-fd:
@ptaffet-jump Once the merge conflict is fixed, this is good to go

Issue #1125, opened 2024-01-05
Move address resolution to net tile


ripatel-fd:
### Problem

Currently, address resolution via netlink/ARP happens in the fd_quic library.
This is problematic, as it broke many tests that rely on simulated networking.
We will also eventually have more tiles serving the network than the "main" quic tile, which will also need address resolution (if in kernel bypass mode).

Finally, operators might want to run Firedancer's networking components with the regular Linux stack. (Because, e.g., their company policy forbids kernel bypass networking) 

### Recommended Change

Move address resolution to the net tile. The net tile is the first tile to handle incoming network traffic and is responsible for internal routing decisions. It is therefore also the appropriate place to resolve addresses for outgoing traffic.

mmcgee-jump:
Fixed by #1162

PR #1162, opened 2024-01-22
move routing/arp to fd_net

nbridge-jump:
Closes https://github.com/firedancer-io/firedancer/issues/1125

### DIFF
--- a/src/app/fdctl/run/tiles/fd_net.c
+++ b/src/app/fdctl/run/tiles/fd_net.c
@@ -6,6 +6,7 @@
 #include "../../../../tango/xdp/fd_xdp.h"
 #include "../../../../tango/xdp/fd_xsk_private.h"
 #include "../../../../util/net/fd_ip4.h"
+#include "../../../../tango/ip/fd_ip.h"
 
 #include <linux/unistd.h>
 
@@ -26,6 +27,7 @@ typedef struct {
   fd_mux_context_t * mux;
 
   uint   src_ip_addr;
+  uchar  src_mac_addr[6];
   ushort allow_ports[ FD_NET_PORT_ALLOW_CNT ];
 
   fd_wksp_t * in_mem;
@@ -36,6 +38,9 @@ typedef struct {
   ulong       out_chunk0;
   ulong       out_wmark;
   ulong       out_chunk;
+
+  fd_ip_t *   ip;
+  long        ip_next_upd;
 } fd_net_ctx_t;
 
 typedef struct {
@@ -44,6 +49,8 @@ typedef struct {
 
   fd_xsk_t * lo_xsk;
   void *     lo_xsk_aio;
+
+  fd_ip_t *  ip;
 } fd_net_init_ctx_t;
 
 FD_FN_CONST static inline ulong
@@ -64,6 +71,7 @@ scratch_footprint( fd_topo_tile_t * tile ) {
     l = FD_LAYOUT_APPEND( l, fd_xsk_align(),     fd_xsk_footprint( FD_NET_MTU, tile->net.xdp_rx_queue_size, tile->net.xdp_rx_queue_size, tile->net.xdp_tx_queue_size, tile->net.xdp_tx_queue_size ) );
     l = FD_LAYOUT_APPEND( l, fd_xsk_aio_align(), fd_xsk_aio_footprint( tile->net.xdp_tx_queue_size, tile->net.xdp_aio_depth ) );
   }
+  l = FD_LAYOUT_APPEND( l, fd_ip_align(), fd_ip_footprint( 0U, 0U ) );
   return FD_LAYOUT_FINI( l, scratch_align() );
 }
 
@@ -160,6 +168,17 @@ before_credit( void * _ctx,
   }
 }
 
+static void
+during_housekeeping( void * _ctx ) {
+  fd_net_ctx_t * ctx = (fd_net_ctx_t *)_ctx;
+  long now = fd_log_wallclock();
+  if( FD_UNLIKELY( now > ctx->ip_next_upd ) ) {
+    ctx->ip_next_upd = now + (long)60e9;
+    fd_ip_arp_fetch( ctx->ip );
+    fd_ip_route_fetch( ctx->ip );
+  }
+}
+
 FD_FN_PURE static int
 route_loopback( uint  tile_ip_addr,
                 ulong sig ) {
@@ -219,6 +238,29 @@ during_frag( void * _ctx,
 }
 
 static void
+send_arp_probe( fd_net_ctx_t * ctx,
+                uint           dst_ip_addr,
+                uint           ifindex ) {
+  uchar          arp_buf[FD_IP_ARP_SZ];
+  ulong          arp_len = 0UL;
+
+  uint           src_ip_addr  = ctx->src_ip_addr;
+  uchar *        src_mac_addr = ctx->src_mac_addr;
+
+  /* prepare arp table */
+  int arp_table_rtn = fd_ip_update_arp_table( ctx->ip, dst_ip_addr, ifindex );
+
+  if( FD_UNLIKELY( arp_table_rtn == FD_IP_SUCCESS ) ) {
+    /* generate a probe */
+    fd_ip_arp_gen_arp_probe( arp_buf, FD_IP_ARP_SZ, &arp_len, dst_ip_addr, fd_uint_bswap( src_ip_addr ), src_mac_addr );
+
+    /* send the probe */
+    fd_aio_pkt_info_t aio_buf = { .buf = arp_buf, .buf_sz = (ushort)arp_len };
+    ctx->tx->send_func( ctx->xsk_aio[ 0 ], &aio_buf, 1, NULL, 1 );
+  }
+}
+
+static void
 after_frag( void *             _ctx,
             ulong              in_idx,
             ulong              seq,
@@ -242,7 +284,69 @@ after_frag( void *             _ctx,
   if( FD_UNLIKELY( route_loopback( ctx->src_ip_addr, *opt_sig ) ) ) {
     ctx->lo_tx->send_func( ctx->xsk_aio[ 1 ], &aio_buf, 1, NULL, 1 );
   } else {
-    ctx->tx->send_func( ctx->xsk_aio[ 0 ], &aio_buf, 1, NULL, 1 );
+    /* extract dst ip */
+    uint dst_ip = fd_uint_bswap( fd_disco_netmux_sig_ip_addr( *opt_sig ) );
+
+    uint  next_hop    = 0U;
+    uchar dst_mac[6]  = {0};
+    uint  if_idx      = 0;
+
+    /* route the packet */
+    /*
+     * determine the destination:
+     *   same host
+     *   same subnet
+     *   other
+     * determine the next hop
+     *   localhost
+     *   gateway
+     *   subnet local host
+     * determine the mac address of the next hop address
+     *   and the local ipv4 and eth addresses */
+    int rtn = fd_ip_route_ip_addr( dst_mac, &next_hop, &if_idx, ctx->ip, dst_ip );
+    if( FD_UNLIKELY( rtn == FD_IP_PROBE_RQD ) ) {
+      /* another fd_net instance might have already resolved this address
+         so simply try another fetch */
+      fd_ip_arp_fetch( ctx->ip );
+      rtn = fd_ip_route_ip_addr( dst_mac, &next_hop, &if_idx, ctx->ip, dst_ip );
+    }
+
+    long now;
+    switch( rtn ) {
+      case FD_IP_PROBE_RQD:
+        /* TODO possibly buffer some data while waiting for ARPs to complete */
+        /* TODO rate limit ARPs */
+        /* TODO add caching of ip_dst -> routing info */
+        send_arp_probe( ctx, next_hop, if_idx );
+
+        /* refresh tables */
+        now = fd_log_wallclock();
+        ctx->ip_next_upd = now + (long)200e3;
+        break;
+      case FD_IP_NO_ROUTE:
+        /* cannot make progress here */
+        break;
+      case FD_IP_SUCCESS:
+        /* set destination mac address */
+        memcpy( ctx->frame, dst_mac, 6UL );
+
+        /* set source mac address */
+        memcpy( ctx->frame + 6UL, ctx->src_mac_addr, 6UL );
+
+        ctx->tx->send_func( ctx->xsk_aio[ 0 ], &aio_buf, 1, NULL, 1 );
+        break;
+      case FD_IP_RETRY:
+        /* refresh tables */
+        now = fd_log_wallclock();
+        ctx->ip_next_upd = now + (long)200e3;
+        /* TODO consider buffering */
+        break;
+      case FD_IP_MULTICAST:
+      case FD_IP_BROADCAST:
+      default:
+        /* should not occur in current use cases */
+        break;
+    }
   }
 
   *opt_filter = 1;
@@ -296,6 +400,10 @@ privileged_init( fd_topo_t *      topo,
                                       tile->net.xdp_aio_depth );
     if( FD_UNLIKELY( !ctx->lo_xsk_aio ) ) FD_LOG_ERR(( "fd_xsk_aio_new failed" ));
   }
+
+  /* init fd_ip */
+  ctx->ip = fd_ip_join( fd_ip_new( FD_SCRATCH_ALLOC_APPEND( l, fd_ip_align(), fd_ip_footprint( 0UL, 0UL ) ),
+                                   0UL, 0UL ) );
 }
 
 static void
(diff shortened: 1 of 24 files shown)


src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -898,14 +815,14 @@ populate_allowed_fds( void * scratch,
                       int *  out_fds ) {
   FD_SCRATCH_ALLOC_INIT( l, scratch );
```
mmcgee-jump, line 816: You can just remove the FD_SCRATCH stuff from here.

src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -898,14 +815,14 @@ populate_allowed_fds( void * scratch,
...
   fd_shred_ctx_t * ctx = FD_SCRATCH_ALLOC_APPEND( l, alignof( fd_shred_ctx_t ), sizeof( fd_shred_ctx_t ) );
+  (void)ctx;
 
   if( FD_UNLIKELY( out_fds_cnt < 3 ) ) FD_LOG_ERR(( "out_fds_cnt %lu", out_fds_cnt ));
```
mmcgee-jump, line 820: out_fds_cnt<2UL

src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -460,7 +420,6 @@ send_shred( fd_shred_ctx_t *      ctx,
 
   memcpy( hdr->eth->dst, dest->mac_addr, 6UL );
 
-  memcpy( hdr->ip4->daddr_c, &dest->ip4, 4UL );
   hdr->ip4->net_id     = fd_ushort_bswap( ctx->net_id++ );
   hdr->ip4->check      = 0U;
   hdr->ip4->check      = fd_ip4_hdr_check( ( fd_ip4_hdr_t const *) FD_ADDRESS_OF_PACKED_MEMBER( hdr->ip4 ) );
```
mmcgee-jump, line 425: Don't need to calculate checksum here? It's going to be wrong anyway right?

src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -128,7 +128,7 @@ typedef struct {
   uint                 src_ip_addr;
   uchar                src_mac_addr[ 6 ];
   ushort               shred_listen_port;
-  fd_ip_t            * ip;
+  uchar const        * shred_signing_key;
```
mmcgee-jump, line 131: Merge isuse?

src/app/fdctl/run/tiles/fd_quic.c
```diff
@@ -749,27 +745,24 @@ populate_allowed_seccomp( void *               scratch,
...
 
 static ulong
 populate_allowed_fds( void * scratch,
                       ulong  out_fds_cnt,
                       int *  out_fds ) {
-  FD_SCRATCH_ALLOC_INIT( l, scratch );
-  fd_quic_ctx_t * ctx = FD_SCRATCH_ALLOC_APPEND( l, alignof( fd_quic_ctx_t ), sizeof( fd_quic_ctx_t ) );
+  (void)scratch;
 
   if( FD_UNLIKELY( out_fds_cnt < 3 ) ) FD_LOG_ERR(( "out_fds_cnt %lu", out_fds_cnt ));
```
mmcgee-jump, line 760: out_fds_cnt<2UL

src/app/fdctl/run/tiles/fd_quic.c
```diff
@@ -749,27 +745,24 @@ populate_allowed_seccomp( void *               scratch,
                           struct sock_filter * out ) {
   FD_SCRATCH_ALLOC_INIT( l, scratch );
```
mmcgee-jump, line 746: Don't need to do FD_SCRATCH_.. here anymore.

src/app/fdctl/run/tiles/fd_quic.c
```diff
@@ -619,8 +618,7 @@ privileged_init( fd_topo_t *      topo,
 
   FD_SCRATCH_ALLOC_INIT( l, scratch );
```
mmcgee-jump, line 619: Don't need FD_SCRATCH_...

src/tango/ip/fd_ip.c
```diff
@@ -348,25 +348,6 @@ fd_ip_route_ip_addr( uchar *   out_dst_mac,
...
-    //  out_dst_mac[3] = 0xffU;
-    //  out_dst_mac[4] = 0xffU;
-    //  out_dst_mac[5] = 0xffU;
-
-    //  *out_ifindex = route_entry->oif;
-
-    //  return FD_IP_BROADCAST;
-    //}
-
-    /* else local unicast */
```
mmcgee-jump, line 368: Maybe keep the comment?

src/app/fdctl/run/tiles/net.seccomppolicy
```diff
@@ -62,3 +80,18 @@ recvmsg: (and (or (eq (arg 0) xsk_fd)
               (eq (arg 1) 0)
               (eq (arg 2) 0)
               (eq (arg 3) MSG_DONTWAIT))
+# In future this call should be moved to the networking tile so it can
```
mmcgee-jump, line 83: Remove

src/app/fdctl/run/tiles/net.seccomppolicy
```diff
@@ -36,13 +41,26 @@ fsync: (eq (arg 0) logfile_fd)
...
 # for entries.  There are two possible XSKs, since we can send packets
 # on a network device or the loopback device.
-sendto: (and (or (eq (arg 0) xsk_fd) 
-                 (eq (arg 0) lo_xsk_fd))
-             (eq (arg 1) 0) 
-             (eq (arg 2) 0)
-             (eq (arg 3) MSG_DONTWAIT)
-             (eq (arg 4) 0) 
-             (eq (arg 5) 0))
+
```
mmcgee-jump, line 44: #

src/app/fdctl/run/tiles/fd_net.c
```diff
@@ -158,6 +167,13 @@ before_credit( void * _ctx,
   for( ulong i=0; i<ctx->xsk_aio_cnt; i++ ) {
     fd_xsk_aio_service( ctx->xsk_aio[i] );
   }
+
+  long now = fd_log_wallclock();
```
mmcgee-jump, line 171: This should be in housekeeping function, not `before_credit` (this is doing fd_log_wallclock every iteration of the runloop)

src/app/fdctl/run/tiles/fd_net.c
```diff
@@ -218,6 +234,29 @@ during_frag( void * _ctx,
   fd_memcpy( ctx->frame, src, sz ); // TODO: Change xsk_aio interface to eliminate this copy
 }
 
+static void
+send_arp_probe( void * _ctx,
+                uint   dst_ip_addr,
+                uint   ifindex ) {
+  fd_net_ctx_t * ctx     = (fd_net_ctx_t *)_ctx;
+  uchar *        arp_buf = ctx->arp_buf;
```
mmcgee-jump, line 242: why not just make this buf a stack local?

src/app/fdctl/run/tiles/fd_net.c
```diff
@@ -218,6 +234,29 @@ during_frag( void * _ctx,
...
+  ulong          arp_len = 0UL;
+
+  uint           src_ip_addr  = ctx->src_ip_addr;
+  uchar *        src_mac_addr = ctx->src_mac_addr;
+
+  /* prepare arp table */
+  int arp_table_rtn = fd_ip_update_arp_table( ctx->ip, dst_ip_addr, ifindex );
+  (void)arp_table_rtn;
```
mmcgee-jump, line 250: Can you add a metric for the result of this update call. And also, the error cases should be handled (don't send the probe).

src/app/fdctl/run/tiles/fd_shred.c
nbridge-jump: no - the memcpy is necessary - it shouldn't have been removed

src/app/fdctl/run/tiles/fd_quic.c
nbridge-jump: nice catch

src/app/fdctl/run/tiles/fd_net.c
nbridge-jump: how often does housekeeping get called?

mmcgee-jump: It's variable, but typically every few millis.

src/app/fdctl/run/tiles/fd_net.c
```diff
@@ -242,7 +281,77 @@ after_frag( void *             _ctx,
...
+          /* cannot make progress here */
+          break;
+        case FD_IP_SUCCESS:
+          /* set destination mac address */
+          memcpy( ctx->frame, dst_mac, 6UL );
+
+          /* set source mac address */
+          memcpy( ctx->frame + 6UL, ctx->src_mac_addr, 6UL );
+
+          ctx->tx->send_func( ctx->xsk_aio[ 0 ], &aio_buf, 1, NULL, 1 );
```
mmcgee-jump, line 336: Do you need to recalculate the checksum?

src/app/fdctl/run/tiles/fd_quic.c
nbridge-jump: `l` is used later

src/app/fdctl/run/tiles/fd_net.c
nbridge-jump: I don't like to have too much on the stack

src/app/fdctl/run/tiles/fd_net.c
nbridge-jump: We want to send the probe anyway. It'll will be quite likely for fd_ip_update_arp_table to fail due to it already having been executed successfully, possibly in a different tile.

src/app/fdctl/run/tiles/fd_net.c
nbridge-jump: the IP checksum doesn't include the ETH header

src/app/fdctl/run/tiles/fd_net.c
```diff
@@ -160,6 +169,17 @@ before_credit( void * _ctx,
   }
 }
 
+static void
+during_housekeeping( void * _ctx ) {
+  fd_net_ctx_t * ctx = (fd_net_ctx_t *)_ctx;
+  long now = fd_log_wallclock();
+  if( now > ctx->ip_next_upd ) {
```
mmcgee-jump, line 176: nit but kbowers style here is like

```suggestion
  if( FD_UNLIKELY( now>ctx->ip_next_upd ) ) {
```

src/app/fdctl/run/tiles/fd_net.c
```diff
@@ -388,6 +512,7 @@ populate_allowed_fds( void * scratch,
   out_fds[ out_cnt++ ] = init_ctx->xsk->xsk_fd;
   if( FD_UNLIKELY( init_ctx->lo_xsk ) )
     out_fds[ out_cnt++ ] = init_ctx->lo_xsk->xsk_fd;
+  out_fds[ out_cnt++ ] = fd_ip_netlink_get(init_ctx->ip)->fd;
```
mmcgee-jump, line 515: another style nit (won't keep making this comment, but it's the style throughout these files at least)

```suggestion
  out_fds[ out_cnt++ ] = fd_ip_netlink_get( init_ctx->ip )->fd;
```

src/app/fdctl/run/tiles/fd_net.c
mmcgee-jump: Am I missing something, or it's 26 bytes? The stack is 8MiB. Having it in the context makes the purpose less clear I think.

src/app/fdctl/run/tiles/fd_net.c
mmcgee-jump: I don't fully follow this. It could be NUD_UNREACHABLE in which case probing is pointless? I presume there are other cases too, if there's already a probe in flight and things like that? As well, we still want a metric for the result code here.

src/app/fdctl/run/tiles/fd_net.c
```diff
@@ -355,6 +476,8 @@ unprivileged_init( fd_topo_t *      topo,
   ulong scratch_top = FD_SCRATCH_ALLOC_FINI( l, 1UL );
   if( FD_UNLIKELY( scratch_top > (ulong)scratch + scratch_footprint( tile ) ) )
     FD_LOG_ERR(( "scratch overflow %lu %lu %lu", scratch_top - (ulong)scratch - scratch_footprint( tile ), scratch_top, (ulong)scratch + scratch_footprint( tile ) ));
+
+  ctx->ip = init_ctx->ip;
```
mmcgee-jump, line 480: Might be better to move this line up a bit, with related code, but at least it shouldn't be after the scratch overflow check which should be last.

src/app/fdctl/run/tiles/fd_net.c
```diff
@@ -405,4 +530,5 @@ fd_tile_config_t fd_tile_net = {
   .scratch_footprint        = scratch_footprint,
   .privileged_init          = privileged_init,
   .unprivileged_init        = unprivileged_init,
+  .mux_during_housekeeping  = during_housekeeping,
```
mmcgee-jump, line 533: nit but stick this line above with the other mux lines

src/app/fdctl/run/tiles/fd_quic.c
mmcgee-jump: Was referring to `ctx` being unused, you can ignore the return code. But this part is wrong anyway and will cause memory corruption, the order of the FD_SCRATCH_* must have the same order for any caller that derives them from `void * scratch` so

* Remove line 132 above `l = FD_LAYOUT_APPEND( l, fd_ip_align(), fd_ip_footprint( 256UL, 256UL ) );`
* Remove this line 620
* Rebase and the fd_alloc is removed as well, so you don't need ctx at all in this function.
* Remove like 654 below `FD_SCRATCH_ALLOC_APPEND( l, fd_ip_align(), fd_ip_footprint( 256UL, 256UL ) );`

src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -885,27 +803,24 @@ populate_allowed_seccomp( void *               scratch,
                           struct sock_filter * out ) {
   FD_SCRATCH_ALLOC_INIT( l, scratch );
```
mmcgee-jump, line 804: Can just remove these 3 scratch lines.

src/app/fdctl/run/tiles/fd_net.c
nbridge-jump: This isn't a style thing. If it's actually unpredictable it's better not to annotate. In this case FD_UNLIKELY is correct

src/app/fdctl/run/tiles/fd_net.c
mmcgee-jump: The scratch overflow check is last in the function so that whatever gets added, it checks the right thing. You could refactor the function to be like
```
init_scratch()
do_work()
check_scratch()
```
This `ctx->ip` would be part of `do_work` so it should go above the scratch check.

src/app/fdctl/run/tiles/fd_net.c
mmcgee-jump: According to Kevin all branches should be annotated, the annotation determines code layout in the binary, it does not affect the branch predictor in the CPU, which can't be hinted.

nbridge-jump: I'll take it up with KB

src/app/fdctl/run/tiles/fd_net.c
nbridge-jump: ctx is in the memory region allocated. You want to modify the region before checking the region is valid?
I'll make it consistent

src/app/fdctl/run/tiles/fd_net.c
nbridge-jump: There is no NUD_UNREACHABLE
ARP sends a broadcast, and hopes for a reply. The failure modes are "can't send a broadcast", and "timed out"
I'll add a metric

mmcgee-jump: It's NUD_NOARP, which does seem like a valid result.

src/tango/quic/fd_quic.h
```diff
@@ -441,7 +440,12 @@ struct fd_quic {
   fd_aio_t aio_rx; /* local AIO */
   fd_aio_t aio_tx; /* remote AIO */
 
-  fd_ip_t * ip;    /* ownership transferred to fd_quic */
+  /* Opaque handles for OpenSSL objects.
+     Owned by fd_quic object (freed on fini).
+     TODO: Instead, provide SSL_CTX object here. */
+
+  void * cert_object;      /* X509 * object */
```
mmcgee-jump, line 447: Merge conflict

mmcgee-jump (changes requested):
Just the one issue about the merge conflict in `fd_quic.h`, otherwise approved

ptaffet-jump (approved):
I reviewed the shred tile integration and verified the merge conflict fix, which was the last question Michael had.

PR #1170, opened 2024-01-26
verify: verify all N signatures of transaction

### LINKED ISSUE #1059: verify: verify all N signatures of transaction

0x0ece:
This PR verifies all signatures in a tx.

We added an interface for batch verification (batch signatures over a single message), but the implementation is the naive one. Efficient implementation will be done in an independent PR.

Fixes #1059.

This includes tests for #1059 and #1068.

### DIFF
--- a/src/ballet/ed25519/fd_ed25519.h
+++ b/src/ballet/ed25519/fd_ed25519.h
@@ -100,6 +100,35 @@ fd_ed25519_verify( void const *  msg,
                    void const *  public_key,
                    fd_sha512_t * sha );
 
+/* fd_ed25519_verify_batch_single_msg verifies a batch of signatures
+   over a single message, according to the ED25519 standard.
+
+   msg is assumed to point to the first byte of a msg_sz byte memory region
+   which holds the message to verify (msg_sz==0 fine, msg==NULL fine if
+   msg_sz==0).
+
+   signatures is assumed to point to the first byte of a memory region
+   which holds the signatures of the message. Each signature is 64-byte long.
+
+   pubkeys is assumed to point to first byte of a memory region
+   that holds the public keys to use to verify these signatures.
+   Each public key is 64-byte long.
+
+   shas is an array of handles of a local join to sha512 calculators.
+
+   batch_sz is the size of signatures, pubkeys and shas.
+   batch_sz must be greater than zero.
+
+   See fd_ed25519_verify for more details. */
+
+int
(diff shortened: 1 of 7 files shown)


src/ballet/txn/fd_txn.h
```diff
@@ -57,12 +57,15 @@
...
 
+   FD_TXN_ACTUAL_SIG_MAX: used to allocate arrays of signatures, pubkeys, etc.
+
    From the spec: "The Solana runtime verifies that the number of signatures
    [stored as a compact-u16] matches the number in the first 8 bits of the
    message header."
    Thus this value must live in the range where compact-u16 and uint8
    representations are identical, hence a max of 127. */
 #define FD_TXN_SIG_MAX               (127UL)
+#define FD_TXN_ACTUAL_SIG_MAX        (12UL)
```
mmcgee-jump, line 68: Where does 12UL come from?

src/ballet/ed25519/fd_ed25519.h
```diff
@@ -100,6 +100,35 @@ fd_ed25519_verify( void const *  msg,
...
+
+   batch_sz is the size of signatures, pubkeys and shas.
+   batch_sz must be greater than zero.
+
+   See fd_ed25519_verify for more details. */
+
+int
+fd_ed25519_verify_batch_single_msg( uchar const   msg[], /* msg_sz */
+                                    ulong const   msg_sz,
+                                    uchar const   signatures[ static 64 ], /* 64 * batch_sz */
```
mmcgee-jump, line 127: Is the `static 64` correct if it's 64 * batch_sz ?

0x0ece: static n (by spec) means at least n, so yes, it's correct. it's basically only saying "it must contain at least 1 signature".

src/ballet/txn/fd_txn.h
0x0ece: It's in the comment - given the max tx size, you can only fit 12x 64-bit signatures.
I personally think it's better to keep the constants close, so we don't risk to change one but not the other.

src/ballet/ed25519/fd_ed25519.h
mmcgee-jump: :+1:

src/ballet/txn/fd_txn.h
mmcgee-jump: Ah, didn't see that thanks

PR #1181, opened 2024-01-29
Fix fddev/txn.c (broken during signer changes)

nbridge-jump:
Calls to the new signing API were added to `fd_quic_new_anonymous` in  `fd_quic_test_helpers.c`
`fddev/txn.c`, however, doesn't use `fd_quic_new_anonymous`, so `sign_fn` was not being set, resulting in a segfault


mmcgee-jump (approved):
Sorry! Thanks

PR #1187, opened 2024-01-30
chore: move fd_fec_resolver into hula

lidatong:


ptaffet-jump:
Can you hang on before merging this so I can understand the goal better? I'm not opposed to `hula` but I don't know that the FEC resolver belongs there.

PR #1191, opened 2024-01-30
tls: use upstream OpenSSL

ripatel-fd:
We no longer depend on the QUIC-specific functionality of OpenSSL thanks to fd_tls.

We can therefore switch to upstream OpenSSL. This allows us to bring OpenSSL bug fixes to Firedancer faster.


mmcgee-jump:
What are the consumers now? Some test clients? I guess soon to be the metrics event writer.

mmcgee-jump (approved):

PR #1192, opened 2024-01-31
Misc changes to fix Clang AVX512 compilation

### LINKED ISSUE #1188: fddev does not compile with Clang
**Problem**

fddev-specific code only compiles only on GCC, not Clang.

**Suggested Fix**

- [ ] Fix the broken code
- [ ] Consider compiling fddev with Clang in CI to prevent regressions

ptaffet-jump:
Clang and GCC are pickier about `_mm512_rol_epi32` having a compile-time constant argument.  Richie's fix using `__builtin_constant_p` works fine with GCC but not with Clang. The best way to address it is to split into e.g. `wwi_rol` and `wwi_rol_variable`. I modified the tests correspondingly.

There was also an unused variable Clang did not like in the PoH tile, and an incorrect alignment argument in `fd_r43x6_ge.c`.

The reason all these issues have gone undetected is that CI was not building with `FD_HAS_AVX512`, so I've fixed that as well.

Fixes #1188.

### DIFF
--- a/src/app/fdctl/run/tiles/fd_poh.c
+++ b/src/app/fdctl/run/tiles/fd_poh.c
@@ -1097,15 +1097,9 @@ after_frag( void *             _ctx,
   }
 
   ulong txn_cnt = (*opt_sz-sizeof(fd_microblock_trailer_t))/sizeof(fd_txn_p_t);
+  fd_txn_p_t * txns = (fd_txn_p_t *)(ctx->_txns);
   ulong sanitized_txn_cnt = 0UL;
-  ulong executed_txn_cnt = 0UL;
-  for( ulong i=0; i<txn_cnt; i++ ) {
-    fd_txn_p_t * txn = (fd_txn_p_t *)(ctx->_txns + i*sizeof(fd_txn_p_t));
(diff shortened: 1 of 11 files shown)


src/app/fdctl/run/tiles/fd_poh.c
```diff
@@ -1098,13 +1098,11 @@ after_frag( void *             _ctx,
 
   ulong txn_cnt = (*opt_sz-sizeof(fd_microblock_trailer_t))/sizeof(fd_txn_p_t);
   ulong sanitized_txn_cnt = 0UL;
-  ulong executed_txn_cnt = 0UL;
   for( ulong i=0; i<txn_cnt; i++ ) {
     fd_txn_p_t * txn = (fd_txn_p_t *)(ctx->_txns + i*sizeof(fd_txn_p_t));
     if( FD_UNLIKELY( !(txn->flags & FD_TXN_P_FLAGS_SANITIZE_SUCCESS) ) ) continue;
     sanitized_txn_cnt++;
     if( FD_UNLIKELY( !(txn->flags & FD_TXN_P_FLAGS_EXECUTE_SUCCESS) ) ) continue;
```
mmcgee-jump, line 1105: Can remove this line as well now

ptaffet-jump:
I think the version of gcc in CI is too old for proper avx512 support

mmcgee-jump (approved):

PR #1193, opened 2024-01-31
introduce the `development.gossip.allow_private_addr` configuration item

marctrem:
From the new documentation:
>  Under normal operating conditions, a validator should always reach out to a host located on the public internet.
> This configuration item allows Firedancer to gossip with nodes located on a private internet (rfc1918).

### DIFF
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -907,3 +907,7 @@ dynamic_port_range = "8900-9000"
         interface1_mac = "52:F1:7E:DA:2C:E1"
         # IP address of the second network namespace.
         interface1_addr = "198.18.0.2"
+    [development.gossip]
+        # Under normal operating conditions, a validator should always reach out to a host located on the public internet.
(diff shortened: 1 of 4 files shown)


src/app/fdctl/config/default.toml
```diff
@@ -895,3 +895,7 @@ dynamic_port_range = "8900-9000"
         interface1_mac = "52:F1:7E:DA:2C:E1"
         # IP address of the second network namespace.
         interface1_addr = "198.18.0.2"
+    [development.gossip]
+        # Under normal operating conditions, a validator should always reach out to a host located on the public internet.
+        # This configuration item allows Firedancer to gossip with nodes located on a private internet (rfc1918).
+        allow_private_addr = false
```
mmcgee-jump, line 901: Can you make this comment say what solana command line flag it toggles (see rest of this file for examples of the language.)

```suggestion
        allow_private_address = false
```

marctrem:
@mmcgee-jump ptal

ripatel-fd:
For the future, ideally please squash

PR #1195, opened 2024-01-31
A couple bug fixes

llamb-jump:
config_t config can be too large for the stack

zero initialize args_t args in fddev/main.c


nbridge-jump (approved):
LGTM

PR #1215, opened 2024-02-02
uninitialized variable args is causing a segfault with the command fdctl configure init sysctl

llamb-jump:


mmcgee-jump:
Where was the actual segfault? This doesn't address the bug in `configure.c` -- since it should set the args values correctly. @llamb-jump

nbridge-jump:
After reading the code, my guess is that `configure_cmd_args` doesn't add a `NULL` terminator to the `stages` array, and `configure_cmd_fn` and others use `NULL` to exit their loop.
I'd rather see the `NULL` added (when necessary) explicitly in `configure_cmd_args`

PR #1222, opened 2024-02-03
Fix U.B. with pointer additions

ripatel-fd:
Pointer arithmetic has some special assumptions.  For example,
adding a non-zero offset to a pointer value is assumed to never
result in a zero pointer.  (Or more generally, pointers cannot
overflow in C.)  Fixes instances of such U.B. by replacing them
with regular integer arithmetic.


kbowers-jump:
Very irritating that, in a world crawling with bugs created by code that doesn't test for safe memory accesses, the new compiler "safety" hotness is to complain about code that tests for safe memory accesses.  Sigh.

kbowers-jump (approved):

Issue #148, opened 2023-02-28
util: improve fd_cstr test coverage


ripatel-fd:
The following functions in //src/util/cstr:test_cstr.c are not covered by unit tests yet:

- fd_cstr_casecmp
- fd_cstr_to_ulong_octal
- fd_cstr_append_printf
- fd_cstr_append_cstr
- fd_cstr_append_cstr_safe
- fd_cstr_hash
- fd_cstr_hash_append

mahmudsudo:
can i take on this ?

Akhil-Rawat:
@ripatel-fd can i get this task assigned .

PR #1237, opened 2024-02-05
Increase test coverage for `fd_cstr`

AbhinavMir:
Potential fix for https://github.com/firedancer-io/firedancer/issues/148


ripatel-fd:
The latest coverage report can be found here. We actually do have nearly perfect test coverage for fd_cstr at this point, so the issue description in #148 is a little outdated. It is true though that these functions have no dedicated unit tests yet. 
https://firedancer-ci-artifacts-public.storage.googleapis.com/cov/util/cstr/index.html

ripatel-fd:
@AbhinavMir It looks like we already have a fix of this in the works that we didn't publish yet. Sorry about that it.

AbhinavMir:
No problem!

Issue #1210, opened 2024-02-02
Support bool flags in fd_env


ripatel-fd:
**Problem**

Many of our CLI tools use boolean flags, like "verify". Our argument syntax would be either `--verify 0` or `--verify 1`.
When passing `--verify true` though, the `true` string fails to parse and defaults to `0`. 
I believe many users would find this confusing.

The established convention for argument handling is GNU these days, which would accept `--verify` (no argument, implying true), `--verify false` or `--verify true`. We can leave the old numeric style in for backwards compatibility. We should probably abort with an error if the argument is not recognized though, since the "failed to parse" value and "false" are indistinguishable by return code.

**Suggested Fix**

- [ ] Add an `int fd_env_strip_cmdline_bool( ... )` function
- [ ] Update usages

ripatel-fd:
@kbowers-jump Does the above sound reasonable to you?

AbhinavMir:
OK, I can tackle this! Thanks Richard!

AbhinavMir:
@ripatel-fd quick question: went thru the source code, couldn't really figure out where this goes. Any hints on that? 😅

ripatel-fd:
@AbhinavMir The right place to add this is here https://github.com/firedancer-io/firedancer/blob/main/src/util/env/fd_env.h (fd_env.h for the function declaration, fd_env.c for the implementation, test_env.c for the unit tests)

AbhinavMir:
Gotcha, what would a failure mode look like? `fprintf(stderr, "Error: Invalid value for %s. Use true/false or 1/0.\n", key);` sounds good? Also, is there a `make` command to format it according to Firedancer's current style? I couldn't specifically find a clang formatter or such. Thx!

I've added my current progress below, but it isn't formatted correctly. Later, I will squash and rebase for clean commit history!

ripatel-fd:
https://github.com/firedancer-io/firedancer/blob/59f1861e077d543d2e72c366a3d06971bd634680/src/util/env/fd_env.h#L65

PR #1253, opened 2024-02-09
add support for cmdline bool support

AbhinavMir:
potential fix for https://github.com/firedancer-io/firedancer/issues/1210

### DIFF
--- a/src/util/env/fd_env.c
+++ b/src/util/env/fd_env.c
@@ -11,6 +11,7 @@
 #include "fd_env.h"
 
 #include <stdlib.h>
+#include <stdio.h>
 
 #define FD_ENV_STRIP_CMDLINE_IMPL( T, what )                                   \
 T                                                                              \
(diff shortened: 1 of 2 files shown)


ripatel-fd:
Thanks @AbhinavMir. Could you add tests to `test_env` and reformat to match the code style? 
You can probably also remove `fd_env_strip_cmdline_contains` and update usages. 
`fd_env_strip_cmdline_bool` a more powerful version of `fd_env_strip_cmdline_contains`.

AbhinavMir:
Hi @ripatel-fd , sorry I forgot to add tests! Doing so now.

AbhinavMir:
Hi @ripatel-fd , sorry I forgot to address the tests! Will do now. Also: Pasting my comments from the issues tab, I think it flew under the radar.

"What would a failure mode look like? `fprintf(stderr, "Error: Invalid value for %s. Use true/false or 1/0.\n", key);` sounds good? 

Also, is there a make command to format it according to Firedancer's current style? I couldn't specifically find a clang formatter or such. Thx!"

ripatel-fd:
> Hi @ripatel-fd , sorry I forgot to address the tests! Will do now. Also: Pasting my comments from the issues tab, I think it flew under the radar.

Great!

> "What would a failure mode look like? `fprintf(stderr, "Error: Invalid value for %s. Use true/false or 1/0.\n", key);` sounds good?

Good question. `fd_env` is declared before `fd_log`, so we can't really do any error reporting. I would probably just default to "false" for now. 

Although a bit verbose, we could also offer an API as such:

```c
char const * flag_cstr = fd_env_strip_cmdline_bool( ... );
/* At this point flag_cstr contains
  1) "false" if not specified by the user, or if user explicitly specified `--flag false` or `--flag 0`
  2) "true" if specified without any argument (`--flag`), or if user explicitly specified `--flag true` or `--flag 1` 
  3) NULL if failed to parse */
if( FD_UNLIKELY( !flag_cstr )  FD_LOG_ERR(( "invalid --flag ..." ));
int flag = fd_cstr_to_bool( flag_cstr );
``` 

> Also, is there a make command to format it according to Firedancer's current style? I couldn't specifically find a clang formatter or such. Thx!"

Unfortunately, we don't support an automated formatter.

This is looking pretty good now. I'm happy to do some belt-sanding for the formatting, API, and testing code. Would you be okay with me merging your code manually, and then pushing changes on top?

ripatel-fd:
Closing this since we have a `fd_env_strip_cmdline_contains` function now. Thank you for the contribution.

PR #1254, opened 2024-02-09
Fix ubsan errors killing QUIC harness in CF

nlgripto:
`src/ballet/ed25519/avx/fd_ed25519_fe.c:682:36: runtime error: left shift of negative value -60211` 
This ubsan complaint was (likely) killing the harness in CLusterFuzz.


ripatel-fd:
Looks like we are missing this macro for UBSan:

https://github.com/firedancer-io/firedancer/blob/10cd42ed1c3e2a3746c01ab4d413adaf665566eb/src/util/sanitize/fd_asan.h#L43-L47

marctrem:
> These are no longer arithmetic right shifts. Consider FD_FN_NO_UBSAN instead or silence UBSan warnings project-wide, as we are not yet UBSan compatible.

I partially agree with this assessment. Conceptually those might not be arithmetic right shifts anymore but still, to get the desired result, you need to use the machine's arithmetic right shift. This should apparently not be done with signed values.

Maybe some other machine implements right shift on signed values in another way that's unexpected? I know that our static asserts would catch that but still, I think the change is worth it.

That's just my 2 cents. Definitely feel free to dismiss.

ripatel-fd:
Right. We don't understand the implications of this change, and it potentially affects cryptography. Therefore we don't just change it to make some linter happy.

nlgripto:
kb:
>Now to deal with such from UBSan and what not that are dealing with busted language specs (that all too often are being used as an excuse by bad compiler writers to drive a mack truck through otherwise sane code), there are some operations in bits that should fix this.
>In bits there is a fd_long_shift_left (and other such) to deal with stuff like this.
>Namely:
FD_FN_CONST static inline T    fd_##T##_shift_left  ( T x, int n      ) { return (T)fd_##UT##_shift_left  ( (UT)x, n ); }        \
And this in turn also handles large n (another area of UB behavior in the standard):
FD_FN_CONST static inline T    fd_##T##_shift_left  ( T x, int n        ) { return (T)(((n>(w-1)) ? ((T)0) : x) << (n&(w-1)));   } \
In principle, since n here is a compile time constant all this should boil down to what the original code was doing.  I don't like needing to throw this much obtuseness and effort for the compiler to do so what was already fine.  But such is the modern way and we make do with what we have.

src/ballet/ed25519/avx/fd_ed25519_fe.c
```diff
@@ -1,6 +1,6 @@
 #include "../fd_ed25519_private.h"
 #include "fd_ed25519_fe_avx.h"
-
+#include "../../../util/bits/fd_bits.h"
```
kbowers-jump, line 3: Unnecessary (already in build space)

src/ballet/ed25519/avx/fd_ed25519_fe.c
```diff
@@ -671,29 +671,29 @@ fd_ed25519_fe_mul121666( fd_ed25519_fe_t *       h,
...
-  long carry9 = (h9 + (1L << 24)) >> 25;
-  long carry1 = (h1 + (1L << 24)) >> 25;
-  long carry3 = (h3 + (1L << 24)) >> 25;
-  long carry5 = (h5 + (1L << 24)) >> 25;
-  long carry7 = (h7 + (1L << 24)) >> 25;
+  long carry9 = (h9 + fd_long_shift_left(1L, 24)) >> 25;
+  long carry1 = (h1 + fd_long_shift_left(1L, 24)) >> 25;
+  long carry3 = (h3 + fd_long_shift_left(1L, 24)) >> 25;
+  long carry5 = (h5 + fd_long_shift_left(1L, 24)) >> 25;
+  long carry7 = (h7 + fd_long_shift_left(1L, 24)) >> 25;
```
kbowers-jump, line 678: Mostly indifferent to this but it doesn't matter to the UBSAN.  And the 1L << 25 didn't get the treatment below.  So, would probably keep this as (1L << 24).  Or in lieu of that, use the fortified shift_left on the 1<<25.

src/ballet/ed25519/avx/fd_ed25519_fe.c
```diff
@@ -671,29 +671,29 @@ fd_ed25519_fe_mul121666( fd_ed25519_fe_t *       h,
...
   long carry8 = (h8 + (1L << 25)) >> 26;
 
-  h1 += carry0;       h0 -= carry0 << 26;
-  h3 += carry2;       h2 -= carry2 << 26;
-  h5 += carry4;       h4 -= carry4 << 26;
-  h7 += carry6;       h6 -= carry6 << 26;
-  h9 += carry8;       h8 -= carry8 << 26;
+  h1 += carry0;       h0 -= fd_long_shift_left(carry0, 26);
+  h3 += carry2;       h2 -= fd_long_shift_left(carry2, 26);
+  h5 += carry4;       h4 -= fd_long_shift_left(carry4 , 26);
```
kbowers-jump: Minor cleanup (but know you've already mentioned this has been fixed).

kbowers-jump (changes requested):

kbowers-jump (approved):

ripatel-fd:
Looks like GitHub messed up the rebase. I rebased this PR to here: https://github.com/firedancer-io/firedancer/pull/1260

PR #1260, opened 2024-02-15
ed25519: fix out-of-bounds shifts

ripatel-fd:
Fixes an issue with shift masks greater than integer bit count, which is U.B. in C.  Detected by UBSan.


nlgripto:
perfect

mmcgee-jump (approved):

PR #1262, opened 2024-02-15
util: add fd_rdrand for x86

ripatel-fd:


mmcgee-jump:
Caller needs to be changed to check the return code.

ripatel-fd:
> Caller needs to be changed to check the return code.

Which caller? The PR only adds the method and a test

mmcgee-jump:
Huh, must have looked at the wrong PR.

ripatel-fd:
> Huh, must have looked at the wrong PR.

No sorry for gaslighting you. I removed the offending commit a few days ago.

PR #1265, opened 2024-02-16
test: ed25519 sigverify wycheproof

marctrem:
I still need to cleanup style.


mmcgee-jump:
Please no submodules without a good justification. Is it absolutely necessary? Can we have a test script which fetches wycheproof, or run this out of repo?

PR #1269, opened 2024-02-19
Remove X.509 parser

ripatel-fd:
Once Solana v1.18 activates on mainnet, we won't need a certificate parser any longer.


mmcgee-jump:
Nice

mmcgee-jump (approved):

PR #1270, opened 2024-02-19
Added processing time metrics for the main tiles

nbridge-jump:


ripatel-fd:
As far as I know, Michael added 'stem' metrics that achieve this. So I'm closing this for now.

Issue #1271, opened 2024-02-20
Bring back ClusterFuzz code coverage report


ripatel-fd:
**Problem**

The current coverage report only shows files that are covered by all the tests checked into main (including the fuzz harnesses).

We have a fuzz corpus on ClusterFuzz with additional coverage.

**Suggested Fix**

Generate a second coverage report just from ClusterFuzz and upload it to codecov

PR #1274, opened 2024-02-20
ci: run daily ClusterFuzz report

ripatel-fd:


lidatong (approved):

ripatel-fd:
Closes https://github.com/firedancer-io/firedancer/issues/1271

PR #1277, opened 2024-02-21
Add option to optimize treap iteration

ptaffet-jump:
Pack was spending a substantial amount of time iterating through a treap in sorted order.  The complexity of this can be reduced from O(log(n)) to O(1) by threading a linked list through the nodes.  Iterating with this linked list also eliminates hard-to-predict branches and cache misses.  Maintaining this linked list doesn't affect the complexity of the other treap operations, but it does increase the constant factor slightly and the space requirements somewhat significantly (an extra two indices), so I made this optional.


mmcgee-jump (approved):
Stamping to unblock, looks ok although I don't understand the treap code. kbowers can review out of band.

PR #1280, opened 2024-02-21
reasm: don't use sig field

mmcgee-jump:


ripatel-fd:
Change looks correct, not sure why it's failing. Will investigate later

ripatel-fd:
Thanks for the patch. I'm closing this and will fix the issues in a separate PR.

PR #1289, opened 2024-02-27
fdctl/fd_poh: fix segfault when trying to update metrics from Rust thread

ptaffet-jump:
Rust can call the `fd_ext` functions from random threads with no `FD_TL` variables. In particular, in rare cases, they might try to update metrics, which needs the thread local `fd_metrics_tl`.

There's no good way to intercept the creation of these threads by the Rust code, so the two choices become to defer the metrics until later or to have the Rust thread assume the identity of the PoH tile when they call into any `fd_ext` functions.  Deferring the metrics until the end of the housekeeping loop seemed cleanest.

Although the deferred metrics are written from one thread and read from another, there's the intervening lock release, with the necessary fences, so it won't race (not that it really matters).

### DIFF
--- a/src/app/fdctl/run/tiles/fd_poh.c
+++ b/src/app/fdctl/run/tiles/fd_poh.c
@@ -393,7 +393,7 @@ typedef struct {
   /* We need to tell pack when we are done with a microblock so that it
      can reschedule (unlock) the accounts that were in it. */
   ulong * pack_busy[ 32 ];
-  
+
   fd_sha256_t * sha256;
   void * bmtree;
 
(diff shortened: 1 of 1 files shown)


src/app/fdctl/run/tiles/fd_poh.c
```diff
@@ -519,13 +541,13 @@ fd_ext_poh_initialize( ulong         hashcnt_duration_ns, /* See clock comments
...
-  
+
   ctx->hashcnt             = tick_height*hashcnt_per_tick;
   ctx->last_hashcnt        = ctx->hashcnt;
   ctx->reset_slot_hashcnt  = ctx->hashcnt;
-  ctx->reset_slot_start_ns = fd_log_wallclock();
+  ctx->reset_slot_start_ns = fd_log_wallclock(); /* safe to call from Rust */
 
-  fd_memcpy( ctx->hash, last_entry_hash, 32UL );
+  memcpy( ctx->hash, last_entry_hash, 32UL );
```
ptaffet-jump, line 550: I think memcpy is faster when the size is small and known at compile time because the compiler can do a better job optimizing it

anwayde (approved):

mmcgee-jump (approved):

PR #1290, opened 2024-02-28
`pack`: allow fuzz_compute_budget_program_parse to build

nlgripto:


ptaffet-jump (approved):

ptaffet-jump:
Could you actually edit the commit message to make it more descriptive? I don't think that will affect the approval

PR #1291, opened 2024-02-28
 fdctl: move getpid1 to sandbox

mmcgee-jump:


ptaffet-jump (dismissed):
The only real issue is the pid nul termination issue, but PIDs are no more than 7 characters, so I think it's actually fine.  I'd still fix it so it doesn't look wrong though.

src/util/sandbox/fd_sandbox.c
```diff
@@ -312,3 +312,16 @@ fd_sandbox_alloc_protected_pages( ulong page_cnt,
   return middle_pages;
 #undef PAGE_SZ
 }
+
+int
+fd_sandbox_getpid( void ) {
```
ptaffet-jump, line 317: I appreciate the new name!

ptaffet-jump (approved):

PR #1293, opened 2024-02-28
fdctl: move key loading to disco

mmcgee-jump:


src/disco/keyguard/fd_keyload.c
```diff
@@ -0,0 +1,91 @@
...
+
+/* Expects that key[i] is writable for i in [0, 1600). */
+static inline uchar *
+read_key( char const * key_path,
+          uchar      * key       ) {
+  int key_fd = open( key_path, O_RDONLY );
+  if( FD_UNLIKELY( key_fd==-1 ) ) {
+    if( FD_UNLIKELY( errno==ENOENT ) ) {
+      FD_LOG_ERR((
+          "The [consensus.identity_path] in your configuration expects a "
```
ptaffet-jump, line 20: Hmmm this error message is specific to the old use

mmcgee-jump: Not sure how to fix this, I added some more comments. We want a good error message, but it's going to be excruciating to plumb an error message all the way from fdctl to this function, and it's virtually only going to be used in that case anyway.

ptaffet-jump (dismissed):

src/disco/keyguard/fd_keyload.c
mmcgee-jump: The problem is, the tiles doing the key loading live in disco too, so you can't even put the error there. The return code needs to be from all the `privileged_init` and _then_ fdctl stringizes it.

ptaffet-jump (approved):

PR #1294, opened 2024-02-29
net: fix bad layout for net xsk

mmcgee-jump:


ptaffet-jump (approved):
I checked the other `FD_LAYOUT_APPEND` calls in this file and they're all good. I wonder if this is the origin of the mysterious `xsk 0` issue.

PR #1295, opened 2024-02-29
fdctl: move topo to disco

mmcgee-jump:


src/app/fdctl/run/tiles/fd_net.c
```diff
@@ -360,6 +360,8 @@ privileged_init( fd_topo_t *      topo,
 
   FD_SCRATCH_ALLOC_INIT( l, scratch );
   fd_net_init_ctx_t * ctx = FD_SCRATCH_ALLOC_APPEND( l, alignof( fd_net_init_ctx_t ), sizeof( fd_net_init_ctx_t ) );
+  FD_SCRATCH_ALLOC_APPEND( l, alignof( fd_net_ctx_t ), sizeof( fd_net_ctx_t ) );
```
ptaffet-jump, line 363: Watch this double change in the rebase

src/app/fdctl/run/tiles/tiles.h
```diff
@@ -92,15 +92,28 @@ typedef struct {
...
 
 FD_FN_CONST fd_tile_config_t *
-fd_topo_tile_to_config( fd_topo_tile_t * tile );
+fd_topo_tile_to_config( fd_topo_tile_t const * tile );
+
+static inline ulong
+fdctl_tile_align( fd_topo_tile_t const * tile ) {
+  fd_tile_config_t * config = fd_topo_tile_to_config( tile );
+  if( FD_LIKELY( config->scratch_align ) ) return config->scratch_align();
+  return 0UL;
```
ptaffet-jump, line 108: Default align of 0 seems like a bad idea. This should probably just error, but if you need a default 1, 8, 32 or something seem much better.

src/disco/topo/fd_topo.c
```diff
@@ -318,14 +314,14 @@ fd_topo_workspace_fill( fd_topo_t *      topo,
...
 
-      fd_tile_config_t * config = fd_topo_tile_to_config( tile );
-      if( FD_LIKELY( config->scratch_align ) ) {
+      ulong align = tile_align( tile );
+      if( FD_LIKELY( align ) ) {
         ulong current_wksp_offset = fd_ulong_align_up( fd_wksp_private_data_off( part_max ), fd_topo_workspace_align() ) + scratch_top;
 
-        ulong desired_scratch_align = config->scratch_align();
+        ulong desired_scratch_align = tile_align( tile );
         ulong pad_to_align = desired_scratch_align - (current_wksp_offset % desired_scratch_align);
```
ptaffet-jump, line 322: Can you switch this to `fd_ulong_align_up`? If `current_wksp_offset` is a multiple of `desired_scratch_align`, then this adds `desired_scratch_align` instead of 0. That's fine, but not sure it is intentional.

src/disco/topo/fd_topo.c
```diff
@@ -318,14 +314,14 @@ fd_topo_workspace_fill( fd_topo_t *      topo,
       fd_topo_tile_t * tile = &topo->tiles[ i ];
       if( FD_LIKELY( tile->wksp_id!=wksp->id ) ) continue;
 
-      fd_tile_config_t * config = fd_topo_tile_to_config( tile );
-      if( FD_LIKELY( config->scratch_align ) ) {
+      ulong align = tile_align( tile );
+      if( FD_LIKELY( align ) ) {
```
ptaffet-jump, line 318: Oh now I see this align of 0 means no memory gets allocated. Are there any tiles with no footprint? I think if align is 1 and footprint is 0, you can get rid of this branch and it won't allocate any space for the tile.

src/app/fdctl/mem.c
```diff
@@ -1,10 +1,11 @@
 #include "fdctl.h"
 
+#include "run/run.h"
+
 void
 mem_cmd_fn( args_t *         args,
             config_t * const config ) {
   (void)args;
 
-  fd_topo_fill( &config->topo, FD_TOPO_FILL_MODE_FOOTPRINT );
```
ptaffet-jump, line 8: I'm not sure why these aren't needed anymore, but I'll trust you on it. It would have failed pretty quickly if they were still needed.

mmcgee-jump: I moved the footprint fill into config.c when it creates the topology. This will change a bit more in another incoming change.

ptaffet-jump (approved):

PR #1296, opened 2024-02-29
ed25519: minor tweaks avx512

0x0ece:
Making some minor changes in avx512 arithmetic:
1. (almost irrelevant) compute 2*Z1 using add instead of mul (no change in the num of instr)
2. modified sub: r = a - b, to compute it as r = a + (p-b), where p-b is positive. This allows to replace fold_singed with fold_unsigned

These came up during a refactor, and (will) allow for more similar logic between ref and avx512 implementations.

No major change on performance. Small improvement across the board, minor loss (<2%) on ed25519_verify, on gcc.
I don't think this is a big issue since the refactor will improve ed25519_verify.
```
                                  gcc old &   gcc new & clang old & clang new
fd_ed25519_public_from_private   8275.419 &  8171.789 & 10235.025 &  9850.557 ns/call
fd_ed25519_sign(128)             9563.785 &  9461.095 & 11581.620 & 11089.120 ns/call
fd_ed25519_sign(1024)           12620.123 & 12512.965 & 14621.085 & 14165.299 ns/call
fd_ed25519_verify(good 128)     24636.156 & 25082.158 & 27513.482 & 26908.014 ns/call
fd_ed25519_verify(good 1024)    26699.027 & 27083.375 & 29546.553 & 29209.562 ns/call
```


src/ballet/ed25519/avx512/fd_r43x6_inl.h
```diff
@@ -152,12 +152,18 @@ FD_PROTOTYPES_BEGIN
 /* FD_R43X6_QUAD_LANE_SUB_FAST does:
      D = [ (imm0 ? (PX-QX) : SX) (imm1 ? (PY-QY) : SY) (imm2 ? (PZ-QZ) : SZ) (imm3 ? (PT-QT) : ST) ]
    imm* should be in [0,1]. */
-
 #define FD_R43X6_QUAD_LANE_SUB_FAST( D, S, imm0,imm1,imm2,imm3, P, Q ) do { \
```
kbowers-jump, line 155: My guess here is this modification is subtracting Q from a "big zero" such that reasonable range limbs, the subtraction doesn't wrap and the add of the resulting "-Q" to "P" results in reasonable range output limbs.  If this is correct, I'm a little bit worried about this as it:

- It doesn't look like this would yield the exact same range limb result as the scalar fd_r43x6_sub_fast.  These quad operations are meant to be identical to their scalar counterparts to make developers and numerical analysis life saner.  (E.g. do all numerical analysis in scalar and turn the crank to group into quads and then turn the crank to convert to fast quads without any new numerical analysis).

- Takes slight more operations per lane than fd_r43x6_sub_fast.

Suggest either:

- Keeping the FD_R43X6_QUAD_LANE_SUB_FAST / fd_r43x6_sub_fast as is and add a FD_R43X6_QUAD_LANE_SUB_ALT / fd_r43x6_quad_lane_sub_alt pair with some documentation about the range limbs expected on entry or exist.

- Update fd_r43x6_sub_fast to match with corresponding documentation changes.

I suspect the former is actually less work (e.g. don't have to go through all the places sub_fast is used and update documentation and analysis around existing usages of sub_fast / QUAD_LANE_SUB_FAST.  But YMMV.

kbowers-jump (changes requested):

src/ballet/ed25519/avx512/fd_r43x6_inl.h
0x0ece: I fixed so that FD_R43X6_QUAD_LANE_SUB_FAST / fd_r43x6_sub_fast match.
I'll do another pass to validate performance, and if I see anything weird I'll do the _alt versions instead.

0x0ece:
Included in #1311

PR #1298, opened 2024-02-29
topo: replace constants with string identifiers

mmcgee-jump:


ptaffet-jump (dismissed):
I worry a little about our ability to get all the strings right, but I think it's fine and the best way to make topologies more dynamic.

lidatong (approved):

PR #1299, opened 2024-03-01
topo: move threaded runner to disco

mmcgee-jump:


src/disco/topo/fd_topo_run.c
```diff
@@ -0,0 +1,280 @@
...
+}
+
+void
+fd_topo_run_single_process( fd_topo_t * topo,
+                            int         solana_labs,
+                            uint        uid,
+                            uint        gid,
+                            fd_topo_run_tile_t (* tile_run )( fd_topo_tile_t * tile ),
+                            ulong (* tile_align )( fd_topo_tile_t const * tile ),
+                            ulong (* tile_footprint )( fd_topo_tile_t const * tile ) ) {
```
ptaffet-jump, line 256: I'm not sure if we have a convention for aligning function pointers, but I'd probably align them a bit better (and elsewhere)

ptaffet-jump (approved):

ptaffet-jump (approved):

PR #1300, opened 2024-03-01
Fixed crash in fd_log_private_stack_discover

asiegel-jt:


jumpsiegel (dismissed):

src/util/log/fd_log.c
```diff
@@ -1455,51 +1455,78 @@ fd_log_private_stack_discover( ulong   stack_sz,
...
+  int filefd;
+  if( FD_UNLIKELY( ( filefd = open( "/proc/self/maps", O_RDONLY ) ) < 0 ) ) {
+    FD_LOG_WARNING(( "open( \"/proc/self/maps\" ) failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+    *_stack0 = 0UL;
+    *_stack1 = 0UL;
+    return;
+  }
 
+  char filebuf[1<<16];
+  size_t filelen = 0;
```
mmcgee-jump, line 1467: use ulong, long etc rather than size_t

src/util/log/fd_log.c
```diff
@@ -1455,51 +1455,78 @@ fd_log_private_stack_discover( ulong   stack_sz,
...
-      }
+  int filefd;
+  if( FD_UNLIKELY( ( filefd = open( "/proc/self/maps", O_RDONLY ) ) < 0 ) ) {
+    FD_LOG_WARNING(( "open( \"/proc/self/maps\" ) failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+    *_stack0 = 0UL;
+    *_stack1 = 0UL;
+    return;
+  }
 
+  char filebuf[1<<16];
```
mmcgee-jump, line 1466: I don't love the change here to read the whole file up-front, since there's no guarantee the maps listing fits in this buffer. Can we keep the old line-based parsing?

mmcgee-jump (changes requested):
inline

src/util/log/fd_log.c
asiegel-jt: that's complicated because we might read partial lines

src/util/log/fd_log.c
asiegel-jt: ok

src/util/log/fd_log.c
mmcgee-jump: The implementation already did streaming? You can just leave it basically as it was, except swap fread for read. Line length is guaranteed to be a certain size, since the file format is fixed; while the number of lines is unlimited.

Also this code is in util so probably shouldn't abort or do other shortcuts like that. You also might want to keep the FD_LOG_WARNING if close() fails.

asiegel-jt: @ripatel-fd @mmcgee-jump the problem is that I need to read a line at a time, and I don't know how long the lines are. This requires a buffering/push back solution. raw read and fd_io_buffered doesn't support that

mmcgee-jump: What was wrong with the existing impl? Also you do know how long the lines are... the file format is fixed, the first 5 fields are fixed width, and the last (pathname) is bounded by PATH_MAX.

asiegel-jt: fopen uses a hidden malloc which was crashing. the problem is that if you are reading fixed size chunks, you'll get partial lines, which then have to be shifted over for the next read. that is too much code for what is a trivial application

asiegel-jt: that's a mystery

mmcgee-jump: I think it's worth adding the code to shift the buffer here -- this code is in util so has a higher quality bar than other parts of the codebase, is intended to be reused by other arbitrary non-Firedancer callers.  If you don't do the line-by-line parser you arbitrarily cap the number of mapped regions a process can have where this code works to ... approx ~750. (After which it will fail silently except for an inscrutable warning message and let the caller proceed to crash later.)

asiegel-jt: done

PR #1302, opened 2024-03-05
topo: make wksp<>obj relations explicit

mmcgee-jump:


src/app/fdctl/config.c
```diff
@@ -469,220 +470,365 @@ username_to_uid( char * username ) {
...
+  fd_topob_link( topo, 1,                                "shred_store",  "shred_store",  0,        128UL,                                    4UL*FD_SHRED_STORE_MTU, 4UL+config->tiles.shred.max_pending_shred_sets );
 
-  topo->link_cnt = link_cnt;
+  fd_topob_link( topo, config->layout.verify_tile_count, "quic_sign",    "quic_sign",    0,        128UL,                                    130UL,                  1UL );
+  fd_topob_link( topo, config->layout.verify_tile_count, "sign_quic",    "sign_quic",    0,        128UL,                                    64UL,                   1UL );
+  fd_topob_link( topo, 1,                                "shred_sign",   "shred_sign",   0,        128UL,                                    32UL,                   1UL );
+  fd_topob_link( topo, 1,                                "sign_shred",   "sign_shred",   0,        128UL,                                    64UL,                   1UL );
 
-  ushort tile_to_cpu[ FD_TILE_MAX ];
+  ushort tile_to_cpu[ FD_TILE_MAX ] = { USHORT_MAX }; /* Unassigned tiles will be floating. */
```
ptaffet-jump, line 661: This only initializes the first element to `USHORT_MAX`. The rest are zeroed

src/app/fdctl/config.c
```diff
@@ -469,220 +470,365 @@ username_to_uid( char * username ) {
...
-  for( ulong i=0; i<config->layout.verify_tile_count; i++ ) TILE_OUT( "quic",   i,   "quic_netmux",  i      );
-  for( ulong i=0; i<config->layout.verify_tile_count; i++ ) TILE_IN(  "verify", i,   "quic_verify",  i,   0, 1 ); /* No reliable consumers, verify tiles may be overrun */
-  for( ulong i=0; i<config->layout.verify_tile_count; i++ ) TILE_IN(  "dedup",  0UL, "verify_dedup", i,   1, 1 );
-  /**/                                                      TILE_IN(  "pack",   0UL, "dedup_pack",   0UL, 1, 1 );
-  /**/                                                      TILE_IN(  "pack",   0UL, "gossip_pack",  0UL, 1, 1 );
-  /**/                                                      TILE_IN(  "pack",   0UL, "poh_shred",    0UL, 1, 1 );
+  /*             topo, tile_cnt,                         tile_name, tile_wksp, cnc_wksp,    metrics_wksp, cpu_idx,                    is_labs, out_link,       out_link_indexer */
+  fd_topob_tile( topo, config->layout.net_tile_count,    "net",     "net",     "metric_in", "metric_in",  tile_to_cpu+topo->tile_cnt, 0,       "net_netmux",   INDEX_FLOAT );
```
ptaffet-jump, line 665: What does `INDEX_FLOAT` mean in this context?

src/app/fdctl/config.c
```diff
@@ -469,220 +470,365 @@ username_to_uid( char * username ) {
...
+  for( ulong i=0; i<config->layout.verify_tile_count; i++ ) fd_topob_tile_in(  topo, "netmux",  0UL,          "metric_in", "quic_netmux",  i,            UNRELIABLE, POLLED ); /* No reliable consumers of networking fragments, may be dropped or overrun */
+  /**/                                                      fd_topob_tile_in(  topo, "netmux",  0UL,          "metric_in", "shred_netmux", 0UL,          UNRELIABLE, POLLED ); /* No reliable consumers of networking fragments, may be dropped or overrun */
+  for( ulong i=0; i<config->layout.verify_tile_count; i++ ) fd_topob_tile_in(  topo, "quic",    i,            "metric_in", "netmux_out",   0UL,          UNRELIABLE, POLLED ); /* No reliable consumers of networking fragments, may be dropped or overrun */
+  for( ulong i=0; i<config->layout.verify_tile_count; i++ ) fd_topob_tile_out( topo, "quic",    i,                         "quic_netmux",  i                                );
+  for( ulong i=0; i<config->layout.verify_tile_count; i++ ) fd_topob_tile_in(  topo, "verify",  i,            "metric_in", "quic_verify",  i,            UNRELIABLE, POLLED ); /* No reliable consumers, verify tiles may be overrun */
```
ptaffet-jump, line 696: I like `RELIABLE` and `POLLED` much better than just `1` like it was before, but it should probably be prefixed like `FD_TOPOB_POLLED` since it's in `disco` to avoid collisions. Since that'll make these lines even longer, consider declaring `net_cnt` and `verify_cnt` as local variables or something.

src/app/fdctl/config.c
```diff
@@ -469,220 +470,365 @@ username_to_uid( char * username ) {
...
+  char path[ 64UL ];
+  FD_TEST( fd_cstr_printf_check( path, sizeof(path), NULL, "poh_shred" ) );
+  FD_TEST( fd_pod_insert_ulong( topo->props, path, poh_shred_obj->id ) );
+
+  for( ulong i=0UL; i<topo->tile_cnt; i++ ) {
+    fd_topo_tile_t * tile = &topo->tiles[ i ];
+    if( FD_UNLIKELY( !strcmp( tile->name, "net" ) ) ) {
+      strncpy( tile->net.app_name, config->name, sizeof(tile->net.app_name) );
+      strncpy( tile->net.interface, config->tiles.net.interface, sizeof(tile->net.interface) );
+      tile->net.xdp_aio_depth = config->tiles.net.xdp_aio_depth;
```
ptaffet-jump, line 777: I know this is basically copied, but would you mind cleaning it up? I think adding blank lines in between the cases and aligning equal signs will help.

src/app/fdctl/configure/workspace.c
```diff
@@ -33,6 +33,56 @@ workspace_path( config_t * const config,
...
+static void
+fdctl_obj_new( fd_topo_t const *     topo,
+               fd_topo_obj_t const * obj ) {
+  if( FD_UNLIKELY( !strcmp( obj->name, "tile" ) ) ) {
+    /* No need to do anything, tiles don't have a new. */
+  } else if( FD_UNLIKELY( !strcmp( obj->name, "mcache" ) ) ) {
+    char path[ 32UL ];
+    FD_TEST( fd_cstr_printf_check( path, sizeof( path ), NULL, "obj.%lu.depth", obj->id ) );
+    ulong depth = fd_pod_query_ulong( topo->props, path, ULONG_MAX );
+    FD_TEST( depth!=ULONG_MAX );
```
ptaffet-jump, line 45: That same helper function I suggested above will help clean this function up

src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -700,11 +701,13 @@ unprivileged_init( fd_topo_t *      topo,
 
   ulong expected_shred_version = tile->shred.expected_shred_version;
   if( FD_LIKELY( !expected_shred_version ) ) {
-    ulong * gossip_shred_version = (ulong*)tile->extra[0];
+    ulong busy_obj_id = fd_pod_query_ulong( topo->props, "poh_shred", ULONG_MAX );
+    FD_TEST( busy_obj_id!=ULONG_MAX );
+    ulong * gossip_shred_version = fd_fseq_join( (void*)((ulong)(topo->workspaces[ topo->objs[ busy_obj_id ].wksp_id ].wksp) + topo->objs[ busy_obj_id ].offset) );
```
ptaffet-jump, line 706: I think a macro or static inline along the lines of `fd_topo_get_laddr( topo, obj_id )` for `(void*)((ulong)(topo->workspaces[ topo->objs[ busy_obj_id ].wksp_id ].wksp) + topo->objs[ busy_obj_id ].offset)` would be worthwhile. The code to get the pointer is kind of gross.

src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -700,11 +701,13 @@ unprivileged_init( fd_topo_t *      topo,
...
     FD_LOG_INFO(( "Waiting for shred version to be determined via gossip." ));
-    while( !expected_shred_version ) {
+    do {
```
ptaffet-jump, line 708: Shred version of 0 is an invalid shred version used for "unknown" or "not yet assigned," but shred version is a `ushort`, so I guess `ULONG_MAX` is fine as a sentinel.

src/disco/topo/fd_topo.h
```diff
@@ -363,55 +386,29 @@ fd_topo_leave_workspaces( fd_topo_t *  topo );
...
+                        fd_topo_wksp_t * wksp );
+
+/* Apply a function to every object in the topology. */
+
+void
+fd_topo_wksp_apply( fd_topo_t * topo,
+                    void (* fn )( fd_topo_t const * topo, fd_topo_obj_t const * obj ) );
 
 /* Same as fd_topo_fill_tile but fills in all tiles in the topology with
    the given mode. */
```
ptaffet-jump, line 409: `mode` is gone

src/disco/topo/fd_topob.c
```diff
@@ -0,0 +1,428 @@
+#include "fd_topob.h"
+
+fd_topo_t
+fd_topob_new( char const * app_name ) {
```
ptaffet-jump, line 4: `fd_topo_t` is pretty big. I think sticking with the normal `footprint/align/new/join/leave/delete` conventions would be the most clear (though it might require combining the `topo` and `topob` in a strange way), but if you don't want to do that, it should be something like `fd_topob_init` that takes a `fd_topo_t * ` as the first argument.

ptaffet-jump: Another FD pattern that might be more applicable is for `join` to return an `fd_topob_t *` (which is maybe just a typedef for `fd_topo_t *`) and then `fd_topob_fini` returns an `fd_topo_t *`.

ptaffet-jump: Then the functions would be named more like `add_*`

ptaffet-jump (commented):
Most of my comments are just ways to clean up the code. The only actual issue is with the floating CPU array initialization. I also think `fd_topob` feels a bit out of place, but just needs a bit more massaging to conform with other more common FD coding patterns.

src/app/fdctl/run/tiles/fd_shred.c
mmcgee-jump: This changed just because we fd_fseq_new with ULONG_MAX, and I didn't want to special case it.

src/disco/topo/fd_topob.c
mmcgee-jump: Leaving this for a later change.

ptaffet-jump (approved):
Stamping to unblock, but `topob` could use some refactoring as discussed. Also, the comments are not as precise as normal.

ptaffet-jump (approved):
Restamping

PR #1313, opened 2024-03-12
topo: decouple quic tile count from verify

mmcgee-jump:


src/app/fdctl/run/tiles/fd_verify.c
```diff
@@ -33,6 +33,19 @@ mux_ctx( void * scratch ) {
...
+before_frag( void * _ctx,
+             ulong  in_idx,
+             ulong  seq,
+             ulong  sig,
+             int *  opt_filter ) {
+  (void)in_idx;
+  (void)sig;
+
+  fd_verify_ctx_t * ctx = (fd_verify_ctx_t *)_ctx;
+  if( FD_LIKELY( (seq % ctx->round_robin_cnt) != ctx->round_robin_idx ) ) *opt_filter = 1;
```
ptaffet-jump, line 46: Maybe I'm thinking too hard about this, but I'd make it something like `fd_ulong_hash( seq + in_idx<< 48)`. With seq, assuming load is distributed very evenly among the quic tiles, you'll get more work collisions than you'd expect with a random distribution (e.g. two quic tiles produce txns for verify n, then they both produce for n+1, etc. until they get out of phase).

ptaffet-jump (approved):
Looks good, feel free to take or leave the comment about hashing

src/app/fdctl/run/tiles/fd_verify.c
mmcgee-jump: Will leave for a different change, there are a few places we RR like this.

PR #1317, opened 2024-03-12
fd_pack: assorted small improvements

ptaffet-jump:
Some performance improvements and general code cleanup.

- Randomize which bank we start scheduling transactions for
- Use the data block limit in addition to the max microblocks per block limit. The former tends to be much looser.
-  Remove the last vestiges of `delay_end_block` and `conflicting_with`.  An earlier version moved transactions that couldn't be scheduled so they weren't considered again until the conflicting transaction had finished.  With the addition of the bitsets, that doesn't make sense from a performance perspective anymore.
- Improved metrics
- Added a function to check some invariants
- Added a better performance test
- Support non-standard chains (not 400 ms/slot, not 64 ticks/slot)
- Refactored releasing references to bits for bitvector, which fixed a small and subtle bug


mmcgee-jump:
Specific changes look good but a few thoughts,

* The number of ticks per block is set at genesis and could (for non-standard chains) be a different value than 64. We shoudl either try to support this, or explicitly assert that it's 64 and fail loudly.

* How does the shredder fail if it would produce too much data due to a pack error? We definitely want to abort the validator in this case before finishing the invalid block.

mmcgee-jump (approved):

mmcgee-jump (changes requested):
See comment, also needs a rebase

src/app/fdctl/run/tiles/fd_poh.c
```diff
@@ -569,8 +569,8 @@ fd_ext_poh_initialize( ulong         hashcnt_duration_ns, /* See clock comments
     /* Low power producer, maximum of one microblock per tick in the slot */
     ctx->max_microblocks_per_slot = ctx->ticks_per_slot;
   } else {
-    ctx->max_microblocks_per_slot = FD_PACK_MAX_MICROBLOCKS_PER_BLOCK;
-    FD_TEST( ctx->hashcnt_per_tick>=FD_PACK_MAX_MICROBLOCKS_PER_BLOCK );
+    /* We can set this to more or less whatever we want. */
```
ptaffet-jump, line 572: It no longer has any impact on shreds. It's just how many hashcnt we want to reserve.

ptaffet-jump:
If we go over the shred limit, the Rust side panics. It may be after we've sent out the block though. Other validators will just reject the block.

src/app/fdctl/run/tiles/fd_poh.c
```diff
@@ -700,9 +700,11 @@ publish_became_leader( fd_poh_ctx_t * ctx,
   uchar * dst = (uchar *)fd_chunk_to_laddr( ctx->pack_out_mem, ctx->pack_out_chunk );
 
   fd_became_leader_t * leader = (fd_became_leader_t *)dst;
-  leader->slot_start_ns = slot_start_ns;
-  leader->bank = ctx->current_leader_bank;
+  leader->slot_start_ns           = slot_start_ns;
```
mmcgee-jump, line 703: Do we still need to pass slot_start_ns?

mmcgee-jump (approved):

src/app/fdctl/run/tiles/fd_poh.c
ptaffet-jump: No, but it might be helpful for metrics in the future, so I left it

PR #1323, opened 2024-03-14
fddev: add tool to dump mcaches to a pcap

ptaffet-jump:


src/app/fddev/dump.c
```diff
@@ -0,0 +1,88 @@
+#include "fddev.h"
+#include "../../util/net/fd_pcap.h"
+#include <stdio.h>
+
+void
+dump_cmd_args( int      * argc,
+               char * * * argv,
+               args_t   * args ) {
+  char const * out_file = fd_env_strip_cmdline_cstr( argc, argv, "--out-file", NULL, "dump.pcap" );
+  char const * link     = fd_env_strip_cmdline_cstr( argc, argv, "--link",     NULL, "all"       );
```
mmcgee-jump, line 10: Don't love using `all` here which someone could name a link, better to use NULL?

mmcgee-jump (dismissed):
Nice!

src/app/fddev/dump.c
ptaffet-jump: Good point. I switched to using the empty string instead.

mmcgee-jump (approved):

PR #1327, opened 2024-03-17
shred: allow multiple tiles

anwayde:


src/disco/shred/fd_shredder.c
```diff
@@ -82,6 +82,31 @@ fd_shredder_delete( void *          mem      ) {
...
+                        ulong           slot ) {
+
+  if( FD_UNLIKELY( entry_batch_sz==0UL ) ) return NULL;
+
+  if( FD_UNLIKELY( slot != shredder->slot ) ) {
+    shredder->data_idx_offset   = 0UL;
+    shredder->parity_idx_offset = 0UL;
+  }
+
+  ulong entry_bytes           =  entry_batch_sz - shredder->offset;
```
ptaffet-jump, line 97:
```suggestion
  ulong entry_bytes           =  entry_batch_sz;
```

src/disco/shred/fd_shredder.c
```diff
@@ -82,6 +82,31 @@ fd_shredder_delete( void *          mem      ) {
...
+  ulong chunk_size            =  fd_ulong_if( entry_bytes>=2UL*31200UL, 31200UL, entry_bytes );
```
ptaffet-jump, line 98: I don't think this is right. You should get rid of `chunk_size`, since in this case it's the same as `entry_batch_sz`.

src/disco/shred/fd_shredder.h
```diff
@@ -164,6 +164,16 @@ fd_shredder_t * fd_shredder_init_batch( fd_shredder_t               * shredder,
                                         ulong                         slot,
                                         fd_entry_batch_meta_t const * meta );
 
+/* fd_shredder_skip_batch updates the shredder state as necessary
+   to skip processing this current batch.  shredder must be a valid
+   local join.  entry_batch_sz must be strictly positive.
+
+   Returns shredder, which will have data and parity shred indices
+   updated without processing the entry. */
```
ptaffet-jump, line 172:
```suggestion
   updated as if the caller had called fd_shredder_init_batch with a batch of the specified size, followed by fd_shredder_next_fec_set exactly fd_shredder_count_fec_sets( entry_batch_sz ) times. */
```

src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -124,6 +124,13 @@ typedef struct {
   fd_fec_resolver_t  * resolver;
   fd_pubkey_t          identity_key[1]; /* Just the public key */
 
+  ulong                round_robin_id;
+  ulong                round_robin_cnt;
+  /* Number of batches we shredded from PoH during the current slot */
```
ptaffet-jump, line 129: "we shredded" seems to suggest maybe just this tile, but this is for any tile, right?

src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -124,6 +124,13 @@ typedef struct {
...
+  ulong                batch_cnt;
+  /* Slot that we last processed from PoH */
```
ptaffet-jump, line 131:
```suggestion
  /* Slot of the most recent microblock we've seen from PoH, or 0 if we haven't seen one yet */
```

src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -392,6 +413,18 @@ during_frag( void * _ctx,
     uchar const * dcache_entry = fd_chunk_to_laddr_const( ctx->net_in_mem, chunk );
     ulong hdr_sz = fd_disco_netmux_sig_hdr_sz( sig );
     FD_TEST( hdr_sz < sz ); /* Should be ensured by the net tile */
+    fd_shred_t const * shred = fd_shred_parse( dcache_entry+hdr_sz, sz-hdr_sz );
+    if( FD_UNLIKELY( !shred ) ) {
+      *opt_filter = 1;
+      return;
+    };
+    /* all shreds in the same FEC set will have the same signature
+       so we can parallelize handling them */
```
ptaffet-jump, line 422:
```suggestion
       so we can round-robin shreds between the shred tiles based on just the signature without splitting individual FEC sets. */
```

ptaffet-jump (commented):
Getting there! Can you add a test for `fd_shredder_skip_batch`? Maybe create two shredders and randomly choose between skipping and shredding, and make sure they produce byte-for-byte identical output. Write the test before you make the changes I suggested to the function, and it should fail, showing the problem with batches that span multiple FEC sets.

src/disco/shred/test_shredder.c
```diff
@@ -141,6 +141,85 @@ test_shredder_pcap( void ) {
...
+      slot++;
+      data_shred_cnt   = 0;
+      parity_shred_cnt = 0;
+    }
+  }
+
+  /* Make all the shredders process the same data and compare the outputs. */
+  fd_fec_set_t _sets[ SHREDDERS ];
+  for( ulong i=0; i<SHREDDERS; i++ ) {
+    for( ulong j=0UL; j<FD_REEDSOL_DATA_SHREDS_MAX;   j++ ) ( &_sets[ i ] )->data_shreds[   j ] = fec_set_memory_1 + 2048UL*j;
```
ptaffet-jump, line 210:
```suggestion
    for( ulong j=0UL; j<FD_REEDSOL_DATA_SHREDS_MAX;   j++ ) sets[ i ].data_shreds[   j ] = fec_set_memory_1 + 2048UL*j;
```

src/disco/shred/test_shredder.c
```diff
@@ -141,6 +141,85 @@ test_shredder_pcap( void ) {
...
+    for( ulong j=0UL; j<FD_REEDSOL_PARITY_SHREDS_MAX; j++ ) ( &_sets[ i ] )->parity_shreds[ j ] = fec_set_memory_2 + 2048UL*j;
+    FD_TEST( fd_shredder_init_batch( shredders[ i ], data, idx, slot, meta ) );
+    FD_TEST( fd_shredder_next_fec_set( shredders[ i ], &_sets[ i ] ) );
+    FD_TEST( fd_shredder_fini_batch( shredders[ i ] ) );
+
+    FD_TEST( !memcmp( ( &_sets[ 0 ] )->data_shreds,   ( &_sets[ i ] )->data_shreds,   FD_REEDSOL_DATA_SHREDS_MAX ) );
+    FD_TEST( !memcmp( ( &_sets[ 0 ] )->parity_shreds, ( &_sets[ i ] )->parity_shreds, FD_REEDSOL_PARITY_SHREDS_MAX ) );
```
ptaffet-jump, line 217: This does not do what you think. This compares the pointers, but also `memcmp` takes a number of bytes, so the third argument is wrong.

src/disco/shred/test_shredder.c
ptaffet-jump: But actually, the problem is deeper than just formatting. Each FEC set is pointing to the same memory, which means they're all going to overwrite each other, and this comparison test wouldn't actually test anything

ptaffet-jump (commented):
Getting there. Test needs a bit more work though.

src/disco/shred/test_shredder.c
ptaffet-jump: But in general, if you want a pointer to the ith element, `(sets+i)` is your friend. If you're going to dereference it immediately with `->`, I'd just go with `sets[i]`.

src/disco/shred/test_shredder.c
```diff
@@ -141,6 +141,109 @@ test_shredder_pcap( void ) {
...
+  ulong idx  = 0UL;
+  ulong slot = 1UL;
+
+  ulong data_sz = 1024UL*1024UL;
+  uchar data    [ 1024UL*1024UL ];
+  for( ulong i=0; i<data_sz; i++ ) data[ i ] = fd_rng_uchar( r );
+
+  /* Randomly choose a shredder to process a batch until the data buffer is exhausted. */
+  while( idx<data_sz ) {
+    ulong sz        = ( fd_rng_ulong_roll( r, 4UL )+1UL )*25000UL;
```
ptaffet-jump, line 185: How about just making this a random number between 1 and 100k? No need for it to be a multiple of 25k.

src/disco/shred/test_shredder.c
```diff
@@ -141,6 +141,109 @@ test_shredder_pcap( void ) {
...
+  FD_TEST( fd_shredder_next_fec_set( shredders[ 0 ], _set ) );
+  FD_TEST( fd_shredder_fini_batch( shredders[ 0 ] ) );
+
+  /* Make all the other shredders process the same data and compare the outputs. */
+  fd_fec_set_t _temp_set[ 1 ];
+  uchar temp_data_shreds  [ 2048UL*FD_REEDSOL_DATA_SHREDS_MAX   ] = { 0 };
+  uchar temp_parity_shreds[ 2048UL*FD_REEDSOL_PARITY_SHREDS_MAX ] = { 0 };
+  for( ulong j=0UL; j<FD_REEDSOL_DATA_SHREDS_MAX;   j++ ) _temp_set->data_shreds[   j ] = temp_data_shreds + 2048UL*j;
+  for( ulong j=0UL; j<FD_REEDSOL_PARITY_SHREDS_MAX; j++ ) _temp_set->parity_shreds[ j ] = temp_parity_shreds + 2048UL*j;
+  for( ulong i=1; i<SHREDDERS; i++ ) {
```
ptaffet-jump, line 228:
```suggestion
  for( ulong i=1UL; i<SHREDDERS; i++ ) {
```

src/disco/shred/test_shredder.c
```diff
@@ -141,6 +141,109 @@ test_shredder_pcap( void ) {
...
+    memset( temp_data_shreds, 0, FD_REEDSOL_DATA_SHREDS_MAX*2048UL );
```
ptaffet-jump, line 229:
```suggestion
    memset( temp_data_shreds,   0, FD_REEDSOL_DATA_SHREDS_MAX*2048UL );
```
Align these

src/disco/shred/test_shredder.c
```diff
@@ -141,6 +141,109 @@ test_shredder_pcap( void ) {
...
+  for( ulong j=0UL; j<FD_REEDSOL_DATA_SHREDS_MAX;   j++ ) _set->data_shreds[   j ] = data_shreds + 2048UL*j;
+  for( ulong j=0UL; j<FD_REEDSOL_PARITY_SHREDS_MAX; j++ ) _set->parity_shreds[ j ] = parity_shreds + 2048UL*j;
+
+  ulong data_shred_cnt   = 0;
+  ulong parity_shred_cnt = 0;
+  ulong idx  = 0UL;
+  ulong slot = 1UL;
+
+  ulong data_sz = 1024UL*1024UL;
+  uchar data    [ 1024UL*1024UL ];
```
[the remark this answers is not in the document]
ptaffet-jump: You can use perf_test_entry_batch (in which case you might want to rename it), or just move these variables to a global scope.

ptaffet-jump (commented):
Great! The only thing that really needs to be fixed is the stack space consumption. Other than that, a few nitpick comments, but the test is greatly improved.

ptaffet-jump (approved):

PR #1330, opened 2024-03-18
ballet: add Rust-compatible UTF-8 validator

ripatel-fd:
Partially reverts commit 8a8c95a3798ec1b17054e698bdc4cdae34855447, which removed the UTF-8 validator last year.


ripatel-fd:
Moving to #1326

PR #1331, opened 2024-03-18
poh: support reset in the middle of leader slot

mmcgee-jump:


src/app/fdctl/run/tiles/fd_pack.c
```diff
@@ -181,8 +178,15 @@ after_credit( void *             _ctx,
          We reuse the slot field when sending this message to send the
```
ptaffet-jump, line 178: Comment needs to be updated

ptaffet-jump (approved):

PR #1338, opened 2024-03-19
shmem: reserve huge and gigantic pages exclusively

mmcgee-jump:
This should prevent mmap() from ever failing due to lack of huge or gigantic pages, if the hugetlbfs mount was successfully created, which moves runtime (fddev) failures to configure time.

Per the kernel documentation below, `size=` on the hugetlbfs mount is a maximum limit, and the mount will be unlimited, so we don't need to provide this.  Calculating the total memory available and attempting to set the maximum to this is especially not useful.

What we really want is to reserve pages from the global huge page pool specific for firedancer use, which is achieved with min_size instead. One unfortunate thing is that not all application pages come from the mount (eg, anonymous stack pages), so we need to leave those ones out of the min_size.

  The size option sets the maximum value of memory (huge pages) allowed
  for that filesystem (/mnt/huge). The size option can be specified in
  bytes, or as a percentage of the specified huge page pool
  (nr_hugepages). The size is rounded down to HPAGE_SIZE boundary.

  The min_size option sets the minimum value of memory (huge pages)
  allowed for the filesystem. min_size can be specified in the same way
  as size, either bytes or a percentage of the huge page pool. At mount
  time, the number of huge pages specified by min_size are reserved for
  use by the filesystem. If there are not enough free huge pages
  available, the mount will fail. As huge pages are allocated to the
  filesystem and freed, the reserve count is adjusted so that the sum of
  allocated and reserved huge pages is always at least min_size.

  The option nr_inodes sets the maximum number of inodes that /mnt/huge
  can use.

  If the size, min_size or nr_inodes option is not provided on command
  line then no limits are set.


src/disco/topo/fd_topo.c
```diff
@@ -252,6 +252,18 @@ fd_topo_gigantic_page_cnt( fd_topo_t * topo ) {
...
+      result += topo->workspaces[ i ].page_cnt;
+    }
+  }
+
+  return result;
+}
+
 FD_FN_PURE ulong
 fd_topo_huge_page_cnt( fd_topo_t * topo ) {
   ulong result = 0UL;
```
ptaffet-jump, line 269: You can replace the loop below with a call to `fd_topo_mounted_huge_page_cnt`

ptaffet-jump: Or what about making this one function with a bool flag whether to include extra (non-wksp) pages or not? `mounted` doesn't seem like the right word to describe these pages to me

ptaffet-jump (dismissed):

ptaffet-jump (approved):

src/util/tile/fd_tile_threads.cxx
```diff
@@ -9,6 +9,7 @@
 #include <sched.h>
 #include <syscall.h>
 #include <sys/resource.h>
+#include <sys/prctl.h>
```
ptaffet-jump, line 12: Is this include still needed?

ptaffet-jump (commented):
Some thoughts, but not really anything that needs to be fixed, other than CI

src/app/fdctl/configure/hugetlbfs.c
```diff
@@ -0,0 +1,300 @@
...
+            FD_LOG_ERR(( "error closing `/proc/self/mounts` (%i-%s)", errno, fd_io_strerror( errno ) ));
+          PARTIALLY_CONFIGURED( "mount `%s` has malformed min_size, expected `min_size=%lu`", mount_path[ i ], required_min_size[ i ] );
+        }
+
+        if( FD_UNLIKELY( min_size<required_min_size[ i ] ) ) {
+          if( FD_UNLIKELY( fclose( fp ) ) )
+            FD_LOG_ERR(( "error closing `/proc/self/mounts` (%i-%s)", errno, fd_io_strerror( errno ) ));
+          PARTIALLY_CONFIGURED( "mount `%s` has min_size `%lu`, expected at least `min_size=%lu`", mount_path[ i ], min_size, required_min_size[ i ] );
+        }
+
```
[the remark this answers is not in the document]
mmcgee-jump: I wanted to do this, but min_size could be >=, not just = so a string compare won't work.

src/app/fdctl/configure/workspace.c
```diff
@@ -84,51 +67,53 @@ init( config_t * const config ) {
...
-          FD_LOG_ERR(( "unlink failed when trying to delete wksp `%s` (%i-%s)", path, errno, fd_io_strerror( errno ) ));
-      }
-    }
-    else if( FD_LIKELY( result && errno == ENOENT ) ) continue;
-    else FD_LOG_ERR(( "stat failed when trying to delete wksp `%s` (%i-%s)", path, errno, fd_io_strerror( errno ) ));
+  (void)config;
+
+  for( ulong i=0UL; i<2UL; i++ ) {
+    char mount_path[ FD_SHMEM_PRIVATE_PATH_BUF_MAX ];
+    FD_TEST( fd_cstr_printf_check( mount_path, FD_SHMEM_PRIVATE_PATH_BUF_MAX, NULL, "%s/.%s", fd_shmem_private_base, MOUNT_NAMES[ i ] ));
```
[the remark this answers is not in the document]
mmcgee-jump: This is kind of broken right now because of the way shmem works. I filed an issue to fix in another PR.

PR #1351, opened 2024-03-21
fdctl: support to affine labs to cores

mmcgee-jump:


ptaffet-jump (approved):
I have never heard of `affine` as a verb, but works for me

PR #1383, opened 2024-03-26
bench: add consensus breaking option to raise block CU limit

mmcgee-jump:


src/app/fdctl/config/default.toml
```diff
@@ -1005,3 +1005,15 @@ dynamic_port_range = "8900-9000"
         #
         #      bencho, benchg1, .... benchgN, benchs
         affinity = "f6"
+
+        # Solana has a hard-coded maximum CUs per block limit of
+        # 48,000,000 which works out to around 180,000 transactions a
```
mmcgee-jump, line 1010: Not right?

src/app/fdctl/run/tiles/fd_pack.c
```diff
@@ -148,13 +151,21 @@ scratch_align( void ) {
 
 FD_FN_PURE static inline ulong
 scratch_footprint( fd_topo_tile_t const * tile ) {
-  (void)tile;
+  fd_pack_limits_t limits[1] = {{
+    .max_cost_per_block        = tile->pack.larger_max_cost_per_block ? LARGER_MAX_COST_PER_BLOCK : FD_PACK_MAX_COST_PER_BLOCK,
+    .max_vote_cost_per_block   = FD_PACK_MAX_VOTE_COST_PER_BLOCK,
```
mmcgee-jump, line 156: Need to update the solana side to leave this the same as well then

ptaffet-jump: I had made the change but not pushed everything. Thanks for catching it!

ptaffet-jump (approved):

PR #1388, opened 2024-03-27
bench: add consensus breaking option to raise shred limits

ptaffet-jump:


src/app/fdctl/config/default.toml
```diff
@@ -1021,9 +1021,24 @@ dynamic_port_range = "8900-9000"
         # second.  When benchmarking, this can be the limiting
         # bottleneck of the system, so this option is provided to raise
         # the limit.  If set to true, the limit will be lifted to
-        # 288,000,000 for a little over 1 million transactions a second.
+        # 288,000,000 for a little over 1 million cheap transactions a
```
mmcgee-jump, line 1024: Sorry my fault. Isn't this actually 480,000 TPS? (6*80,000) Can we update the number to be over 1M ?

src/app/fdctl/config/default.toml
```diff
@@ -1021,9 +1021,24 @@ dynamic_port_range = "8900-9000"
...
+        # second.
         #
         # This option should not be used in production, and would cause
         # consensus violations and a consensus difference between the
         # validator and the rest of the network.
         larger_max_cost_per_block = false
+
+        # Solana has a consensus-agreed-upon limit of 32k data and
```
mmcgee-jump, line 1032: 32768? 32,000? 32KiB ?

mmcgee-jump (approved):

src/app/fdctl/config/default.toml
ptaffet-jump: I updated it to 13x, which should be over 1M

src/app/fdctl/config/default.toml
ptaffet-jump: Thanks, I clarified it's 32*1024

mmcgee-jump (approved):

PR #1408, opened 2024-03-29
quic: remove packet buffer from conn state

ripatel-fd:
Moves the scratch buffer for packet protection to the QUIC tile state.
Further reduces the fd_quic_conn_t size to 6440 bytes.


src/waltz/quic/fd_quic_private.h
```diff
@@ -111,6 +111,9 @@ struct __attribute__((aligned(16UL))) fd_quic_state_private {
 
   /* last arp/routing tables update */
   ulong ip_table_upd;
+
+  /* Scratch space for packet protection */
+  uchar                   crypt_scratch[FD_QUIC_MTU];
```
nbridge-jump, line 116: I sized the buffer to allow for block ciphers... I suppose that isn't necessary here?

nbridge-jump (approved):

PR #1409, opened 2024-03-29
quic: share object pools across connections

### LINKED ISSUE #1407: quic: inflight packet pool should be shared across conns
**Problem**

Currently, every QUIC connection owns an exclusive pool of inflight packets. 
This pool is used to track what has been sent to the peer which is eventually used when processing ACKs.

The fdctl default config file specifies a depth of 65536 packets across 2048 conns, which is about 67 GB (!) of memory.
This scales badly with connection table size and limits the per-connection burst rate.

**Suggested Fix**

Instead, let each quic tile manage a shared pool of inflight packets. 

The pool size is determined by the max packet burst rate in a time window of end-to-end latency. 
E.g. for 1Mpps burst rate and 500ms end-to-end latency, the pool size is 250MB.

ripatel-fd:
Fixes #1407 

- **quic: remove test_tls_quic_both**
- **quic: simplify fd_rng_t usage**
- **quic: move ACK pool to tile**


ripatel-fd:
This PR is stale. I will redo it.

PR #1417, opened 2024-04-01
double free map entry in test

marctrem:


ripatel-fd:
@marctrem What is the context of this PR? Did you find a vulnerability relating to a double free?

PR #1420, opened 2024-04-01
build: report firedancer client id and version over gossip

### LINKED ISSUE #1379: Switch gossip validator name to Firedancer

mmcgee-jump:
Fixes #1379


ripatel-fd:
I think this is a typical example of a change that should be isolated to a 1.5 release branch, but not be on main

PR #1423, opened 2024-04-01
Create executable binary for executing Protobuf instructions

mjain-jumptrading:
Creates binary that accepts Protobuf message files and executes them through Firedancer. Useful for being able to step through and debug individual instructions through testing suites.


ripatel-fd:
Closing this since we now have the ability to debug the shared library target

PR #1428, opened 2024-04-02
fdctl: better workspace ENOMEM diagnostics

mmcgee-jump:


src/app/fdctl/configure/hugetlbfs.c
```diff
@@ -109,6 +110,56 @@ init( config_t * const config ) {
   }
 }
 
+static void
+cmdline( char * buf,
+         ulong  buf_len ) {
```
ptaffet-jump, line 115: `len` typically means strlen. This should be `buf_sz`

src/app/fdctl/configure/workspace.c
```diff
@@ -14,7 +14,12 @@
 static void
 init_perm( fd_caps_ctx_t *  caps,
            config_t * const config ) {
-  ulong mlock_limit = fd_topo_mlock( &config->topo );
+  ulong mlock_limit = 0UL;
```
ptaffet-jump, line 17: Does this need to be a sum instead of a max if running in single process mode?

ptaffet-jump (dismissed):
Great! Some nits, but it's fine to merge if you'd like

src/app/fdctl/configure/workspace.c
mmcgee-jump: Nope, these workspaces are created and then unmapped one by one.

ptaffet-jump (approved):

PR #1440, opened 2024-04-03
Changed max_inflight_quic_packets to 2048

nbridge-jump:


mmcgee-jump:
Why is this decreasing?

nbridge-jump:
> Why is this decreasing?

It never needed to be that high. I didn't mean for the increase to get checked in

PR #1442, opened 2024-04-03
Changed fd_quic_crypto_encrypt to avoid FD_LOG_ERR

nbridge-jump:


nbridge-jump:
Addresses https://github.com/firedancer-io/auditor-internal/issues/26

mmcgee-jump (approved):

PR #1444, opened 2024-04-04
spammer: support for standalone benchmarking

### LINKED ISSUE #1370: Make bench tool able to run standalone as well

anwayde:
Closes #1370

### DIFF
--- a/src/app/fdctl/fdctl.h
+++ b/src/app/fdctl/fdctl.h
@@ -69,6 +69,18 @@ typedef union {
   struct {
     char name[ 13UL ];
   } flame;
+
+  struct {
+    char    affinity[ AFFINITY_SZ ];
+    uint    tpu_ip;
+    uint    rpc_ip;
+    ushort  tpu_port;
+    ushort  rpc_port;
+    ulong   accounts;
+    ulong   connections;
+    ulong   benchg;
+    ulong   benchs;
+  } spammer;
 } args_t;
 
 typedef struct fd_caps_ctx fd_caps_ctx_t;
--- a/src/app/fddev/tiles/fd_benchs.c
+++ b/src/app/fddev/tiles/fd_benchs.c
@@ -86,7 +86,7 @@ privileged_init( fd_topo_t *      topo,
 
   ushort port = 12000;
 
-  ctx->conn_cnt = fd_topo_tile_name_cnt( topo, "quic" );
+  ctx->conn_cnt = tile->benchs.conn_cnt;
   FD_TEST( ctx->conn_cnt <=sizeof(ctx->conn_fd)/sizeof(*ctx->conn_fd) );
   for( ulong i=0UL; i<ctx->conn_cnt ; i++ ) {
     int conn_fd = socket( AF_INET, SOCK_DGRAM, 0 );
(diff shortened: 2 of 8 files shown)


mmcgee-jump:
I think my main hesitation is growing the config TOML for new commands. It's more for configuring the production validator, and I stuck the bench stuff in there as a bit of a convenience because it's reasonably related when you are benching the validator that you plan to run.

For this command, though, it stands on its own. I think we could rename it `spammer` maybe, and make the arguments command line arguments, rather than being in the TOML file.

src/disco/topo/fd_topob.c
```diff
@@ -120,7 +120,7 @@ fd_topob_tile( fd_topo_t *    topo,
   if( FD_UNLIKELY( topo->tile_cnt>=FD_TOPO_MAX_TILES ) ) FD_LOG_ERR(( "too many tiles" ));
 
   ulong kind_id = 0UL;
-  for( ulong i=0UL; i<topo->link_cnt; i++ ) {
```
mmcgee-jump, line 123: Can you make this a separate PR? Think it should go in ASAP.

src/app/fddev/tiles/fd_benchs.c
```diff
@@ -86,7 +86,7 @@ privileged_init( fd_topo_t *      topo,
 
   ushort port = 12000;
 
-  ctx->conn_cnt = fd_topo_tile_name_cnt( topo, "quic" );
```
mmcgee-jump, line 89: This change makes sense for the spammer, but not when running the existing bench command, since it should auto detect this from the topo still.

mmcgee-jump (changes requested):

src/app/fddev/tiles/fd_benchs.c
anwayde: Updated, plumbed it through using `layout.quic_tile_count`.

src/app/fdctl/fdctl.h
```diff
@@ -68,6 +68,18 @@ typedef union {
   struct {
     char name[ 13UL ];
   } flame;
+
+  struct {
+    const char * affinity;
```
mmcgee-jump, line 73: These should be arrays in the struct, and copy the string in there. You don't know if lifetime of those arguments.

src/app/fdctl/config/default.toml
```diff
@@ -981,7 +981,7 @@ dynamic_port_range = "8900-9000"
         # genesis, but it might be immediately overriden at runtime if
         # the related features which increase this value are enabled.
         # The features are like update_hashes_per_tick2.
-        hashes_per_tick = 1
+        hashes_per_tick = 12500
```
mmcgee-jump, line 984: Let's leave this for another change. Needs more thought what the default should be (this prevents early epoch rollovers in development).

mmcgee-jump (changes requested):
Nice! thanks for changes, just one minor thing then looks good

src/app/fdctl/fdctl.h
```diff
@@ -69,6 +69,18 @@ typedef union {
   struct {
     char name[ 13UL ];
   } flame;
+
+  struct {
+    char    affinity[ AFFINITY_SZ ];
+    char    tpu_ip[ 32 ];
```
mmcgee-jump, line 75: These can be `uint` and do the parse in the args fn.

mmcgee-jump:
Another nit but LGTM

mmcgee-jump (dismissed):

mmcgee-jump (approved):

PR #1453, opened 2024-04-05
config: parse both uid and gid

### LINKED ISSUE #1398: Check that fdctl/fddev works properly when `gid!=uid`
Nick got an error while trying to set up a new box in which his uid and gid are different

anwayde:
Fixes #1398


mmcgee-jump:
This isn't quite the right fix I think, we actually want to keep gid=uid here, since it determines how we permission files (we want only for the solana user, not group). I think you can just relax the check that the running uid and gid are different?

mmcgee-jump (approved):

PR #1463, opened 2024-04-08
ci: remove unused code

ripatel-fd:
No idea what the HAS_APT_GET identifier does.
Does not appear to be used.


ripatel-fd:
Nvm, ignore, lol

PR #1467, opened 2024-04-09
config: specify uid and gid

### LINKED ISSUE #1454: config: take uid and gid from config file, not username
It's difficult to map username to a uid/gid, because sometimes we need to talk to sssd and do other such nonsense. Instead, just take a UID/GID instead.

anwayde:
Fixes #1454

### DIFF
--- a/book/guide/configuring.md
+++ b/book/guide/configuring.md
@@ -24,6 +24,10 @@ validator might look like:
 ```sh [bash]
 # /home/firedancer/config.toml
 user = "firedancer"
+# id -u firedancer
+uid  = 1000
+# id -g firedancer
+gid  = 1000
 [gossip]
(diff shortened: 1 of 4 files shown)


mmcgee-jump:
I think it's probably not the right approach. If we can't get the username from the UID/GID then there's duplicate information in the config file (username + uid + gid) and it can cause subtle security / other issues if they don't line up.

Probably the right thing to do here is take the username from config as before, but query nscd as in musl, eg https://elixir.bootlin.com/musl/latest/source/src/passwd/getpw_a.c#L20

mmcgee-jump (changes requested):
comment

mmcgee-jump:
Replaced by #1863

PR #1472, opened 2024-04-10
docs: officialize compiler support

marctrem:


book/guide/getting-started.md
```diff
@@ -28,9 +28,10 @@ Firedancer must be built from source and requires the following,
 
  - A linux kernel version 5.7 or higher, or with support for
    `BPF_OBJ_PIN`.
- - GCC version 8.3 or higher.
```
mmcgee-jump, line 31: This is suggested, not required, per the language above. Suggest, something like: `GCC version 8.3 or higher, although only GCC version 11, 12, and 13 are supported and tested by the Firedancer developers.`

doc/getting-started.md
```diff
@@ -1,6 +1,11 @@
 Getting started
 ===============
 
+Compiler support
+----------------
+
+Officially supported compilers for production use are the latest releases of _gcc 11_, _gcc 12_ and _gcc 13_.
```
mmcgee-jump, line 7: Don't love the word official here since it's a community project. Something like "The Firedancer development team tests and supports the following compilers"

README.md
```diff
@@ -22,6 +22,12 @@ The below describes building Frankendancer from scratch and running it
 optimized on a stock Linux image. You will need basic development tools
 like `make`, `gcc` along with `rustc`, and `clang`.
 
+Officially supported compilers for production use are the latest releases of _gcc 11_, _gcc 12_ and _gcc 13_.
+Other compilers are likely to work but are not tested or secured by the Firedancer team.
```
mmcgee-jump, line 26: nit: no newline here

README.md
```diff
@@ -22,6 +22,12 @@ The below describes building Frankendancer from scratch and running it
...
+
+
+The latest release of Clang 17 is supported for development and tooling.
```
mmcgee-jump, line 29: I think better to just say "Building with clang is supported for static analysis and other development use cases, but not recommended for production."

mmcgee-jump (dismissed):
comments inline

README.md
marctrem: I don't think that we're doing static analysis with clang (outside of the checks that are implicit to building).

book/guide/getting-started.md
riptl: Agree with @mmcgee-jump.

mmcgee-jump:
Addressed in #1991

PR #1473, opened 2024-04-10
shmem: use unified mount path

### LINKED ISSUE #1375: Get `huge_page_mount_path` working correctly
These two options in the config TOML don't quite work correctly, since the shmem code hardcodes a path, and expects both .gigantic and .huge to be under that path.

anwayde:
Fixes #1375

### DIFF
--- a/src/app/fdctl/config.c
+++ b/src/app/fdctl/config.c
@@ -234,8 +234,7 @@ static int parse_key_value( config_t *   config,
   ENTRY_UINT  ( ., layout,              bank_tile_count                                           );
   ENTRY_UINT  ( ., layout,              shred_tile_count                                          );
 
-  ENTRY_STR   ( ., hugetlbfs,           gigantic_page_mount_path                                  );
-  ENTRY_STR   ( ., hugetlbfs,           huge_page_mount_path                                      );
+  ENTRY_STR   ( ., hugetlbfs,           mount_path                                                );
 
   ENTRY_STR   ( ., tiles.net,           interface                                                 );
   ENTRY_STR   ( ., tiles.net,           xdp_mode                                                  );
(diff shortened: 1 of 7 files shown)


src/util/shmem/fd_shmem.h
```diff
@@ -503,8 +503,11 @@ fd_shmem_page_sz_to_cstr( ulong page_sz );
 /* These functions are for fd_shmem internal use only. */
 
 void
-fd_shmem_private_boot( int *    pargc,
-                       char *** pargv );
+fd_shmem_boot( int *    pargc,
```
mmcgee-jump, line 506: We don't really change util code like this, it's an API for other consumers outside Firedancer. You should use the API that exists already, aka. passing in a --shmem-path.

src/app/fdctl/config.c
```diff
@@ -1125,6 +1124,28 @@ config_parse( int *      pargc,
   if( FD_UNLIKELY( getgid() != 0 && config->gid != getgid() ) )
     FD_LOG_ERR(( "running as gid %i, but config specifies gid %i", getgid(), config->gid ));
 
+  const char * shmem_path = fd_env_strip_cmdline_cstr( pargc, pargv, "--shmem-path","FD_SHMEM_PATH", NULL );
```
mmcgee-jump, line 1127: I don't think we don't want to support arguments like this from fdctl. You can drop this.

src/app/fdctl/config.c
```diff
@@ -1125,6 +1124,28 @@ config_parse( int *      pargc,
...
+  if( FD_UNLIKELY( shmem_path ) ) {
+    strncpy( config->hugetlbfs.mount_path, shmem_path, sizeof( config->hugetlbfs.mount_path )-1 );
+  }
+  ulong len = strlen( config->hugetlbfs.mount_path );
```
mmcgee-jump, line 1131: This is a bit error prone and we don't need to do it, just append `/.gigantic` like below.

mmcgee-jump (changes requested):
comments inline

src/app/fdctl/main1.c
```diff
@@ -170,6 +172,14 @@ fdctl_boot( int *        pargc,
       strncpy( config->log.path, log_path, sizeof( config->log.path ) - 1 );
   }
 
+  char * args[ MAX_ARGC+2 ];
+  for( int i=0; i<*pargc; i++ ) args[ i ] = *( *pargv+i );
+  /* Override --shmem-path value from the config */
+  args[ *pargc   ] = "--shmem-path";
```
mmcgee-jump, line 178: --shmem-path is the only possible argument to fd_shmem_private_boot, so you don't need to copy args here. You can just pass a 3 element array of `["--shmem-path", path, NULL]`

src/app/fdctl/main1.c
```diff
@@ -218,6 +228,8 @@ static config_t config;
 int
 main1( int     argc,
        char ** _argv ) {
+
+  if( FD_UNLIKELY( argc >= MAX_ARGC ) ) FD_LOG_ERR(( "too many arguments (%i)", argc ));
```
mmcgee-jump, line 232: Not needed

src/app/fdctl/config.c
```diff
@@ -921,6 +920,19 @@ config_parse( int *      pargc,
   if( FD_UNLIKELY( getgid() != 0 && config->gid != getgid() ) )
     FD_LOG_ERR(( "running as gid %i, but config specifies gid %i", getgid(), config->gid ));
 
+  ulong len = strlen( config->hugetlbfs.mount_path );
+  if( FD_UNLIKELY( !len ) ) FD_LOG_ERR(( "Bad path for [hugetlbfs.mount_path] in config" ));
```
mmcgee-jump, line 924: `[hugetlbfs.mount_path] must be non-empty in your configuration file`

mmcgee-jump (changes requested):
.

mmcgee-jump (approved):

PR #1474, opened 2024-04-10
ethtool: set individual channels as well

### LINKED ISSUE #1427: ethtool: set individual channels if combined is not available
```/usr/sbin/ethtool --set-channels eno1 combined 1
netlink error: requested channel count exceeds maximum (offset 36)
netlink error: Invalid argument```

```[nbridge@fpip3-compute0951 ~]$ ethtool -l eno1
Channel parameters for eno1:
Pre-set maximums:
RX:             4
TX:             4
Other:          n/a
Combined:       n/a
Current hardware settings:
RX:             4
TX:             1
Other:          n/a
Combined:       n/a```

```[nbridge@emny-ossdev-firedancer11 firedancer]$ ethtool -l ens3f0
Channel parameters for ens3f0:
Pre-set maximums:
RX:             n/a
TX:             n/a
Other:          1
Combined:       64
Current hardware settings:
RX:             n/a
TX:             n/a
Other:          1
Combined:       1```

anwayde:
Fixes #1427

### DIFF
--- a/src/app/fdctl/configure/ethtool.c
+++ b/src/app/fdctl/configure/ethtool.c
@@ -77,15 +77,40 @@ init_device( const char * device,
     FD_LOG_ERR(( "error configuring network device, ioctl(SIOCETHTOOL,ETHTOOL_GCHANNELS) failed (%i-%s)",
                  errno, fd_io_strerror( errno ) ));
 
+  uchar success = 0;
+
   channels.combined_count = combined_channel_count;
   channels.cmd = ETHTOOL_SCHANNELS;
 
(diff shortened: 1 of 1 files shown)


mmcgee-jump:
I don't think we understand why some NICs support combined channels vs. not, so we should figure that out before making a change like this. The reason it happens will indicate the correct fix.

I also think quite likely, we want to query for the maximum of combined / rx / tx, and eg, if max_combined==1 && max_rx>1 then it's one of these special NICs, and we set the rx channel rather than combined -- rather than this try/catch/retry kind of control flow you have here instead.

mmcgee-jump (changes requested):
See comment

PR #1491, opened 2024-04-11
Fix system program mismatch (remove system transfer zero check)

mjain-jump:
Solana removed system transfer zero checks [here](https://github.com/solana-labs/solana/pull/34076).


ripatel-fd:
Let's not clean up the feature gate yet. We have tests still using it.

PR #1503, opened 2024-04-14
Genesis creation

ripatel-fd:
- **flamenco: add type utils**
- **add fd_genesis_create API**


mmcgee-jump:
I think this would look a lot cleaner without the pod, and just a regular struct, but otherwise LGTM.

mmcgee-jump (approved):

PR #1527, opened 2024-04-18
Adding accessor to change the logging concept of what the thread is

jumpsiegel:


src/util/log/fd_log.c
```diff
@@ -98,6 +98,12 @@ fd_log_private_thread_id_next( void ) {
 static FD_TL ulong fd_log_private_thread_id;      /* 0 at thread start */
 static FD_TL int   fd_log_private_thread_id_init; /* 0 at thread start */
 
+void
+fd_log_private_thread_id_set_init_reset( ulong thread_id ) {
```
mmcgee-jump, line 102: This doesn't seem to make a lot of sense. Why would you set the thread ID and then set the boolean indicating if the thread ID has been set to false at the same time? The new thread ID is never going to be used. I think whatever caller uses this is actually intending to use thread_id_set and this is a bug? But unclear without seeing the caller, can you include that here as well?

mmcgee-jump (changes requested):

src/util/log/fd_log.c
jumpsiegel: The user of this API placed it into fdctl/run.c

`int                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
main_pid_namespace( void * _args ) {                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
  struct pidns_clone_args * args = _args;                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
  if( FD_UNLIKELY( close( args->pipefd[ 0 ] ) ) ) FD_LOG_ERR(( "close() failed (%i-%s)", errno, fd_io_strerror( errno ) ));                                                                                                                                                                                                                                                                                                                                                                                                                 
  if( FD_UNLIKELY( -1!=args->closefd ) ) {                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
    if( FD_UNLIKELY( close( args->closefd ) ) ) FD_LOG_ERR(( "close() failed (%i-%s)", errno, fd_io_strerror( errno ) ));                                                                                                                                                                                                                                                                                                                                                                                                                   
  }                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
  config_t * const config = args->config;                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
  fd_log_thread_set( "pidns" );                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
  ulong pid = (ulong)getpid1(); /* Need to read /proc again.. we got a new PID from clone */                                                                                                                                                                                                                                                                                                                                                                                                                                                
  fd_log_private_group_id_set( pid );                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
  fd_log_private_thread_id_set_init_reset( pid );

mmcgee-jump: Yes, this is a bug. It doesn't set the thread_id to pid, but clears it, so the next time someone retrieves the thread ID it will be returned incorrectly. I think the. caller here just wants to do `fd_log_private_thread_id_set( pid )` ? It actually could potentially cause workspace corruption, since tiles all have a "real" thread ID of 0 in the naemspace.

jumpsiegel: @llamb-jump ?  what was the issue being solved?

jumpsiegel:
@llamb-jump ? what was the issue being solved?

PR #1528, opened 2024-04-18
improve asan support for a concept of workspace away (deepasan)

jumpsiegel:


mmcgee-jump (changes requested):
I'm a bit skeptical of these workspace ASAN changes. They were originally written at request of 1.5 after a recommendation from the security team, but they aren't on by default, and the structure of the DEEP_ASAN flag is a bit suspicious.

Let me review this one a bit closer and see if it's meeting our needs.

ripatel-fd:
Closing, already merged

PR #1529, opened 2024-04-18
global tile specific flag to get things to cleanly shut down

jumpsiegel:
This is a process specific shutdown for tiles/threads.  This should be integrated with other shutdown standards with the goal of cleanly shutting down the entire system

### DIFF
--- a/src/util/tile/fd_tile.c
+++ b/src/util/tile/fd_tile.c
@@ -3,6 +3,8 @@
 
 #include "fd_tile_private.h"
 
+volatile int fd_tile_shutdown_flag = 0;
+
 int
 fd_cpuset_getaffinity( ulong         pid,
                        fd_cpuset_t * mask ) {
--- a/src/util/tile/fd_tile.h
+++ b/src/util/tile/fd_tile.h
@@ -167,6 +167,8 @@ fd_tile_private_boot( int *    pargc,
 void
 fd_tile_private_halt( void );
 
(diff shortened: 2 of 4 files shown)


mmcgee-jump (changes requested):
I think fd_tile is probably not the right place for this shutdown flag. Exiting the executing task seems to be the model that's already supported by the code as written, (e.g. via the tile CNC) and supports threading / process models used by both 1.5 and other jump teams internally when calling this util code.

jumpsiegel:
What are your specific recommended changes?  What is the correct API we should call for setting a shutdown request and/or checking to see if it is set?

mmcgee-jump:
I think the design of tile exec, and tpool exec as intended is that clean exit is done by the child task that's executing returning to the parent, not by the parent simply halting execution. There are a few reasons that design is better.

You can conform to that design by exiting tiles with the CNC object (changing it into the shutdown state), and adding a specific flag for worker items of the tpool which have long loops you want to be able to exit during the loop. Many worker items wouldn't need it, since they exit reasonably quickly anyway.

jumpsiegel:
I also agree... this cross object/abstraction boundries in inconsistent ways.

asiegel-jt:
Regardless the scratch changes are needed to get a leak free shutdown

ripatel-fd:
I'm going to close this for now since clean shutdown is possible without 'low-level' modifications to tpool.

PR #1535, opened 2024-04-18
simplify use of the poh apis

jumpsiegel:
This one feeds sketchy.  It  eliminates an object but breaks it away from the API style of FD.


jumpsiegel:
This is not used by the poh tile. It is only used by the runtime.  I improved the documentation a little but also added a TODO to revisit this during the documentation day effort

PR #1536, opened 2024-04-18
microblock/entry abstraction

jumpsiegel:


mmcgee-jump (changes requested):
I think we already have this in the shredder code. @ptaffet-jump can these be merged?

ripatel-fd:
> I think we already have this in the shredder code. @ptaffet-jump can these be merged?

@mmcgee-jump Yes, this code is duplicated, but we can't update all the flamenco usages at the moment. I would rather have both for a short while and unify them when flamenco is in tree

lheeger-jump (approved):

jumpsiegel:
I squashed and updated the comments to remind us to zap this once the merge is complete

mmcgee-jump (approved):

PR #1537, opened 2024-04-18
convenience methods to speed up sha256 operations

jumpsiegel:


mmcgee-jump (approved):
Could use some better docs but lgtm

PR #1538, opened 2024-04-18
new textstream object

jumpsiegel:
Where are the comments?  This needs to be attacked during documentation day


lheeger-jump:
We will improve code quality via documentation after the merge is complete.

lheeger-jump (approved):

ripatel-fd:
Nevermind, can't actually move textstream since it is used in waltz, flamenco, and disco. The only common dependency of those is util. So I begrudgingly accept.

Issue #1540, opened 2024-04-18
net: device busy warnings


anwayde:
We sometimes get these errors:
```
WARNING 04-18 11:12:36.768756 109658 15   net:0 src/waltz/ip/fd_netlink.c(392): netlink returned data with error: 16 Device or resource busy
WARNING 04-18 11:13:36.768023 109658 15   net:0 src/waltz/ip/fd_netlink.c(630): netlink returned data with error: 16 Device or resource busy
```

We also get some error messages for infiniband interfaces:
```
WARNING 04-18 11:15:36.767996 109658 15   net:0 src/waltz/ip/fd_netlink.c(689): Neighbor entry has LL address with other than 6 byte MAC address
NOTICE  04-18 11:15:36.768040 109658 15   net:0 src/waltz/ip/fd_netlink.c(208): netlink diagnostic rtattr:  NDA_DST(1): 0x0797f1b7 NDA_LLADDR(2): 0x00001149fe8000000000000008c0eb0300f57a9f NDA_PROBES(4): 0x01000000 NDA_CACHEINFO(3): 0x04000000e1ff430092ff430001000000
```

ptaffet-jump:
Fixed by #1612

PR #1547, opened 2024-04-18
ARP - removed some warnings. Continue in the presence of EBUSY

nbridge-jump:
Addresses https://github.com/firedancer-io/firedancer/issues/1540


mmcgee-jump:
Why does it return EBUSY ?

nbridge-jump:
Superceded

PR #1557, opened 2024-04-19
sometimes you just want to watch the dance

jumpsiegel:


ripatel-fd:
Let's put this one on hold until we can ensure that libmicrohttpd is license-compatible with Firedancer. Most likely not.

PR #1566, opened 2024-04-20
solcap

ripatel-fd:


src/flamenco/capture/fd_solcap.pb.h
lheeger-jump, line 1: we should probably add these auto generated files to git attributes so GH flags them... not necessary just nice to collapse these diffs.

lheeger-jump (approved):
LGTM

src/flamenco/capture/fd_solcap.pb.h
CantelopePeel: sure

PR #1572, opened 2024-04-21
feature gates

ripatel-fd:
- **features: update feature map**
- **features: track hardcoded**


ptaffet-jump (approved):
Skimmed, looks good. Stamping to keep merge rolling

PR #1580, opened 2024-04-22
fddev: print genesis hash

ripatel-fd:


alpeng-jump (dismissed):

ptaffet-jump:
This function already exists here: https://github.com/firedancer-io/firedancer/blob/7a34721bb44f482ce716aeb8b7a71c765860b060/src/app/fddev/genesis_hash.h#L9

ripatel-fd:
@alpeng-jump Would you like to take over and finish this PR? Let's deduplicate the logic to compute the shred version and genesis hash as Philip suggested.

ptaffet-jump (approved):
Approving on behalf of @ripatel-fd

alpeng-jump:
Sorry was asleep then - thanks for picking up the changes @ripatel-fd

PR #1615, opened 2024-04-24
More banks hash mismatch fixes

jumpsiegel:


ripatel-fd:
Good find

ripatel-fd:
This change is fine but GitHub CI runners are down yet again.

PR #1621, opened 2024-04-24
Add new tiles

arjain4:


lheeger-jump (approved):

lheeger-jump:
These will be converted to new system in a follow up

mmcgee-jump:
I don't think copy pasting everything is a long term sustainable strategy for maintaining this code. This probably needs to be rationalized and discussed or stay on a branch.

mmcgee-jump (changes requested):
.

PR #1629, opened 2024-04-24
Fix tpool in main

asiegel-jt:


jumpsiegel (approved):

src/util/tpool/fd_tpool.cxx
```diff
@@ -29,12 +30,21 @@ fd_tpool_private_worker( int     argc,
   FD_VOLATILE( worker->state ) = FD_TPOOL_WORKER_STATE_BOOT;
   FD_COMPILER_MFENCE();
 
+  if( scratch_sz ) {
+    if( !scratch ) {
+      scratch = aligned_alloc( FD_SCRATCH_SMEM_ALIGN, fd_ulong_align_up( scratch_sz, FD_SCRATCH_SMEM_ALIGN ) );
```
mmcgee-jump, line 35: Is this change necessary? Can't the caller do the aligned alloc? Having tpool take an implicit / hidden dependency on malloc is the kind of stuff we have hated about Rust.

mmcgee-jump (changes requested):
Not sure about tpool allocating.

jumpsiegel:
this will give us a pretty huge performance improvement in the ci tests... looking forward to this

src/util/tpool/fd_tpool.cxx
asiegel-jt: Worker threads need to allocate scratch space and this is the only way to do it.

mmcgee-jump:
Can you check with Kevin if he is OK with this design?

PR #1633, opened 2024-04-25
remaining_balance underflow check re-ordering.

anarcheuz:
Re-ordering the check to avoid confusion. `remaining_balance` should not be in scope if an underflow occurs.
No security issue.


ripatel-fd:
Still conflicts ...

PR #1638, opened 2024-04-25
tiles: add repair tile

### LINKED ISSUE #1648: Repair: Use the stakes ci data structure to handle stake weights
The repair library just grabs the last sent weights instead of parsing two msgs into stakes ci as is required.

arjain4:
Adds the repair tile for executing the solana repair protocol.

Closes #1648


src/app/fdctl/run/tiles/fd_repair.c
lheeger-jump, line 1: Add a comment at the top explaining what the tile does.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,730 @@
...
+  } else {
+    FD_LOG_ERR(( "proto %lu not handled %lu", proto, in_idx ));
+    *opt_filter = 1;
+    return;
+  }
+
+  *pkt_sz = sz;
+  fd_memcpy( pkt, dcache_entry, *pkt_sz );
+  // fd_memcpy( pkt, dcache_entry+hdr_sz, sz-hdr_sz );
+  // ctx->shred_buffer_sz = sz-hdr_sz;
```
lheeger-jump, line 464:
```suggestion
```

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,730 @@
...
+  uchar * pkt;
+  ulong * pkt_sz;
+  if( FD_UNLIKELY( proto == DST_PROTO_REPAIR ) ) {
+    pkt = ctx->repair_buffer;
+    pkt_sz = &ctx->repair_buffer_sz;
+  } else {
+    FD_LOG_ERR(( "proto %lu not handled %lu", proto, in_idx ));
+    *opt_filter = 1;
+    return;
+  }
```
lheeger-jump, line 459: redundant

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,730 @@
...
+    peer_addr.addr = ip;
+    peer_addr.port = fd_ushort_bswap( hdr->udp->net_sport );
+
+    fd_repair_settime( ctx->repair, fd_log_wallclock() );
+    fd_repair_continue( ctx->repair );
+    fd_repair_recv_packet( ctx->repair, ctx->repair_buffer + hdr_sz, ctx->repair_buffer_sz - hdr_sz, &peer_addr );
+  } else {
+    FD_LOG_ERR(( "port %lu not handled %lu", proto, in_idx ));
+    *opt_filter = 1;
+    return;
```
lheeger-jump, line 522: ditto, redundant

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,730 @@
...
+
+  ulong out_cnt = 0;
+  out_fds[ out_cnt++ ] = 2; /* stderr */
+  if( FD_LIKELY( -1!=fd_log_private_logfile_fd() ) )
+    out_fds[ out_cnt++ ] = fd_log_private_logfile_fd(); /* logfile */
+  return out_cnt;
+}
+
+fd_topo_run_tile_t fd_tile_repair = {
+  .mux_flags                = FD_MUX_FLAG_COPY,
```
lheeger-jump, line 716: add manual publish flag

lheeger-jump (changes requested):

lheeger-jump (dismissed):

src/disco/fd_disco_base.h
```diff
@@ -10,6 +10,7 @@
 #define DST_PROTO_TPU_UDP  (1UL)
 #define DST_PROTO_TPU_QUIC (2UL)
 #define DST_PROTO_SHRED    (3UL)
+#define DST_PROTO_REPAIR   (4UL)
```
arjain4, line 13: this will eventually get used in fd_net.c for clarity

src/disco/topo/fd_topo.h
```diff
@@ -142,6 +142,7 @@ typedef struct {
       ushort shred_listen_port;
       ushort quic_transaction_listen_port;
       ushort legacy_transaction_listen_port;
+      ushort repair_listen_port;
```
mmcgee-jump, line 145: Dont you need both serve and listen? Or this should clarify which one

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,690 @@
...
+#include "../../../../util/net/fd_udp.h"
+
+#define NET_IN_IDX      0
+#define CONTACT_IN_IDX  1
+#define STAKE_IN_IDX    2
+#define STORE_IN_IDX    3
+
+#define STORE_OUT_IDX   0
+
+#define MAX_REPAIR_PEERS 40200UL
```
mmcgee-jump, line 33: How is this determined? What happens if someone spams?

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,690 @@
...
+
+#define SMAX    (1UL << 30UL)
+#define SDEPTH  (1UL << 11UL)
+
+struct __attribute__((packed)) fd_shred_dest_wire {
```
mmcgee-jump, line 38: Looks copy-pasted? Move to disco and share

src/disco/topo/fd_topo.h
```diff
@@ -210,6 +211,14 @@ typedef struct {
     struct {
       ulong accounts_cnt;
     } benchg;
+
+    struct {
+      char    repair_my_intake_addr[ 22 ];   // len('255.255.255.255:65535') == 22
```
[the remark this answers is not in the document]
arjain4: We are staying away from making config changes in these PRs, there will be a follow up for getting all the config stuff right since it might require a large refactor.

src/disco/topo/fd_topo.h
arjain4: For the purposes of filtering via the net tile we only need listen in this ctx. I removed the net tile change i made from this PR which is why this feels unclear.

src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -8,6 +8,7 @@
 #include "../../../../disco/keyguard/fd_keyload.h"
 #include "../../../../flamenco/leaders/fd_leaders.h"
 #include "../../../../waltz/ip/fd_ip.h"
+#include "../../../../disco/tvu/util.h"
```
mmcgee-jump, line 11: Shred tile is part of TPU as well as TVU, this would likely cause merge pain for this tile in future, and `util.h` is not a convincing place to move this. You can just use fd_disco.h or something for now.

src/disco/topo/fd_topo.h
mmcgee-jump: I meant to say serve or intake? It's unclear from "listen" which one and I still don't know..

src/disco/topo/fd_topo.h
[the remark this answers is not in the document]
mmcgee-jump: Maybe.. but it seems simple to fix here since you're adding it. The port doesn't get parsed unless you `run` the app, which prevents `configure check` from flagging invalid ip addresses, among other issues. You can see how other IP addresses are parsed in config.c and loaded into tile config.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  fd_valloc_t valloc = fd_alloc_virtual( alloc );
+
+  /* Gossip set up */
+
+  // TODO: actually get a reasonable seed
+  ulong seed = 42;
+  ctx->repair = fd_repair_join( fd_repair_new( ctx->repair, seed, valloc ) );
+
+  FD_LOG_NOTICE(( "repair my addr - intake addr: %s, serve_addr: %s", tile->repair.repair_my_intake_addr, tile->repair.repair_my_serve_addr ));
+  FD_TEST( resolve_hostport( tile->repair.repair_my_intake_addr, &ctx->repair_my_intake_addr ) );
```
mmcgee-jump, line 617: I don't understand why this parses a string like "host:port" when the config definition has a separate ip address and port fields? It doesn't look like the format is what it expects?

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  uchar src_mac_addr[6];
+
+  /* Includes Ethernet, IP, UDP headers */
+  ulong repair_buffer_sz;
+  uchar repair_buffer[ FD_NET_MTU ];
+};
+typedef struct fd_repair_tile_ctx fd_repair_tile_ctx_t;
+
+static fd_gossip_peer_addr_t *
+resolve_hostport( const char * str /* host:port */, fd_gossip_peer_addr_t * res ) {
```
mmcgee-jump, line 123: This function is problematic in a few ways and could probably just be deleted outright

 * The hand rolled string parsing is... suspicious at best, and doesn't appear necessary either.
 * You can't call gethostname here, you're inside the sandbox. Gethostname is also not going to work well for what you want to do. I'd suggest just starting with an ip address field, and defaulting it to what gets determined by the net tile anyway.
 * It's not in a firedancer code style, and eg htons can be fd_ushort_bswap, bad indentation etc

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+    return NULL;
+  }
+  res->port = htons( (ushort)port );
+
+  return res;
+}
+
+FD_FN_CONST static inline ulong
+scratch_align( void ) {
+  return 4096UL;
```
mmcgee-jump, line 163: Does this need to be 4096? I didn't look at fd_repair_align, but it just should be max of the align of things under scratch_footprint

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  return (void*)fd_ulong_align_up( (ulong)scratch, alignof(fd_repair_tile_ctx_t) );
+}
+typedef struct __attribute__((packed)) {
+  fd_eth_hdr_t eth[1];
+  fd_ip4_hdr_t ip4[1];
+  fd_udp_hdr_t udp[1];
+} eth_ip_udp_t;
+
+static inline void
+populate_packet_header_template( eth_ip_udp_t * pkt,
```
mmcgee-jump, line 194: Copy pasted from shred? Probably time to move these to a utility location, util/net would be appropriate for this IMO.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  memcpy( pkt->ip4->saddr_c, &src_ip, 4UL );
+  memset( pkt->ip4->daddr_c, 0,       4UL ); /* varies by shred */
+
+  pkt->udp->net_sport = fd_ushort_bswap( src_port );
+  pkt->udp->net_dport = (ushort)0; /* varies by shred */
+  pkt->udp->net_len   = fd_ushort_bswap( (ushort)(payload_sz + sizeof(fd_udp_hdr_t)) );
+  pkt->udp->check     = (ushort)0;
+}
+
+static void FD_FN_UNUSED
```
mmcgee-jump, line 219:
```suggestion
static void
```

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+send_packet( fd_repair_tile_ctx_t * ctx,
```
mmcgee-jump, line 220: nit: align args

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+             uint    ip,
```
mmcgee-jump, line 221:
```suggestion
             uint      dst_ip_addr,
             ushort dst_port,
```

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  eth_ip_udp_t * hdr = (eth_ip_udp_t *)packet;
+  uchar mac[6] = {0};
+  populate_packet_header_template( hdr, payload_sz, ctx->repair_my_intake_addr.addr, mac, ctx->repair_intake_listen_port );
+
+  hdr->udp->net_dport = port;
+
+  memcpy( hdr->eth->dst, mac, 6UL );
+  memcpy( hdr->ip4->daddr_c, &ip, 4UL );
+
+  // TODO: LML handle checksum correctly
```
mmcgee-jump, line 237: ??

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  ulong packet_sz = payload_sz + sizeof(eth_ip_udp_t);
+  fd_memcpy( packet+sizeof(eth_ip_udp_t), payload, payload_sz );
+
+  hdr->udp->check = fd_ip4_udp_check( *(uint *)FD_ADDRESS_OF_PACKED_MEMBER( hdr->ip4->saddr_c ), 
+                                      *(uint *)FD_ADDRESS_OF_PACKED_MEMBER( hdr->ip4->daddr_c ), 
+                                      (fd_udp_hdr_t const *)FD_ADDRESS_OF_PACKED_MEMBER( hdr->udp ), 
+                                      packet + sizeof(eth_ip_udp_t) );
+
+  ulong tspub = fd_frag_meta_ts_comp( fd_tickcount() );
+  ulong sig = fd_disco_netmux_sig( ip, port, FD_NETMUX_SIG_MIN_HDR_SZ, DST_PROTO_OUTGOING, (ushort)0 );
```
mmcgee-jump, line 249: The args here don't seem to make sense, `FD_NETMUX_SIG_MIN_HDR_SZ` in not a valid dst ip address?

src/app/fdctl/run/tiles/fd_repair.c
mmcgee-jump: Or are these meant to be src? I can't tell from the broken netmux_sig call below

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  ctx->net_out_seq   = fd_seq_inc( ctx->net_out_seq, 1UL );
+  ctx->net_out_chunk = fd_dcache_compact_next( ctx->net_out_chunk, packet_sz, ctx->net_out_chunk0, ctx->net_out_wmark );
+}
+
+static inline void
+handle_new_cluster_contact_info( fd_repair_tile_ctx_t * ctx,
+                                 uchar const    * buf ) {
+  ulong const * header = (ulong const *)fd_type_pun_const( buf );
+
+  ulong dest_cnt = header[ 0 ];
```
mmcgee-jump, line 260:
```suggestion
  ulong dest_cnt = FD_LOAD( ulong, header );
```

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  ulong tspub = fd_frag_meta_ts_comp( fd_tickcount() );
+  ulong sig = fd_disco_netmux_sig( ip, port, FD_NETMUX_SIG_MIN_HDR_SZ, DST_PROTO_OUTGOING, (ushort)0 );
+  fd_mcache_publish( ctx->net_out_mcache, ctx->net_out_depth, ctx->net_out_seq, sig, ctx->net_out_chunk, packet_sz, 0UL, tsorig, tspub );
+  ctx->net_out_seq   = fd_seq_inc( ctx->net_out_seq, 1UL );
+  ctx->net_out_chunk = fd_dcache_compact_next( ctx->net_out_chunk, packet_sz, ctx->net_out_chunk0, ctx->net_out_wmark );
+}
+
+static inline void
+handle_new_cluster_contact_info( fd_repair_tile_ctx_t * ctx,
+                                 uchar const    * buf ) {
```
mmcgee-jump, line 257: nit: align args

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  ulong const * header = (ulong const *)fd_type_pun_const( buf );
+
+  ulong dest_cnt = header[ 0 ];
+
+  if( dest_cnt >= MAX_REPAIR_PEERS ) {
+    FD_LOG_ERR(( "Cluster nodes had %lu destinations, which was more than the max of %lu", dest_cnt, MAX_REPAIR_PEERS ));
```
mmcgee-jump, line 263: Probably shouldn't bring down the validator in this case? Or is it invalid for gossip to publish more than this?

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  }
+
+  fd_shred_dest_wire_t const * in_dests = fd_type_pun_const( header+1UL );
```
mmcgee-jump, line 266: Not sure why you need to type pun? You can just cast the bytes.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+    fd_repair_add_active_peer( ctx->repair, &repair_peer, in_dests[i].pubkey );
+  }
+}
+
+static inline void
+handle_new_repair_requests( fd_repair_tile_ctx_t * ctx,
+                            uchar const    * buf,
+                            ulong buf_sz ) {
+
+  fd_repair_request_t const * repair_reqs = (fd_repair_request_t const *)fd_type_pun_const( buf );
```
mmcgee-jump, line 283: Do not need to type pun

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  ulong repair_req_cnt = buf_sz;
+  // FD_LOG_WARNING(("Repair requests %lu received %lu", repair_req_cnt, buf_sz));
+  for( ulong i = 0; i < repair_req_cnt; i++ ) {
```
mmcgee-jump, line 286: nit kbowers style

```suggestion
  for( ulong i=0UL; i<repair_req_cnt; i++ ) {
```

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+}
+
+static inline void
+handle_new_repair_requests( fd_repair_tile_ctx_t * ctx,
+                            uchar const    * buf,
+                            ulong buf_sz ) {
+
+  fd_repair_request_t const * repair_reqs = (fd_repair_request_t const *)fd_type_pun_const( buf );
+  ulong repair_req_cnt = buf_sz;
+  // FD_LOG_WARNING(("Repair requests %lu received %lu", repair_req_cnt, buf_sz));
```
mmcgee-jump, line 285: Is it a good time to remove these stray commented logging?

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+        break;
+      }
+      case FD_REPAIR_REQ_TYPE_NEED_ORPHAN: {
+        rc = fd_repair_need_orphan( ctx->repair, repair_req->slot );
+        break;
+      }
+    }
+    
+    if( rc < 0 ) {
+      FD_LOG_WARNING(( "failed to issue repair request" ));
```
mmcgee-jump, line 305: This is presumably not the right thing to do. Can an external attacker cause rc<0 ? Then logging warning is bad, as it flushes disk and stalls the tile. Is failure ok here at all? Needs comments and documentation on what's going on.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  }
+
+}
+
+static inline void
+handle_new_stake_weights( fd_repair_tile_ctx_t * ctx,
+                          uchar const    * buf ) {
+  ulong const * header = (ulong const *)fd_type_pun_const( buf );
+
+  ulong stakes_cnt = header[ 1 ];
```
mmcgee-jump, line 316: FD_LOAD per above

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+                          uchar const    * buf ) {
+  ulong const * header = (ulong const *)fd_type_pun_const( buf );
+
+  ulong stakes_cnt = header[ 1 ]; 
+
+  if( stakes_cnt >= MAX_REPAIR_PEERS ) {
+    FD_LOG_ERR(( "Cluster nodes had %lu stake weights, which was more than the max of %lu", stakes_cnt, MAX_REPAIR_PEERS ));
+  }
+
+  fd_stake_weight_t const * in_stake_weights = fd_type_pun_const( header+4UL );
```
mmcgee-jump, line 322: Actually this stuff is kind of suspect. You should probably just define a type for these overall, eg,

fd_stake_weight_t {
    ulong cnt;
    .... weights;
}

and cast (no need to type pun). You also probably want to validate the sz and not read out of bounds? You can see how the existing tiles do this for the right style, for all of the above parsing functions.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  fd_repair_set_stake_weights( ctx->repair, in_stake_weights, stakes_cnt );
+}
+
+
+static void 
+repair_send_packet( uchar const * msg, 
+                    size_t msglen, 
+                    fd_gossip_peer_addr_t const * addr, 
+                    void * arg ) {
+  ulong tsorig = fd_frag_meta_ts_comp( fd_tickcount() );
```
mmcgee-jump, line 332: I don't think this makes sense as tsorig? Which should be the time the request was received which eventually resulted in the outgoing packet (if there was one at all)

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  send_packet( arg, addr->addr, fd_ushort_bswap( addr->port ), msg, msglen, tsorig );
+}
+
+static void
+repair_shred_deliver( fd_shred_t const * shred,
+                      ulong shred_sz, 
+                      fd_repair_peer_addr_t const * from FD_PARAM_UNUSED,
```
mmcgee-jump, line 339: Why bother with unused args here?

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+}
+
+static void
+repair_shred_deliver( fd_shred_t const * shred,
+                      ulong shred_sz, 
+                      fd_repair_peer_addr_t const * from FD_PARAM_UNUSED, 
+                      fd_pubkey_t const * id FD_PARAM_UNUSED, 
+                      void * arg ) {
+  fd_repair_tile_ctx_t * ctx = (fd_repair_tile_ctx_t *)arg;
+  ulong tsorig = fd_frag_meta_ts_comp( fd_tickcount() );
```
mmcgee-jump, line 343: Per above, this is `tspub` not `tsorig`, which would be the timetamp of the request we received that caused this response.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  ctx->repair = FD_SCRATCH_ALLOC_APPEND( l, fd_repair_align(), fd_repair_footprint() );
+
+  void * smem = FD_SCRATCH_ALLOC_APPEND( l, fd_scratch_smem_align(), fd_scratch_smem_footprint( SMAX ) );
+  void * fmem = FD_SCRATCH_ALLOC_APPEND( l, fd_scratch_fmem_align(), fd_scratch_fmem_footprint( SDEPTH ) );
+
+  FD_TEST( ( !!smem ) & ( !!fmem ) );
+  fd_scratch_attach( smem, fmem, SMAX, SDEPTH );
+  
+  ulong scratch_top = FD_SCRATCH_ALLOC_FINI( l, 1UL );
+  if( FD_UNLIKELY( scratch_top > (ulong)scratch + scratch_footprint( tile ) ) )
```
mmcgee-jump, line 529: Prefer to do this check at the bottom of the init function like the other tiles (in case someone adds another scratch alloc below)

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+    FD_LOG_ERR(( "scratch overflow %lu %lu %lu", scratch_top - (ulong)scratch - scratch_footprint( tile ), scratch_top, (ulong)scratch + scratch_footprint( tile ) ));
+  
+  ctx->wksp = topo->workspaces[ topo->objs[ tile->tile_obj_id ].wksp_id ].wksp;
+
+  void * alloc_shmem = fd_wksp_alloc_laddr( ctx->wksp, fd_alloc_align(), fd_alloc_footprint(), 3UL );
```
mmcgee-jump, line 534: You don't need to place the alloc object out of the workspace, it should be part of the context object.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  FD_TEST( ctx->repair_serve_listen_port!=0 );
+
+  fd_topo_link_t * netmux_link = &topo->links[ tile->in_link_id[ 0 ] ];
+
+  ctx->net_in    = topo->workspaces[ topo->objs[ netmux_link->dcache_obj_id ].wksp_id ].wksp;
+  ctx->chunk  = fd_disco_compact_chunk0( ctx->net_in );
+  ctx->wmark  = fd_disco_compact_wmark ( ctx->net_in, netmux_link->mtu );
+
+  // TODO: make configurable
+  FD_TEST( getrandom( ctx->identity_private_key, 32UL, 0 ) == 32UL );
```
mmcgee-jump, line 552: Is this supposed to use a random public key? Not the validator identity key?

Also this will fail. getrandom is not available in the sandbox for this tile.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  ctx->repair_intake_listen_port = tile->repair.repair_intake_listen_port;
+  ctx->repair_serve_listen_port = tile->repair.repair_serve_listen_port;
+
+  FD_TEST( ctx->repair_intake_listen_port!=0 );
+  FD_TEST( ctx->repair_serve_listen_port!=0 );
+
+  fd_topo_link_t * netmux_link = &topo->links[ tile->in_link_id[ 0 ] ];
+
+  ctx->net_in    = topo->workspaces[ topo->objs[ netmux_link->dcache_obj_id ].wksp_id ].wksp;
+  ctx->chunk  = fd_disco_compact_chunk0( ctx->net_in );
```
mmcgee-jump, line 548: net_in_chunk0

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  ctx->wmark  = fd_disco_compact_wmark ( ctx->net_in, netmux_link->mtu );
```
mmcgee-jump, line 549: net_in_wmark

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+
+  fd_repair_settime( ctx->repair, fd_log_wallclock() );
+  fd_repair_continue( ctx->repair );
+}
+
+static void
+unprivileged_init( fd_topo_t *      topo,
+                   fd_topo_tile_t * tile,
+                   void *           scratch ) {
+  fd_flamenco_boot( NULL, NULL );
```
mmcgee-jump, line 496: I wouldn't trust this glibc crap to succeed in the sandbox, does this need it? Much better to not use this.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  ctx->store_out_mcache = store_out->mcache;
+  ctx->store_out_sync   = fd_mcache_seq_laddr( ctx->store_out_mcache );
+  ctx->store_out_depth  = fd_mcache_depth( ctx->store_out_mcache );
+  ctx->store_out_seq    = fd_mcache_seq_query( ctx->store_out_sync );
+  ctx->store_out_chunk0 = fd_dcache_compact_chunk0( fd_wksp_containing( store_out->dcache ), store_out->dcache );
+  ctx->store_out_mem    = topo->workspaces[ topo->objs[ store_out->dcache_obj_id ].wksp_id ].wksp;
+  ctx->store_out_wmark  = fd_dcache_compact_wmark( ctx->store_out_mem, store_out->dcache, store_out->mtu );
+  ctx->store_out_chunk  = ctx->store_out_chunk0;
+
+  fd_memcpy( ctx->src_mac_addr, tile->repair.src_mac_addr, 6 );
```
mmcgee-jump, line 579: Maybe move this actual initialization above all the link boilerplate

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  fd_memcpy( out_shred, shred, shred_sz );
+  
+  ulong tspub = fd_frag_meta_ts_comp( fd_tickcount() );
+  fd_mcache_publish( ctx->store_out_mcache, ctx->store_out_depth, ctx->store_out_seq, 1, ctx->store_out_chunk,
+    shred_sz, 0UL, tsorig, tspub );
+  ctx->store_out_seq   = fd_seq_inc( ctx->store_out_seq, 1UL );
+  ctx->store_out_chunk = fd_dcache_compact_next( ctx->store_out_chunk, shred_sz, ctx->store_out_chunk0, ctx->store_out_wmark );}
+
+static void
+repair_shred_deliver_fail( fd_pubkey_t const * id FD_PARAM_UNUSED,
```
mmcgee-jump, line 355: Why can this happen? What does it mean for a shred to fail to be delivered?

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+#include "../../../../util/fd_util.h"
+#include "../../../../disco/tvu/util.h"
+
+#include <unistd.h>
+#include <arpa/inet.h>
+#include <linux/unistd.h>
+#include <sys/random.h>
+#include <netdb.h>
+#include <netinet/in.h>
+#include <sys/socket.h>
```
mmcgee-jump, line 19: Doesn't look used?

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+                      fd_pubkey_t const * id FD_PARAM_UNUSED, 
+                      void * arg ) {
+  fd_repair_tile_ctx_t * ctx = (fd_repair_tile_ctx_t *)arg;
+  ulong tsorig = fd_frag_meta_ts_comp( fd_tickcount() );
+  
+  fd_shred_t * out_shred = fd_chunk_to_laddr( ctx->store_out_mem, ctx->store_out_chunk );
+  fd_memcpy( out_shred, shred, shred_sz );
+  
+  ulong tspub = fd_frag_meta_ts_comp( fd_tickcount() );
+  fd_mcache_publish( ctx->store_out_mcache, ctx->store_out_depth, ctx->store_out_seq, 1, ctx->store_out_chunk,
```
mmcgee-jump, line 349: You need to use `fd_mux_publish` since this is the primary out it's managed by the mux, what you have here breaks flow control and likely the general control flow of the mux tile also.

src/app/fdctl/run/tiles/fd_repair.c
mmcgee-jump, line 343: This is tspub, not tsorig, just leave the field 0 if you don't want to find tsorig

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+            ctx->repair_req_in_chunk0, ctx->repair_req_in_wmark ));
+
+    uchar const * dcache_entry = fd_chunk_to_laddr_const( ctx->repair_req_in_mem, chunk );
+    handle_new_repair_requests( ctx, dcache_entry, sz );
+    return;
+  }
+  
+  if( FD_UNLIKELY( chunk<ctx->chunk || chunk>ctx->wmark || sz>FD_NET_MTU ) ) {
+    FD_LOG_ERR(( "chunk %lu %lu corrupt, not in range [%lu,%lu]", chunk, sz, ctx->chunk, ctx->wmark ));
+    *opt_filter = 1;
```
mmcgee-jump, line 418: FD_LOG_ERR terminates the application, you don't need the stuff after this

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  }
+
+  uchar const * dcache_entry = fd_chunk_to_laddr_const( ctx->net_in, chunk );
+  ulong  hdr_sz = fd_disco_netmux_sig_hdr_sz( sig );
+  FD_TEST( hdr_sz < sz ); /* Should be ensured by the net tile */
+
+  ctx->repair_buffer_sz = sz;
+  fd_memcpy( ctx->repair_buffer, dcache_entry, sz );
+
+  *opt_filter = 0;
```
mmcgee-jump, line 429: Not needed

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+
+  return;
```
mmcgee-jump, line 431: Not needed

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+
+  fd_repair_tile_ctx_t * ctx = (fd_repair_tile_ctx_t *)_ctx;
+
+  if( FD_UNLIKELY( in_idx==CONTACT_IN_IDX ) ) {
+    if( FD_UNLIKELY( chunk<ctx->contact_in_chunk0 || chunk>ctx->contact_in_wmark ) )
+      FD_LOG_ERR(( "chunk %lu %lu corrupt, not in range [%lu,%lu]", chunk, sz,
+            ctx->contact_in_chunk0, ctx->contact_in_wmark ));
+
+    uchar const * dcache_entry = fd_chunk_to_laddr_const( ctx->contact_in_mem, chunk );
+    handle_new_cluster_contact_info( ctx, dcache_entry );
```
mmcgee-jump, line 392: We typically don't handle data like this in during_frag, because you can get overrun your consumer does not know how to detect corruption. You want to copy the data out to a buffer, then process it in after_frag. Same for the below cases.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+            ulong *            opt_tsorig FD_PARAM_UNUSED,
+            int *              opt_filter,
+            fd_mux_context_t * mux        FD_PARAM_UNUSED) {
+
+  fd_repair_tile_ctx_t * ctx = (fd_repair_tile_ctx_t *)_ctx;
+
+  *opt_filter = 1;
+
+  if( FD_UNLIKELY( in_idx==CONTACT_IN_IDX ) ) {
+    return;
```
mmcgee-jump, line 450: See above comment, the frags actually need to be handled here

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+            ulong *            opt_sig,
+            ulong *            opt_chunk  FD_PARAM_UNUSED,
+            ulong *            opt_sz     FD_PARAM_UNUSED,
+            ulong *            opt_tsorig FD_PARAM_UNUSED,
+            int *              opt_filter,
+            fd_mux_context_t * mux        FD_PARAM_UNUSED) {
+
+  fd_repair_tile_ctx_t * ctx = (fd_repair_tile_ctx_t *)_ctx;
+
+  *opt_filter = 1;
```
mmcgee-jump, line 447: ?? Not needed and will cause the tile metrics to be wrong

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  ulong hdr_sz = fd_disco_netmux_sig_hdr_sz( *opt_sig );
+  eth_ip_udp_t * hdr = (eth_ip_udp_t *)ctx->repair_buffer;
+
+  fd_repair_peer_addr_t peer_addr;
+  peer_addr.l = 0;
+  peer_addr.addr = ip;
+  peer_addr.port = fd_ushort_bswap( hdr->udp->net_sport );
+
+  fd_repair_settime( ctx->repair, fd_log_wallclock() );
+  fd_repair_continue( ctx->repair );
```
mmcgee-jump, line 471: Does this turn the internal engine of the repair module? You probably don't want to tie this to incoming frags since you might never get any, but still want to turn the engine. Continue should probably go in after_credit?

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+  fd_repair_recv_packet( ctx->repair, ctx->repair_buffer + hdr_sz, ctx->repair_buffer_sz - hdr_sz, &peer_addr );
+}
+
+static void
+privileged_init( fd_topo_t *      topo    FD_PARAM_UNUSED,
```
mmcgee-jump, line 476: Dead code

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+                 fd_topo_tile_t * tile    FD_PARAM_UNUSED,
+                 void *           scratch FD_PARAM_UNUSED) {
+
+}
+
+static void
+during_housekeeping( void * _ctx ) {
+  fd_repair_tile_ctx_t * ctx = (fd_repair_tile_ctx_t *)_ctx;
+
+  fd_mcache_seq_update( ctx->store_out_sync, ctx->store_out_seq );
```
mmcgee-jump, line 486: This is managed by the mux, you don't need to update the seq here. You do need to update the net out seq instead.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+    FD_LOG_ERR( ( "fd_alloc_join failed" ) ); 
+  }
+
+  fd_valloc_t valloc = fd_alloc_virtual( alloc );
+
+  /* Gossip set up */
+
+  // TODO: actually get a reasonable seed
+  ulong seed = 42;
+  ctx->repair = fd_repair_join( fd_repair_new( ctx->repair, seed, valloc ) );
```
mmcgee-jump, line 614: Does this thing really need to allocate? And if it does... why does it only need 1GiB? Does it fail if it hits that limit? Can an attacker trigger that and crash firedancer?

src/app/fdctl/run/tiles/fd_repair.c
asiegel-jt: My repair and gossip implementations use host:port

src/app/fdctl/run/tiles/fd_repair.c
arjain4: shifted to using stake ci

src/app/fdctl/run/tiles/fd_repair.c
arjain4: I didn't write any of the repair code but afaict, this is a callback passed into the library code so it just needs to exist. From a quick read through, does not look like we can hit a case which triggers this from the repair tile, but just having a log for completeness makes sense. This becomes a more important callback for replay

arjain4: As for when this happens, the condition is when the "needed_table" is full (i.e. too many reqs)

src/app/fdctl/run/tiles/fd_repair.c
arjain4: continue primarily sends repair requests and prints stats. i don't think this instance needs to shift

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,675 @@
...
+
+}
+
+static void
+during_housekeeping( void * _ctx ) {
+  fd_repair_tile_ctx_t * ctx = (fd_repair_tile_ctx_t *)_ctx;
+
+  fd_mcache_seq_update( ctx->store_out_sync, ctx->store_out_seq );
+
+  fd_repair_settime( ctx->repair, fd_log_wallclock() );
```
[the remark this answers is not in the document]
arjain4: this instance of continue is for regularly shuffling a part of the internal state and logging i think. i am inclined to believe that is best suited to housekeeping, but i could be wrong. would look to @lheeger-jump for clarification on this

arjain4:
Addressed all the actionable comments from @mmcgee-jump, thank you for the review. There's some open questions which i have tried to address, but @lheeger-jump and @asiegel-jt might have more comprehensive thoughts

src/app/fdctl/run/tiles/fd_repair.c
arjain4: I might need some clarification from @asiegel-jt on the reasoning behind this setup. It would also be nice to get a quick explanation on what "gets determined by the net tile" vs gethostname means in reality (or with an example).

src/app/fdctl/run/tiles/fd_repair.c
asiegel-jt: the continue is to trigger delayed actions which are time dependent, such as shuffling peers or sending pushes

src/app/fdctl/run/tiles/fd_repair.c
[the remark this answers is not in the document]
asiegel-jt: it is also used for sending pushes and delayed requests (due to throttling)

asiegel-jt (approved):

src/app/fdctl/run/tiles/fd_repair.c
asiegel-jt: New peers get ignored in that case

src/app/fdctl/run/tiles/fd_repair.c
asiegel-jt: this was a convenient syntax for command line parsing. hard coding ip addresses in configs is fragile and error prone.

src/app/fdctl/run/tiles/fd_repair.c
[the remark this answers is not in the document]
asiegel-jt: why are we calling settime/continue in two places? housekeeping should be enough I think

src/app/fdctl/run/tiles/fd_repair.c
lheeger-jump: this function is already gone on my integration branch. wait up for that where we really clean this up.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,655 @@
...
+  eth_ip_udp_t * hdr = (eth_ip_udp_t *)packet;
+  uchar mac[6] = {0};
+  populate_packet_header_template( hdr, payload_sz, ctx->repair_my_intake_addr.addr, mac, ctx->repair_intake_listen_port );
+
+  hdr->udp->net_dport = dst_port;
+
+  memcpy( hdr->eth->dst, mac, 6UL );
+  memcpy( hdr->ip4->daddr_c, &dst_ip_addr, 4UL );
+
+  // TODO: LML handle checksum correctly
```
lheeger-jump, line 210:
```suggestion
```

checksum is correct

src/app/fdctl/run/tiles/fd_repair.c
lheeger-jump: yea we can warn and return @arjain4

src/app/fdctl/run/tiles/fd_repair.c
lheeger-jump: this can be made a debug log

src/app/fdctl/run/tiles/fd_repair.c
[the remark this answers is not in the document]
lheeger-jump: this can move to after_credit

src/app/fdctl/run/tiles/fd_repair.c
lheeger-jump: this is fixed on the integration branch for both gossip and repair. recommend leaving as is and the follow up will come very shortly

src/app/fdctl/run/tiles/fd_repair.c
arjain4: In a comment above, alex mentioned that too many peers should err

lheeger-jump (changes requested):

lheeger-jump (approved):

src/app/fdctl/run/tiles/fd_repair.c
mmcgee-jump: How is "too many peers" determined? Is 40200 an arbitrary number we picked? Can you justify why that is the right number? If I sent 40201 gossip clients at the validator what happens? Let's add some commentary at least. FD_LOG_ERR terminates the application so this sounds like a DOS vector.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+#include "../../../../util/tmpl/fd_map_giant.c"
+
+
+struct fd_repair_tile_ctx {
+  fd_repair_t * repair;
+  fd_repair_config_t repair_config;
+
+  fd_repair_peer_addr_t repair_my_intake_addr;
+  fd_repair_peer_addr_t repair_my_serve_addr;
+  ushort                repair_intake_listen_port;
```
mmcgee-jump, line 69: The peer_addr_t already has a port, why do you have another one?

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+}
+
+/* Contact info table */
+#define MAP_NAME     fd_contact_info_table
+#define MAP_KEY_T    fd_pubkey_t
+#define MAP_KEY_EQ   fd_pubkey_eq
+#define MAP_KEY_HASH fd_pubkey_hash
+#define MAP_KEY_COPY fd_pubkey_copy
+#define MAP_T        fd_contact_info_elem_t
+#include "../../../../util/tmpl/fd_map_giant.c"
```
mmcgee-jump, line 60: Not sure you need to use fd_map_giant here? Do you need persistent key location? I think fd_map_dynamic is probably better?

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+struct fd_repair_tile_ctx {
+  fd_repair_t * repair;
+  fd_repair_config_t repair_config;
+
+  fd_repair_peer_addr_t repair_my_intake_addr;
+  fd_repair_peer_addr_t repair_my_serve_addr;
+  ushort                repair_intake_listen_port;
+  ushort                repair_serve_listen_port;
+
+  uchar       identity_private_key[ 32 ];
```
mmcgee-jump, line 72: Flagging, I think you already switched to signing tile?

src/util/net/util.h
```diff
@@ -0,0 +1,43 @@
...
+#define HEADER_fd_src_util_net_util_h
+
+#include "fd_udp.h"
+#include "fd_eth.h"
+
+typedef struct __attribute__((packed)) {
+  fd_eth_hdr_t eth[1];
+  fd_ip4_hdr_t ip4[1];
+  fd_udp_hdr_t udp[1];
+} eth_ip_udp_t;
```
ptaffet-jump, line 11: This definitely needs to be renamed, since it's no longer a static thing. The function needs to be renamed as well.

src/util/net/util.h
```diff
@@ -0,0 +1,43 @@
...
+
+  pkt->ip4->verihl       = FD_IP4_VERIHL( 4U, 5U );
+  pkt->ip4->tos          = (uchar)0;
+  pkt->ip4->net_tot_len  = fd_ushort_bswap( (ushort)(payload_sz + sizeof(fd_ip4_hdr_t)+sizeof(fd_udp_hdr_t)) );
+  pkt->ip4->net_frag_off = fd_ushort_bswap( FD_IP4_HDR_FRAG_OFF_DF );
+  pkt->ip4->ttl          = (uchar)64;
+  pkt->ip4->protocol     = FD_IP4_HDR_PROTOCOL_UDP;
+  pkt->ip4->check        = 0U;
+  memcpy( pkt->ip4->saddr_c, &src_ip, 4UL );
+  memset( pkt->ip4->daddr_c, 0,       4UL ); /* varies by shred */
```
ptaffet-jump, line 33:
```suggestion
  memset( pkt->ip4->daddr_c, 0,       4UL );
```

src/util/net/util.h
```diff
@@ -0,0 +1,43 @@
...
+
+typedef struct __attribute__((packed)) {
+  fd_eth_hdr_t eth[1];
+  fd_ip4_hdr_t ip4[1];
+  fd_udp_hdr_t udp[1];
+} eth_ip_udp_t;
+
+FD_PROTOTYPES_BEGIN
+
+void
```
ptaffet-jump, line 15:
```suggestion
eth_ip_udp_t *
```
It should probably return pkt.

ptaffet-jump: Also needs a doc comment

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+             uint                   dst_ip_addr,
+             ushort                 dst_port,
+             uchar const *          payload,
+             ulong                  payload_sz,
+             ulong                  tsorig ) {
+  uchar * packet = fd_chunk_to_laddr( ctx->net_out_mem, ctx->net_out_chunk );
+
+  eth_ip_udp_t * hdr = (eth_ip_udp_t *)packet;
+  uchar mac[6] = {0};
+  populate_packet_header_template( hdr, payload_sz, ctx->repair_my_intake_addr.addr, mac, ctx->repair_intake_listen_port );
```
mmcgee-jump, line 203: The purpose of this function is to "pre-stage" the packet during unprivileged_init so you don't need to rewrite the whole thing every time, then you can just memcpy the bytes into the packet buffer. See how fd_shred does this and follow along.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+
+  hdr->udp->net_dport = dst_port;
+
+  memcpy( hdr->eth->dst, mac, 6UL );
```
mmcgee-jump, line 207: Can you follow the `hfd->ip4->net_id` fragmentation thing same as fd_shred here.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+  return FD_LAYOUT_FINI( l, scratch_align() );
+}
+
+FD_FN_CONST static inline void *
+mux_ctx( void * scratch ) {
+  return (void*)fd_ulong_align_up( (ulong)scratch, alignof(fd_repair_tile_ctx_t) );
+}
+
+static void
+send_packet( fd_repair_tile_ctx_t * ctx,
```
mmcgee-jump, line 193: `send_repair_response` ?

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+
+static inline void
+handle_new_cluster_contact_info( fd_repair_tile_ctx_t * ctx,
+                                 uchar const *          buf ) {
+  ulong const * header = (ulong const *)fd_type_pun_const( buf );
+
+  ulong dest_cnt = FD_LOAD( ulong, header );
+
+  if( dest_cnt >= MAX_REPAIR_PEERS ) {
+    FD_LOG_WARNING(( "Cluster nodes had %lu destinations, which was more than the max of %lu", dest_cnt, MAX_REPAIR_PEERS ));
```
mmcgee-jump, line 235: Can the gossip tile send more than this many? Is it an error condition, or should you just drop the peers after MAX_REPAIR_PPERS from the table?

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+}
+
+static inline void
+handle_new_cluster_contact_info( fd_repair_tile_ctx_t * ctx,
+                                 uchar const *          buf ) {
+  ulong const * header = (ulong const *)fd_type_pun_const( buf );
+
+  ulong dest_cnt = FD_LOAD( ulong, header );
+
+  if( dest_cnt >= MAX_REPAIR_PEERS ) {
```
mmcgee-jump, line 234:
```suggestion
  if( FD_UNLIKELY( dest_cnt >= MAX_REPAIR_PEERS ) ) {
```

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+    FD_LOG_WARNING(( "Cluster nodes had %lu destinations, which was more than the max of %lu", dest_cnt, MAX_REPAIR_PEERS ));
+    return;
+  }
+
+  fd_shred_dest_wire_t const * in_dests = (fd_shred_dest_wire_t const *)( header+1UL );
+
+  for( ulong i=0UL; i<dest_cnt; i++ ) {
+    fd_repair_peer_addr_t repair_peer = {
+      .addr = in_dests[i].ip4_addr,
```
mmcgee-jump, line 243: You need to validator the buffer size against the offsets to make sure it's reading out of bounds

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+  ctx->repair_config.deliver_fail_fun = repair_shred_deliver_fail;
+  ctx->repair_config.send_fun = repair_send_packet;
+
+  if( fd_repair_set_config( ctx->repair, &ctx->repair_config ) ) {
+    FD_LOG_ERR( ( "error setting gossip config" ) );
+  }
+
+  fd_repair_update_addr( ctx->repair, &ctx->repair_my_intake_addr, &ctx->repair_my_serve_addr );
+
+  fd_repair_settime( ctx->repair, fd_log_wallclock() );\
```
mmcgee-jump, line 610:
```suggestion
  fd_repair_settime( ctx->repair, fd_log_wallclock() );
```

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+  void * smem = FD_SCRATCH_ALLOC_APPEND( l, fd_scratch_smem_align(), fd_scratch_smem_footprint( SMAX ) );
+  void * fmem = FD_SCRATCH_ALLOC_APPEND( l, fd_scratch_fmem_align(), fd_scratch_fmem_footprint( SDEPTH ) );
+
+  FD_TEST( ( !!smem ) & ( !!fmem ) );
+  fd_scratch_attach( smem, fmem, SMAX, SDEPTH );
+  
+  ctx->wksp = topo->workspaces[ topo->objs[ tile->tile_obj_id ].wksp_id ].wksp;
+
+  void * alloc_shmem = FD_SCRATCH_ALLOC_APPEND( l, fd_alloc_align(), fd_alloc_footprint() );
+  if( FD_UNLIKELY( !alloc_shmem ) ) {
```
mmcgee-jump, line 515: This can't fail.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+  if( FD_UNLIKELY( tile->out_link_id_primary == ULONG_MAX ) )
+    FD_LOG_ERR(( "repair tile has no primary output link" ));
+
+  /* Scratch mem setup */
+
+  FD_SCRATCH_ALLOC_INIT( l, scratch );
+  fd_repair_tile_ctx_t * ctx = FD_SCRATCH_ALLOC_APPEND( l, alignof(fd_repair_tile_ctx_t), sizeof(fd_repair_tile_ctx_t) );
+  ctx->repair = FD_SCRATCH_ALLOC_APPEND( l, fd_repair_align(), fd_repair_footprint() );
+
+  void * smem = FD_SCRATCH_ALLOC_APPEND( l, fd_scratch_smem_align(), fd_scratch_smem_footprint( SMAX ) );
```
mmcgee-jump, line 506: Can you just inline these constants here and document them. Who is using the scratch, and why do you know it's going to be large enough to serve the consumers?

mmcgee-jump: (SMAX, SDEPTH)

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+
+static inline void
+handle_new_repair_requests( fd_repair_tile_ctx_t * ctx,
+                            uchar const *          buf,
+                            ulong                  buf_sz ) {
+
+  fd_repair_request_t const * repair_reqs = (fd_repair_request_t const *) buf;
+  ulong repair_req_cnt = buf_sz;
+
+  for( ulong i=0; i<repair_req_cnt; i++ ) {
```
mmcgee-jump, line 259:
```suggestion
  for( ulong i=0UL; i<repair_req_cnt; i++ ) {
```

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+    fd_repair_request_t const * repair_req = &repair_reqs[i];
```
mmcgee-jump, line 260: You could read out of bounds here if the buf_sz < this offset. Can you validate the reads

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+  if( FD_UNLIKELY( !alloc ) ) {
+    FD_LOG_ERR( ( "fd_alloc_join failed" ) ); 
+  }
+
+  fd_valloc_t valloc = fd_alloc_virtual( alloc );
+
+  /* Gossip set up */
+
+  // TODO: actually get a reasonable seed
+  ulong seed = 42;
```
mmcgee-jump, line 592: Initialize seed during `privileged_init` with getrandom(). This hash table can be attacker controlled for hash, so you need to secure it.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+    ulong  hdr_sz = fd_disco_netmux_sig_hdr_sz( sig );
+    FD_TEST( hdr_sz < sz ); /* Should be ensured by the net tile */
+
+  } else {
+    FD_LOG_WARNING(("Unknown in_idx %lu for repair", in_idx));
+    *opt_filter = 1;
+    return;
+  }
+
+  ctx->buffer_sz = sz;
```
mmcgee-jump, line 402: You don't need to save this, the sz will be provided to after_frag `(*opt_sz)`

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+             ulong  sig,
+             ulong  chunk,
+             ulong  sz,
+             int *  opt_filter ) {
+
+  fd_repair_tile_ctx_t * ctx = (fd_repair_tile_ctx_t *)_ctx;
+  uchar const * dcache_entry;
+  if( FD_UNLIKELY( in_idx==CONTACT_IN_IDX ) ) {
+    if( FD_UNLIKELY( chunk<ctx->contact_in_chunk0 || chunk>ctx->contact_in_wmark ) ) {
+      FD_LOG_WARNING(( "chunk %lu %lu corrupt, not in range [%lu,%lu]", chunk, sz,
```
mmcgee-jump, line 356: These should all log error and abort. No need to filter the frag. Internal packet corruption is a critical error.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+  } else if ( FD_LIKELY( in_idx == NET_IN_IDX ) ) {
+    if( FD_UNLIKELY( chunk<ctx->net_in_chunk0 || chunk>ctx->net_in_wmark || sz>FD_NET_MTU ) ) {
+      FD_LOG_WARNING(( "chunk %lu %lu corrupt, not in range [%lu,%lu]", chunk, sz, ctx->net_in_chunk0, ctx->net_in_wmark ));
+      *opt_filter = 1;
+      return;
+    }
+
+    dcache_entry = fd_chunk_to_laddr_const( ctx->net_in_mem, chunk );
+    ulong  hdr_sz = fd_disco_netmux_sig_hdr_sz( sig );
+    FD_TEST( hdr_sz < sz ); /* Should be ensured by the net tile */
```
mmcgee-jump, line 394: This check is not useful here since the bytes can be torn (it can pass and later be invalid anyway).

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+
+  } else {
+    FD_LOG_WARNING(("Unknown in_idx %lu for repair", in_idx));
```
mmcgee-jump, line 397: Also an error

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+static inline void
+handle_new_stake_weights( fd_repair_tile_ctx_t * ctx ) {
+  ulong stakes_cnt = ctx->stake_ci->scratch->staked_cnt;
+
+  if( stakes_cnt >= MAX_REPAIR_PEERS ) {
+    FD_LOG_ERR(( "Cluster nodes had %lu stake weights, which was more than the max of %lu", stakes_cnt, MAX_REPAIR_PEERS ));
+  }
+
+  fd_stake_weight_t const * in_stake_weights = ctx->stake_ci->stake_weight;
+  fd_repair_set_stake_weights( ctx->repair, in_stake_weights, stakes_cnt );
```
mmcgee-jump, line 293: I can't see this function to verify if doesn't take a lifetime interest in the stake weights. Can you PR this piece before the tile that uses it?

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+    }
+  }
+
+}
+
+static inline void
+handle_new_stake_weights( fd_repair_tile_ctx_t * ctx ) {
+  ulong stakes_cnt = ctx->stake_ci->scratch->staked_cnt;
+
+  if( stakes_cnt >= MAX_REPAIR_PEERS ) {
```
mmcgee-jump, line 288: This isn't possible with stake_ci, which has a limit of 40,000 peers. Do you expect more than that? Then the stake_ci might need to change to support this.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+                      void *                        arg ) {
+  fd_repair_tile_ctx_t * ctx = (fd_repair_tile_ctx_t *)arg;
+  
+  fd_shred_t * out_shred = fd_chunk_to_laddr( ctx->store_out_mem, ctx->store_out_chunk );
+  fd_memcpy( out_shred, shred, shred_sz );
+  
+  ulong tspub = fd_frag_meta_ts_comp( fd_tickcount() );
+  ulong sig = 0UL;
+  fd_mux_publish( ctx->mux, sig, ctx->store_out_chunk, shred_sz, 0UL, 0UL, tspub );
+  ctx->store_out_chunk = fd_dcache_compact_next( ctx->store_out_chunk, shred_sz, ctx->store_out_chunk0, ctx->store_out_wmark );}
```
mmcgee-jump, line 320:
```suggestion
  ctx->store_out_chunk = fd_dcache_compact_next( ctx->store_out_chunk, shred_sz, ctx->store_out_chunk0, ctx->store_out_wmark );
  }
```

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,656 @@
...
+                      fd_pubkey_t const *           id FD_PARAM_UNUSED, 
+                      void *                        arg ) {
+  fd_repair_tile_ctx_t * ctx = (fd_repair_tile_ctx_t *)arg;
+  
+  fd_shred_t * out_shred = fd_chunk_to_laddr( ctx->store_out_mem, ctx->store_out_chunk );
+  fd_memcpy( out_shred, shred, shred_sz );
+  
+  ulong tspub = fd_frag_meta_ts_comp( fd_tickcount() );
+  ulong sig = 0UL;
+  fd_mux_publish( ctx->mux, sig, ctx->store_out_chunk, shred_sz, 0UL, 0UL, tspub );
```
mmcgee-jump, line 319: How many shreds can be requested to deliver per turn of the repair_continue? You might only have 1 flow control credit here (due to burst being set to 1 below), so if repair_continue tries to deliver two shreds this could corrupt the shreds.

src/app/fdctl/run/tiles/fd_repair.c
arjain4: Not in here unfortunately, the signing is in the integration branch and is still wip.

src/app/fdctl/run/tiles/fd_repair.c
arjain4: This isn't even used it turns out...
removing

src/app/fdctl/run/tiles/fd_repair.c
arjain4: This is def an error condition, but as we discussed earlier we don't want to shut down on this. Arbitrarily dropping peers from the table seems dangerous to me.

src/app/fdctl/run/tiles/fd_repair.c
arjain4: This check is actually a duplicate of one inside stake_ci. Removing entirely

src/app/fdctl/run/tiles/fd_repair.c
arjain4: Not sure what you mean, there is no change made to the library function so it's not possible to PR this before the repair tile.

src/app/fdctl/run/tiles/fd_repair.c
arjain4: Are you suggesting that i have local constants here because these are used in the footprint as well. Will add a note explaining how the scratch space is used

src/app/fdctl/run/tiles/fd_repair.c
arjain4: The deliver function only allows for 1 shred to be delivered

src/app/fdctl/run/tiles/fd_repair.c
arjain4: I don't think pulling changes ad-hoc from a branch which is still a wip is a good idea. Suggest coming back to this in the follow up PR.

src/flamenco/repair/fd_repair.h
```diff
@@ -10,6 +10,9 @@
 #define FD_REPAIR_DELIVER_FAIL_TIMEOUT -1
 #define FD_REPAIR_DELIVER_FAIL_REQ_LIMIT_EXCEEDED -2
 
+/* Maximum size of a network packet */
+#define PACKET_DATA_SIZE 1232
```
mmcgee-jump, line 14:
```suggestion
#define FD_REPAIR_MAX_PACKET_SIZE 1232
```

src/util/net/util.h
```diff
@@ -0,0 +1,50 @@
+#ifndef HEADER_fd_src_util_net_util_h
+#define HEADER_fd_src_util_net_util_h
+
+#include "fd_udp.h"
+#include "fd_eth.h"
+
+typedef struct __attribute__((packed)) {
```
mmcgee-jump, line 7: Can you just still all of this in `fd_udp.h` and prefix things like they are there? Eg, this struct should be named `fd_udp_ip_eth_hdr_t`

src/util/net/util.h
```diff
@@ -0,0 +1,50 @@
...
+  fd_eth_hdr_t eth[1];
+  fd_ip4_hdr_t ip4[1];
+  fd_udp_hdr_t udp[1];
+} fd_net_hdrs_t;
+
+FD_PROTOTYPES_BEGIN
+
+/**
```
mmcgee-jump, line 15: Just because this is util code can you match the formatting for the comment, eg:

```
/* Helper method to ....
    and other text... */

fd_net_hdrs_t *
```

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,621 @@
...
+#define CONTACT_IN_IDX  1
+#define STAKE_IN_IDX    2
+#define STORE_IN_IDX    3
+
+#define STORE_OUT_IDX   0
+
+#define MAX_REPAIR_PEERS 40200UL
+
+/**
+ * Scratch space is used by the repair library to allocate an
```
mmcgee-jump, line 35: Why does the repair library not declare its scratch requirements? These constants shouldn't live in the tile, and this comment is kind of pointless -- it doesn't say _why_ these constants are valid.

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,621 @@
...
+#define SMAX    (1UL << 30UL)
+#define SDEPTH  (1UL << 11UL)
+
+
+struct fd_repair_tile_ctx {
+  fd_repair_t * repair;
+  fd_repair_config_t repair_config;
+  ulong repair_seed;
+
+  fd_repair_peer_addr_t repair_my_intake_addr;
```
mmcgee-jump, line 48:
```suggestion
  fd_repair_peer_addr_t intake_addr;
  fd_repair_peer_addr_t serve_addr;
```

src/app/fdctl/run/tiles/fd_repair.c
```diff
@@ -0,0 +1,621 @@
...
+             ulong  seq FD_PARAM_UNUSED,
+             ulong  sig FD_PARAM_UNUSED,
+             ulong  chunk,
+             ulong  sz,
+             int *  opt_filter FD_PARAM_UNUSED ) {
+
+  fd_repair_tile_ctx_t * ctx = (fd_repair_tile_ctx_t *)_ctx;
+  uchar const * dcache_entry;
+  if( FD_UNLIKELY( in_idx==CONTACT_IN_IDX ) ) {
+    if( FD_UNLIKELY( chunk<ctx->contact_in_chunk0 || chunk>ctx->contact_in_wmark ) ) {
```
mmcgee-jump, line 340: You need to check that sz is less than the MTU here as well like all the other tiles, or corruption is still possible (chunk<wmark but *sz>MTU)

mmcgee-jump: (same for the other branches)

src/app/fdctl/run/tiles/fd_repair.c
mmcgee-jump: I don't understand what's going on here. Are you saying the store tile is sending frags with the meaning of the `sz` field changed to be the number of requests rather than the actual fragment size? Even if you did that (which you shouldn't), it's still wrong - you aren't validating that the memory you are reading here would be in bounds. I can trivially construct a fragment (eg, by sending you a packet with sz field of ULONG_MAX) which crashes your tile.

src/app/fdctl/run/tiles/fd_repair.c
arjain4: I am happy to shift these to the library header and rename them. Unfortunately, these constants were picked last year (most likely by Alex) and my guess is the reasoning was "these look large enough for this table." It's also possible that they were picked once and the scratch requirements changed since. I am not comfortable changing them since we know they work, and there isn't a concrete reasoning yet. I can make sure that while testing our topology on the integration branch we mess around with these or reason about them properly and update this comment, but this will have to do for now.

src/util/net/util.h
arjain4: I am not sure this is the best idea. I would then have to tie fd_udp.h to fd_eth.h as well to pull in the header definition. This struct uses all three and having a separate file makes most sense. I also think fd_net_hdrs is a more succinct but clear name

src/app/fdctl/run/tiles/fd_repair.c
arjain4: Adding a todo for this, the MTUs aren't set in stone yet

src/util/net/util.h
```diff
@@ -0,0 +1,50 @@
...
+ * Helper method to populate a header template
+ * containing ethernet, UDP and IP headers with
+ * default values. Used for pre-staging headers.
+*/
+fd_net_hdrs_t *
+fd_net_create_packet_header_template( fd_net_hdrs_t * pkt,
```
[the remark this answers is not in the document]
arjain4: As per above comment and Phillip's suggestion, I think using the more generic "net" prefix for these is better.

src/util/net/util.h
mmcgee-jump: This one part if util code so making it very precise and a higher bar than our other changes is important. There are no `util.h` type of files in util/, everything has a purpose (take a look around). So this should similarly be named in a purposeful way.

src/app/fdctl/run/tiles/fd_repair.c
arjain4: Updated the comment

lheeger-jump (approved):
I am going to approve this as it seems like we have had a lot of really good feedback we can use to apply to the other tiles. Thanks @arjain4 and @mmcgee-jump

src/util/net/util.h
arjain4: renamed

lheeger-jump (approved):

PR #1644, opened 2024-04-25
disable_optimizer: disable optimization via environmental variable

jumpsiegel:


config/extra/with-optimization.mk
```diff
@@ -1,4 +1,9 @@
+ifeq ($(DISABLE_OPTIMIZATION),)
```
mmcgee-jump, line 1:
```suggestion
ifeq ($(FD_DISABLE_OPTIMIZATION),)
```

mmcgee-jump (changes requested):
Thank you! Very useful

config/extra/with-optimization.mk
jumpsiegel: good suggestion.. done

mmcgee-jump (approved):

PR #1649, opened 2024-04-26
fd_shred: add metrics to tile and fd_fec_resolver

ptaffet-jump:
Add the following metric types:
```
# HELP shred_cluster_contact_info_cnt Number of contact infos in the cluster contact info message
# HELP shred_microblocks_abandoned The number of microblocks that were abandoned because we switched slots without finishing the current slot
# HELP shred_batch_sz The size (in bytes) of each microblock batch that is shredded
# HELP shred_batch_microblock_cnt The number of microblocks in each microblock batch that is shredded
# HELP shred_shredding_duration_seconds Duration of producing one FEC set from the shredder
# HELP shred_add_shred_duration_seconds Duration of verifying and processing one shred received from the network
# HELP shred_shred_processed_bad_slot The result of processing a thread from the network (Shred was for a slot for which we don't know the leader)
# HELP shred_shred_processed_parse_failed The result of processing a thread from the network (Shred parsing failed)
# HELP shred_shred_processed_rejected The result of processing a thread from the network (Shred was invalid for one of many reasons)
# HELP shred_shred_processed_ignored The result of processing a thread from the network (Shred was ignored because we had already received or reconstructed it)
# HELP shred_shred_processed_okay The result of processing a thread from the network (Shred accepted to an incomplete FEC set)
# HELP shred_shred_processed_completes The result of processing a thread from the network (Shred accepted and resulted in a valid, complete FEC set)
# HELP shred_fec_set_spilled The number of FEC sets that were spilled because they didn't complete in time and we needed space
# HELP shred_shred_rejected_initial The number shreds that were rejected before any resources were allocated for the FEC set
# HELP shred_fec_rejected_fatal The number of FEC sets that were rejected for reasons that cause the whole FEC set to become invalid
```


src/disco/shred/fd_fec_resolver.c
```diff
@@ -402,6 +407,7 @@ int fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
       bmtrlist_push_head( bmtree_free_list, bmtree_mem );
       return FD_FEC_RESOLVER_SHRED_REJECTED;
     }
+    FD_MCNT_INC( SHRED, SHRED_REJECTED_INITIAL, (ulong)(-1L) ); /* Undo bad speculation */
```
mmcgee-jump, line 410: This is not valid, a counter cannot ever decrease in prometheus or the collector will assume that the application has restarted.

src/disco/shred/fd_fec_resolver.c
```diff
@@ -568,6 +575,8 @@ int fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
     return FD_FEC_RESOLVER_SHRED_REJECTED;
   }
 
+  FD_MCNT_INC( SHRED, FEC_REJECTED_FATAL, (ulong)(-1L) ); /* It wasn't rejected, so undo the bad speculation */
```
mmcgee-jump, line 578: As above

src/app/fdctl/run/tiles/fd_metric.c
```diff
@@ -263,6 +263,9 @@ prometheus_print( fd_topo_t * topo,
   PRINT( "\n" );
   result = prometheus_print1( topo, out, out_len, "store", FD_METRICS_STORE_TOTAL, FD_METRICS_STORE, PRINT_TILE );
   if( FD_UNLIKELY( result<0 ) ) return result;
+  PRINT( "\n" );
```
mmcgee-jump, line 266: I have some notion to put `shred` before `store` in the output to preserve the ordering of the tiles in a fragment flow.

mmcgee-jump (changes requested):
.

src/disco/shred/fd_fec_resolver.c
ptaffet-jump: Ah, I forgot the metrics tile could actually observe the count in between these. I was thinking it required metric write, but that's not true.

src/app/fdctl/run/tiles/fd_metric.c
ptaffet-jump: Yeah, seems reasonable.

mmcgee-jump (approved):

PR #1652, opened 2024-04-26
Change value alignment to 8 in funky

asiegel-jt:
Kevin found an issue with unaligned fd_alloc calls


jumpsiegel (approved):

jumpsiegel (approved):

kbowers-jump:
To clarify (from internal discussion), there's no issue fine grained allocations but no objection to this default (it matches typical malloc behavior ... it might help performance in some situations but be slightly less memory efficient in others ... YMMV).

PR #1655, opened 2024-04-27
added quic support to fddev bench. Numerous QUIC fixes.

nbridge-jump:
bench code needs the following, but putting these into another PR:
  quic_enable - hard coded flag to use quic rather than udp: should be configurable
  only supports one connection. Each benchs should have a connection for each fd_quic tile

This PR should help us move forward with the demo

### DIFF
--- a/src/app/fddev/tiles/fd_benchs.c
+++ b/src/app/fddev/tiles/fd_benchs.c
@@ -1,9 +1,72 @@
+/* _GNU_SOURCE for recvmmsg and sendmmsg */
+#define _GNU_SOURCE
+
 #include "../../fdctl/run/tiles/tiles.h"
+#include "../../../waltz/xdp/fd_xsk_aio.h"
+#include "../../../waltz/quic/fd_quic.h"
+#include "../../../waltz/tls/test_tls_helper.h"
 
 #include <linux/unistd.h>
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <netinet/in.h>
+#include <string.h>
+#include <unistd.h>
+#include <poll.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <time.h>
+
+/* max number of buffers batched for receive */
+#define IO_VEC_CNT 128
+
+
+struct signer_ctx {
+  fd_sha512_t sha512[ 1 ];
+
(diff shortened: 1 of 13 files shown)


src/app/fddev/tiles/fd_benchs.c
```diff
@@ -1,32 +1,337 @@
...
+
+  signer_ctx_t     signer_ctx;
+  fd_quic_t *      quic;
+  ushort           quic_port;
+  fd_quic_conn_t * quic_conn;
+  const fd_aio_t * quic_rx_aio;
+  ulong            no_stream;
+
+  // vector receive members
+  struct mmsghdr rx_msgs[IO_VEC_CNT];
```
mmcgee-jump, line 93: Can we move all this QUIC stuff out to another file or like, `fd_quic_client_t` ? I feel like there could be a simple minimal interface which the bench tile can call without needing all of this stuff in its namespace.

src/app/fddev/tiles/fd_benchs.c
```diff
@@ -1,32 +1,337 @@
...
+};
+typedef struct signer_ctx signer_ctx_t;
+
+
+static void
+signer( void *        _ctx,
+        uchar         signature[ static 64 ],
+        uchar const   payload[ static 130 ] ) {
+  fd_tls_test_sign_ctx_t * ctx = (fd_tls_test_sign_ctx_t *)_ctx;
+  fd_ed25519_sign( signature, payload, 130UL, ctx->public_key, ctx->private_key, ctx->sha512 );
```
mmcgee-jump, line 44: I'm not sure about signing on the benchs. The design here is that benchg is scalable and does signing, and benchs just sends on the socket (which is very expensive). If you are mixing those two things it makes the scaling story for the bench topology much more complicated.

nbridge-jump: you think doing it in bencho is a better choice?

nbridge-jump: you know, this signing only occurs during handshaking

mmcgee-jump: Oh, hmm. Then after that it's just TLS or whatever?

PR #1656, opened 2024-04-27
util: belt-sand fd_deque API additions

ripatel-fd:
- Add missing documentation
- Make oob peek_index U.B. instead of NULL return
- Clean up method names
- Unit test reverse iterator and peek index
- Unit test pop index
- Fix deque binary search in vote program


mmcgee-jump:
Can you split the util / flamenco commits?

ripatel-fd:
> Can you split the util / flamenco commits?

@mmcgee-jump No, sorry, one commit can't work without the other and they would get squashed together anyways. And then mess up the squashed commit's message. I preferred to keep rebase but the runtime team overruled me. Their argument was that Git bisect will work more reliably if not split up.

ibhatt-jumptrading (approved):

PR #1659, opened 2024-04-27
runtime: vote program hardening, support 'tower sync'

ripatel-fd:
- types: fd_treap and fd_deque alloc helper to accept custom max
- sysvar: define max element count for 'slot hashes', 'recent block hashes', and 'stake history'
- types: change collection 'max' to 'min', update deserializer to always allocate enough capacity for collections
- types: remove 'SMALL_DEQUE' error
- program: support vote program 'tower sync' instructions

A note on hardening:

Previously, Firedancer used Agave's size limits for various dynamically sized bincode data structures, such as the lockouts deque of the vote instruction and vote state.

A malicious user could provide a oversize structure in instruction data though.  Similarly, a fuzzer could provide a corrupted vote state, though this is assumed to not be possible on a real network.

We did not handle this case correctly and would instead fail to deserialize the input (potentially leading to a different execution result).

The real item count limit of dynamically sized structures in native programs is derived from size limitations:  instruction data size is ~1.2kB max, account size is 10MiB max.

This patch updates fd_types to consider the previous 'max' item count limits as minimums instead.  If any larger structure is provided, the deserializer will allocate as much as needed.  (Provably bounded as the deserializer always advances when it makes an allocation, and verifies that the serialization is valid before allocating)

This handles all cases correctly.  If the user provides 0 vote lockouts, we allocate 31 (the minimum), which is enough space to correctly execute native program business logic, which might append more entries to the structure.  If the user provides 100 vote lockouts, we allocate 100, as that exceeds the minimum.

However, we do still need to ensure that the native program never attempts to add more items to a collection in the oversize case. This is out of scope for this PR.

### DIFF
--- a/src/flamenco/types/fd_types.c
+++ b/src/flamenco/types/fd_types.c
@@ -1552,8 +1552,7 @@ int fd_stake_history_decode_preflight(fd_bincode_decode_ctx_t * ctx) {
   int err;
   ulong fd_stake_history_treap_len;
   err = fd_bincode_uint64_decode(&fd_stake_history_treap_len, ctx);
-  if ( FD_UNLIKELY(err) ) return err;
-  if ( fd_stake_history_treap_len > FD_STAKE_HISTORY_MAX ) return FD_BINCODE_ERR_SMALL_DEQUE;
+  if( FD_UNLIKELY( err ) ) return err;
   for (ulong i = 0; i < fd_stake_history_treap_len; ++i) {
     err = fd_stake_history_entry_decode_preflight( ctx );
     if ( FD_UNLIKELY ( err ) ) return err;
@@ -1563,8 +1562,9 @@ int fd_stake_history_decode_preflight(fd_bincode_decode_ctx_t * ctx) {
 void fd_stake_history_decode_unsafe(fd_stake_history_t* self, fd_bincode_decode_ctx_t * ctx) {
   ulong fd_stake_history_treap_len;
   fd_bincode_uint64_decode_unsafe(&fd_stake_history_treap_len, ctx);
-  self->pool = fd_stake_history_pool_alloc( ctx->valloc );
-  self->treap = fd_stake_history_treap_alloc( ctx->valloc );
+  ulong fd_stake_history_treap_max = fd_ulong_max( fd_stake_history_treap_len, FD_STAKE_HISTORY_MIN );
+  self->pool = fd_stake_history_pool_alloc( ctx->valloc, fd_stake_history_treap_max );
+  self->treap = fd_stake_history_treap_alloc( ctx->valloc, fd_stake_history_treap_max );
   for (ulong i = 0; i < fd_stake_history_treap_len; ++i) {
     fd_stake_history_entry_t * ele = fd_stake_history_pool_ele_acquire( self->pool );
     fd_stake_history_entry_new( ele );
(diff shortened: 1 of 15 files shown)


src/flamenco/types/fd_types.c
```diff
@@ -1521,8 +1520,9 @@ int fd_stake_history_decode_preflight(fd_bincode_decode_ctx_t * ctx) {
 void fd_stake_history_decode_unsafe(fd_stake_history_t* self, fd_bincode_decode_ctx_t * ctx) {
   ulong fd_stake_history_treap_len;
   fd_bincode_uint64_decode_unsafe(&fd_stake_history_treap_len, ctx);
-  self->pool = fd_stake_history_pool_alloc( ctx->valloc );
-  self->treap = fd_stake_history_treap_alloc( ctx->valloc );
+  ulong fd_stake_history_treap_max = fd_ulong_max( fd_stake_history_treap_len, FD_STAKE_HISTORY_MIN );
+  self->pool = fd_stake_history_pool_alloc( ctx->valloc, fd_stake_history_treap_max );
```
topointon-jump, line 1524: > A malicious user could provide a oversize structure in instruction data though. Similarly, a fuzzer could provide a corrupted vote state, though this is assumed to not be possible on a real network.
>
> We did not handle this case correctly and would instead fail to deserialize the input (potentially leading to a different execution result).

Just so I understand, is the "potentially different execution result" you're referring to here a case where Agave succeeds and we fail? I imagine it's possible to construct a case where Agave "drops" the extra values, and we fail at the deserialization stage. Just curious as to exactly what you saw whilst fuzzing.

src/flamenco/types/gen_stubs.py
```diff
@@ -546,7 +546,7 @@ def __init__(self, container, json):
         self.name = json["name"]
         self.element = json["element"]
         self.compact = ("modifier" in json and json["modifier"] == "compact")
-        self.max = (json["max"] if "max" in json else None)
+        self.min = json.get("min", None)
```
topointon-jump, line 549: Instead of providing max instead of min, should we instead give the option to provide max OR min?

I'm a bit concerned that in the future developers may inadvertently use these data structures outside of the native programs, even if they aren't today, leading to unbounded allocations. In those cases, max should be used. Not sure how much this paranoia is justified.

topointon-jump, line 549: Do you mind copying the explanation from the PR description as to why the deserialisation is provably bounded, so that future developers are cognisant of this?

asiegel-jt: We can't stop people from mis-using data types.

src/flamenco/types/fd_types.c
topointon-jump: Thanks for the context!

src/flamenco/types/gen_stubs.py
topointon-jump: Maybe in `gen_stubs.py`? Or in fd_types.h, once. A comment in gen_stubs is probably fine though :)

topointon-jump (approved):

src/flamenco/types/gen_stubs.py
topointon-jump: Maybe we should add this function in a follow-up PR, so we have easy-to-understand limits on the allocations?

PR #1662, opened 2024-04-29
contrib: ledger conformance cli tool

alpeng-jump:
Introduces a CLI to consolidate bank hash mismatch debugging operations


llamb-jump:
We can speed up the `gsutil cp` commands with:

```
gsutil -o GSUtil:parallel_composite_upload_threshold=1 -m cp -r ...
```

jumpsiegel (dismissed):

contrib/ledger-tests/solcap.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
ibhatt-jumptrading, line 1: Can we add a cli flag for verbosity levels.  A lot of the transaction information gets printed with -v 4

alpeng-jump:
@llamb-jump We might be able to speed up the uploads with that, but object composition comes with some limitations too ( https://cloud.google.com/storage/docs/parallel-composite-uploads). Let's stick with the existing command since we don't expect to have too many uploads, and these are fairly small as they are minimized one block on each side.

PR #1670, opened 2024-04-29
fd_netlink - Handling multipart netlink responses

nbridge-jump:


src/waltz/ip/fd_netlink.c
```diff
@@ -588,136 +588,146 @@ fd_nl_load_arp_table( fd_nl_t *           nl,
...
+	/* error occurred - don't switch to new routing table */
 
+	return FD_IP_ERROR; /* return failure */
       }
 
-      if( entry->flags & FD_NL_ARP_FLAGS_UNSUPPORTED ) {
-        break;
+      /* we're done if:
+       *        type is NLMSG_DONE
+       *    OR  nlmsg_flags does not contain NLMSG_DONE */
```
mmcgee-jump, line 648:
```suggestion
       *    OR  nlmsg_flags does not contain NLM_F_MULTI */
```

mmcgee-jump:
Can you squash

src/waltz/ip/fd_netlink.c
```diff
@@ -588,136 +588,146 @@ fd_nl_load_arp_table( fd_nl_t *           nl,
...
+
+	rat = RTA_NEXT( rat, ratmsglen );
       }
 
-      rat = RTA_NEXT( rat, ratmsglen );
-    }
+      /* at present, each entry must have at least the following:
+	 FD_NL_ARP_FLAGS_IP_ADDR
+	 FD_NL_ARP_FLAGS_MAC_ADDR
+	 FD_NL_ARP_FLAGS_IFINDEX
```
ptaffet-jump, line 720: Is this supposed to be enforced?

src/waltz/ip/fd_netlink.c
```diff
@@ -588,136 +588,146 @@ fd_nl_load_arp_table( fd_nl_t *           nl,
...
 
+	return FD_IP_ERROR; /* return failure */
       }
 
-      if( entry->flags & FD_NL_ARP_FLAGS_UNSUPPORTED ) {
-        break;
+      /* we're done if:
+       *        type is NLMSG_DONE
+       *    OR  nlmsg_flags does not contain NLM_F_MULTI */
+      done |= ( h->nlmsg_type == NLMSG_DONE ) | !( h->nlmsg_flags & NLM_F_MULTI );
```
ptaffet-jump, line 649: If `h->nlmsg_type == NLMSG_DONE`, should we break? I think the while loop below might have 0 iterations, but we'll still insert an entry.

nbridge-jump: The flags NLMSG_DONE is set on a valid message, so we need to finish iterating thru the attributes.

ptaffet-jump (approved):

PR #1675, opened 2024-04-29
tiles: adds remote signing to sign tile for repair and gossip

arjain4:


src/disco/keyguard/fd_keyguard_match.c
```diff
@@ -161,6 +161,8 @@ FD_FN_PURE int
 fd_keyguard_payload_matches_gossip_msg( uchar const * data,
                                         ulong         sz ) {
 
+  // TODO: this causes potential ambiguity with the shred messages
+  if ( sz==32 ) return 1;
```
lheeger-jump, line 165: what should we do about this? is it an actual issue?

lheeger-jump (approved):

src/disco/keyguard/fd_keyguard_match.c
arjain4: I copied the todo over, but like i said when we added this todo, I don't think it is an issue. The gossip msg method is only hit for gossip messages (marked with gossip role), so there's no real ambiguity

PR #1683, opened 2024-04-30
runtime: fix bugs in Protobuf instruction context input / output

mjain-jump:
* Refactor spitting out Protobuf messages
* Fix bugs causing mismatches when creating instruction context

### DIFF
--- a/src/flamenco/runtime/tests/fd_exec_instr_test.c
+++ b/src/flamenco/runtime/tests/fd_exec_instr_test.c
@@ -5,6 +5,8 @@
 #include "../fd_acc_mgr.h"
 #include "../fd_account.h"
 #include "../fd_executor.h"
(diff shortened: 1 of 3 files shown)


src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -224,7 +229,7 @@ _context_create( fd_exec_instr_test_runner_t *        runner,
   txn_ctx->slot_ctx  = slot_ctx;
   txn_ctx->funk_txn  = funk_txn;
   txn_ctx->acc_mgr   = acc_mgr;
-  txn_ctx->valloc    = fd_scratch_virtual();
+  txn_ctx->valloc    = fd_libc_alloc_virtual();
```
topointon-jump, line 232: why are we using libc here?

PR #1687, opened 2024-04-30
afl-gcc-plugin machine targets

nlgripto:
, modified with-fuzz to only add those flags when we're not using AFLGCC


ripatel-fd:
@nlgripto Do you still want to merge this PR?

nlgripto:
yeah I haven't tested it since I PR'd it but it needs to happen at some
point

ripatel-fd:
Closing stale PR

PR #1691, opened 2024-05-01
Rejiggered tpool API to reflect actual use case

asiegel-jt:
The existing tpool API makes invalid assumptions about how tpool and tiles should interact. They do not reflect the real use case. This PR changes the construction of tpools to allow a tpool to run inside a tile, and allows a single tile to span multiple cores.

In particular, the replay tile wants to own 10 cores, not 1, so it can efficiently execute very large numbers of small tasks.


src/util/tpool/test_tpool.c
```diff
@@ -195,60 +175,28 @@ main( int     argc,
 
   FD_LOG_NOTICE(( "Testing push" ));
 
-  ulong tile_cnt = fd_tile_cnt();
+  ulong tile_cnt = 64;
   tpool = fd_tpool_init( tpool_mem, tile_cnt ); FD_TEST( tpool );
```
jumpsiegel, line 179: Kevin, should we just use the number of cpu's on the box?

jumpsiegel (approved):

jumpsiegel:
Obviously, we will squash this before we merge...

src/util/tpool/test_tpool.c
asiegel-jt: Actually, when I set this to 65, the test gets stuck. I was hoping Kevin could help me figure out why. It was very confusing.

mmcgee-jump (changes requested):
> /* tpool provides APIs to group sets of tiles together for ultra low
> overhead and high scalability launching of thread parallel jobs.

As with all kbowers code I have seen, the design and comment were already correct, and hacking the API a little bit doesn't change the functionality tpool provides.  Why can't we update the caller to distribute jobs to tiles correctly? It would take an hour or two.

mmcgee-jump:
The way this should work is that there are N `tvu_worker` tiles, rather than threadpool threads.  You can set these up correctly with a somewhat simple change to the topology,

```
runtime_obj = fd_topob_add_obj( shared_object_for_runtime )
for ( i=0; i<num_replay_worker_tiles; i++) {
  fd_topob_tile( replay_worker, i, ...);
  fd_topob_uses_obj( replay_worker, i, runtime_obj, MODE_WRITE );
}
```

Now you have N worker tiles, with their own sandbox and CPU core assigned from program affinity, which show up in the monitor and can have metrics and all the other goodness, and they also map in your "runtime_obj" to their address spaces so they can share state just like before.

src/util/tpool/test_tpool.c
```diff
@@ -195,60 +175,28 @@ main( int     argc,
 
   FD_LOG_NOTICE(( "Testing push" ));
 
-  ulong tile_cnt = fd_tile_cnt();
+  ulong tile_cnt = 64;
```
lheeger-jump, line 178: this seems wrong

lheeger-jump (changes requested):

asiegel-jt:
> The way this should work is that there are N `tvu_worker` tiles, rather than threadpool threads. You can set these up correctly with a somewhat simple change to the topology,
> 
> ```
> runtime_obj = fd_topob_add_obj( shared_object_for_runtime )
> for ( i=0; i<num_replay_worker_tiles; i++) {
>   fd_topob_tile( replay_worker, i, ...);
>   fd_topob_uses_obj( replay_worker, i, runtime_obj, MODE_WRITE );
> }
> ```
> 
> Now you have N worker tiles, with their own sandbox and CPU core assigned from program affinity, which show up in the monitor and can have metrics and all the other goodness, and they also map in your "runtime_obj" to their address spaces so they can share state just like before.

This approach would be vastly more complicated and less efficient. We are dispatching small tasks which access a shared data structure.

asiegel-jt:
> > /* tpool provides APIs to group sets of tiles together for ultra low
> > overhead and high scalability launching of thread parallel jobs.
> 
> As with all kbowers code I have seen, the design and comment were already correct, and hacking the API a little bit doesn't change the functionality tpool provides. Why can't we update the caller to distribute jobs to tiles correctly? It would take an hour or two.

It would take much longer than an hour or two. We would have to refactor a bunch of shared data structures and implement very complicated message passing. This work would buy us no additional functionality.

mmcgee-jump:
You would still be dispatching small tasks that access a shared data structure! There is no performance impact, and no message passing. All code stays the same, except you are launching tiles that access shared memory rather than tpool threads.

As for it taking longer, the problems caused by this tpool hack in tvu have been ongoing for months now, and seem to just create more work than they ever end up solving. It also does buy you a lot of new functionality, for example support for monitoring, telemetry, and and metrics tools, as well as standard configuration support and using the shared affinity assignment.

ripatel-fd:
@asiegel-jt Could you squash your commits before we merge? Thanks

PR #1692, opened 2024-05-01
flamenco: adds txn generation methods

arjain4:
Transaction generation for sending votes etc

### DIFF
--- a/src/flamenco/txn/fd_txn_generate.c
+++ b/src/flamenco/txn/fd_txn_generate.c
@@ -1,0 +1,144 @@
+#include "fd_txn_generate.h"
+
+/* Message header type */
+struct __attribute__((packed)) fd_txn_message_hdr {
+  uchar num_signatures;
+  uchar num_readonly_signatures;
+  uchar num_readonly_unsigned;
+};
+
+typedef struct fd_txn_message_hdr fd_txn_message_hdr_t;
+
+static fd_txn_instr_t *
+fd_txn_instr_meta_generate( uchar * out_buf,
+                            uchar program_id,
+                            ushort acct_cnt,
+                            ushort data_sz,
+                            ushort acct_off,
+                            ushort data_off ) {
+    fd_txn_instr_t * out_instr = (fd_txn_instr_t *) out_buf;
+    out_instr->program_id = program_id;
+    out_instr->acct_cnt   = acct_cnt;
+    out_instr->data_sz    = data_sz;
+    out_instr->acct_off   = acct_off;
+    out_instr->data_off   = data_off;
+    return out_instr;
+}
+
+ulong
+fd_txn_base_generate( uchar out_txn_meta[ static FD_TXN_MAX_SZ ],
+                      uchar out_txn_payload[ static FD_TXN_MTU ],
+                      ulong num_signatures,
+                      fd_txn_accounts_t * accounts,
+                      uchar * opt_recent_blockhash ) {
+
+  FD_TEST(num_signatures <= 128);
+  *out_txn_payload = (uchar)num_signatures;
+  /* Fill out txn metadata */
+  fd_txn_t * txn_meta             = (fd_txn_t *) out_txn_meta;
+  txn_meta->acct_addr_cnt         = accounts->acct_cnt;
+  txn_meta->readonly_signed_cnt   = accounts->readonly_signed_cnt;
+  txn_meta->readonly_unsigned_cnt = accounts->readonly_unsigned_cnt;
+  /* Number of signatures is encoded as a compact u16 but
+     can always be encoded using 1 byte here. */
+  txn_meta->message_off           = (ushort)(num_signatures * FD_TXN_SIGNATURE_SZ  + 1);
+  txn_meta->signature_off         = (ushort)1UL;
+  txn_meta->instr_cnt             = 0;
+
+  FD_TEST(txn_meta->acct_addr_cnt <= 128);
+  txn_meta->acct_addr_off        = (ushort)(txn_meta->message_off + (sizeof(fd_txn_message_hdr_t)) + 1);
+  txn_meta->recent_blockhash_off = (ushort)(txn_meta->acct_addr_off + (txn_meta->acct_addr_cnt * FD_TXN_ACCT_ADDR_SZ));
+
+  /* Fill message header in txn payload */
+  uchar * write_ptr = out_txn_payload + txn_meta->message_off;
+  fd_txn_message_hdr_t msg_header = { .num_signatures = accounts->signature_cnt,
+                                      .num_readonly_signatures = accounts->readonly_signed_cnt,
+                                      .num_readonly_unsigned   = accounts->readonly_unsigned_cnt };
+  memcpy( write_ptr, &msg_header, sizeof(fd_txn_message_hdr_t) ) ;
+  write_ptr += sizeof(fd_txn_message_hdr_t);
+
+  /* Write number of accounts */
+  *write_ptr = (uchar)txn_meta->acct_addr_cnt;
+  write_ptr += 1;
+
+  /* Write accounts list to txn payload */
+  ulong signers_write_sz = FD_TXN_ACCT_ADDR_SZ * (accounts->signature_cnt - accounts->readonly_signed_cnt);
+  fd_memcpy( write_ptr, accounts->signers_w, signers_write_sz );
+  write_ptr += signers_write_sz;
+
+  fd_memcpy( write_ptr, accounts->signers_r, FD_TXN_ACCT_ADDR_SZ * accounts->readonly_signed_cnt );
+  write_ptr += FD_TXN_ACCT_ADDR_SZ * accounts->readonly_signed_cnt;
+
+  ulong non_signers_write_sz = FD_TXN_ACCT_ADDR_SZ * (ulong)(accounts->acct_cnt - accounts->readonly_unsigned_cnt - accounts->signature_cnt);
+  fd_memcpy( write_ptr, accounts->non_signers_w, non_signers_write_sz);
+  write_ptr += non_signers_write_sz;
+
+  fd_memcpy( write_ptr, accounts->non_signers_r, FD_TXN_ACCT_ADDR_SZ * accounts->readonly_unsigned_cnt );
+  write_ptr += FD_TXN_ACCT_ADDR_SZ * accounts->readonly_unsigned_cnt;
+  FD_TEST( (ushort)((ulong)write_ptr - (ulong)out_txn_payload) == txn_meta->recent_blockhash_off );
+
+  /* Write recent blockhash */
+  if( FD_LIKELY( opt_recent_blockhash ) ) {
+    memcpy( write_ptr, opt_recent_blockhash, FD_TXN_BLOCKHASH_SZ );
+  } else {
+    memset( write_ptr, 0UL, FD_TXN_BLOCKHASH_SZ );
+  }
+  write_ptr += FD_TXN_BLOCKHASH_SZ;
+
+  return (ulong)(write_ptr - out_txn_payload);
+}
+
+ulong
+fd_txn_add_instr( uchar * txn_meta_ptr,
+                  uchar out_txn_payload[ static FD_TXN_MTU ],
+                  uchar program_id,
+                  uchar const * accounts,
+                  ulong accounts_sz,
+                  uchar const * instr_buf,
+                  ulong instr_buf_sz ) {
+
+  fd_txn_t * txn_meta = (fd_txn_t *) txn_meta_ptr;
+  FD_TEST( txn_meta->instr_cnt < FD_TXN_INSTR_MAX );
+  FD_TEST( txn_meta->recent_blockhash_off != 0 );
+
+  uchar * instr_start = out_txn_payload + txn_meta->recent_blockhash_off + FD_TXN_BLOCKHASH_SZ;
+  txn_meta->instr_cnt++;
+  uchar * write_ptr       = instr_start;
+
+  uchar compact_instr_cnt_sz = (uchar) fd_cu16_enc( (ushort)txn_meta->instr_cnt, write_ptr );
+  FD_TEST( compact_instr_cnt_sz == 1 );
+
+  write_ptr += compact_instr_cnt_sz;
+
+  /* Calculate offset of next instruction. */
+  if ( txn_meta->instr_cnt > 1 ) {
+    write_ptr = out_txn_payload + txn_meta->instr[txn_meta->instr_cnt-2].data_off + txn_meta->instr[txn_meta->instr_cnt-2].data_sz;
+  }
+
+  instr_start = write_ptr;
+
+  *write_ptr = program_id;
+  write_ptr += sizeof(uchar);
+
+  uchar compact_accts_len_sz = (uchar) fd_cu16_enc( (ushort)accounts_sz, write_ptr );
+  write_ptr += compact_accts_len_sz;
+
+  fd_memcpy( write_ptr, accounts, accounts_sz );
+  write_ptr += accounts_sz;
+
+  ushort data_sz = (ushort)instr_buf_sz;
+  uint compact_data_len_sz = fd_cu16_enc( data_sz, write_ptr );
+  write_ptr += compact_data_len_sz;
+
+  /* Copy data buffer over */
+  ushort data_off = (ushort) (write_ptr - out_txn_payload);
+  fd_memcpy( write_ptr, instr_buf, data_sz );
+  write_ptr += data_sz;
+  ushort acct_off = (ushort) (instr_start + sizeof(uchar) + compact_accts_len_sz - out_txn_payload);
+
+  (void) fd_txn_instr_meta_generate( (uchar*)&txn_meta->instr[txn_meta->instr_cnt-1], 
+                                      program_id, 
+                                      (ushort)accounts_sz,
+                                      data_sz, acct_off, data_off );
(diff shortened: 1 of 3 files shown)


src/flamenco/txn/fd_txn_generate.c
```diff
@@ -0,0 +1,136 @@
...
+  uchar compact_accts_len_sz = (uchar) fd_cu16_enc( (ushort)accounts_len, compact_accts_len );
+  fd_memcpy( write_ptr, compact_accts_len, compact_accts_len_sz );
+  write_ptr += compact_accts_len_sz;
+
+  fd_memcpy( write_ptr, accounts, accounts_len );
+  write_ptr += accounts_len;
+
+  // Build instruction data
+  uchar instr_buf[FD_TXN_MTU];
+  ushort data_sz = (*instr_fun)( instr_buf, opt_build_args, opt_args_len );
```
ptaffet-jump, line 119: I'm not sure why this callback-style API makes more sense than just taking a pointer. The callback doesn't have any extra information at this point than the caller of this method would have, right? The one potential advantage of the callback API I could see is the ability to serialize directly into the transaction buffer to save a memcpy, but you're not doing that here, so I'm confused.

src/flamenco/txn/fd_txn_generate.h
```diff
@@ -0,0 +1,87 @@
...
+*/
+
+#include "../../ballet/txn/fd_txn.h"
+#include "../../ballet/txn/fd_compact_u16.h"
+#include "../../flamenco/types/fd_types_custom.h"
+#include "../../flamenco/types/fd_types.h"
+
+#define SIGNATURE_SZ 64
+#define ACCOUNT_SZ   32
+#define BLOCKHASH_SZ 32
```
ptaffet-jump, line 14: You should probably use the pre-existing definitions for these, but if not, they definitely need to be prefixed, since they're #defines in a header.

src/flamenco/txn/fd_txn_generate.h
```diff
@@ -0,0 +1,87 @@
+/**
+ * Provides utility methods to create txn templates for
+ * pre-staging, as well as a mechanism to build out an 
+ * entire transaction with instructions.
+*/
```
ptaffet-jump, line 5: Firedancer comments start with `/*` and don't have `*`s on subsequent lines. The ending `*/` can be on the same line as the text.

src/flamenco/txn/fd_txn_generate.h
```diff
@@ -0,0 +1,87 @@
...
+typedef struct fd_txn_accounts fd_txn_accounts_t;
+
+// Message header type
+struct __attribute__((packed)) fd_txn_message_hdr {
+  uchar num_signatures;
+  uchar num_readonly_signatures;
+  uchar num_readonly_unsigned;
+};
+
+typedef struct fd_txn_message_hdr fd_txn_message_hdr_t;
```
ptaffet-jump, line 40: This isn't meant for use by the callers of this API, right? You can move the definition into the .c file

src/flamenco/txn/fd_txn_generate.h
```diff
@@ -0,0 +1,87 @@
...
+ * Method used to create a template for a txn (useful for pre-staging and re-use)
+*/
+ulong fd_txn_base_generate( uchar out_txn_meta[ static FD_TXN_MAX_SZ ],
+                            uchar out_txn_payload[ static FD_TXN_MTU ],
+                            ulong num_signatures,
+                            fd_txn_accounts_t * accounts,
+                            uchar * opt_recent_blockhash );
+
+/**
+ * Method used for adding an instruction to a txn being generated.
```
ptaffet-jump, line 77: I guess this is in flamenco and not disco, but it seems like a general-purpose function like this probably needs more comprehensive comments. What does the return value mean? Can it fail? Is `opt_build_args==NULL` okay? Only if `opt_args_len==0`? Given the type of accounts is `uchar *`, I assume these are indices to the list passed to `base_generate`, but that should be mentioned.

src/flamenco/txn/fd_txn_generate.h
```diff
@@ -0,0 +1,87 @@
...
+*/
+ulong fd_txn_add_instr( uchar * txn_meta_ptr,
+                        uchar out_txn_payload[ static FD_TXN_MTU ],
+                        uchar program_id,
+                        uchar * accounts,
```
ptaffet-jump, line 82:
```suggestion
                        uchar const * accounts,
```
Right? And below for args? Or no?

arjain4:
@ptaffet-jump thank you for the review, this should have been marked as a draft actually. Started cleaning up the code and will look into simplifying these APIs. The first version was just hacked up a few months ago to prove we could send votes to solana. The design comments will be particularly helpful as we start thinking about how we'll use these again

src/flamenco/txn/fd_txn_generate.c
arjain4: I wrote this to allow for the memcpy to be saved but this was before I knew about the compact u16 stuff :( Can't use the txn buffer directly because we need the size of the instr. Will re-think this

src/flamenco/txn/fd_txn_generate.c
```diff
@@ -0,0 +1,136 @@
...
+
+  fd_txn_t * txn_meta = (fd_txn_t *) txn_meta_ptr;
+  FD_TEST( txn_meta->instr_cnt < FD_TXN_INSTR_MAX );
+  FD_TEST( txn_meta->recent_blockhash_off != 0 );
+
+  uchar * instr_start = out_txn_payload + txn_meta->recent_blockhash_off + BLOCKHASH_SZ;
+  txn_meta->instr_cnt++;
+  uchar compact_instr_cnt[3];
+  uchar compact_instr_cnt_sz = (uchar) fd_cu16_enc( (ushort)txn_meta->instr_cnt, compact_instr_cnt );
+  FD_TEST( compact_instr_cnt_sz == 1 );
```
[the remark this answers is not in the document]
arjain4: This was originally designed to provide flexibility in terms of adding as well as removing instructions (which is to say having a variable number of instructions added to a pre-staged txn). Will discuss the current use case and need with Charlie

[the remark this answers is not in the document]
arjain4: `FD_TEST( txn_meta->instr_cnt < FD_TXN_INSTR_MAX );` this is the check which protects against the overwrite and crash. If we reach 64 it is already a crash condition

src/flamenco/txn/fd_txn_generate.c
arjain4: Switched to using an instr buf pointer

src/flamenco/txn/fd_txn_generate.c
```diff
@@ -0,0 +1,158 @@
...
+  txn_meta->acct_addr_cnt         = accounts->acct_cnt;
+  txn_meta->readonly_signed_cnt   = accounts->readonly_signed_cnt;
+  txn_meta->readonly_unsigned_cnt = accounts->readonly_unsigned_cnt;
+  /* Number of signatures is encoded as a compact u16 but
+     can always be encoded using 1 byte here. */
+  txn_meta->message_off           = (ushort)(num_signatures * FD_TXN_SIGNATURE_SZ  + 1);
+  txn_meta->signature_off         = (ushort)1UL;
+  txn_meta->instr_cnt             = 0;
+
+  FD_TEST(txn_meta->acct_addr_cnt <= 128);
```
ptaffet-jump, line 48: Hmm maybe I accidentally mislead you here. If acct_addr_cnt ==128, then it doesn't fit in 1 byte. Is just asserting (and documenting) it must be strictly < 128 okay?

src/flamenco/txn/fd_txn_generate.c
```diff
@@ -0,0 +1,158 @@
...
+
+  fd_txn_t * txn_meta = (fd_txn_t *) txn_meta_ptr;
+  FD_TEST( txn_meta->instr_cnt < FD_TXN_INSTR_MAX );
+  FD_TEST( txn_meta->recent_blockhash_off != 0 );
+
+  uchar * instr_start = out_txn_payload + txn_meta->recent_blockhash_off + FD_TXN_BLOCKHASH_SZ;
+  txn_meta->instr_cnt++;
+  uchar * write_ptr       = instr_start;
+
+  uchar compact_instr_cnt_sz = (uchar) fd_cu16_enc( (ushort)txn_meta->instr_cnt, write_ptr );
```
ptaffet-jump, line 108: Why make this a `uchar`?

src/flamenco/txn/fd_txn_generate.c
```diff
@@ -0,0 +1,158 @@
...
+  FD_TEST( compact_instr_cnt_sz == 1 );
+
+  write_ptr += compact_instr_cnt_sz;
+
+  /* Calculate offset of next instruction. */
+  if ( txn_meta->instr_cnt > 1 ) {
```
ptaffet-jump, line 114:
```suggestion
  if( FD_LIKELY( txn_meta->instr_cnt > 1 ) ) {
```

ptaffet-jump: (I think >= 2 is more clear, since you subtract 2 below, but they're obviously equivalent)

arjain4: For votes it's the other way around actually, changing to UNLIKELY

src/flamenco/txn/fd_txn_generate.c
arjain4: remnant from cleanup

src/flamenco/txn/fd_txn_generate.c
```diff
@@ -0,0 +1,158 @@
...
+    write_ptr = out_txn_payload + txn_meta->instr[txn_meta->instr_cnt-2].data_off + txn_meta->instr[txn_meta->instr_cnt-2].data_sz;
+  }
+
+  instr_start = write_ptr;
+
+  *write_ptr = program_id;
+  write_ptr += sizeof(uchar);
+
+  uchar compact_accts_len_sz = (uchar) fd_cu16_enc( (ushort)accounts_sz, write_ptr );
```
ptaffet-jump, line 123: Again, why the cast to uchar here?

src/flamenco/txn/fd_txn_generate.c
arjain4: yes, this condition will never actually even be hit with how we use this, so i'm updating the condition

src/flamenco/txn/fd_txn_generate.c
arjain4: this is some sort of remnant from pre-cleanup

src/flamenco/txn/fd_txn_generate.c
```diff
@@ -0,0 +1,158 @@
...
+
+  ushort data_sz = (ushort)instr_buf_sz;
+  uint compact_data_len_sz = fd_cu16_enc( data_sz, write_ptr );
+  write_ptr += compact_data_len_sz;
+
+  /* Copy data buffer over */
+  ushort data_off = (ushort) (write_ptr - out_txn_payload);
+  fd_memcpy( write_ptr, instr_buf, data_sz );
+  write_ptr += data_sz;
+  ushort acct_off = (ushort) (instr_start + sizeof(uchar) + compact_accts_len_sz - out_txn_payload);
```
ptaffet-jump, line 137: Why not just move this line a few lines up and calculate it the same way as `data_off` ?

src/flamenco/txn/fd_txn_generate.c
```diff
@@ -0,0 +1,158 @@
...
+}
+
+void
+fd_txn_reset_instrs( uchar * txn_meta_ptr,
+                     uchar out_txn_payload[ static FD_TXN_MTU ] ) {
+  fd_txn_t * txn_meta = (fd_txn_t *)txn_meta_ptr;
+  ulong instr_start   = txn_meta->recent_blockhash_off + FD_TXN_BLOCKHASH_SZ;
+
+  ushort num_instrs   = txn_meta->instr_cnt;
+  ulong instrs_sz     = txn_meta->instr[num_instrs-1].data_off + txn_meta->instr[num_instrs-1].data_sz - instr_start;
```
ptaffet-jump, line 153: What if num_instrs is 0?

src/flamenco/txn/fd_txn_generate.c
```diff
@@ -0,0 +1,158 @@
...
+
+  fd_memset( (out_txn_payload + instr_start), 0UL, instrs_sz );
+  fd_memset( txn_meta->instr, 0UL, num_instrs * sizeof(fd_txn_instr_t) );
```
ptaffet-jump, line 156: The only necessary part is zeroing out the payload's instruction count, which is hidden in the first memset.

src/flamenco/txn/fd_txn_generate.h
```diff
@@ -0,0 +1,55 @@
...
+                      uchar out_txn_payload[ static FD_TXN_MTU ],
+                      ulong num_signatures,
+                      fd_txn_accounts_t * accounts,
+                      uchar * opt_recent_blockhash );
+
+/* Method used for adding an instruction to a txn being generated.
+   The accounts param is a list of indices to the accounts in the txn.
+   The instruction buffer contains the data for the instruction to
+   be added. */
+ulong
```
ptaffet-jump, line 41: What does this return value mean?

ptaffet-jump (approved):

src/flamenco/txn/fd_txn_generate.c
```diff
@@ -0,0 +1,160 @@
...
+  txn_meta->instr_cnt++;
+  uchar * write_ptr       = instr_start;
+
+  uint compact_instr_cnt_sz = fd_cu16_enc( (ushort)txn_meta->instr_cnt, write_ptr );
+  FD_TEST( compact_instr_cnt_sz == 1 );
+
+  write_ptr += compact_instr_cnt_sz;
+
+  /* Calculate offset of next instruction. */
+  if ( FD_UNLIKELY( txn_meta->instr_cnt > 1 ) ) {
```
yhzhangjump, line 116: I am not sure why this is `txn_meta->instr_cnt > 1` instead of `txn_meta->instr_cnt > 0`. It seems that, even if `txn_meta->instr_cnt == 1`, we need to recalculate `write_ptr`.

src/flamenco/txn/fd_txn_generate.c
```diff
@@ -0,0 +1,160 @@
...
+  /* Copy data buffer over */
+  ushort data_off = (ushort) (write_ptr - out_txn_payload);
+  fd_memcpy( write_ptr, instr_buf, data_sz );
+  write_ptr += data_sz;
+
+  (void) fd_txn_instr_meta_generate( (uchar*)&txn_meta->instr[txn_meta->instr_cnt-1], 
+                                      program_id, 
+                                      (ushort)accounts_sz,
+                                      data_sz, acct_off, data_off );
+  return (ulong)(instr_start - out_txn_payload);
```
yhzhangjump, line 145: It seems that you are returning the old txn size instead of the new txn size. Is this deliberate? It seems more natural to return the new txn size.
Also, you never check whether we are writing more than `FD_TXN_MTU` bytes to `out_txn_payload`. It may be useful to guarantee that this function never writes overflow `out_txn_payload`, for security purposes.

src/flamenco/txn/fd_txn_generate.c
```diff
@@ -0,0 +1,160 @@
...
+}
+
+void
+fd_txn_reset_instrs( uchar * txn_meta_ptr,
```
yhzhangjump, line 149: Maybe call it `fd_txn_remove_instrs`?

src/flamenco/txn/fd_txn_generate.c
```diff
@@ -0,0 +1,160 @@
...
+  *out_txn_payload = (uchar)num_signatures;
+
+  /* Fill out txn metadata */
+  fd_txn_t * txn_meta             = (fd_txn_t *) out_txn_meta;
+  txn_meta->acct_addr_cnt         = accounts->acct_cnt;
+  txn_meta->readonly_signed_cnt   = accounts->readonly_signed_cnt;
+  txn_meta->readonly_unsigned_cnt = accounts->readonly_unsigned_cnt;
+  /* Number of signatures is encoded as a compact u16 but
+     can always be encoded using 1 byte here. */
+  txn_meta->message_off           = (ushort)(num_signatures * FD_TXN_SIGNATURE_SZ  + 1);
```
yhzhangjump, line 46: If num_signatures is higher than 12, then this offset is already going beyond `FD_TXN_MTU`. It seems useful to guarantee that this function would never write beyond `FD_TXN_MTU` bytes from `out_txn_payload`.

src/flamenco/txn/fd_txn_generate.c
arjain4: the write_ptr is already at the right offset for that case

src/flamenco/txn/fd_txn_generate.c
arjain4: this returns the size of the just added instr

src/flamenco/txn/fd_txn_generate.c
arjain4: We have it setup such that the caller can specify a larger buffer. That can be changed in a follow up PR if you'd like to take a look at that

src/flamenco/txn/fd_txn_generate.c
arjain4: remove is used for removing a single element across firedancer so i steered clear of that for this

PR #1698, opened 2024-05-01
build: add no-solana extra

lheeger-jump:
This adds the `no-solana` extra build flag which will disable solana build and stub out any external calls. Also preps the build for our no solana version of fddev (adding deps).


arjain4 (commented):
EOF new lines

lheeger-jump:
I have already spoken to multiple people about this change. Some of the upcoming changes make it easier to move to separate binaries. This is not that change and I do not have time to restructure the fdctl (rather, there is not enough review capacity to rewrite it right now).

lheeger-jump:
> EOF new lines

fixed

ripatel-fd:
> This is not that change and I do not have time to restructure the fdctl (rather, there is not enough review capacity to rewrite it right now).

@lheeger-jump Yes completely agree. Practical quick fixes first, polish later.

src/app/fdctl/Local.mk
```diff
@@ -40,7 +41,7 @@ $(call add-objs,configure/xdp,fd_fdctl)
 $(call add-objs,configure/ethtool,fd_fdctl)
 $(call add-objs,configure/workspace,fd_fdctl)
 
-$(call make-bin-rust,fdctl,main,fd_fdctl fd_disco fd_flamenco fd_quic fd_tls fd_ip fd_reedsol fd_ballet fd_waltz fd_tango fd_util solana_validator)
+$(call make-bin-rust,fdctl,main,fd_fdctl fd_choreo fd_disco fd_flamenco fd_funk fd_quic fd_tls fd_ip fd_reedsol fd_ballet fd_waltz fd_tango fd_util solana_validator, $(SECP256K1_LIBS))
```
mmcgee-jump, line 44: Guard these changes with `FD_HAS_NO_SOLANA` same as fddev?

mmcgee-jump: (Then move the FD_HAS_SECP256K1 check to the same place

mmcgee-jump:
Also if you're adding this can you get rid of the "topos" change thing in the config toml, and just compile in a different topology depending on this flag?

lheeger-jump:
> Also if you're adding this can you get rid of the "topos" change thing in the config toml, and just compile in a different topology depending on this flag?

Logan and I are already working on a separate PR to get this done.

src/app/fdctl/Local.mk
lheeger-jump: done

arjain4 (approved):

mmcgee-jump (approved):

PR #1701, opened 2024-05-01
VM fixes

ripatel-fd:
[vm: add instruction test suite](https://github.com/firedancer-io/firedancer/commit/6ff19c6ba530ec31ec4e2dbcf5b27e37d622aa7b) 

Resolve FIXME to move fd_vm_setup_state_for_execution to init.

[vm: bug fixes](https://github.com/firedancer-io/firedancer/commit/3b52ea43d7445a781e6bbf843481e9e2fbaf214a) 

- Fix some ALU32 sign extension bugs
- Fix div by zero being silent, introduce SIGFPE
- Fix oob shift handling

### DIFF
--- a/src/flamenco/vm/fd_vm.c
+++ b/src/flamenco/vm/fd_vm.c
@@ -27,6 +27,7 @@ fd_vm_strerror( int err ) {
   case FD_VM_ERR_SIGBUS:    return "SIGBUS misaligned memory address";
   case FD_VM_ERR_SIGRDONLY: return "SIGRDONLY illegal write";
   case FD_VM_ERR_SIGCOST:   return "SIGCOST compute unit limit exceeded";
+  case FD_VM_ERR_SIGFPE:    return "SIGFPE division by zero";
 
   /* VM syscall error codes */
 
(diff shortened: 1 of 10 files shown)


src/flamenco/vm/fd_vm.c
```diff
@@ -393,12 +393,15 @@ fd_vm_init(
   vm->trace = trace;
   vm->sha = sha;
 
-  // TODO: call fd_vm_setup_state_for_execution here once it is trace-aware
+  /* Unpack the configuration */
+  int err = fd_vm_setup_state_for_execution( vm );
+  if( FD_UNLIKELY( err != FD_VM_SUCCESS ) ) {
+    return NULL;
+  }
```
topointon-jump, line 417: We need to also switch to use `fd_vm_init` everywhere, otherwise this will break the ledger tests. `fd_vm_setup_state_for_execution` is currently not being called by the bpf loader.

topointon-jump (changes requested):
Thanks for helping with this! 🙌 

At the moment I've commented out the ledger tests which are not passing on the `vm-belt-sanding-public` branch, so CI should be green for `vm-belt-sanding-public`. Could we only merge PRs into `vm-belt-sanding-public` which are green? Thanks so much.

PR #1703, opened 2024-05-01
topo: add concrete sized, aligned, topology objs

lheeger-jump:
Adds an explicit function for adding sized and aligned object to topology. This allows us to add objects like blockstore and funk to a topology without special case logic for every single object in a topology.

### DIFF
--- a/src/app/fdctl/config.c
+++ b/src/app/fdctl/config.c
@@ -535,6 +535,11 @@ fd_topo_tile_to_config( fd_topo_tile_t const * tile ) {
 ulong
 fdctl_obj_align( fd_topo_t const *     topo,
                  fd_topo_obj_t const * obj ) {
+  #define VAL(name) (__extension__({                                                               \
+    ulong __x = fd_pod_queryf_ulong( topo->props, ULONG_MAX, "obj.%lu.%s", obj->id, name );      \
+    if( FD_UNLIKELY( __x==ULONG_MAX ) ) FD_LOG_ERR(( "obj.%lu.%s was not set", obj->id, name )); \
+    __x; }))
+
   if( FD_UNLIKELY( !strcmp( obj->name, "tile" ) ) ) {
     fd_topo_tile_t const * tile = NULL;
     for( ulong i=0UL; i<topo->tile_cnt; i++ ) {
@@ -559,9 +564,10 @@ fdctl_obj_align( fd_topo_t const *     topo,
   } else if( FD_UNLIKELY( !strcmp( obj->name, "metrics" ) ) ) {
     return FD_METRICS_ALIGN;
   } else {
-    FD_LOG_ERR(( "unknown object `%s`", obj->name ));
-    return 0UL;
+    ulong align = VAL("align");
+    return align;
   }
+#undef VAL
 }
 
 ulong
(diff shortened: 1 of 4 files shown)


src/app/fdctl/configure/workspace.c
```diff
@@ -48,7 +48,14 @@ fdctl_obj_new( fd_topo_t const *     topo,
   } else if( FD_UNLIKELY( !strcmp( obj->name, "metrics" ) ) ) {
     fd_metrics_new( laddr, VAL("in_cnt"), VAL("out_cnt") );
   } else {
-    FD_LOG_ERR(( "unknown object `%s`", obj->name ));
+    /* if the object specified the properities necessary to be concrete */
+    ulong align = fd_pod_queryf_ulong( topo->props, ULONG_MAX, "obj.%lu.align", obj->id );
```
mmcgee-jump, line 52: ??? This does not new the object. Who is going to call new?

src/app/fdctl/config.c
```diff
@@ -559,9 +564,10 @@ fdctl_obj_align( fd_topo_t const *     topo,
   } else if( FD_UNLIKELY( !strcmp( obj->name, "metrics" ) ) ) {
     return FD_METRICS_ALIGN;
   } else {
-    FD_LOG_ERR(( "unknown object `%s`", obj->name ));
-    return 0UL;
+    ulong align = VAL("align");
```
mmcgee-jump, line 567: This should happen before the other checks... a/k/a at the top of this function.  If the object already specifies a concrete align, then we don't need to calculate it.

src/app/fdctl/config.c
```diff
@@ -596,8 +602,8 @@ fdctl_obj_footprint( fd_topo_t const *     topo,
   } else if( FD_UNLIKELY( !strcmp( obj->name, "metrics" ) ) ) {
     return FD_METRICS_FOOTPRINT( VAL("in_cnt"), VAL("out_cnt") );
   } else {
-    FD_LOG_ERR(( "unknown object `%s`", obj->name ));
-    return 0UL;
+    ulong sz = VAL("sz");
```
mmcgee-jump, line 605: Same as above

src/app/fdctl/config.c
```diff
@@ -615,6 +621,11 @@ fdctl_obj_loose( fd_topo_t const *     topo,
     }
     fd_topo_run_tile_t * config = fd_topo_tile_to_config( tile );
     if( FD_LIKELY( config->loose_footprint ) ) return config->loose_footprint( tile );
+  } else {
+    ulong loose_sz = fd_pod_queryf_ulong( topo->props, ULONG_MAX, "obj.%lu.%s", obj->id, "loose" );
```
mmcgee-jump, line 625: As aobve

mmcgee-jump (changes requested):
.

src/app/fdctl/configure/workspace.c
lheeger-jump: The user.

lheeger-jump: it explicitly doesnt happen here. It is your responsibility to set the object up where you actually intend to use it.

src/app/fdctl/config.c
lheeger-jump: done

src/app/fdctl/config.c
lheeger-jump: done

src/app/fdctl/config.c
lheeger-jump: done

src/app/fdctl/configure/workspace.c
lheeger-jump: Added a better comment there.

mmcgee-jump (approved):

PR #1705, opened 2024-05-02
topo: add guard pages around workspaces

### LINKED ISSUE #989: Add guard pages around workspaces
See #968 . The workspaces are mapped to a random address, but it's possible (about 1/400,000) that gigantic pages would end up adjacent, and so an overrun could step into another workspace. This can be prevented by adding guard pages around each workspace.

alpeng-jump:
Fixes #989. Prevent overruns from stepping into another workspace by adding guard pages around each workspace.

### DIFF
--- a/src/disco/topo/fd_topo.c
+++ b/src/disco/topo/fd_topo.c
@@ -1,3 +1,4 @@
+#define _GNU_SOURCE
 #include "fd_topo.h"
 
 #include "../fd_disco_base.h"
@@ -10,6 +11,7 @@
 #include <errno.h>
 #include <unistd.h>
 #include <sys/stat.h>
+#include <sys/mman.h>
 
 void
 fd_topo_join_workspace( fd_topo_t *      topo,
@@ -74,6 +76,29 @@ fd_topo_leave_workspaces( fd_topo_t * topo ) {
 
 extern char fd_shmem_private_base[ FD_SHMEM_PRIVATE_BASE_MAX ];
 
+static int
+fd_topo_grab_region( ulong addr,
+                     ulong size ) {
+  void *mmap_ret;
+  int err;
+
+  mmap_ret = mmap( (void*)addr, size, PROT_READ, MAP_ANON|MAP_PRIVATE, -1, 0 );
+
+  if( mmap_ret == MAP_FAILED )
+    return 0;
+
+  /* Only call munmap on failure case. On success we want to keep the mapping */
+  if( (ulong)mmap_ret != addr ) {
+    err = munmap( mmap_ret, size );
+    if ( err == -1 ) {
+      FD_LOG_ERR(( "failed to unmap temporary mapping, munmap() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+    }
+    return 0;
+  }
+
+  return 1;
+}
+
 int
 fd_topo_create_workspace( fd_topo_t *      topo,
                           fd_topo_wksp_t * wksp,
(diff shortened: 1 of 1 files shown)


asiegel-jt (dismissed):

mmcgee-jump:
I don't think this is the direction to go with this, it adds 2GiB of mlocked memory overhead to every wksp which is over like 40GiB. These pages need to stay unmapped. You can make them without paging in by doing mmap(MAP_FIXED) around the prior allocation, not making it bigger.

asiegel-jt:
It seems like workspaces using gigantic pages will naturally be adjacent in vm space. I suspect allocating guard pages will often fail.

mmcgee-jump:
I don't think that's true? Workspace addresses are random, due to another security change.

mmcgee-jump:
Yeah, this happens in sequence as well, so the other workspaces won't be allocated yet anyway. They will randomize their address until it succeeds.

src/util/shmem/fd_shmem.h
```diff
@@ -269,6 +269,19 @@ int
...
+   address with the passed size. If the return value of `mmap` equals
+   the passed address this means the area of memory was unmapped
+   previously and we havesuccesfully "grabbed" the region. We can then
+   call `mmap` with MAP_FIXED over the region and be certain no
+   corruption occurs. If the return value of `mmap` does not return the
+   passed address this means that the passed region is already at least
+   partially mapped and we cannot grab it. */
+
+int
+fd_shmem_grab_region( ulong addr,
```
mmcgee-jump, line 282: This isn't really an interface on fd_shmem, I would not expose this, and leave the code in util unchanged.  You can just copy paste the grabber code into fd_topo it's a bit cleaner for now.

mmcgee-jump (changes requested):
.

mmcgee-jump:
Unless I'm doing something wrong this change doesn't appear to work. I checked it out and ran it and there are no guard pages around the workspaces. You can see this by running, eg the below,

So it doesn't work but it also silently fails, which is not what we want for security sensitive changes. I would tidy this up and make the error handling clearer and more explicit.

```
[mmcgee@emch-ossdev-firedancer8 firedancer]$ sudo cat /proc/234938/maps
00400000-05573000 r-xp 00000000 fd:01 135427458                          /home/mmcgee/firedancer/build/native/gcc/bin/fddev
05773000-0594b000 r--p 05173000 fd:01 135427458                          /home/mmcgee/firedancer/build/native/gcc/bin/fddev
0594b000-05954000 rw-p 0534b000 fd:01 135427458                          /home/mmcgee/firedancer/build/native/gcc/bin/fddev
05954000-0861c000 rw-p 00000000 00:00 0 
182edfc00000-182edfe00000 rw-s 00000000 00:2f 1993885                    /mnt/.fd/.huge/fd1_metric_in.wksp
23fd1bc00000-23fd1be00000 rw-s 00000000 00:2f 1993892                    /mnt/.fd/.huge/fd1_verify.wksp
3c4740000000-3c4780000000 r--s 00000000 00:30 2011142                    /mnt/.fd/.gigantic/fd1_quic_verify.wksp
49b3c0dff000-49b3c0e00000 ---p 00000000 00:00 0 
49b3c0e00000-49b3c1600000 rw-s 00200000 00:2f 2074106                    /mnt/.fd/.huge/fd1_stack_verify1
49b3c1600000-49b3c1601000 ---p 00000000 00:00 0 
6acac0000000-6acb00000000 rw-s 00000000 00:30 2011143                    /mnt/.fd/.gigantic/fd1_verify_dedup.wksp
7f2298e00000-7f2299600000 rw-p 00000000 00:00 0 
7f22997b3000-7f22997ca000 r-xp 00000000 fd:01 138665019                  /usr/lib64/libgcc_s-8-20210514.so.1
```

mmcgee-jump (changes requested):
.

mmcgee-jump:
I'm also realizing, on the rare occasion that workspaces end up adjacent this code would just fail anyway, rather than making guard pages, since it wouldn't be able to grab the region.

You probably will need a `with_guard` flag to fd_wksp_create that makes sure the grabbed region has enough padding on either side to avoid this.

ripatel-fd:
Linux 6.13 has a new feature for this.

```
       MADV_GUARD_INSTALL (since Linux 6.13)
              Install a lightweight guard region into the range specified
              by addr and size, causing any read or write in the range to
              result in a SIGSEGV signal being raised.
```

ripatel-fd:
Closing stale PR

PR #1710, opened 2024-05-02
features: update timely_vote_credits

0x0ece:
Context: https://github.com/anza-xyz/agave/pull/127

We need to update our test ledgers:
- v18-small (fails at slot 2)
- helloworld (fails at slot 101)


jumpsiegel (approved):

ripatel-fd:
I guess we'll have to regenerate the test ledger. Very annoying, thanks Anza.

topointon-jump:
I will make v18-small-2 with the new public keys for the feature, and we can remove helloworld as it's a subset of the other ledger tests.

0x0ece:
Merged into the other PR, to keep everything in one place.

PR #1714, opened 2024-05-02
Add run_cluster script

topointon-jump:
Porting this over to the public repo.

---

Simple python script to make running Solana multi-node and solana-test-validator clusters from a local Solana checkout more convenient. Useful for creating test ledgers, and will be later used to generate more elaborate tests. To kill the cluster, simply kill the script.

This is written in an async style to enable us to use it in more complicated async scripts later.


topointon-jump:
Mind re-approving @ripatel-fd? 🙏

topointon-jump:
@ripatel-fd  They get squashed as part of the merge queue IIUC, let's see

edit: sorry didn't realise this had been changed back!

PR #1728, opened 2024-05-03
flamenco: fix gossip ping token verification

lheeger-jump:
Issue was the formulation of the ping token was not mapped to the verfication correctly.


arjain4 (approved):
Wow whatta fix

PR #1729, opened 2024-05-03
util: static linkage for fd_net_create_packet_header_template

lheeger-jump:
Without this the linker complains about the symbol when it actually gets used.


mmcgee-jump (approved):

mmcgee-jump:
Clean the commits? Not sure why that merge is in there.

mmcgee-jump (approved):

PR #1733, opened 2024-05-03
Fix underflow error in binary search

mjain-jump:
Case if `start = 0, end = 1`, and `slot` is lower than all values in the deque. When mid = 0, `end = mid - 1` which causes an underflow and thus a segfault / infinite loop.

Similarly, if `slot` is greater than all deque values, we may try to access a value at `vote_state->votes[len]` which would cause an out of bounds memory access.


topointon-jump (approved):
As discussed, the fuzz failure which caused this will be put into a corpus to backtest against. When @mjain-jump's pipeline is ready, this failure will be converted into a unit test and added to CI.

lidatong (approved):

lidatong:
crashes during the check itself... good catch thanks

PR #1736, opened 2024-05-03
flamenco: unifying ledger test code path + on demand genesis ingest

ibhatt-jumptrading:


jumpsiegel (approved):

jumpsiegel:
Can we generate a solcap file for slot 0?  We should try this (and fix it in a different PR if we cannot).  Right now I believe genesis is broken for agave 2.0 and it is due to the fact that some of the bogus accounts are feature flag controlled and we just don't support them (zk token program for example).  We should ask agave to remove them and/or we should look to see if the feature is enabled and throw a clear error on why the banks hash does not match...

ibhatt-jumptrading:
> Can we generate a solcap file for slot 0? We should try this (and fix it in a different PR if we cannot). Right now I believe genesis is broken for agave 2.0 and it is due to the fact that some of the bogus accounts are feature flag controlled and we just don't support them (zk token program for example). We should ask agave to remove them and/or we should look to see if the feature is enabled and throw a clear error on why the banks hash does not match...

Yes, I will look at this later today

asiegel-jt (approved):

PR #1738, opened 2024-05-03
flamenco: handling vm hash zero length case

ibhatt-jumptrading:
1. Shouldn't be erroring out on vm_translate on zero length cases
2. Adding FIXME in function to handle within fd_vm_translate_vm_to_host_private


lheeger-jump (approved):

topointon-jump:
I remember fixing several issues related to this in the vm-belt-sanding branch. 

The memory translation functions in vm-belt-sanding don't error out when the size is zero, instead they are treated as "undefined behaviour". That is, they don't error, but the return value can't be relied upon. This is acceptable, as there should be no accesses for regions with sizes of 0. 

However it does mean the caller needs to explicitly handle the case where the size is 0, as this PR has done. See https://github.com/firedancer-io/firedancer/blob/3a226fa1ebe73055cd8e0f3123a9f7333c6372a2/src/flamenco/vm/fd_vm_private.h#L162-L164

topointon-jump:
cc @0x0ece for awareness, as this change was in the crypto vm syscalls

PR #1742, opened 2024-05-06
Add two fields to fd_topo_run_tile_t:

llamb-jump:
`int for_tpool;`
      when set this a call to `run_tile_thread()` will not spawn a thread so the tpool can handle the thread spawning later
  `int   (*main                    )( void );`
      When not NULL this function is called instead of `fd_mux_tile()`.

### DIFF
--- a/src/disco/topo/fd_topo_run.c
+++ b/src/disco/topo/fd_topo_run.c
@@ -161,22 +161,27 @@ fd_topo_run_tile( fd_topo_t *          topo,
   if( FD_UNLIKELY( tile_run->lazy ) ) lazy = tile_run->lazy( tile_mem );
 
   fd_rng_t rng[1];
-  fd_mux_tile( tile->cnc,
-               tile_run->mux_flags,
-               polled_in_cnt,
-               in_mcache,
-               in_fseq,
-               tile->out_link_id_primary == ULONG_MAX ? NULL : topo->links[ tile->out_link_id_primary ].mcache,
-               out_cnt_reliable,
-               out_fseq,
-               tile_run->burst,
-               0,
-               lazy,
-               fd_rng_join( fd_rng_new( rng, 0, 0UL ) ),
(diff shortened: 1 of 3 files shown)


asiegel-jt (approved):

src/disco/topo/fd_topo_run.c
```diff
@@ -161,22 +161,27 @@ fd_topo_run_tile( fd_topo_t *          topo,
...
-               0,
-               lazy,
-               fd_rng_join( fd_rng_new( rng, 0, 0UL ) ),
-               fd_alloca( FD_MUX_TILE_SCRATCH_ALIGN, FD_MUX_TILE_SCRATCH_FOOTPRINT( polled_in_cnt, out_cnt_reliable ) ),
-               ctx,
-               &callbacks );
-  FD_LOG_ERR(( "tile run loop returned" ));
+  int ret = 0;
+  if( FD_LIKELY( tile_run->main == NULL ) ) {
+    ret = fd_mux_tile( tile->cnc,
```
mmcgee-jump, line 166: This is not the right way to do this, and has a few problems. You can see them above: the threads don't get their own metrics, or sandbox, the IDs might be assigned incorrectly and cause deadlocks.

You should spawn actual tiles for the worker threads, but just run them in the same process. You can do this with a new flag on the topo, and using the existing system that run_solana does.

mmcgee-jump (changes requested):
.

src/disco/topo/fd_topo_run.c
mmcgee-jump: fd_topo_run launcher is only used for debugging, as well, so this is broken for the production fdctl boot, which is in `run.c`

asiegel-jt: You can't actually spawn the tile thread this early. You need to wait until tpool is created.

PR #1744, opened 2024-05-06
add fd_tile_private_map_boot() to fd_tile.h for later tpool use in fddev

llamb-jump:
`fd_tile_private_map_boot` will allow a tile to boot a tpool on the CPUs specified by `tile_to_cpu`. These CPUs have tiles running on them with `for_tpool = 1` from https://github.com/firedancer-io/firedancer/pull/1742.


asiegel-jt (approved):

mmcgee-jump (changes requested):
This change is not needed, see comment in the other PR.

PR #1745, opened 2024-05-06
flamenco: fixing nonce account check

ibhatt-jumptrading:
1. Reimplementing has_nonce check
2. Uncommenting ledger test

### DIFF
--- a/src/flamenco/runtime/program/fd_system_program_nonce.c
+++ b/src/flamenco/runtime/program/fd_system_program_nonce.c
@@ -961,51 +961,97 @@ fd_load_nonce_account( fd_exec_txn_ctx_t const *   txn_ctx,
   return 1;
 }
 
-
+/* https://github.com/firedancer-io/solana/blob/4b31032e68f85848b02fcc4c9e580d57f32ec04b/runtime/src/bank.rs#L4755-L4756 */
 int
-fd_has_nonce_account( fd_exec_txn_ctx_t const *   txn_ctx,
-                       int *                      perr ) {
+fd_has_nonce_account( fd_exec_txn_ctx_t const * txn_ctx, int * perr ) {
+  ushort recent_blockhash_off = txn_ctx->txn_descriptor->recent_blockhash_off;
+  fd_hash_t * recent_blockhash = (fd_hash_t *)((uchar *)txn_ctx->_txn_raw->raw + recent_blockhash_off);
 
-  *perr = 0;
+  fd_block_block_hash_entry_t * hashes_deque   = txn_ctx->slot_ctx->slot_bank.recent_block_hashes.hashes;
+  fd_block_block_hash_entry_t * last_blockhash = deq_fd_block_block_hash_entry_t_peek_tail( hashes_deque );
+  fd_hash_t * last_blockhash_hash = &last_blockhash->blockhash;
+  
 
-  fd_txn_t const *      txn_descriptor = txn_ctx->txn_descriptor;
-  fd_rawtxn_b_t const * txn_raw        = txn_ctx->_txn_raw;
-
-  if( txn_descriptor->instr_cnt == 0 ) {
-    *perr = FD_EXECUTOR_INSTR_ERR_INVALID_ARG;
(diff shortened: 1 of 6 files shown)


src/flamenco/runtime/program/fd_system_program_nonce.c
```diff
@@ -919,6 +919,7 @@ fd_load_nonce_account( fd_exec_txn_ctx_t const *   txn_ctx,
   uchar const * instr_data = fd_txn_get_instr_data( txn_instr, txn_raw->raw );
   uchar const * instr_acct_idxs = fd_txn_get_instr_accts( txn_instr, txn_raw->raw );
 
+  /* TODO: Likely this check must be updated to support other nonce related instructions */
```
lheeger-jump, line 922: what does this mean

jumpsiegel: I assume he is referring to the change he did below where "has_nonce_account" is also checking to see if the instruction contained is one of the nonce instructions..

jumpsiegel: fd_load_nonce_account is only used from fd_runtime_txn_lamports_per_signature.  This function looks to be the equiv of bank.rs:get_fee_for_message.  This calls check_message_for_nonce (bank.rs:3610) which calls get_durable_nonce in message/sanitized.rs which looks like it only works against a AdvanceNonceAccount instruction which matches the code.  So, I "believe" THIS code is correct and you should remove the todo

Now,  bank.rs:get_fee_for_message looks like it gives preference to the lamports_per_signature in the blockhash queue over the lamprots_per_signature contained in the nonce account.  It looks like fd_runtime_txn_lamports_per_signature does exactly the reverse...  This says we have a bug in fd_runtime_txn_lamports_per_signature

src/flamenco/runtime/program/fd_system_program_nonce.c
```diff
@@ -989,23 +989,18 @@ fd_has_nonce_account( fd_exec_txn_ctx_t const *   txn_ctx,
...
-  uchar const * instr_acct_idxs = fd_txn_get_instr_accts( txn_instr, txn_raw->raw );
-
-  if( FD_UNLIKELY(
-      txn_instr->data_sz != 4UL ||
-      FD_LOAD( uint, instr_data ) != (uint)fd_system_program_instruction_enum_advance_nonce_account ) ) {
-    *perr = FD_EXECUTOR_INSTR_ERR_INVALID_ACC_DATA;
-    return 0;
+  uint instr_data_val;
+  fd_memcpy( &instr_data_val, instr_data, sizeof(uint) );
+
```
jumpsiegel, line 994: Could you embed the URL (as a comment) here to where the equiv solana code is?

jumpsiegel (approved):

src/flamenco/runtime/program/fd_system_program_nonce.c
```diff
@@ -961,51 +961,97 @@ fd_load_nonce_account( fd_exec_txn_ctx_t const *   txn_ctx,
...
+/* https://github.com/firedancer-io/solana/blob/4b31032e68f85848b02fcc4c9e580d57f32ec04b/runtime/src/bank.rs#L4755-L4756 */
 int
-fd_has_nonce_account( fd_exec_txn_ctx_t const *   txn_ctx,
-                       int *                      perr ) {
+fd_has_nonce_account( fd_exec_txn_ctx_t const * txn_ctx, int * perr ) {
+  ushort recent_blockhash_off = txn_ctx->txn_descriptor->recent_blockhash_off;
+  fd_hash_t * recent_blockhash = (fd_hash_t *)((uchar *)txn_ctx->_txn_raw->raw + recent_blockhash_off);
 
-  *perr = 0;
+  fd_block_block_hash_entry_t * hashes_deque   = txn_ctx->slot_ctx->slot_bank.recent_block_hashes.hashes;
```
arjain4, line 970: I think this comes from the block hash queue: https://github.com/firedancer-io/solana/blob/4b31032e68f85848b02fcc4c9e580d57f32ec04b/runtime/src/bank.rs#L4654

Use slot_ctx->slot_bank.block_hash_queue

PR #1748, opened 2024-05-06
runtime tests: fix err==-1 misinterpreted as success

0x0ece:


lheeger-jump (approved):

0x0ece:
@ripatel-fd no, it's the opposite. generic error is -1, so before it was 0, now it's 1.

I have a corresponding change in https://github.com/firedancer-io/solfuzz-agave/pull/14

PR #1750, opened 2024-05-07
build: add perf extra

lheeger-jump:


config/extra/with-perf.mk
```diff
@@ -0,0 +1,2 @@
+CPPFLAGS+=-fno-omit-frame-pointer -fno-inline
```
mmcgee-jump, line 1: fno-omit-frame-pointer is defaulted on for perf on all builds, I believe? I made that change.

lheeger-jump: i see it for some other things but not for perf in specific. This is just for explicitly perf profiling.

arjain4 (approved):

PR #1760, opened 2024-05-07
ci: firedancer integration test

lheeger-jump:


.github/workflows/test_firedancer.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          count: 128
+
+      - name: build
+        run: |
+          ./contrib/make-j fddev all
+      - name: test firedancer - local cluster
+        run: |
+          sudo prlimit --pid=$$ --nofile=1048576
+          sudo prlimit --pid=$$ --memlock=unlimited
```
mmcgee-jump, line 28: You don't need these, fddev handles nofile and memlock.

lheeger-jump, line 28:
```suggestion
```

lheeger-jump: done

arjain4 (approved):

.github/workflows/test_firedancer.yml
lheeger-jump: nvm, undid. these are set up for a local cluster we set up for the script. fddev doesnt do this until much later.

arjain4 (approved):

PR #1762, opened 2024-05-07
fdctl: tiles specify their own run loop

mmcgee-jump:


src/app/fdctl/run/tiles/fd_poh.c
```diff
@@ -1725,25 +1720,55 @@ unprivileged_init( fd_topo_t *      topo,
 static long
 lazy( fd_topo_tile_t * tile ) {
   (void)tile;
-  /* See explanation in fd_pack */
+  
   return 128L * 300L;
 }
```
ptaffet-jump, line 1725: This is unnecessary now, right?

ptaffet-jump (approved):
Seems fine.
As I went through it, I realized that we should probably think a bit more about all the lazy values instead of just using 128*300, but that's neither here nor there.

ripatel-fd:
This is unmergeable / severely out of date.

PR #1784, opened 2024-05-08
Fix fuzzer / conformance memory issue

mjain-jump:
Memset `smem` to 0 before the beginning of every sol_compat execution

### DIFF
--- a/src/flamenco/runtime/tests/fd_exec_sol_compat.c
+++ b/src/flamenco/runtime/tests/fd_exec_sol_compat.c
@@ -89,6 +89,18 @@ sol_compat_instr_execute_v1( uchar *       out,
                              ulong         in_sz ) {
 
   ulong fmem[ 64 ];
+  /*
+  FIXME: Previously, there were issues with getting different execution results from this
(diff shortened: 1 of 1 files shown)


src/flamenco/runtime/tests/fd_exec_sol_compat.c
```diff
@@ -89,6 +89,7 @@ sol_compat_instr_execute_v1( uchar *       out,
                              ulong         in_sz ) {
 
   ulong fmem[ 64 ];
+  memset( smem, 0, smax );
```
topointon-jump, line 92: This is a hack, if this solves your problem it means there is likely some bad memory access further on. However, for now I would say it's fine to keep it in `sol_compat_instr_execute_v1`, with a FIXME comment to address it later.

topointon-jump (dismissed):

src/flamenco/runtime/tests/fd_exec_sol_compat.c
topointon-jump: Does this also significantly slow down test execution?

topointon-jump (approved):

Issue #1770, opened 2024-05-08
Reimplement and use faster status cache


mmcgee-jump:

mmcgee-jump:
Not planned for 1.5

PR #1790, opened 2024-05-08
bank: rewrite status cache

mmcgee-jump:
The status cache has two main issues,

 (1) It's not particularly concurrent, and causes slowdowns when running
     with highly parallel bank execution.

 (2) It has unbounded memory consumption, and can cause out of memory
     conditions when it needs to store large numbers of transactions.

It is rewritten in Firedancer for performance.  The general design is there are two operations which need to be highly concurrent, insert and query, and everything else is rare and happens between slots and can largely just lock the whole thing.

The base case for insert is optimized to two always-uncontended, and one very lightly contended compare and swap.  Query is fully lockless.

The transaction result storage is combined between the snapshot service cache, and the query lookup cache, which more than halves the memory usage and the memory use is fixed up front.

Related to #1770

### DIFF
--- a/src/disco/bank/fd_txncache.c
+++ b/src/disco/bank/fd_txncache.c
@@ -1,0 +1,744 @@
+#include "fd_txncache.h"
+
+#define SORT_NAME        sort_slot_ascend
+#define SORT_KEY_T       ulong
+#define SORT_BEFORE(a,b) (a)<(b)
+#include "../../util/tmpl/fd_sort.c"
+
+/* The number of transactions in each page.  This needs to be high
+   enough to amoritze the cost of caller code reserving pages from,
+   and returning pages to the pool, but not so high that the memory
+   wasted from blockhashes with only one transaction is significant. */
+
+#define FD_TXNCACHE_TXNS_PER_PAGE (16384UL)
+
+/* The number of unique entries in the hash lookup table for each
+   blockhash.  A higher value here uses more memory but enables faster
+   lookups. */
+
+#define FD_TXNCACHE_BLOCKCACHE_MAP_CNT (524288UL)
+
+/* The number of unique entries in the hash lookup table for each
+   (slot, blockhash).  This prevents all the entries needing to be in
+   one slotcache list, so insertions can happen concurrently. */
+
+#define FD_TXNCACHE_SLOTCACHE_MAP_CNT (1024UL)
+
+struct fd_txncache_private_txn {
+  uint  blockcache_next; /* Pointer to the next element in the blockcache hash chain containing this entry from the pool. */
+  uint  slotblockcache_next;  /* Pointer to the next element in the slotcache hash chain containing this entry from the pool. */
+
+  ulong slot;            /* Slot that the transaction was executed.  A transaction might be in the cache
+                            multiple times if it was executed in a different slot on different forks.  The
+                            same slot will not appear multiple times however. */
+  uchar txnhash[ 20 ];   /* The transaction hash, truncated to 20 bytes.  The hash is not always the first 20
+                            bytes, but is 20 bytes starting at some arbitrary offset given by the key_offset value
+                            of the containing by_blockhash entry. */
+  uchar result[ 40 ];    /* The result of executing the transaction. */
+};
+
+typedef struct fd_txncache_private_txn fd_txncache_private_txn_t;
+
+struct fd_txncache_private_txnpage {
+  ushort                    free; /* The number of free txn entries in this page. */
+  fd_txncache_private_txn_t txns[ FD_TXNCACHE_TXNS_PER_PAGE][ 1 ]; /* The transactions in the page. */
+};
+
+typedef struct fd_txncache_private_txnpage fd_txncache_private_txnpage_t;
+
+struct fd_txncache_private_blockcache {
+  uchar blockhash[ 32 ]; /* The actual blockhash of these transactions. */
+  ulong lowest_slot;     /* The lowest slot we have seen that contains a transaction referencing this blockhash.
+                            The blockhash entry will not be purged until the lowest rooted slot is at least 150
+                            slots higher than this. */
+  ulong txnhash_offset;  /* To save memory, the Agave validator decided to truncate the hash of transactions stored in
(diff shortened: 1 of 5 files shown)


src/disco/bank/fd_txncache.c
```diff
@@ -0,0 +1,741 @@
...
+}
+
+FD_FN_CONST ulong
+fd_txncache_footprint( ulong max_rooted_slots,
+                       ulong max_live_slots,
+                       ulong max_txn_per_slot ) {
+  if( FD_UNLIKELY( max_rooted_slots<1UL || max_live_slots<1UL ) ) return 0UL;
+  if( FD_UNLIKELY( max_live_slots<max_rooted_slots ) ) return 0UL;
+  if( FD_UNLIKELY( max_txn_per_slot<1UL ) ) return 0UL;
+  if( FD_UNLIKELY( !fd_ulong_is_pow2( max_live_slots || !fd_ulong_is_pow2( max_txn_per_slot ) ) ) ) return 0UL;
```
ptaffet-jump, line 198:
```suggestion
  if( FD_UNLIKELY( !fd_ulong_is_pow2( max_live_slots ) || !fd_ulong_is_pow2( max_txn_per_slot ) ) ) return 0UL;
```

src/disco/bank/fd_txncache.c
```diff
@@ -0,0 +1,741 @@
...
+    fd_txncache_query_t const * query = &queries[ i ];
+    fd_txncache_private_blockcache_t * blockcache;
+    int result = fd_txncache_find_blockhash( tc, query->blockhash, &blockcache );
+    if( FD_UNLIKELY( result!=FD_TXNCACHE_FIND_FOUND ) ) {
+      continue;
+    }
+
+    ulong txnhash_offset = blockcache->txnhash_offset;
+    ulong head_hash = FD_LOAD( ulong, query->txnhash+txnhash_offset ) % FD_TXNCACHE_BLOCKCACHE_MAP_CNT;
+    for( uint head=blockcache->heads[ head_hash ]; head!=UINT_MAX; head=tc->txnpages[ head/FD_TXNCACHE_TXNS_PER_PAGE ].txns[ head%FD_TXNCACHE_TXNS_PER_PAGE ]->blockcache_next ) {
```
ptaffet-jump, line 697: You could consider a macro for `tc->txnpages[ head/FD_TXNCACHE_TXNS_PER_PAGE ].txns[ head%FD_TXNCACHE_TXNS_PER_PAGE ]` since it shows up several times in places it would be nice if it weren't so long

src/disco/bank/fd_txncache.c
```diff
@@ -0,0 +1,741 @@
...
+                            same slot will not appear multiple times however. */
+  uchar txnhash[ 20 ];   /* The transaction hash, truncated to 20 bytes.  The hash is not always the first 20
+                            bytes, but is 20 bytes starting at some arbitrary offset given by the key_offset value
+                            of the containing by_blockhash entry. */
+  uchar result[ 40 ];    /* The result of executing the transaction. */
+};
+
+typedef struct fd_txncache_private_txn fd_txncache_private_txn_t;
+
+struct fd_txncache_private_txnpage {
```
ptaffet-jump, line 42: You should mention explicitly here that all the txns refer to the same blockhash

src/disco/bank/fd_txncache.c
```diff
@@ -0,0 +1,741 @@
...
+  ushort                    free; /* The number of free txn entries in this page. */
```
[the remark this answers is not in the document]
ptaffet-jump: I haven't finished understanding it, so I'm not sure this is possible, but what about making a linked list of the pages for a blockhash using these bytes. Then  you could get rid of the `pages` pointer and freeing a list of pages would be O(1).

src/disco/bank/fd_txncache.c
```diff
@@ -0,0 +1,741 @@
...
+
+void
+fd_txncache_register_root_slot( fd_txncache_t * tc,
+                                ulong           slot ) {
+  fd_rwlock_write( tc->lock );
+
+  ulong idx;
+  for( idx=0UL; idx<tc->root_slots_cnt; idx++ ) {
+    if( FD_UNLIKELY( tc->root_slots[ idx ]==slot ) ) goto unlock;
+    if( FD_UNLIKELY( tc->root_slots[ idx ]>slot ) ) break;
```
ptaffet-jump, line 379: I think this means root slots must be registered in increasing order and re-registering a root slot is a no-op. Both of those should be documented in the function comment.

src/disco/bank/fd_txncache.c
```diff
@@ -0,0 +1,741 @@
...
+}
+
+void
+fd_txncache_register_root_slot( fd_txncache_t * tc,
+                                ulong           slot ) {
+  fd_rwlock_write( tc->lock );
+
+  ulong idx;
+  for( idx=0UL; idx<tc->root_slots_cnt; idx++ ) {
+    if( FD_UNLIKELY( tc->root_slots[ idx ]==slot ) ) goto unlock;
```
ptaffet-jump, line 378: Comment that re-registering a root slot is a no-op in the function comment

src/disco/bank/fd_txncache.c
```diff
@@ -0,0 +1,741 @@
...
+  fd_rwlock_t lock[ 1 ]; /* The txncache is a concurrent structure and will be accessed by multiple threads
+                            concurrently.  Insertion and querying only take a read lock as they can be done
+                            lockless but all other operations will take a write lock internally. */
+
+  ulong  root_slots_max;
+  ulong  live_slots_max;
+  ushort txnpages_per_blockhash_max;
+
+  ulong   root_slots_cnt; /* The number of root slots being tracked in the below array. */
+  ulong * root_slots; /* The highest N slots that have been rooted.  These slots are
```
ptaffet-jump, line 102: Could you add a comment about the way this is used? I was definitely expecting something more like a ring buffer with far fewer memmoves, and that initially confused me.

arjain4 (dismissed):

arjain4 (approved):

PR #1811, opened 2024-05-09
fuzz, flamenco: Add clock, epoch schedule, and rent default sysvars to fuzz harness

mjain-jump:
Solfuzz-agave sets default values for these clock, epoch schedule, and rent sysvars if they are not present in the accounts list. We introduce the same for Firedancer within the fuzz harness context creation.

### DIFF
--- a/src/flamenco/runtime/tests/fd_exec_instr_test.c
+++ b/src/flamenco/runtime/tests/fd_exec_instr_test.c
@@ -11,6 +11,7 @@
 #include "../context/fd_exec_slot_ctx.h"
 #include "../context/fd_exec_txn_ctx.h"
 #include "../sysvar/fd_sysvar_recent_hashes.h"
+#include "../sysvar/fd_sysvar_cache.c"
 #include "../../../funk/fd_funk.h"
 #include "../../../util/bits/fd_float.h"
 #include <assert.h>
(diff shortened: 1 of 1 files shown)


src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -315,8 +338,36 @@ _context_create( fd_exec_instr_test_runner_t *        runner,
...
+  /* We create mock accounts for each of the sysvars and hardcode the data fields before loading it into the account manager */
+  /* We use Agave sysvar defaults for data field values */
+
+  /* Clock */
+  // https://github.com/firedancer-io/solfuzz-agave/blob/agave-v2.0/src/lib.rs#L466-L474
+  uchar clock_raw_data[40] = {10, 0, 0, 0, 0, 0, 0, 0, // slot = 10
+                              0, 0, 0, 0, 0, 0, 0, 0,  // epoch_start_timestamp = 0
+                              0, 0, 0, 0, 0, 0, 0, 0,  // epoch = 0
+                              0, 0, 0, 0, 0, 0, 0, 0,  // leader_schedule_epoch = 0
+                              0, 0, 0, 0, 0, 0, 0, 0}; // unix_timestamp = 0
```
topointon-jump, line 351: If it's not a big pain, do you mind constructing a new `fd_sol_sysvar_clock_t` instance and use the `fd_sol_sysvar_clock_encode` API to encode this into the raw data?

It'll be easier to both write and read and quicker to add new ones in the future.

If it's a pain to change don't worry :)

topointon-jump (dismissed):

topointon-jump (approved):

PR #1812, opened 2024-05-09
fuzz, flamenco: Remove lamport overflow check in fuzz harness

mjain-jump:
This check was added within the fuzz harness since we used to abort on a instruction account lamport sum overflow, but that is no longer the case, so this check is incorrect. 

We should be checking for overflows within the new lamport sum function, and returning an error as appropriate.

### DIFF
--- a/src/flamenco/runtime/info/fd_instr_info.c
+++ b/src/flamenco/runtime/info/fd_instr_info.c
@@ -90,6 +90,10 @@ fd_instr_any_signed( fd_instr_info_t const * info,
   return is_signer;
 }
 
+/*
+  TODO: We should modify this function to handle overflows / other issues similar to Agave.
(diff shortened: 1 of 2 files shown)


src/flamenco/runtime/info/fd_instr_info.c
```diff
@@ -90,6 +90,10 @@ fd_instr_any_signed( fd_instr_info_t const * info,
   return is_signer;
 }
 
+/*
+  TODO: We should modify this function to handle overflows / other issues similar to Agave.
+  https://github.com/anza-xyz/agave/blob/9706a6464665f7ebd6ead47f0d12f853ccacbab9/sdk/src/transaction_context.rs#L407
+*/
 ulong
```
topointon-jump, line 97: Great comment, thank-you for the link!

topointon-jump (approved):

topointon-jump (approved):

PR #1820, opened 2024-05-10
vote_loop_fix: Calling the pop_idx_tail with an invalid filter item c…

jumpsiegel:
…an cause an infinite loop


ripatel-fd:
commit squash would be nice

PR #1824, opened 2024-05-10
genesis: add options for warmup epochs and bootstrap stake

anwayde:


src/app/fdctl/config/default.toml
```diff
@@ -1011,6 +1011,16 @@ dynamic_port_range = "8900-9000"
         # The amount of SOL to pre-fund each account with.
         fund_initial_amount_lamports = 50000000000000
 
+        # The amount of stake for the bootstrap validator in lamports.
+        bootstrap_stake_lamports = 5000000000
```
mmcgee-jump, line 1015: Is this what solana calls it? It seems pretty imprecise, what is "bootstrap"? Should this just be `vote_account_stake_lamports` ? I would expand the comment to specify this clearly as well.

mmcgee-jump (changes requested):
.

src/app/fdctl/config/default.toml
anwayde: Yep, this is what solana calls it: https://github.com/anza-xyz/agave/blob/master/genesis/src/main.rs#L223-L229. Bootstrap validator is the term they use to refer to the validator that starts up from the genesis. It makes sense to me, but I'm okay with calling it whatever makes more sense. It sets the amount that is staked on the (vote account of the) bootstrap validator.

mmcgee-jump: We can probably do better than the Solana CLI in this case. `fund_vote_account_lamports` is the clearest to me.

mmcgee-jump (approved):

PR #1834, opened 2024-05-11
quic - updates to benchs-quic for stability

nbridge-jump:
Moved service ratio code and changed the ratio. This allows multiple transactions to be included in each QUIC packet.
Added in calls to service in additional places, including whenever we have no quic connection. This avoids losing many transactions in the event a connection is closed.
Also removed the DONTWAIT flag from sendmmsg, which seems to have no effect.

This branches off nbridge/quic-ms14-fixes, although there should be no overlap


nbridge-jump:
CI is failing with "too many tiles". I didn't change the number of tiles

ripatel-fd (approved):

PR #1845, opened 2024-05-13
CI - run ledger tests in parallel

llamb-jump:
Tested on all the runners:
`ci10` https://github.com/firedancer-io/firedancer/actions/runs/9071073359/job/24924081545
`ci11` https://github.com/firedancer-io/firedancer/actions/runs/9071069869/job/24924073646
`ci12` https://github.com/firedancer-io/firedancer/actions/runs/9071068595/job/24924068899
`ci13` https://github.com/firedancer-io/firedancer/actions/runs/9071066887/job/24924064166
`ci14` https://github.com/firedancer-io/firedancer/actions/runs/9070984426/job/24923826054


jumpsiegel (dismissed):

.github/workflows/test_ledgers.yml
```diff
@@ -0,0 +1,33 @@
...
+          count_huge: 1000 # TODO: this is required until we can handle anonymouse workspaces and loose huge pages in fddev
+
+      - name: build
+        run: |
+          make -j bin
+      - name: test replay ledgers
+        run: |
+          sudo prlimit --pid=$$ --nofile=1048576
+          sudo prlimit --pid=$$ --memlock=unlimited
+          python3.11 ./src/flamenco/runtime/tests/run_ledger_tests_all.py ./src/flamenco/runtime/tests/run_ledger_tests_all.txt
```
topointon-jump, line 33: Could we maybe use `make run-runtime-test` instead?

We have https://github.com/firedancer-io/firedancer/blob/main/src/flamenco/runtime/tests/Local.mk which has a list of ledger tests in, but now we have 2 lists of ledger tests which could be confusing.

Would be great if we had an authoritative list of the ledger tests.

Thanks so much :)

jumpsiegel (approved):

PR #1860, opened 2024-05-14
flamenco: checkpoint on mismatch

ibhatt-jumptrading:
1. deprecating checkpt_slot option
2. adding cli option for checkpointing at last rooted slot upon mismatch


lheeger-jump (approved):
You can do anything you put your mind to! ⭐️

PR #1866, opened 2024-05-14
Booting fddev off a funky workspace checkpoint

asiegel-jt:


src/app/fdctl/run/tiles/fd_replay.c
```diff
@@ -510,33 +521,52 @@ unprivileged_init( fd_topo_t *      topo,
...
-  if (shmem == NULL)
-    FD_LOG_ERR(( "failed to allocate a funky" ));
-  funk = fd_funk_join( fd_funk_new( shmem, FD_FUNK_MAGIC, ctx->funk_seed, tile->replay.txn_max, tile->replay.index_max ) );
-  if (funk == NULL) {
-    fd_wksp_free_laddr(shmem);
-    FD_LOG_ERR(( "failed to allocate a funky" ));
+  fd_funk_t * funk = NULL;
+  void * shmem = NULL;
+  ctx->snapshot = tile->replay.snapshot;
+  if ( strncmp(ctx->snapshot, "wksp:", 5) == 0 ) {
```
arjain4, line 527: I think we should have a separate field for the wksp name. As far as I can tell snapshot is just getting used as either a snapshot file or a wksp file name which is unclean. These checks should become consistent with all the other arg related checks, instead of checking for a prefix.

src/util/wksp/fd_wksp_io.c
```diff
@@ -492,3 +492,61 @@ fd_wksp_restore( fd_wksp_t *  wksp,
 # undef RBUF_FOOTPRINT
 # undef RBUF_ALIGN
 }
+
+int
+fd_wksp_restore_preview( char const * path,
+                         uint *       seed,
```
arjain4, line 498: nit: rename params to `out_seed` etc for clarity on return params

arjain4 (dismissed):

ibhatt-jumptrading (approved):

src/app/fdctl/run/tiles/fd_replay.c
asiegel-jt: urls will use the http: prefix for snapshot, so I'm just following this convention

lheeger-jump: We need to know the slot number from the snapshot filename. We should probably not be doing HTTP requests this late in start up.

arjain4: Granted the prefix convention might change, but until it does we should have what is current. And even with different prefixes, I don't think this field should be used for both cases. Or it should be renamed to reflect that it can be either

ibhatt-jumptrading (dismissed):

mmcgee-jump (dismissed):
Please remember to squash the commits.

lheeger-jump:
Marked as draft, clogging CI

jumpsiegel (approved):

lheeger-jump (approved):

PR #1885, opened 2024-05-16
ledger: codecov

kbhargava-jump:


.github/codecov.yml
```diff
@@ -9,3 +9,5 @@ flags:
     carryforward: true
   ledgers:
     carryforward: true
+  nightly_ledgers:
```
kbhargava-jump, line 12: carry forward doesn't seem to work yet with my setup, need to look into

kbhargava-jump:
cleaning up

PR #1888, opened 2024-05-16
runtime: fix discriminant in delegate

kbhargava-jump:


lidatong (changes requested):
i dont think that's right. set_state is initializing the stake state. new_stake_state is stack allocated. is there some test failing?

PR #1903, opened 2024-05-17
flamenco: adding slot check to program execution

ibhatt-jumptrading:


ibhatt-jumptrading:
addressed in bpf loader rewrite

PR #1909, opened 2024-05-17
flamenco: gossip version set, no more tmp contact info

lheeger-jump:


src/flamenco/gossip/fd_gossip.c
```diff
@@ -853,7 +805,7 @@ fd_gossip_random_pull( fd_gossip_t * glob, fd_pending_event_arg_t * arg ) {
   (void)arg;
 
   /* Try again in 5 sec */
-  fd_pending_event_t * ev = fd_gossip_add_pending(glob, glob->now + (long)5e9);
+  fd_pending_event_t * ev = fd_gossip_add_pending(glob, glob->now + (long)100e6);
```
arjain4, line 808: what is this change all about?

arjain4 (approved):
looks good

src/flamenco/gossip/fd_gossip.c
lheeger-jump: it was too slow!

lheeger-jump: we will adjust it

PR #1913, opened 2024-05-19
Add TOML fuzz vectors

ripatel-fd:
This PR is fairly large - We might want to move this to a dumping ground repo


ripatel-fd:
@mmcgee-jump I'm assuming you don't want this merged?

ripatel-fd:
@marctrem @nlgripto Can you add these to fuzzcorp? Many thanks

PR #1929, opened 2024-05-22
fuzz, flamenco: fix default sysvar cache loading and add slot bank slot

mjain-jump:
In the Agave fuzz harness, we load default sysvars if there was any failure in decoding the account data. In FD if the data for a sysvar account is corrupt, we load it anyways but don't set the sysvars. We should set default sysvar values if they don't exist.

We now set the slot bank slot as well.

### DIFF
--- a/src/flamenco/runtime/tests/fd_exec_instr_test.c
+++ b/src/flamenco/runtime/tests/fd_exec_instr_test.c
@@ -14,6 +14,7 @@
 #include "../../../funk/fd_funk.h"
 #include "../../../util/bits/fd_float.h"
 #include <assert.h>
+#include "../sysvar/fd_sysvar_cache.h"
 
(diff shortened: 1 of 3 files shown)


src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -14,6 +14,7 @@
 #include "../../../funk/fd_funk.h"
 #include "../../../util/bits/fd_float.h"
 #include <assert.h>
+#include "../sysvar/fd_sysvar_cache.c"
```
topointon-jump, line 17: Can we move the types to the header file instead please? Thanks so much

topointon-jump (changes requested):

topointon-jump (approved):

PR #1932, opened 2024-05-23
keyguard: fix support for multiple signing methods

ripatel-fd:
Previously, support for ed25519_sign(sha256(x)) signing was
implemented ad-hoc in a way that broke ambiguity proofs.

This patch adds the signature scheme as an explicit param to the
signing request body.  Also removes various arbitrary hardcoded
parameters and edge cases from the sign tile.

### DIFF
--- a/src/app/fdctl/run/tiles/fd_sign.c
+++ b/src/app/fdctl/run/tiles/fd_sign.c
@@ -25,6 +25,7 @@ typedef struct {
 
   int               in_role[ MAX_IN ];
   uchar *           in_data[ MAX_IN ];
+  ushort            in_mtu [ MAX_IN ];
 
   fd_sign_out_ctx_t out[ MAX_IN ];
 
@@ -73,9 +74,11 @@ during_frag_sensitive( void * _ctx,
   fd_sign_ctx_t * ctx = (fd_sign_ctx_t *)_ctx;
   FD_TEST( in_idx<MAX_IN );
 
-  int role = ctx->in_role[ in_idx ];
-  if( sz>FD_KEYGUARD_SIGN_REQ_MTU ) {
-    FD_LOG_EMERG(( "oversz signing request (role=%d sz=%lu)", role, sz ));
+  int  role = ctx->in_role[ in_idx ];
+  uint mtu  = ctx->in_mtu [ in_idx ];
+
+  if( sz>mtu ) {
+    FD_LOG_EMERG(( "oversz signing request (role=%d sz=%lu mtu=%u)", role, sz, mtu ));
   }
   fd_memcpy( ctx->_data, ctx->in_data[ in_idx ], sz );
 }
@@ -218,7 +221,9 @@ unprivileged_init_sensitive( fd_topo_t *      topo,
     fd_topo_link_t * in_link = &topo->links[ tile->in_link_id[ i ] ];
     fd_topo_link_t * out_link = &topo->links[ tile->out_link_id[ i ] ];
 
+    if( in_link->mtu > FD_KEYGUARD_SIGN_REQ_MTU ) FD_LOG_CRIT(( "oversz link[%lu].mtu=%lu", i, in_link->mtu ));
     ctx->in_data[ i ] = in_link->dcache;
+    ctx->in_mtu [ i ] = (ushort)in_link->mtu;
 
     ctx->out[ i ].mcache = out_link->mcache;
     ctx->out[ i ].data   = out_link->dcache;
--- a/src/app/fdctl/run/tiles/fd_sign.c
+++ b/src/app/fdctl/run/tiles/fd_sign.c
@@ -23,7 +23,7 @@ typedef struct {
 typedef struct {
   uchar             _data[ FD_KEYGUARD_SIGN_REQ_MTU ];
 
-  ulong             in_role [ MAX_IN ];
+  int               in_role[ MAX_IN ];
   uchar *           in_data[ MAX_IN ];
 
   fd_sign_out_ctx_t out[ MAX_IN ];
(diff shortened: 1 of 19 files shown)


src/app/fdctl/run/tiles/fd_sign.c
```diff
@@ -111,57 +103,48 @@ after_frag( void *             _ctx,
...
-      if( FD_UNLIKELY( !fd_keyguard_payload_authorize( ctx->_data, *opt_sz, FD_KEYGUARD_ROLE_REPAIR ) ) ) {
-        FD_LOG_EMERG(( "fd_keyguard_payload_authorize failed" ));
-      }
-      fd_ed25519_sign( ctx->out[ in_idx ].data, ctx->_data, *opt_sz, ctx->public_key, ctx->private_key, ctx->sha512 );
-      break;
-    }
-    default:
-      FD_LOG_CRIT(( "unexpected link role %lu", ctx->in_role[ in_idx ] ));
+  if( FD_UNLIKELY( !fd_keyguard_payload_authorize( ctx->_data, sz, role, sign_type ) ) ) {
+    FD_LOG_HEXDUMP_WARNING(( "failed signing request", ctx->_data, sz ));
```
mmcgee-jump, line 121: I don't like adding this log because it adds a potential vector for the other process to try and corrupt (eg, passing a lot of data and overflowing a log buffer or something).

Obviously that probably isn't possible, but keeping these vectors minimized is important for the signer, printing user provided buffers doesn't seem smart here.

src/app/fdctl/run/tiles/fd_sign.c
```diff
@@ -111,57 +103,48 @@ after_frag( void *             _ctx,
...
+    FD_LOG_EMERG(( "fd_keyguard_payload_authorize failed (role=%d sign_type=%d)", role, sign_type ));
+  }
+
+  switch( role ) {
```
mmcgee-jump, line 125: Probably do this check before authorize?

src/disco/keyguard/fd_keyguard_match.c
```diff
@@ -183,22 +193,20 @@ fd_keyguard_payload_matches_gossip_msg( uchar const * data,
...
-                                   ulong         sz ) {
+                                   ulong         sz,
+                                   int           sign_type ) {
+
+  if( sign_type != FD_KEYGUARD_SIGN_TYPE_ED25519 ) return 0;
+
   switch( sz ) {
 
   /* Merkle shreds signing payloads always 32 byte */
+  /* FIXME: Sign Merkle shreds using SIGN_TYPE_SHA256_ED25519 (!!!) */
```
mmcgee-jump, line 209: Why is this important? Can we file an issue for 1.5?

mmcgee-jump (changes requested):
.

src/disco/keyguard/fd_keyguard_match.c
mmcgee-jump: thanks

src/app/fdctl/run/tiles/fd_sign.c
```diff
@@ -72,32 +72,24 @@ during_frag( void * _ctx,
...
+  case FD_KEYGUARD_ROLE_TLS:
+  case FD_KEYGUARD_ROLE_GOSSIP:
+  case FD_KEYGUARD_ROLE_REPAIR:
+    break;
+  default:
+    FD_LOG_CRIT(( "unexpected link role %d", role ));
+  }
+
+  if( sz>FD_KEYGUARD_SIGN_REQ_MTU ) {
+    FD_LOG_WARNING(( "oversz signing request (role=%d sz=%lu)", role, sz ));
```
mmcgee-jump, line 88: Again, just because of keyguard I think we should abort here. This points to a programmer error in the caller, which is one more bug away from swiping the key.

mmcgee-jump (changes requested):
.

arjain4 (dismissed):
Prune msg requests might get rejected. Please test against testnet

src/app/fdctl/run/tiles/fd_sign.c
```diff
@@ -72,32 +72,24 @@ during_frag( void * _ctx,
...
+  case FD_KEYGUARD_ROLE_LEADER:
+  case FD_KEYGUARD_ROLE_TLS:
+  case FD_KEYGUARD_ROLE_GOSSIP:
+  case FD_KEYGUARD_ROLE_REPAIR:
+    break;
+  default:
+    FD_LOG_CRIT(( "unexpected link role %d", role ));
+  }
+
+  if( sz>FD_KEYGUARD_SIGN_REQ_MTU ) {
```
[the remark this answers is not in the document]
mmcgee-jump: Oh right lol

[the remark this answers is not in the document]
mmcgee-jump: (I wrote this :'( )

mmcgee-jump (approved):

src/disco/keyguard/fd_keyguard_authorize.c
```diff
@@ -0,0 +1,165 @@
...
+
+int
+fd_keyguard_payload_authorize( fd_keyguard_authority_t const * authority,
+                               uchar const *                   data,
+                               ulong                           sz,
+                               int                             role,
+                               int                             sign_type ) {
+
+  if( sz > FD_KEYGUARD_SIGN_REQ_MTU ) {
+    FD_LOG_WARNING(( "oversz signing request (role=%d sz=%lu)", role, sz ));
```
mmcgee-jump, line 82: Just LOG_EMERG here, programmer error in sign tile is fatal

src/app/fdctl/run/tiles/fd_sign.c
```diff
@@ -73,35 +73,11 @@ during_frag_sensitive( void * _ctx,
...
-        FD_LOG_WARNING(("Corrupt repair signing message with size %lu", sz));
-        *opt_filter = 1;
-        return;
-      }
-      fd_memcpy( ctx->_data, ctx->in_data[ in_idx ], sz );
-      break;
-    default:
-      FD_LOG_CRIT(( "unexpected link role %lu", ctx->in_role[ in_idx ] ));
+  int role = ctx->in_role[ in_idx ];
+  if( sz>FD_KEYGUARD_SIGN_REQ_MTU ) {
```
mmcgee-jump, line 77: Trivial nit but it's the sign tile so fair: this could in some weird cases copy past the end of the input buffer (if dcache size of the LEADER ink is 32UL, and attacker passes 64UL here it would pass, and copy 32 off the end);.

In practice it can't happen and wouldn't do anything anyway, but I think you should restrict this more tightly.

mmcgee-jump (changes requested):
.

mmcgee-jump (approved):

PR #1944, opened 2024-05-23
gossip: use the same gossip listen port value everywhere

### LINKED ISSUE #1942: gossip: use the same port number for the listen port in the tile as the library config value
This is just annoying, we can use the same value.

arjain4:
Closes #1942


llamb-jump (dismissed):

src/app/fdctl/configure/xdp.c
```diff
@@ -86,7 +86,7 @@ init( config_t * const config ) {
     config->tiles.quic.regular_transaction_listen_port,
     config->tiles.quic.quic_transaction_listen_port,
     config->tiles.shred.shred_listen_port,
-    config->tiles.gossip.gossip_listen_port,
+    config->gossip.port,
```
mmcgee-jump, line 89: This change (already) breaks 1.5 since you are going to intercept gossip traffic from reaching the agave code, even when running 1.5 topology.

mmcgee-jump (changes requested):
.

asiegel-jt (approved):

lheeger-jump (approved):

PR #1945, opened 2024-05-23
Module Descriptions

cavemanloverboy:
Every time I navigate this repo I forget what these are lol.

I started this off. Feel free to add, correct, or ignore. Up to y'all.


cavemanloverboy:
i've just been informed this info is in `docs/`. Maybe a simlink to this location would be good

PR #1952, opened 2024-05-26
adding new metrics for replay in storei tile

llamb-jump:


src/app/fdctl/run/tiles/fd_replay.c
```diff
@@ -508,6 +516,10 @@ after_credit( void *             _ctx,
 static void
 during_housekeeping( void * _ctx ) {
   fd_replay_tile_ctx_t * ctx = (fd_replay_tile_ctx_t *)_ctx;
+  FD_MCNT_SET( REPLAY, SNAPSHOT_STATUS_SNAPSHOT_BEGIN, 0 );
```
mmcgee-jump, line 519: I'm not quite sure what these counters are for, but this can't be right. Resetting them to zero will make prometheus think the application exited and restarted, counters should only increment.

llamb-jump: It made the visualization easier to understand IMO, I removed these lines since they're confusing

lheeger-jump (approved):

mmcgee-jump (changes requested):
Can we hold off on this one? Needs a discussion before merging, as 1.5 / 2.x probably shouldn't share metrics definitions.

llamb-jump:
how about splitting `metrics.xml` into `metrics-frankendancer.xml` and `metrics-firedancer.xml`?

mmcgee-jump (approved):

PR #1960, opened 2024-05-29
flamenco: remove fixup memory from epoch ctx

### LINKED ISSUE #1734: flamenco: epoch ctx fixup memory uses dynamic allocation to copy
The fixup memory method copies from a dynamic decode. This causes confusion since we have only one case of this and it would be better to just decode without dynamic allocation.

arjain4:
removes fixup memory and adds a new solana vote account type to store only the required fields.
Closes #1734


lheeger-jump (dismissed):

lheeger-jump (approved):

ripatel-fd:
@arjain4 Can you add some documentation how epoch context memory management works now? 
Where do all these objects live during genesis restore, snapshot restore, and normal operation?

PR #1963, opened 2024-05-29
deploy_mode flag in ELF Loader

ravyu-jump:
The Agave ELF loader introduced additional checks that would fail on (certain) existing mainnet programs. Since it is impossible to retroactively enforce these checks on already deployed programs, a guard flag is used to enable these checks only when deploying programs. This PR implements the equivalent for FD.

### DIFF
--- a/src/ballet/sbpf/fd_sbpf_loader.c
+++ b/src/ballet/sbpf/fd_sbpf_loader.c
@@ -235,7 +235,8 @@ fd_sbpf_load_phdrs( fd_sbpf_elf_info_t *  info,
 static int
 fd_sbpf_load_shdrs( fd_sbpf_elf_info_t *  info,
                     fd_sbpf_elf_t const * elf,
-                    ulong                 elf_sz ) {
+                    ulong                 elf_sz,
+                    bool                  elf_deploy_checks ) {
 
   /* File Header */
   ulong const eh_offset = 0UL;
@@ -277,6 +278,7 @@ fd_sbpf_load_shdrs( fd_sbpf_elf_info_t *  info,
      of the rodata segment.  This is the minimal virtual address range
      that spans all sections.  The offset of each section in virtual
      addressing and file addressing is guaranteed to be the same. */
+  ulong segment_start = FD_SBPF_MM_PROGRAM_ADDR;  /* Lower bound of segment virtual address */
   ulong segment_end    = 0UL;  /* Upper bound of segment virtual address */
   ulong tot_section_sz = 0UL;  /* Size of all sections */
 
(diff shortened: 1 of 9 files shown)


src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -133,7 +134,8 @@ FD_PROTOTYPES_BEGIN
 fd_sbpf_elf_info_t *
 fd_sbpf_elf_peek( fd_sbpf_elf_info_t * info,
                   void const *         bin,
-                  ulong                bin_sz );
+                  ulong                bin_sz,
+                  bool                 deploy_mode );
```
topointon-jump, line 138: Can we document here what `deploy_mode` is?

src/ballet/sbpf/fuzz_sbpf_loader.c
```diff
@@ -52,7 +52,7 @@ LLVMFuzzerTestOneInput( uchar const * data,
     fd_sbpf_syscalls_insert( syscalls, *x );
 
   /* Load program */
-  int res = fd_sbpf_program_load( prog, data, size, syscalls );
+  int res = fd_sbpf_program_load( prog, data, size, syscalls, true );
```
topointon-jump, line 55: We should add `deploy_mode` to the protobuf so that the fuzzer can check both configurations.

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -235,7 +235,8 @@ fd_sbpf_load_phdrs( fd_sbpf_elf_info_t *  info,
 static int
 fd_sbpf_load_shdrs( fd_sbpf_elf_info_t *  info,
                     fd_sbpf_elf_t const * elf,
-                    ulong                 elf_sz ) {
+                    ulong                 elf_sz,
+                    bool                  deploy_mode ) {
```
topointon-jump, line 239: Can we rename `deploy_mode` to `elf_deploy_checks`?

topointon-jump (commented):
Nice!

src/ballet/sbpf/fuzz_sbpf_loader.c
ravyu-jump: Good point!

ravyu-jump: This particular fuzzer is no longer used (and should be removed at some point), but I will work on this in solfuzz

topointon-jump: Yep sorry, I was referring to solfuzz!

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -366,7 +368,11 @@ fd_sbpf_load_shdrs( fd_sbpf_elf_info_t *  info,
 
       ulong sh_actual_size = shdr_get_loaded_size( &shdr[ i ] );
       ulong sh_virtual_end = sh_addr + sh_actual_size;
-      REQUIRE( sh_addr        == sh_offset               );
+
+      /* https://github.com/solana-labs/rbpf/blob/v0.8.0/src/elf.rs#L426 */
+      if ( elf_deploy_checks ){
```
topointon-jump, line 373:
```suggestion
      if ( FD_UNLIKELY( elf_deploy_checks ) ){
```
Could you add these to all the deploy checks?

topointon-jump (dismissed):
Nice!!

topointon-jump (approved):

PR #1965, opened 2024-05-30
update deps.sh

1000xsh:
ensure fetch before install: if install is requested but fetch is not explicitly listed, the script runs fetch before proceeding with install.

some optimizations to accumulates commands and executes them in sequence.

### DIFF
--- a/deps.sh
+++ b/deps.sh
@@ -485,43 +485,27 @@ install () {
 }
 
 ACTION=0
+COMMANDS=()
 while [[ $# -gt 0 ]]; do
   case $1 in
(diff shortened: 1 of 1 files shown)


ripatel-fd:
Thank you for the contribution. The root cause of this issue was incorrect documentation. Operators should always run `./deps.sh` or `./deps.sh nuke`. It is not necessary to pass any other arguments or environment variables. They mostly exist for developers.

Running fetch before install is unwanted behavior for developers. I might need to quickly install local patches to dependency code such as adding debug symbols.

PR #1968, opened 2024-05-31
ci: build fdctl

anwayde:
* We should build `fdctl` during our CI runs
* Based on requirements, we cannot build `fdctl` on `minimal`, `noarch64` or `noarch128`, hence the minimal capable build is updated to `linux_gcc_x86_64`
* We currently don't require secp256k so, added build conditions wherever that library is called during `all` target

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,16 +17,15 @@ jobs:
       # targets from here, not add more.
       matrix:
         machine:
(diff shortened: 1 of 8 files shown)


mmcgee-jump (approved):

lheeger-jump (approved):
I am displeased that we are not building with noarch due to the desire to maintain a nice base for compiling under Arm, PPC, and BPF/SBPF.

We need to resurrect this in the future.

PR #1974, opened 2024-05-31
flamenco, fuzz: capture instr execution return data

ravyu-jump:


kbhargava-jump:
tested this with solana conformance (originally saw `return_data` was not being generated in fd output logs which was causing failures -> now those are passing when comparing with solana)

lheeger-jump (approved):

PR #1978, opened 2024-05-31
flamenco: adding resize delta check

ibhatt-jumptrading:
We were missing the resize delta check which is a transaction level limit on how much data can be changed.


ripatel-fd:
Maybe add a handwritten unit test with a system program invocation that reallocs too much?

ibhatt-jumptrading:
> Maybe add a handwritten unit test with a system program invocation that reallocs too much?

Creating issue for this, will be a lot easier to do when unit test framework that is currently wip is finished

lheeger-jump (approved):

PR #2001, opened 2024-06-04
flamenco, vm: adding memory gap checks for address host translation

ibhatt-jumptrading:
In the address to host translation, we currently don't check against if we are in a gapped region in memory. This adds the check along with a ledger test that tests this case. It is a temporary holdover before moving over and implementing this check in vm-belt-sanding

### DIFF
--- a/src/flamenco/vm/fd_vm_context.c
+++ b/src/flamenco/vm/fd_vm_context.c
@@ -172,9 +172,30 @@ fd_vm_translate_vm_to_host_private( fd_vm_exec_context_t *  ctx,
   ulong end_addr = start_addr + sz;
 
   ulong host_addr = 0UL;
+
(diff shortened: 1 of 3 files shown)


src/flamenco/vm/fd_vm_context.c
```diff
@@ -172,9 +172,24 @@ fd_vm_translate_vm_to_host_private( fd_vm_exec_context_t *  ctx,
   ulong end_addr = start_addr + sz;
 
   ulong host_addr = 0UL;
+
+  /* Note: vm gap shift is calculated when a memory region is created. However,
+     it is always == 63 unless the memory region is writable and gapped. This is
+     onlt the case for the stack region in memory which has a calculated value
+     of 12 for the vm_gap_shift. */
+  ulong vm_gap_shift = 0UL;
```
topointon-jump, line 180: Could you add permalinks for where these are in the Agave source?

ibhatt-jumptrading: Yeah was doing that as you sent the message haha

topointon-jump (approved):

PR #2003, opened 2024-06-04
add sol-compat style test

0x0ece:


contrib/test/run_test_vectors.sh
```diff
@@ -26,14 +26,31 @@ else
   cd ../..
 fi
 
-find dump/test-vectors/instr/fixtures -type f -name '*.fix' -exec ./$OBJDIR/unit-test/test_exec_instr {} + > $LOG_PATH/test_vectors_exec 2>&1
-failed=`grep -w FAIL $LOG_PATH/test_vectors_exec | wc -l`
+find dump/test-vectors/instr/fixtures -type f -name '*.fix' -exec ./$OBJDIR/unit-test/test_exec_instr --log-path $LOG_PATH/test_exec_instr --log-level-stderr 4 {} +
```
0x0ece, line 29: The idea is to call test_exec_sol_compat also here

src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -1012,17 +1013,25 @@ fd_exec_vm_syscall_test_run( fd_exec_instr_test_runner_t *          runner,
   effects->stack->size = (uint)FD_VM_STACK_MAX;
   fd_memcpy( effects->stack->bytes, vm->stack, FD_VM_STACK_MAX );
 
-  effects->inputdata = FD_SCRATCH_ALLOC_APPEND(
-    l, alignof(uchar), PB_BYTES_ARRAY_T_ALLOCSIZE( input_data_sz ) );
-  effects->inputdata->size = (uint)input_data_sz;
-  fd_memcpy( effects->inputdata->bytes, vm->input, input_data_sz );
+  if( input_data_sz ) {
```
0x0ece, line 1016: There's a difference between null vs 0-length array.
By making 0-length array null, the encoded pb is the same, so we can do binary comparison (if we want).

PR #2004, opened 2024-06-05
ci: removing ledger test that is strictly a subset

ibhatt-jumptrading:
This ledger test is not needed and it can cause a race condition with the other folder of the same name causing CI to be flaky


topointon-jump (approved):
🙌

PR #2014, opened 2024-06-06
contrib: support for scripting reproducible ledgers

kbhargava-jump:
Tool to support custom ledger creation


ibhatt-jumptrading (dismissed):

contrib/create-ledgers/src/main.rs
```diff
@@ -0,0 +1,99 @@
...
+fn example_ledger(client: &RpcClient, payer: &Keypair, program_data: &Vec<u8>, account_data: &Vec<u8>) {
+    // Set Up Buffer Accounts
+    let buffer_account_deploy = programs::set_up_buffer_account_transactions(&client, &payer, &program_data);
+    let buffer_account_upgrade = programs::set_up_buffer_account_transactions(&client, &payer, &program_data);
+    let buffer_account_redeploy = programs::set_up_buffer_account_transactions(&client, &payer, &program_data);
+
+    // Deploy Program
+    let (program_account, deploy_program_instructions) = programs::deploy_program_instructions(&client, &payer, None, &buffer_account_deploy, program_data.len());
+    let transaction = utils::create_message_and_sign(&deploy_program_instructions, &payer, vec![&payer, &program_account], client.get_latest_blockhash().unwrap());
+    let _ = client.send_and_confirm_transaction(&transaction);
```
topointon-jump, line 23: We should fail if any of these calls fail.

ibhatt-jumptrading: We want some transactions to fail though

topointon-jump: sure! for cases where we expect them to succeed, we should surface the error not fail silently and continue

contrib/ledger-gen/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+        .expect("Payer file path must be provided as 'payer=/path/to/file'");
+    let payer = read_keypair_file(&payer_path).unwrap();
+
+    // Set Up Program and Account Data
+    let program_data = utils::read_and_verify_elf("helloworld.so").unwrap();
+    let account_data = vec![0u8; 4];
+
+    // ----------------------- ONLY CHANGE BELOW THIS LINE -----------------------
+    ledgers::example_ledger(&rpc_client, &arc_client, &payer, &program_data, &account_data);
+    ledgers::deploy_invoke_same_slot(&rpc_client, &arc_client, &payer, &program_data, &account_data);
```
ravyu-jump, line 52: Put these in a separate function (or the above setup logic, just to make it clear what the user should be looking at)

contrib/ledger-gen/src/programs.rs
```diff
@@ -0,0 +1,260 @@
...
+/// * `program_data` - Program Data
+///
+/// # Returns
+/// * `buffer_account` - Buffer Account Keypair
+///
+/// # Examples
+/// ```
+/// let buffer_account = programs::set_up_buffer_account_transactions(&client, &payer, &program_data);
+/// ```
+pub fn set_up_buffer_account_transactions(client: &Arc<RpcClient>, payer: &Keypair, program_data: &Vec<u8>) -> Keypair {
```
ravyu-jump, line 42: Should probably make the side effects (creating and sending transactions relevant to account creation) clear here. Also rename this to `set_up_buffer_account`

contrib/ledger-gen/src/programs.rs
```diff
@@ -0,0 +1,260 @@
...
+            client.get_inner_client().clone(),
+            "ws://localhost:8900/",
+            solana_client::tpu_client::TpuClientConfig::default(),
+            cache,
+        );
+    let tpu_client = client
+        .runtime()
+        .block_on(tpu_client_fut)
+        .expect("Should return a valid tpu client");
+
```
ravyu-jump, line 102: Would it make sense to have this as a seperate function/moved to utils? Seems like it could be handy in the future (and decomposing improves readability)

kbhargava-jump: dont want parallel transactions elsewhere

contrib/ledger-gen/src/programs.rs
kbhargava-jump: renaming

contrib/ledger-gen/src/main.rs
kbhargava-jump: fine as it, main should just call what ledger we want. nothing else. setup is minimal so fine to leave in main

ibhatt-jumptrading (approved):

PR #2015, opened 2024-06-06
Echo vote transactions through UDP

yhzhangjump:


src/disco/consensus/test_gossip_echo_vote.c
```diff
@@ -140,7 +145,72 @@ gossip_send_fun( uchar const *                 data,
   }
 }
 
+void send_udp_pkt( void * pkt, ulong pkt_sz, fd_wksp_t * wksp, uint dst_ip, ushort dst_port ) {
```
lidatong, line 148: sorry you probably didn't know this. we actually already have a `fd_udpsock_send` in `fd_udpsock`.
https://github.com/firedancer-io/firedancer/blob/demo-v6/src/waltz/udpsock/fd_udpsock.c#L57-L63

 it uses the `aio` abstraction documented here https://github.com/firedancer-io/firedancer/blob/demo-v6/src/waltz/aio/fd_aio.h#L1

can you try refactoring to use that?

lidatong (changes requested):
good work so far see comment

lidatong (approved):

PR #2018, opened 2024-06-06
sbpf: bounds check entry_pc against rodata_sz

llamb-jump:
Addressing the crashes in https://github.com/firedancer-io/auditor-internal/issues/91

The check is basically a copy of https://github.com/firedancer-io/firedancer/blob/32b9530a2115b282b6bb8078b077b81c9cd9019e/src/ballet/sbpf/fd_sbpf_loader.c#L964

### DIFF
--- a/src/ballet/sbpf/fd_sbpf_loader.c
+++ b/src/ballet/sbpf/fd_sbpf_loader.c
@@ -418,22 +418,23 @@ fd_sbpf_load_shdrs( fd_sbpf_elf_info_t *  info,
 
   /* Convert entrypoint offset to program counter */
 
+  info->rodata_sz        = (uint)segment_end;
+  info->rodata_footprint = (uint)elf_sz;
+
   ulong entry_off = fd_ulong_sat_sub( elf->ehdr.e_entry, shdr_text->sh_addr );
   ulong entry_pc = entry_off / 8UL;
   REQUIRE( fd_ulong_is_aligned( entry_off, 8UL ) );
+  REQUIRE( entry_pc < ( info->rodata_sz / 8UL ) );
   info->entry_pc = (uint)entry_pc;
 
   if( (info->shndx_dynstr)>=0 ) {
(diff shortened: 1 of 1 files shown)


llamb-jump:
Does https://github.com/firedancer-io/solana-conformance support these ASAN crashes?

ravyu-jump:
> Does https://github.com/firedancer-io/solana-conformance support these ASAN crashes?

Nope. Like with UBSAN, the dlib loader used there doesn't deal with symbolic references to these libraries

ravyu-jump (approved):

ravyu-jump:
I think this might be a better spot for the check: https://github.com/firedancer-io/firedancer/blob/9293a968b1349cb670a4f72d880e3df700a72916/src/ballet/sbpf/fd_sbpf_loader.c#L422

ravyu-jump:
Just realized we already have a check here

https://github.com/firedancer-io/firedancer/pull/2018/files#diff-04cfcb3b63022935578d4fe4b2c44dd463e58fa1f8ef8e8899c794ed9be8e863R409

(investigating)

ravyu-jump:
Left a comment in https://github.com/firedancer-io/auditor-internal/issues/91

ravyu-jump (approved):
LGTM,

The real issue is that the .text section header's sh_type was mutated to `SHT_NOBITS` by the fuzzer in these crash files.  `shdr_get_loaded_size( shdr_text )` returns 0 there. Might want to think about a fix for that at some point, since there are a couple issues that arise:

- `sh_actual_size` is set to 0 [here](https://github.com/firedancer-io/firedancer/blob/main/src/ballet/sbpf/fd_sbpf_loader.c#L369). Among other things, this causes `rodata_sz` to not encompass the .text region, resulting in `calldests` being under-allocated. 
- `text_cnt` is set to 0.  This is used by the VM so there maybe further ramifications.

ravyu-jump:
Left a follow-up comment here: https://github.com/firedancer-io/auditor-internal/issues/91#issuecomment-2153451285

PR #2031, opened 2024-06-07
add precompiles to CI

0x0ece:


contrib/test/run_test_vectors.sh
```diff
@@ -2,7 +2,7 @@
 
 # this assumes fd_ledger has already been built
 
-set -x
+set -xe
```
0x0ece, line 5: -e will fail if any of the scripts fail, definitely useful to avoid regressions

contrib/test/run_test_vectors.sh
```diff
@@ -42,18 +48,20 @@ else
   exit 0
 fi
```
0x0ece, line 49: I just reverted my changes copying from main. This way when we rebase this part doesn't create issues.

@ravyu-jump note that the elf is not really checked, but the `set -e` above will catch issues.

contrib/test/run_test_vectors.sh
```diff
@@ -42,18 +48,20 @@ else
...
 
-find dump/test-vectors/syscall/fixtures -type f -name '*.fix' -exec ./$OBJDIR/unit-test/test_exec_sol_compat {} + > $LOG_PATH/test_vectors_exec 2>&1
-test_res=$?
-failed=`grep -wR FAIL $LOG_PATH | wc -l`
-echo "Total failed: $failed"
+TESTS=dump/test-vectors/syscall/fixtures
+LOG=$LOG_PATH/test_exec_syscalls
+find $TESTS -type f -name '*.fix' -exec ./$OBJDIR/unit-test/test_exec_sol_compat --log-path $LOG {} +
+if [ "$?" != "0" ]
```
0x0ece, line 54: test_exec_sol_compat already checks if any test fails, and together with the set -e it's safe to just call it and let it fail CI.

I added the if just as an example and to terminate early (but if someone copy&paste and forgets the if, we're good).
There's no need to tail the error logs, because test_exec_sol_compat outputs warnings.

contrib/test/run_test_vectors.sh
ravyu-jump: This will probably cause the test on the icelake machine to fail since it runs with asan and ubsan enabled

contrib/test/run_test_vectors.sh
ravyu-jump: yeah there is actually an issue with elf test vectors atm where we added new fields but didn't update the existing fixtures so I expect this test to fail. Working on updating the fixtures right now.

ripatel-fd:
Converting to draft due to merge conflicts

PR #2060, opened 2024-06-10
flamenco: system program null seed checks

llamb-jump:
https://github.com/firedancer-io/auditor-internal/issues/100

### DIFF
--- a/src/flamenco/runtime/program/fd_system_program.c
+++ b/src/flamenco/runtime/program/fd_system_program.c
@@ -651,6 +651,9 @@ fd_system_program_execute( fd_exec_instr_ctx_t ctx ) {
     if( FD_UNLIKELY( err ) ) return err;
   } while(0);
 
+  if( FD_UNLIKELY( !ctx.instr || !ctx.instr->data ) )
+    return FD_EXECUTOR_INSTR_ERR_FATAL;
+
   /* Deserialize the SystemInstruction enum */
   uchar * data = ctx.instr->data;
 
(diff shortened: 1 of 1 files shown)


src/flamenco/runtime/program/fd_system_program.c
```diff
@@ -651,6 +651,9 @@ fd_system_program_execute( fd_exec_instr_ctx_t ctx ) {
     if( FD_UNLIKELY( err ) ) return err;
   } while(0);
 
+  if( FD_UNLIKELY( !ctx.instr || !ctx.instr->data ) )
+    return FD_EXECUTOR_INSTR_ERR_FATAL;
```
llamb-jump, line 655: @CantelopePeel what should be returned here when these are `NULL`?

mjain-jump, line 655: If we're already this far into execution I think this check is redundant

src/flamenco/runtime/program/fd_system_program.c
```diff
@@ -683,7 +686,7 @@ fd_system_program_execute( fd_exec_instr_ctx_t ctx ) {
     break;
   }
   case fd_system_program_instruction_enum_create_account_with_seed: {
-    if( !VERIFY_SEED_UTF8( instruction.inner.create_account_with_seed.seed ) )
+    if( !instruction.inner.create_account_with_seed.seed || !VERIFY_SEED_UTF8( instruction.inner.create_account_with_seed.seed ) )
```
mjain-jump, line 689: This shouldn't be necessary at this point - the deserialization shouldn't allow this to be null (otherwise deserializing would fail)

mjain-jump (changes requested):

mjain-jump:
I don't think this PR is necessary - these null checks should be caught already before these bits get executed

PR #2064, opened 2024-06-10
flamenco: fix acceptable_reference_epoch_credits head out of bounds

llamb-jump:
https://github.com/firedancer-io/auditor-internal/issues/102

### DIFF
--- a/src/flamenco/runtime/program/fd_stake_program.c
+++ b/src/flamenco/runtime/program/fd_stake_program.c
@@ -640,6 +640,7 @@ acceptable_reference_epoch_credits( fd_vote_epoch_credits_t * epoch_credits,
       ulong vote_epoch = deq_fd_vote_epoch_credits_t_peek_index( epoch_credits, i )->epoch;
       if( vote_epoch != epoch ) { return 0; }
       epoch = fd_ulong_sat_sub( epoch, 1 );
+      if( i == 0 ) break;
     }
     return 1;
   } else {


llamb-jump:
https://github.com/firedancer-io/auditor-internal/issues/102

src/flamenco/runtime/program/fd_stake_program.c
```diff
@@ -640,6 +640,7 @@ acceptable_reference_epoch_credits( fd_vote_epoch_credits_t * epoch_credits,
       ulong vote_epoch = deq_fd_vote_epoch_credits_t_peek_index( epoch_credits, i )->epoch;
       if( vote_epoch != epoch ) { return 0; }
       epoch = fd_ulong_sat_sub( epoch, 1 );
+      if( i == 0 ) break;
```
topointon-jump, line 643: When you fix bugs found on fuzzing, do you mind also putting the fixture in test-vectors so we don't regress? Thanks 🙏

topointon-jump (changes requested):
Can we add the corresponding test fixture to test-vectors, to make sure that the fix is correct and we don't regress in future? 🙏

llamb-jump:
https://github.com/firedancer-io/test-vectors/blob/main/instr/fixtures/stake/a8c850f61c076fecd5fffd6d7ce21871eab030da.fix

topointon-jump (approved):

PR #2067, opened 2024-06-11
`CI`: on_main_push.yml: kick off solfuzz CI

nlgripto:
When a new commit lands on main, we want to kick solfuzz CI.


nlgripto:
@lheeger-jump do you want me to put the content in a separate file as the other entries in this workflow have done?

topointon-jump (approved):

ripatel-fd (changes requested):
@nlgripto Can we use webhooks instead? (e.g. to a lambda function specified by your team). That way, the configuration is entirely on AR's side and decoupled from our CI config.

ripatel-fd (approved):

PR #2068, opened 2024-06-11
fdctl: silence topology message when frankendancer

mmcgee-jump:


ptaffet-jump (approved):
I think your new print message is better and would be fine unconditionally, but this is okay too

PR #2076, opened 2024-06-12
Speedup funk record table with memoize

asiegel-jt:


lidatong (commented):
i'm ok with this change but needs @kbowers-jump approval for change to util

kbowers-jump (approved):
Some polishing to consider in the future (not important enough to hold up this PR):
- Annotate hash == ele->MAP_HASH branches with likely / unlikely.
- Eliminate the list_private_idx as memoization seems to eliminate the need for it.
- Rename the unit test to test_map_giant_memo (instead of mem).
- Added extra validation of the memo in the unit test.
- Clean up query2 (unrelated to this PR but it is just a const cast variant of query).
- Add static assert that the funk txn footprint is the expect amount (much as there current is for the funk rec footprint).

lidatong (approved):

lheeger-jump (approved):
I am a little dubious of how this actually speeds things up. Can you please post a benchmark of the improvement (before and after)?

lheeger-jump (changes requested):
Sorry meant to set this as request changes

lheeger-jump (dismissed):

asiegel-jt:
> Sorry meant to set this as request changes

It saves 6 seconds out of 330 in my test. Not impressive but it doesn't hurt.

lidatong (approved):

PR #2098, opened 2024-06-13
fdctl: rebuild version file if Solana root Cargo.toml changes

ripatel-fd:


src/app/fdctl/version_new.h
```diff
@@ -0,0 +1,3 @@
+#define FDCTL_MAJOR_VERSION 0UL
```
mmcgee-jump, line 1: What is this file?

PR #2109, opened 2024-06-14
flamenco: cast safely with fd_rust_cast_double_to_ulong

llamb-jump:
https://github.com/firedancer-io/auditor-internal/issues/105

also remove `fd_rent_exempt` since it doesn't appear to be used anywhere


kbhargava-jump:
we probably don't want to remove `fd_rent_exempt` in case for future use

also can we add the crash files (from the two audits) to fixtures if they work now

llamb-jump:
I removed `fd_rent_exempt` since it's a duplicate of `fd_rent_exempt_minimum_balance2` we should have one or the other. 

We don't have any crash files associated with this PR, the crash files associated with double to ulong casting where landed in https://github.com/firedancer-io/test-vectors/pull/22.

ripatel-fd (approved):

PR #2127, opened 2024-06-18
Move download_incremental.sh to contrib

ripatel-fd:
Reduces repo root pollution


src/app/fdctl/run/tiles/fd_replay.c
```diff
@@ -686,7 +686,7 @@ read_snapshot( void * _ctx, char const * snapshotfile, char const * incremental
 
       /* Open the command for reading. */
       char cmd[128];
-      snprintf( cmd, sizeof( cmd ), "./download_incremental.sh %s", incremental );
+      snprintf( cmd, sizeof( cmd ), "./contrib/download_incremental.sh %s", incremental );
```
jumpsiegel, line 689: We should think about a better way of finding this script... like a environmental variable or something?

PR #2132, opened 2024-06-18
flamenco: fix asan issue with bpf_loader_v3 deploy_program

llamb-jump:
https://github.com/firedancer-io/auditor-internal/issues/109

We don't do `fd_ulong_sat_add( SIZE_OF_PROGRAM, programdata_len )` because that value is wrong and [only used for the program cache](https://github.com/anza-xyz/agave/blob/da6f7f2d317ed978701b08c3cc3e8d62756e565a/programs/bpf_loader/src/lib.rs#L146) (which we don't have).

https://github.com/firedancer-io/test-vectors/pull/24

### DIFF
--- a/src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
+++ b/src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
@@ -1,5 +1,7 @@
 #include "fd_bpf_loader_v3_program.h"
 
+/* For additional context see https://solana.com/docs/programs/deploying#state-accounts */
+
 #include "../fd_pubkey_utils.h"
 #include "../../../ballet/base58/fd_base58.h"
 #include "../../../ballet/sbpf/fd_sbpf_loader.h"
@@ -134,7 +136,18 @@ calculate_heap_cost( ulong heap_size, ulong heap_cost, int round_up_heap_size, i
   #undef KIBIBYTE_MUL_PAGES_SUB_1
 }
 
-/* https://github.com/anza-xyz/agave/blob/574bae8fefc0ed256b55340b9d87b7689bcdf222/programs/bpf_loader/src/lib.rs#L105-L171 */
+/* https://github.com/anza-xyz/agave/blob/574bae8fefc0ed256b55340b9d87b7689bcdf222/programs/bpf_loader/src/lib.rs#L105-L171
+
+   Our arguments to deploy_program are different from the Agave version because we handle the caching of deployed programs differently.
+   In Firedancer we lack the concept of ProgramCacheEntryType entirely https://github.com/anza-xyz/agave/blob/114d94a25e9631f9bf6349c4b833d7900ef1fb1c/program-runtime/src/loaded_programs.rs#L158
+
+   In Agave there is a separate caching structure that is used to store the deployed programs. In Firedancer the deployed, validated program
+   is stored as metadata for the account in the funk record.
+
+   See https://github.com/firedancer-io/firedancer/blob/9c1df680b3f38bebb0597e089766ec58f3b41e85/src/flamenco/runtime/program/fd_bpf_loader_v3_program.c#L1640
+   for how we handle the concept of 'LoadedProgramType::DelayVisibility' in Firedancer.
+
+   As a concrete example, our version of deploy_program does not have the 'account_size' argument because we do not update the funk record here. */
 int
 deploy_program( fd_exec_instr_ctx_t * instr_ctx,
                 uchar * const         programdata,
@@ -815,10 +828,8 @@ process_loader_upgradeable_instruction( fd_exec_instr_ctx_t * instr_ctx ) {
       return FD_EXECUTOR_INSTR_ERR_ACC_DATA_TOO_SMALL;
     }
 
-    uchar * programdata_data = buffer->data + buffer_data_offset;
-    ulong   programdata_size = fd_ulong_sat_add( SIZE_OF_PROGRAM, programdata_len );
-
-    err = deploy_program( instr_ctx, programdata_data, programdata_size );
+    uchar * buffer_data = buffer->data + buffer_data_offset;
+    err = deploy_program( instr_ctx, buffer_data, buffer_data_len );
     if( FD_UNLIKELY( err ) ) {
       FD_LOG_WARNING(( "Failed to deploy program" ));
       return err;
@@ -1034,16 +1045,16 @@ process_loader_upgradeable_instruction( fd_exec_instr_ctx_t * instr_ctx ) {
       FD_LOG_WARNING(( "Invalid ProgramData account" ));
       return FD_EXECUTOR_INSTR_ERR_INVALID_ACC_DATA;
     }
-    ulong programdata_len = programdata->meta->dlen;
 
     /* https://github.com/anza-xyz/agave/blob/574bae8fefc0ed256b55340b9d87b7689bcdf222/programs/bpf_loader/src/lib.rs#L825-L845 */
     /* Load and verify the program bits */
-    ulong programdata_size = fd_ulong_sat_add( SIZE_OF_PROGRAM, programdata_len );
     if( FD_UNLIKELY( buffer_data_offset>buffer->meta->dlen ) ) {
       FD_LOG_WARNING(( "Buffer data offset is out of bounds" ));
       return FD_EXECUTOR_INSTR_ERR_ACC_DATA_TOO_SMALL;
     }
-    err = deploy_program( instr_ctx, buffer->data+buffer_data_offset, programdata_size );
+
+    uchar * buffer_data = buffer->data + buffer_data_offset;
+    err = deploy_program( instr_ctx, buffer_data, buffer_data_len );
     if( FD_UNLIKELY( err ) ) {
       FD_LOG_WARNING(( "Failed to deploy program" ));
       return err;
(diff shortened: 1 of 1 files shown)


src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
```diff
@@ -815,8 +815,8 @@ process_loader_upgradeable_instruction( fd_exec_instr_ctx_t * instr_ctx ) {
       return FD_EXECUTOR_INSTR_ERR_ACC_DATA_TOO_SMALL;
     }
 
-    uchar * programdata_data = buffer->data + buffer_data_offset;
-    ulong   programdata_size = fd_ulong_sat_add( SIZE_OF_PROGRAM, programdata_len );
+    uchar * programdata_data = buffer->data    + buffer_data_offset;
+    ulong   programdata_size = programdata_len - buffer_data_offset;
```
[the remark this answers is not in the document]
llamb-jump: What comment would you want to see here given the PR description? The original was the divergent behavior

src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
```diff
@@ -815,9 +815,13 @@ process_loader_upgradeable_instruction( fd_exec_instr_ctx_t * instr_ctx ) {
...
 
+    /* We use programdata_size instead of UpgradeableLoaderState::size_of_program().saturating_add(programdata_len)
+       because that is the length of our buffer, and that argument to deploy_program (account_size) in the rust code is not used
+       in our version of deploy_program because we do not have a program cache
+       https://github.com/anza-xyz/agave/blob/da6f7f2d317ed978701b08c3cc3e8d62756e565a/programs/bpf_loader/src/lib.rs#L146 */
```
ravyu-jump, line 824: Need to rewrite this to reflect changes above

src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
```diff
@@ -815,9 +815,13 @@ process_loader_upgradeable_instruction( fd_exec_instr_ctx_t * instr_ctx ) {
...
     err = deploy_program( instr_ctx, programdata_data, programdata_size );
```
ravyu-jump, line 825:
```suggestion
    err = deploy_program( instr_ctx, buffer_data, buffer_data_len );
```

src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
```diff
@@ -1038,12 +1042,19 @@ process_loader_upgradeable_instruction( fd_exec_instr_ctx_t * instr_ctx ) {
...
     /* Load and verify the program bits */
-    ulong programdata_size = fd_ulong_sat_add( SIZE_OF_PROGRAM, programdata_len );
     if( FD_UNLIKELY( buffer_data_offset>buffer->meta->dlen ) ) {
       FD_LOG_WARNING(( "Buffer data offset is out of bounds" ));
       return FD_EXECUTOR_INSTR_ERR_ACC_DATA_TOO_SMALL;
     }
-    err = deploy_program( instr_ctx, buffer->data+buffer_data_offset, programdata_size );
+
+    uchar * programdata_data = buffer->data    + buffer_data_offset;
+    ulong   programdata_size = programdata_len - buffer_data_offset;
```
ravyu-jump, line 1051:
```suggestion
    uchar * buffer_data = buffer->data    + buffer_data_offset;
```

src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
```diff
@@ -1038,12 +1042,19 @@ process_loader_upgradeable_instruction( fd_exec_instr_ctx_t * instr_ctx ) {
...
+
+    /* We use programdata_size instead of UpgradeableLoaderState::size_of_program().saturating_add(programdata_len)
+       because that is the length of our buffer, and that argument to deploy_program (account_size) in the rust code is not used
+       in our version of deploy_program because we do not have a program cache
+       https://github.com/anza-xyz/agave/blob/da6f7f2d317ed978701b08c3cc3e8d62756e565a/programs/bpf_loader/src/lib.rs#L146 */
+    err = deploy_program( instr_ctx, programdata_data, programdata_size );
```
ravyu-jump, line 1057:
```suggestion
    err = deploy_program( instr_ctx, buffer_data, buffer_data_len );
```

src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
```diff
@@ -815,9 +815,13 @@ process_loader_upgradeable_instruction( fd_exec_instr_ctx_t * instr_ctx ) {
       return FD_EXECUTOR_INSTR_ERR_ACC_DATA_TOO_SMALL;
     }
 
-    uchar * programdata_data = buffer->data + buffer_data_offset;
-    ulong   programdata_size = fd_ulong_sat_add( SIZE_OF_PROGRAM, programdata_len );
+    uchar * programdata_data = buffer->data    + buffer_data_offset;
+    ulong   programdata_size = programdata_len - buffer_data_offset;
```
ravyu-jump, line 819:
```suggestion
    uchar * buffer_data = buffer->data    + buffer_data_offset;
```

ravyu-jump (changes requested):
Breaking it down:

 FD's deploy_program takes in `programdata_sz` alongside `programdata_data`, while agave's equivalent just takes a slice (which contains size/length information).

Where the confusion comes from is Agave's deploy_program also takes in an `account_size` argument, which is a different type of size value specific to their program cache implementation. It combines both sizes of the program account (36bytes) and the programdata account. We mistakenly took that for the size of the byte-code we are deploying.

For the first two cases where we invoke `deploy_program`, `programdata_len` is initialized to `PROGRAMDATA_METADATA_SIZE + max_deploy_length`. So the initialization of `programdata_size` to `programdata_len - buffer_data_offset` doesn't make sense since programdata metadata and buffer metadata are different (45 and 37 bytes respectively). 

Furthermore, `max_deploy_length` is the maxium [length that a program can be uploaded to](https://github.com/anza-xyz/agave/blob/3ee204b5cba93211fa0e07dc9238372eeb7933db/sdk/program/src/loader_upgradeable_instruction.rs#L80C9-L80C64). The size of the actual byte-code may be smaller than this value, so passing in `programdata_size` to deploy_checks is also (probably) inaccurate. To be accurate (and to actually follow what Agave [does](https://github.com/anza-xyz/agave/blob/3ee204b5cba93211fa0e07dc9238372eeb7933db/programs/bpf_loader/src/lib.rs#L660)), we should be passing in `buffer_data` pointer and length instead. `programdata_data` is actually already a pointer to the buffer data, so that change is effectively a rename for accuracy. 

This does not apply to the third (and last) place where `deploy_program` is invoked: theextend_program instruction. There is no notion of buffer accounts here. There should be (please correct me if I am wrong) no issue passing a larger-than-actual-size value in the `deploy_program` call in this case, as the bytecode would have been checked during deployment. The check at deployment guarantees two things:

1. Byte-code meets minimum size requirements
2. No out-of-bounds accesses
    - This is what lets us specify larger sizes here

This is a nitpick but `deploy_program` probably needs a rename. 

The [Solana documentation](https://solana.com/docs/programs/deploying#state-accounts) was very useful in understanding the roles of the different state accounts

src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
ravyu-jump: Also the review comments are out of order, if you scroll to the last comment you will see where `buffer_data` comes from

ravyu-jump: This is equivalent to this in Agave: https://github.com/anza-xyz/agave/blob/3ee204b5cba93211fa0e07dc9238372eeb7933db/programs/bpf_loader/src/lib.rs#L839

src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
ravyu-jump: This is equivalent to this in Agave:  https://github.com/anza-xyz/agave/blob/3ee204b5cba93211fa0e07dc9238372eeb7933db/programs/bpf_loader/src/lib.rs#L839

src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
```diff
@@ -1492,7 +1503,15 @@ process_loader_upgradeable_instruction( fd_exec_instr_ctx_t * instr_ctx ) {
...
-    err = deploy_program( instr_ctx, programdata_account->data + PROGRAMDATA_METADATA_SIZE, fd_ulong_sat_add( SIZE_OF_PROGRAM, new_len ) );
+
+    uchar * programdata_data = programdata_account->data + PROGRAMDATA_METADATA_SIZE;
+    ulong   programdata_size = new_len                   - PROGRAMDATA_METADATA_SIZE;
+
+    /* We use programdata_size instead of UpgradeableLoaderState::size_of_program().saturating_add(programdata_len)
+       because that is the length of our buffer, and that argument to deploy_program (account_size) in the rust code is not used
+       in our version of deploy_program because we do not have a program cache
+       https://github.com/anza-xyz/agave/blob/da6f7f2d317ed978701b08c3cc3e8d62756e565a/programs/bpf_loader/src/lib.rs#L146 */
+    err = deploy_program( instr_ctx, programdata_data, programdata_size );
```
ibhatt-jumptrading, line 1514: This seems to be an incorrect permalink

llamb-jump: I removed it and moved the comment to `deploy_program`

src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
llamb-jump: moved comment to top of deploy program

src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
llamb-jump: updated

src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
```diff
@@ -134,7 +136,8 @@ calculate_heap_cost( ulong heap_size, ulong heap_cost, int round_up_heap_size, i
   #undef KIBIBYTE_MUL_PAGES_SUB_1
 }
 
-/* https://github.com/anza-xyz/agave/blob/574bae8fefc0ed256b55340b9d87b7689bcdf222/programs/bpf_loader/src/lib.rs#L105-L171 */
+/* https://github.com/anza-xyz/agave/blob/574bae8fefc0ed256b55340b9d87b7689bcdf222/programs/bpf_loader/src/lib.rs#L105-L171
+   Our arguments to deploy_program are different from the rust version because we handle the caching of deployed programs differently. */
```
ibhatt-jumptrading, line 140: can you describe how the two program caches are different

ravyu-jump (approved):
LGTM. Thanks!

ravyu-jump:
@topointon-jump @ibhatt-jumptrading could either of you take a look as well?

ibhatt-jumptrading (approved):
lgtm

PR #2134, opened 2024-06-18
use offsetof to reduce UBSAN false positives

llamb-jump:
UBSAN gets confused on our use of code like 

`(ulong)&(((DEQUE_(private_t) *)NULL)->deque)` 

which equivalent to 

`offsetof(DEQUE_(private_t), deque)`

This PR changes all of those in `util/tmpl` which removes a large number of UBSAN false positives.


ripatel-fd (changes requested):
It's fine to use https://en.cppreference.com/w/c/types/offsetof (`stddef.h`)

ripatel-fd (approved):

PR #2136, opened 2024-06-19
test_sandbox_build: harden for alarm return codes

jumpsiegel:


ripatel-fd (approved):
Trailing whitespace but otherwise looks good

topointon-jump (approved):

PR #2141, opened 2024-06-20
Tidy up runtime dependency graph

ripatel-fd:
- **runtime: split parts of fd_runtime.c into fd_runtime_init.c**
- **runtime: split executor error codes into fd_executor_err.h**
- **runtime: Optimize imports**

### DIFF
--- a/src/app/ledger/Local.mk
+++ b/src/app/ledger/Local.mk
@@ -1,7 +1,11 @@
 ifdef FD_HAS_ROCKSDB
 
 ifdef FD_HAS_ZSTD
+ifdef FD_HAS_SECP256K1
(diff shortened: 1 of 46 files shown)


ripatel-fd:
Marking as high priority because people have been getting a bunch of link errors with libsecp256k1.a

jumpsiegel:
My concern about this is it is going to touch a great deal of the same files of the vm belt sanding branch which is days/minutes away from being merged.... I see cu consuming functions/etc being moved around.  Could we hold off on merging this until tom merges the vm stuff?

ripatel-fd:
> My concern about this is it is going to touch a great deal of the same files of the vm belt sanding branch which is days/minutes away from being merged.... I see cu consuming functions/etc being moved around. Could we hold off on merging this until tom merges the vm stuff?

Tom mentioned on Slack that there don't appear to be many conflicts. @topointon-jump can you confirm here?

jumpsiegel:
tom says all good.. :)

topointon-jump:
@jumpsiegel I had a look and it's fine to merge this, don't wait on me :)

jumpsiegel (approved):

lheeger-jump (approved):

PR #2145, opened 2024-06-20
loader_v3_inverted_check: uninitialized check is inverted

jumpsiegel:


topointon-jump (approved):
Could you add the fixture to the list when you have it? 🙏

PR #2150, opened 2024-06-20
Use protosol for nanopb definitions

mjain-jump:


ravyu-jump:
suggestion: add `*.proto` to `.gitignore`

src/flamenco/runtime/fd_executor.c
```diff
@@ -466,11 +466,9 @@ static void
 export_account_state( fd_borrowed_account_t * borrowed_account,
                       fd_exec_test_acct_state_t * output_account ) {
     // Address
-    output_account->has_address = true;
     fd_memcpy(output_account->address, borrowed_account->pubkey, sizeof(fd_pubkey_t));
 
     // Lamports
-    output_account->has_lamports = true;
```
ravyu-jump, line 473: Why are we removing these?

src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -120,15 +120,6 @@ _load_account( fd_borrowed_account_t *           acc,
...
                fd_exec_test_acct_state_t const * state ) {
-  /* Although missing addresses / owners may be interpreted as 0-set bytes,
-     the Agave fuzz harness fails to parse these messages out and fails
-     before execution. The following check is meant to keep behavior
-     consistent between fuzz harnesses.
-   */
-  if( !state->has_address || !state->has_owner ) {
-    return 0;
-  }
-
```
ravyu-jump, line 131: does removing this affect anyth?

mjain-jump:
> suggestion: add `*.proto` to `.gitignore`

There's other protobuf files like `fd_solcap.proto` which are FD specific and outside our scope, so we probably shouldn't do that

src/flamenco/runtime/tests/fetch_and_generate.sh
```diff
@@ -0,0 +1,9 @@
+if [ ! -d protosol ]; then
+  git clone --depth=1 https://github.com/firedancer-io/protosol.git
+else
+  cd protosol
+  git pull
+  cd ..
+fi
+
+nanopb_generator.py -I ../.. -I ./protosol/proto -L "" -C ./protosol/proto/*.proto
```
ravyu-jump, line 9: Idea: can we have nanopb_generator a part of the 3rd party deps installed in `deps.sh` and have this point to that?

ravyu-jump: lmk if you want me to look into this

ravyu-jump:
> Requirements: nanopb (generator tool located in nanopb/generator/nanopb_generator.py)

So should we have `nanopb` installed in `src/flamenco/runtime/tests`?

src/flamenco/runtime/fd_executor.c
mjain-jump: I removed explicitly optional types from protobuf definitions since this is default. Removing the optional keyword removes generation of `has_xxx` traits within protobuf message structs

ravyu-jump:
> > suggestion: add `*.proto` to `.gitignore`
> 
> There's other protobuf files like `fd_solcap.proto` which are FD specific and outside our scope, so we probably shouldn't do that

You can add a `.gitignore` within a subdirectory and make it only apply to files in there

src/flamenco/runtime/tests/fd_exec_instr_test.c
mjain-jump: This should be fine

src/flamenco/runtime/fd_executor.c
ravyu-jump: any implications to existing fuzzers/tests that use these?

src/flamenco/runtime/tests/fetch_and_generate.sh
mjain-jump: That wouldn't be a bad idea

mjain-jump:
> > > suggestion: add `*.proto` to `.gitignore`
> > 
> > 
> > There's other protobuf files like `fd_solcap.proto` which are FD specific and outside our scope, so we probably shouldn't do that
> 
> You can add a `.gitignore` within a subdirectory and make it only apply to files in there

The `protosol` directory and the associated .proto files aren't included in this PR. Should we include it in the gitignore anyways to prevent someone from accidentally committing the repo?

src/flamenco/runtime/tests/fetch_and_generate.sh
topointon-jump: we should definitely do that, some older versions of nanopb have trouble compiling the newer protobuf format

topointon-jump (approved):
🎉

topointon-jump (approved):

PR #2154, opened 2024-06-20
capture: add a stub implementation of fd_solcap_writer_t

ripatel-fd:
Allows executing Solana transactions on non-hosted targets.

### DIFF
--- a/src/flamenco/capture/fd_solcap_writer_stub.c
+++ b/src/flamenco/capture/fd_solcap_writer_stub.c
@@ -1,0 +1,84 @@
+#include "fd_solcap_writer.h"
+
+/* This file provides a stub implementation of fd_solcap_writer for
+   non-hosted targets. */
+
+struct fd_solcap_writer {
+  uchar dummy;
+};
+
+ulong
+fd_solcap_writer_align( void ) {
+  return alignof(fd_solcap_writer_t);
+}
+
+ulong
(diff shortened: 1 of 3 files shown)


src/flamenco/capture/fd_solcap_writer_stub.c
lheeger-jump, line 1: Can you use FD_PARAM_UNUSED instead of the void cast.

lheeger-jump (dismissed):

src/flamenco/capture/fd_solcap_writer_stub.c
ripatel-fd: @lheeger-jump hmm I'm not sure if that works because that might cause the declaration and definition to have different function signatures.

lheeger-jump: You just do it in the definition, no?

ripatel-fd: I'll try. I'm personally more of a fan of the `(void)` syntax since it's more portable and util uses it. 
But I'll change it to use the param attribute.

https://github.com/firedancer-io/firedancer/blob/2bf20d526ad9bd6122c53ac49c79390a7746cbd1/src/util/tpool/test_tpool.c#L130

lidatong (approved):

PR #2163, opened 2024-06-21
test_list: switch to using preset lists for test-vectors

jumpsiegel:


topointon-jump (dismissed):
Nice!! Can we call these `elf-loader-fixtures.list` and `instr-fixtures.list` to make it clear they correspond to directories in test-vectors? 🙌

ravyu-jump (approved):
Good idea

mjain-jump (approved):

PR #2172, opened 2024-06-21
we arent merging this just testing a ci job

llamb-jump:


ripatel-fd:
Last time I touched CI, all the workflows had manual workflow_dispatch triggers on them that allowed you to trigger them via Actions UI btw.

llamb-jump:
I didn't see a way to run `test` in Actions which was surprising https://github.com/firedancer-io/firedancer/blob/ll/add-fixture-for-%232132/.github/workflows/tests.yml

ripatel-fd:
@lheeger-jump please fix

PR #2177, opened 2024-06-23
More FreeBSD fixes

ripatel-fd:
Now builds without errors on FreeBSD 14.1:

```
export MACHINE=freebsd_clang_noarch128
gmake -j
```


ripatel-fd:
Will open a new PR

PR #2191, opened 2024-06-24
deps: snappy, rocksdb to compile on mac

0x0ece:
Upgraded deps, so that snappy compiles on mac.
For rocksdb, removing `haswell` leads to uninitialized issues on avx512, so added the `-Wno-uninitialized`


ripatel-fd:
@0x0ece Needs to be rebased to resolve conflicts.

ripatel-fd (approved):

PR #2192, opened 2024-06-24
Fixed download from load balanced servers

asiegel-jt:


src/flamenco/snapshot/fd_snapshot_http.c
```diff
@@ -75,19 +76,13 @@ fd_snapshot_http_new( void *               mem,
...
-  p = fd_cstr_append_text( p, hdr_part1, sizeof(hdr_part1)-1 );
-
-  p = fd_cstr_append_printf( p, FD_IP4_ADDR_FMT ":%u",
-                             FD_IP4_ADDR_FMT_ARGS( dst_ipv4 ), dst_port );
-  static char const hdr_part2[] =
-    "\r\n"
+    "host: %s\r\n"
     "\r\n";
-  p = fd_cstr_append_text( p, hdr_part2, sizeof(hdr_part2)-1 );
+  p = fd_cstr_append_printf( p, hdr, dst_str );
```
ripatel-fd, line 85: Please avoid use of unsafe printf. This isn't bounds checked

ripatel-fd: (Was fine for IPv4 because the longest IP address is still pretty short, but more scary for a user provided string via redirects)

src/flamenco/snapshot/fd_snapshot_loader.c
```diff
@@ -256,6 +256,8 @@ fd_snapshot_src_parse( fd_snapshot_src_t * src,
     cstr[ m_hostname->rm_eo ] = '\0';
     char * hostname = cstr + m_hostname->rm_so;
 
+    strncpy( src->http.dest, hostname, sizeof(src->http.dest)-1 );
```
ripatel-fd, line 259: Needs code to set the trailing 0 byte

lidatong (approved):

lidatong (changes requested):
content length is bugged

lidatong (approved):

PR #2194, opened 2024-06-24
tests: enable test vectors for syscalls and precompiles

0x0ece:


ravyu-jump:
Do we want a whitelist file for these as well?

PR #2196, opened 2024-06-24
New fdctl parser

ripatel-fd:


src/ballet/toml/fd_toml.h
```diff
@@ -43,8 +44,12 @@ FD_PROTOTYPES_BEGIN
      inline table  | x={a=1,b=2} | subpod
      inline array  | x=[1,2]     | subpod (keys %d formatted cstrs)
      bool          | true        | int
-     integer       | -3          | ulong (two's complement)
-     float         | 3e-3        | double
+     dec int       | -3          | long
+     hex int       | 0x3         | ulong
```
mmcgee-jump, line 48: IMO these should map to `long` as well, according to the TOML spec there is no ulong, and the type of a field shouldn't change just because the user changed it from 0x1 to 1

src/ballet/toml/fd_toml.h
```diff
@@ -43,8 +44,12 @@ FD_PROTOTYPES_BEGIN
...
+     oct int       | 0o3         | ulong
+     bin int       | 0b11        | ulong
+     float         | 3e-3        | float
+     float special | "+inf"      | cstr ("-inf", "+inf", "nan")
```
mmcgee-jump, line 52: I also don't see why this needs to exist? Just map +inf to the +Infinity float encoding?

src/ballet/toml/fd_toml.h
```diff
@@ -43,8 +44,12 @@ FD_PROTOTYPES_BEGIN
      inline table  | x={a=1,b=2} | subpod
      inline array  | x=[1,2]     | subpod (keys %d formatted cstrs)
      bool          | true        | int
```
mmcgee-jump, line 46: IMO this is a limitation of the pod that breaks the API, since consumers won't be able to tell if it was specified as `true` in the TOML or `1` (we probably wouldn't want to accept `1` for booleans).

I feel like fd_pod should almost add a bool type because of this.

src/ballet/toml/fd_toml.c
```diff
@@ -1282,14 +1293,18 @@ fd_float_parse_float_special( fd_toml_parser_t * parser,
   fd_toml_advance_inline( parser, 3UL );
 
   if( 0==memcmp( str, "inf", 3 ) ) {
-    FD_LOG_WARNING(( "float infinity is unsupported" ));
-    *pres = base * FLT_MAX;
+    if( FD_UNLIKELY( !fd_pod_insert_cstr( parser->pod, parser->key, neg ? "-inf" : "+inf" ) ) ) {
```
mmcgee-jump, line 1296: why not just construct a inf valued float?

src/app/fdctl/config_parse.c
```diff
@@ -0,0 +1,358 @@
...
+# define FDCTL_CFG_MAX_DEPTH (16)
+
+static ulong
+fdctl_pod_find_leftover_recurse( uchar *       pod,
+                                 char const ** stack,
+                                 ulong         depth ) {
+
+  if( FD_UNLIKELY( depth+1 >= FDCTL_CFG_MAX_DEPTH ) ) {
+    FD_LOG_WARNING(( "configuration file has too many nested keys" ));
+    return 0UL;
```
mmcgee-jump, line 116: I think this should probably return 1?

src/app/fdctl/config_parse.c
```diff
@@ -0,0 +1,358 @@
...
+  char * c   = fd_cstr_init( path );
+  char * end = path + 64*FDCTL_CFG_MAX_DEPTH - 1;
+  for( ulong j=0UL; j<depth; j++ ) {
+    char const * key     = stack[j];
+    ulong        key_len = strlen( key );
+    if( c+key_len+1 >= end ) {
+      c = fd_cstr_append_text( c, "...", 3UL );
+      break;
+    }
+    c = fd_cstr_append_text( c, key, key_len );
```
mmcgee-jump, line 152: How do you know the key will fit in the buffer?

src/app/fdctl/config_parse.c
```diff
@@ -0,0 +1,358 @@
...
+    depth--;
+  }
+
+  return 0UL;
+}
+
+static int
+fdctl_pod_find_leftover( uchar * pod ) {
+
+  static char const * stack[ FDCTL_CFG_MAX_DEPTH ];
```
mmcgee-jump, line 138: You can do this without the gross `stack` here because the recursive function call is creating a stack for you.

Just have the function take a

```
struct parent_pod {
  char const * key;
  struct parent_pod * parent;
}
```

stick one on the stack in each nested fn call and refer it to the parent, then you can print when you hit an unknown key (you also don't need the max recursion depth and other kinda checks)

mmcgee-jump (changes requested):
.

src/ballet/toml/fd_toml.c
ripatel-fd: @mmcgee-jump We compile with `-ffast-math`. Under that model, INF and NAN do not exist.

src/app/fdctl/config_parse.c
ripatel-fd: There is a bounds check 3 lines above

src/app/fdctl/config_parse.c
ripatel-fd: @mmcgee-jump This doesn't work because as the call stack unwinds that stack memory becomes free. Memory accesses to such memory would fail under ASan.

src/ballet/toml/fd_toml.h
ripatel-fd: I agree, but this is out of scope for this change. 
When using the TOML parser the integer literal 1 gets encoded as a LONG though, not an INT.

src/ballet/toml/fd_toml.h
ripatel-fd: See my other comment.

src/ballet/toml/fd_toml.h
ripatel-fd: Sounds good

src/app/fdctl/config_parse.c
mmcgee-jump: You would print the warning deep while in the stack though, not after

src/ballet/toml/fd_toml.h
ripatel-fd: Changed the code to reject NaN and infinity with error.

src/app/fdctl/config_parse.c
ripatel-fd: Makes sense, good idea

mmcgee-jump (approved):

PR #2220, opened 2024-06-26
fdctl: make version detection more robust

### LINKED ISSUE #2190: version.h is flakey
I've ran into bugs where version.h contains something along the lines of `#define FD_MINOR_VERSION UL` on several OSes.
I suspect it's a race condition trying to build fdctl for the first time when Rust 1.75.0 is not yet installed.

When this condition occurs Make fails. Rebuilding does not fix the problem.
The resolution is to delete `version.h` and then rebuild. 

This bug needs to be investigated further. 
We should probably replace the Makefile code that generates `version.h` with a more robust bash script.

ripatel-fd:
- Move version number generation from Makefile to script (POSIX compliant) 
- Add extensive error checking
- Remove conflicting FD_VERSION macros from fd_util_base.h

Closes #2190


src/app/fdctl/with-version.mk
```diff
@@ -1,22 +0,0 @@
...
-# which release candidate branch this is. Different Firedancer release
-# branches could point to the same Solana Labs patch.
-FIREDANCER_VERSION_MINOR := $(VERSION_MINOR)$(shell printf "%02d" $(VERSION_PATCH))
-
-# For Frankendancer, we stuff the entire Solana Labs version that we are
-# linking to in the patch version.  This transforms, for example, a full
-# Solana Labs version of "1.17.8" to a minor version of "11708".
-FIREDANCER_VERSION_PATCH := $(shell grep -Po "(?<=^version = \").*(?=\")" "solana/Cargo.toml" | awk -F. '{ printf "%d%02d%02d\n", $$1, $$2, $$3 }')
-
-export FIREDANCER_VERSION_MAJOR
```
mmcgee-jump, line 17: These are used by Agave to access the version and report it to gossip etc. Can't be completely removed

mmcgee-jump: Oh nvm

contrib/tag-release.py
```diff
@@ -1,32 +1,16 @@
...
-        else:
-            print('Error: version.mk file is not well formatted')
-            exit(1)
-
-    if version_major is None or version_minor is None or version_patch is None:
-        print('Error: version.mk file is not well formatted')
-        exit(1)
+    version_txt = open('src/app/fdctl/version.txt', 'r').read().strip()
+    version_parts = version_txt.split('.')
+    version_major = int(version_parts[0])
```
mmcgee-jump, line 4: `version_major` is always 0, the version in this file is actually <minor>.<patch>

contrib/tag-release.py
```diff
@@ -47,15 +31,9 @@ def main():
 
     solana_version = f'{solana_version_major}{solana_version_minor:02d}{solana_version_patch:02d}'
 
-    with open('src/app/fdctl/version.mk', 'w') as f:
```
mmcgee-jump, line 50: You need to write back `<minor>.<patch>` to the version.txt as part of this, so tag-release keeps incrementing it.

PR #2227, opened 2024-06-26
ci: fix clusterfuzz publish script

ravyu-jump:
Submodules need to be fetched so that fdctl's version.h [generator script](https://github.com/firedancer-io/firedancer/blob/7541b4bb0645a311922132f60959f03e30aa0bf7/src/app/fdctl/with-version.mk#L15) can correctly determine FIREDANCER_VERSION_PATCH

This fixes the clusterfuzz ci scripts


ravyu-jump:
~~@nlgripto should we just remove these scripts entirely? Given that we don't use clusterfuzz~~

I believe we are still using clusterfuzz

nlgripto (approved):

PR #2230, opened 2024-06-26
empty_lockouts: address some empty lockouts and states

jumpsiegel:


ripatel-fd (changes requested):
`run-test-vectors] Error 123`

topointon-jump (approved):

PR #2234, opened 2024-06-26
disco: txn cache fixes and changes

arjain4:


lheeger-jump (approved):

src/disco/bank/fd_txncache.h
```diff
@@ -197,6 +197,7 @@ struct fd_txncache_insert {
   uchar const * blockhash;
   uchar const * txnhash;
   ulong         slot;
+  ulong         txnhash_offset;
```
mmcgee-jump, line 200: This doesn't make sense on the insert API, typically in the insert case the caller does not know the txnhash_offset.

You should add a separate snapshot_restore type functionality and pass a txnhash_offset only there.

src/disco/bank/fd_txncache.c
```diff
@@ -345,25 +346,66 @@ fd_txncache_delete( void * shtc ) {
...
+  memcpy( tc->txnpages_free+tc->txnpages_free_cnt, tc->blockcache[ idx ].pages, tc->blockcache[ idx ].pages_cnt*sizeof(ushort) );
+  tc->txnpages_free_cnt += tc->blockcache[ idx ].pages_cnt;
+
+  /* Check if this a probed entry or not. If this is an entry inserted
+     at the intended "hash index" we still need to check if there are
+     any probed entries and shift one to this index. */
+  ulong hash_idx = FD_LOAD(ulong, tc->blockcache[ idx ].blockhash);
+  if( hash_idx%tc->live_slots_max != idx ) return;
+
+  /* We linear search for an entry which hashed to the same index and move it to
```
mmcgee-jump, line 362: This fix isn't correct. Relocating an probed entry might then cause a gap that needs to itself have a reprobed item moved into it.

PR #2237, opened 2024-06-26
quic: handle RESET_STREAM frame

ripatel-fd:
See RFC 9000 Section 19.4

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -6280,17 +6280,81 @@ fd_quic_frame_handle_ecn_counts_frag(
 
 static ulong
 fd_quic_frame_handle_reset_stream_frame(
-    void * context,
+    void *                         vp_context,
     fd_quic_reset_stream_frame_t * data,
-    uchar const * p,
-    ulong p_sz) {
-  (void)context;
-  (void)data;
-  (void)p;
-  (void)p_sz;
-  /* ack-eliciting */
-  /* TODO implement */
-  return FD_QUIC_PARSE_FAIL;
+    uchar const *                  p    FD_PARAM_UNUSED,
+    ulong                          p_sz FD_PARAM_UNUSED
+) {
+
+  /* The peer uses the RESET_STREAM frame to indicate it abandons its
+     send interest on the stream. */
+
+  fd_quic_frame_context_t * context = vp_context;
+  fd_quic_t *               quic    = context->quic;
+  fd_quic_conn_t *          conn    = context->conn;
+  fd_quic_pkt_t *           pkt     = context->pkt;
+
+  ulong stream_id        = data->stream_id;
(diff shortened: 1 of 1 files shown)


src/waltz/quic/fd_quic.c
```diff
@@ -6253,17 +6253,82 @@ fd_quic_frame_handle_ecn_counts_frag(
...
+  /* Notify client that local recv interest is reset */
+  stream->state |= FD_QUIC_STREAM_STATE_RX_FIN;
+  if( stream->state & FD_QUIC_STREAM_STATE_TX_FIN || !bidir ) {
+    fd_quic_stream_free( quic, conn, stream, FD_QUIC_NOTIFY_RESET );  /* calls notify */
+  } else {
+    fd_quic_cb_stream_notify( quic, stream, stream->context, FD_QUIC_NOTIFY_RESET );
+  }
+
+ack:
+  pkt->ack_flag |= ACK_FLAG_RQD;
```
ripatel-fd, line 6330: I'm only setting the ACK flag on successful handling of the frame. 
Should I also set it when abandoning the connection with an error?

nbridge-jump: I doubt it makes much difference. 
As a technicality and in general, it should probably ACK only-if all the frames were handled successfully. If any frames were faulty or ignored/unprocessed, then none of the frames should have any effect, and the ACK should not be sent. In any case, if some frame were unable to be processed, we should probably close the connection with an error, like internal-error or protocol-violation, or some more specific error.

src/waltz/quic/fd_quic.c
```diff
@@ -6253,17 +6253,82 @@ fd_quic_frame_handle_ecn_counts_frag(
...
+    if( FD_UNLIKELY( !bidir ) ) {
+      /* This stream type can only be initiated locally and is uni-
+         directional local-to-peer.  See RFC 9000 Section 19.4:
+         > An endpoint that receives a RESET_STREAM frame for a
+         > send-only stream MUST terminate the connection with error
+         > STREAM_STATE_ERROR. */
+      fd_quic_conn_error( conn, FD_QUIC_CONN_REASON_STREAM_STATE_ERROR, __LINE__ );
+      return FD_QUIC_PARSE_FAIL;
+    } else {
+      /* No-op, can't reason about peer-initiated bidirectional stream
```
nbridge-jump, line 6289: This is in the local_initiated+bidir branch

src/waltz/quic/fd_quic.c
nbridge-jump: In this case, I would not send ACK

ripatel-fd:
Before merging, I have to add a unit test for this one

nbridge-jump (approved):

ripatel-fd:
Closing this because I'm about to send a PR that makes this redundant.

PR #2244, opened 2024-06-27
treap_buffer: Allocate some buffer room

jumpsiegel:
lets give ourselves some space for if you have 64 voters and want to add 1 more


anarcheuz:
@jumpsiegel just wondering, is it maybe possible that more than 1 element past the real size is added to the treap ? since for some of the changes we are using values such as 1_000 in the fd_ulong_max(...), couldn't we also change FD_VOTE_AUTHORIZED_VOTERS_MIN and FD_STAKE_HISTORY_MIN to a big value too ?

src/flamenco/types/gen_stubs.py
```diff
@@ -874,7 +874,7 @@ def emitDecodeUnsafe(self):
             print(f'  fd_bincode_uint64_decode_unsafe( &{self.name}_len, ctx );', file=body)
 
         if self.minalloc > 0:
-            print(f'  self->{self.name}_pool = {mapname}_alloc( ctx->valloc, fd_ulong_max({self.name}_len, {self.minalloc} ) );', file=body)
+            print(f'  self->{self.name}_pool = {mapname}_alloc( ctx->valloc, fd_ulong_max({self.name}_len + 1, {self.minalloc} ) );', file=body)
```
topointon-jump, line 877: Should we change this in the vote program, and not in the stub generator?

ripatel-fd:
@jumpsiegel This looks super out of date. Closing. Feel free to reopen.

PR #2247, opened 2024-06-27
flamenco, util: add missing add-hdr declarations 

ravyu-jump:
This allows flamenco to be used as a library


ripatel-fd (approved):

ripatel-fd:
Sorry @mmcgee-jump

PR #2254, opened 2024-06-28
replay: get leader contact info and send vote txns to the leader TPU

yhzhangjump:
This PR creates 2 links: a `gossip->replay` link passes contact information (IP/port of TPU vote) to replay tile and a `replay->net` link helps the replay tile send vote txns. 

I wonder whether we should merge this PR -- most of the code should be removed later and moved to the new tile that Liam is working on. It also doesn't hurt to merge it because the code will only be executed if `consensus.vote = true` in the toml.

### DIFF
--- a/src/app/fdctl/run/tiles/fd_gossip.c
+++ b/src/app/fdctl/run/tiles/fd_gossip.c
@@ -35,6 +35,7 @@
 #define PACK_OUT_IDX    2
 #define SIGN_OUT_IDX    3
 #define REPLAY_OUT_IDX  4
+#define VOTE_CONTACT_OUT_IDX    5
 
 #define CONTACT_INFO_PUBLISH_TIME_NS ((long)5e9)
 
@@ -123,6 +124,16 @@ struct fd_gossip_tile_ctx {
   ulong       replay_out_chunk0;
   ulong       replay_out_wmark;
   ulong       replay_out_chunk;
+
+  fd_frag_meta_t * vote_contact_out_mcache;
+  ulong *          vote_contact_out_sync;
+  ulong            vote_contact_out_depth;
+  ulong            vote_contact_out_seq;
+
+  fd_wksp_t * vote_contact_out_mem;
+  ulong       vote_contact_out_chunk0;
+  ulong       vote_contact_out_wmark;
+  ulong       vote_contact_out_chunk;
 
   fd_wksp_t *     wksp;
   fd_gossip_peer_addr_t gossip_my_addr;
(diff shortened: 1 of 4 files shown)


src/app/fdctl/run/tiles/fd_gossip.c
```diff
@@ -35,6 +35,7 @@
 #define PACK_OUT_IDX    2
 #define SIGN_OUT_IDX    3
 #define REPLAY_OUT_IDX  4
+#define VOTE_CONTACT_OUT_IDX    5
```
lidatong, line 38: this should just be contact out, not really specific to vote

src/app/fdctl/run/tiles/fd_replay.c
```diff
@@ -520,19 +588,74 @@ after_frag( void *             _ctx,
...
+      hdr->udp->net_dport   = fd_ushort_bswap( dest->port );
+      hdr->udp->net_len     = fd_ushort_bswap( (ushort)(vote_txn_sz + sizeof(fd_udp_hdr_t)) );
+      hdr->udp->check       = fd_ip4_udp_check( *(uint *)FD_ADDRESS_OF_PACKED_MEMBER( hdr->ip4->saddr_c ),
+                                                *(uint *)FD_ADDRESS_OF_PACKED_MEMBER( hdr->ip4->daddr_c ),
+                                                (fd_udp_hdr_t const *)FD_ADDRESS_OF_PACKED_MEMBER( hdr->udp ),
+                                                packet + sizeof(fd_net_hdrs_t) );
+
+      /* Send the vote transaction through net tile */
+      ulong tspub = fd_frag_meta_ts_comp( fd_tickcount() );
+      ulong tsorig = fd_frag_meta_ts_comp( fd_tickcount() );
```
lidatong, line 648: can you put all this in a function `send_vote` in the replay tile that takes in the tile ctx and vote_txn as arguments ?

lidatong (changes requested):

src/app/fdctl/run/tiles/fd_gossip.c
yhzhangjump: Hmm, it's actually true that the contact info sent out through this link is specific to TPU vote.

That's why you could find `ele->contact_info.tpu_vote.addr` and `ele->contact_info.tpu_vote.port` here: https://github.com/firedancer-io/firedancer/pull/2254/files#diff-4990d28d56483a6cad7d22a6dd0d2a4fd5c14196dfa762c00176b62e7aaf4a83R494

src/app/fdctl/run/tiles/fd_replay.c
yhzhangjump: Done. The code is much cleaner now.

lidatong (approved):

PR #2263, opened 2024-07-01
quic: gracefully ignore STOP_SENDING frame

ripatel-fd:
TPU only uses short-lived streams so we shouldn't ever see a
STOP_SENDING frame from a peer.  This patch changes frame handling
logic to ignore but ACK STOP_SENDING frames instead of dropping
the packet.

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -6257,17 +6257,14 @@ fd_quic_frame_handle_reset_stream_frame(
 
 static ulong
 fd_quic_frame_handle_stop_sending_frame(
-    void * context,
+    void *                         vp_context,
     fd_quic_stop_sending_frame_t * data,
-    uchar const * p,
-    ulong p_sz) {
-  (void)context;
-  (void)data;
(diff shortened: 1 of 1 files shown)


src/waltz/quic/fd_quic.c
```diff
@@ -6257,16 +6257,13 @@ fd_quic_frame_handle_reset_stream_frame(
...
-  (void)p;
-  (void)p_sz;
-  /* ack-eliciting */
-  /* TODO implement */
+    uchar const *                  p,
+    ulong                          p_sz ) {
+  (void)data; (void)p; (void)p_sz;
+  fd_quic_frame_context_t * context = vp_context;
+  context->pkt->ack_flag |= ACK_FLAG_RQD;  /* ack-eliciting */
   return FD_QUIC_PARSE_FAIL;
```
nbridge-jump, line 6267: You need to return SUCCESS here

nbridge-jump (changes requested):

nbridge-jump (approved):

PR #2265, opened 2024-07-01
fdctl: fix config bugs

ripatel-fd:
- Ensure fdctl_cfg_from_env zero-initializes config_t
- Fix incorrect return code in fdctl_pod_to_cfg
- Add unit test rejecting invalid config keys
- Add fdctl_cfg_validate API to validate fields from default.toml
- Add unit test asserting that parsing default.toml works
- Add unit test selectively overriding a field

### DIFF
--- a/src/app/fdctl/config_parse.c
+++ b/src/app/fdctl/config_parse.c
@@ -1,5 +1,7 @@
 #include "config_parse.h"
 
+FD_IMPORT_BINARY( fdctl_default_config, "src/app/fdctl/config/default.toml" );
+
 /* Pod query utils ****************************************************/
 
 static int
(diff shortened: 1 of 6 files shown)


src/app/fdctl/config_parse.c
```diff
@@ -1,5 +1,7 @@
 #include "config_parse.h"
 
+FD_IMPORT_BINARY( fdctl_default_config, "src/app/fdctl/config/default.toml" );
```
mmcgee-jump, line 3: The makefile has an explicit link dependency for this you need to update

ripatel-fd: Ah, awesome, we should do this everywhere we use FD_IMPORT

jumpsiegel (approved):

PR #2285, opened 2024-07-02
ci: use rpath to link libraries

anwayde:


lheeger-jump (commented):
I believe you can just do EXTRAS="rpath" no? no need for the extra variable.

lheeger-jump (approved):

PR #2307, opened 2024-07-03
fd_stake_ci: degrade gracefully if cluster has >40k staked nodes

ptaffet-jump:
This is a pretty subtle change, and we need to test it more thoroughly before merging.
The Rust side is also only pushed to a branch at the moment.


mmcgee-jump:
I think the dcache size constants in fd_frankendancer.c need to change as well?

ptaffet-jump:
Good catch, thanks! I also fixed the test and the place in the replay tile that sends stake weights over.

lheeger-jump (dismissed):

mmcgee-jump (approved):

mmcgee-jump (approved):

PR #2313, opened 2024-07-05
flamenco, vm: direct mapping

ibhatt-jumptrading:
direct mapping

### DIFF
--- a/src/flamenco/runtime/fd_account.c
+++ b/src/flamenco/runtime/fd_account.c
@@ -71,16 +71,19 @@ fd_account_set_lamports( fd_exec_instr_ctx_t const * ctx,
 
   /* An account not owned by the program cannot have its blanace decrease */
   if( FD_UNLIKELY( ( !fd_account_is_owned_by_current_program( ctx->instr, account->const_meta ) ) &&
-                   ( lamports < account->const_meta->info.lamports ) ) )
+                   ( lamports < account->const_meta->info.lamports ) ) ) {
     return FD_EXECUTOR_INSTR_ERR_EXTERNAL_ACCOUNT_LAMPORT_SPEND;
+  }
 
   /* The balance of read-only may not change */
-  if( FD_UNLIKELY( !fd_instr_acc_is_writable_idx( ctx->instr, instr_acc_idx ) ) )
+  if( FD_UNLIKELY( !fd_instr_acc_is_writable_idx( ctx->instr, instr_acc_idx ) ) ) {
     return FD_EXECUTOR_INSTR_ERR_READONLY_LAMPORT_CHANGE;
+  }
 
   /* The balance of executable accounts may not change */
-  if( FD_UNLIKELY( fd_account_is_executable( account->const_meta ) ) )
+  if( FD_UNLIKELY( fd_account_is_executable( account->const_meta ) ) ) {
     return FD_EXECUTOR_INSTR_ERR_EXECUTABLE_LAMPORT_CHANGE;
+  }
 
   /* Don't touch the account if the lamports do not change */
   if( lamports==account->const_meta->info.lamports ) {
@@ -137,8 +140,6 @@ fd_account_set_data_from_slice( fd_exec_instr_ctx_t const * ctx,
                                 uchar const *               data,
                                 ulong                       data_sz ) {
 
-  int err = FD_EXECUTOR_INSTR_SUCCESS;
-
   fd_borrowed_account_t * account = NULL;
   do {
     int err = fd_instr_borrowed_account_view_idx( ctx, (uchar)instr_acc_idx, &account );
@@ -147,15 +148,19 @@ fd_account_set_data_from_slice( fd_exec_instr_ctx_t const * ctx,
     }
   } while(0);
 
-  if( !fd_account_can_data_be_resized( ctx, account->const_meta, data_sz, &err ) ) {
+  int err;
+  if( FD_UNLIKELY( !fd_account_can_data_be_resized( ctx, account->const_meta, data_sz, &err ) ) ) {
     return err;
   }
 
-  if( !fd_account_can_data_be_changed( ctx->instr, instr_acc_idx, &err ) ) {
+  if( FD_UNLIKELY( !fd_account_can_data_be_changed( ctx->instr, instr_acc_idx, &err ) ) ) {
     return err;
   }
 
-  if( !fd_account_update_accounts_resize_delta( ctx, instr_acc_idx, data_sz, &err ) ) {
+  /* touch() */
+  account->meta->slot = ctx->slot_ctx->slot_bank.slot;
+
+  if( FD_UNLIKELY( !fd_account_update_accounts_resize_delta( ctx, instr_acc_idx, data_sz, &err ) ) ) {
     return err;
   }
 
@@ -166,11 +171,10 @@ fd_account_set_data_from_slice( fd_exec_instr_ctx_t const * ctx,
     }
   } while(0);
 
-  /* self.touch() */
-  account->meta->slot = ctx->slot_ctx->slot_bank.slot;
-
-  assert( account->meta->dlen >= data_sz );
+  /* AccountSharedData::set_data_from_slice() */
+  account->meta->dlen = data_sz;
   fd_memcpy( account->data, data, data_sz );
+
   return FD_EXECUTOR_INSTR_SUCCESS;
 }
 
--- a/src/flamenco/runtime/program/fd_bpf_loader_serialization.c
+++ b/src/flamenco/runtime/program/fd_bpf_loader_serialization.c
@@ -4,92 +4,235 @@
 #pragma GCC diagnostic ignored "-Wformat"
 #pragma GCC diagnostic ignored "-Wformat-extra-args"
 
-/**
- * num accounts
- * serialized accounts
- * instr data len
- * instr data
- * program id public key
-*/
-// 64-bit aligned
+/* As a general note, copy_account_data implies that direct mapping is not being
+   used/is inactive. This file is responsible for serializing and deserializing
+   the input region of the BPF virtual machine. The input region contains
+   instruction information, account metadata, and account data. The high level
+   format is as follows: 
+   
+   [ account 1 metadata, account 1 data, account 2 metadata, account 2 data, ...,
+     acccount N metadata, account N data, instruction info. ]
+
+  This format by no means comprehensive, but it should give an idea of how 
+  the input region is laid out. When direct mapping is not enabled, the input
+  region is stored as a single contigious buffer. This buffer in the host 
+  address space is then mapped to the VM virtual address space (the range 
+  starting with 0x400...). This means to serialize into the input region, we
+  need to copy in the account metadata and account data into the buffer for
+  each account. Everything must get copied out after execution is complete. 
+  A consequence of this is that a memcpy for the account data is required 
+  for each serialize and deserialize operation: this can potentially become
+  expensive if there are many accounts and many nested CPI calls. Also, the 
+  entire memory region is treated as writable even though many accounts are
+  read-only. This means that for all read-only accounts, a memcmp must be done
+  while deserializing to make sure that the account (meta)data has not changed.
+
+  Direct mapping offers a solution to this by introducing a more sophisticated
+  memory translation protocol. Now the account data is not copied into a single
+  contiguous buffer, but instead a borrowed account's data is directly mapped
+  into the VM's virtual address space. The host memory for the input region is
+  now represented by a list of fragmented memory regions. These sub regions
+  also have different write permissions. This should solve the problem of 
+  having to memcpy/memcmp account data regions (which can be up to 10MiB each).
+  There is some nuance to this, as the account data can be resized. This means
+  that memcpys for account data regions can't totally be avoided. */
+
+/* Add a new memory region to represent the input region. All of the memory
+   regions here have sorted virtual addresses. These regions may or may not
+   correspond to an account's data region. If it corresponds to metadata,
+   the pubkey for the rgion will be NULL. */
+static void
+new_input_mem_region( fd_vm_input_region_t * input_mem_regions,
+                      uint *                 input_mem_regions_cnt,
+                      const uchar *          buffer,
+                      ulong                  region_sz,
+                      uint                   is_writable,
+                      fd_pubkey_t * const    pubkey ) {
+
+  /* The start vaddr of the new region should be equal to start of the previous
+     region added to its size. */
+  ulong vaddr_offset = *input_mem_regions_cnt==0UL ? 0UL : input_mem_regions[ *input_mem_regions_cnt-1U ].vaddr_offset + 
+                                                           input_mem_regions[ *input_mem_regions_cnt-1U ].region_sz;
+  input_mem_regions[ *input_mem_regions_cnt ].is_writable  = is_writable;
+  input_mem_regions[ *input_mem_regions_cnt ].haddr        = (ulong)buffer;
+  input_mem_regions[ *input_mem_regions_cnt ].region_sz    = (uint)region_sz;
+  input_mem_regions[ *input_mem_regions_cnt ].vaddr_offset = vaddr_offset;
+  input_mem_regions[ *input_mem_regions_cnt ].pubkey       = pubkey;
+  (*input_mem_regions_cnt)++;
+}
+
+/* https://github.com/anza-xyz/agave/blob/b5f5c3cdd3f9a5859c49ebc27221dc27e143d760/programs/bpf_loader/src/serialization.rs#L93-L130 */
+/* This function handles casing for direct mapping being enabled as well as if
+   the alignment is being stored. In the case where direct mapping is not 
+   enabled, we copy in the account data and a 10KiB buffer into the input reigon.
+   These both go into the same memory buffer. However, when direct mapping is
+   enabled, the account data and resizing buffers are represented by two
+   different memory regions. In both cases, padding is used to maintain 8 byte
+   alignment. If alignment is not required, then a resizing buffer is not used
+   as the deprecated loader doesn't allow for resizing accounts. */
+void
+write_account( fd_exec_instr_ctx_t *     instr_ctx,
+               fd_borrowed_account_t *   account,
+               uchar                     instr_acc_idx,
+               uchar * *                 serialized_params,
+               uchar * *                 serialized_params_start,
+               fd_vm_input_region_t *    input_mem_regions,
+               uint *                    input_mem_regions_cnt,
+               fd_vm_acc_region_meta_t * acc_region_metas,
+               int                       is_aligned,
+               int                       copy_account_data ) {
+  if( copy_account_data ) {
+    /* Copy the account data into input region buffer */
+    fd_memcpy( *serialized_params, account->const_data, account->const_meta->dlen );
+    *serialized_params += account->const_meta->dlen;
+
+    if( FD_LIKELY( is_aligned ) ) {
+      /* Zero out padding bytes and max permitted data increase */
+      ulong align_offset = fd_ulong_align_up( account->const_meta->dlen, FD_BPF_ALIGN_OF_U128 ) - account->const_meta->dlen;
+      fd_memset( *serialized_params, 0, MAX_PERMITTED_DATA_INCREASE + align_offset );
+      *serialized_params += MAX_PERMITTED_DATA_INCREASE + align_offset;
+    }
+  } else { /* direct_mapping == true */
+    /* First, push on the region for the metadata that has just been serialized.
+       This function will push the metadata in the serialized_params from
+       serialized_params_start to serialized_params as a region to the input
+       memory regions array. */
+    /* TODO: This region always has length of 96 and this can be set as a constant. */
+
+    ulong region_sz = (ulong)(*serialized_params) - (ulong)(*serialized_params_start);
+    new_input_mem_region( input_mem_regions, input_mem_regions_cnt, *serialized_params_start, region_sz, 1, NULL );
+
+    /* Next, push the region for the account data if there is account data. We 
+       intentionally omit copy on write as a region type. */
+    int err = 0;
+    uint is_writable = (uint)(fd_account_can_data_be_changed( instr_ctx->instr, instr_acc_idx, &err ) && !err);
+
+    /* Update the mapping from instruction account index to memory region index.
+       This is an optimization to avoid redunant lookups to find accounts. */
+    acc_region_metas[instr_acc_idx] = (fd_vm_acc_region_meta_t){ .region_idx          = *input_mem_regions_cnt, 
+                                                                 .has_data_region     = !!account->const_meta->dlen,
+                                                                 .has_resizing_region = (uchar)is_aligned };
+    
+    if( account->const_meta->dlen ) {
+      new_input_mem_region( input_mem_regions, input_mem_regions_cnt, account->const_data, account->const_meta->dlen, is_writable, account->pubkey );
+    }
+
+    if( FD_LIKELY( is_aligned ) ) {
+      /* Finally, push a third region for the max resizing data region. This is
+         done even if there is no account data. This must be aligned so padding
+         bytes must be inserted. This resizing region is also padded to result
+         in 8 byte alignment for the combination of the account data region with
+         the resizing region.
+         
+         We add the max permitted resizing limit along with 8 bytes of padding
+         to the serialization buffer. However, the padding bytes are used to
+         maintain alignment in the VM virtual address space. */
+      ulong align_offset = fd_ulong_align_up( account->const_meta->dlen, FD_BPF_ALIGN_OF_U128 ) - account->const_meta->dlen;
+
+      fd_memset( *serialized_params, 0, MAX_PERMITTED_DATA_INCREASE + FD_BPF_ALIGN_OF_U128 );
+
+      /* Leave a gap for alignment */
+      uchar * region_buffer = *serialized_params + (FD_BPF_ALIGN_OF_U128 - align_offset);
+      ulong   region_sz     = MAX_PERMITTED_DATA_INCREASE + align_offset;
+      new_input_mem_region( input_mem_regions, input_mem_regions_cnt, region_buffer, region_sz, is_writable, account->pubkey );
+
+      *serialized_params += MAX_PERMITTED_DATA_INCREASE + FD_BPF_ALIGN_OF_U128;
+    }
+    *serialized_params_start = *serialized_params;
+  }
+}
+
 uchar *
-fd_bpf_loader_input_serialize_aligned( fd_exec_instr_ctx_t ctx,
-                                       ulong *             sz,
-                                       ulong *             pre_lens ) {
-  ulong serialized_size = 0;
+fd_bpf_loader_input_serialize_aligned( fd_exec_instr_ctx_t       ctx,
+                                       ulong *                   sz,
+                                       ulong *                   pre_lens,
+                                       fd_vm_input_region_t *    input_mem_regions,    
+                                       uint *                    input_mem_regions_cnt,
+                                       fd_vm_acc_region_meta_t * acc_region_metas,
+                                       int                       copy_account_data ) {
   uchar const * instr_acc_idxs = ctx.instr->acct_txn_idxs;
-  fd_pubkey_t * txn_accs = ctx.txn_ctx->accounts;
+  fd_pubkey_t * txn_accs       = ctx.txn_ctx->accounts;
 
-  uchar acc_idx_seen[256];
-  ushort dup_acc_idx[256];
-  memset(acc_idx_seen, 0, sizeof(acc_idx_seen));
-  memset(dup_acc_idx, 0, sizeof(dup_acc_idx));
+  uchar acc_idx_seen[256] = {0};
+  ushort dup_acc_idx[256] = {0};
 
+  /* https://github.com/anza-xyz/agave/blob/b5f5c3cdd3f9a5859c49ebc27221dc27e143d760/programs/bpf_loader/src/serialization.rs#L429-L459 */
+  ulong serialized_size = 0UL;
   serialized_size += sizeof(ulong);
-  for( ushort i = 0; i < ctx.instr->acct_cnt; i++ ) {
+  /* First pass is to calculate size of buffer to allocate */
+  for( ushort i=0; i<ctx.instr->acct_cnt; i++ ) {
     uchar acc_idx = instr_acc_idxs[i];
 
     serialized_size++; // dup byte
     if( FD_UNLIKELY( acc_idx_seen[acc_idx] ) ) {
-      serialized_size += 7; // pad to 64-bit alignment
+      serialized_size += 7UL; // pad to 64-bit alignment
     } else {
       acc_idx_seen[acc_idx] = 1;
-      dup_acc_idx[acc_idx] = i;
-      fd_pubkey_t * acc = &txn_accs[acc_idx];
+      dup_acc_idx[acc_idx]  = i;
+      fd_pubkey_t *           acc      = &txn_accs[acc_idx];
       fd_borrowed_account_t * view_acc = NULL;
       int read_result = fd_instr_borrowed_account_view( &ctx, acc, &view_acc );
       fd_account_meta_t const * metadata = view_acc->const_meta;
 
-      ulong acc_data_len = 0;
-      if ( FD_LIKELY( read_result == FD_ACC_MGR_SUCCESS ) ) {
+      ulong acc_data_len = 0UL;
+      if( FD_LIKELY( read_result==FD_ACC_MGR_SUCCESS ) ) {
         acc_data_len = metadata->dlen;
-      } else if ( FD_UNLIKELY( read_result == FD_ACC_MGR_ERR_UNKNOWN_ACCOUNT ) ) {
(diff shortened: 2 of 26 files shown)


src/flamenco/runtime/program/fd_bpf_loader_serialization.c
```diff
@@ -224,6 +231,9 @@ fd_bpf_loader_input_deserialize_aligned( fd_exec_instr_ctx_t ctx,
     fd_borrowed_account_t * view_acc = NULL;
     /* Intentionally ignore the borrowed account error because it is not used */
     fd_instr_borrowed_account_view( &ctx, acc, &view_acc );
+    if( FD_UNLIKELY( !view_acc ) ) {
+      return FD_EXECUTOR_INSTR_ERR_INVALID_ACC_DATA;
+    }
```
topointon-jump, line 236: Doesn't this change the behaviour? Previously, if the account was unknown it was silently skipped, now we throw an error.

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
```diff
@@ -274,46 +287,65 @@ fd_bpf_loader_input_deserialize_aligned( fd_exec_instr_ctx_t ctx,
...
-        int modify_err = fd_instr_borrowed_account_modify( &ctx, acc, post_len, &modify_acc );
-        if( FD_UNLIKELY( modify_err != FD_ACC_MGR_SUCCESS ) ) {
-          fd_valloc_free( ctx.valloc, buffer );
-          return FD_EXECUTOR_INSTR_ERR_ACC_BORROW_FAILED;
-        }
-
-        ulong pre_len = pre_lens[i];
-
-        start += fd_ulong_align_up( pre_len, FD_BPF_ALIGN_OF_U128 );
+        if( copy_account_data ) {
```
topointon-jump, line 290: nit: wonder if it would be clearer to use `direct_mapping_enabled` everywhere instead of `copy_account_data`.

topointon-jump (commented):
Nice! 

A bit concerned about the changes to fd_account_set_data_from_slice, let's talk about these before merging.

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
```diff
@@ -40,56 +40,61 @@ fd_bpf_loader_input_serialize_aligned( fd_exec_instr_ctx_t ctx,
...
+      serialized_size += sizeof(uchar)               // is_signer
+                       + sizeof(uchar)               // is_writable
+                       + sizeof(uchar)               // is_executable
+                       + sizeof(uint)                // original_data_len
+                       + sizeof(fd_pubkey_t)         // key
+                       + sizeof(fd_pubkey_t)         // owner
+                       + sizeof(ulong)               // lamports
+                       + sizeof(ulong)               // data_len
+                       + MAX_PERMITTED_DATA_INCREASE
+                       + sizeof(ulong);              // rent_epoch
```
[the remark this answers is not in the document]
ibhatt-jumptrading: I kind of disagree. It's bad code style and it makes it harder to read/debug the code.

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
ibhatt-jumptrading: Was just matching up to agave logic

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
```diff
@@ -123,20 +128,21 @@ fd_bpf_loader_input_serialize_aligned( fd_exec_instr_ctx_t ctx,
...
+      /* https://github.com/anza-xyz/agave/blob/b5f5c3cdd3f9a5859c49ebc27221dc27e143d760/programs/bpf_loader/src/serialization.rs#L465 */
       fd_account_meta_t const * metadata = view_acc->const_meta;
       uchar const * acc_data             = view_acc->const_data;
 
       uchar is_signer = (uchar)fd_instr_acc_is_signer_idx( ctx.instr, (uchar)i );
       FD_STORE( uchar, serialized_params, is_signer );
       serialized_params += sizeof(uchar);
 
-      uchar is_writable = (uchar)(fd_instr_acc_is_writable_idx( ctx.instr, (uchar)i ) && !fd_pubkey_is_sysvar_id( acc ) && !fd_pubkey_is_builtin_program( acc ));
+      uchar is_writable = (uchar)fd_instr_acc_is_writable_idx( ctx.instr, (uchar)i );
```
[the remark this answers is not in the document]
ibhatt-jumptrading: My best guess is that this was a hack before we addressed the demotion logic properly. I don't think a comment in the code is necessary as this matches up exactly to what agave does

src/flamenco/runtime/fd_account.c
```diff
@@ -135,40 +135,17 @@ fd_account_set_data_from_slice( fd_exec_instr_ctx_t const * ctx,
...
-
-  if( !fd_account_can_data_be_changed( ctx->instr, instr_acc_idx, &err ) ) {
-    return err;
-  }
-
-  if( !fd_account_update_accounts_resize_delta( ctx, instr_acc_idx, data_sz, &err ) ) {
-    return err;
-  }
-
   do {
```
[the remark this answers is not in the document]
ibhatt-jumptrading: I will discuss this offline with you

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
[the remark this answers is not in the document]
ibhatt-jumptrading: I'll seperate out codestyle fixes in the future into a different pr

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
ibhatt-jumptrading: We should not silently skip when we deserialize and can't find an account. This was intentional

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
```diff
@@ -4,92 +4,231 @@
...
+    uchar         acc_idx = instr_acc_idxs[i];
+    fd_pubkey_t * acc     = &txn_accs[acc_idx];
 
     if( FD_UNLIKELY( acc_idx_seen[acc_idx] && dup_acc_idx[acc_idx] != i ) ) {
-      // Duplicate
-      FD_STORE( ulong, serialized_params, 0 );
+      /* Duplicate. Store 8 byte buffer to maintain alignment but store the
+         account index in the first byte.*/
+      FD_STORE( ulong, serialized_params, 0UL );
       FD_STORE( uchar, serialized_params, (uchar)dup_acc_idx[acc_idx] );
```
topointon-jump, line 220: Are these two the correct way round? Agave serializes them in little-endian: https://github.com/anza-xyz/agave/blob/36323b6dcd3e29e4d6fe6d73d716a3f33927148b/programs/bpf_loader/src/serialization.rs#L488-L489

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
```diff
@@ -4,92 +4,231 @@
...
-  serialized_size += sizeof(ulong)
-      + ctx.instr->data_sz
-      + sizeof(fd_pubkey_t);
-  uchar * serialized_params = fd_valloc_malloc( ctx.valloc, 8UL, serialized_size );
-  uchar * serialized_params_start = serialized_params;
+  serialized_size += sizeof(ulong)        // data len 
+                  +  ctx.instr->data_sz
+                  +  sizeof(fd_pubkey_t); // program id
+
+  uchar * serialized_params            = fd_valloc_malloc( ctx.valloc, FD_BPF_ALIGN_OF_U128, serialized_size );
```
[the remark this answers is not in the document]
topointon-jump: Also, should this buffer be 1-byte aligned? As we write to it at unaligned addresses.

topointon-jump (commented):
Nice work!!

Halfway through review 😄

src/flamenco/vm/fd_vm_private.h
```diff
@@ -184,16 +196,86 @@ fd_vm_mem_cfg( fd_vm_t * vm ) {
...
 
    fd_vm_mem_haddr_fast is when the vaddr is for use when it is already
-   known that the vaddr region has a valid mapping. */
+   known that the vaddr region has a valid mapping.
+   
+   These assumptions don't hold if direct mapping is enabled since input
+   region lookups become O(log(n)). */
+
+static inline ulong
+fd_vm_get_input_mem_region_idx( fd_vm_t const * vm, ulong offset ) {
```
topointon-jump, line 205: What is the intended behaviour if offset is outside the range of all memory regions? Should this function propagate an error? It would be great to document this, as well as the supported number of regions etc.

src/flamenco/vm/fd_vm_private.h
```diff
@@ -184,16 +196,86 @@ fd_vm_mem_cfg( fd_vm_t * vm ) {
...
+  while( left<right ) {
+    mid = (left+right) / 2U;
+    if( offset>=vm->input_mem_regions[mid].vaddr_offset ) {
+      /* Exit condition: found the right region */
+      if( mid+1==vm->input_mem_regions_cnt || offset<vm->input_mem_regions[mid+1].vaddr_offset ) {
+        break;
+      }
+      left = mid + 1U;
+    } else {
+      right = mid - 1U;
```
topointon-jump, line 219: This can underflow if we have only 1 region and `offset < vm->input_mem_regions[mid].vaddr_offset`. I know that practically this can never happen, but it's probably a good idea to guard against the underflow in this function in case that invariant changes (defense in depth).

src/flamenco/vm/fd_vm_private.h
```diff
@@ -184,16 +196,86 @@ fd_vm_mem_cfg( fd_vm_t * vm ) {
...
+    }
+  }
+  if( left==right ) {
+    mid = left;
+  }
+  left = mid;
+  return left;
```
topointon-jump, line 226: I think we can just return mid here, no need for these lines of code.

src/flamenco/vm/fd_vm.h
```diff
@@ -18,6 +18,30 @@ typedef struct fd_vm fd_vm_t;
...
+/* fd_vm_acc_to_mem_t holds metadata about a given account.  An array of these
+   structs will map an instruction account index to its respective input memory
+   region location. */
+
+struct __attribute((aligned(8UL))) fd_vm_acc_to_mem {
+   uint  region_idx;
+   uchar has_data_region;
+   uchar has_resizing_region;        
+};
+typedef struct fd_vm_acc_to_mem fd_vm_acc_to_mem_t;
```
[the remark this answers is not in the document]
ibhatt-jumptrading: I was racking my head to think of a better name. I like this more

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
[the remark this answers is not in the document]
ibhatt-jumptrading: I think the start of the buffer should be 8 byte aligned. Good catch with the buffer size

src/flamenco/vm/fd_vm_private.h
ibhatt-jumptrading: isn't the same because of exit conditions. Will clean up though

src/flamenco/vm/fd_vm_private.h
ibhatt-jumptrading: no longer applicable

src/flamenco/runtime/program/fd_bpf_loader_v3_program.c
```diff
@@ -376,9 +378,13 @@ execute( fd_exec_instr_ctx_t * instr_ctx, fd_sbpf_validated_program_t * prog ) {
...
-  ulong input_sz = 0;
-  ulong pre_lens[ 256UL ];
-  uchar * input = fd_bpf_loader_input_serialize_aligned( *instr_ctx, &input_sz, pre_lens );
+  ulong                input_sz              = 0UL;
+  ulong                pre_lens[256]         = {0};
+  fd_vm_input_region_t mem_regions[1000]     = {0}; /* We can have a max of (3 * num accounts + 1) regions */
+  fd_vm_acc_to_mem_t   mem_regions_accs[256] = {0}; /* instr acc idx to idx */
+  uint                 mem_regions_cnt       = 0U;
+  int                  direct_mapping        = FD_FEATURE_ACTIVE( instr_ctx->slot_ctx, bpf_account_data_direct_mapping );
+  uchar * input = fd_bpf_loader_input_serialize_aligned( *instr_ctx, &input_sz, pre_lens, mem_regions, &mem_regions_cnt, mem_regions_accs, !direct_mapping );
```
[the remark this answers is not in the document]
topointon-jump: @ibhatt-jumptrading I already created this ticket: https://github.com/firedancer-io/firedancer-private/issues/960

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
```diff
@@ -520,6 +745,7 @@ VM_SYSCALL_CPI_ENTRYPOINT( void *  _vm,
     return FD_VM_ERR_INSTR_ERR;
   }
   
+  //FD_LOG_NOTICE(("LEAVING CPI"));
```
topointon-jump, line 748: nit: remove

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
```diff
@@ -501,13 +695,44 @@ VM_SYSCALL_CPI_ENTRYPOINT( void *  _vm,
...
+      }
+    }
+  }
+
   /* Update the caller accounts with any changes made by the callee during CPI execution */
   for( ulong i=0UL; i < caller_accounts_to_update_len; i++ ) {
-    fd_pubkey_t const * callee = &vm->instr_ctx->instr->acct_pubkeys[callee_account_keys[i]];
-    err = VM_SYSCALL_CPI_UPDATE_CALLER_ACC_FUNC(vm, &acc_infos[caller_accounts_to_update[i]], callee);
-    if( FD_UNLIKELY( err ) ) return err;
+    if( fd_instr_acc_is_writable_idx( vm->instr_ctx->instr, callee_account_keys[i] ) ) {
```
topointon-jump, line 731: Similarly, where is this logic in Agave?

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
```diff
@@ -501,13 +695,44 @@ VM_SYSCALL_CPI_ENTRYPOINT( void *  _vm,
...
+  /* https://github.com/anza-xyz/agave/blob/b5f5c3cdd3f9a5859c49ebc27221dc27e143d760/programs/bpf_loader/src/syscalls/cpi.rs#L1128-L1145 */
+  /* Update all account permissions before updating the account data updates.
+     We have inlined the anza function update_caller_account_perms here.
+     TODO: consider factoring this out */
+  if( FD_FEATURE_ACTIVE( vm->instr_ctx->slot_ctx, bpf_account_data_direct_mapping ) ) {
+    for( ulong i=0UL; i<vm->instr_ctx->instr->acct_cnt; i++ ) {
+      /* Anza only even attemps to update the account permissions if it is a
+         "caller account". Only writable accounts are caller accounts. */
+      if( fd_instr_acc_is_writable_idx( vm->instr_ctx->instr, i ) ) {
+
```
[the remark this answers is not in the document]
ibhatt-jumptrading: https://github.com/firedancer-io/solana/blob/508f325e19c0fd8e16683ea047d7c1a85f127e74/programs/bpf_loader/src/syscalls/cpi.rs#L939-L943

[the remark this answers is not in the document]
ibhatt-jumptrading: I'll add it inline

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
ibhatt-jumptrading: adding permalink. basically we only need to operate on writable accounts because non-writable accounts can't be updated

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
ibhatt-jumptrading: Removing a bunch of prints i forgot to remove

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
```diff
@@ -315,54 +385,177 @@ TODO: error codes
...
+    if( realloc_bytes_used ) {
+      /* We intentionally do a load in the case where we are writing to because
+         we want to ignore the write checks. We load from the first byte of the
+         resizing region */
+      ulong resizing_idx = vm->acc_region_metas[ instr_acc_idx ].region_idx;
+      if( vm->acc_region_metas[ instr_acc_idx ].has_data_region ) {
+        resizing_idx++;
+      }
+      uchar * to_slice   = (uchar*)vm->input_mem_regions[ resizing_idx ].haddr;
+      uchar * to_slice_2 = (uchar*)FD_VM_MEM_HADDR_LD( vm, caller_acc_metadata_vm_addr+original_len, alignof(uchar), sizeof(uchar) );
```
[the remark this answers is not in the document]
ibhatt-jumptrading: removed already 🙂

src/flamenco/vm/syscall/fd_vm_syscall_cpi.c
```diff
@@ -417,6 +417,10 @@ TODO: check_align is set wrong in the runtime, ensure that it is set correctly:
   ulong FD_EXPAND_THEN_CONCAT2(decl, _vm_addr) = acc_info->data_addr; \
   ulong FD_EXPAND_THEN_CONCAT2(decl, _len) = acc_info->data_sz;
 
+#define VM_SYSCALL_CPI_ACC_INFO_METADATA( vm, acc_info, decl ) \
+  ulong FD_EXPAND_THEN_CONCAT2(decl, _vm_addr) = acc_info->data_addr; \
+  ulong FD_EXPAND_THEN_CONCAT2(decl, _len) = acc_info->data_sz;
```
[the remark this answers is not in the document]
ibhatt-jumptrading: I don't want to combine a potentially illegal store with a metadata access. That's why I included a separate macro. Another approach could be to split the data load and metadata load into 2 macros and get rid of VM_SYSCALL_CPI_ACC_INFO_DATA.

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
```diff
@@ -315,54 +382,170 @@ TODO: error codes
...
+    ulong post_len = callee_acc_rec->const_meta->dlen;
+
+    /* Do additional handling in the case where the data size has changed in
+       the course of the callee's CPI. */
+    if( prev_len!=post_len ) {
+      /* There is an illegal data overflow if the post len is greater than the
+         original data len + the max resizing limit (10KiB). Can't resize the
+         account if the deprecated loader is being used */
+      ulong max_increase = vm->is_deprecated ? 0UL : 10240UL;
+      if( FD_UNLIKELY( post_len>fd_ulong_sat_add( original_len, max_increase ) ) ) {
```
topointon-jump, line 496: We probably want to cast `original_len` to a ulong here

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
```diff
@@ -315,54 +382,170 @@ TODO: error codes
...
+         space in the account resizing region. */
+      if( post_len<prev_len && prev_len>original_len ) {
+        ulong dirty_realloc_start = fd_ulong_max( post_len, original_len );
+        ulong dirty_realloc_len   = fd_ulong_sat_sub( prev_len, dirty_realloc_start );
+        /* We don't have to worry about multiple region writes here since we
+           know the amount to zero out is located in the account's data
+           resizing region. We intentionally write to the pointer despite
+           loading it in because we assume that the permissions were changed
+           in the callee. */
+        uchar * dirty_region = (uchar*)FD_VM_MEM_HADDR_LD( vm, caller_acc_metadata_vm_addr + dirty_realloc_start,
```
topointon-jump, line 511: This should be `FD_VM_MEM_HADDR_ST`

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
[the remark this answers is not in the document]
topointon-jump: awesome thank-you!

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
ibhatt-jumptrading: Agave converts it to a non-writable region then converts it back. It doesn't make sense to do this if we can just do a load to get the raw pointer

topointon-jump: I don't think they do - they convert it to a _writable_ region to zero it out, then set it back to the same permissions it had before: https://github.com/anza-xyz/agave/blob/36323b6dcd3e29e4d6fe6d73d716a3f33927148b/programs/bpf_loader/src/syscalls/cpi.rs#L1399-L1400

In our case, we are performing a write through a pointer obtained from `FD_VM_MEM_HADDR_LD`, which is an API violation. We should use `FD_VM_MEM_HADDR_ST` to obtain the pointer, as we are performing a write through it.

I'm not suggesting we change the regions metadata like they do, just that we use our memory translation APIs correctly.

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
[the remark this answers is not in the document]
topointon-jump: I think this should be 1-byte aligned, as when we serialize into it we write at arbitrary offsets.

src/flamenco/vm/fd_vm_private.h
topointon-jump: Should we not throw an error or return a sentinel value if the offset is invalid?

src/flamenco/vm/fd_vm_private.h
```diff
@@ -184,16 +196,87 @@ fd_vm_mem_cfg( fd_vm_t * vm ) {
...
+fd_vm_find_input_mem_region( fd_vm_t const * vm, 
+                             ulong           offset,
+                             ulong           sz, 
+                             uchar           write,
+                             ulong           sentinel,
+                             uchar *         is_multi_region ) {
+
+  /* Binary search to find the correct memory region.  If direct mapping is not
+     enabled, then there is only 1 memory region which spans the input region. */
+  ulong region_idx = fd_vm_get_input_mem_region_idx( vm, offset );
```
topointon-jump, line 240: Should we return the sentinel value here if the region is not found?

src/flamenco/vm/fd_vm_private.h
```diff
@@ -184,16 +196,87 @@ fd_vm_mem_cfg( fd_vm_t * vm ) {
...
+                                              (uint)fd_ulong_sat_sub( offset, vm->input_mem_regions[ region_idx ].vaddr_offset ) );
+
+  if( FD_UNLIKELY( write && vm->input_mem_regions[ region_idx ].is_writable==0U ) ) {
+    return sentinel; /* Illegal write */
+  }
+
+  *is_multi_region = 0;
+  while( FD_UNLIKELY( bytes_left>bytes_in_cur_region ) ) {
+    *is_multi_region = 1;
+    FD_LOG_DEBUG(( "Size of access spans multiple memory regions" ));
```
topointon-jump, line 253: nit: remove?

src/flamenco/vm/fd_vm_private.h
```diff
@@ -260,35 +458,40 @@ static inline void fd_vm_mem_st_8( ulong haddr, ulong  val ) { memcpy( (void *)h
...
-#define FD_VM_MEM_HADDR_LD( vm, vaddr, align, sz ) (__extension__({                                       \
-    fd_vm_t const * _vm     = (vm);                                                                       \
-    ulong           _vaddr  = (vaddr);                                                                    \
-    int             _sigbus = _vm->check_align & (!fd_ulong_is_aligned( _vaddr, (align) ));               \
-    ulong           _haddr  = fd_vm_mem_haddr( vm, _vaddr, (sz), _vm->region_haddr, _vm->region_ld_sz, 0, 0UL ); \
-    if( FD_UNLIKELY( (!_haddr) | _sigbus) ) return FD_VM_ERR_SIGSEGV;                                     \
-    (void const *)_haddr;                                                                                 \
+   syscall has already done preflight checks on them).
+   
+   FIXME: These macros currently don't support multi region loads/stores */
```
topointon-jump, line 463: Do we need to change these to support multi-region loads/stores before merging? These are used everywhere inside the syscalls, and a lot of the size values are user-provided, so it's not impossible they could span multiple regions.

src/flamenco/vm/fd_vm_private.h
```diff
@@ -385,6 +588,21 @@ fd_vm_log_append( fd_vm_t *    vm,
   return vm;
 }
 
+/* fd_vm_get_acc_data_region_idx returns an index into the list of input
+   memory regions. If an account data region does not exist, it returns
+   the index of its resizing region. If a pubkey can't be found it will
+   return 0. Although there is a valid memory region with index 0, it 
+   can never correspond to a memory region for an account. */
+static inline ulong
+fd_vm_get_acc_data_region_idx( fd_vm_t * vm, fd_pubkey_t const * pubkey ) {
```
topointon-jump, line 597: It looks like this is unused

src/flamenco/vm/fd_vm_interp.c
```diff
@@ -68,6 +70,8 @@ fd_vm_exec_trace( fd_vm_t * vm ) {
 
   fd_vm_shadow_t * FD_RESTRICT shadow = vm->shadow;
 
+  uchar is_multi_region = 0;
```
topointon-jump, line 73: Could we not make this a global variable, but instead scope it to each instruction? Since it's only set-and-read inside a single instruction at a time. Having it here is confusing, because it makes it seem like it's part of the vm's state, whereas it's actually part of the instructions state. Minimising the scope of variables makes code easier to understand and improves performance.

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
```diff
@@ -125,21 +125,21 @@ callee account we look up the account in the borrowed accounts cache and update
...
-
+                                      uchar                             instr_acc_idx,
+                                      fd_pubkey_t const *               callee_acc_pubkey ) {
   /* Consume compute units for the account data access */
 
   /* FIXME: do we also need to consume the compute units if the account is not known? */
-  VM_SYSCALL_CPI_ACC_INFO_DATA( vm, account_info, caller_acc_data );
-  (void)caller_acc_data_vm_addr;
+
+  VM_SYSCALL_CPI_ACC_INFO_METADATA( vm, account_info, caller_acc_metadata );
```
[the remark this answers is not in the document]
ibhatt-jumptrading: I disagree. Calling it caller_acc_data to me implies that it is the account's data

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
```diff
@@ -150,42 +150,109 @@ VM_SYCALL_CPI_UPDATE_CALLEE_ACC_FUNC( fd_vm_t * vm,
...
+    ulong region_idx = vm->acc_region_metas[ instr_acc_idx ].region_idx;
+    uint original_len = vm->acc_region_metas[ instr_acc_idx ].has_data_region ? 
+                        vm->input_mem_regions[ region_idx ].region_sz : 0U;
+    ulong prev_len           = callee_acc->const_meta->dlen;
+    ulong post_len           = caller_acc_metadata_len;
+
+    int err;
+    if( fd_account_can_data_be_resized( vm->instr_ctx, callee_acc->meta, post_len, &err ) &&
+        fd_account_can_data_be_changed( vm->instr_ctx->instr, instr_acc_idx, &err ) ) {
+
```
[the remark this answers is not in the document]
ibhatt-jumptrading: Yes we do on line 246

src/flamenco/vm/fd_vm_private.h
ibhatt-jumptrading: It get's handled later in the function

src/flamenco/vm/fd_vm_private.h
ibhatt-jumptrading: I think this is fine

src/flamenco/vm/fd_vm_interp.c
ibhatt-jumptrading: sure

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
topointon-jump: This still feels off to me, we are mis-using the API. I wonder if we should create a new `_ST_UNCHECKED` macro or something to make the semantics clear. Just worried someone will make this stricter in the future and that will break things.

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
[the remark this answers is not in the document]
topointon-jump: The generated variables: `caller_acc_metadata_vm_addr` and `caller_acc_metadata_len` read like they refer to the addr and size of the metadata, not the data.

Changing it to `caller_acc_data` would generate `caller_acc_data_vm_addr` and `caller_acc_data_len` which is more correct.

src/flamenco/vm/fd_vm_private.h
ibhatt-jumptrading: we don't need to. made a small change to macros to support this

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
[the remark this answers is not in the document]
topointon-jump: If we enter the loop then we ignore the value of err (and overwrite it later). If err is non-zero that invalidates the fd_account_ calls, so we can't be sure that we're entering the loop correctly. We should check the error before potentially entering the loop.

src/flamenco/vm/fd_vm_private.h
topointon-jump: Do you mind adding a comment explaining how? The offset can potentially be any value as it's user-provided.

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
```diff
@@ -315,54 +382,174 @@ TODO: error codes
...
+    ulong realloc_bytes_used = fd_ulong_sat_sub( post_len, original_len );
+    if( realloc_bytes_used ) {
+      /* We intentionally do a load in the case where we are writing to because
+         we want to ignore the write checks. We load from the first byte of the
+         resizing region */
+      ulong resizing_idx = vm->acc_region_metas[ instr_acc_idx ].region_idx;
+      if( vm->acc_region_metas[ instr_acc_idx ].has_data_region ) {
+        resizing_idx++;
+      }
+      uchar * to_slice   = (uchar*)vm->input_mem_regions[ resizing_idx ].haddr;
```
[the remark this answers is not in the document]
ibhatt-jumptrading: post_len can't be greater than original len as an invariant of the deprecated loader. I'll add a check for safety

src/flamenco/vm/fd_vm_private.h
ibhatt-jumptrading: I documented that the caller is responsible for handling it. It does so in the relevant syscalls where it gets called and in fd_vm_find_input_mem_region

src/flamenco/vm/fd_vm_private.h
ibhatt-jumptrading: I have a comment, but I'll also add a note at the top of the function
```
    if( FD_UNLIKELY( region_idx==vm->input_mem_regions_cnt ) ) {
      return sentinel; /* Access is too large */
    }
```

src/flamenco/vm/fd_vm_private.h
topointon-jump: Don't we need to error out if `_is_multi` is true? Also can we document why we don't need to support multi-region loads in the syscalls (because Agave's memory translation logic doesn't).

src/flamenco/vm/fd_vm_interp.c
topointon-jump: thank-you!

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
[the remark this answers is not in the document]
ibhatt-jumptrading: I think it's okay, but I don't feel strongly so I'll change it

src/flamenco/vm/fd_vm_private.h
ibhatt-jumptrading: I did

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
ibhatt-jumptrading: adding a st_write_unchecked

PR #2323, opened 2024-07-05
fix(choreo, flamenco): use const types for voter pubkey

yhzhangjump:


src/app/fdctl/run/tiles/fd_replay.c
```diff
@@ -206,8 +206,8 @@ struct fd_replay_tile_ctx {
   uint poh_init_done;
 
   int         vote;
-  fd_pubkey_t validator_identity_pubkey[ 1 ];
-  fd_pubkey_t vote_acct_addr[ 1 ];
+  fd_pubkey_t const validator_identity_pubkey[ 1 ];
+  fd_pubkey_t const vote_acct_addr[ 1 ];
```
lheeger-jump, line 210: i am not super crazy about making the members const here. my reasoning being that we will want to make them non-const for hotswapping the keys. but for now... this is fine.

```suggestion
  // TODO: make these keys safely hotswappable, and therefore not const.
  fd_pubkey_t const validator_identity_pubkey[ 1 ];
  fd_pubkey_t const vote_acct_addr[ 1 ];
```

lheeger-jump (changes requested):

lidatong (changes requested):
dont make these const or you cant assign on init

src/app/fdctl/run/tiles/fd_replay.c
yhzhangjump: Thanks and I have changed them back to non-const.

lidatong (approved):

lheeger-jump (approved):

PR #2359, opened 2024-07-09
tls: provide safe default for CSPRNG

ripatel-fd:
- Fixes an unlikely crash in fd_quic when CSPRNG is starved
- Simplifies fd_tls public API by providing a safe default if the
  user doesn't provide an rng callback

### DIFF
--- a/src/waltz/tls/fd_tls.h
+++ b/src/waltz/tls/fd_tls.h
@@ -135,11 +135,7 @@ typedef void
    reasonable amount of data ahead of time.  NULL return value implies
    inability to keep up with demand for random values.  In this case,
    function should return NULL.  Function should minimize side effects
-   (notably, should not log).
-
-   TODO API considerations:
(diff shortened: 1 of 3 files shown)


0x0ece (approved):

src/waltz/tls/fd_tls.h
```diff
@@ -230,7 +219,7 @@ fd_tls_sign( fd_tls_sign_t const * sign,
    across multiple TLS handshakes. */
 
 struct fd_tls {
-  fd_tls_rand_t       rand;
+  fd_tls_rand_t       rand;  /* may be NULL */
```
ripatel-fd, line 222: @0x0ece recommended against nullable function pointers at runtime here. I still have to think about the most robust fix. We should probably have an `fd_tls_init` that fills out defaults and errors if any mandatory field is left uninitialized.

PR #2363, opened 2024-07-10
Various fixes for basic v20 support

jumpsiegel:
@yhzhangjump and I added basic v2.0 support for both snapshots and genesis


jumpsiegel:
The cluster version is a 4 digit number (1.18.5) => 1185.  This tells us the version of solana used to create the genesis block.  If you are booting from a snapshot, the version is not required (at present)

ripatel-fd (approved):

ripatel-fd (approved):

PR #2369, opened 2024-07-10
ci: cache solana submodule

anwayde:


ripatel-fd:
Uploading from Equinix to the remote GitHub cache is probably slower than just building from scratch. I suggest using ccache instead.

lheeger-jump (approved):

PR #2388, opened 2024-07-11
flamenco: fix segfault when dumping instructions

mjain-jump:
Seed address field for accounts was not getting set correctly


kbhargava-jump (approved):
ty for fixing

PR #2402, opened 2024-07-12
status_cache: make workspace compatible

arjain4:


lheeger-jump:
Any chance you can get the 23 GB on loose to be part of footprint?

lidatong (approved):

PR #2415, opened 2024-07-15
QUIC - various fixes

nbridge-jump:
Account for ethernet frames with padding
Disable PING keep alive
Encryption of short packet number. Not used, but helps testing of decryption of short packet numbers


ripatel-fd (changes requested):
.

ripatel-fd (approved):

PR #2416, opened 2024-07-15
Second attempt at cleaning up epoch_bank decode

asiegel-jt:


jumpsiegel (approved):

arjain4 (commented):
I don't see fd_types.json getting modified but fd_types.c is modified. If those changes are made by hand it might cause issues when someone tries to re-gen the file

src/flamenco/types/fd_types_custom.c
```diff
@@ -172,17 +172,15 @@ void fd_solana_vote_account_decode_unsafe( fd_solana_vote_account_t * self, fd_b
   ulong data_len;
   fd_bincode_uint64_decode_unsafe( &data_len, ctx );
   if( data_len ) {
-    uchar * data = fd_valloc_malloc( ctx->valloc, 8UL, data_len );
```
arjain4, line 175: This change was made because we don't want to keep a variable sized solana account in here. Does this change that?

arjain4 (changes requested):
Both commits are the same, squash

src/flamenco/types/fd_types_custom.c
arjain4: I think we need to do a 
```
uchar * data = fd_scratch_alloc(...);
```

src/flamenco/types/fd_types_custom.c
```diff
@@ -172,17 +172,15 @@ void fd_solana_vote_account_decode_unsafe( fd_solana_vote_account_t * self, fd_b
...
-
     FD_SCRATCH_SCOPE_BEGIN {
       /* Deserialize content */
       fd_vote_block_timestamp_t vote_ts;
       fd_vote_state_versioned_t vs[1];
       fd_bincode_decode_ctx_t decode =
-          { .data    = data,
-            .dataend = data + data_len,
-            .valloc  = fd_scratch_virtual() };
+        { .data    = (uchar *)ctx->data,
```
arjain4, line 180: We cannot decode into the ctx here, this method is designed to save only the fields we need and keeping around the entire variable length data buffer

src/flamenco/types/fd_types_custom.c
asiegel-jt: this was just copying the memory for no purpose. it works just fine without

arjain4 (approved):

PR #2424, opened 2024-07-15
disco: use offsets in txncache

arjain4:


src/disco/bank/fd_txncache.c
```diff
@@ -324,61 +332,17 @@ fd_txncache_join( void * shtc ) {
...
-  }
-
-  fd_txncache_t * tc = (fd_txncache_t *)shtc;
-
-  if( FD_UNLIKELY( tc->magic!=FD_TXNCACHE_MAGIC ) ) {
-    FD_LOG_WARNING(( "bad magic" ));
-    return NULL;
-  }
+  uchar * base = (uchar *)tc;
+  tc->root_slots = (ulong *)( base + tc->root_slots_off );
```
mmcgee-jump, line 336: ?? This isn't right, the different joins share this memory still. You need to remove these pointers from the struct and just reference everything from the offset always (probably with helper fns, see eg, the pattern used by fd_cnc_app_laddr

https://github.com/firedancer-io/firedancer/blob/main/src/tango/cnc/fd_cnc.h#L211

mmcgee-jump (changes requested):
.

src/disco/bank/fd_txncache.c
arjain4: this change is made, testing against mainnet snapshot

lheeger-jump (approved):
I think this looks good but @mmcgee-jump might have more feedback

PR #2428, opened 2024-07-16
fuzz: various fix for the syscall fuzzer harness

anarcheuz:
Fix protobuf following protosol change.

Fix various bugs in the syscall fuzzer harness.

### DIFF
--- a/src/flamenco/runtime/tests/fd_exec_instr_test.c
+++ b/src/flamenco/runtime/tests/fd_exec_instr_test.c
@@ -1,3 +1,4 @@
+
 #include "generated/invoke.pb.h"
 #undef FD_SCRATCH_USE_HANDHOLDING
 #define FD_SCRATCH_USE_HANDHOLDING 1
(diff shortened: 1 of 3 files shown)


src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -1192,4 +1228,8 @@ fd_exec_vm_syscall_test_run( fd_exec_instr_test_runner_t *          runner,
 
   *output = effects;
   return actual_end - (ulong)output_buf;
+
+error:
+  _context_destroy( runner, ctx );
+  return 0;
```
topointon-jump, line 1234: Might be worth using a cleanup attribute here: https://github.com/firedancer-io/firedancer/blob/main/CONTRIBUTING.md#83-complex-function-exit

src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -773,6 +774,25 @@ _diff_effects( fd_exec_instr_fixture_diff_t * check ) {
            violation) */
 }
 
+fd_sbpf_syscalls_t *
+lookup_syscall_func( fd_sbpf_syscalls_t *syscalls,
```
topointon-jump, line 778: Can we make this static?

topointon-jump (commented):
Awesome! Can you squash into a single commit with a scoped title? Then I'll approve

src/flamenco/runtime/tests/fd_exec_instr_test.c
anarcheuz: yes

src/flamenco/runtime/tests/fd_exec_instr_test.c
anarcheuz: As far as I know the cleanup attribute can only receive 1 argument. Not sure if it would be clearer if I have to create a new type to encapsulate `runner` and `ctx`. What do you think @topointon-jump ?

topointon-jump: I'm not against creating a new type, but it's up to you. The only reason I suggest it is because it may help with memory leaks whilst fuzzing.

topointon-jump (approved):

PR #2451, opened 2024-07-17
fdctl: don't load config to run --help or --version

mmcgee-jump:


src/app/fdctl/main1.c
```diff
@@ -231,6 +231,20 @@ main1( int     argc,
   char ** argv = _argv;
   argc--; argv++;
 
+  /* Short circuit evaluating help and version commands so that we don't
+     need to load and evaluate the entire config file to run them.
+     This is useful for some operators in CI environments where, for
+     example, they want to show the version or validate the produced
+     binary without yet setting up the full TOML. */
+
+  if( FD_UNLIKELY( argc==2 && (!strcmp( argv[ 1 ], "help" ) || !strcmp( argv[ 1 ], "--help" )) ) ) {
```
ptaffet-jump, line 240: I think the indices are off by one because of the `argc--; argv++` line above

ripatel-fd: I usually add this to the first line of main:

`for( int i=1; i<argc; i++ ) if( 0==strcmp( argv[i], "--help" ) ) return usage();`

ptaffet-jump (approved):

Issue #1438, opened 2024-04-02
Handle precompiles properly


ptaffet-jump:
Ed25519SigVerify111111111111111111111111111  and KeccakSecp256k11111111111111111111111111111 are precompiles, which means they are handled differently from other instructions. Make sure we handle them properly.

0x0ece:
TODO: add call to verify_precompiles

PR #2453, opened 2024-07-17
bank: verify precompiles for transactions

ptaffet-jump:
Fixes [#1438](https://github.com/firedancer-io/firedancer/issues/1438)
Though pack's cost estimation should be updated to work with them


src/disco/metrics/metrics.xml
```diff
@@ -213,6 +213,8 @@ metric introduced.
     <counter name="TransactionSanitizeFailure" summary="Number of transactions that failed to sanitize." />
     <counter name="TransactionNotExecutedFailure" summary="Number of transactions that did not execute. This is different than transactions which fail to execute, which make it onto the chain." />
 
+    <counter name="PrecompileVerifyFailed" summary="Number of transactions that failed precompile verification and thus will not execute." />
```
mmcgee-jump, line 216: nit: maybe keep naming consistent with the above ones, `PrecompileVerifyFailure`

src/app/fdctl/run/tiles/fd_bank.c
```diff
@@ -182,6 +183,12 @@ after_frag( void *             _ctx,
     ctx->metrics.txn_load_address_lookup_tables[ result ]++;
     if( FD_UNLIKELY( result!=FD_BANK_ABI_TXN_INIT_SUCCESS ) ) continue;
 
+    int precompile_result = fd_ext_bank_verify_precompiles( ctx->_bank, abi_txn );
+    if( FD_UNLIKELY( precompile_result ) ) {
+      FD_MCNT_INC( BANK_TILE, PRECOMPILE_VERIFY_FAILED, 1 );
```
mmcgee-jump, line 188: nit: could be a lot of cache traffic if someone spams txns that fail precompiles, they also don't pay right

mmcgee-jump (dismissed):

src/app/fdctl/run/tiles/fd_bank.c
ptaffet-jump: I think each precompile takes tens of microseconds of computation, so the cache traffic will be in the noise

src/disco/metrics/metrics.xml
ptaffet-jump: Good call. I renamed it.

mmcgee-jump (approved):

PR #2458, opened 2024-07-18
Backport: Fix QUIC RX stream size limits

ripatel-fd:
Backport of #2457


mmcgee-jump:
For now just make fixes in main, and we will review for cherry picking regularly (this will change after v0.1)

ripatel-fd:
@mmcgee-jump This is cherry-picked, but PRs would allow reviewing what exactly is cherry-picked

PR #2461, opened 2024-07-18
fuzz, flamenco: migrate sol compat targets to wksp allocators + add wksp leak detection

mjain-jump:
For all targets...
* Migrate from libc to wksp allocators for fuzzing / CI tests
* Add leak detection for fuzzing / CI tests (now aborts on a wksp leak)

Tested so far in...
- CI 
- Instruction fuzzing

### DIFF
--- a/src/flamenco/runtime/tests/fd_exec_instr_test.c
+++ b/src/flamenco/runtime/tests/fd_exec_instr_test.c
@@ -51,9 +51,6 @@ static FD_TL char _report_prefix[100] = {0};
 #include "../../../util/tmpl/fd_sort.c"
 #include "../../vm/fd_vm_base.h"
 
-static uchar * data_wksp_ptrs[256] = {0};
-static ulong data_wksp_ptrs_idx = 0;
-
 struct __attribute__((aligned(32UL))) fd_exec_instr_test_runner_private {
   fd_funk_t * funk;
 };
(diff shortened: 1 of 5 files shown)


src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -159,10 +154,8 @@ static int
 _context_create( fd_exec_instr_test_runner_t *        runner,
                  fd_exec_instr_ctx_t *                ctx,
                  fd_exec_test_instr_context_t const * test_ctx,
+                 fd_alloc_t *                         alloc,
```
topointon-jump, line 157: We should probably pass a `fd_valloc_t` so we can easily swap this out in future.

src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -190,15 +183,15 @@ _context_create( fd_exec_instr_test_runner_t *        runner,
...
-  fd_exec_slot_ctx_t *  slot_ctx      = fd_exec_slot_ctx_join ( fd_exec_slot_ctx_new ( slot_ctx_mem, fd_libc_alloc_virtual() ) );
+  fd_exec_slot_ctx_t *  slot_ctx      = fd_exec_slot_ctx_join ( fd_exec_slot_ctx_new ( slot_ctx_mem, fd_alloc_virtual( alloc ) ) );
   fd_exec_txn_ctx_t *   txn_ctx       = fd_exec_txn_ctx_join  ( fd_exec_txn_ctx_new  ( txn_ctx_mem   ) );
 
   assert( epoch_ctx );
   assert( slot_ctx  );
 
   ctx->slot_ctx   = slot_ctx;
   ctx->txn_ctx    = txn_ctx;
-  txn_ctx->valloc = fd_libc_alloc_virtual();
```
topointon-jump, line 201: It might be a good idea to keep support for both allocators, in case we want to temporarily switch back at some point in the future.

topointon-jump (approved):

PR #2474, opened 2024-07-19
sandbox: use privileged userns on ubuntu 24.04

mmcgee-jump:


src/util/sandbox/fd_sandbox.c
```diff
@@ -40,18 +40,29 @@
...
-  (void)arg;
+check_unshare_eacces_main( void * _arg ) {
+  ulong arg = (ulong)_arg;
+  uint desired_uid = (uint)((arg >>  0UL) & 0xFFFFUL);
+  uint desired_gid = (uint)((arg >> 32UL) & 0xFFFFUL);
 
+  fd_sandbox_private_switch_uid_gid( desired_uid, desired_gid );
   int result = unshare( CLONE_NEWUSER );
-  if( -1==result && errno==EACCES ) return 999;
+  if( -1==result && errno==EACCES ) return 255;
```
ptaffet-jump, line 55: Why not just return and check for EACCES instead of an arbitrary number? That at least communicates the intent.

ptaffet-jump (approved):

PR #2477, opened 2024-07-19
dedup: use randomized hash

### LINKED ISSUE #990: Verify security of deduplication strategy in dedup/verify tiles
These tiles use the first 8 bytes of the hash as a key for checking for duplicates. If a user can generate collisions, they could censor another users transactions, and if they can generate the 0 (NULL) value it might corrupt the tcache.

Hash with some random seed private to the validator should solve this.

#987

0x0ece:
Fixes #990

### DIFF
--- a/src/app/fdctl/run/tiles/fd_dedup.c
+++ b/src/app/fdctl/run/tiles/fd_dedup.c
@@ -44,6 +44,8 @@ typedef struct {
   ulong       out_chunk0;
   ulong       out_wmark;
   ulong       out_chunk;
+
+  ulong       hashmap_seed;
 } fd_dedup_ctx_t;
 
 FD_FN_CONST static inline ulong
@@ -130,6 +132,17 @@ after_frag( void *             _ctx,
 
   fd_dedup_ctx_t * ctx = (fd_dedup_ctx_t *)_ctx;
 
+  ulong payload_sz = *opt_sz;
+  ulong txn_off = fd_ulong_align_up( payload_sz, 2UL );
+
+  /* Ensure sufficient trailing space for parsing. */
+  if( FD_UNLIKELY( txn_off>FD_TPU_DCACHE_MTU - FD_TXN_MAX_SZ - sizeof(ushort)) ) {
+    FD_LOG_ERR(( "got malformed txn (sz %lu) insufficient space left in dcache for fd_txn_t", payload_sz ));
+  }
+
+  uchar    * dcache_entry = fd_chunk_to_laddr( ctx->out_mem, ctx->out_chunk );
+  fd_txn_t * txn = (fd_txn_t *)(dcache_entry + txn_off);
+
   if ( FD_UNLIKELY( in_idx < ctx->unparsed_in_cnt ) ) {
     /* Transactions coming in from these links are not parsed.
 
@@ -140,16 +153,6 @@ after_frag( void *             _ctx,
     /* Increment on GOSSIP_IN_IDX but not VOTER_IN_IDX */
     FD_MCNT_INC( DEDUP, GOSSIPED_VOTES_RECEIVED, 1UL - in_idx );
 
-    ulong payload_sz = *opt_sz;
-    ulong txn_off = fd_ulong_align_up( payload_sz, 2UL );
-
-    /* Ensure sufficient trailing space for parsing. */
-    if( FD_UNLIKELY( txn_off>FD_TPU_DCACHE_MTU - FD_TXN_MAX_SZ - sizeof(ushort)) ) {
-      FD_LOG_ERR(( "got malformed txn (sz %lu) insufficient space left in dcache for fd_txn_t", payload_sz ));
-    }
-
-    uchar    * dcache_entry = fd_chunk_to_laddr( ctx->out_mem, ctx->out_chunk );
-    fd_txn_t * txn = (fd_txn_t *)(dcache_entry + txn_off);
     ulong txn_t_sz = fd_txn_parse( dcache_entry, payload_sz, txn, NULL );
     if( FD_UNLIKELY( !txn_t_sz ) ) {
       FD_LOG_ERR(( "fd_txn_parse failed for vote transactions that should have been sigverified" ));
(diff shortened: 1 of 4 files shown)


src/app/fdctl/run/topos/fd_firedancer.c
```diff
@@ -405,6 +405,7 @@ fd_topo_firedancer( config_t * _config ) {
 
     } else if( FD_UNLIKELY( !strcmp( tile->name, "dedup" ) ) ) {
       tile->dedup.tcache_depth = config->tiles.dedup.signature_cache_size;
+      FD_TEST( fd_rng_secure( &tile->dedup.hashmap_seed, 8U ) );
```
mmcgee-jump, line 408: Do this in privileged_init of the tile, not here

src/disco/topo/fd_topo.h
```diff
@@ -167,6 +167,7 @@ typedef struct {
 
     struct {
       ulong tcache_depth;
+      ulong hashmap_seed;
```
mmcgee-jump, line 170: Not here, just in the ctx with privileged_init

src/app/fdctl/run/tiles/fd_verify.c
```diff
@@ -199,6 +199,9 @@ unprivileged_init( fd_topo_t *      topo,
   ctx->out_wmark  = fd_dcache_compact_wmark ( ctx->out_mem, topo->links[ tile->out_link_id_primary ].dcache, topo->links[ tile->out_link_id_primary ].mtu );
   ctx->out_chunk  = ctx->out_chunk0;
 
+  /* All verify tiles use the same hashmap_seed as dedup */
+  ctx->hashmap_seed = tile->dedup.hashmap_seed;
```
mmcgee-jump, line 203: I don't think this is valid (dedup state is not present for verify tile), but see below: just initialize it in the tile itself

mmcgee-jump (changes requested):
.

mmcgee-jump:
Can you also just discard the txn if the sig comes out to 0? I think we dont want to insert the NULL value to the map for safety

src/app/fdctl/run/topos/fd_firedancer.c
0x0ece: fixed

src/disco/topo/fd_topo.h
0x0ece: fixed

src/app/fdctl/run/tiles/fd_verify.c
0x0ece: fixed

0x0ece:
> Can you also just discard the txn if the sig comes out to 0? I think we dont want to insert the NULL value to the map for safety

tcache is safe to use. `FD_TCACHE_INSERT` does a query, and if you query 0 (NULL) it returns true == dedup.

src/app/fdctl/run/tiles/fd_dedup.c
```diff
@@ -130,6 +132,17 @@ after_frag( void *             _ctx,
...
+  ulong payload_sz = *opt_sz;
+  ulong txn_off = fd_ulong_align_up( payload_sz, 2UL );
+
+  /* Ensure sufficient trailing space for parsing. */
+  if( FD_UNLIKELY( txn_off>FD_TPU_DCACHE_MTU - FD_TXN_MAX_SZ - sizeof(ushort)) ) {
+    FD_LOG_ERR(( "got malformed txn (sz %lu) insufficient space left in dcache for fd_txn_t", payload_sz ));
+  }
+
+  uchar    * dcache_entry = fd_chunk_to_laddr( ctx->out_mem, ctx->out_chunk );
+  fd_txn_t * txn = (fd_txn_t *)(dcache_entry + txn_off);
```
0x0ece, line 144: In both paths we expect txn at the same memory address. From verify, tx is already parsed and txn filled, from gossip it's not.

src/app/fdctl/run/tiles/fd_dedup.c
```diff
@@ -172,22 +175,13 @@ after_frag( void *             _ctx,
                     payload_sz, txn_t_sz ));
     }
 
-    /* Assert that the payload_sz includes all signatures.
-       We don't need them all for dedup. */
-    ushort sig_off = txn->signature_off;
-    ulong sig_end_off = sig_off + FD_TXN_SIGNATURE_SZ * txn->signature_cnt;
```
0x0ece, line 178: This check is redundant because we just did it in fd_txn_parse.

src/app/fdctl/run/tiles/fd_dedup.c
```diff
@@ -172,22 +175,13 @@ after_frag( void *             _ctx,
...
-    /* Write signature for dedup. */
-    ulong txn_sig = FD_VERIFY_DEDUP_TAG_FROM_PAYLOAD_SIG( dcache_entry + sig_off );
-    *opt_sig = txn_sig;
-
     /* Write new size for mcache. */
     *opt_sz = new_sz;
   }
 
+  ushort sig_off = txn->signature_off;
+  *opt_sig = fd_hashmap_hash( dcache_entry + sig_off, ctx->hashmap_seed );
```
0x0ece, line 183: Is it useful to write `*opt_sig` or should we just have a temp var?

src/app/fdctl/run/tiles/fd_verify.c
```diff
@@ -135,22 +137,12 @@ after_frag( void *             _ctx,
                   payload_sz, txn_t_sz ));
   }
 
-  /* We need to access signatures and accounts, which are all before the recent_blockhash_off.
-     We assert that the payload_sz includes all signatures and account pubkeys we need. */
-  ushort recent_blockhash_off = txn_t->recent_blockhash_off;
```
0x0ece, line 140: This check is redundant because we just did it in fd_txn_parse.

src/app/fdctl/run/tiles/fd_verify.c
```diff
@@ -135,22 +137,12 @@ after_frag( void *             _ctx,
...
-  ulong txn_sig;
-  int res = fd_txn_verify( ctx, txn, (ushort)payload_sz, txn_t, &txn_sig );
+  int res = fd_txn_verify( ctx, txn, (ushort)payload_sz, txn_t );
   if( FD_UNLIKELY( res!=FD_TXN_VERIFY_SUCCESS ) ) {
     *opt_filter = 1; /* Signature verification failed. */
     return;
   }
 
-  ulong tspub = (ulong)fd_frag_meta_ts_comp( fd_tickcount() );
-  fd_mux_publish( mux, txn_sig, ctx->out_chunk, new_sz, 0UL, *opt_tsorig, tspub );
```
0x0ece, line 153: Verify no longer needs to publish `txn_sig`. This value is now different for every tile (and different than inside dedup).

src/app/fdctl/run/tiles/fd_verify.h
```diff
@@ -38,16 +38,25 @@ typedef struct {
...
+         ( (fd_ulong_hash( seed ^ (1UL<<4) ^ kul[4] ) ^ fd_ulong_hash( seed ^ (1UL<<5) ^ kul[5] ) ) ^
+           (fd_ulong_hash( seed ^ (1UL<<6) ^ kul[6] ) ^ fd_ulong_hash( seed ^ (1UL<<7) ^ kul[7] ) ) ); /* tons of ILP */
+}
 
 static inline int
 fd_txn_verify( fd_verify_ctx_t * ctx,
                uchar const *     udp_payload,
                ushort const      payload_sz,
-               fd_txn_t const *  txn,
-               ulong *           opt_sig ) {
```
0x0ece, line 50: This opt_sig doesn't need to be propagated anymore, so we can remove it for simplicity.

0x0ece:
New version in #2496

PR #2478, opened 2024-07-19
quic: fix check for oversize transactions

ptaffet-jump:


mmcgee-jump (approved):

src/app/fdctl/run/tiles/fd_quic.c
```diff
@@ -310,7 +310,7 @@ after_frag( void *             _ctx,
       return;
     }
 
-    if( FD_UNLIKELY( *opt_sz>FD_TPU_MTU ) ) {
+    if( FD_UNLIKELY( *opt_sz-network_hdr_sz>FD_TPU_MTU ) ) {
```
mmcgee-jump, line 313: Do you need to check for underflow?

ptaffet-jump: I think the check above it ensures it can't underflow

ripatel-fd (approved):
LGTM. CI failed for unrelated reasons.

PR #2484, opened 2024-07-22
flamenco, vm: fix try_find_program_address iteration count

topointon-jump:
Agave uses [1,255] bump seeds, whereas we use [0,255].

Reported by Ben Hawkes


ibhatt-jumptrading (commented):
I'm pretty sure it was right before. https://github.com/solana-labs/solana/blob/2afde1b028ed4593da5b6c735729d8994c4bfac6/programs/bpf_loader/src/syscalls/mod.rs#L727. As you can see they start with a bump seed of 255 and loop from 0 to 255 inclusive to try the full range from [0,255] inclusive. However, With this change we loop from [0,254] so our bump seed will range from [1,254]

ibhatt-jumptrading (approved):
Turns out ..= is used for inclusive iteration

PR #2485, opened 2024-07-22
Multithreaded global account hash

asiegel-jt:


src/app/ledger/main.c
```diff
@@ -82,7 +82,6 @@ struct fd_ledger_args {
   int                   verify_funk;             /* verify funk before execution starts */
   uint                  verify_acc_hash;         /* verify account hash from the snapshot */
   uint                  check_acc_hash;          /* check account hash by reconstructing with data */
```
jumpsiegel, line 84: In v2.0, we have to reconstruct the account_hash every time and we are no longer given the hash in the snapshot so.. maybe we should just remove the check_acc_hash path?

jumpsiegel (approved):

jumpsiegel (approved):

PR #2488, opened 2024-07-22
deps: add missing agave dependencies

anwayde:


mmcgee-jump (approved):

ripatel-fd (changes requested):
We depend on [pkgconf](https://packages.ubuntu.com/focal/pkgconf), not pkg-config.

ripatel-fd:
@anwayde Can you test this with a Ubuntu 22.04 VM and/or podman containers and let me know if everything is fine? 
Please note that Docker builds work just fine on main without this patch, which seems to behave differently than real hardware.

ripatel-fd (approved):

PR #2496, opened 2024-07-23
dedup: use randomized hash

0x0ece:
Fixes https://github.com/firedancer-io/firedancer/issues/990

### DIFF
--- a/src/app/fdctl/run/tiles/fd_dedup.c
+++ b/src/app/fdctl/run/tiles/fd_dedup.c
@@ -44,6 +44,8 @@ typedef struct {
   ulong       out_chunk0;
   ulong       out_wmark;
   ulong       out_chunk;
+
+  ulong       hashmap_seed;
 } fd_dedup_ctx_t;
 
 FD_FN_CONST static inline ulong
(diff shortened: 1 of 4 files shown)


src/ballet/hashmap/polymur-hash.h
```diff
@@ -0,0 +1,295 @@
+/*
+    PolymurHash version 2.0
```
mmcgee-jump, line 2: Please add a section to `NOTICE` regarding this files source, and its license.

Also please note here the commit hash you took and modified for future reference

src/app/fdctl/run/tiles/fd_verify.h
```diff
@@ -38,9 +39,9 @@ typedef struct {
   ulong       out_chunk0;
   ulong       out_wmark;
   ulong       out_chunk;
-} fd_verify_ctx_t;
 
-#define FD_VERIFY_DEDUP_TAG_FROM_PAYLOAD_SIG(payload_sig_p) FD_LOAD( ulong, (payload_sig_p) )
+  fd_hashmap_t hashmap;
```
mmcgee-jump, line 43: I don't really like calling this `fd_hashmap_t` which implies it's a container for the data (the actual map itself), rather than just a hashing algorithm. Maybe just use `fd_hasher_t` ?

src/app/fdctl/run/tiles/fd_dedup.c
```diff
@@ -172,22 +187,13 @@ after_frag( void *             _ctx,
...
-    /* Write signature for dedup. */
-    ulong txn_sig = FD_VERIFY_DEDUP_TAG_FROM_PAYLOAD_SIG( dcache_entry + sig_off );
-    *opt_sig = txn_sig;
-
     /* Write new size for mcache. */
     *opt_sz = new_sz;
   }
 
+  /* Compute hashmap(signature) for dedup. */
+  *opt_sig = fd_hashmap_hash( &ctx->hashmap, dcache_entry + txn->signature_off, 64UL );
```
mmcgee-jump, line 195: Should this just be `ulong sighash = fd_hashmap_hash` ? Seems unclear to use `*opt_sig` as a temporary variable here

mmcgee-jump (changes requested):
.

src/app/fdctl/run/tiles/fd_verify.h
0x0ece: it's the state of the hash function, exactly like fd_sha?_t

src/app/fdctl/run/tiles/fd_dedup.c
0x0ece: it's not temporary, it's used below... changed the code to make it more clear

0x0ece: fixed

src/ballet/hashmap/polymur-hash.h
0x0ece: Actually removed PolymurHash.

mmcgee-jump (approved):

PR #2498, opened 2024-07-23
submodule: use agave submodule

anwayde:


mmcgee-jump:
could be good to do it here: can you update all the "Solana Labs" strings in like, the TOML file and book etc to say Agave as well?

mmcgee-jump (changes requested):
.

src/disco/topo/fd_topo.h
```diff
@@ -90,7 +90,7 @@ typedef struct {
   ulong id;                     /* The ID of this tile.  Indexed from [0, tile_cnt).  When placed in a topology, the ID must be the index of the tile in the tiles list. */
   char  name[ 7UL ];            /* The name of this tile.  There can be multiple of each tile name in a topology. */
   ulong kind_id;                /* The ID of this tile within its name.  If there are n tile of a particular name, they have IDs [0, N).  The pair (name, kind_id) uniquely identifies a tile, as does "id" on its own. */
-  int   is_labs;                /* If the tile needs to run in the Solana Labs (Anza) address space or not. */
+  int   is_agave;                /* If the tile needs to run in the Agave (Anza) address space or not. */
```
mmcgee-jump, line 93: nit: extra whitespace

src/app/fdctl/config_parse.c
```diff
@@ -407,7 +408,7 @@ fdctl_cfg_validate( config_t * cfg ) {
   CFG_HAS_NON_ZERO( snapshots.incremental_snapshot_interval_slots );
 
   CFG_HAS_NON_EMPTY( layout.affinity );
-  CFG_HAS_NON_EMPTY( layout.solana_labs_affinity );
+  CFG_HAS_NON_EMPTY( layout.agave_affinity );
```
mmcgee-jump, line 411: What's the behavior here if they specify solana_labs_affinity but not agave_affinity? It will produce an empty value error right? I think we want to allow that, so operator doesn't have to update their config.

anwayde: https://github.com/firedancer-io/firedancer/pull/2498/files#diff-4063f4668e77e6c1002f6c54c0ef5d9fd728a2f3474fd2132abc7a770c033466R664 addresses that. So if `agave_affinity` is non-empty, we will use that. If it is empty (which by default it is), it'll take the value of `solana_labs_affinity`, so operators won't have to update their config.

mmcgee-jump (approved):

lheeger-jump (approved):

PR #2504, opened 2024-07-24
Revert native cpi check

jumpsiegel:


ripatel-fd (approved):

ripatel-fd:
@jumpsiegel can you make a github issue for future reference? I'm still worried about feature activations that add new native programs...

PR #2509, opened 2024-07-24
bpf_loader_set_authority_check: add missing feature check

jumpsiegel:


ibhatt-jumptrading (approved):
this is only needed for the fuzzer, the feature was activated a few months ago

PR #2513, opened 2024-07-24
flamenco, fuzz: Fuzz VM CPI syscalls

ravyu-jump:
~~Introduces a "stubbed" version of `libfd_exec_sol_compat` that wraps around `fd_execute_instr` for CPI syscall fuzzing.~~

Supporting CPI syscalls implements the following:
- Performing a full `_instr_context_create` (i.e, `is_syscall = false`) specifically for CPI calls 
- Summing up account lamports during `_instr_context_create`
- Capturing of data from input_data_regions into SyscallEffects
- Tool to dump `SyscallContext` at CPI entrypoint

Further cleanup:
- Fixed scratch frame mismanagement, particularly use-after-free when writing to output buffer

### DIFF
--- a/src/flamenco/runtime/fd_executor.h
+++ b/src/flamenco/runtime/fd_executor.h
@@ -13,6 +13,16 @@
 
 FD_PROTOTYPES_BEGIN
 
+/* Create an InstrContext protobuf struct from a given 
+   transaction context and instr info. 
+
+   NOTE: Calling this function requires the caller to have a scratch 
+   frame ready and pushed (see dump_instr_to_protobuf) */
+void
+fd_create_instr_context_protobuf_from_instructions( fd_exec_test_instr_context_t       * instr_context,
+                                                    fd_exec_txn_ctx_t            const * txn_ctx,
+                                                    fd_instr_info_t              const * instr );
+
 /* fd_exec_instr_fn_t processes an instruction.  Returns an error code
    in FD_EXECUTOR_INSTR_{ERR_{...},SUCCESS}. */
 
--- a/src/flamenco/runtime/tests/fd_exec_sol_compat.c
+++ b/src/flamenco/runtime/tests/fd_exec_sol_compat.c
@@ -93,10 +93,17 @@ sol_compat_check_wksp_usage( void ) {
   }
 }
 
+sol_compat_features_t const *
+sol_compat_get_features_v1( void ) {
+  return &features;
+}
+
 fd_exec_instr_test_runner_t *
 sol_compat_setup_scratch_and_runner( void * fmem ) {
   // Setup scratch
   fd_scratch_attach( smem, fmem, smax, 64UL );
+  /* Push frame */
+  fd_scratch_push();
 
   // Setup test runner
   void * runner_mem = fd_wksp_alloc_laddr( wksp, fd_exec_instr_test_runner_align(), fd_exec_instr_test_runner_footprint(), WKSP_TAG );
@@ -109,6 +116,9 @@ void
 sol_compat_cleanup_scratch_and_runner( fd_exec_instr_test_runner_t * runner ) {
   /* Cleanup test runner */
   fd_wksp_free_laddr( fd_exec_instr_test_runner_delete( runner ) );
+
+  /* Pop frame */
+  fd_scratch_pop();
   /* Cleanup scratch */
   fd_scratch_detach( NULL );
 }
@@ -152,19 +162,20 @@ sol_compat_execute_wrapper( fd_exec_instr_test_runner_t * runner,
                             void * input,
                             void ** output,
                             exec_test_run_fn_t * exec_test_run_fn ) {
-  // fd_scratch_push();
-  do {
-    ulong out_bufsz = 100000000;  /* 100 MB */
-    void * out0 = fd_scratch_prepare( 1UL );
-    assert( out_bufsz < fd_scratch_free() );
-    fd_scratch_publish( (void *)( (ulong)out0 + out_bufsz ) );
+  
+  assert( fd_scratch_prepare_is_safe( 1UL ) );
+  ulong out_bufsz = 100000000;  /* 100 MB */
+  void * out0 = fd_scratch_prepare( 1UL );
+  assert( out_bufsz < fd_scratch_free() );
+  fd_scratch_publish( (void *)( (ulong)out0 + out_bufsz ) );
+
+  FD_SCRATCH_SCOPE_BEGIN {
     ulong out_used = exec_test_run_fn( runner, input, output, out0, out_bufsz );
     if( FD_UNLIKELY( !out_used ) ) {
       *output = NULL;
       break;
     }
-  } while(0);
-  // fd_scratch_pop();
+  } FD_SCRATCH_SCOPE_END;
 }
 
 /*
@@ -484,10 +495,6 @@ sol_compat_elf_loader_v1( uchar *       out,
   return ok;
 }
 
-sol_compat_features_t const *
-sol_compat_get_features_v1( void ) {
-  return &features;
-}
 
 int
 sol_compat_vm_syscall_execute_v1( uchar *       out,
@@ -530,7 +537,7 @@ int
 sol_compat_vm_validate_v1(  uchar *       out,
                             ulong *       out_sz,
                             uchar const * in,
-                            ulong         in_sz) {
+                            ulong         in_sz ) {
   // Setup
   ulong fmem[ 64 ];
   fd_exec_instr_test_runner_t * runner = sol_compat_setup_scratch_and_runner( fmem );
@@ -561,4 +568,15 @@ sol_compat_vm_validate_v1(  uchar *       out,
   sol_compat_check_wksp_usage();
 
   return ok;
+}
+
+/* We still need a separate entrypoint since other harnesses (namely sfuzz-agave) 
+   do something other than wrap their vm_syscall equivalent */
+int
+sol_compat_vm_cpi_syscall_v1( uchar *       out,
+                              ulong *       out_sz,
+                              uchar const * in,
+                              ulong         in_sz ) {
+  /* Just a wrapper to vm_syscall_execute_v1 */
+  return sol_compat_vm_syscall_execute_v1( out, out_sz, in, in_sz );
 }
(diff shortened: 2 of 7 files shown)


src/flamenco/vm/syscall/fd_vm_syscall_cpi.c
```diff
@@ -4,6 +4,105 @@
...
+                            pb_byte_t const* buf,
+                            ulong bufsz ) {
+  region->size = bufsz;
+  region->bytes_const = buf;
+  cb->arg = region;
+  cb->funcs.encode = write_bytes_callback;
+}
+
+static inline void
+dump_vm_cpi_state(fd_vm_t *vm,
```
topointon-jump, line 30: Could we add a comment explaining what this is?

src/flamenco/runtime/fd_executor.h
```diff
@@ -13,6 +13,11 @@
 
 FD_PROTOTYPES_BEGIN
 
+void
+fd_create_instr_context_protobuf_from_instructions( fd_exec_test_instr_context_t * instr_context,
+                                                 fd_exec_txn_ctx_t const *txn_ctx,
+                                                 fd_instr_info_t const *instr );
+
```
ravyu-jump, line 20: @mjain-jump when and why did we remove this decl?

src/flamenco/runtime/tests/Local.mk
```diff
@@ -10,6 +11,7 @@ $(call add-objs,fd_exec_sol_compat,fd_flamenco)
 $(call make-unit-test,test_exec_instr,test_exec_instr,fd_flamenco fd_funk fd_ballet fd_util fd_disco,$(SECP256K1_LIBS))
 $(call make-unit-test,test_exec_sol_compat,test_exec_sol_compat,fd_flamenco fd_funk fd_ballet fd_util fd_disco,$(SECP256K1_LIBS))
 $(call make-shared,libfd_exec_sol_compat.so,fd_exec_sol_compat,fd_flamenco fd_funk fd_ballet fd_util fd_disco,$(SECP256K1_LIBS))
+$(call make-shared,libfd_exec_sol_compat_stubbed.so,fd_exec_sol_compat,fd_flamenco fd_funk fd_ballet fd_util fd_disco,$(SECP256K1_LIBS) $(WRAP_FLAGS))
```
ripatel-fd, line 14: Why disable fd_execute_instr? Why not just make one binary with everything enabled?

ravyu-jump: But yeah having a second binary is unfortunate, and I'm still not sure if we can mock anything on the Agave end. So this might change.

ripatel-fd: @ravyu-jump Makes sense. Can we move this out to a separate PR? I believe there are better ways to do this, e.g. LD_PRELOADing certain symbols. The wrap hack looks rather non-portable so is likely to break portability.

ravyu-jump: Sure, I can do that. 

> LD_PRELOADing certain symbols
Wouldn't this restrict wrapping to only when using the .so file? IIRC our test-vectors CI runs a statically linked executable instead of the .so so that wouldn't work. But of course open to other ideas

> wrap hack looks rather non-portable
Both gcc and clang support this, unless you mean architecture-wise?

ripatel-fd: > Wouldn't this restrict wrapping to only when using the .so file? IIRC our test-vectors CI runs a statically linked executable instead of the .so so that wouldn't work.

The wrap hack is currently only applied to the shared library which is why I suggested LD_PRELOAD instead.
If we need this for test executables too why not just fix this in the source code? We could trivially indirect the instruction executor call.

> Both gcc and clang support this, unless you mean architecture-wise?

It's a link option, so must be supported by bfd, ld.lld, mold, etc. On most systems the linker is provided by the system and you can't replace it as easily as the compiler.

ripatel-fd (changes requested):
Please fix code style of new code

See https://github.com/firedancer-io/firedancer/blob/main/CONTRIBUTING.md

src/flamenco/runtime/tests/Local.mk
```diff
@@ -1,6 +1,7 @@
 $(call add-hdrs,generated/context.pb.h,generated/elf.pb.h,generated/invoke.pb.h,generated/txn.pb.h,generated/vm.pb.h)
 $(call add-objs,generated/context.pb generated/elf.pb generated/invoke.pb generated/txn.pb generated/vm.pb,fd_flamenco)
 
+WRAP_FLAGS += -Xlinker --wrap=fd_execute_instr
```
topointon-jump, line 4: Could we add a comment explaining this wrapper approach?

src/flamenco/runtime/tests/Local.mk
topointon-jump: As most users don't actually need this binary, maybe we can conditionally build this for the few systems that the core devs use? To avoid breaking the whole build.

src/flamenco/runtime/tests/generated/invoke.pb.h
```diff
@@ -39,6 +39,9 @@ typedef struct fd_exec_test_instr_context {
     fd_exec_test_slot_context_t slot_context;
     bool has_epoch_context;
     fd_exec_test_epoch_context_t epoch_context;
+    /* FIXME: move to future transaction context protobuf
+ Tag is intentionally set to 12 DO NOT CHANGE */
+    uint64_t heap_size;
```
topointon-jump, line 44: Need to regenerate this file

src/flamenco/vm/syscall/fd_vm_syscall_cpi.c
```diff
@@ -4,6 +4,133 @@
...
+#include <errno.h>
+#include "../../nanopb/pb_encode.h"
+#include "../../runtime/tests/generated/vm.pb.h"
+#include "../../runtime/tests/fd_exec_instr_test.h"
+
+#define STRINGIFY(x) TOSTRING(x)
+#define TOSTRING(x) #x
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-function"
```
topointon-jump, line 19: FD_FN_UNUSED

src/flamenco/vm/fd_vm_base.h
```diff
@@ -79,6 +79,16 @@
...
+/* Error codes related to CPI syscall.
+   FIXME: Should this be in fd_executor_err.h instead?
+          Or a separate file for syscall errors? */
+
+#define FD_VM_CPI_ERR_TOO_MANY_SIGNERS (-39) /* detected too many signers */
+#define FD_VM_CPI_ERR_TOO_MANY_ACC_INFOS (-40) /* detected too many account infos */
+#define FD_VM_CPI_ERR_INSTR_TOO_LARGE (-41) /* detected too many account infos meta */
+#define FD_VM_CPI_ERR_INSTR_DATA_TOO_LARGE (-42) /* detected instruction data too large */
+#define FD_VM_CPI_ERR_TOO_MANY_ACC_METAS (-43) /* detected too many account metas */
+
```
topointon-jump, line 91: nit: align

src/flamenco/runtime/fd_executor.h
topointon-jump: could we add a doccomment?

src/flamenco/runtime/tests/fd_exec_sol_compat.c
```diff
@@ -578,3 +589,12 @@ sol_compat_vm_validate_v1(  uchar *       out,
 
   return ok;
 }
+
+int
+sol_compat_vm_cpi_syscall_v1( uchar *       out,
```
topointon-jump, line 594: can you add a comment explaining why we need a different entrypoint

src/flamenco/runtime/tests/fd_exec_sol_compat.c
```diff
@@ -166,7 +181,7 @@ sol_compat_execute_wrapper( fd_exec_instr_test_runner_t * runner,
                             void * input,
                             void ** output,
                             exec_test_run_fn_t * exec_test_run_fn ) {
-  // fd_scratch_push();
+  fd_scratch_push();
```
topointon-jump, line 184: Can we use the macros here?

src/flamenco/runtime/tests/fd_exec_sol_compat.c
```diff
@@ -24,6 +24,11 @@ typedef struct {
   ulong   supported_feature_cnt;
 } sol_compat_features_t;
 
+typedef struct {
+  uint16_t validator_type;
+} sol_compat_metadata_t;
+
+static sol_compat_metadata_t metadata = { .validator_type = 1 /* Firedancer */ };
```
topointon-jump, line 31: potentially this could be a protobuf message..?

src/flamenco/runtime/tests/Local.mk
ripatel-fd: I agree @topointon-jump, but the wrap linker hack is not the right way to achieve this.

ravyu-jump: Moved this to a separate PR: https://github.com/firedancer-io/firedancer/pull/2601

ripatel-fd (approved):

ravyu-jump:
@ripatel-fd, had to re-implement capturing of input data regions after the direct-mapping merge. Sorry about that here's the diff: https://github.com/firedancer-io/firedancer/compare/0a3fd0c1b6bfd429f065277e4152c847c0fb6eeb..c1b4187d4794d3b3b8a42efe4c74bb81a8d60189

src/flamenco/runtime/tests/fd_exec_sol_compat.c
```diff
@@ -163,8 +177,7 @@ sol_compat_execute_wrapper( fd_exec_instr_test_runner_t * runner,
       *output = NULL;
       break;
     }
-  } while(0);
-  // fd_scratch_pop();
+  } FD_SCRATCH_SCOPE_END;
```
ripatel-fd, line 180: IIRC these functions publish in to the caller's scratch frame so freeing the frame here could cause memory corruption (use after free)

ravyu-jump: @ripatel-fd https://github.com/firedancer-io/firedancer/compare/7421edee062eace051669dd85a7f8d20110d3bc7..c810ce395d4b828970bce8c441e6a3ddcaeec0b5

I think this just fixes the obvious use-after-free. Will probably need to do a bit of restructuring with how we use scratch frames here at some point.

ripatel-fd (approved):

PR #2514, opened 2024-07-24
docs: tidy up and add metrics page

### LINKED ISSUE #1011: Add documentation for metrics
Automatically generate metrics documentation pages from the metrics XML definition

mmcgee-jump:
Fixes #1011

### DIFF
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -2,7 +2,7 @@
 # so that multiple Firedancer instances can run side by side without
 # conflicting when they need to share a system or kernel namespace.
 # When starting a Firedancer instance, it will potentially load, reset,
-# or overwrite any state created by a prior, or currently running
+# or overwrite any state created by a prior or currently running
 # instance with the same name.
 name = "fd1"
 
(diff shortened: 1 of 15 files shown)


src/app/fdctl/config/default.toml
```diff
@@ -11,11 +11,11 @@ name = "fd1"
...
 # restrictive sandbox, drop all privileges, and switch to the user given
-# here.  When running the configuration steps of `fdctl configure`, data
-# will be permissioned as read/write for this user, irrespective of
-# which user is performing the configuration.
+# here.  When running the configuration steps of `fdctl configure` data
+# will be permissioned so that is it writable for this user and not the
+# user which is performing the configuration.
 #
-# Firedancer requires nothing from this user, and it should be as
+# Firedancer requires nothing from this user and it should be as
```
ptaffet-jump, line 18:
```suggestion
# Firedancer requires nothing from this user, and it should be as
```
I think the comma is right here because "it should be as minimally permissioned as possible" is a whole sentence.

ptaffet-jump (commented):
A few small things I noticed when proofreading

ptaffet-jump (approved):

PR #2520, opened 2024-07-26
match sol_log_64 output

two-heart:
Currently, these are example logs from Firedancer and agave:

```
Program log: 5d724c44c0a108c8 ac246eae6b5f6c8a 7bd2aac7cd318b02 b8fb78d524538540 7a3034c712804303
Program log: 0x5d724c44c0a108c8, 0xac246eae6b5f6c8a, 0x7bd2aac7cd318b02, 0xb8fb78d524538540, 0x7a3034c712804303
```

### DIFF
--- a/src/flamenco/vm/syscall/fd_vm_syscall_util.c
+++ b/src/flamenco/vm/syscall/fd_vm_syscall_util.c
@@ -101,7 +101,8 @@ fd_vm_syscall_sol_log_64( void *  _vm,
   ulong  msg_max = fd_vm_log_prepare_max( vm );
   ulong  msg_len;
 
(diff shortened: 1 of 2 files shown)


topointon-jump (dismissed):

ripatel-fd (changes requested):
You missed the commas

two-heart:
Oh wow, you are right 🤦 sec

ripatel-fd (approved):

ripatel-fd (changes requested):
@two-heart Can you fix the tests? Looks like we need to increase some hardcoded size of a log buffer.

ripatel-fd (approved):

PR #2528, opened 2024-07-29
initial set of hardening patches

anarcheuz:


anarcheuz:
@two-heart @nlgripto @marctrem please review/improve.

anarcheuz:
Unsure how to bound-check for `TREAP_(ele) and TREAP_(ele_const)` as we do not have the implementation of TREAP_T. Could be a pool or anything else.

jumpsiegel:
Turns out, the wrapping behavior of the queue's are an intended behavior and it is unclear if these hardening checks are valid.  I think we need to understand what tests are being tripped by the current behavior.  It. is also obvious we need to update the documentation so avoid this confusion in the future

two-heart:
@jumpsiegel I suggest that we add add a `DEQUE_CHECK_WRAP_HANDHOLDING` that default to false, but the user can set it to true, when wrapping is not intended. Example:
```c
#define DEQUE_NAME fd_tower_votes
#define DEQUE_T    fd_tower_vote_t
#define DEQUE_MAX  FD_TOWER_VOTE_MAX
#define DEQUE_CHECK_WRAP_HANDHOLDING 1
#include "../../util/tmpl/fd_deque.c"
```
@anarcheuz you should use `FD_LOG_ERR` to be consistent with the other handholding code and have no space after the if

ibhatt-jumptrading:
Add unit tests to the corresponding utils

ripatel-fd:
Closing since https://github.com/firedancer-io/firedancer/pull/4001 attempts to fix the same problem.
Please reopen if you think this is a mistake.

PR #2551, opened 2024-07-30
test-vectors: fix asan/ubsan complains

0x0ece:
- [x] enable warn logs for all tests
- [x] remove unnecessary warn logs (cluttering output / hiding real issues)
- [x] curve25519: fix ulong load alignment
- [x] sha256: fix uint bswap alignment
- [x] utf8: handle null pointer
- [x] programs: handle null instr->data
- [x] zk-sdk: fix uint load alignment
- [x] bincode: handle null pointer
- [x] sol_compat: fix asan complaints with fn interface
- [x] test_exec_instr: fix page size and boundaries

### DIFF
--- a/src/ballet/sha256/fd_sha256.c
+++ b/src/ballet/sha256/fd_sha256.c
@@ -349,16 +349,15 @@ fd_sha256_fini( fd_sha256_t * sha,
 
   /* Unpack the result into md (annoying bswaps here) */
 
-  uint * hash = (uint *)_hash;
-  hash[0] = fd_uint_bswap( state[0] );
-  hash[1] = fd_uint_bswap( state[1] );
-  hash[2] = fd_uint_bswap( state[2] );
-  hash[3] = fd_uint_bswap( state[3] );
-  hash[4] = fd_uint_bswap( state[4] );
-  hash[5] = fd_uint_bswap( state[5] );
-  hash[6] = fd_uint_bswap( state[6] );
-  hash[7] = fd_uint_bswap( state[7] );
-  return _hash;
+  state[0] = fd_uint_bswap( state[0] );
+  state[1] = fd_uint_bswap( state[1] );
+  state[2] = fd_uint_bswap( state[2] );
+  state[3] = fd_uint_bswap( state[3] );
+  state[4] = fd_uint_bswap( state[4] );
+  state[5] = fd_uint_bswap( state[5] );
+  state[6] = fd_uint_bswap( state[6] );
+  state[7] = fd_uint_bswap( state[7] );
+  return memcpy( _hash, state, 32 );
 }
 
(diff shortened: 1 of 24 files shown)


jumpsiegel (approved):

src/ballet/sha256/fd_sha256.c
```diff
@@ -349,16 +349,15 @@ fd_sha256_fini( fd_sha256_t * sha,
...
-  return _hash;
+  state[0] = fd_uint_bswap( state[0] );
+  state[1] = fd_uint_bswap( state[1] );
+  state[2] = fd_uint_bswap( state[2] );
+  state[3] = fd_uint_bswap( state[3] );
+  state[4] = fd_uint_bswap( state[4] );
+  state[5] = fd_uint_bswap( state[5] );
+  state[6] = fd_uint_bswap( state[6] );
+  state[7] = fd_uint_bswap( state[7] );
+  return fd_memcpy( _hash, state, 32 );
```
ripatel-fd, line 360:
```suggestion
  return memcpy( _hash, state, 32 );
```

Normal memcpy is preferred here because it allows the compiler to generate faster unaligned SIMD stores. 
fd_memcpy is actually slower here because the `state` variable is in SIMD registers, not in memory.

ripatel-fd (approved):

src/ballet/bn254/fd_bn254.c
```diff
@@ -166,7 +166,9 @@ fd_bn254_g1_add_syscall( uchar       out[64],
     return -1;
   }
   uchar FD_ALIGNED buf[128] = { 0 };
-  fd_memcpy( buf, in, in_sz );
+  if( in_sz ) {
+    fd_memcpy( buf, in, in_sz );
+  }
```
ripatel-fd, line 171: @0x0ece Sorry, would you mind reverting this change? 
Unlike stdlib memcpy, fd_memcpy is actually fine with a 0 size argument for all supported build targets (test_util_base verifies that). The reason is that 0 sz stdlib memcpy is fine on glibc, GCC, and Clang.

There are still some "insane" targets like CBMC or UBSan that complain on 0 sz stdlib memcpy, but for those, fd_memcpy does this branch under the hood.

ripatel-fd (changes requested):
Looks great but I suggest reverting the conditional fd_memcpy change for simplicity.

src/ballet/bn254/fd_bn254.c
0x0ece: @ripatel-fd I've got this issue with:
```
MACHINE=linux_clang_x86_64 EXTRAS="asan ubsan rpath" make run-test-vectors
```

ripatel-fd (approved):

PR #2566, opened 2024-08-01
Detect mismatching Ed25519 pubkey in fd_keyload

### LINKED ISSUE #2554: fd_keyload_load should check that keypair matches
Ed25519 security breaks down when signing messages with a public key that doesn't match the private key.
Loading certain key files can result in compromise of the private key.

This can happen due to user error (e.g. user uses a broken script to turn a private key into a JSON key file).
And should thus be fixed.

ripatel-fd:
Defends against user error by checking that private key and public key
match in the key pair file.

Adds end-to-end tests for fd_keyload.

Closes https://github.com/firedancer-io/firedancer/issues/2554

### DIFF
--- a/src/disco/keyguard/fd_keyload.c
+++ b/src/disco/keyguard/fd_keyload.c
@@ -68,6 +68,19 @@ read_key( char const * key_path,
 #undef KEY_SZ
 #undef KEY_PARSE_ERR
 
+  /* Verify the public key */
+  fd_sha512_t _sha[1];
+  fd_sha512_t * sha = fd_sha512_join( fd_sha512_new( _sha ) );
+  if( FD_UNLIKELY( !sha ) ) FD_LOG_ERR(( "could not validate keypair, fd_sha512 join failed" ));
+  uchar check_public_key[ 32 ];
+  fd_ed25519_public_from_private( check_public_key, key, sha );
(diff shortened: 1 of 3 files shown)


jumpsiegel (approved):

src/disco/keyguard/fd_keyload.c
```diff
@@ -68,6 +68,19 @@ read_key( char const * key_path,
...
 #undef KEY_PARSE_ERR
 
+  /* Verify the public key */
+  fd_sha512_t _sha[1];
+  fd_sha512_t * sha = fd_sha512_join( fd_sha512_new( _sha ) );
+  if( FD_UNLIKELY( !sha ) ) FD_LOG_ERR(( "could not validate keypair, fd_sha512 join failed" ));
+  uchar check_public_key[ 32 ];
+  fd_ed25519_public_from_private( check_public_key, key, sha );
+  if( FD_UNLIKELY( memcmp( check_public_key, key+32, 32 ) ) )
+    FD_LOG_EMERG(( "!!! INVALID KEY FILE !!!\n"
```
mmcgee-jump, line 78: Will dump the stack which might have private key material in it due to use of sscanf / other glibc crap.

mmcgee-jump (changes requested):
.

PR #2571, opened 2024-08-01
txn: remove allow_zero_signatures flag

ripatel-fd:
allow_zero_signatures was an ugly hack added by the runtime team,
only used for testing.  We have since moved on to better testing
so we can remove this hack.


jumpsiegel (approved):
since we now test at the instruction level, this is not required anymore... good change

PR #2583, opened 2024-08-02
Hardcode reject_callx_r10 feature

ripatel-fd:
The feature reject_callx_r10 is enabled on all clusters and thus should be hardcoded

Depends on https://github.com/anza-xyz/agave/pull/2414


ravyu-jump (approved):

ripatel-fd:
We'll merge this PR when Agave releases a new version with https://github.com/anza-xyz/agave/pull/2414 merged.

ripatel-fd:
reject_callx_r10 was hardcoded as enabled in Agave v2.1.0.
https://github.com/anza-xyz/agave/pull/2884

topointon-jump (approved):

PR #2585, opened 2024-08-02
flamenco: remove return on failed fd_vote_get_state

kbhargava-jump:
solana does not fail if this line returns an error:
```
let vote_state = vote_account.get_state::<VoteStateVersions>();
```


jumpsiegel (dismissed):

ripatel-fd (changes requested):
Wait, how is this safe? Doesn't this leave `vote_state` in an undefined state?

kbhargava-jump:
> Wait, how is this safe? Doesn't this leave `vote_state` in an undefined state?

ive added the get state logic to be later and added the check back in

ripatel-fd (dismissed):

ibhatt-jumptrading (approved):

PR #2589, opened 2024-08-04
flamenco, capture: fix solcap at epoch boundary

topointon-jump:
This is a short-term fix for solcap at mainnet epoch boundaries, where the size of the accounts table needs to be quite large (~1000000). We size it to be the smallest power of 2 greater than the required capacity. We also change the initialization of the writer to avoid smashing the stack.

Note that this is a short-term fix. A better long-term solution would be to dynamically size the table as the size of this account table varies dramatically between epoch boundary slots and non-epoch boundary slots.

### DIFF
--- a/src/flamenco/capture/fd_solcap_proto.h
+++ b/src/flamenco/capture/fd_solcap_proto.h
@@ -144,7 +144,7 @@ typedef struct fd_solcap_account_tbl fd_solcap_account_tbl_t;
 /* FD_SOLCAP_ACC_TBL_CNT is the number of entries that fit in the in-
    memory buffer for the account table. */
 
-#define FD_SOLCAP_ACC_TBL_CNT (8192U)
+#define FD_SOLCAP_ACC_TBL_CNT (2097152U)
 
 /* FD_SOLCAP_FILE_META_FOOTPRINT is the max size of the FileMeta
    Protobuf struct. */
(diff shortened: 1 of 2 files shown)


ripatel-fd:
> A better long-term solution would be to dynamically size the table as the size of this account table varies dramatically between epoch boundary slots and non-epoch boundary slots.

@topointon-jump We generally try to avoid dynamic sized allocations. We should instead adapt the file format to support streaming with constant size memory buffers.

Currently, the writer produces a stream like so. The in-memoy buffer serves to remember the written account offsets to they can later be placed in a table. 

```
[ account record ] <--------+
[ account record ]          |
[ account record ]          |
[ ... ]                     |
[ account table chunk ] ----+
```

A simple way to support streaming would be linked lists:

```
[ account record ] <--------+
[ account record ]          |
[ account record ]          |
[ ... ]                     |
[ account table chunk 1 ] --+---+
                                | 
[ account record ] <--------+   |
[ account record ]          |   |
[ account record ]          |   |
[ ... ]                     |   |
[ account table chunk 2 ] --+   |
                    /\          | 
                    ++----------+
```

topointon-jump:
Closing this as it's not worth the slowdown

ibhatt-jumptrading:
I actually think this PR would be a useful short-term fix to get in

topointon-jump:
@ibhatt-jumptrading I'm bundling something similar into the epoch rewards PR

PR #2590, opened 2024-08-04
jit: initial commit

ripatel-fd:
Adds first pieces of a JIT compiler.
The ALU and jump instruction classes are mostly complete.
Nothing is tested and many pieces are still missing.


src/flamenco/vm/fd_vm_jitproto.c
```diff
@@ -0,0 +1,1259 @@
+/*
```
topointon-jump, line 1: Would you be able to put all of the JIT-related code in a subdirectory? So it's clear to auditors and contributors what code is meant for production and what code is experimental. Thanks so much 🙌

ripatel-fd: Good idea. Done

topointon-jump (approved):

deps.sh
```diff
@@ -110,11 +110,15 @@ fetch () {
     checkout_repo zlib      https://github.com/madler/zlib            "v1.3.1"
     checkout_repo rocksdb   https://github.com/facebook/rocksdb       "v9.4.0"
     checkout_repo snappy    https://github.com/google/snappy          "1.2.1"
+    checkout_repo luajit    https://github.com/LuaJIT/LuaJIT          "v2.0.5"
   fi
 }
 
 check_fedora_pkgs () {
   local REQUIRED_RPMS=( perl autoconf gettext-devel automake flex bison cmake clang gmp-devel protobuf-compiler llvm-toolset lcov systemd-devel pkgconf )
+  if [[ $DEVMODE == 1 ]]; then
```
ravyu-jump, line 119: I might have messed up something, but running `./deps.sh +dev` didn't warn me about missing these.

ravyu-jump (changes requested):
Build broke with the moving around of files

deps.sh
ripatel-fd: good find, i had a typo.

ravyu-jump (dismissed):

ravyu-jump (approved):

PR #2595, opened 2024-08-05
Drop tiny TPU/UDP packets

ripatel-fd:
Ensure that incoming UDP packets can fit at least a minimum size transaction.

### DIFF
--- a/src/app/fdctl/run/tiles/fd_quic.c
+++ b/src/app/fdctl/run/tiles/fd_quic.c
@@ -148,7 +148,7 @@ mux_ctx( void * scratch ) {
 static void
 legacy_stream_notify( fd_quic_ctx_t * ctx,
                       uchar *         packet,
-                      uint            packet_sz ) {
+                      ulong           packet_sz ) {
 
   fd_mux_context_t * mux = ctx->mux;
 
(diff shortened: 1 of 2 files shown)


ptaffet-jump:
I think the net tile should throw out anything that's not UDP, but a UDP payload of 0 bytes seems fine to me.

src/app/fdctl/run/tiles/fd_quic.c
```diff
@@ -304,22 +304,29 @@ after_frag( void *             _ctx,
     fd_aio_send( ctx->quic_rx_aio, &pkt, 1, NULL, 1 );
   } else if( FD_LIKELY( proto==DST_PROTO_TPU_UDP ) ) {
     ulong network_hdr_sz = fd_disco_netmux_sig_hdr_sz( *opt_sig );
-    if( FD_UNLIKELY( *opt_sz<network_hdr_sz ) ) {
```
ptaffet-jump, line 307: I don't think this change is necessary, but it's okay

ptaffet-jump (approved):
Sorry, I thought the changes to quic were changes to net for some reason. This looks good to me.

PR #2601, opened 2024-08-05
flamenco, fuzz: stub out fd_execute_instr with --wrap

ravyu-jump:
For certain fuzz harnesses (namely, the CPI syscall), we want to control the scope of the fuzzer (which is coverage-guided) by stubbing out certain functions. This PR uses ld's `--wrap` functionality to stub out `fd_execute_instr`. 

This is currently a simple stub, but the plan is to simulate the effects of an `fd_execute_instr` call so that the CPI clean up stage isn't effectively a NO-OP. Wrap allows us to redefine functions in a separate translation unit, which would give the fuzz harnesses more control over what it can do (particularly, access to static variables within TUs).

### DIFF
--- a/.github/workflows/clusterfuzz.yml
+++ b/.github/workflows/clusterfuzz.yml
@@ -85,9 +85,16 @@ jobs:
           ls ${{ env.OBJ_DIR }}/fuzz-test
           ls ${{ env.OBJ_DIR }}/lib
 
-      - name: upload lib artifact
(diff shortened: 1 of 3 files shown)


ripatel-fd:
Btw, in case the virtual call API for process_instruction takes too long happy to just merge this instead

ripatel-fd (approved):
Looks great

PR #2610, opened 2024-08-06
fix postinc off-by-one in memcmp syscall

two-heart:
fixes https://github.com/firedancer-io/auditor-internal/issues/172


ibhatt-jumptrading (approved):
thanks for the catch!

PR #2613, opened 2024-08-06
quic - fixed default.toml - stream_pool was too small to function proerly

nbridge-jump:
stream pool was so small that all the available streams were allocated to unused connections, and remain there


mmcgee-jump:
Can we add some documentation to the TOML, and perhaps flag errors on boot for misconfigured cases here? This seems like an easy footgun.

mmcgee-jump (changes requested):
.

nbridge-jump:
Superseded by https://github.com/firedancer-io/firedancer/pull/2627

PR #2635, opened 2024-08-07
flamenco: keep finalized slots in blockstore as many as possible

yhzhangjump:


lheeger-jump:
Looking good, needs testing on testnet.

PR #2636, opened 2024-08-07
disco: add epoch forks

arjain4:


src/choreo/forks/fd_epoch_forks.h
lheeger-jump, line 1: document this data structure! why is it needed, what is it for?

src/disco/store/fd_store.c
```diff
@@ -226,7 +226,7 @@ fd_store_shred_insert( fd_store_t * store,
   if( FD_UNLIKELY( rc < FD_BLOCKSTORE_OK ) ) {
     FD_LOG_ERR( ( "failed to insert shred. reason: %d", rc ) );
   } else if ( rc == FD_BLOCKSTORE_OK_SLOT_COMPLETE ) {
-    FD_LOG_WARNING(("BLOCK COMPLETE"));
+    FD_LOG_DEBUG(("BLOCK COMPLETE"));
```
lheeger-jump, line 229: rm this log

lheeger-jump (changes requested):
Please add a stub unit test like the new `test_trusted_slots.c` and add it to Local.mk

lidatong (changes requested):
can you move out of choreo into disco? choreo is really just meant for consensus rules

otherwise looks good superficially to me

lheeger-jump (approved):

PR #2639, opened 2024-08-07
fix: send back input and rodata memory regions for differential fuzzing

anarcheuz:


ibhatt-jumptrading (approved):

mjain-jump (approved):

0x0ece:
FYI, the test vectors that are failing were generated when solfuzz-agave didn't output rodata (and firedancer required rodata != empty). So the context contains rodata, the effects doesn't (because of the old solfuzz-agave) and the computed effects now contain rodata (because of this PR).

Long story short, I'm going to regenerate the fixtures. So we can temp turn these tests off and I'll add them back in a separated PR.

mjain-jump (approved):

mjain-jump (approved):

0x0ece (approved):

src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -1791,15 +1791,23 @@ fd_exec_vm_syscall_test_run( fd_exec_instr_test_runner_t * runner,
...
-  //   fd_memcpy( effects->inputdata->bytes, vm->input, input_data_sz );
-  // } else {
-  //   effects->inputdata = NULL;
-  // }
-  effects->inputdata = NULL;
+  if( input_data_sz ) {
+    effects->inputdata = FD_SCRATCH_ALLOC_APPEND(
+      l, alignof(uint), PB_BYTES_ARRAY_T_ALLOCSIZE( input_data_sz ) );
+    effects->inputdata->size = (uint)input_data_sz;
+    fd_memcpy( effects->inputdata->bytes, (const void*)vm->input_mem_regions[0].haddr, input_data_sz );
```
topointon-jump, line 1798: We should return and compare all the input regions, not just the first one.

PR #2647, opened 2024-08-08
flamenco: use proper function for deriving PDA

mjain-jump:
Fixtures added here: https://github.com/firedancer-io/test-vectors/commit/27e1eaf71de0634458018763567dcafd73dca32b


topointon-jump (approved):
This is a great find! Go fuzzing 🔥

PR #2652, opened 2024-08-08
pack, bank: rebate unused CUs (v0.1)

ptaffet-jump:
Rebase of #2651 to merge onto the v0.1 branch

> Notify pack for transactions that fail to sanitize or executing, and for transactions that use fewer than the requested CUs. If pack is still packing for the same block, it can then schedule additional transactions to use those CUs.

> While I was looking closely at all this, I also noticed a place where our cost estimate differed from Agave. Since enforcing the cost model on replay is not part of consensus, this is not a critical issue, but it seemed worth fixing.


ptaffet-jump:
We're going to do this the normal way. No need for a separate PR.

PR #2657, opened 2024-08-09
quic: fix frame_type validation

two-heart:
Thanks to Immunefi user @ret2happy for reporting this bug. Fixes https://github.com/firedancer-io/auditor-internal/issues/181


ripatel-fd (changes requested):
The link in the commit message is broken

ripatel-fd (approved):

PR #2680, opened 2024-08-12
flamenco: combining prepare and execute tpool task

ibhatt-jumptrading:
Combining tpool tasks for better runtime performance.


ibhatt-jumptrading:
redid this pr

PR #2696, opened 2024-08-14
Remove unnecessary code

xinyijun:
1. Move the call to fcntl( pipefd, F_SETFD, 0 ) into the child process; the parent process will be not impacted, the call to fcntl( pipefd, F_SETFD, FD_CLOEXEC ) can be removed.
2. Remove the while loop around the system call poll. poll only returns 0 if we set a non-negative timeout and expires before any watched file descriptor becomes available. Since we use -1 as a timeout, this would never happen. The option WNOHANG and spurious wakeup logic are also removed because once the poll returns a value other than -1, we are guaranteed that there must be at least one process exit.
3. Remove statement config->topo = *topo. Both the 2 pointers point to the same object and the topo is a big-size struct; assigning it to itself is unnecessary and might be wasteful.

### DIFF
--- a/src/app/fdctl/run/run.c
+++ b/src/app/fdctl/run/run.c
@@ -127,10 +127,11 @@ create_clone_stack( void ) {
 static int
 execve_agave( int config_memfd,
                     int pipefd ) {
-  if( FD_UNLIKELY( -1==fcntl( pipefd, F_SETFD, 0 ) ) ) FD_LOG_ERR(( "fcntl(F_SETFD,0) failed (%i-%s)", errno, fd_io_strerror( errno ) ));
   pid_t child = fork();
   if( FD_UNLIKELY( -1==child ) ) FD_LOG_ERR(( "fork() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
   if( FD_LIKELY( !child ) ) {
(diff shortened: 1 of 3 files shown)


mmcgee-jump:
Thanks for the contribution! I didn't look closely but the tests are failing, I think the config->topo removal at least might not be right?

xinyijun:
Thanks for helping review and trigger the tests. It was caused by replacing "if( FD_UNLIKELY( WEXITSTATUS( wstatus ) ) )" with "else": it caused the parent process(pidns) exits with 1 when the child process exits normally.  Fixed.

xinyijun:
@mmcgee-jump, can you help to review this PR? Thanks!

mmcgee-jump:
The code is correct as written. In future, for core boot changes like this, please separate them and and provide a thorough justification for why it's needed.

From `man 2 poll`
> BUGS
>       See the discussion of spurious readiness notifications under the BUGS section of select(2).

From `main 2 select`
> BUGS
> Under Linux, select() may report a socket file descriptor as "ready for reading", while nevertheless a subsequent read blocks.  This could for example happen when data has arrived but upon
> examination has wrong checksum and is discarded.  There may be other circumstances in which a file descriptor is spuriously reported as ready.  Thus it may be safer to  use  O_NONBLOCK  on
> sockets that should not block.

xinyijun:
Got it. Thank you for the detailed explanation.

PR #2698, opened 2024-08-14
flamenco: fix buffer size check in bpf loader

two-heart:
fixes https://github.com/firedancer-io/auditor-internal/issues/192


jumpsiegel (approved):

two-heart:
Independently fixed in commit afc6f592

PR #2702, opened 2024-08-14
quic: fix casting precedence in fd_tls_transcript_load

two-heart:
This was not a real bug. A handshake will never get this big, but it is good practice to do casting right.

Drive-by: comment typos


ripatel-fd (approved):

ripatel-fd:
Thank you

PR #2707, opened 2024-08-14
flamenco: allocating all instr infos from txn_ctx

ibhatt-jumptrading:
1. removing bad stack allocation for instr infos (would've led to bad memory accesses in process sibling syscall)
2. fixing bug in CPI that doesn't error out for unknown accounts


ripatel-fd (approved):
Great change, thanks for getting rid of the dynamic allocation mess

kbhargava-jump (approved):

PR #2720, opened 2024-08-15
bank: enable extended tx metadata storage when rpc requires it

0x0ece:
When RPC is enabled, even with extended tx metadata storage, `logMessages` is not returned.

This patch propagates the RPC config to bank tile, so that when we invoke [bank.load_and_execute_transactions](https://github.com/firedancer-io/agave/pull/2/files#diff-ed47b4a0198313377e091bb3957bbbc63d937805426d1b2b6de39d0a50d32a0cR292) we can enable logging.

For context, to enable RPC:
```
[rpc]
port = 8899
full_api = true
transaction_history = true
extended_tx_metadata_storage = true
```


mmcgee-jump:
This should be behind the `transaction_history` config flag, not `extended_tx_metadata_storage` (you can trace this back on the Agave side to see that). But also I prefer to not modify the ABI for this, precisely because of the risk of issues like this.

Just move the `fd_ext_bank_load_and_execute_txns` to `committer.rs`, same as `fd_ext_bank_pre_balance_info` and then access `committer.transaction_status_sender_enabled()` for these fields instead.

mmcgee-jump (changes requested):
.

0x0ece:
Updated:
- moved `fd_ext_bank_load_and_execute_txns` to `committer.rs`
- therefore removed all other changes to fd code

mmcgee-jump (approved):

Issue #2746, opened 2024-08-19
Insufficient shred validation


ptaffet-jump:
A malicious leader can create shreds that Firedancer accepts but will cause Rust to panic when inserted into the blockstore. We need to fuzz this aggressively.

PR #2750, opened 2024-08-19
shred: add extra shred validation

ptaffet-jump:
Partially addresses #2746, but more fuzzing needed.


two-heart:
Also fixes https://github.com/firedancer-io/auditor-internal/issues/205

mmcgee-jump (dismissed):

mmcgee-jump (approved):

PR #2759, opened 2024-08-21
quic - rv20 - disallow new conn id in coalesced packets

nbridge-jump:
RFC 9000 recommends skipping QUIC packets in a single datagram when the dst conn id is different than the first in the datagram.


ripatel-fd:
Could you add tests?

PR #2767, opened 2024-08-21
Add vu_permute2 wrapper for shufps instruction

ripatel-fd:


ripatel-fd:
Merged https://github.com/firedancer-io/firedancer/commit/e546b8d49ef2bcf23cd7a8569adfde40da27eb8a

PR #2771, opened 2024-08-21
flamenco, runtime: fix reading epoch context from snapshot

topointon-jump:
Using the snapshot manifest memory doesn't work with the statically allocated epoch context. We instead copy from the statically allocated epoch context when loading from snapshots.


topointon-jump:
Converting this to draft to add support for incremental snapshots

lheeger-jump (approved):

PR #2801, opened 2024-08-27
fdctl: randomize fd_quic_conn_id_hash_seed

ripatel-fd:


src/app/fdctl/run/tiles/fd_quic.c
```diff
@@ -527,6 +527,12 @@ privileged_init( fd_topo_t *      topo,
   uchar const * public_key = fd_keyload_load( tile->quic.identity_key_path, 1 /* public key only */ );
   fd_memcpy( ctx->identity_public_key, public_key, 32UL );
 
+  FD_ONCE_BEGIN {
```
mmcgee-jump, line 530: Why FD_ONCE? There is only one thread

ripatel-fd: @mmcgee-jump In case fdctl is run with 'everything in one process' mode for debugging

ripatel-fd: @mmcgee-jump The difference is that the seed is different on every reboot. It's fine if threads share the same seed

mmcgee-jump: I don't understand the reasoning here. In production the seeds are going to be different for different threads, so development should match that behavior. You're introducing a pointless difference between development and production.

If you're saying the correct behavior is the QUIC tiles share a seed, then we should make this change work multi process too.

I think though, you can just remove the FD_ONCE?

ripatel-fd:
I'm working on a better fix

PR #2805, opened 2024-08-27
util: add list merge APIs

ripatel-fd:


ripatel-fd:
@lidatong said

>  for consistency merge is used for that op in other structures

ripatel-fd:
Not needed for now

PR #2812, opened 2024-08-28
runtime: remove duplicate self.touch() (nit)

two-heart:
just a duplicated statement, that caused no bug


ripatel-fd (changes requested):
This was broken before and it's still broken. `account->meta` at line 43 is not writable because `view_idx` only provides `const_meta`, but not `meta`. 

Line 43 should be deleted, line 52 should stay

ripatel-fd (approved):

ripatel-fd (approved):

PR #2823, opened 2024-08-29
flamenco, fuzz: CPI execution side-effects

ravyu-jump:
This PR does two tangential things:

## 1. Mock CPI execution
With `__wrap_fd_execute`, we can simulate the execution of a CPI by modifying account states with a provided `InstrEffects` message. 

## 2. Dumping account state in CPI calls
To generate seed corpus with execution effects, the CPI dump tool was modified to capture post-CPI instruction account data. 

Also added a script that temporarily applies a patch file that dumps CPI state during CPI syscall.


ravyu-jump:
Wrong branch pushed

PR #2832, opened 2024-09-03
quic: remove tx_buf from conn object

ripatel-fd:
Reduces the size of fd_quic_conn_t by 2kB by removing the datagram
payload buffer.

This buffer does not outlive an fd_quic_conn_tx function call in
the happy path.  Only if there is a transient send failure (due to
XDP rings being full), the tx_buf contents get preserved from one
call to conn_tx to the next.

With this patch, datagram payloads are discarded on transmit failure.
(A retransmit is still attempted since the client won't be able to
generate an ACK)


ripatel-fd:
Stale PR

PR #2833, opened 2024-09-03
quic: simplify ACK generation

ripatel-fd:
- Ignore peer ACKs of our ACKs
- Don't retransmit our ACKs if peer didn't ACK them
- Replace fd_quic_ack_t linked list with a simpler ring buffer
- Limit amount of disjoint ACK ranges generated in response to peer
  packets with gapped packet numbers
- Improve complexity of ACK generation algorithms

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -48,22 +48,6 @@
 #include "../../util/tmpl/fd_map_dynamic.c"
 
 
-/* FD_QUIC_OPT_EXTEND_SENT
- * This is a compile time option to specify whether ack ranges are extended even
- * when they have already been sent. This may have performance implications, but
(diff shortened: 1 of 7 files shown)


src/waltz/quic/fd_quic.c
```diff
@@ -4218,7 +3964,7 @@ fd_quic_conn_tx( fd_quic_t *      quic,
 
     /* did we add any frames? */
 
-    if( !pkt_meta->flags ) {
```
ripatel-fd, line 4221: If we only add ACK frames then pkt_meta->flags==0.
In this case we should probably free the pkt_meta and treat the outgoing ACK as fire-and-forget.

ripatel-fd:
This has been merged in https://github.com/firedancer-io/firedancer/pull/3140

PR #2835, opened 2024-09-03
flamenco, vm: skip pubkeys which span multiple regions

topointon-jump:
User-provided pubkeys should never span multiple regions, these should ~~be skipped.~~ error out


ravyu-jump:
Can't generate a fixture ATM because the crashfile that produced this error produces another error (which needs to be cleaned up) when these fixes are applied. I will take care of that separately. Approving this.

ravyu-jump (dismissed):

topointon-jump:
This PR depends on @ravyu-jump's upcoming CPI error PR

ibhatt-jumptrading (approved):

PR #2846, opened 2024-09-04
quic: fix pkt number gaps and pkt_meta leaks

ripatel-fd:
- Fixes gaps in packet numbers when generating packets.
  This allows the peer to send more optimized ACK frames.

- Fixes an edge case where a pkt_meta object could get leaked.


nbridge-jump:
This looks fine, but execution isn't supposed to go into the main part of `fd_quic_conn_tx` if there is nothing to send.
`fd_quic_tx_enc_level` should only return a valid enc_level if there is data to be sent.

nbridge-jump (approved):

PR #2863, opened 2024-09-05
flamenco: fix account resize delta double add

mjain-jump:
Agave modifies a copy of the variable here, and then increments the actual variable later. We mistakenly perform a double add and incorrectly throw an error


topointon-jump (approved):

jumpsiegel (approved):

jumpsiegel:
another head scratcher

PR #2891, opened 2024-09-09
fix oom crash in rpc

asiegel-jt:


jumpsiegel (approved):

mmcgee-jump:
This isn't an OOM, the hcache is fixed size. What it means is you tried to send a response to an RPC request that was larger than the size of the data region, which should be somewhat difficult to do.

You either need to find the response that was so long and why it was so long, or just increase the size of the buffer if this length of response is expected. It should be several gigabytes, so very strange if this is happening.

asiegel-jt:
> This isn't an OOM, the hcache is fixed size. What it means is you tried to send a response to an RPC request that was larger than the size of the data region, which should be somewhat difficult to do.
> 
> You either need to find the response that was so long and why it was so long, or just increase the size of the buffer if this length of response is expected. It should be several gigabytes, so very strange if this is happening.

regardless, crashing isn't acceptable

mmcgee-jump:
Well, the hcache size you're using looks to be 100MB. Do you expect 100MB responses, if not there's some other bug.

PR #2898, opened 2024-09-09
flamenco: make rent epoch logic conformant with Agave

mjain-jump:
Rent epoch setting logic is now conformant with Agave's logic, with hyperlinks included.


jumpsiegel (approved):

ibhatt-jumptrading (dismissed):
this is great

topointon-jump (approved):

PR #2907, opened 2024-09-10
pack: fix typo that prevented transactions from ever expiring

ptaffet-jump:
Assuming `now` is in `[LONG_MIN, LONG_MAX]` but never wraps around, then `x = (ulong)now+TIME_OFFSET` is in `[0, ULONG_MAX]`. We want to compute `x-transaction_lifetime_ticks`, but saturating at 0 instead of underflowing.  The right way to do that is `max(x, transaction_lifetime_ticks) - transaction_lifetime_ticks`, which is in `[0, ULONG_MAX-transaction_lifetime_ticks]`.


mmcgee-jump (approved):

mmcgee-jump:
Can you pick to v0.1 ?

PR #2912, opened 2024-09-11
store: disable blockstore starting from a certain slot

anwayde:
Note that there are no shreds for slot 0, so setting `disable_blockstore_from_slot = 1` has exactly the same effect as `disable_blockstore = true`.

### DIFF
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -1217,14 +1217,24 @@ dynamic_port_range = "8900-9000"
         # Frankendancer currently depends on the Agave blockstore
         # component for storing block data to disk.  This component can
         # frequently be a bottleneck when testing the throughput of the
-        # leader pipeline, so this option is provided to disable it.
+        # leader pipeline, so this option is provided to disable it
+        # starting from a certain slot.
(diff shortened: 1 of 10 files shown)


src/app/fdctl/config/default.toml
```diff
@@ -1202,14 +1202,19 @@ dynamic_port_range = "8900-9000"
...
+        # starting from a certain slot.
         #
         # This option should not be used in a production network; it
         # causes the validator to not be able to serve repair requests,
         # snapshots, or participate in other consensus critical
         # operations.  It is only useful for benchmarking the leader
         # TPU performance in a single node cluster case.
-        disable_blockstore = false
+        #
+        # A value of 0 means this option will not be used.  Any other
```
mmcgee-jump, line 1214: Could you add some commentary on what the slot number is for (aka. generating a first snapshot in benchmarking)

mmcgee-jump (changes requested):
.

ptaffet-jump:
I think it needs to be set in the firedancer topo as well

mmcgee-jump (approved):

PR #2824, opened 2024-08-29
flamenco, fuzz: CPI execution mocking

ravyu-jump:
## Mock CPI execution
With `__wrap_fd_execute`, we can simulate the execution of a CPI by modifying account states with a provided `InstrEffects` message. 

Separate PR for dumping of account states in replay: https://github.com/firedancer-io/firedancer/pull/2931
~~## 2. Dumping account state in CPI calls~~
~~To generate seed corpus with execution effects, the CPI dump tool was modified to capture post-CPI instruction account data.~~

~~Also added a script that temporarily applies a patch file that dumps CPI state during CPI syscall.~~

PR #2931, opened 2024-09-17
flamenco: dump CPI instruction effects to protobuf

ravyu-jump:
# Dumping account state in CPI calls
To generate seed corpus with execution effects, the CPI dump tool was modified to capture post-CPI instruction account data.

Also added a script that temporarily applies a patch file that dumps CPI state during CPI syscall.

Relevant fuzz PR: https://github.com/firedancer-io/firedancer/pull/2824


mjain-jump:
Implemented already for all syscalls.

PR #2936, opened 2024-09-18
Fix UAF in fd_quic tile

ripatel-fd:
Fixes a assertion failure when closing a conn with streams that
got overrun.

Adds more logging to assertion failures and hardens the
fd_tpu_reasm_cancel function.


mmcgee-jump (dismissed):

mmcgee-jump (approved):

ptaffet-jump (approved):
Forwarding Michael's stamp

PR #2941, opened 2024-09-19
Remove direct mapping support for ABIv1

ripatel-fd:
This PR removes parts of the "direct mapping" implementation that are specific to ABIv1. 

Leaves intact the CPI and serialization routines because those might be reused for ABIv2 when it is ready.

Not yet ready for merge.
This should be merged only after testnet disables the direct mapping feature.

### DIFF
--- a/src/flamenco/vm/fd_vm.h
+++ b/src/flamenco/vm/fd_vm.h
@@ -18,9 +18,9 @@ typedef struct fd_vm fd_vm_t;
 struct fd_vm_shadow { ulong r6; ulong r7; ulong r8; ulong r9; ulong pc; };
 typedef struct fd_vm_shadow fd_vm_shadow_t;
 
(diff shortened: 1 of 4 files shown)


ibhatt-jumptrading:
More work is being done on the agave direct mapping on abi v1, not sure if it makes sense to remove any of this just yet...

ripatel-fd:
> More work is being done on the agave direct mapping on abi v1, not sure if it makes sense to remove any of this just yet...

@ibhatt-jumptrading Agave just did a testnet restart deactivating ABIv1 direct mapping. So direct mapping is off on all clusters. Anza also promised to not trigger a reactivation. But we should probably coordinate ABIv1 DM removal from Firedancer with removal on the Anza side.

ripatel-fd:
Anza changed their mind recently. They are now planning to activate ABIv1

PR #2943, opened 2024-09-20
fixes for explorer on mainnet with rpc

asiegel-jt:


ripatel-fd (approved):

ripatel-fd:
Stamped but didn't review

PR #2956, opened 2024-09-23
pack: update metrics periodically

mmcgee-jump:


ptaffet-jump (approved):
Great! We might even want to remove the metric updates from schedule_next_microblock at some point, but this seems good.

PR #2958, opened 2024-09-23
flamenco, fuzz: fix heap size checks

ravyu-jump:


src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -263,7 +263,7 @@ fd_exec_test_instr_context_create( fd_exec_instr_test_runner_t *        runner,
 
   /* Initial variables */
   txn_ctx->loaded_accounts_data_size_limit = FD_VM_LOADED_ACCOUNTS_DATA_SIZE_LIMIT;
-  txn_ctx->heap_size                       = fd_ulong_max( txn_ctx->heap_size, FD_VM_HEAP_SIZE ); /* FIXME: bound this to FD_VM_HEAP_MAX?*/
+  txn_ctx->heap_size                       = FD_VM_HEAP_MAX;
```
0x0ece, line 266: I think it should be:
```
fd_ulong_min( 
  fd_ulong_max( txn_ctx->heap_size, FD_VM_HEAP_DEFAULT ),
  FD_VM_HEAP_MAX )
```
i.e. by default an ix has 32kB of heap (DEFAULT). If the ix was set with extra heap, we use that. At most 256kB heap (MAX)?

(the other change below looks correct)

ravyu-jump: `txn_ctx->heap_size` isn't initialized, looking at the code above. So the suggestion would return an arbitrary value that we cannot control. This can be a problem with differential fuzzing.

I do see the point that `FD_VM_HEAP_DEFAULT` should be, well, the default value. So I'll change it to that. This line of code has gone through a full circle 🙃

ravyu-jump: We do have a separate mechanism for controlling heap sizes in cases where we are interested in doing so (i.e., VM fuzzers), so I think setting it to default here is fine.

src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -1728,7 +1728,7 @@ fd_exec_vm_syscall_test_run( fd_exec_instr_test_runner_t * runner,
     }
   }
 
-  if (input->vm_ctx.heap_max > FD_VM_HEAP_DEFAULT) {
+  if ( input->vm_ctx.heap_max > FD_VM_HEAP_MAX ) {
```
mjain-jump, line 1731:
```suggestion
  if( input->vm_ctx.heap_max > FD_VM_HEAP_MAX ) {
```

mjain-jump (approved):

PR #2963, opened 2024-09-25
add %64J replacement

two-heart:


src/flamenco/fd_flamenco_base.h
```diff
@@ -22,18 +22,22 @@
...
 
     fd_pubkey_t pk = ... ;
     printf("%s", FD_BASE58_ENCODE_32( pk ) );
 
    The temp buffer is allocated on the stack and therefore invalidated
    when the function this is used in returns. */
-#define FD_BASE58_ENCODE_32(x) (__extension__({            \
-  char * _out = (char *)alloca( FD_BASE58_ENCODED_32_SZ ); \
-  fd_base58_encode_32( (uchar const *)(x), NULL, _out );   \
-  }))
```
ripatel-fd, line 36: I prefer having this than going through a second macro, IMO it makes the behavior a bit more readable

```
#define FD_BASE58_ENCODE_32(x) (__extension__({            \
  char * _out = (char *)alloca( FD_BASE58_ENCODED_64_SZ ); \
  fd_base58_encode_64( (uchar const *)(x), NULL, _out );   \
  }))

#define FD_BASE58_ENCODE_32(x) (__extension__({            \
  char * _out = (char *)alloca( FD_BASE58_ENCODED_64_SZ ); \
  fd_base58_encode_64( (uchar const *)(x), NULL, _out );   \
  }))
```

src/flamenco/fd_flamenco_base.h
```diff
@@ -22,18 +22,22 @@
 */
 #define FD_DEFAULT_AGAVE_CLUSTER_VERSION (2000)
 
-/* FD_BASE58_ENCODE_32 is a shorthand for fd_base58_encode_32, including
-   defining a temp buffer.  Useful for printf-like functions.  Example:
+#define _FD_TO_BASE58( x, bits ) __extension__({               \
```
ripatel-fd, line 25: This should probably be 'sz' not 'bits' since it's a byte size, not a bit count

src/flamenco/fd_flamenco_base.h
ptaffet-jump: Heads up that Richie's comment has some copy-pasting errors. The _32 version of the macro should be
```
#define FD_BASE58_ENCODE_32(x) (__extension__({            \
  char * _out = (char *)alloca( FD_BASE58_ENCODED_32_SZ ); \
  fd_base58_encode_32( (uchar const *)(x), NULL, _out );   \
  }))
```

ripatel-fd (approved):

PR #2964, opened 2024-09-25
quic: don't retransmit ACKs

ripatel-fd:
Instead of storing sent ACKs in pkt_meta, now immediately reclaims
ack objects in conn_tx.  This reduces pressure on the ack pool for
fd_quic tiles.


mmcgee-jump (approved):

ripatel-fd:
Closing this since it's already implemented by https://github.com/firedancer-io/firedancer/pull/3140

PR #2966, opened 2024-09-25
quic: fix uninitialized read in fuzz_quic_wire

ripatel-fd:
fd_quic_decode_one_rtt(one_rtt) requires one_rtt->dst_conn_id_len
to be initialized.  The fuzzer fails to do this.  Production code
in fd_quic.c already uses fd_quic_decode_one_rtt correctly.

Thanks to ImmuneFi User gln for reporting this bug.


two-heart (commented):
lgtm

mmcgee-jump (approved):

PR #2972, opened 2024-09-26
Fix OOB read in test_fec_resolver

ripatel-fd:
Also add MSan support


ripatel-fd:
Depends on #2971

ripatel-fd:
CI checks appear to be stuck.

PR #2973, opened 2024-09-26
Fix various MSan false positives and buggy unit tests

ripatel-fd:


mmcgee-jump (dismissed):

ptaffet-jump:
Can you try with just this diff for `test_bmtree.c`?


```
diff --git a/src/ballet/bmtree/test_bmtree.c b/src/ballet/bmtree/test_bmtree.c
index 90576fee6..4325e8e91 100644
--- a/src/ballet/bmtree/test_bmtree.c
+++ b/src/ballet/bmtree/test_bmtree.c
@@ -84,6 +84,11 @@ test_inclusion( ulong leaf_cnt ) {

   fd_bmtree_node_t leaf[1];
   fd_memset( leaf->hash, 0, 20UL );
+  /* In the leaf_cnt==1 case, the root is just the full 32-byte hash of
+     the payload, so sanitizers may complain if the last bytes are
+     uninitialized.  In the leaf_cnt>1 case, we want to make sure we
+     never use more than the first 20 bytes of the hash. */
+  if( FD_UNLIKELY( leaf_cnt==1UL ) ) fd_memset( leaf->hash+20UL, 'X', 32UL-20UL );
   for( ulong i=0UL; i<leaf_cnt; i++ ) {
     FD_TEST( fd_bmtree_commit_leaf_cnt( tree )==i );
     FD_STORE( ulong, leaf->hash, i );
```

ripatel-fd:
@ptaffet-jump I reverted the bmtree change for now.

nlgripto (approved):

PR #2976, opened 2024-09-27
bm_tree: fix cmp with uninitialized memory in tests

two-heart:
Otherwise, this will trip off msan (now part of CI)


ripatel-fd:
Duplicate of https://github.com/firedancer-io/firedancer/pull/2973

PR #2980, opened 2024-09-28
util: fix map_dynamic MSan false positive

ripatel-fd:
map_dynamic with large keys, memoize==1, key_equal_is_slow==1
relies on speculative reads of the MAP_HASH, which can be
uninitialized in some cases.

### DIFF
--- a/src/util/tmpl/fd_map_dynamic.c
+++ b/src/util/tmpl/fd_map_dynamic.c
@@ -337,8 +337,16 @@ MAP_(new)( void *  shmem,
 
   MAP_T * slot = map->slot; FD_COMPILER_FORGET( slot );
 
-  for( ulong slot_idx=0UL; slot_idx<slot_cnt; slot_idx++ )
+  for( ulong slot_idx=0UL; slot_idx<slot_cnt; slot_idx++ ) {
     slot[ slot_idx ].MAP_KEY = (MAP_KEY_NULL);
+
+  # if MAP_MEMOIZE && MAP_KEY_EQUAL_IS_SLOW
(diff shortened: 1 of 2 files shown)


lheeger-jump (approved):

lheeger-jump:
do you want to guard the poisin with HAS_ASAN?

ripatel-fd:
> do you want to guard the poisin with HAS_ASAN?

@lheeger-jump Poison ops are no-op if MSan / ASan are not available. 
So no, I'd rather keep the code as concise as possible.

src/util/tmpl/fd_map_dynamic.c
```diff
@@ -337,9 +337,17 @@ MAP_(new)( void *  shmem,
...
   MAP_T * slot = map->slot; FD_COMPILER_FORGET( slot );
 
-  for( ulong slot_idx=0UL; slot_idx<slot_cnt; slot_idx++ )
+  for( ulong slot_idx=0UL; slot_idx<slot_cnt; slot_idx++ ) {
     slot[ slot_idx ].MAP_KEY = (MAP_KEY_NULL);
 
+  # if MAP_MEMOIZE && MAP_KEY_EQUAL_IS_SLOW
+    /* Allow speculative reads of uninitialized MAP_HASH value. */
+    fd_msan_unpoison( &slot[ slot_idx ].MAP_HASH, sizeof(MAP_HASH_T) );
+    slot[ slot_idx ].MAP_HASH = 0;
```
ripatel-fd, line 346: Line 346 was an accident

PR #2983, opened 2024-09-28
util: add MSan support for wksp and scratch

ripatel-fd:


ripatel-fd:
Leaving as draft for now because the slowdown appears to be excessive

PR #2984, opened 2024-09-30
remove X.509 parser corpus

two-heart:
PR #1269 removed the X.509 parser and fuzzer, therefore we do not need the fuzz corpus anymore


ripatel-fd:
We still have an X.509 parser. It's just 100x simpler.
https://github.com/firedancer-io/firedancer/blob/00e259f3ecb03bce2b6fa84f4aa41216122cf71c/src/ballet/x509/fd_x509_mock.c#L163

We have a CBMC proof for correctness here, so we probably don't need a fuzzer anymore: https://github.com/firedancer-io/firedancer/blob/main/verification/proofs/x509_mock/fd_x509_mock_harness.c

ripatel-fd (approved):

PR #2988, opened 2024-09-30
runtime: remove "touch" verbiage

ripatel-fd:
BorrowedAccount::touch() is a no-op in recent Agave versions.
Update Firedancer code to be less confusing about the behavior of
touch by updating comments and removing redundant statements.


ripatel-fd:
Bank hash mismatch... I guess touch does something after all

PR #2992, opened 2024-09-30
Don't bleed printf pragmas from headers

ripatel-fd:


two-heart:
Not required when https://github.com/firedancer-io/firedancer/pull/2993/ is merged

ripatel-fd:
Replaced by #2993

PR #2998, opened 2024-10-01
sha256: msan support

two-heart:
prevents a msan false positive in fuzz_tls


mmcgee-jump (dismissed):

jumpsiegel (dismissed):

ripatel-fd (changes requested):
fd_msan wrappers should stay static inlines, not extern inlines

PR #3004, opened 2024-10-01
sign: add support for block engine challenge

mmcgee-jump:


ptaffet-jump (approved):
We need to ask about clarifying the spec to make sure this is always correct, but seems right as far as I know

PR #3009, opened 2024-10-01
http: improve server performance

mmcgee-jump:


ptaffet-jump (approved):
Stamping to unblock since Yufeng is out. I didn't read it carefully

PR #3010, opened 2024-10-01
flamenco: set error kind properly for instructions

mjain-jump:
VM and instruction error codes were being mixed up due to error kinds not being set correctly for instructions.

### DIFF
--- a/src/flamenco/runtime/context/fd_exec_instr_ctx.h
+++ b/src/flamenco/runtime/context/fd_exec_instr_ctx.h
@@ -40,6 +40,11 @@ struct __attribute__((aligned(8UL))) fd_exec_instr_ctx {
 
 FD_PROTOTYPES_BEGIN
 
+#define FD_INSTR_ERR_FOR_LOG_INSTR( instr_ctx, err ) (__extension__({ \
+    instr_ctx->txn_ctx->exec_err = err;                               \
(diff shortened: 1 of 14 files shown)


src/flamenco/runtime/context/fd_exec_instr_ctx.h
```diff
@@ -40,6 +40,11 @@ struct __attribute__((aligned(8UL))) fd_exec_instr_ctx {
 
 FD_PROTOTYPES_BEGIN
 
+#define FD_INSTR_ERR_FOR_LOG_INSTR( instr_ctx, err ) (__extension__({ \
```
0x0ece, line 43: There is already FD_VM_ERR_FOR_LOG_INSTR

src/flamenco/runtime/fd_executor.c
```diff
@@ -919,15 +922,19 @@ fd_execute_instr( fd_exec_txn_ctx_t * txn_ctx,
       /* Log success */
       fd_log_collector_program_success( ctx );
     } else {
-      if( !txn_ctx->exec_err ) {
-        FD_TXN_ERR_FOR_LOG_INSTR( txn_ctx, exec_result, txn_ctx->instr_err_idx );
```
0x0ece, line 923: Why removing this? This should add the error for all native programs instead of doing it for every program.

PR #3019, opened 2024-10-03
metric: generate more metrics code, use new http server

mmcgee-jump:


src/app/fdctl/run/tiles/metric.seccomppolicy
```diff
@@ -1,32 +1,69 @@
...
+close: (not (or (eq (arg 0) 2)
+                (eq (arg 0) logfile_fd)
+                (eq (arg 0) metrics_socket_fd)))
 
-# server: serving metric values over HTTP requires polling conns
-poll
+# server: serving pages over HTTP requires polling connections
+#
+# arg 2 is the timeout.
+poll: (eq (arg 2) 0)
```
ptaffet-jump, line 69: Do you want to do a similar filter on the file descriptor as for `sendto`, etc. here?

mmcgee-jump: I don't think we can with the current filter implementation, since you provide a list of `pollfd` structs. See `man 2 poll`

ptaffet-jump: Oh good point. I forgot what the syscall looked like.

src/disco/metrics/fd_prometheus.c
```diff
@@ -0,0 +1,147 @@
...
+                   ulong                     metrics_cnt,
+                   fd_metrics_meta_t const * metrics,
+                   int                       print_mode ) {
+  for( ulong i=0UL; i<metrics_cnt; i++ ) {
+    fd_metrics_meta_t const * metric = &metrics[ i ];
+    fd_http_server_printf( http, "# HELP %s %s\n# TYPE %s %s\n", metric->name, metric->desc, metric->name, fd_metrics_meta_type_str( metric ) );
+
+    for( ulong j=0UL; j<topo->tile_cnt; j++ ) {
+      fd_topo_tile_t const * tile = &topo->tiles[ j ];
+      if( FD_LIKELY( tile_name!=NULL && strcmp( tile->name, tile_name ) ) ) continue;
```
ptaffet-jump, line 53: No need to fix it now, but could you add a `FIXME` that this makes printing each with `PRINT_TILE` O(n^2) instead of O(n)?

src/app/fdctl/run/tiles/fd_metric.c
```diff
@@ -458,20 +123,11 @@ unprivileged_init( fd_topo_t *      topo,
...
-
-  ctx->fds[ MAX_CONNS ].fd = ctx->socket_fd;
-  ctx->fds[ MAX_CONNS ].events = POLLIN | POLLOUT;
-
   ulong scratch_top = FD_SCRATCH_ALLOC_FINI( l, 1UL );
   if( FD_UNLIKELY( scratch_top > (ulong)scratch + scratch_footprint( tile ) ) )
     FD_LOG_ERR(( "scratch overflow %lu %lu %lu", scratch_top - (ulong)scratch - scratch_footprint( tile ), scratch_top, (ulong)scratch + scratch_footprint( tile ) ));
 
-  FD_LOG_NOTICE(( "Prometheus metrics endpoint listening on port %u", tile->metric.prometheus_listen_port ));
+  FD_LOG_WARNING(( "Prometheus metrics endpoint listening at http://0.0.0.0:%u/metrics", tile->metric.prometheus_listen_port ));
```
ptaffet-jump, line 130: I prefer notice over warning for this, because it's not a bad thing. Was the thought to make sure it was firewalled appropriately?

ptaffet-jump (dismissed):

src/app/fdctl/run/tiles/fd_metric.c
mmcgee-jump: I think it's good to highlight these lines for now, since we have a good amount of boot spam. When I clean that spam up we can downgrade back to notice.

src/disco/metrics/gen_metrics.py
```diff
@@ -134,8 +137,12 @@ def parse_metrics(xml_data):
     (enums, metrics) = parse_metrics(xml_data)
     os.makedirs('generated', exist_ok=True)  # Ensure the directory exists
 
+    tiles = sorted(list(set([x.tile for x in metrics if x.tile != 'all'])))
+    agave_tiles = sorted(list(set([x.tile for x in metrics if x.tile != 'all' and not x.full])))
+    non_agave_tiles = sorted(list(set([x.tile for x in metrics if x.tile != 'all' and x.full])))
```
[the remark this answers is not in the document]
mmcgee-jump: Will resolve in a future change that rewrites this script.

ptaffet-jump (approved):
Fantastic! It always took me two tries to add metrics for a new tile, but this makes it a lot easier

PR #3020, opened 2024-10-03
fdctl: tidy up keys command

mmcgee-jump:


src/app/fdctl/keys.c
```diff
@@ -46,72 +47,76 @@ keys_cmd_args( int *    pargc,
...
+void FD_FN_SENSITIVE
 generate_keypair( char const *     keyfile,
                   config_t * const config,
                   int              use_grnd_random ) {
-  uchar keys[ 64 ];
-
-  uint flags = use_grnd_random ? GRND_RANDOM : 0;
+  uint flags = use_grnd_random ? GRND_RANDOM : 0U;
 
+  uchar private_key[ 64 ];
```
ptaffet-jump, line 56: This should be called keypair if it's 64 bytes and private key if it's 32 bytes

src/app/fdctl/keys.c
```diff
@@ -46,72 +47,76 @@ keys_cmd_args( int *    pargc,
...
   long bytes_produced = 0L;
   while( FD_LIKELY( bytes_produced<32 ) ) {
```
ptaffet-jump, line 58:
```suggestion
  while( FD_LIKELY( bytes_produced<32L ) ) {
```

src/app/fdctl/keys.c
```diff
@@ -46,72 +47,76 @@ keys_cmd_args( int *    pargc,
...
-      FD_LOG_ERR(( "could not create keypair as the keyfile `%s` already exists", keyfile ));
-    else
-      FD_LOG_ERR(( "could not create keypair, fopen(%s) failed (%i-%s)", keyfile, errno, fd_io_strerror( errno ) ));
+  if( FD_LIKELY( !gid && setegid( config->gid ) ) ) FD_LOG_ERR(( "setegid() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+  if( FD_LIKELY( !uid && seteuid( config->uid ) ) ) FD_LOG_ERR(( "seteuid() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+
+  /* find last `/` in keyfile and zero it */
+  char * last_slash = strrchr( keyfile, '/' );
+  if( FD_LIKELY( last_slash ) ) {
+    *last_slash = '\0';
```
ptaffet-jump, line 79: `keyfile` is a const pointer. Modifying it is a bit sketchy, but I don't know if there's a better way. Copying it is a bit of a pain.

ptaffet-jump (commented):
Some nits, but overall I like the idea of getting `stdio` out of the way for sensitive data

src/app/fdctl/keys.c
```diff
@@ -46,72 +47,77 @@ keys_cmd_args( int *    pargc,
...
-    if( FD_LIKELY( errno == EEXIST ) )
-      FD_LOG_ERR(( "could not create keypair as the keyfile `%s` already exists", keyfile ));
-    else
-      FD_LOG_ERR(( "could not create keypair, fopen(%s) failed (%i-%s)", keyfile, errno, fd_io_strerror( errno ) ));
+  if( FD_LIKELY( !gid && setegid( config->gid ) ) ) FD_LOG_ERR(( "setegid() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+  if( FD_LIKELY( !uid && seteuid( config->uid ) ) ) FD_LOG_ERR(( "seteuid() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+
+  /* find last `/` in keyfile and zero it */
+  char keyfile_copy[ PATH_MAX ];
+  fd_memcpy( keyfile_copy, keyfile, sizeof( keyfile_copy ) );
```
ptaffet-jump, line 78: I think `strncpy` be better so that it doesn't overrun `keyfile`, though that's not possible with current uses, so I'll stamp. I think really what this is suggesting is that the type of `keyfile` should be `char const keyfile[ static PATH_MAX ]` or something. I think just using strncpy is easier.

ptaffet-jump (dismissed):

ptaffet-jump (approved):

PR #3022, opened 2024-10-03
agave: remove unneeded RUST_LOG config

mmcgee-jump:


ptaffet-jump (commented):
Could you keep a comment with the line? I would sometimes adjust it, to e.g. print trace level information about a component, and it was nice to remember how to set it

mmcgee-jump:
These directives don't work anymore since we don't use env_logger.

ptaffet-jump (approved):

PR #3023, opened 2024-10-03
poh: silence noisy init message

mmcgee-jump:


ptaffet-jump (approved):
Great! I think this confused people sometimes.

PR #3032, opened 2024-10-03
fdctl: allow reliable consumer from non-primray out

mmcgee-jump:


lheeger-jump (commented):
Does this actually do reliable consuming though? You have to manage it yourself still?

mmcgee-jump:
The producer will need to implement flow control before publishing, since the mux won't do it automatically, but otherwise it works fine.

ptaffet-jump (approved):

PR #3045, opened 2024-10-06
build: link libsecp256k1.a

cavemanloverboy:
using the secp256k1 extra does not add the linker flag `LDFLAGS+=$(OPT)/lib/libsecp256k1.a` required for linking

### DIFF
--- a/config/extra/with-secp256k1.mk
+++ b/config/extra/with-secp256k1.mk
@@ -1,6 +1,7 @@
 ifneq (,$(wildcard $(OPT)/lib/libsecp256k1.a))
 FD_HAS_SECP256K1:=1
 CFLAGS+=-DFD_HAS_SECP256K1=1
+LDFLAGS+=$(OPT)/lib/libsecp256k1.a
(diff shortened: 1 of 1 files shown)


lheeger-jump:
Which targets were you having issues with? We dont want to link in secp256k1 into every target

cavemanloverboy:
> Which targets were you having issues with? We dont want to link in secp256k1 into every target

I wrote some tests in `test_tower.c`.

As for this PR, I was just following what is done for the other extras:

https://github.com/firedancer-io/firedancer/blob/aebc6dac237e6ac5dc5af46383318484e45f8840/config/extra/with-zstd.mk#L4

The extra `.mk` file is only included when specifying `EXTRAS=secp256k1`

ripatel-fd:
The proper way to link libsecp256k1 is like so:

```
$(call make-unit-test,test_secp256k1,test_secp256k1,fd_ballet fd_util,$(SECP256K1_LIBS))
```

(The fourth argument of a make-bin or make-unit-test directive)

PR #3050, opened 2024-10-07
turn on MSAN for ci builds

nlgripto:


ripatel-fd (changes requested):
msan requires a deps.sh installation with the +msan flag
Needs more CI plumbing to call `./deps.sh +msan` and cache the `opt-msan` prefix.

PR #3053, opened 2024-10-07
gui: implement simple GUI backend

mmcgee-jump:


book/api/websocket.md
```diff
@@ -0,0 +1,1093 @@
...
+average from the prior 750 slots, or around five minutes. Live here
+means the estimate is republished whenever it changes, which is when
+a new slot is confirmed on the currently active fork.
+
+#### `summary.skip_rate`
+| frequency       | type       | example                                 |
+|-----------------|------------|-----------------------------------------|
+| *Once* + *Live* | `SkipRate` | `{"epoch": 522, "skip_rate": 0.456172}` |
+
+The skip rate of an epoch is the ratio of `skipped_slots/total_slots`
```
ptaffet-jump, line 415: What does it publish if `total_slots==0`?

mmcgee-jump:
TODO: add comment to config about needing the account index thing to show validator names / icons

book/api/websocket.md
```diff
@@ -0,0 +1,1093 @@
...
+                "quic_overrun": 45,
+                "quic_quic_invalid": 12,
+                "quic_udp_invalid": 13,
+                "verify_overrun": 2059,
+                "verify_parse": 14,
+                "verify_failed": 4092,
+                "verify_duplicate": 128,
+                "dedup_duplicate": 87,
+                "pack_invalid": 14,
+                "pack_retained": 1985,
```
ptaffet-jump, line 501: Do we need `pack_expired`?

book/api/websocket.md
```diff
@@ -0,0 +1,1093 @@
...
+}
+```
+
+:::
+
+**`LiveTilePrimaryMetric`**
+| Field               | Type                | Description
+|---------------------|---------------------|------------
+| next_leader_slot    | `number\|null`      | The next leader slot |
+| tile_primary_metric | `TilePrimaryMetric` | Per-tile-type primary metrics since the end of the previous leader slot |
```
ptaffet-jump, line 587: These are averaged since the end of the previous leader slot?

src/app/fdctl/Local.mk
```diff
@@ -169,6 +173,70 @@ $(OBJDIR)/bin/agave-ledger-tool: agave/target/$(RUST_PROFILE)/agave-ledger-tool
...
+	for file in $$(find src/app/fdctl/dist -type f); do \
+		counter=$$((counter + 1)); \
+	done; \
+	echo "" >> src/app/fdctl/run/tiles/generated/http_import_dist.h; \
+	echo "struct fd_http_static_file {" >> src/app/fdctl/run/tiles/generated/http_import_dist.h; \
+	echo "    const char * name;" >> src/app/fdctl/run/tiles/generated/http_import_dist.h; \
+	echo "    const uchar * data;" >> src/app/fdctl/run/tiles/generated/http_import_dist.h; \
+	echo "    ulong const * data_len;" >> src/app/fdctl/run/tiles/generated/http_import_dist.h; \
+	echo "    const uchar * zstd_data;" >> src/app/fdctl/run/tiles/generated/http_import_dist.h; \
+	echo "    ulong const * zstd_data_len;" >> src/app/fdctl/run/tiles/generated/http_import_dist.h; \
```
ptaffet-jump, line 198: I think the standard is `type const *`

src/app/fdctl/Local.mk
```diff
@@ -169,6 +173,70 @@ $(OBJDIR)/bin/agave-ledger-tool: agave/target/$(RUST_PROFILE)/agave-ledger-tool
...
+			echo "        .zstd_data = file$${counter}_zstd," >> src/app/fdctl/run/tiles/generated/http_import_dist.c; \
+			echo "        .zstd_data_len = &file$${counter}_zstd_sz" >> src/app/fdctl/run/tiles/generated/http_import_dist.c; \
+		else \
+			echo "        .zstd_data = NULL," >> src/app/fdctl/run/tiles/generated/http_import_dist.c; \
+			echo "        .zstd_data_len = 0UL" >> src/app/fdctl/run/tiles/generated/http_import_dist.c; \
+		fi; \
+		echo "    }," >> src/app/fdctl/run/tiles/generated/http_import_dist.c; \
+		counter=$$((counter + 1)); \
+	done; \
+	echo "};" >> src/app/fdctl/run/tiles/generated/http_import_dist.c; \
```
ptaffet-jump, line 238: Consider a bash script that does all this. Seems kind of gross here.

src/app/fdctl/run/tiles/fd_gui.c
```diff
@@ -0,0 +1,342 @@
...
+  .max_ws_recv_frame_len = FD_HTTP_SERVER_GUI_MAX_WS_RECV_FRAME_LEN,
+  .max_ws_send_frame_cnt = FD_HTTP_SERVER_GUI_MAX_WS_SEND_FRAME_CNT,
+  .outgoing_buffer_sz    = FD_HTTP_SERVER_GUI_OUTGOING_BUFFER_SZ,
+};
+
+typedef struct {
+  fd_topo_t * topo;
+
+  fd_gui_t * gui;
+  uchar      buf[ 8UL+40200UL*(58UL+12UL*34UL) ] __attribute__((aligned(8)));
```
ptaffet-jump, line 54: A comment about the size of this buffer would be helpful

src/app/fdctl/run/tiles/fd_gui.c
```diff
@@ -0,0 +1,342 @@
...
+  uchar * src = (uchar *)fd_chunk_to_laddr( ctx->in_mem, chunk );
+
+   /* ... todo... sigh, sz is not correct since it's too big */
+  if( sig==FD_PLUGIN_MSG_GOSSIP_UPDATE || sig==FD_PLUGIN_MSG_VOTE_ACCOUNT_UPDATE || sig==FD_PLUGIN_MSG_VALIDATOR_INFO ) sz = 8UL + 40200UL*(58UL+12UL*34UL);
+  else if( sig==FD_PLUGIN_MSG_LEADER_SCHEDULE ) sz = 40UL + 40200UL*40UL;
+
+  if( FD_UNLIKELY( chunk<ctx->in_chunk0 || chunk>ctx->in_wmark || sz>sizeof( ctx->buf ) ) )
+    FD_LOG_ERR(( "chunk %lu %lu corrupt, not in range [%lu,%lu]", chunk, sz, ctx->in_chunk0, ctx->in_wmark ));
+
+  fd_memcpy( ctx->buf, src, sz );
```
ptaffet-jump, line 122: FYI, this is a 20+MB memcpy!

src/app/fdctl/run/tiles/fd_gui.c
```diff
@@ -0,0 +1,342 @@
...
+  (void)in_idx;
+  (void)seq;
+  (void)opt_chunk;
+  (void)opt_tsorig;
+  (void)opt_filter;
+  (void)mux;
+
+  fd_gui_ctx_t * ctx = (fd_gui_ctx_t *)_ctx;
+
+  fd_gui_plugin_message( ctx->gui, *opt_sig, ctx->buf, *opt_sz );
```
ptaffet-jump, line 144: Is `*opt_sz` accurate here? From `during_frag` it seems like no. The value computed in `during_frag` isn't passed back to the mux.

src/app/fdctl/run/tiles/fd_gui.c
```diff
@@ -0,0 +1,342 @@
...
+}
+
+static fd_http_server_response_t
+gui_http_request( fd_http_server_request_t const * request ) {
+  if( FD_UNLIKELY( request->method!=FD_HTTP_SERVER_METHOD_GET ) ) {
+    return (fd_http_server_response_t){
+      .status            = 400,
```
ptaffet-jump, line 151: There's a special `HTTP 405 Method not allowed` which you might want to use here

src/app/fdctl/run/tiles/fd_gui.c
```diff
@@ -0,0 +1,342 @@
...
+      .upgrade_websocket = 0,
+    };
+  }
+
+  for( ulong i=0UL; i<sizeof(STATIC_FILES)/sizeof(STATIC_FILES[0]); i++ ) {
+    if( !strcmp( request->path, STATIC_FILES[ i ].name ) ||
+        (!strcmp( STATIC_FILES[ i ].name, "/index.html" ) &&
+          (!strcmp( request->path, "/" ) ||
+           !strcmp( request->path, "/leaderSchedule" ) ||
+           !strcmp( request->path, "/gossip" ))) ) {
```
ptaffet-jump, line 175: I think it's kind of strange to have these checks inside the loop, since they're loop invariant

src/app/fdctl/run/tiles/fd_gui.c
```diff
@@ -0,0 +1,342 @@
...
+
+      char const * content_type = NULL;
+
+      char const * ext = strrchr( STATIC_FILES[ i ].name, '.' );
+      if( FD_LIKELY( ext ) ) {
+        if( !strcmp( ext, ".html" ) ) content_type = "text/html; charset=utf-8";
```
ptaffet-jump, line 181: Why not put this in the struct generated by `frontend`?

src/app/fdctl/run/tiles/fd_gui.c
ptaffet-jump, line 175: I found this confusing initially. I recommend hoisting the checks outside the loop, like 
`int is_page = !strcmp(...) || ...`
Not sure if `is_page` is the right name for the variable, but you get the idea.

src/app/fdctl/run/tiles/fd_gui.c
```diff
@@ -0,0 +1,342 @@
...
+        else if( !strcmp( ext, ".css" ) ) content_type = "text/css";
+        else if( !strcmp( ext, ".js" ) ) content_type = "application/javascript";
+        else if( !strcmp( ext, ".svg" ) ) content_type = "image/svg+xml";
+        else if( !strcmp( ext, ".woff" ) ) content_type = "font/woff";
+        else if( !strcmp( ext, ".woff2" ) ) content_type = "font/woff2";
+      }
+
+      char const * cache_control = NULL;
+      if( FD_LIKELY( !strncmp( request->path, "/assets", 7 ) ) ) cache_control = "public, max-age=31536000, immutable";
```
ptaffet-jump, line 190: I think `public` is unnecessary here because there's no authentication, but it looks like it won't hurt

src/app/fdctl/run/tiles/fd_plugin.c
```diff
@@ -0,0 +1,202 @@
...
+  ulong * dst = (ulong *)fd_chunk_to_laddr( ctx->out_mem, ctx->out_chunk );
+
+  ulong _sz = sz;
+  if( FD_UNLIKELY( in_idx==1UL ) ) _sz = 8UL + 40200UL*(58UL+12UL*34UL);
+  else if( FD_UNLIKELY( in_idx==2UL ) ) _sz = 40UL + 40200UL*40UL; /* ... todo... sigh, sz is not correct since it's too big */
+
+  if( FD_UNLIKELY( chunk<ctx->in[ in_idx ].chunk0 || chunk>ctx->in[ in_idx ].wmark || _sz>ctx->in[ in_idx ].mtu ) )
+    FD_LOG_ERR(( "chunk %lu %lu corrupt, not in range [%lu,%lu]", chunk, sz, ctx->in[ in_idx ].chunk0, ctx->in[ in_idx ].wmark ));
+
+  fd_memcpy( dst, src, _sz );
```
ptaffet-jump, line 66: Same comment as above. This is a huge memcpy that's going to create a decent amount of cache traffic, and it seems a bit unnecessary.

src/app/fdctl/run/tiles/fd_poh.c
```diff
@@ -526,6 +529,107 @@ static fd_poh_ctx_t * fd_poh_global_ctx;
...
+    FD_SPIN_PAUSE();
+  }
+}
+
+static void
+poh_link_publish( poh_link_t *  link,
+                  ulong         sig,
+                  uchar const * data,
+                  ulong         data_len ) {
+  while( FD_UNLIKELY( !FD_VOLATILE_CONST( link->mcache ) ) ) FD_SPIN_PAUSE();
```
ptaffet-jump, line 588: I know this is not new code, but is it safe to assume that when the tile boots, `link->mcache` will be `NULL`? Will it not be uninitialized memory in a workspace?

src/app/fdctl/run/tiles/fd_poh.c
```diff
@@ -962,6 +1069,39 @@ fd_ext_poh_reset( ulong         completed_bank_slot, /* The slot that successful
...
+      FD_TEST( leader_before_reset );
+    } else {
+      /* 2. We are reset onto a different slot. If we were leader
+            before, we should first end that slot, then begin the new
+            one if we are newly leader now. */
+      if( FD_LIKELY( leader_before_reset ) ) {
+        uchar data[ 16 ];
+        FD_STORE( ulong, data, slot_before_reset );
+        FD_STORE( ulong, data+8UL, ctx->cus_used );
+        poh_link_publish( &poh_plugin, FD_PLUGIN_MSG_SLOT_END, data, 16UL );
```
ptaffet-jump, line 1088: I think something like
```suggestion
        struct {
          ulong slot;
          ulong cus;
          } end = { slot_before_reset, ctx->cus_used };
          poh_link_publish( &poh_plugin, FD_PLUGIN_MSG_SLOT_END, (char const *)&end, sizeof(end) );
```


```suggestion
        uchar data[ 16 ];
        FD_STORE( ulong, data, slot_before_reset );
        FD_STORE( ulong, data+8UL, ctx->cus_used );
        poh_link_publish( &poh_plugin, FD_PLUGIN_MSG_SLOT_END, data, 16UL );
```

ptaffet-jump: But I guess perhaps that's a violation of strict aliasing? So then the alternative would be 
`union {
  uchar bytes[16];
  struct {
    ulong slot;
    ulong cus;
  } v;
} end = { .v = { slot_before_reset, ctx->cus_used } };
poh_link_publish( &poh_plugin, FD_PLUGIN_MSG_SLOT_END, end.bytes, sizeof(end) );`
It's not clear to me this is less gross, but I'll leave it for your consideration

src/app/fdctl/run/tiles/fd_poh.c
```diff
@@ -526,6 +529,107 @@ static fd_poh_ctx_t * fd_poh_global_ctx;
...
+poh_link_t gossip_plugin;
+poh_link_t start_progress_plugin;
+poh_link_t vote_listener_plugin;
+poh_link_t poh_plugin;
+
+static void
+poh_link_wait_credit( poh_link_t * link ) {
+  if( FD_LIKELY( link->cr_avail ) ) return;
+
+  while( 1 ) {
```
ptaffet-jump, line 567: This spin loop scares me, but as we've discussed, there's not an easy solution. I'm still inclined to make the plugin links unreliable, and then if the plugin gets overrun, disable it the plugin until the next restart with a message that it was too slow

src/app/fdctl/run/tiles/gui.seccomppolicy
```diff
@@ -0,0 +1,69 @@
...
+# clients, which are the log file, STDOUT, and the listening socket
+# itself.
+close: (not (or (eq (arg 0) 2)
+                (eq (arg 0) logfile_fd)
+                (eq (arg 0) gui_socket_fd)))
+
+# server: serving pages over HTTP requires polling connections
+#
+# arg 2 is the timeout.
+poll: (or (eq (arg 2) 0))
```
ptaffet-jump, line 69: The `or` here is redundant but doesn't cause any problems

src/app/fddev/dev.c
```diff
@@ -119,6 +119,12 @@ update_config_for_dev( config_t * const config ) {
                                    NULL,
                                    "%s/vote-account.json",
                                    config->scratch_directory ) );
+
+  fd_topo_tile_t * gui = &config->topo.tiles[ fd_topo_find_tile( &config->topo, "gui", 0UL ) ];
```
ptaffet-jump, line 123: Does this need a check to make sure it found it? I think it's possible to run `fddev` with the GUI disabled in the TOML, right?

src/disco/gui/fd_gui_printf.h
```diff
@@ -0,0 +1,91 @@
+#include "fd_gui.h"
+
```
ptaffet-jump, line 2: Could you include a doc comment about these? You don't need an individual comment for each function. E.g. all these that take no parameters could be grouped

book/api/websocket.md
yufeng-jump: 0.  Will clarify.

book/api/websocket.md
yufeng-jump: Some are.  Others are more point-in-time values in nature.  Will clarify.

src/app/fdctl/run/tiles/fd_pack.c
```diff
@@ -287,7 +287,7 @@ insert_from_extra( fd_pack_ctx_t * ctx ) {
   extra_txn_deq_remove_head( ctx->extra_txn_deq );
 
   /* Unpack the time stashed in the CUs field */
-  long receive_time = (long)((((ulong)insert->executed_cus)<<32) | ((ulong)insert->requested_cus));
+  long receive_time = (long)((((ulong)insert->pack_cu.non_execution_cus)<<32) | ((ulong)insert->pack_cu.requested_execution_cus));
```
ptaffet-jump, line 290: We should probably just add another field to the union for this use, which would make this code much less gross anyway

book/api/metrics-generated.md
```diff
@@ -105,6 +105,7 @@
 ## Dedup Tile
 | Metric | Type | Description |
 |--------|------|-------------|
+| dedup_&#8203;transaction_&#8203;dedup_&#8203;failure | `counter` | Count of transactions that failed to deduplicate in the dedup stage |
```
mmcgee-jump, line 108: TODO: Fix for still republishing all the GUI nodes every update

src/app/fdctl/run/tiles/fd_poh.c
mmcgee-jump: This is fixed now for the poh_plugin links.

src/app/fdctl/run/tiles/fd_poh.c
mmcgee-jump: These are in bss, so they are zero initialized

src/app/fdctl/Local.mk
mmcgee-jump: Can be fixed in a later change

src/app/fdctl/run/tiles/fd_gui.c
mmcgee-jump: We might want to dynamically figure this out? Not sure. This build stuff can be cleaned up later.

src/disco/gui/fd_gui_printf.c
```diff
@@ -0,0 +1,1172 @@
...
+static void 
+jsonp_double( fd_gui_t *   gui,
+              char const * key,
+              double       value ) {
+  if( FD_LIKELY( key ) ) fd_http_server_printf( gui->http, "\"%s\":%.2f,", key, value );
+  else                   fd_http_server_printf( gui->http, "%.2f,", value );
+}
+
+static void
+jsonp_sanitize_str( fd_http_server_t * http,
```
ptaffet-jump, line 74: I spent some time thinking about this code's behavior on unicode strings, and I think it's fine. All bytes that are part of a non-ASCII character have the high bit set, so this will never replace a byte in the middle of a multi-byte codepoint.

src/disco/gui/fd_gui_printf.c
```diff
@@ -0,0 +1,1172 @@
...
+    if( FD_UNLIKELY( data[ (http->stage_off%http->oring_sz)+i ] < 0x20 ||
+                     data[ (http->stage_off%http->oring_sz)+i ] == '"' ||
+                     data[ (http->stage_off%http->oring_sz)+i ] == '\\' ) ) {
+      data[ (http->stage_off%http->oring_sz)+i ] = ' ';
+    }
+  }
+}
+
+FD_FN_UNUSED static void
+print_char_buf_to_buf_escape_hex(const char *buf, ulong buf_sz, char *out_buf, ulong out_buf_sz) {
```
ptaffet-jump, line 89: The code style is off here

src/disco/gui/fd_gui_printf.c
```diff
@@ -0,0 +1,1172 @@
...
+  /* Prints invisible chars as escaped hex codes */
+  ulong idx = 0;
+  for( ulong i = 0; i < buf_sz; ++i ) {
+    uchar c = (uchar)buf[i];
```
ptaffet-jump, line 93: Not sure why this is cast to `uchar`. `isprint` takes an int. You can just keep it as a `char`

src/disco/gui/fd_gui_printf.c
```diff
@@ -0,0 +1,1172 @@
...
+    if( isprint( c ) ) {
+      if( idx < out_buf_sz - 1 ) {
+        out_buf[idx++] = (char)c;
+      }
+    } else {
+      if( idx < out_buf_sz - 4UL ) { // \xXX takes 4 characters
```
ptaffet-jump, line 99: If `out_buf_sz<4UL`, this will underflow and be true. You can write it as `idx+4UL<out_buf_sz`. However you change it, do the same for the `isprint` true case

src/disco/gui/fd_gui_printf.c
```diff
@@ -0,0 +1,1141 @@
...
+    if( FD_LIKELY( val ) ) {
+      fd_http_server_printf( gui->http, "\"" );
+      ulong start_len = gui->http->stage_len;
+      fd_http_server_printf( gui->http, "%s", val );
+      jsonp_sanitize_str( gui->http, start_len );
+      fd_http_server_printf( gui->http, "\"," );
+    } else {
+      fd_http_server_printf( gui->http, "null," );
+    }
+  }
```
ptaffet-jump, line 116: What do you think about this? One extra call to printf in the good case, but less repetition.
```suggestion
  if( FD_LIKELY( key ) ) {
    fd_http_server_printf( gui->http, "\"%s\":", key );
  }
  if( FD_LIKELY( val ) ) {
    fd_http_server_printf( gui->http, "\"", key );
    ulong start_len = gui->http->stage_len;
    fd_http_server_printf( gui->http, "%s", val );
    jsonp_sanitize_str( gui->http, start_len );
    fd_http_server_printf( gui->http, "\"," );
  } else {
    fd_http_server_printf( gui->http, "null," );
  }
 ```

File: src/app/fdctl/config/default.toml [hunk 27]
```diff
@@ -1015,6 +1015,22 @@ dynamic_port_range = "8900-9000"
...
 
+    # The gui tile receives data from the validator and serves an HTTP
+    # endpoint to clients to view it.
+    [tiles.gui]
+        # If the GUI is enabled.  If the GUI is enabled, by default it
+        # will not show names or icons for validators unless the
+        # program-id index is also enabled in this TOML file by setting
+        #
+        # [ledger.account_indexes]
+        #   ["program-id"]
```
ptaffet-jump, line 1026:
```suggestion
        # [ledger]
        #   account_indexes = ["program-id"]
```

.github/workflows/tests.yml
```diff
@@ -56,7 +56,7 @@ jobs:
             run-integration-tests: false
           - test-case: linux_clang_x86_64
             machine: linux_clang_x86_64
-            label: X64
+            label: ci1
```
mmcgee-jump, line 59: undo

src/disco/gui/fd_gui_printf.c
```diff
@@ -0,0 +1,1130 @@
...
+                                            char const * key,
+                                            ulong        id ) {
+  jsonp_open_object( gui, NULL );
+  jsonp_string( gui, "topic", topic );
+  jsonp_string( gui, "key", key );
+  jsonp_ulong( gui, "id", id );
+}
+
+void
+fd_gui_printf_close_query_response_envelope( fd_gui_t * gui ) {
```
ptaffet-jump, line 150: This function is named so differently from `jsonp_close_envelope`, even though they're actually the same. Do you want to standardize the naming?

src/disco/gui/fd_gui_printf.c
```diff
@@ -0,0 +1,1130 @@
...
+  jsonp_close_object( gui );
+  jsonp_strip_trailing_comma( gui );
+}
+
+void
+fd_gui_printf_null_query_response( fd_gui_t *   gui,
+                                   char const * topic,
+                                   char const * key,
+                                   ulong        id ) {
+  jsonp_open_object( gui, NULL );
```
ptaffet-jump, line 160: You can use `fd_gui_printf_open_query_response_envelope` and `fd_gui_printf_close_query_response_envelope` here

ptaffet-jump (approved):
Stamping to unblock. We should probably review a bit more thoroughly later.

PR #3059, opened 2024-10-08
config: add auto affinity option

mmcgee-jump:


src/disco/topo/fd_topo.h
```diff
@@ -333,6 +333,9 @@ typedef struct fd_topo_t {
   fd_topo_link_t links[ FD_TOPO_MAX_LINKS ];
   fd_topo_tile_t tiles[ FD_TOPO_MAX_TILES ];
   fd_topo_obj_t  objs[ FD_TOPO_MAX_OBJS ];
+
+  ulong          agave_affinity_cnt;
+  ulong          agave_affinity_cpu_idx[ 1024 ];
```
ptaffet-jump, line 338:
```suggestion
  ulong          agave_affinity_cpu_idx[ FD_TILE_MAX ];
```
And should they be ushort? Some places they are and some they aren't.

src/disco/topo/fd_topob.c
```diff
@@ -392,6 +393,95 @@ validate( fd_topo_t const * topo ) {
...
+void
+fd_topob_auto_layout( fd_topo_t * topo ) {
+  /* Incredibly simple automatic layout system for now ... just assign
+     tiles to CPU cores in NUMA sequential order, except for a few tiles
+     which should be floating. */
+
+  char const * FLOATING[] = {
+    "metric",
+    "bencho",
+    "bhole",  // FIREDANCER only
```
ptaffet-jump, line 405:
```suggestion
    "bhole",  /* FIREDANCER only */
```
And similar?

src/disco/topo/fd_topob.c
```diff
@@ -392,6 +393,95 @@ validate( fd_topo_t const * topo ) {
...
+    "thread", // FIREDANCER only
+    "sender", // FIREDANCER only
+  };
+
+  for( ulong i=0UL; i<topo->tile_cnt; i++ ) {
+    fd_topo_tile_t * tile = &topo->tiles[ i ];
+    tile->cpu_idx = ULONG_MAX;
+  }
+
+  ulong cpu_ordering[ USHORT_MAX ] = { 0UL };
```
ptaffet-jump, line 435: Not a huge deal, but this is kind of big for the stack and to 0 like that. 512 kB. Is `FD_TILE_MAX` enough?

src/disco/topo/fd_topob.c
```diff
@@ -392,6 +393,95 @@ validate( fd_topo_t const * topo ) {
...
+  ulong num_cpus = fd_numa_cpu_cnt();
+
+  ulong next_cpu_idx = 0UL;
+  for( ulong i=0UL; i<fd_numa_node_cnt(); i++ ) {
```
ptaffet-jump, line 439: `fd_numa_node_cnt` does syscalls and stuff, so probably cache it in a variable like `num_cpus`

src/disco/topo/fd_topob.c
```diff
@@ -392,6 +393,95 @@ validate( fd_topo_t const * topo ) {
...
+    for( ulong j=0UL; j<num_cpus; j++ ) {
+      ulong numa_node = fd_numa_node_idx( j );
+      if( FD_UNLIKELY( numa_node!=i ) ) continue;
+      cpu_ordering[ next_cpu_idx++ ] = j;
```
ptaffet-jump, line 443: Make it abort if it goes past the end of the array instead of being an OOB write

ptaffet-jump: You can do this in O(n) with two passes (count CPUs per numa node, prefix sums, then assign) but it doesn't matter since this is startup and the number of CPUs/numa nodes is small

src/disco/topo/fd_topob.c
```diff
@@ -392,6 +393,95 @@ validate( fd_topo_t const * topo ) {
...
+
+  FD_TEST( next_cpu_idx==num_cpus );
+
+  ulong cpu_idx = 0UL;
+  for( ulong i=0UL; i<sizeof(ORDERED)/sizeof(ORDERED[0]); i++ ) {
+    for( ulong j=0UL; j<topo->tile_cnt; j++ ) {
+      fd_topo_tile_t * tile = &topo->tiles[ j ];
+      if( !strcmp( tile->name, ORDERED[ i ] ) ) {
+        if( FD_UNLIKELY( cpu_idx>=num_cpus ) ) {
+          FD_LOG_WARNING(( "auto layout cannot affine tile `%s:%lu` because all the CPUs are already assigned. Leaving it floating", tile->name, tile->kind_id ));
```
ptaffet-jump, line 455:
```suggestion
          FD_LOG_WARNING(( "auto layout cannot set affinity for tile `%s:%lu` because all the CPUs are already assigned. Leaving it floating", tile->name, tile->kind_id ));
```
I remember you like using affine as a verb, but I checked the dictionary, and I don't see that as a real word 🤣

src/app/fdctl/run/topos/fd_frankendancer.c
```diff
@@ -81,15 +80,26 @@ fd_topo_frankendancer( config_t * config ) {
...
+
+  if( FD_UNLIKELY( is_auto_affinity != is_agave_auto_affinity ||
+                   is_auto_affinity != is_bench_auto_affinity ) ) {
+    FD_LOG_ERR(( "The CPU affinity string in the configuration file under [layout.affinity], [layout.agave_affinity], and [development.bench.affinity] must all be set to 'auto' or all be set to a specific CPU affinity string." ));
+  }
+
+  ulong affinity_tile_cnt = 0UL;
+  if( FD_LIKELY( !is_auto_affinity ) ) affinity_tile_cnt = fd_tile_private_cpus_parse( config->layout.affinity, parsed_tile_to_cpu );
 
   ulong tile_to_cpu[ FD_TILE_MAX ];
```
ptaffet-jump, line 96: In the auto case, this should be zero-ed or set to USHORT_MAX so that when we read it below, it doesn't get garbage. I know the garbage will be overwritten, but that still seems safer.

src/disco/topo/fd_topob.c
ptaffet-jump: Are you sure we want this behavior though? If all the cores have processes pinned with high priority, won't the floating tiles get very little CPU time? In the non-auto case, we just error out and print a message that they need more cores.

src/disco/topo/fd_topob.c
```diff
@@ -392,6 +393,95 @@ validate( fd_topo_t const * topo ) {
...
+        } else {
+          tile->cpu_idx = cpu_ordering[ cpu_idx++ ];
+        }
+      }
+    }
+  }
+
+  for( ulong i=0UL; i<topo->tile_cnt; i++ ) {
```
ptaffet-jump, line 463:
```suggestion
  /* Make sure all the tiles we haven't set are supposed to be floating. */
  for( ulong i=0UL; i<topo->tile_cnt; i++ ) {
```

src/disco/topo/fd_topob.c
```diff
@@ -392,6 +393,95 @@ validate( fd_topo_t const * topo ) {
...
+    "benchg",
+    "benchs",
+    "net",
+    "quic",
+    "verify",
+    "dedup",
+    "pack",
+    "bank",
+    "poh",
+    "pohi",   // FIREDANCER only
```
ptaffet-jump, line 418: You can use `FD_HAS_AGAVE` to get the right list at compile time if you like

ptaffet-jump (commented):
A few small things, but a nice change!

src/disco/topo/fd_topob.c
mmcgee-jump: This is number of cores on the machine, I guess it's probably fine.

ptaffet-jump (approved):

PR #3065, opened 2024-10-08
test_tower unit test

cavemanloverboy:
Adds a unit test for tower. This test fills a tower with votes for sequential slots, and then checks the case for a vote which pops expired votes and the case where a new root is published.

### DIFF
--- a/src/choreo/tower/Local.mk
+++ b/src/choreo/tower/Local.mk
@@ -1,5 +1,5 @@
 ifdef FD_HAS_INT128
 $(call add-hdrs,fd_tower.h)
 $(call add-objs,fd_tower,fd_choreo)
-$(call make-unit-test,test_tower,test_tower,fd_choreo fd_flamenco fd_ballet fd_funk fd_util)
+$(call make-unit-test,test_tower,test_tower,fd_choreo fd_flamenco fd_ballet fd_funk fd_util,$(SECP256K1_LIBS))
 endif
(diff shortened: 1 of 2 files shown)


cavemanloverboy:
This won't compile on main without https://github.com/firedancer-io/firedancer/pull/3045 (which is why [#3045](https://github.com/firedancer-io/firedancer/pull/3045) was made in the first place, actually)

lidatong:
thanks for the contribution! good to get unit tests for tower started

try modifying the `Local.mk` with this

```
ifdef FD_HAS_INT128
ifdef FD_HAS_SECP256K1
$(call add-hdrs,fd_tower.h)
$(call add-objs,fd_tower,fd_choreo)
$(call make-unit-test,test_tower,test_tower,fd_choreo fd_flamenco fd_funk fd_ballet fd_util,$(SECP256K1_LIBS))
endif
endif
```

ripatel-fd:
Thank you for the contribution @cavemanloverboy. The libsecp256k1 dependency is unfortunate. @lidatong's suggestion will fix the linker issue. Once this is merged, I'll clean up the dependency graph so the test no longer depends on secp256k1.

cavemanloverboy:
Sounds good, just updated the .mk

lidatong (approved):

cavemanloverboy:
forgot the conditional in the mk... i'll add this when i get off my flight

PR #3067, opened 2024-10-08
fix: rw lock misbehavior

lheeger-jump:


mmcgee-jump:
Also revert the volatile change (or do it properly with FD_VOLATILE ( ... ))

mmcgee-jump (changes requested):
.

jumpsiegel:
What does the "." comment mean?

lheeger-jump:
Fixed

mmcgee-jump (approved):

PR #3069, opened 2024-10-08
metrics: add tile context switch counters

mmcgee-jump:


src/app/fdctl/monitor/monitor.c
```diff
@@ -350,6 +357,8 @@ run_monitor( config_t * const config,
       PRINT( " | " ); printf_stale   ( &buf, &buf_sz, (long)(0.5+ns_per_tic*(double)(toc - cur->cnc_heartbeat)), 1e8 /* 100 millis */ );
       PRINT( " | " ); printf_heart   ( &buf, &buf_sz, cur->cnc_heartbeat, prv->cnc_heartbeat        );
       PRINT( " | " ); printf_sig     ( &buf, &buf_sz, cur->cnc_signal,    prv->cnc_signal           );
+      PRINT( " | " ); printf_rate    ( &buf, &buf_sz, 1e9, 0., cur->nivcsw, prv->nivcsw, dt );
+      PRINT( " | " ); printf_rate    ( &buf, &buf_sz, 1e9, 0., cur->nvcsw,  prv->nvcsw,  dt );
```
ptaffet-jump, line 361: Can you align it better? Probably add spaces to the 2 previous and following lines to correspond to the 1e9 and 0

src/app/fdctl/run/tiles/fd_cswtch.c
```diff
@@ -0,0 +1,197 @@
...
+static void
+privileged_init( fd_topo_t *      topo,
+                 fd_topo_tile_t * tile,
+                 void *           scratch ) {
+  (void)tile;
+
+  FD_SCRATCH_ALLOC_INIT( l, scratch );
+  fd_cswtch_ctx_t * ctx = FD_SCRATCH_ALLOC_APPEND( l, alignof( fd_cswtch_ctx_t ), sizeof( fd_cswtch_ctx_t ) );
+
+  FD_TEST( topo->tile_cnt<FD_TOPO_MAX_TILES );
```
ptaffet-jump, line 46: Heads up that here the check is against `FD_TOPO_MAX_TILES` and the arrays in the context are declared with `FD_TILE_MAX`.  It's okay because `FD_TOPO_MAX_TILES < FD_TILE_MAX`, but it seems off

src/app/fdctl/run/tiles/fd_cswtch.c
```diff
@@ -0,0 +1,197 @@
...
+    if( FD_UNLIKELY( -1==lseek( ctx->status_fds[ i ], 0, SEEK_SET ) ) ) FD_LOG_ERR(( "lseek failed (%i-%s)", errno, strerror( errno ) ));
+
+    char contents[ 4096 ];
+    ulong contents_len = 0UL;
+
+    while( 1 ) {
+      if( FD_UNLIKELY( contents_len>=sizeof( contents ) ) ) FD_LOG_ERR(( "contents overflow" ));
+      long n = read( ctx->status_fds[ i ], contents + contents_len, sizeof( contents ) - contents_len );
+      if( FD_UNLIKELY( -1==n ) ) FD_LOG_ERR(( "read failed (%i-%s)", errno, strerror( errno ) ));
+      if( FD_UNLIKELY( 0==n ) ) break;
```
ptaffet-jump, line 113: Likely?

src/app/fdctl/run/tiles/fd_cswtch.c
```diff
@@ -0,0 +1,197 @@
...
+          FD_SPIN_PAUSE();
+          continue;
+        }
+      }
+
+      char path[ 64 ];
+      FD_TEST( fd_cstr_printf_check( path, sizeof( path ), NULL, "/proc/%lu/task/%lu/status", pid, tid ) );
+      ctx->status_fds[ i ] = open( path, O_RDONLY );
+      ctx->metrics[ i ] = fd_metrics_tile( metrics );
+      if( FD_UNLIKELY( -1==ctx->status_fds[ i ] ) ) FD_LOG_ERR(( "open failed (%i-%s)", errno, strerror( errno ) ));
```
ptaffet-jump, line 70: It seems like `/proc` [can be mounted with various options](https://docs.kernel.org/filesystems/proc.html#chapter-4-configuring-procfs) that may make this unreadable. We might want to have an option to disable it or to fail more gracefully.

src/app/fdctl/run/tiles/fd_cswtch.c
```diff
@@ -0,0 +1,197 @@
...
+      if( FD_UNLIKELY( -1==n ) ) FD_LOG_ERR(( "read failed (%i-%s)", errno, strerror( errno ) ));
+      if( FD_UNLIKELY( 0==n ) ) break;
+      contents_len += (ulong)n;
+    }
+
+    int found_voluntary = 0;
+    int found_involuntary = 0;
+
+    char * line = contents;
+    while( 1 ) {
```
ptaffet-jump, line 121: I checked to see if there was a better way to get this info from the kernel, but it doesn't seem so.

ptaffet-jump (commented):
I like the idea, but I think it's a lot to dedicate a whole core to it. Can we roll the functionality into the metrics tile? Can we make it a floating tile by default?

src/app/fdctl/run/tiles/fd_cswtch.c
mmcgee-jump: We can add that if it is a problem in practice, I don't think it's very likely anyone remounts /proc with weird options.

mmcgee-jump:
Yah this PR is supposed to land after the auto one, so we can have the tile floating.

ptaffet-jump (approved):

PR #3078, opened 2024-10-09
Fix stream handling to ensure all connections always have at least a minimum number of streams

nbridge-jump:
Also adds a test which simply keeps a specified number of connections open and idle.


src/waltz/quic/tests/Local.mk
```diff
@@ -28,6 +28,7 @@ $(call run-unit-test,test_quic_crypto)
 $(call make-unit-test,test_quic_client_flood,test_quic_client_flood,fd_quic fd_tls fd_ballet fd_waltz fd_util)
 $(call make-unit-test,test_quic_server,test_quic_server,            fd_quic fd_tls fd_ballet fd_waltz fd_util)
 $(call make-unit-test,test_quic_txns,  test_quic_txns,              fd_quic fd_tls fd_ballet fd_waltz fd_util)
+$(call make-unit-test,test_quic_idle_conns, test_quic_idle_conns,   fd_quic fd_tls fd_ballet fd_waltz fd_util)
```
mmcgee-jump, line 31: Any reason these aren't added with run-unit-test so it runs in CI as well?

mmcgee-jump (approved):

PR #3082, opened 2024-10-10
feat(eqvoc): equivocation tile

lidatong:
Introduce a tile and logic (fd_eqvoc) to detect equivocating shreds.

There are 2 cases to consider:
1. Equivocation for the same FEC set
2. Equivocation for overlapping FEC sets

The first case is trivially checked by making sure the signatures of
all shreds in the same FEC set are the same.

The second case is trickier. Two (or more) FEC sets might overlap in
their shred indices. Because there are two separate FEC sets, we expect
the signatures to be different. However, this still implies
equivocation, because in order for FEC sets to overlap the same shred
idx occurs twice. See fd_eqvoc.h for details.

The tile is currently mostly a stub to setup the links and topo. The
full impl will come with subsequent PRs.


mmcgee-jump:
Why is this a tile? Is it expensive to do this work?

lidatong:
> Why is this a tile? Is it expensive to do this work?

it has to examine every shred (before resolving the fec set) as well rx / tx potentially spammy gossip msgs

contrib/test/setup_fd_cluster.sh
```diff
@@ -5,6 +5,7 @@ IFS=$'\n\t'
 PRIMARY_IP=$(ip -o -4 addr show scope global | awk '{ print $4 }' | cut -d/ -f1)
 RPC_URL="http://$PRIMARY_IP:8899/"
 
+rm -rf ../test-ledger
```
lheeger-jump, line 8: rm

contrib/test/setup_fd_cluster.sh
```diff
@@ -71,7 +72,8 @@ genesis_args+=(--upgradeable-program NanoToken1111111111111111111111111111111111
 
 echo "${genesis_args[@]}"
 
-GENESIS_OUTPUT=$(./agave/target/release/solana-genesis \
+pwd
+GENESIS_OUTPUT=$(/data/chali/agave/release/solana-genesis \
```
lheeger-jump, line 76: make parameterizeable with reasonable default

contrib/test/setup_fd_cluster.sh
```diff
@@ -92,7 +94,7 @@ GENESIS_HASH=$(echo $GENESIS_OUTPUT | grep -o -P '(?<=Genesis hash:).*(?=Shred v
 SHRED_VERSION=$(echo $GENESIS_OUTPUT | grep -o -P '(?<=Shred version:).*(?=Ticks per slot:)' | xargs)
 _PRIMARY_INTERFACE=$(ip route show default | awk '/default/ {print $5}')
 
-RUST_LOG=debug ./agave/target/release/agave-validator \
+RUST_LOG=debug /data/chali/agave/release/agave-validator \
```
lheeger-jump, line 97: ditto

lheeger-jump (changes requested):

lheeger-jump (approved):

PR #3090, opened 2024-10-11
tiles: rewrite mux to stem

mmcgee-jump:


ripatel-fd (changes requested):
Large refactors like these should come with unit tests. 
fd_stem doesn't appear to have any test coverage

mmcgee-jump:
That's coming in a separate change, this is getting too big.

ripatel-fd:
> That's coming in a separate change, this is getting too big.

Changing test_mux to test_stem instead of deleting it might reduce the diff

mmcgee-jump:
That test is a bit limited, and it also didn't run in CI.

mmcgee-jump:
todo: switch the printer for counter/gauge to use converter as well

src/disco/stem/fd_stem.c
```diff
@@ -0,0 +1,668 @@
...
+  }
+
+  /* out frag stream init */
+
+  cr_avail = 0UL;
+
+  out_depth  = (ulong *)FD_SCRATCH_ALLOC_APPEND( l, alignof(ulong), out_cnt*sizeof(ulong) );
+  out_seq    = (ulong *)FD_SCRATCH_ALLOC_APPEND( l, alignof(ulong), out_cnt*sizeof(ulong) );
+
+  ulong cr_min = fd_ulong_if( !out_cnt, 128UL, ULONG_MAX );
```
ptaffet-jump, line 267: Maybe call this `depth_min`

ptaffet-jump: And then keep the variable formerly called `cr_max`. It's just confusing to see a variable change names from `cr_max` to `cr_min` without changing roles

ptaffet-jump (approved):
I looked at `fd_stem` a little, and the rest of the PR almost not at all, but I'm stamping to unblock

lheeger-jump (approved):

ripatel-fd (dismissed):

ripatel-fd:
Kicking CI to rerun failed jobs. (flaky CI environment)

ripatel-fd (approved):

ptaffet-jump (approved):
Restamping

PR #3094, opened 2024-10-11
fdctl: remove concrete objects

lheeger-jump:
- **fdctl: remove concrete objects**
- **contrib: fix configuration of firedancer localnet script**


mmcgee-jump (changes requested):
Remove the code in `fdctl_obj_new` that checks for these concrete objects. You also need to remove the `loose_footprint` from the relevant workspaces now as well I think?

lheeger-jump:
> Remove the code in fdctl_obj_new that checks for these concrete objects

Done

>  You also need to remove the loose_footprint from the relevant workspaces now as well I think?

What do you mean?

mmcgee-jump (approved):
Thanks!

PR #3102, opened 2024-10-11
ci: unreserve huge/gigantic pages before test

mmcgee-jump:


yufeng-jump:
Can probably get rid of the reset if we are doing a fini.

topointon-jump (approved):

ripatel-fd (approved):

PR #3105, opened 2024-10-11
quic: remove fd_quic_stream_send gather list

ripatel-fd:
fd_quic_stream_send previously accepted a gather list of chunks.
All those chunks are then sent over the same stream.

This commit simplifies the API to instead only take one chunk.
A bug is fixed in fddev/txn.c that incorrectly sends multiple txns
over the same stream.
Otherwise, all other callers only sent one chunk per call anyways.


mmcgee-jump (approved):
lgtm

ravyu-jump (approved):

PR #3115, opened 2024-10-14
Hardcode account_hash_ignore_slot feature gate

ripatel-fd:


ripatel-fd:
CI might fail if there are tests that set this feature gate to disabled

kbhargava-jump:
let's hold off on this right now

jumpsiegel:
Could we hold on this until after the lthash branch.  This is pure merge conflict for me

ripatel-fd:
> let's hold off on this right now

why?

jumpsiegel:
because I am going to have to make these changes, line for line, again during my merge as I fix all the conflicts.  We should instead just do them on the lthash branch.

ripatel-fd:
> because I am going to have to make these changes, line for line, again during my merge as I fix all the conflicts. We should instead just do them on the lthash branch.

Yes, I understand, it's a good reason. But I wanted to ask why Kunal objects to the changes.

PR #3124, opened 2024-10-14
quic - changed defaults.toml max_concurrent_handshakes description and value

nbridge-jump:
`max_concurrent_handshakes` is no longer bound to the value of `max_concurrent_connections`. So this PR fixes the description and sets the value to a conservative number.

### DIFF
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -843,7 +843,7 @@ dynamic_port_range = "8900-9000"
         # be accepted.
         #
         # This must be >= 2 and also a power of 2.
-        max_concurrent_connections = 2048
+        max_concurrent_connections = 65536
 
         # QUIC allows for multiple streams to be multiplexed over a
         # single connection.  This option sets the maximum number of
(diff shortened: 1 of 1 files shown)


mmcgee-jump:
Can you update the default for `max_concurrent_connections` to be correct as well?

mmcgee-jump (changes requested):
.

nbridge-jump:
> Can you update the default for `max_concurrent_connections` to be correct as well?

I'm not sure what you're asking for. This value isn't being used in mainnet. Is that what you want?

mmcgee-jump:
The default values we provide should be sane for mainnet, which 2048 for max_concurrent_connections clearly is not.

nbridge-jump:
> The default values we provide should be sane for mainnet, which 2048 for max_concurrent_connections clearly is not.

Just accept this PR. Increasing max_concurrent_connections to mainnet levels broke some tests. So I'll work on fixing all that in another PR. No need to hold up this change, which is clearly an improvement on main

nbridge-jump:
Ignore that. Looks like the issue was already solved. Will squash the commits

mmcgee-jump (approved):

PR #3126, opened 2024-10-15
flamenco, runtime: align all vm regions to 16-bytes

topointon-jump:
All vm memory regions need to be 8-byte aligned. This has consensus implications - Agave's vm checks that host physical addresses are aligned during address translation. If the host address is unaligned than address translation will fail. We need to match this behaviour.

Direct mapping means that we have to ensure that the account data region of borrowed accounts are therefore 8-byte aligned.

### DIFF
--- a/src/ballet/bn254/fd_bn254.c
+++ b/src/ballet/bn254/fd_bn254.c
@@ -257,6 +257,12 @@ fd_bn254_pairing_is_one_syscall( uchar       out[32],
       sz = 0;
     }
   }
+  if( sz>0 ) {
+    fd_bn254_fp12_t tmp[1];
+    fd_bn254_miller_loop( tmp, p, q, sz );
+    fd_bn254_fp12_mul( r, r, tmp );
+    sz = 0;
+  }
 
   /* Compute the final exponentiation */
   fd_bn254_final_exp( r, r );
--- a/src/flamenco/types/fd_types.h
+++ b/src/flamenco/types/fd_types.h
@@ -405,13 +405,13 @@ typedef struct fd_solana_account_stored_meta_off fd_solana_account_stored_meta_o
 #define FD_SOLANA_ACCOUNT_STORED_META_OFF_FOOTPRINT sizeof(fd_solana_account_stored_meta_off_t)
 #define FD_SOLANA_ACCOUNT_STORED_META_OFF_ALIGN (8UL)
 
-/* Encoded Size: Fixed (56 bytes) */
+/* Encoded Size: Fixed (52 bytes) */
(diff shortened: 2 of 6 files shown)


src/flamenco/runtime/fd_borrowed_account.c
```diff
@@ -33,8 +34,8 @@ fd_borrowed_account_resize( fd_borrowed_account_t * borrowed_account,
   // TODO: Check for max accounts size?
   uchar * new_raw_data = (uchar *)buf;
 
-  ulong old_sz = sizeof(fd_account_meta_t)+borrowed_account->meta->dlen;
-  ulong new_sz = sizeof(fd_account_meta_t)+dlen;
+  ulong old_sz = fd_ulong_align_up( sizeof(fd_account_meta_t), FD_VM_HOST_REGION_ALIGN ) +borrowed_account->meta->dlen;
+  ulong new_sz = fd_ulong_align_up( sizeof(fd_account_meta_t), FD_VM_HOST_REGION_ALIGN )+dlen;
   fd_memcpy( new_raw_data, borrowed_account->const_meta, old_sz );
```
jumpsiegel, line 39: the mind boggles

src/flamenco/runtime/fd_executor.c
```diff
@@ -1266,6 +1267,14 @@ fd_executor_setup_borrowed_accounts_for_txn( fd_exec_txn_ctx_t * txn_ctx ) {
...
         borrowed_account->meta->info.rent_epoch = ULONG_MAX;
       }
+    } else {
+      /* When direct mapping is enabled, the account data has to be 16-byte aligned, otherwise memory translation inside the vm
+         will fail non-deterministically. */
+      if ( FD_UNLIKELY( FD_FEATURE_ACTIVE( txn_ctx->slot_ctx, bpf_account_data_direct_mapping ) ) ) {
+        uchar * aligned_const_data = fd_valloc_malloc( txn_ctx->valloc, FD_VM_HOST_REGION_ALIGN, borrowed_account->const_meta->dlen );
+        fd_memcpy( aligned_const_data, borrowed_account->const_data, borrowed_account->const_meta->dlen );
+        borrowed_account->const_data = aligned_const_data;
+      }
```
jumpsiegel, line 1277: instead of always copying, couldn't we check to see if const_data is already aligned and only copy if it is not?  Or, is it never aligned 16 byte aligned due to the size of the header?  This feels like we should either tweak the header or hack funk.  memcpy's are already a large portion of our execution time

jumpsiegel (changes requested):

0x0ece (approved):

PR #3132, opened 2024-10-16
`ci`: set BUILDDIR such that clusterfuzz.yml fuzzer binaries end up in `clang-fuzz/`

nlgripto:
This should fix the CI build for the "Clusterfuzz" fuzzers


nlgripto:
alternative: change the `find` command to go into the appropriate machine dir inside `build/`

ripatel-fd (approved):

PR #3137, opened 2024-10-16
quic: remove flow control limits

ripatel-fd:
- Remove ahead-of-time stream assignment
- Simplify stream count metrics
- Remove flow control limits
- Introduce stream concurrency bounds
- Further remove support for bidirectional streams

Thanks to recent fixes to fragmentation patterns in Agave's QUIC
client, we can make stricter assumptions about stream framing
patterns.  Agave clients now promise to use stream IDs in sequence.
Anytime a stream ID increments, the previous stream has already
entered FIN state.  On the server-side stream concurrency only
exceeds 1 in case of IP packet reordering along the network path.

This patch removes ahead-of-time assignment of stream IDs.
Thereby, flow control limits become independent of concurrency
limits.  Specifying a new flow control mechanism is out of scope
for this PR, so MAX_STREAMS and MAX_DATA are set to 2^62-1.

The definition of 'max_concurrent_streams_per_connection' is
changed to control concurrency only instead of flow control.
If a client exceeds this limit, their traffic is silently dropped.


src/waltz/quic/tests/test_quic_conformance.c
```diff
@@ -45,20 +49,219 @@ test_quic_stream_limit_enforcement( fd_quic_sandbox_t * sandbox,
...
   FD_TEST( conn->state  == FD_QUIC_CONN_STATE_ABORT );
   FD_TEST( conn->reason == FD_QUIC_CONN_REASON_STREAM_LIMIT_ERROR );
 }
 
+/* Ensure that worst-case stream pool allocations are bounded.
+
+   This is a custom protocol restriction that goes beyond QUIC's limits. */
+
+static __attribute__ ((noinline)) void
+test_quic_stream_concurrency( fd_quic_sandbox_t * sandbox,
```
ripatel-fd, line 71: Just noticed I didn't complete this test

src/waltz/quic/tests/test_quic_conformance.c
```diff
@@ -45,20 +49,219 @@ test_quic_stream_limit_enforcement( fd_quic_sandbox_t * sandbox,
...
+    FD_TEST( sz!=FD_QUIC_ENCODE_FAIL );
+    buf[ sz++ ] = '0';
+    fd_quic_sandbox_send_lone_frame( sandbox, conn, buf, sz );
+    FD_TEST( conn->state == FD_QUIC_CONN_STATE_ACTIVE );
+  }
+  FD_TEST( sandbox->quic->metrics.stream_rx_event_cnt == 1UL );
+  FD_TEST( sandbox->quic->metrics.stream_active_cnt == 0UL );
+}
+
+/* reorder: ensure that reordered fragments arrive without corruption */
```
ripatel-fd, line 234: Missing test

mmcgee-jump (approved):

PR #3157, opened 2024-10-18
flamenco, log: use variadics in fd_log_collector_msg_many

ravyu-jump:
As we start digging into CPI mismatches I figured we'd extend `fd_log_collector_msg_many` to take it a variable number of arguments. Works fine on existing invocations, and also tested locally on a call with >2 msgs (to be pushed in a separate PR).


0x0ece (approved):
LGTM, the only thing I'm not sure is the use of 2x va_start and 1x va_end. It's possible that it works because the 2nd va_start is inside a scope, but just to be safer I'd put a va_end for each va_start, I left comments in the code.

0x0ece (approved):

PR #3166, opened 2024-10-18
gui: moving window average and max for tile primary metrics

yufeng-jump:
Currently live values are computed from the end of the prior leader slot.  Change to a 1-second moving average.  Also change queries for our slots to return max within slot, so this would tell us for example the peak number of QUIC connections or peak fullness of the pack buffer that we observed within a slot.

### DIFF
--- a/book/api/websocket.md
+++ b/book/api/websocket.md
@@ -587,21 +587,18 @@ potential underflow.
 | Field               | Type                | Description
 |---------------------|---------------------|------------
 | next_leader_slot    | `number\|null`      | The next leader slot |
-| tile_primary_metric | `TilePrimaryMetric` | Per-tile-type primary metrics.  Some of these are point-in-time values (P), and some are aggregated since the end of the previous leader slot (A) |
+| tile_primary_metric | `TilePrimaryMetric` | Per-tile-type primary metrics.  Some of these are point-in-time values (P), and some are 1-second moving window averages (W) |
 
 **`TilePrimaryMetric`**
 | Field   | Type     | Description |
 |---------|----------|-------------|
-| net_in  | `number` | Ingress bytes per second (P) |
+| net_in  | `number` | Ingress bytes per second (W) |
 | quic    | `number` | Active QUIC connections (P) |
-| verify  | `number` | Fraction of transactions that failed sigverify (A) |
-| dedup   | `number` | Fraction of transactions deduplicated (A) |
+| verify  | `number` | Fraction of transactions that failed sigverify (W) |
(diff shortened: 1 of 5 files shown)


.github/actions/hugepages/action.yml
```diff
@@ -17,7 +17,6 @@ runs:
       run: |
         set -x
         sudo src/util/shmem/fd_shmem_cfg fini || true
-        sudo src/util/shmem/fd_shmem_cfg reset || true
```
mmcgee-jump, line 20: Can you move all the CI changes out to another PR if they are still needed?

book/api/websocket.md
```diff
@@ -586,21 +586,21 @@ potential underflow.
 | Field               | Type                | Description
 |---------------------|---------------------|------------
 | next_leader_slot    | `number\|null`      | The next leader slot |
-| tile_primary_metric | `TilePrimaryMetric` | Per-tile-type primary metrics.  Some of these are point-in-time values (P), and some are aggregated since the end of the previous leader slot (A) |
+| tile_primary_metric | `TilePrimaryMetric` | Per-tile-type primary metrics.  Some of these are point-in-time values (P), and some are moving window averages (W) |
```
mmcgee-jump, line 589: Mention that it's 1s moving average

mmcgee-jump: Also is there a separate part of the docs clarifying that it's the max when querying a historical slot?

yufeng-jump: Yes, in the doc for SlotResponse

book/api/websocket.md
```diff
@@ -587,21 +587,21 @@ potential underflow.
...
+| verify  | `number` | Fraction of transactions that failed sigverify (W) |
+| dedup   | `number` | Fraction of transactions deduplicated (W) |
 | pack    | `number` | Fraction of pack buffer filled (P) |
-| bank    | `number` | Execution TPS (P) |
-| poh     | `number` | Fraction of time spent hashing (P) |
-| shred   | `number` | Shreds processed per second (P) |
-| store   | `number` | 50% percentile latency (A) |
-| net_out | `number` | Egress bytes per second (P) |
+| bank    | `number` | Execution TPS (W) |
+| poh     | `number` | Fraction of time spent hashing (W) |
```
mmcgee-jump, line 601: Looks like we don't send a few of these (e.g. poh, shred, ..) remove them here I think

mmcgee-jump:
Replaced by https://github.com/firedancer-io/firedancer/pull/3440

PR #3169, opened 2024-10-21
syscall: always test alignment, like solfuzz-agave

0x0ece:


src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -1770,7 +1770,12 @@ fd_exec_vm_syscall_test_run( fd_exec_instr_test_runner_t * runner,
   if ( !vm ) {
     goto error;
   }
-  uchar is_deprecated = !memcmp( input_instr_ctx->program_id, &fd_solana_bpf_loader_deprecated_program_id, sizeof(fd_pubkey_t) );
+  uchar is_deprecated = false; /* tl;dr: solfuzz-agave always checks alignemnt.
```
topointon-jump, line 1773: Let's chat about this, I feel like we should set the owner to the deprecated loader if `check_align` is false

topointon-jump (approved):

PR #3171, opened 2024-10-21
flamenco, vm: implement correct checks for stack gap access violations

mjain-jump:


ibhatt-jumptrading (dismissed):
good find

src/flamenco/vm/fd_vm_private.h
```diff
@@ -309,25 +309,35 @@ fd_vm_mem_haddr( fd_vm_t const *    vm,
...
     /* To account for the fact that we have gaps in the virtual address space but not in the 
        physical address space, we need to subtract from the offset the size of all the virtual
        gap frames underneath it.
        
-       https://github.com/solana-labs/rbpf/blob/b503a1867a9cfa13f93b4d99679a17fe219831de/src/memory_region.rs#L147-L149 */
-    ulong gap_mask = 0xFFFFFFFFFFFFF000;
+       https://github.com/solana-labs/rbpf/blob/v0.8.5/src/memory_region.rs#L137 */
+    ulong gap_mask = ( 0xFFFFFFFFFFFFFFFFUL << vm_gap_shift );
     offset = ( ( offset & gap_mask ) >> 1 ) | ( offset & ~gap_mask ); 
   }
```
topointon-jump, line 342: This whole block is only needed if the gaps are enabled, so you can keep the inside as-is and just add the direct mapping condition to the if statement.

mjain-jump: I think it's still necessary - if direct mapping is enabled but the access is not in a stack frame gap, the offset would be calculated incorrectly.

topointon-jump: the offset adjustment is just to account for the fact we have virtual stack gaps but not physical ones - if there are no virtual gaps no adjustment _should_ be needed

src/flamenco/runtime/tests/fd_exec_sol_compat.c
```diff
@@ -45,7 +45,7 @@ sol_compat_init( void ) {
   char ** argv_ = argv;
   setenv( "FD_LOG_PATH", "", 1 );
   fd_boot( &argc, &argv_ );
-  fd_log_level_logfile_set(5);
```
ibhatt-jumptrading, line 48: fix

src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -163,7 +163,7 @@
   shadow[ frame_cnt ].r9 = reg[9];                                                                          \
   shadow[ frame_cnt ].pc = pc;                                                                              \
   if( FD_UNLIKELY( ++frame_cnt>=frame_max ) ) goto sigstack; /* Note: untaken branches don't consume BTB */ \
-  reg[10] += FD_VM_STACK_FRAME_SZ + FD_VM_STACK_GUARD_SZ
```
ibhatt-jumptrading, line 166: We ideally shouldn't call slot_ctx from the actual interpreter. the better approach is probably to add a field into the vm indicating if direct mapping is enabled or not

src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -663,7 +663,7 @@
     reg[8]   = shadow[ frame_cnt ].r8;
     reg[9]   = shadow[ frame_cnt ].r9;
     pc       = shadow[ frame_cnt ].pc;
-    reg[10] -= FD_VM_STACK_FRAME_SZ + FD_VM_STACK_GUARD_SZ;
```
ibhatt-jumptrading, line 666: same thin here

src/flamenco/vm/fd_vm_private.h
```diff
@@ -313,10 +313,10 @@ fd_vm_mem_haddr( fd_vm_t const *    vm,
   ulong region    = fd_ulong_min( vaddr_hi, 5UL );
   ulong offset    = vaddr & 0xffffffffUL;
 
-  /* Stack memory regions have 4kB unmapped "gaps" in-between each frame.
+  /* Stack memory regions have 4kB unmapped "gaps" in-between each frame (only if direct mapping is disabled).
     https://github.com/solana-labs/rbpf/blob/b503a1867a9cfa13f93b4d99679a17fe219831de/src/memory_region.rs#L141
     */
-  if ( FD_UNLIKELY( region == 2UL ) ) {
+  if ( FD_UNLIKELY( region == 2UL && !FD_FEATURE_ACTIVE( vm->instr_ctx->slot_ctx, bpf_account_data_direct_mapping ) ) ) {
```
ibhatt-jumptrading, line 319: and here

src/flamenco/vm/fd_vm_interp_core.c
ibhatt-jumptrading: Also we shouldn't probably add a conditional here if it can be avoided. Let's determine the stack guard size when the vm is instantiated

src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -163,7 +163,7 @@
   shadow[ frame_cnt ].r9 = reg[9];                                                                          \
   shadow[ frame_cnt ].pc = pc;                                                                              \
   if( FD_UNLIKELY( ++frame_cnt>=frame_max ) ) goto sigstack; /* Note: untaken branches don't consume BTB */ \
-  reg[10] += FD_VM_STACK_FRAME_SZ + FD_VM_STACK_GUARD_SZ
+  reg[10] += FD_VM_STACK_FRAME_SZ + ( FD_FEATURE_ACTIVE( vm->instr_ctx->slot_ctx, bpf_account_data_direct_mapping ) ? 0 : FD_VM_STACK_GUARD_SZ )
```
topointon-jump, line 166: can we move this into a variable and put it at the top of fd_vm_interp_core.c or fd_vm_interp.c?

ibhatt-jumptrading (changes requested):
small comments

topointon-jump (commented):
Do you mind adding the fixture also? 🙏

src/flamenco/vm/fd_vm_private.h
ibhatt-jumptrading: this wasn't resolved

ibhatt-jumptrading (approved):

PR #3176, opened 2024-10-21
gui: broadcast updated skip rate more frequently

yufeng-jump:


src/disco/gui/fd_gui.c
```diff
@@ -1312,6 +1312,17 @@ fd_gui_handle_rooted_slot( fd_gui_t * gui,
     slot->level = FD_GUI_SLOT_LEVEL_ROOTED;
     fd_gui_printf_slot( gui, parent_slot );
     fd_http_server_ws_broadcast( gui->http );
+
+    /* Broadcast new skip rate if one of our slots got rooted. */
```
mmcgee-jump, line 1316: I think rooted is the wrong place for this, should be if a slot is "completed" or flips to skipped (or back from skipped).

mmcgee-jump (changes requested):
.

src/disco/gui/fd_gui.c
```diff
@@ -1289,6 +1289,17 @@ fd_gui_handle_completed_slot( fd_gui_t * gui,
   } else if( FD_UNLIKELY( gui->epoch.has_epoch[ 1 ] && _slot==gui->epoch.epochs[ 1 ].end_slot ) ) {
     gui->epoch.epochs[ 1 ].end_time = slot->completed_time;
   }
+
+  /* Broadcast new skip rate if one of our slots got completed. */
```
mmcgee-jump, line 1293: This still isn't right, this doesn't happen if the slot flips state skipped/unskipped. You need to handle this in `fd_gui_handle_reset_slot`. It actually looks like there's already code there to do this, so it must just have a bug?

yufeng-jump: Not sure I understand.  There's code already in reset slot handling that deals with state flips.  We weren't getting skip rate updates on non-skipped slots presumably because the slots start off as non-skipped, and so there are no state flips down the road, and so the backend doesn't send a skip rate update, unless we run into a skipped slot that triggers the state flip code path.  Am I missing something here?

mmcgee-jump: Hmm I see, this makes sense.

mmcgee-jump (approved):

PR #3177, opened 2024-10-21
gui: fix slot completed updates

yufeng-jump:
We are seeing slots starting off in an optimistically confirmed state in the GUI, rather than the completed state that a slot should usually start from.  This is because of a bug relating to the timing of replay completion relative to optimistic confirmation and the handling of slot states in the GUI when those two events happen out of order, i.e. when a higher slot is optimistically confirmed before replay completes for a lower slot.  This PR fixes that.

### DIFF
--- a/src/disco/gui/fd_gui.c
+++ b/src/disco/gui/fd_gui.c
@@ -38,7 +38,7 @@ fd_gui_new( void *             shmem,
     return NULL;
   }
 
-  if( FD_UNLIKELY( topo->tile_cnt>128UL ) ) {
+  if( FD_UNLIKELY( topo->tile_cnt>FD_GUI_TILE_TIMER_TILE_CNT ) ) {
     FD_LOG_WARNING(( "too many tiles" ));
     return NULL;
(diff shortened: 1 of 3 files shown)


src/disco/gui/fd_gui.c
```diff
@@ -1284,6 +1284,9 @@ fd_gui_handle_completed_slot( fd_gui_t * gui,
     slot->compute_units   = compute_units;
   }
 
+  fd_gui_printf_slot( gui, _slot );
```
mmcgee-jump, line 1287: I think the not publishing here is by design, since the slot will get published when it becomes part of the active fork chain. Otherwise every slot gets published twice.

mmcgee-jump (changes requested):
.

src/disco/gui/fd_gui.c
yufeng-jump: I see that makes sense

mmcgee-jump (approved):

PR #3182, opened 2024-10-22
sbpf_loader_fix_strtab: duplicate check for strtab

jumpsiegel:
We were missing a duplicate check


topointon-jump (commented):
Could you add the fixture? 🙏

mjain-jump (approved):

topointon-jump (dismissed):

mjain-jump (approved):

PR #3183, opened 2024-10-22
net: fix assertion failure

mmcgee-jump:


ptaffet-jump (dismissed):
We should understand why this is non-zero and fix it later, but for now this is fine.

ptaffet-jump (approved):

PR #3187, opened 2024-10-22
flamenco, vm: fix OOB in tracing code

topointon-jump:
The `pc>=text_cnt` check happens in `FD_VM_INTERP_INSTR_EXEC`, _after_ `fd_vm_trace_event_exe` is called. This leads to an OOB in the tracing code. Moving the call to create the trace event after the check guarantees that pc is always valid.

### DIFF
--- a/src/flamenco/vm/fd_vm_interp_core.c
+++ b/src/flamenco/vm/fd_vm_interp_core.c
@@ -174,7 +174,7 @@ interp_exec:
   /* Note: when tracing or optimizing for code footprint, all
      instruction execution starts here such that this is only point
      where exe tracing diagnostics are needed. */
-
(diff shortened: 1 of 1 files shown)


src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -170,6 +170,8 @@
   goto interp_exec; /* Silly but to avoid unused label warning in some configurations */
 interp_exec:
 
+  FD_VM_INTERP_INSTR_EXEC;
```
ravyu-jump, line 173: `FD_VM_INTERP_INSTR_EXEC` performs a jump, which means the tracing stuff below this can never get executed

ravyu-jump (approved):

PR #3194, opened 2024-10-23
Use a memory mapped backing file in funk

asiegel-jt:


src/util/shmem/fd_shmem_user.c
```diff
@@ -488,7 +488,7 @@ fd_shmem_join_anonymous( char const * name,
   join_info = fd_shmem_private_map_query_by_addr( fd_shmem_private_map, a0, a1, NULL );
   if( FD_UNLIKELY( join_info ) ) {
     FD_SHMEM_UNLOCK;
-    FD_LOG_WARNING(( "%s join memory already mapped", name ));
+    FD_LOG_WARNING(( "(%s,%lx,%lx) memory already mapped as (%s,%lx,%lx)", name, a0, a1, join_info->name, (ulong)join_info->shmem, (ulong)join_info->shmem + join_info->page_cnt*join_info->page_sz ));
```
mmcgee-jump, line 491: looks like stray debugging code? maybe revert this file

asiegel-jt: this change is intended to make the message more helpful

mmcgee-jump: These warnings are going to be read by operators, not developers (who can just dump the variables as you've done). I don't know what an operator would do with any of this, and it makes it inconsistent with the rest of the file.

jumpsiegel (approved):

PR #3205, opened 2024-10-24
net - add support for additional ip addresses

nbridge-jump:
Required for some specific configurations

### DIFF
--- a/src/app/fdctl/config.c
+++ b/src/app/fdctl/config.c
@@ -532,6 +532,7 @@ fdctl_cfg_from_env( int *      pargc,
   config->is_live_cluster = cluster != FD_CONFIG_CLUSTER_UNKNOWN;
 
   if( FD_UNLIKELY( config->development.netns.enabled ) ) {
+    /* not currently supporting multihoming on netns */
     if( FD_UNLIKELY( strcmp( config->development.netns.interface0, config->tiles.net.interface ) ) )
       FD_LOG_ERR(( "netns interface and firedancer interface are different. If you are using the "
                    "[development.netns] functionality to run Firedancer in a network namespace "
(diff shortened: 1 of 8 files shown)


src/app/fdctl/run/topos/fd_frankendancer.c
```diff
@@ -361,6 +361,11 @@ fd_topo_initialize( config_t * config ) {
       tile->net.quic_transaction_listen_port   = config->tiles.quic.quic_transaction_listen_port;
       tile->net.legacy_transaction_listen_port = config->tiles.quic.regular_transaction_listen_port;
 
+      /* multihome support */
+      ulong multi_cnt = tile->net.multihome.cnt = config->tiles.net.multihome.src_ip_addr_cnt;
+      for( ulong j = 0; j < multi_cnt; ++j ) {
+        tile->net.multihome.src_ip_addr[j] = config->tiles.net.multihome.src_ip4_addr[j];
```
anwayde, line 367: does this need to be a `fd_memcpy`?

anwayde:
We also need to add this to `src/app/fdctl/run/topos/fd_firedancer.c`

src/app/fdctl/run/topos/fd_frankendancer.c
nbridge-jump: fd_memcpy is not great for small copies. Also, we expect multi_cnt to be 0 or 1.
Lastly, this is in initialization where this kind of concern is unimportant.

anwayde (dismissed):

src/app/fdctl/config/default.toml
```diff
@@ -820,6 +820,10 @@ dynamic_port_range = "8900-9000"
         # this really be configurable?
         send_buffer_size = 16384
 
+        # Additional addresses this validator needs to be able to
```
mmcgee-jump, line 823: Don't need a nested namespace `multihome` for this, just stick it under net tile same as the other options.

Can you also add a detailed comment a validator operator who doesn't know what "multihome" is could use to understand this option, and why it might need to be set?

mmcgee-jump: Maybe call it `multihome_ip_addrs` ? I believe the convention in this config is array member names are plural as well.

src/app/fdctl/config.c
```diff
@@ -570,6 +571,18 @@ fdctl_cfg_from_env( int *      pargc,
 
     config->tiles.net.ip_addr = iface_ip;
     mac_address( config->tiles.net.interface, config->tiles.net.mac_addr );
+
+    /* support for multihomed hosts */
```
mmcgee-jump, line 575: Can you check that none of the multi-homed IP addresses are the same here, and also that none of them are the same as the iface_ip we calculated above. (I think the net tile will fail with an opaque error if these are true).

src/app/fdctl/config.c
```diff
@@ -570,6 +571,18 @@ fdctl_cfg_from_env( int *      pargc,
...
+    ulong multi_cnt = config->tiles.net.multihome.src_ip_addr_cnt;
+    for( ulong j = 0; j < multi_cnt; ++j ) {
+      int success = fd_cstr_to_ip4_addr( config->tiles.net.multihome.src_ip_addr[j],
+          &config->tiles.net.multihome.src_ip4_addr[j] );
+      if( !success ) {
+        FD_LOG_ERR(( "Failed to convert net.multihome address %s to ipv4",
```
mmcgee-jump, line 581: > "configuration option [tiles.net.multihome_ip_addrs] specifies malformed IP address `%s`"

src/app/fdctl/config.c
```diff
@@ -570,6 +571,18 @@ fdctl_cfg_from_env( int *      pargc,
 
     config->tiles.net.ip_addr = iface_ip;
     mac_address( config->tiles.net.interface, config->tiles.net.mac_addr );
+
+    /* support for multihomed hosts */
+    ulong multi_cnt = config->tiles.net.multihome.src_ip_addr_cnt;
```
mmcgee-jump, line 576:
```suggestion
  if( FD_UNLIKELY( multi_cnt>FD_NET_MAX_SRC_ADDR ) ) FD_LOG_ERR(( "configuration option [tiles.net.multihome_ip_addrs] specifies too many additional IP addresses, at most %d are allowed" ));
```

mmcgee-jump (changes requested):
Sorry for the nits ... operator facing configuration options and error messages should be very thorough

mmcgee-jump (approved):

anwayde (approved):

PR #3223, opened 2024-10-24
gui: fix missing waterfall for the first slot after boot

yufeng-jump:


src/disco/gui/fd_gui.h
```diff
@@ -165,7 +165,7 @@ struct fd_gui_slot {
 
   int leader_state;
 
-  ulong prior_leader_slot;
+  fd_gui_txn_waterfall_t waterfall_begin[ 1 ];
```
mmcgee-jump, line 168: I don't love doing this because it bloats memory use (we duplicate every waterfall_t twice, one end and one begin.) Can we keep the prior_leader_slot reference, but for the first slot just set it to ULONG_MAX and assume reference waterfall is all zeroes.

mmcgee-jump (changes requested):
.

src/disco/gui/fd_gui.h
mmcgee-jump: Actually nvm, this is like 20 mb.

mmcgee-jump (approved):

src/disco/gui/fd_gui.c
```diff
@@ -1096,6 +1091,7 @@ fd_gui_handle_slot_end( fd_gui_t * gui,
 
   fd_gui_txn_waterfall_snap( gui, slot->waterfall_end );
   fd_gui_tile_prime_metric_snap( gui, slot->waterfall_end, slot->tile_prime_metric_end );
+  memcpy( slot->waterfall_begin, gui->summary.txn_waterfall_reference, sizeof(slot->waterfall_begin) );
```
mmcgee-jump, line 1094: Don't you need to set waterfall_begin to all zeros for the first slot though? Nothing else has ended so it will be random garbage.

mmcgee-jump (changes requested):
.

src/disco/gui/fd_gui.h
yufeng-jump: Haha at one point I advocated for this prior_leader_slot approach and you were advocating for storing both begin and end.  I think you made a good point back then that having begin stored makes waterfall more available for validators that only have very few rotations per epoch.  For such validators, the prior leader slot might have been overwritten when they try to look at some leader slot.

yufeng-jump: But I agree this duplication is bad for validators that have frequent leader slots.  Maybe we just do something akin to tile timers.  We assume that we're gonna store 4096 leader slot waterfalls, and we point into that storage.

src/disco/gui/fd_gui.c
mmcgee-jump: Ah so it does!

PR #3229, opened 2024-10-25
gui: flush pack metrics before sampling

mmcgee-jump:


src/app/fdctl/run/tiles/fd_pack.c
```diff
@@ -376,6 +376,13 @@ after_credit( fd_pack_ctx_t *     ctx,
       fd_done_packing_t * done_packing = fd_chunk_to_laddr( ctx->out_mem, ctx->out_chunk );
       done_packing->microblocks_in_slot = ctx->slot_microblock_cnt;
 
+      /* We have a race window with the GUI, where if the slot is ending
+         it will snap these metrics to draw the waterfall, but see them
+         outdated because housekeeping hasn't run.  For now just update
+         them here, but PoH should eventually flush the pipeline before
+         ending the slot. */
+      metrics_write( ctx );
+
```
ptaffet-jump, line 385: Hmm sorry I mislead you. Actually this is not necessary because pack only updates the time estimate during housekeeping, so at least in the normal case, the leader->non-leader transition happens in the first `after_credit` after a housekeeping. I guess the strangeness with the metrics is not a race condition.

PR #3233, opened 2024-10-25
quic: remove ack bit vector

ripatel-fd:
Removes a bit vector from fd_quic_stream which remembers which
bytes were ACKed.  The stream frame handler only supports in-order
receive anyways.

Improves test_quic_bw performance 2.46 => 2.65


mmcgee-jump (approved):

ripatel-fd:
Closing for now

PR #3250, opened 2024-10-29
quic: rewrite service queue

ripatel-fd:
Greatly improves performance at high conn counts.

### DIFF
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -906,6 +906,9 @@ dynamic_port_range = "8900-9000"
         # longer than this threshold.
         idle_timeout_millis = 2000
 
+        # Max delay for outgoing ACKs.
+        ack_delay_millis = 50
+
         # QUIC retry is a feature to combat new connection request
         # spamming.  See rfc9000 8.1.2 for more details.  This flag
         # determines whether the feature is enabled in the validator.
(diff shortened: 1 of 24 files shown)


src/waltz/quic/fd_quic.c
```diff
@@ -463,14 +420,13 @@ fd_quic_init( fd_quic_t * quic ) {
...
-    FD_LOG_WARNING(( "NULL service_queue" ));
-    return NULL;
+  for( uint j=0U; j<FD_QUIC_SVC_CNT; j++ ) {
+    state->svc_queue[j].head = UINT_MAX;
+    state->svc_queue[j].tail = UINT_MAX;
   }
-  state->rb_service_queue_root = NULL;
+  state->svc_delay[ FD_QUIC_SVC_INSTANT ] = 0UL;
+  state->svc_delay[ FD_QUIC_SVC_ACK_TX  ] = (ulong) 50e6;
+  state->svc_delay[ FD_QUIC_SVC_WAIT    ] = (ulong)500e6;
```
ripatel-fd, line 429: This should be configurable

src/app/fdctl/config/default.toml
```diff
@@ -906,6 +906,10 @@ dynamic_port_range = "8900-9000"
         # longer than this threshold.
         idle_timeout_millis = 2000
 
+        # Max delay for outgoing ACKs.
+        # A value of zero selects the hardcoded default ACK delay.
```
mmcgee-jump, line 910: Suggest not doing this. The default is the value in this file, there shouldn't be some other second level default if you set it to zero. Also would an operator ever even want to configure this?

mmcgee-jump (approved):

PR #3268, opened 2024-10-30
gui: add mutli sparklines to frontend

mmcgee-jump:


ptaffet-jump (approved):
Looks entirely autogenerated

PR #3274, opened 2024-10-31
flamenco, vm: check signers seeds are of correct length in cpi

topointon-jump:
Because we check the signer seeds length in our CPI pre-flight checks, we need to perform the same length check Agave does in the address translation.


contrib/test/test-vectors-fixtures/cpi-fixtures/cpi.list
```diff
@@ -1,2 +1,3 @@
 dump/test-vectors/cpi/fixtures/callee_not_executable_log.fix
 dump/test-vectors/cpi/fixtures/cpi_prepare_missing_acct.fix
+dump/test-vectors/cpi/fixtures/18387a7e499df53e20db216f4520bca9ce0727e7_1582213.fix
```
ravyu-jump, line 3: In future the names of these fixtures should describe what exactly they test for (see above)

ravyu-jump (approved):

Issue #2205, opened 2024-06-25
Unbreak net tile parallelism


ripatel-fd:
Firedancer main as of abbe48de18469415df8bfd49b4a0f3eda561233b only supports running with one net tile.

There are two bugs preventing it from scaling:
- The /proc/pid/fd handover mechanism for BPF maps is broken (needs to be deleted) 
- quic<=>net topology code crashes

PR #3278, opened 2024-10-31
net: allow multiple net tiles

mmcgee-jump:
Closes https://github.com/firedancer-io/firedancer/issues/2205


src/waltz/xdp/fd_xdp1.c
ripatel-fd, line 1: Can we delete the other code?

ripatel-fd (dismissed):

ripatel-fd (approved):

PR #3282, opened 2024-11-01
quic: only support one DCID

ripatel-fd:
Decreases the fd_quic_conn_t footprint by about 600 bytes.

Removes support for the NEW_CONNECTION_ID frame.  fd_quic only ever
uses the first available connection ID anyways, and doesn't
implement RETIRE_CONNECTION_ID.  The conn ID state is thus dead
weight.


lidatong (approved):
so this is just removing support for connection id migration and stuff like that from our impl? if I recall from the rfc it’s for switching the same conn from mobile to desktop and it really isn’t FDs use case to have that so makes sense

ripatel-fd:
@lidatong Yes, should have clarified - This relates to the conn IDs that we can address the peer with, i.e. when we send outgoing packets. As of main today, we accept the 'NEW_CONNECTION_ID' frame and even store the new IDs we get. However, we always pick the first conn ID, and we don't support 'RETIRE_CONNECTION_ID'. 
This violates the spec and may be fixed in the future. Since we don't support 'active connection migration', there isn't really any reason to support conn ID rotation.

PR #3291, opened 2024-11-04
Use ppoll(2) instead of poll(2)

ripatel-fd:
The 'poll' Linux syscall is not available on some Linux architectures
like arm64, unlike 'ppoll'.  Since our seccomp policy compiler only
supports one policy across all architectures, use 'ppoll' instead.

Also fixes a make dependency issue that prevents 'make seccomp-policies'
from completing if the generated headers are broken.

Co-authored-by: Charles Moyes <cmoyes@jumptrading.com>


ripatel-fd:
I have a different idea that won't require changes to the x86 code

PR #3296, opened 2024-11-04
fuzz, flamenco, bpf: do not prematurely check executable status when validating programs

mjain-jump:
The program blacklist grows far too large in production when running in devnet / testnet. Functionality we've added in bpf cache reloading to mitigate this has other side effects, e.g. conflicts with read-only executable account promotion. It's simpler and more efficient for now to (unfortunately) tie our BPF cache back into consensus and query it when trying to execute the program.


src/flamenco/runtime/program/fd_bpf_loader_program.c
```diff
@@ -1916,10 +1910,19 @@ fd_bpf_loader_program_execute( fd_exec_instr_ctx_t * ctx ) {
...
-    /* This should NEVER fail. */
+    /* Sadly, we have to tie the cache in with consensus. We tried our best to avoid this,
+       but Agave's program loading logic is too complex to solely rely on checks without
+       significant redundancy. 
+       
+       For example, devnet and testnet have older programs that were deployed before stricter ELF / VM validation
+       checks were put in place, causing these older programs to fail newer validation checks and
+       be unexecutable. `fd_bpf_scan_and_create_bpf_program_cache_entry()` will populate our BPF program
+       cache correctly, but now, we have no way of checking if this validation passed or not here without
+       querying our program cache, otherwise we would have to copy-paste our validation checks here. */
```
ibhatt-jumptrading, line 1921: Only thing I would add here is some more information that is along the lines of:

"If we were to fail here, that implies that the program either failed to load or fails bytecode verification."

I think just be generally more explicit about what behavior we check for via this cache

ibhatt-jumptrading (approved):

PR #3298, opened 2024-11-04
fuzz: capture only writable accounts

yufeng-jump:


mjain-jump (commented):
Two things to do:
1. Regenerate fixtures with solfuzz-agave including your writable account capturing changes
2. Bump the test-vectors commit SHA on FD and re-run test vectors

mjain-jump (approved):

PR #3302, opened 2024-11-05
tpu: add resolv tile for resolving aluts

mmcgee-jump:


src/disco/bank/fd_bank_abi.c
```diff
@@ -266,10 +265,82 @@ is_upgradeable_loader_present( fd_txn_t * txn, uchar * payload, sanitized_txn_ab
...
+    result = fd_address_lookup_table_state_is_lookup_table( table );
+    if( FD_UNLIKELY( !result ) ) return FD_BANK_ABI_TXN_INIT_ERR_ACCOUNT_UNINITIALIZED;
+
+    if( FD_UNLIKELY( (data_sz-56UL)%32UL ) ) return FD_BANK_ABI_TXN_INIT_ERR_INVALID_ACCOUNT_DATA;
+
+    ulong addresses_len = (data_sz-56UL)/32UL;
+    fd_acct_addr_t const * addresses = fd_type_pun_const( data+56UL );
+
+    ulong deactivation_slot = table->inner.lookup_table.meta.deactivation_slot;
+    if( FD_UNLIKELY( deactivation_slot!=ULONG_MAX && (deactivation_slot+512UL)<slot ) ) return FD_BANK_ABI_TXN_INIT_ERR_ACCOUNT_NOT_FOUND;
```
ptaffet-jump, line 319: I'm not an expert on this, but I don't think this handles skipped slots properly. Since this is only the leader pipeline, it's not the end of the world if it rejects a tiny fraction of transactions that reference an address lookup table deactivated between 512 and 512*(1+skip rate) slots ago, but it at least needs a comment to that effect

ptaffet-jump (approved):

ptaffet-jump (approved):

PR #3306, opened 2024-11-05
add codeql to ci

two-heart:


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,48 @@
+name: "CodeQL PR Analysis"
+
+on:
+    workflow_call:
+    workflow_dispatch:
+
+jobs:
+  analyze:
+    name: Analyze (${{matrix.language}})
+    runs-on: 'ubuntu-latest'
```
ripatel-fd, line 10: If this check takes longer than 2 minutes or so, we should run it on-prem

two-heart: I agree, see slack channel. I just can't test on on-prem runners and don't want to PR something I am not sure is working.

topointon-jump (dismissed):
Following Anway's suggestion, we will run this on GitHub runners for now. Then when this is stable we will move it to on-prem.

anwayde (dismissed):

ravyu-jump:
@two-heart looks like the workflow script is broken?

two-heart:
Fixed the workflow file. Let me tune some queries such that we are maybe a bit faster.

ravyu-jump (approved):

PR #3321, opened 2024-11-05
flamenco: allow dead accounts to be borrowed

yufeng-jump:
Error code mismatch on a CPI fixture.  FD errors out as missing account, and agave errors out as account not executable.  This is due to a mismatch in account borrowing behavior.  FD does extra dead account check that agave doesn't.  This commit fixes that.

### DIFF
--- a/src/flamenco/runtime/context/fd_exec_txn_ctx.h
+++ b/src/flamenco/runtime/context/fd_exec_txn_ctx.h
@@ -161,6 +161,18 @@ int
 fd_txn_borrowed_account_view_idx( fd_exec_txn_ctx_t * ctx,
                                   uchar idx,
                                   fd_borrowed_account_t * * account );
+
+/* Same as above, except that this function doesn't check if the account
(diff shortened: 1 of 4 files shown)


src/flamenco/runtime/context/fd_exec_txn_ctx.h
```diff
@@ -161,6 +161,17 @@ int
 fd_txn_borrowed_account_view_idx( fd_exec_txn_ctx_t * ctx,
                                   uchar idx,
                                   fd_borrowed_account_t * * account );
+
+/* Same as above, except that this function doesn't check if the account
+   is dead (0 balance, 0 data, etc.) or not. When agave obtains a
+   borrowed account, it doesn't always check if the account is dead or
+   not. So this function allows us to more closely emulate that
+   behavior. */
```
topointon-jump, line 169: Should we add the permalink to this utility function too?

topointon-jump (approved):

ibhatt-jumptrading (approved):

PR #3329, opened 2024-11-06
quic: further optimize packet number parser

ripatel-fd:
For all valid QUIC v1 packets, there are at least 16 bytes after
the packet number field.  (The AES-GCM MAC tag)

Thus, we can use a trailing read to make fd_quic_pktnum_decode
branchless.


ptaffet-jump (commented):
Looks great! I'd keep `sz` a ulong. That's more normal, and I don't see any noticeable difference in the assembly it produces. I didn't do a fully exhaustive check though, so if you see a difference, I'm happy to stamp as is.

two-heart (approved):
lgtm from a security pov

PR #3343, opened 2024-11-06
http: fix treap removal issue

mmcgee-jump:


ptaffet-jump (approved):
I don't remember the full context, but this seems right

PR #3349, opened 2024-11-07
xdp: update bpf maps with multihome ips

anwayde:


ripatel-fd (commented):
Can we remove the term 'multihome' from the source? 
Just having a list of IP addresses should be enough.

lheeger-jump:
Lets address the immediate issue. We can rename in a followup.

lheeger-jump (approved):

PR #3352, opened 2024-11-07
quic: use random TLS public key

ripatel-fd:
Agave TPU clients ignore the server's TLS public key.
Thus, we can generate a random throwaway key on startup instead of
reusing the identity key.  This improves our security architecture
and removes a potential performance bottleneck with the keyguard tile.

### DIFF
--- a/contrib/tango.lua
+++ b/contrib/tango.lua
@@ -78,8 +78,6 @@ local link_hashes = {
 	[0xdb6a44] = "poh_shred",
 	[0xb8e9a5] = "crds_shred",
 	[0x6e5d41] = "shred_store",
-	[0x863a9a] = "quic_sign",
-	[0x9b7f2f] = "sign_quic",
 	[0xc650c9] = "shred_sign",
 	[0xd408b5] = "sign_shred"       
 }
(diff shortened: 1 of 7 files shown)


nbridge-jump (approved):
Nice
LGTM

mmcgee-jump (changes requested):
Can't you also remove FD_KEYGUARD_PAYLOAD_TLS_CV and all the related code (e.g. ambiguity proof etc?)

ripatel-fd:
> Can't you also remove FD_KEYGUARD_PAYLOAD_TLS_CV and all the related code (e.g. ambiguity proof etc?)

@mmcgee-jump I think we should keep this logic. We'll need it for client-side code in the future. When sending transactions to Agave servers, you get better quotas if you authenticate yourself with the identity key.

mmcgee-jump:
We can just bring the code back then? I don't mind leaving things in place that aren't security sensitive, but any surface area reduction in signing tile is a win.

two-heart (approved):
lgtm from a security pov

ripatel-fd:
> We can just bring the code back then? I don't mind leaving things in place that aren't security sensitive, but any surface area reduction in signing tile is a win.

@mmcgee-jump  I removed the authorize logic from the keyguard. The ambiguity detection logic should stay in, since anytime we change that, we have to rerun CBMC proofs. The full validator team is about to use this code any minute. They will need a QUIC client to implement the submitTransaction RPC. 

This is a big improvement as-is, let's merge it first? And if nobody uses that signing payload ambiguity logic in a month, we can always remove it later.

mmcgee-jump (approved):

PR #3355, opened 2024-11-07
quic: fix FD_QUIC_FRAME_TYPES off-by-one

two-heart:
\+ drive-by typo fixes

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -5798,7 +5798,7 @@ fd_quic_frame_handle_conn_close_0_frame(
     return FD_QUIC_PARSE_FAIL;
   }
 
-  /* the information here can be invaluble for debugging */
+  /* the information here can be invaluable for debugging */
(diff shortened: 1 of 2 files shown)


src/waltz/quic/fd_quic.c
```diff
@@ -859,7 +859,7 @@ fd_quic_handle_v1_frame( fd_quic_t *       quic,
 
   fd_quic_frame_context_t frame_context[1] = {{ quic, conn, pkt }};
 
-  /* Frame ID is technically a varint but it's sufficient to look at the
+  /* Frame ID is technically a variant, but it's sufficient to look at the
```
nlgripto, line 862:
```suggestion
  /* Frame ID is technically a varint, but it's sufficient to look at the
```

nbridge-jump:
Not a typo - padding is allowed. No objection tho

nbridge-jump (approved):
It's definitely better this way

PR #3359, opened 2024-11-08
resolv: filter expired transactions

mmcgee-jump:


src/app/fdctl/run/tiles/fd_resolv.c
```diff
@@ -167,10 +254,18 @@ unprivileged_init( fd_topo_t *      topo,
...
   ctx->round_robin_idx = tile->kind_id;
 
+  ctx->completed_slot = 0UL;
+  ctx->blockhash_ring_idx = 0UL;
+
   if( FD_LIKELY( !tile->kind_id ) ) _fd_ext_resolv_tile_cnt = ctx->round_robin_cnt;
 
   ctx->root_bank = NULL;
 
+  memset( &ctx->metrics, 0, sizeof( ctx->metrics ) );
```
ptaffet-jump, line 264: I think you should zero out the ring as well. I know for prod it will be in pre-zeroed pages, but it's safer for the first 4k slots when you're deleting a blockhash based on uninitialized memory.

src/app/fdctl/run/tiles/fd_resolv.c
```diff
@@ -133,6 +199,27 @@ after_frag( fd_resolv_ctx_t *   ctx,
   uchar    const * payload = dcache_entry;
   fd_txn_t const * txn     = (fd_txn_t const *)( dcache_entry + fd_ulong_align_up( payload_sz, 2UL ) );
 
+  /* If we can't find the recent blockhash ... it means one of two things,
+  
+     (1) It's really old (more than 28 minutes) or just non-existent.
+     (2) It's really new (we haven't seen the bank yet).
```
ptaffet-jump, line 205: Or it's a durable nonce transaction, but the logic for this is still right. Just update the comment.

src/app/fdctl/run/tiles/fd_resolv.c
```diff
@@ -119,11 +163,33 @@ after_frag( fd_resolv_ctx_t *   ctx,
...
+        ctx->root_slot = frag->slot;
+        break;
+      }
+      case 1: {
+        fd_completed_bank_t * frag = (fd_completed_bank_t *)ctx->_bank_msg;
+
+        blockhash_map_t * entry = map_query( ctx->blockhash_map, ctx->blockhash_ring[ ctx->blockhash_ring_idx%4096UL ], NULL );
+        if( FD_LIKELY( entry ) ) map_remove( ctx->blockhash_map, entry );
+
+        fd_memcpy( ctx->blockhash_ring[ ctx->blockhash_ring_idx%4096UL ].b, frag->hash, 32UL );
```
ptaffet-jump, line 181: Normal memcpy is probably better here, since size is small and known at compile time. The compiler normally has an easier time optimizing it.

ptaffet-jump (commented):
Some nits, but great!

ptaffet-jump (approved):

PR #3363, opened 2024-11-08
pack: schedule min size microblocks by default

ptaffet-jump:


ptaffet-jump:
This kills benchmark performance but seems to improve mainnet performance. Do we want to try to do something adaptive? It'll be hard to tune without more data.

mmcgee-jump (approved):

PR #3376, opened 2024-11-09
Add sandbox arm64 support

ripatel-fd:
The 'poll' Linux syscall is not available on some Linux architectures
like arm64, unlike 'ppoll'.

Also fixes a make dependency issue that prevents 'make seccomp-policies'
from completing if the generated headers are broken.

Co-authored-by: Charles Moyes <cmoyes@jumptrading.com>


src/util/sandbox/fd_sandbox.c
```diff
@@ -26,7 +26,7 @@
 #error "Target operating system is unsupported by seccomp."
 #endif
 
-#if !defined(__x86_64__)
+#if !defined(__x86_64__) && !defined(__aarch64__)
```
mmcgee-jump, line 29: Is this the only change needed? Doesn't the compiler need to produce a bit of different instructions in the programs as well?

mmcgee-jump: I know there are obscure bugs in a lot of seccomp programs around like, x86 syscall numbers having 2 variants and things like that.

riptl: @mmcgee-jump This is the only change needed to unbreak the arm64 build with default flags. There are certainly other problems with Arm support (assumptions on a TSO memory model in our codebase), and the seccomp policies are all untested. But we've had reports that 'fddev' ran fine with this patch at one point.

mmcgee-jump (approved):

marctrem (commented):
👍

PR #3381, opened 2024-11-11
deps: install rustup version as specified in agave submodule

anwayde:


deps.sh
```diff
@@ -287,13 +287,17 @@ check () {
...
         curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
         source "$HOME/.cargo/env"
-        rustup toolchain add 1.75.0
         ;;
       *)
         echo "[-] Skipping rustup install"
         ;;
     esac
   fi
+  rust_toolchain=$(grep "channel" agave/rust-toolchain.toml | awk '{print $NF}' | tr -d '"')
```
mmcgee-jump, line 296: Should this stay in the same "Y" choice response area though?

mmcgee-jump (changes requested):
.

deps.sh
anwayde: I put it outside thinking if someone already has rust installed but does not have the correct toolchain installed, we would run into the same weird race condition.. But if we think then that's the operator's choice (as they already installed rust), then I can put it back inside that block.

mmcgee-jump (approved):

PR #3390, opened 2024-11-11
flamenco: update slots_elapsed and program_id logic

kbhargava-jump:
summary:
* when `rent_epoch>epoch`, the slots elapsed should be 0 when calculating rent due
* when `program_account==fd_solana_native_loader_id`, agave has `account_indices` as [] so it fails when trying to add the instruction to the instruction stack with `UnsupportedProgramId`
* changing order in `close_lookup_table` checks and removing one extra check we had
* agave uses a `acccount_found` variable in their load_transactions_accounts and it is always truw if the program is `is_sysvar_instruction`


jumpsiegel (dismissed):

src/flamenco/runtime/fd_executor.c
```diff
@@ -399,14 +399,27 @@ fd_executor_load_transaction_accounts( fd_exec_txn_ctx_t * txn_ctx ) {
...
       return FD_RUNTIME_TXN_ERR_PROGRAM_ACCOUNT_NOT_FOUND;
     }
 
     /* https://github.com/anza-xyz/agave/blob/v2.0.9/svm/src/account_loader.rs#L307-309 */
     if( FD_UNLIKELY( !memcmp( program_account->pubkey->key, fd_solana_native_loader_id.key, sizeof(fd_pubkey_t) ) ) ) {
-      continue;
+      /* When program account pubkey is the native loader, Agave sets the program_indices of the LoadedTransaction to [] with
+         no errors until trying to push the specific instruction onto the instruction stack */
+      txn_ctx->null_program_id = 1;
+      return FD_RUNTIME_EXECUTE_SUCCESS;
```
mjain-jump, line 412: does this also happen when there are multiple instructions where NativeLoader is the program id of one of them? returning here seems sus considering that Agave continues to iterate over all of the instructions in the closure.

https://github.com/anza-xyz/agave/blob/ced98f1ebe73f7e9691308afa757323003ff744f/svm/src/account_loader.rs#L297-L300

kbhargava-jump: yah, you're right. maybe we need to just have a `program_indices` list as well in the transaction context

src/flamenco/runtime/fd_executor.c
```diff
@@ -363,6 +363,7 @@ fd_executor_load_transaction_accounts( fd_exec_txn_ctx_t * txn_ctx ) {
   fd_epoch_schedule_t const * schedule = fd_sysvar_cache_epoch_schedule( txn_ctx->slot_ctx->sysvar_cache );
   ulong                       epoch    = fd_slot_to_epoch( schedule, txn_ctx->slot_ctx->slot_bank.slot, NULL );
 
+  uchar accounts_found[txn_ctx->accounts_cnt];
```
mjain-jump, line 366: The usage of this looks a bit iffy to me. The default behavior in Agave is "found is true unless EXPLICITLY set here: https://github.com/anza-xyz/agave/blob/v2.0.9/svm/src/account_loader.rs#L251-L277"

Look at how the default value is set: https://github.com/anza-xyz/agave/blob/v2.0.9/svm/src/account_loader.rs#L311-L314

Your logic seems to be operating on an opposite assumption and I'm not entirely convinced it's correct. This needs a long comment explaining when exactly `accounts_found` is set to true vs false, and the logic should follow after.

mjain-jump: Maybe your logic is correct, but it's not entirely clear in what cases `account_found` is true or false. Can you clearly document this?

src/flamenco/runtime/fd_executor.c
```diff
@@ -394,31 +400,29 @@ fd_executor_load_transaction_accounts( fd_exec_txn_ctx_t * txn_ctx ) {
...
       return FD_RUNTIME_TXN_ERR_PROGRAM_ACCOUNT_NOT_FOUND;
     }
 
-    /* https://github.com/anza-xyz/agave/blob/v2.0.9/svm/src/account_loader.rs#L307-309 */
-    if( FD_UNLIKELY( !memcmp( program_account->pubkey->key, fd_solana_native_loader_id.key, sizeof(fd_pubkey_t) ) ) ) {
-      /* When program account pubkey is the native loader, Agave sets the program_indices of the LoadedTransaction to [] with
-         no errors until trying to push the specific instruction onto the instruction stack */
-      txn_ctx->null_program_id = 1;
-      return FD_RUNTIME_EXECUTE_SUCCESS;
+    /* https://github.com/anza-xyz/agave/blob/v2.0.9/svm/src/account_loader.rs#L307-L309 */
```
ibhatt-jumptrading, line 414: Make this comment more explicit + clearer

src/flamenco/runtime/fd_executor.c
```diff
@@ -394,31 +400,29 @@ fd_executor_load_transaction_accounts( fd_exec_txn_ctx_t * txn_ctx ) {
   ushort      program_owners_cnt = 0;
 
   /* https://github.com/anza-xyz/agave/blob/v2.0.9/svm/src/account_loader.rs#L297-L358 */
+
+  txn_ctx->has_program_id = 0;
   for( ushort i=0; i<instr_cnt; i++ ) {
     fd_txn_instr_t const * instr = &txn_ctx->txn_descriptor->instr[i];
-
+    // fd_instr_info_t      * instr_info = &txn_ctx->instr_infos[i];
```
ibhatt-jumptrading: remove

src/flamenco/runtime/context/fd_exec_txn_ctx.h
```diff
@@ -128,8 +128,7 @@ struct __attribute__((aligned(8UL))) fd_exec_txn_ctx {
 
   fd_spad_t * spad;
 
-  /* Used to track whether the transaction has program ids or not */
-  uint null_program_id;
+  uchar has_program_id;
```
ibhatt-jumptrading, line 131: comment

src/flamenco/runtime/fd_executor.c
kbhargava-jump: from older commit

src/flamenco/runtime/fd_executor.c
```diff
@@ -370,8 +370,11 @@ fd_executor_load_transaction_accounts( fd_exec_txn_ctx_t * txn_ctx ) {
   fd_epoch_schedule_t const * schedule = fd_sysvar_cache_epoch_schedule( txn_ctx->slot_ctx->sysvar_cache );
   ulong                       epoch    = fd_slot_to_epoch( schedule, txn_ctx->slot_ctx->slot_bank.slot, NULL );
 
+  uchar accounts_found[txn_ctx->accounts_cnt];
   for( ulong i=1UL; i<txn_ctx->accounts_cnt; i++ ) {
     fd_borrowed_account_t * acct = NULL;
+    /* https://github.com/anza-xyz/agave/blob/ced98f1ebe73f7e9691308afa757323003ff744f/svm/src/account_loader.rs#L220 */
+    accounts_found[i] = true;
```
mjain-jump, line 377: nit: accounts_found[i] = 1

src/flamenco/runtime/fd_executor.c
```diff
@@ -387,6 +390,19 @@ fd_executor_load_transaction_accounts( fd_exec_txn_ctx_t * txn_ctx ) {
       acct->starting_lamports = acct->meta->info.lamports;
     }
 
+    /* https://github.com/anza-xyz/agave/blob/v2.0.9/svm/src/account_loader.rs#L222-L277 */
+    int is_sysvar_instruction = !memcmp( acct->pubkey->key, fd_sysvar_instructions_id.key, sizeof(fd_pubkey_t) );
+    int account_exists = fd_acc_exists( acct->const_meta );
+    int is_fee_payer = (i == 0);
```
mjain-jump, line 396: you dont need this (look at the for loop bounds)

kbhargava-jump: thanks

mjain-jump (approved):

PR #3407, opened 2024-11-12
quic - fix key phase change

nbridge-jump:
This fixes an issue whereby the client initiates a key phase change, but it doesn't complete properly, and we lose the connection.
Also, it removes a line that broke the build.

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -3659,7 +3659,7 @@ fd_quic_conn_tx( fd_quic_t *      quic,
         uchar sb = conn->spin_bit;
 
         /* one_rtt has a short header */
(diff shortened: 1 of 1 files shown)


src/ballet/secp256r1/fd_secp256r1.c
```diff
@@ -1,7 +1,8 @@
 #include "fd_secp256r1.h"
 
 #include <stdint.h>
-#include <s2n-bignum.h>
+// this breaks the build:
+//#include <s2n-bignum.h>
```
mmcgee-jump, line 5: This is fixed by another PR landing now, you can revert

mmcgee-jump (dismissed):
.

ripatel-fd (approved):

PR #3412, opened 2024-11-13
verify: handle overrun from quic gracefully

mmcgee-jump:


ripatel-fd:
@mmcgee-jump We fixed the root cause here. I'd prefer to keep the strict checks in
https://github.com/firedancer-io/firedancer/pull/3414

PR #3427, opened 2024-11-14
config: add s2n-bignum submodule and integrate into build system

ripatel-fd:


ripatel-fd:
Exploring a different option that does not require a submodule

PR #3437, opened 2024-11-14
quic - ensure varint encoding handles overflow properly

nbridge-jump:
If a value beyond 2^62-1 was supplied to fd_quic_varint_encode, it would assume the 62 low order bits contained the value.
This fix defensively limits the value to at most 2^62-1, thereby avoiding truncation.

### DIFF
--- a/src/waltz/quic/templ/fd_quic_parse_util.h
+++ b/src/waltz/quic/templ/fd_quic_parse_util.h
@@ -6,15 +6,21 @@
 #include "../fd_quic_common.h"
 
 static inline uint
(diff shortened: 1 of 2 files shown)


ripatel-fd:
Hmm ... ideally we would never attempt to encode out of bound integers

nbridge-jump:
> Hmm ... ideally we would never attempt to encode out of bound integers

But we did. And who knows, maybe it could become part of an exploit.

ripatel-fd (approved):

PR #3444, opened 2024-11-15
quic: nits

two-heart:
Some stylistic improvements I found while auditing QUIC.


ripatel-fd (approved):
Thank you, this is very much appreciated.

PR #3456, opened 2024-11-16
Don't check in zstd compressed files

ripatel-fd:
Compress gui assets on startup to reduce git bloat.


src/app/fdctl/run/tiles/fd_gui.c
```diff
@@ -268,6 +269,30 @@ privileged_init( fd_topo_t *      topo,
...
+    char const * ext = strrchr( f->name, '.' );
+    if( !ext ) continue;
+    if( !strcmp( ext, ".html" ) ||
+        !strcmp( ext, ".css"  ) ||
+        !strcmp( ext, ".js"   ) ||
+        !strcmp( ext, ".svg"  ) ) {}
+    else continue;
+
+    ulong   zstd_bufsz = *f->data_len;
+    uchar * zstd_buf   = malloc( zstd_bufsz );
```
ripatel-fd, line 285: @mmcgee-jump I know usage of `malloc` is generally frowned upon. 
But this is one-time initialization code, so I think it gets a pass.

ripatel-fd: Would still be nice to have these files huge page backed

mmcgee-jump: I don't like using ZSTD in privileged init here. Can we move this code to `unprivileged_init`? Also you can just do `fd_wksp_alloc( fd_wksp_containing( gui_ctx ), zstd_bufsz )` to get the memory in the GUI gigantic pages. You could also set `loose_sz` to some minimum value to ensure these allocs succeed.

mmcgee-jump (changes requested):
.

src/app/fdctl/run/tiles/fd_gui.c
ripatel-fd: done

src/app/fdctl/run/tiles/fd_gui.c
```diff
@@ -270,6 +288,51 @@ privileged_init( fd_topo_t *      topo,
...
+    if( !ext ) continue;
+    if( !strcmp( ext, ".html" ) ||
+        !strcmp( ext, ".css"  ) ||
+        !strcmp( ext, ".js"   ) ||
+        !strcmp( ext, ".svg"  ) ) {}
+    else continue;
+
+    ulong   zstd_bufsz = fd_spad_alloc_max( spad, 1UL );
+    uchar * zstd_buf   = fd_spad_prepare( spad, 1UL, zstd_bufsz );
+    ulong zstd_sz = ZSTD_compress( zstd_buf, zstd_bufsz, f->data, *f->data_len, 19 );
```
ripatel-fd, line 321: This seems to do malloc() under the hood. Need to switch to the static compression API

src/app/fdctl/run/tiles/fd_gui.c
ripatel-fd: @mmcgee-jump I'm working on the change now. It has about 100 LOC additional code bloat.

src/app/fdctl/run/tiles/fd_gui.c
ripatel-fd: Fixed

mmcgee-jump (dismissed):

ptaffet-jump:
What about compressing them at compile time? Would that require another dependency?

mmcgee-jump (dismissed):

riptl (approved):

Issue #3461, opened 2024-11-16
fd_quic_tile tracer


ripatel-fd:
This is a tracking issue for an fddev subcommand that traces QUIC events to a log file.
Possibly in qlog format.

PR #3462, opened 2024-11-16
fddev quic-trace

ripatel-fd:
- **Add internal API for fd_quic_tile**
- **Add fddev quic-trace scaffold**

Relates to https://github.com/firedancer-io/firedancer/issues/3461

### DIFF
--- a/src/app/fddev/quic_trace/fd_quic_trace_main.c
+++ b/src/app/fddev/quic_trace/fd_quic_trace_main.c
@@ -16,13 +16,8 @@
 #include "fd_quic_trace.h"
 #include "../fddev.h"
 #include "../../fdctl/run/tiles/fd_quic_tile.h"
+#include "../../../disco/stem/fd_stem.h"
 #include "../../../tango/fseq/fd_fseq.h"
-
-/* Define global variables */
-
-fd_quic_ctx_t         fd_quic_trace_ctx;
-fd_quic_ctx_t const * fd_quic_trace_ctx_remote;
-ulong **              fd_quic_trace_target_fseq;
 
 void
 quic_trace_cmd_args( int *    pargc,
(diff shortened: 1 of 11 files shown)


src/app/fddev/quic_trace/fd_quic_trace_rx_tile.c
```diff
@@ -0,0 +1,159 @@
...
+
+#define STEM_BURST (1UL)
+#define STEM_CALLBACK_CONTEXT_TYPE  void
+#define STEM_CALLBACK_CONTEXT_ALIGN 1
+#define STEM_CALLBACK_BEFORE_FRAG   before_frag
+#define STEM_CALLBACK_DURING_FRAG   during_frag
+#define STEM_CALLBACK_AFTER_FRAG    after_frag
+#include "../../../disco/stem/fd_stem.c"
+
+fd_topo_run_tile_t fd_tile_quic_trace_rx = {
```
mmcgee-jump, line 155: I don't think you need this? Not used anywhere.

src/app/fddev/quic_trace/fd_quic_trace_main.c
```diff
@@ -0,0 +1,111 @@
...
+  /* Redirect metadata writes to dummy buffers.
+     Without this hack, stem_run would attempt to write metadata updates
+     into the target topology which is read-only. */
+
+  /* ... redirect metric updates */
+  ulong * metrics = aligned_alloc( FD_METRICS_ALIGN, FD_METRICS_FOOTPRINT( quic_tile->in_cnt, quic_tile->out_cnt ) );
+  if( !metrics ) FD_LOG_ERR(( "out of memory" ));
+  fd_metrics_register( metrics );
+
+  /* ... redirect fseq updates */
```
mmcgee-jump, line 98: I don't think this does anything? The QUIC tile has already. booted once you do this, so wouldn't read these fseq pointers again?

mmcgee-jump: Oh lol I see how this works ... kind of an interesting strategy to just pretend it's a QUIC tile, but I guess works for now.

ripatel-fd: @mmcgee-jump I don't see any other way. We need to consume net frags but we also need decryption keys. 
If we skip ahead of the real QUIC tile then the packets we see might not have encryption keys available.

src/app/fddev/quic_trace/fd_quic_trace_rx_tile.c
ripatel-fd: @mmcgee-jump I'd like to keep it. I want this to work as a real time eventually

src/app/fddev/quic_trace/fd_quic_trace_main.c
ripatel-fd: Oh I thought your comment was about something else. 
Eventually, I want to make it less hacky and make it a real tile. But I wasn't going to bother with these fdctl APIs just for a debugging tool

PR #3480, opened 2024-11-19
SIMD-0161: sbpf version

0x0ece:
Some impl details:
- At the end of the day we need SBPF version inside the VM, to customize its instructions (in future SIMDs/PRs)
- The VM is instantiated in 2 places: BPF Loader **deploy** and **execute**. Unfortunately, before instantiating the VM, they handle the program pretty differently (even though the variable name is `prog` in both functions, the type is different). Fortunately, they both use `fd_sbpf_elf_peek` to retrieve info from the ELF header, so we can tweak that function to extract the version.
- Note that in both deploy and execute (and also when we create prog, passed to execute) we have access to the slot, so we can check feature gates (feature gates are not specified yet, but they'll arrive soon)
- At the low SBPF level, everything is controlled by max_version. Right now max_version==1, i.e. we only accept ELF with version==1 (the SIMD-0161 is a bit ambiguous on this point, but both FD and Agave code prior to this PR are enforcing that version==1). As we roll out features, max_version will be progressively increased via feature gates.


src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -472,18 +474,21 @@ fd_sbpf_load_shdrs( fd_sbpf_elf_info_t *  info,
   return 0;
 }
 
-static int
```
0x0ece, line 475: This fn was static int, but the wrapped in a fn returning pointer. And not used anywhere else. So I figured to simplify and just keep the pointer one.

src/flamenco/vm/fd_vm_private.h
```diff
@@ -74,6 +74,30 @@ struct __attribute__((packed)) fd_vm_vec {
...
+#define FD_VM_SBPF_ENABLE_PQR                         (3UL)  /* SIMD-0174 */
+#define FD_VM_SBPF_DISABLE_NEG                        (3UL)
+#define FD_VM_SBPF_SWAP_SUB_REG_IMM_OPERANDS          (3UL)
+#define FD_VM_SBPF_EXPLICIT_SIGN_EXTENSION_OF_RESULTS (3UL)
+#define FD_VM_SBPF_STATIC_SYSCALLS                    (4UL)  /* SIMD-0176 */
+#define FD_VM_SBPF_STRICTER_CONTROLFLOW               (4UL)  /* SIMD-XXXX */
+#define FD_VM_SBPF_REJECT_RODATA_STACK_OVERLAP        (4UL)
+#define FD_VM_SBPF_ENABLE_ELF_VADDR                   (4UL)
+
+#define FD_VM_SBPF_HAS( vm, feature )   ((vm)->sbpf_version >= (feature))
```
0x0ece, line 99: This is not used yet, will be used in future SIMDs/PRs. I figured to include it now, we'll tweak it as we build features.

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -499,33 +504,29 @@ _fd_sbpf_elf_peek( fd_sbpf_elf_info_t * info,
...
   };
 
   fd_sbpf_elf_t const * elf = (fd_sbpf_elf_t const *)bin;
   int err;
 
   /* Validate file header */
-  if( FD_UNLIKELY( (err=fd_sbpf_check_ehdr( &elf->ehdr, elf_sz ))!=0 ) )
-    return err;
+  if( FD_UNLIKELY( (err=fd_sbpf_check_ehdr( &elf->ehdr, elf_sz, sbpf_max_version ))!=0 ) )
+    return NULL;
```
jumpsiegel, line 516: I worry a little bit that we are throwing away the err information here and this change will throw a bunch of fuzz errors that we will have to fix later... Was the returned err masked out in the caller?

src/ballet/sbpf/fuzz_sbpf_loader.c
```diff
@@ -34,7 +34,7 @@ LLVMFuzzerTestOneInput( uchar const * data,
                         ulong         size ) {
 
   fd_sbpf_elf_info_t info;
-  if( FD_UNLIKELY( !fd_sbpf_elf_peek( &info, data, size, 0 ) ) )
+  if( FD_UNLIKELY( !fd_sbpf_elf_peek( &info, data, size, 0, FD_SBPF_MAX_VERSION ) ) )
     return -1;
```
jumpsiegel, line 38: Yes, I guess it was masked out...

src/ballet/sbpf/fd_sbpf_loader.c
0x0ece: Agreed. We can bring there errors back again if needed. I think we're already fuzzing elf loader, so prob good for now. 🤞

PR #3494, opened 2024-11-19
slot 0 is a valid argument value for "wait for supermajority" when starting Agave validator

fkouteib:
**Problem:**
In Agave, 'validator_config.wait_for_supermajority' is of type option. So the default value is none not zero. Slot 0 is a valid setting for this config. So it needs to be passed in when starting agave if user requested in firedancer toml config. 

**Solution:**
Remove non-zero conditional check on appending '--wait-for-supermajority' to agave restart params.

Requesting this to be patched on Frankendancer release branch as well.

### DIFF
--- a/src/app/fdctl/run/run_agave.c
+++ b/src/app/fdctl/run/run_agave.c
@@ -75,11 +75,9 @@ agave_boot( config_t * config ) {
   if( !config->consensus.poh_speed_test ) ADD1( "--no-poh-speed-test" );
   if( strcmp( config->consensus.expected_genesis_hash, "" ) )
     ADD( "--expected-genesis-hash", config->consensus.expected_genesis_hash );
-  if( config->consensus.wait_for_supermajority_at_slot ) {
-    ADDU( "--wait-for-supermajority", config->consensus.wait_for_supermajority_at_slot );
-    if( strcmp( config->consensus.expected_bank_hash, "" ) )
-      ADD( "--expected-bank-hash", config->consensus.expected_bank_hash );
-  }
+  ADDU( "--wait-for-supermajority", config->consensus.wait_for_supermajority_at_slot );
+  if( strcmp( config->consensus.expected_bank_hash, "" ) )
+    ADD( "--expected-bank-hash", config->consensus.expected_bank_hash );
   if( config->consensus.expected_shred_version )
     ADDH( "--expected-shred-version", config->consensus.expected_shred_version );
   if( !config->consensus.wait_for_vote_to_start_leader )


fkouteib:
@mmcgee-jump any concerns with this change? Can it be included in the next FrD release?

mmcgee-jump:
Don't we need to update default.toml to set this to off by default, though? It looks like with this change it would, by default, wait for a supermajority at slot 0, rather than not waiting?

fkouteib:
Yes we do. My bad, I missed it as I was using my own config file. It looks like FD CI caught it as well. One of the CI tests failed for having wait_for_supermajority set but not having an expected bank hash set.

fkouteib:
@mmcgee-jump bumping this. CI failure and review feedback addressed. thanks.

src/app/fdctl/config/default.toml
```diff
@@ -514,12 +514,12 @@ dynamic_port_range = "8900-9000"
...
-    # If nonzero, after processing the ledger, and the next slot is the
+    # If set, after processing the ledger, and the next slot is the
     # provided value, wait until a supermajority of stake is visible on
     # gossip before starting proof of history.  This option is passed to
-    # the Agave client with the `--wait-for-supermajority`
-    # argument.
-    wait_for_supermajority_at_slot = 0
+    # the Agave client with the `--wait-for-supermajority` argument.
+    #
+    # wait_for_supermajority_at_slot = 0
```
mmcgee-jump: The TOML has no notion of "set" or "unset". The value will just default to zero anyway. Why is 0 a valid argument here?

mmcgee-jump (changes requested):
.

src/app/fdctl/config/default.toml
fkouteib: At high level, we want to enforce a specific bank hash and shred version at slot 0 for starting a test cluster with a new empty ledger we just created and captured the bank hash from.

The scenarios where I ran into a problem:
1. bootstrap from Agave with warmup epochs enabled in genesis, then have a mix of Agave and FD nodes join the cluster. Once FD nodes share exceeds Agave's, the cluster won't root slot 0 nor come online. Agave nodes wait for supermajority indefinitely. This was the original reason for this PR. We have since moved away from enabling warmup epochs for other reasons.
2. Recently, I tried to bootstrap a FD-only cluster from FD node [with (mostly) the same setup I use for Agave bootstrap nodes]. Once again hitting this problem because slot 0 never gets rooted and cluster never comes online.  Bootstrap node produces blocks up to block 92 then stalls. Other nodes skip their leader slots because they haven't landed a vote. With the latest iteration of the patch, this setup is now working and the cluster comes online.

At implementation level, this argument is stored as 'uint' in FD, but as 'Option' in Agave. So we can use 0 as a valid value, but FD uses 0 as equivalent to None to disable the feature.

I put up a new patch proposing that FD use UINT32_MAX as the "disable feature" value and allow all other positive values including zero as valid slot number to wait at.

PR #3500, opened 2024-11-20
flamenco: logfile level override via env var

nlgripto:


ravyu-jump:
This feels like something that should be handled by whatever is loading the .so

PR #3509, opened 2024-11-21
ci: improvements

kbhargava-jump:


.github/workflows/on_nightly.yml
```diff
@@ -10,3 +10,7 @@ jobs:
   coverage-report:
     uses: ./.github/workflows/coverage_report.yml
     secrets: inherit
+  codeql:
```
two-heart, line 13: This won't work. The CodeQL query just provides suggestions in a PR but does not raise an error, like for example unit tests do. See https://github.com/two-heart/firedancer/pull/2 for how this looks like. Instead, with the same reasoning, we should remove it from the merge queue tests.
If we really want to move it into nightly, which I am not fundamentally against, we need to enable SARIF upload again. Moving it to nightly with results upload would mean viewing results in the security tab instead of getting immediate inline feedback in the PR.
I tuned the CodeQL queries to run about as long as the ledger tests, and we would get significant speedup when moving it from a GitHub runner to an on-premise box. See comment in https://github.com/firedancer-io/firedancer/pull/3306. We could even reuse built artifact.

kbhargava-jump: talked offline and make necessary changes :)

PR #3515, opened 2024-11-21
flamenco: keep sysvar cache in sync with multiple epoch rewards sysvar updates within a slot

mjain-jump:
When distributing partitioned epoch rewards, there are multiple updates to the epoch rewards sysvar within a single slot, which updates the sysvar account but NOT the sysvar cache. When a subsequent update is performed within a single slot, the update ends up happening on a stale cache entry and causing the updated stale value to get written back into the account, causing a mismatching epoch rewards sysvar account. We should make sure to restore the sysvar cache before we perform subsequent updates on a sysvar within a single slot.

### DIFF
--- a/src/flamenco/runtime/sysvar/fd_sysvar_epoch_rewards.c
+++ b/src/flamenco/runtime/sysvar/fd_sysvar_epoch_rewards.c
@@ -49,6 +49,9 @@ fd_sysvar_epoch_rewards_read(
   return result;
 }
 
+/* Since there are multiple sysvar epoch rewards updates within a single slot,
+   we need to ensure that the cache stays updated after each change (versus with other
+   sysvars which only get updated once per slot and then synced up after) */
 void
 fd_sysvar_epoch_rewards_distribute(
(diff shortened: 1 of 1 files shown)


src/flamenco/rewards/fd_rewards.c
```diff
@@ -1071,6 +1071,8 @@ fd_distribute_partitioned_epoch_rewards(
 
     /* If we have finished distributing rewards, set the status to inactive */
     if ( fd_ulong_sat_add( height, 1UL ) >= distribution_end_exclusive ) {
+        /* Sync the epoch rewards sysvar cache entry with the account */
+        fd_sysvar_cache_restore_epoch_rewards( slot_ctx->sysvar_cache, slot_ctx->acc_mgr, slot_ctx->funk_txn );
```
topointon-jump, line 1075: Maybe we should put this inside `fd_sysvar_epoch_rewards_distribute` instead

topointon-jump (approved):

PR #3516, opened 2024-11-21
flamenco, vm: reorder callimm checks to match Agave

ravyu-jump:
To best-effort-match Agave's error path we re-arrange the order of checks in the callimm `0x85`


topointon-jump (approved):
Fantastic work 💪

PR #3529, opened 2024-11-22
SIMD-0166: SBPF dynamic stack frames

0x0ece:
[SIMD-0166](https://github.com/solana-foundation/solana-improvement-documents/blob/c5fba114ae4081ad39080f3c6bb62ee7e4dea9f3/proposals/0166-dynamic-stack-frames.md): SBPF Dynamic stack frames

Implementation details:
- Enabling dynamic stack frames automagically disables gaps. This should be coherent with the work done for direct mapping.
- Minor cleanup of the register checks, including R11. Agave has a specific error for R10 that we don't have, but since all errors are mismatched I don't think we really need to add one.
- Stack alignment set to 64, I made comments in SIMD-0166 about that. If we decide to reduce it, we just need to change it in the macro.

### DIFF
--- a/src/flamenco/vm/fd_vm.c
+++ b/src/flamenco/vm/fd_vm.c
@@ -408,10 +408,22 @@ fd_vm_validate( fd_vm_t const * vm ) {
     case FD_INVALID: default: return FD_VM_ERR_INVALID_OPCODE;
     }
 
-    if( FD_UNLIKELY( instr.src_reg>10 ) ) return FD_VM_ERR_INVALID_SRC_REG; /* FIXME: MAGIC NUMBER */
+    /* Check registers
+       https://github.com/solana-labs/rbpf/blob/v0.8.5/src/verifier.rs#L177 */
 
-    int is_invalid_dst_reg = instr.dst_reg > ((validation_code == FD_CHECK_ST) ? 10 : 9); /* FIXME: MAGIC NUMBER */
-    if( FD_UNLIKELY( is_invalid_dst_reg ) ) return FD_VM_ERR_INVALID_DST_REG;
+    /* Source register */
+    if( FD_UNLIKELY( instr.src_reg>10 ) ) return FD_VM_ERR_INVALID_SRC_REG;
+
+    /* Special R11 register */
+    if( instr.dst_reg==11U
+        && FD_VM_SBPF_DYNAMIC_STACK_FRAMES( sbpf_version )
+        && instr.opcode.raw == 0x07
+        && ( instr.imm % FD_VM_SBPF_DYNAMIC_STACK_FRAMES_ALIGN )==0 )
+        continue;
+
+    /* Destination register. */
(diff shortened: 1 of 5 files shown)


src/flamenco/vm/fd_vm.c
```diff
@@ -408,10 +408,22 @@ fd_vm_validate( fd_vm_t const * vm ) {
...
+    /* Special R11 register */
+    if( instr.dst_reg==11U
+        && FD_VM_SBPF_DYNAMIC_STACK_FRAMES( sbpf_version )
+        && instr.opcode.raw == 0x07
+        && ( instr.imm % FD_VM_SBPF_DYNAMIC_STACK_FRAMES_ALIGN )==0 )
+        continue;
+
+    /* Destination register. */
+    if( FD_UNLIKELY( instr.dst_reg==10U && validation_code != FD_CHECK_ST ) ) return FD_VM_ERR_INVALID_DST_REG;
+    if( FD_UNLIKELY( instr.dst_reg > 10U ) ) return FD_VM_ERR_INVALID_DST_REG;
```
topointon-jump, line 426: I think these destination register checks need to be conditioned on whether dynamic stack frames are enabled: https://github.com/solana-labs/rbpf/blob/c16589fcdfe8de65029a51bb28a72be0bbf29701/src/verifier.rs#L187

src/flamenco/vm/fd_vm.c
```diff
@@ -408,10 +408,22 @@ fd_vm_validate( fd_vm_t const * vm ) {
...
+       https://github.com/solana-labs/rbpf/blob/v0.8.5/src/verifier.rs#L177 */
+
+    /* Source register */
+    if( FD_UNLIKELY( instr.src_reg>10 ) ) return FD_VM_ERR_INVALID_SRC_REG;
+
+    /* Special R11 register */
+    if( instr.dst_reg==11U
+        && FD_VM_SBPF_DYNAMIC_STACK_FRAMES( sbpf_version )
+        && instr.opcode.raw == 0x07
+        && ( instr.imm % FD_VM_SBPF_DYNAMIC_STACK_FRAMES_ALIGN )==0 )
```
topointon-jump, line 421: Just so I understand, this check isn't in Agave yet, but will be added? Might be worth adding a TODO to go back and check that the implementation matches theirs once they have it.

topointon-jump (commented):
Nice! Only concern is the R10 register check - everything else looks great!

src/flamenco/vm/fd_vm.c
0x0ece: no, the conditions on all registers but r11 are true for all sbpf versions.
the line you linked corresponds to the line(s) below (and FD_VM_SBPF_DYNAMIC_STACK_FRAMES is part of our check).

src/flamenco/vm/fd_vm.c
0x0ece: the align is not in agave yet but they're adding it before release.

topointon-jump (approved):

PR #3535, opened 2024-11-25
quic: crude clock_gettime replacement

ripatel-fd:
clock_gettime consumes significant CPU time.
This patch hacks together an RDTSC based replacement.


riptl (approved):

ripatel-fd:
Tested this with a quic-go client and confirmed that ACKs arrive within 1ms tolerance of max_ack_delay

PR #3547, opened 2024-11-26
flamenco: fixing fee  payer writable account semantics

ibhatt-jumptrading:
1. writable accounts check was in the wrong location so I moved it
2. clarified writable account semantics
3. fixed fee payer handling -- the account only needs to be writable in the message. We don't care to check for demotion on the fee payer


src/flamenco/runtime/context/fd_exec_txn_ctx.c
```diff
@@ -259,3 +267,52 @@ void
 fd_exec_txn_ctx_reset_return_data( fd_exec_txn_ctx_t * txn_ctx ) {
   txn_ctx->return_data.len = 0;
 }
+
+/* https://github.com/anza-xyz/agave/blob/v2.1.1/sdk/program/src/message/versions/v0/loaded.rs#L162 */
```
ibhatt-jumptrading, line 271: for reviewer: i just moved these functions from fd_account to here.

ibhatt-jumptrading: and i updated the permalink and documentiation

topointon-jump (approved):

PR #3549, opened 2024-11-26
secp256r1: ecdsa sig verify

0x0ece:


src/ballet/ed25519/fd_ed25519_user.c
```diff
@@ -215,6 +218,7 @@ fd_ed25519_verify( uchar const   msg[], /* msg_sz */
   fd_ed25519_point_t Rcmp[1];
   fd_ed25519_point_neg( Aprime, Aprime );
   fd_ed25519_double_scalar_mul_base( Rcmp, k, Aprime, S );
+  // edwards25519_scalarmuldouble( (void *)Rcmp, (void *)k, (void *)Aprime, (void *)S );
```
ripatel-fd, line 221: Not sure if this is supposed to stay here or is just a comment for testing

0x0ece: oh no, this file shouldn't be here... now my secret project is reveled 🤣

ripatel-fd (approved):

PR #3568, opened 2024-11-27
flamenco, gossip: insert full crds value into value table on recv_crds

ravyu-jump:
We incorrectly only pushed the CRDS data, not the full value. This causes (among other things) push to break since it forwards values without signature encoded.


topointon-jump (approved):
🔥 🔥 🔥

PR #3575, opened 2024-11-28
runtime: remove unreachable code

ripatel-fd:
Cleans up error paths. Removes unreachable code. In some cases, replaces fatal errors with warnings.


ripatel-fd:
@topointon-jump I'm marking this as draft because I'm not confident that these changes are free of regressions. But hopefully these are useful regardless for future reference.

PR #3580, opened 2024-11-29
vm v2: add feature gates

0x0ece:
https://github.com/anza-xyz/agave/pull/3830

### DIFF
--- a/src/flamenco/runtime/program/fd_bpf_loader_program.c
+++ b/src/flamenco/runtime/program/fd_bpf_loader_program.c
@@ -252,7 +252,9 @@ deploy_program( fd_exec_instr_ctx_t * instr_ctx,
 
   /* Load executable */
   fd_sbpf_elf_info_t  _elf_info[ 1UL ];
-  fd_sbpf_elf_info_t * elf_info = fd_sbpf_elf_peek( _elf_info, programdata, programdata_size, deploy_mode, FD_SBPF_MAX_VERSION );
+  uint min_sbpf_version, max_sbpf_version;
+  fd_bpf_get_sbpf_versions( &min_sbpf_version, &max_sbpf_version, instr_ctx->slot_ctx );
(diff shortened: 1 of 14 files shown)


src/flamenco/runtime/program/fd_bpf_loader_program.c
```diff
@@ -252,7 +252,9 @@ deploy_program( fd_exec_instr_ctx_t * instr_ctx,
 
   /* Load executable */
   fd_sbpf_elf_info_t  _elf_info[ 1UL ];
-  fd_sbpf_elf_info_t * elf_info = fd_sbpf_elf_peek( _elf_info, programdata, programdata_size, deploy_mode, FD_SBPF_MAX_VERSION );
+  uint min_sbpf_version, max_sbpf_version;
+  fd_bpf_get_sbpf_versions( &min_sbpf_version, &max_sbpf_version, instr_ctx->slot_ctx );
+  fd_sbpf_elf_info_t * elf_info = fd_sbpf_elf_peek( _elf_info, programdata, programdata_size, deploy_mode, min_sbpf_version, max_sbpf_version );
```
topointon-jump, line 257: If a program specifies an SBPF version that hasn't been activated yet, what is the intended behaviour? From the code it looks like we treat this program as if it was an SBPF V0 program - is this intentional? Should we error out in this case, or attempt to execute it? What does Agave do?

0x0ece: if you try to load an invalid version, elf_info will return as NULL, like if any other validation fails.
Tests are in `test_sbpf_elf_peek`.

Lichtso:
During program deployment the minimum version must be raised to the maximum version:
https://github.com/anza-xyz/agave/pull/3830/files#diff-3617e23d40791b23531dc3bdab64afc5c8b783a712393b777e44f2ff41b9e3f4R250-R251

PR #3590, opened 2024-12-02
poh: add grace period for online nodes

mmcgee-jump:


src/app/fdctl/run/tiles/fd_poh.c
```diff
@@ -482,6 +471,9 @@ typedef struct {
   /* If an in progress frag should be skipped */
   int skip_frag;
 
+  ulong active_descendants_len;
+  ulong active_descendants[ 4096UL ];
```
ptaffet-jump, line 475: These are slot numbers? Can you add a comment about them?

src/app/fdctl/run/tiles/fd_poh.c
```diff
@@ -842,36 +847,36 @@ fd_ext_poh_reached_leader_slot( ulong * out_leader_slot,
...
-          /* We were the leader in the previous slot, so also no need for
-             a grace period.  We wouldn't get here if we were still
-             processing the prior slot so begin new one immediately. */
-          fd_ext_poh_write_unlock();
-          return 1;
-        }
+  long now_ns = fd_log_wallclock();
+  long expected_start_time_ns = ctx->reset_slot_start_ns + (long)((double)(ctx->next_leader_slot-ctx->reset_slot)*ctx->slot_duration_ns);
+
+  if( FD_UNLIKELY( now_ns<expected_start_time_ns+1200000000L ) ) {
```
ptaffet-jump, line 853: Does this 1.2 seconds need to be a multiple of `slot_duration_ns` or is it okay to be hardcoded?

src/app/fdctl/run/tiles/fd_poh.c
```diff
@@ -842,36 +847,36 @@ fd_ext_poh_reached_leader_slot( ulong * out_leader_slot,
...
+    int wait_graceful = 0;
+    for( ulong i=0UL; i<ctx->active_descendants_len; i++ ) {
+      if( FD_LIKELY( ctx->active_descendants[ i ]>=ctx->reset_slot && ctx->active_descendants[ i ]<ctx->next_leader_slot ) ) {
+        wait_graceful = 1;
+      } else if( FD_UNLIKELY( ctx->active_descendants[ i ]>=ctx->next_leader_slot ) ) {
+        /* Oops, we waited too long and a leader after us started
+           publishing to try and skip us.  Just start our slot
+           immediately, we might win ... */
+        wait_graceful = 0;
```
ptaffet-jump, line 862: I think we can do all this logic only knowing max(active_descendants), right?. That would make `fd_ext_poh_update_active_descendants` a lot simpler and faster

src/app/fdctl/run/tiles/fd_poh.c
```diff
@@ -842,36 +847,36 @@ fd_ext_poh_reached_leader_slot( ulong * out_leader_slot,
...
-    return 0;
+    if( FD_LIKELY( wait_graceful ) ) {
+      /* If one of the leaders between the reset slot and our leader
+         slot is in the process of publishing (they have a descendant
+         bank that is in progress of being replayed), then keep waiting.
+         We probably wouldn't get a leader slot out before they
+         finished.
+
+         Unless... we are past the deadline to start our slot by more
+         than 1.6 seconds, in which case we should probably start it to
```
ptaffet-jump, line 875: The constant is 1.2 seconds. Is that right?

ptaffet-jump (commented):
Seems like a good change, but I think it can be simplified

ptaffet-jump (approved):

PR #3594, opened 2024-12-03
pack: pace CUs through block

ptaffet-jump:
Currently implements a piecewise linear CU schedule, but it's factored so that it can be adjusted very easily.

### DIFF
--- a/src/ballet/pack/fd_pack_pacing.h
+++ b/src/ballet/pack/fd_pack_pacing.h
@@ -1,0 +1,78 @@
+#ifndef HEADER_fd_src_ballet_pack_fd_pack_pacing_h
+#define HEADER_fd_src_ballet_pack_fd_pack_pacing_h
+
+/* One of the keys to packing well is properly pacing CU consumption.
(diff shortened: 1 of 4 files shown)


src/ballet/pack/fd_pack_pacing.h
```diff
@@ -0,0 +1,78 @@
+#ifndef HEADER_fd_src_ballet_pack_fd_pack_pacing_h
+#define HEADER_fd_src_ballet_pack_fd_pack_pacing_h
+
+/* One of the keys to packing well is properly pacing CU consumption.
+   Without pacing, pack will end up filling the block with non-ideal
+   transactions.  Since at the current limits, the banks can execute a
+   block worth of CUs in a fraction of the block time, without pacing,
+   any lucrative transactions that arrive at towards the end of a block
```
mmcgee-jump, line 8: Can we clarify this comment? The case of lost revenue is mostly not due to valuable transactions arriving at the end, but due to all the value being in one conflicting set that doesn't get to execute very much.

mmcgee-jump (approved):

PR #3595, opened 2024-12-03
Add fd_avx_wh.h (16x ushort)

ripatel-fd:
Add AVX ushort wrappers for ops load, store, extract, insert, neg,
abs, min, max, add, sub, mul, eq, and ne.

### DIFF
--- a/src/ballet/reedsol/fd_reedsol_pi.c
+++ b/src/ballet/reedsol/fd_reedsol_pi.c
@@ -69,9 +69,6 @@
 #include "../../util/simd/fd_sse.h"
 
 #define ws_adjust_sign(a,b) _mm256_sign_epi16( (a), (b) ) /* scales elements in a by the sign of the corresponding element of b */
-#define ws_shl(a,imm)       _mm256_slli_epi16( (a), (imm) )
-#define ws_and(a,b)         _mm256_and_si256(    (a), (b) )
-#define ws_shru(a,imm)      _mm256_srli_epi16( (a), (imm) )
 
 static inline ws_t
 ws_mod255( ws_t x ) {
@@ -79,7 +76,7 @@ ws_mod255( ws_t x ) {
      (x%255) == 0xFF & ( x + (x*0x8081)>>23).
      We need at least 31 bits of precision for the product, so
      mulh_epu16 is perfect. */
-  return ws_and( ws_bcast( 0xFF ), ws_add( x, ws_shru( ws_mulhi( x, ws_bcast( (short)0x8081 ) ), 7 ) ) );
+  return ws_and( ws_bcast( 0xFF ), ws_add( x, ws_shru( wh_mulhi( x, ws_bcast( (short)0x8081 ) ), 7 ) ) );
 }
 
 /* The following macros implement the unscaled Fast Walsh-Hadamard
@@ -468,7 +465,7 @@ fd_reedsol_private_gen_pi_16( uchar const * is_erased,
      (x%255) == 0xFF & ( x + (x*0x8081)>>23).
      We need at least 31 bits of precision for the product, so
      mulh_epu16 is perfect. */
-  log_pi = ws_and( ws_bcast( 0xFF ), ws_add( log_pi, ws_shru( ws_mulhi( log_pi, ws_bcast( (short)0x8081 ) ), 7 ) ) );
+  log_pi = ws_and( ws_bcast( 0xFF ), ws_add( log_pi, ws_shru( wh_mulhi( log_pi, ws_bcast( (short)0x8081 ) ), 7 ) ) );
   /* Now 0<=log_pi < 255 */
 
   /* Since our FWHT implementation is unscaled, we've computed a value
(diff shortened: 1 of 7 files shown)


src/util/simd/test_avx_common.c
```diff
@@ -415,6 +415,73 @@ int wv_test( wv_t v, ulong v0, ulong v1, ulong v2, ulong v3 ) {
...
+  wh_stu( m+16,  h ); /* Unaligned store to aligned   */
+  wh_stu( m+33,  h ); /* Unaligned store to aligned+1 */
+  wh_stu( m+50,  h ); /* Unaligned store to aligned+2 */
+  wh_stu( m+67,  h ); /* Unaligend store to aligned+3 */
+  wh_stu( m+84,  h ); /* Unaligned store to aligned+4 */
+  wh_stu( m+101, h ); /* Unaligned store to aligned+5 */
+  wh_stu( m+118, h ); /* Unaligned store to aligned+6 */
+  wh_stu( m+135, h ); /* Unaligned store to aligned+7 */
+
+  l = wh_ld(  m     ); if( wc_pack( wh_eq( h, l ) )!=255 ) return 0;
```
ptaffet-jump, line 452: I don't think this is right, since `wh_eq` doesn't return a proper conditional. This will only look at 8 of the comparisons, not 16.

src/util/simd/fd_avx_wh.h
```diff
@@ -0,0 +1,93 @@
...
+
+#define wh_neg(a) _mm256_sub_epi16( _mm256_setzero_si256(), (a) ) /* [ -a0  -a1  ... -a15]   (twos complement handling) */
+#define wh_abs(a) (a)                                             /* [ |a0| |a1| ... |a15| ] (twos complement handling) */
+
+#define wh_min(a,b) _mm256_min_epu16(   (a), (b) ) /* [ min(a0,b0) min(a1,b1) ... min(a15,a15) ] */
+#define wh_max(a,b) _mm256_max_epu16(   (a), (b) ) /* [ max(a0,b0) max(a1,b1) ... max(a15,a15) ] */
+#define wh_add(a,b) _mm256_add_epi16(   (a), (b) ) /* [ a0+b0      a1+b1      ... a15+b15      ] */
+#define wh_sub(a,b) _mm256_sub_epi16(   (a), (b) ) /* [ a0-b0      a1-b1      ... a15-b15      ] */
+#define wh_mul(a,b) _mm256_mullo_epi16( (a), (b) ) /* [ a0*b0      a1*b1      ... a15*b15      ] */
+
```
ptaffet-jump, line 89: Would `_mm256_mulhi_epu16` be useful too?

ripatel-fd: @ptaffet-jump I have no uses for it currently. As far as I know, we don't have mulhi wrappers for other lane widths. Let's add it in a follow up PR once needed.

src/util/simd/test_avx_common.c
ripatel-fd: Good point. I should use _mm256_movemask_epi8 instead of wc_pack here.

ripatel-fd: Fixed

src/util/simd/fd_avx_wh.h
ripatel-fd: Added mulhi

src/ballet/reedsol/fd_reedsol_pi.c
```diff
@@ -68,22 +68,7 @@
 
 #include "../../util/simd/fd_sse.h"
 
-#define ws_t __m256i
-#define ws_add(a,b)         _mm256_add_epi16( (a), (b) )
-#define ws_sub(a,b)         _mm256_sub_epi16( (a), (b) )
-#define ws_bcast(s0)        _mm256_set1_epi16( (s0) )
 #define ws_adjust_sign(a,b) _mm256_sign_epi16( (a), (b) ) /* scales elements in a by the sign of the corresponding element of b */
-#define ws_mullo(a,b)       _mm256_mullo_epi16( (a), (b) )
-#define ws_mulhi(a,b)       _mm256_mulhi_epu16( (a), (b) )
```
ripatel-fd, line 77: @ptaffet-jump fd_avx_ws.h uses `_mm256_mulhi_epi16` (sign extend a, b before multiplying) 
What should we do here? Maybe just change reedsol_pi.c to use wh_mulhi?

ptaffet-jump: I think reedsol should be using ushort, but I'm not sure why it says short everywhere. I need to try to refresh my memory.

ripatel-fd: @ptaffet-jump You probably had ws=>ushort when you wrote the code. I picked ws=>short and wh=>ushort when I wrote my wrappers. Then I realized you already picked ws. Please let me know what we should do about this

ripatel-fd: Changed fd_reedsol_pi to use wh_mulhi

ptaffet-jump (approved):
I first thought that your change to reedsol_pi was a strange way of doing it, but now I think it's the right way to do it. The vectors are truly signed shorts, but before each mulhi, I ensured that the values were all >= 0. Thanks for this fix!

PR #3630, opened 2024-12-04
tpu: a few more metrics/diagnostic visibility improvements

ptaffet-jump:
Split into multiple commits because we might want to roll back some of these in the future


src/disco/metrics/metrics.xml
```diff
@@ -319,6 +319,7 @@ metric introduced.
   <counter name="NoBankDrop" summary="Count of transactions dropped because the bank was not available" />
   <counter name="LutResolved" enum="LutResolveResult" summary="Count of address lookup tables resolved" />
   <counter name="BlockhashExpired" summary="Count of transactions that failed to resolve because the blockhash was expired" />
+  <counter name="BlockhashUnknown" summary="Count of transactions with an unknown blockhash. These may be very recent, very old, nonces, or bogus." />
```
mmcgee-jump, line 322: Can you add this to existing GUI sankey thing as well (no need for frontend changes, just report it to the same bucket).

mmcgee-jump: Oh nvm

mmcgee-jump (approved):

PR #3656, opened 2024-12-06
flamenco, vm: (SIMD-0182) consume requested CUs for VM Interp failures

ravyu-jump:
WIP: Do not merge until
- [x] Anza creates separate feature gate for this: https://github.com/anza-xyz/agave/pull/3994
- [x] ~~Generate fixtures + fuzz runs~~ Will be done once Anza cuts a release for the feature gate

Implements [SIMD-0182](https://github.com/solana-foundation/solana-improvement-documents/pull/182).

[Agave implementation](https://github.com/anza-xyz/agave/pull/3751/files) for reference


src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -1131,18 +1131,18 @@
...
-sigexit:     /* err current */         /* ic current */      /* cu current */  goto interp_halt;
+sigtext:     err = FD_VM_ERR_SIGTEXT;     FD_VM_INTERP_FAULT;                     goto interp_halt;
+sigtextbr:   err = FD_VM_ERR_SIGTEXT;     /* ic current */      /* cu current */  goto interp_halt;
+sigsplit:    err = FD_VM_ERR_SIGSPLIT;    FD_VM_INTERP_FAULT;                     goto interp_halt;
+sigcall:     err = FD_VM_ERR_SIGCALL;     /* ic current */      /* cu current */  goto interp_halt;
+sigstack:    err = FD_VM_ERR_SIGSTACK;    /* ic current */      /* cu current */  goto interp_halt;
+sigill:      err = FD_VM_ERR_SIGILL;      FD_VM_INTERP_FAULT;                     goto interp_halt;
+sigsegv:     err = FD_VM_ERR_SIGSEGV;     FD_VM_INTERP_FAULT;                     goto interp_halt;
+sigcost:     err = FD_VM_ERR_SIGCOST;     /* ic current */      cu = 0UL;         goto interp_halt;
+sigsyscall:  err = FD_VM_ERR_SIGSYSCALL;  /* ic current */      /* cu current */  goto interp_halt;
```
0x0ece, line 1142: I think that this changes the semantic? Syscalls return a typed error. If we change it to "syscall error", then we loose the type.

FWIW I think the SIMD is very unclear because they rely on SyscallError which is a rust thing, and it's also very unclear when it's thrown vs not. For example, memory violation inside a syscall... is that syscall error or ebpf!?

If we want to avoid syscalls (and I think we should), I think we need to store another var that says "error_because_of_syscall", not reuse the existing err.

ravyu-jump: > If we want to avoid syscalls (and I think we should), I think we need to store another var that says "error_because_of_syscall", not reuse the existing err.

This `FD_VM_ERR_SIGSYSCALL` is a new VM error code specifically for the purpose you mentioned. Do we need a separate variable?

> FWIW I think the SIMD is very unclear because they rely on SyscallError which is a rust thing, and it's also very unclear when it's thrown vs not. For example, memory violation inside a syscall... is that syscall error or ebpf!?

Yeah it's a little confusing. I am taking it as "read the error code output of the VM execution and determine if it was caused by a syscall" situation

0x0ece: Makes sense

0x0ece (dismissed):

topointon-jump (approved):

src/flamenco/vm/fd_vm_interp_core.c
topointon-jump: Yeah, it's about whether or not the error was thrown in the middle of a basic block or not. Syscall errors are thrown at boundaries of basic blocks, so we don't have to set consume all the requested in the failure cases.

> For example, memory violation inside a syscall... is that syscall error or ebpf!?

This is classified as a syscall error because of _where_ it was thrown. I do agree that the SIMD can be made clearer.

PR #3658, opened 2024-12-09
codeql: add fd_value_elem data encoding check

two-heart:
catch bugs like the one fixed in #3568

Credits the query idea: @ravyu-jump


ravyu-jump:
Amazing. Thank you so much !

PR #3660, opened 2024-12-09
Epoch opt

jumpsiegel:


src/flamenco/runtime/fd_runtime.c
```diff
@@ -4445,6 +4403,8 @@ void fd_process_new_epoch(
     fd_exec_slot_ctx_t *slot_ctx,
     ulong parent_epoch )
 {
+  FD_LOG_NOTICE(( "fd_process_new_epoch start" ));
```
topointon-jump, line 4406: nit: remove

topointon-jump (commented):
Looks great! Next step is definitely to get rid of the delegations cache

topointon-jump (approved):

PR #3668, opened 2024-12-10
Fix fd_redblack symbol visibility

ripatel-fd:
Fixes an issue where fd_redblack produces externally linked symbols
even if the implementation style is set to 'local use only'.


src/util/tmpl/fd_redblack.c
```diff
@@ -404,6 +404,12 @@ FD_PROTOTYPES_END
 
 #endif /* REDBLK_IMPL_STYLE==0 || REDBLK_IMPL_STYLE==1 */
```
anwayde, line 405: minor nit: change comment to match

anwayde (approved):

PR #3675, opened 2024-12-11
tile, gossip: check that incoming vote txns are single instruction

ravyu-jump:
Addresses: https://github.com/firedancer-io/auditor-internal/issues/255


topointon-jump (approved):

ravyu-jump:
~~Going to pipe vote txns through verify, switching this to draft for now~~
Doing that in a separate PR

PR #3680, opened 2024-12-11
events: add variable entry size circular buffer

mmcgee-jump:


src/disco/events/fd_circbuf.h
```diff
@@ -0,0 +1,65 @@
...
+struct __attribute__((aligned(FD_CIRCBUF_ALIGN))) fd_circbuf_private {
+  ulong cnt;
+  ulong head;
+  ulong tail;
+  ulong size;
+
+  struct {
+    ulong drop_cnt;
+  } metrics;
+};
```
ptaffet-jump, line 17:
```suggestion
  /* padding out to 4k here ... */
};
```
I missed the fact this struct is so large and was confused about alignment.

src/disco/events/fd_circbuf.h
```diff
@@ -0,0 +1,65 @@
...
+fd_circbuf_delete( void * shbuf );
+
+/* fd_circbuf_push_back appends a message of size sz into the circular
+   buffer, evicting any old messages if they would be overwritten when
+   the buffer wraps around.  Returns the address of the memory contents
+   in the buffer on success, or NULL on failure.  The only reason for
+   failure is if the requested sz (along with the message metadata)
+   exceeds the size of the entire buffer and can't fit. */
+
+uchar *
```
ptaffet-jump, line 49: I think you need that `align<=4096`

src/disco/events/fd_circbuf.c
```diff
@@ -0,0 +1,160 @@
...
+
+  ulong required = fd_ulong_align_up( sizeof( fd_circbuf_message_t ), align ) + footprint;
+  if( FD_UNLIKELY( required>circbuf->size ) ) {
+    FD_LOG_WARNING(( "tried to push message which was too large %lu>%lu", required, circbuf->size ));
+    return NULL;
+  }
+
+  uchar * buf = (uchar *)(circbuf+1);
+
+  ulong next = 0UL;
```
ptaffet-jump, line 123: This seems more like `current` than `next`. I was confused for a bit

src/disco/events/fd_circbuf.c
```diff
@@ -0,0 +1,160 @@
+#include "fd_circbuf.h"
+
+struct __attribute__((aligned(8UL))) fd_circbuf_message_private {
+  ulong align;
+  ulong footprint;
+  ulong next;
```
ptaffet-jump, line 6: Does the next field store unique information? Is it just current+footprint with the proper alignment?

src/disco/events/fd_circbuf.h
```diff
@@ -0,0 +1,65 @@
...
+#define HEADER_fd_src_disco_events_fd_circbuf_h
+
+#include "../fd_disco_base.h"
+
+#define FD_CIRCBUF_ALIGN (4096UL)
+
+struct __attribute__((aligned(FD_CIRCBUF_ALIGN))) fd_circbuf_private {
+  ulong cnt;
+  ulong head;
+  ulong tail;
```
ptaffet-jump, line 11: I think it's worth adding some comments. These are the offsets relative to the end of this struct of the first and last message, respectively. In data structures like this, tail is often a pointer to after the end of the last message, so it's worth noting that this is to the start of the last message

src/disco/events/fd_circbuf.h
```diff
@@ -0,0 +1,65 @@
+#ifndef HEADER_fd_src_disco_events_fd_circbuf_h
+#define HEADER_fd_src_disco_events_fd_circbuf_h
+
+#include "../fd_disco_base.h"
+
+#define FD_CIRCBUF_ALIGN (4096UL)
+
+struct __attribute__((aligned(FD_CIRCBUF_ALIGN))) fd_circbuf_private {
+  ulong cnt;
```
ptaffet-jump, line 9: Non-evicted count

src/disco/events/fd_circbuf.c
```diff
@@ -0,0 +1,160 @@
...
+    if( FD_UNLIKELY( (start<to && end>from) ) ) {
+      circbuf->cnt--;
+      circbuf->metrics.drop_cnt++;
+      if( FD_LIKELY( !circbuf->cnt ) ) circbuf->head = circbuf->tail = 0UL;
+      else                             circbuf->tail = tail->next;
+
+      continue;
+    }
+
+    break;
```
ptaffet-jump, line 98: `else break;` seems a bit more normal to me, but I'm not sure if you have a reason for writing it this way

src/disco/events/test_circbuf.c
```diff
@@ -0,0 +1,123 @@
+#include "../fd_disco.h"
+#include "fd_circbuf.h"
+
+void
+test_simple1( void ) {
+  uchar buf[ 32UL+1024UL ];
```
ptaffet-jump, line 6: Doesn't it need to be 4k+128?

src/disco/events/test_circbuf.c
```diff
@@ -0,0 +1,123 @@
...
+  fd_circbuf_t * circbuf = fd_circbuf_join( fd_circbuf_new( buf, 128 ) );
+  FD_TEST( circbuf );
+
+  fd_rng_t _rng[1];
+  fd_rng_t * rng = fd_rng_join( fd_rng_new( _rng, 0U, 0UL ) );
+
+  for( ulong i=0UL; i<8192UL*8192UL; i++ ) {
+    uchar * msg = fd_circbuf_push_back( circbuf, fd_ulong_pow2( (int)fd_rng_ulong_roll( rng, 5 ) ), 1UL+fd_rng_ulong_roll( rng, 64 ) );
+    FD_TEST( msg );
```
ptaffet-jump, line 15: What about calling verify here?

src/disco/events/fd_circq.h
```diff
@@ -0,0 +1,99 @@
...
+   One more complication is that the circular buffer must store
+   metadata about the messages in the data buffer itself, as it does not
+   have a separate metadata region.  The structure of the buffer then
+   looks as follows:
+
+    +-------+-----+------+-----+-------+-----+-------+
+    + meta0 | pad | msg0 | pad | meta1 | pad | meta2 |
+    +-------+-----+------+-----+-------+-----+-------+
+     ^  |           ^  |         ^  |          ^  ^
+     |  +-----next--+  +---next--+  +----------+  |
```
ptaffet-jump, line 21: Is this line right? `meta0->next` should be `meta1`?

PR #3685, opened 2024-12-12
gui: skip rate and first processed slot

yufeng-jump:


src/disco/gui/fd_gui_printf.c
```diff
@@ -60,12 +60,23 @@ jsonp_long( fd_gui_t *   gui,
   else                   fd_http_server_printf( gui->http, "%ld,", value );
 }
 
+#define JSONP_DOUBLE_N(n)                                                                     \
+FD_FN_UNUSED static void                                                                      \
+jsonp_double_##n(fd_gui_t *   gui,                                                            \
```
mmcgee-jump, line 65: Why? Just print the whole thing at full precision, the frontend can round

mmcgee-jump (changes requested):
.

src/disco/gui/fd_gui_printf.c
mmcgee-jump: Actually I don't even know why we are printing skip rate here. Just print two ulongs, `slots_processed`, `slots_skipped`

mmcgee-jump (changes requested):
.

PR #3692, opened 2024-12-12
tile, gossip: rewire dedup out to verify

ravyu-jump:


src/app/fdctl/run/tiles/fd_dedup.c
```diff
@@ -16,8 +16,7 @@
    checks the transaction signature field for duplicates and filters
    them out. */
 
-#define GOSSIP_IN_IDX (0UL) /* Frankendancer and Firedancer */
-#define VOTER_IN_IDX  (1UL) /* Firedancer only */
+#define VOTER_IN_IDX  (0UL) /* Firedancer only */
```
mmcgee-jump, line 19: Doesn't this break frankendancer?

mmcgee-jump (changes requested):
.

src/app/fdctl/run/tiles/fd_dedup.c
ravyu-jump: yes it does lol

src/app/fdctl/run/tiles/fd_dedup.c
```diff
@@ -16,8 +16,17 @@
    checks the transaction signature field for duplicates and filters
    them out. */
 
-#define GOSSIP_IN_IDX (0UL) /* Frankendancer and Firedancer */
-#define VOTER_IN_IDX  (1UL) /* Firedancer only */
+#ifdef FD_HAS_NO_AGAVE
```
mmcgee-jump, line 19: Not the correct way to do this ... why can't you keep the current mechanism?

mmcgee-jump (changes requested):
.

src/app/fdctl/run/tiles/fd_dedup.c
ravyu-jump: > why can't you keep the current mechanism?

do you mean:
```
#define GOSSIP_IN_IDX (0UL) /* Frankendancer only */
#define VOTER_IN_IDX  (0UL) /* Firedancer only */
```

Without topology awareness we need to make the check for gossip link optional. The current check (in `unprivileged_init`) makes it mandatory for both topologies. I didn't want to break that assumption so I went this way. But if it's not actually a strict requirement I can change that. 

I'm also curious: what's the correct way to weave in topology-aware blocks of code?

src/app/fdctl/run/tiles/fd_dedup.c
```diff
@@ -163,7 +161,7 @@ after_frag( fd_dedup_ctx_t *    ctx,
     if( FD_UNLIKELY( !txn_t_sz ) ) FD_LOG_ERR(( "fd_txn_parse failed for vote transactions that should have been sigverified" ));
```
topointon-jump, line 161: Erroring out here is now dangerous, as unparsed transactions can now come from gossip, so there is no guarantee that the parsing will succeed.

We should probably either call `fd_txn_parse` inside the gossip tile (in `verify_vote_txn`?), or not error out here. I think the first option is probably better.

topointon-jump (dismissed):

src/app/fdctl/run/tiles/fd_gossip.c
```diff
@@ -317,15 +317,23 @@ gossip_deliver_fun( fd_crds_data_t * data, void * arg ) {
       return;
     }
 
-    uchar * vote_txn_msg = fd_chunk_to_laddr( ctx->dedup_out_mem, ctx->dedup_out_chunk );
+    uchar * vote_txn_msg = fd_chunk_to_laddr( ctx->verify_out_mem, ctx->verify_out_chunk );
     ulong vote_txn_sz    = gossip_vote->txn.raw_sz;
     memcpy( vote_txn_msg, gossip_vote->txn.raw, vote_txn_sz );
 
+    /* DEFENSE-IN-DEPTH: While the underlying txn message is technically verifed
```
mmcgee-jump, line 324: This comment doesn't seem to make sense. You've only verified one signature with the CRDS message? And it might be for bytes that aren't the actual transaction bytes. Also what does the TODO mean? What impact would it have on "verify traffic" ?

mmcgee-jump: There is no performance impact of sending gossiped votes to sigverify, it's not a material volume, and you need to verify them... verifying the CRDS is not the same as verifying the transaction.

src/app/fdctl/run/tiles/fd_dedup.c
```diff
@@ -212,32 +211,8 @@ unprivileged_init( fd_topo_t *      topo,
...
-    }
-
-    /* ...followed by a sequence of verify_dedup links */
-    for( ulong i=unparsed_in_cnt; i<tile->in_cnt; i++ ) {
-      if( FD_UNLIKELY( strcmp( topo->links[ tile->in_link_id[ i ] ].name, "verify_dedup" ) ) ) {
-        FD_LOG_ERR(( "dedup tile has unexpected input links %lu %lu %s",
-                     tile->in_cnt, i, topo->links[ tile->in_link_id[ i ] ].name ));
-      }
-    }
+  if( FD_UNLIKELY( tile->in_cnt == 0UL ) ) {
```
mmcgee-jump, line 214: I wouldn't even bother to check this, topology constructor can do what they like.

ravyu-jump: sure

ravyu-jump:
Marking this as draft until this is merged in: https://github.com/firedancer-io/firedancer/pull/3763

mmcgee-jump (approved):

PR #3699, opened 2024-12-13
funk: leaving tombstones in published records

ibhatt-jumptrading:
Now funk records when removed will not lose their data and will be published into the root of funk. However, they will contain the erase flag. This will dneote that they will be tombstones. This is needed for snapshotting as well as for debugging

### DIFF
--- a/src/funk/fd_funk_rec.c
+++ b/src/funk/fd_funk_rec.c
@@ -320,7 +320,14 @@ fd_funk_rec_insert( fd_funk_t *               funk,
     fd_funk_rec_t * rec = fd_funk_rec_map_query( rec_map, pair, NULL );
 
     if( FD_UNLIKELY( rec ) ) { /* Already a record present */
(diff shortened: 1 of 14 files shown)


src/funk/fd_funk_rec.c
```diff
@@ -319,8 +319,17 @@ fd_funk_rec_insert( fd_funk_t *               funk,
...
 
+
+
     if( FD_UNLIKELY( rec ) ) { /* Already a record present */
-      if( FD_UNLIKELY( rec->flags & FD_FUNK_REC_FLAG_ERASE ) ) FD_LOG_CRIT(( "memory corruption detected (bad flags)" ));
+
+      /* However, if the record is marked for erasure, reset the flag and
+         return the record. */
+      if( rec->flags & FD_FUNK_REC_FLAG_ERASE ) {
+        rec->flags &= ~FD_FUNK_REC_FLAG_ERASE;
```
asiegel-jt, line 329: you need to free the entry value here with fd_funk_val_flush

asiegel-jt (approved):

jumpsiegel (approved):

PR #3705, opened 2024-12-16
ghost, tower: more idiomatic new and join

two-heart:


lidatong (changes requested):
so i intentionally dont want to use scratch_alloc because i intend to migrate both these data structures to be gaddr / offset-based which wont work with this pattern for `new` and `join`

kbhargava-jump:
talked offline. will take a look at localnet run on this PR once CI load is less (not a required check anyways)

PR #3711, opened 2024-12-16
simd-0178: static syscalls

0x0ece:
https://github.com/solana-foundation/solana-improvement-documents/pull/178

Added unit tests for static syscalls. I'll add test-vectors when simd-0179 is also completed as many tests depend on both.

### DIFF
--- a/src/flamenco/vm/fd_vm.c
+++ b/src/flamenco/vm/fd_vm.c
@@ -156,22 +156,29 @@ fd_vm_strerror( int err ) {
 int
 fd_vm_validate( fd_vm_t const * vm ) {
 
+  ulong sbpf_version = vm->sbpf_version;
+
   /* A mapping of all the possible 1-byte sBPF opcodes to their
      validation criteria. */
 
-# define FD_VALID       ((uchar)0) /* Valid opcode */
-# define FD_CHECK_JMP   ((uchar)1) /* Validation should check that the instruction is a valid jump */
-# define FD_CHECK_END   ((uchar)2) /* Validation should check that the instruction is a valid endianness conversion */
-# define FD_CHECK_ST    ((uchar)3) /* Validation should check that the instruction is a valid store */
-# define FD_CHECK_LDQ   ((uchar)4) /* Validation should check that the instruction is a valid load-quad */
-# define FD_CHECK_DIV   ((uchar)5) /* Validation should check that the instruction is a valid division by immediate */
-# define FD_CHECK_SH32  ((uchar)6) /* Validation should check that the immediate is a valid 32-bit shift exponent */
-# define FD_CHECK_SH64  ((uchar)7) /* Validation should check that the immediate is a valid 64-bit shift exponent */
-# define FD_INVALID     ((uchar)8) /* The opcode is invalid */
-# define FD_CHECK_CALLX ((uchar)9) /* Validation should check that callx has valid register number */
-# define FD_CHECK_CALLX_DEPR ((uchar)10) /* Older / deprecated FD_CHECK_CALLX */
+# define FD_VALID               ((uchar)0) /* Valid opcode */
+# define FD_CHECK_JMP_V3        ((uchar)1) /* Validation should check that the instruction is a valid jump (v3+) */
+# define FD_CHECK_END           ((uchar)2) /* Validation should check that the instruction is a valid endianness conversion */
+# define FD_CHECK_ST            ((uchar)3) /* Validation should check that the instruction is a valid store */
+# define FD_CHECK_LDQ           ((uchar)4) /* Validation should check that the instruction is a valid load-quad */
+# define FD_CHECK_DIV           ((uchar)5) /* Validation should check that the instruction is a valid division by immediate */
+# define FD_CHECK_SH32          ((uchar)6) /* Validation should check that the immediate is a valid 32-bit shift exponent */
+# define FD_CHECK_SH64          ((uchar)7) /* Validation should check that the immediate is a valid 64-bit shift exponent */
+# define FD_INVALID             ((uchar)8) /* The opcode is invalid */
+# define FD_CHECK_CALL_REG      ((uchar)9) /* Validation should check that callx has valid register number */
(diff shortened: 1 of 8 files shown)


src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -682,22 +693,19 @@
...
-          fault handling. */
-
-      if( FD_UNLIKELY( imm==0x71e3cf81U ) ){
+      /* Special case to handle entrypoint.
+         ebpf::hash_symbol_name(b"entrypoint") = 0xb00c380, and
+         fd_pchash_inverse( 0xb00c380U ) = 0x71e3cf81U */
+      if( FD_UNLIKELY( imm==0x71e3cf81U ) ) {
         FD_VM_INTERP_STACK_PUSH;
-        pc = entry_pc; /* FIXME: MAGIC NUMBER */
+        pc = entry_pc - 1;
```
jumpsiegel, line 701: Interesting change...

src/flamenco/vm/fd_vm_private.h
```diff
@@ -100,17 +102,103 @@ typedef struct fd_vm_vec fd_vm_vec_t;
...
-#define FD_VM_SBPF_ENABLE_ELF_VADDR(v)             ( v >= FD_SBPF_V3 )
+/* SIMD-0189 */
+#define FD_VM_SBPF_ENABLE_STRICTER_ELF_HEADERS(v)  ( v >= FD_SBPF_V3 )
+#define FD_VM_SBPF_ENABLE_LOWER_BYTECODE_VADDR(v)  ( v >= FD_SBPF_V3 )
 
 #define FD_VM_SBPF_DYNAMIC_STACK_FRAMES_ALIGN      (64U)
 
 #define FD_VM_OFFSET_MASK (0xffffffffUL)
 
+static const uint FD_VM_SBPF_STATIC_SYSCALLS_LIST[] = {
```
jumpsiegel, line 115: I know these are the fd_murmur3_32 of the syscall names... Does this table exist some place in solana as well?

jumpsiegel (dismissed):

src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -861,21 +867,70 @@
...
-    if( FD_UNLIKELY( !frame_cnt ) ) goto sigexit; /* Exit program */
-    frame_cnt--;
-    reg[6]   = shadow[ frame_cnt ].r6;
-    reg[7]   = shadow[ frame_cnt ].r7;
-    reg[8]   = shadow[ frame_cnt ].r8;
-    reg[9]   = shadow[ frame_cnt ].r9;
-    reg[10]  = shadow[ frame_cnt ].r10;
-    pc       = shadow[ frame_cnt ].pc;
-  FD_VM_INTERP_BRANCH_END;
+  FD_VM_INTERP_BRANCH_BEGIN(0x95) { /* FD_SBPF_OP_CALL_IMM */
```
topointon-jump, line 870:
```suggestion
  FD_VM_INTERP_BRANCH_BEGIN(0x95) { /* FD_SBPF_OP_SYSCALL */
```

src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -861,21 +867,70 @@
...
+    vm->frame_cnt = frame_cnt;
+
+    /* Do the syscall.  We use ret reduce the risk of the syscall
+       accidentally modifying other registers (note however since a
+       syscall has the vm handle it still do arbitrary modifications
+       to the vm state) and the risk of a pointer escape on reg from
+       inhibiting compiler optimizations (this risk is likely low in
+       as this is the only point in the whole interpreter core that
+       calls outside this translation unit). */
+
```
topointon-jump, line 898: All this logic is now duplicated from 0x85depr - is it worth consolidating these, so bug fixes in one make it into the other?

src/flamenco/vm/fd_vm_interp_core.c
0x0ece: note the old `pc--` on line 712. I found it a bit less readable, so I moved the -1 inside the branches.

src/flamenco/vm/fd_vm_private.h
0x0ece: in theory, in v3 we no longer need a hash map, we just need an array, because the indexes go from a hash to a sequential id.
the complexity is that we have to track what's active vs not.

two options:
1. create an array, set active/disabled syscalls inside that array, look up the syscall in the array
2. create an array that we use to lookup inside the existing map (where activation is already handled)

1. is one lookup faster, but requires us to keep in sync both map and array, at least until we have to support anything before v3. 2. has the advantage that there's a single place that's the source of truth for activation, at the price of a dbl lookup.

I chose 2 because v3 is pretty far away in the future, and v0..v2 will be in use for a pretty long time even after v3 is active. makes sense?

(agave has their own function registry with a completely different internal impl.)

src/flamenco/vm/fd_vm.c
```diff
@@ -391,19 +412,44 @@ fd_vm_validate( fd_vm_t const * vm ) {
...
+    case FD_CHECK_CALL_REG_DEPR: {
       if( FD_UNLIKELY( instr.imm > 9 ) ) {
         return FD_VM_ERR_INVALID_REG;
       }
       break;
     }
 
+    /* https://github.com/solana-labs/rbpf/blob/4ad935be/src/verifier.rs#L411-L418 */
+    case FD_CHECK_CALL_IMM: {
+      ulong target_pc = (ulong)( (long)i + (long)(int)instr.imm + 1 );
```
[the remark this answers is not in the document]
0x0ece: good catch!

src/flamenco/vm/fd_vm_interp_core.c
0x0ece: added macro FD_VM_INTERP_SYSCALL_EXEC (too many var to make a function)

topointon-jump (approved):

PR #3712, opened 2024-12-16
feat(shredder): always pad last FEC set to 32 data shreds

lidatong:
Ensure the last batch generates >= 32 data shreds by padding with 0s.

This is important for https://github.com/anza-xyz/feature-gate-tracker/issues/19 otherwise our blocks won't get voted on

### DIFF
--- a/src/app/fdctl/run/tiles/fd_shred.c
+++ b/src/app/fdctl/run/tiles/fd_shred.c
@@ -382,11 +382,27 @@ during_frag( fd_shred_ctx_t * ctx,
     ctx->send_fec_set_idx = ULONG_MAX;
     if( FD_UNLIKELY( last_in_batch )) {
       if( FD_UNLIKELY( ctx->batch_cnt%ctx->round_robin_cnt==ctx->round_robin_id ) ) {
-        /* If it's our turn, shred this batch */
+        /* If it's our turn, shred this batch. FD_UNLIKELY because shred tile cnt generally >= 2 */
         ulong batch_sz = sizeof(ulong)+ctx->pending_batch.pos;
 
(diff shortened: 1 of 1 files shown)


src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -382,11 +382,27 @@ during_frag( fd_shred_ctx_t * ctx,
...
+          /* Ensure the last batch generates >= 32 data shreds by
+             padding with 0s. Because the last FEC set is "oddly sized"
+             we only expect this code path to execute for blocks
+             containing less data than can fill 32 data shred payloads
+             (hence FD_UNLIKELY).
+
+             See documentation for FD_SHREDDER_NORMAL_FEC_SET_PAYLOAD_SZ
+             for further context. */
+
+          fd_memset( ctx->pending_batch.raw + sizeof(ulong) + batch_sz, 0, FD_SHREDDER_NORMAL_FEC_SET_PAYLOAD_SZ - batch_sz );
```
ptaffet-jump, line 402:
```suggestion
          fd_memset( ctx->pending_batch.payload + ctx->pending_batch.pos, 0, FD_SHREDDER_NORMAL_FEC_SET_PAYLOAD_SZ - batch_sz );
```

ptaffet-jump (commented):
I think you have one small mistake that will lead to reading 8 uninitialized bytes. Writing it this way makes it more clear it's correct.

lidatong:
> I think you have one small mistake that will lead to reading 8 uninitialized bytes. Writing it this way makes it more clear it's correct.

good point thanks

ptaffet-jump (approved):

Issue #3710, opened 2024-12-16
fd_tpool_wait should have FD_COMPILER_MFENCE() at end of function


GaneshRapolu:
fd_tpool_private_worker(..) has an FD_COMPILER_MFENCE after executing the task and setting state to IDLE. 
https://github.com/firedancer-io/firedancer/blob/main/src/util/tpool/fd_tpool.cxx#L74

This prevents compiler reordering of writes done within the task with setting of worker state.

This should be matched with a  FD_COMPILER_MFENCE after the read of worker state observing IDLE in fd_tpool_wait; that would ensure that reads done after observing the worker state are not re-ordered by the compiler with the read of the worker state. However no such compiler fence exists in the impl today which is likely just an oversight.

https://github.com/firedancer-io/firedancer/blob/fcacb50a7195c47cca147ace8db20d125b4761e6/src/util/tpool/fd_tpool.h#L813

Although unnecessary (??) for many use patterns, it likely simplest  (and also mirrors other uses of the compiler fence in many other places; ex fd_fseq_query) to also add one at the start of fd_tpool_wait.

An example of this bug potentially affecting correctness is the recent checkpoint restore_v2 impl which reads err and dirty after calling fd_tpool_wait.

https://github.com/firedancer-io/firedancer/blob/main/src/util/wksp/fd_wksp_restore_v2.c#L421

Also as hinted by kbowers; compiler fences like these are sufficient (for acquire-release patterns or seq-lock style patterns) only on TSO architectures (ex. x86); on arm you would need actual fences or use instructions that have embedded ordering (load acquire; store release etc). But fixing that can/should be part of a later holistic rework to support non-x86 deployments.

cc: @kbowers-jump

kbowers-jump:
No objection here to having fd_tpool_wait using pre- and post-compiler mfences under the hood.   Good catch.

GaneshRapolu:
Now that I am looking at the code once more;

fd_tpool_exec
https://github.com/firedancer-io/firedancer/blob/main/src/util/tpool/fd_tpool.h#L801
is doing the right thing with pre  & post compiler fences and also a compiler fence between setting the worker args and setting worker state to EXEC. 


However, fd_tpool_private_worker https://github.com/firedancer-io/firedancer/blob/main/src/util/tpool/fd_tpool.cxx#L55 does not have a corresponding compiler fence between reading EXEC as worker state and reading the worker args.

Pretty sure that is not right theoretically either.

[...]

PR #3760, opened 2024-12-19
Stronger fd_tpool compiler memory fencing guarantees

kbowers-jump:
As per the discussion in issue #3710, gave the fd_tpool API stronger compiler memory fencing guarantees and updated FD_SPIN_PAUSE documentation to be explicit about its current (lack of) fencing guarantees.

In the process, this fixes a latent bug caused by missing compiler memory fence on targets without FD_HAS_X86.  The missing fence could cause the the compiler to "optimize" the tpool worker task dispatch outer loop into an infinite loop for worker threads in tpools containing 2 or more tiles within the same process on targets without FD_HAS_X86.

With this, in the vast majority of circumstances (e.g. not writing low level concurrency tests/benchmarks), normal code (i.e. single threaded non-conflicting) should "just work" without use of volatile, FD_VOLATILE, FD_COMPILER_MFENCE, FD_ATOMIC, etc when thread parallelized via fd_tpool / FD_FOR_ALL / FD_MAP_REDUCE (and the unit test was updated accordingly).

### DIFF
--- a/src/util/fd_util_base.h
+++ b/src/util/fd_util_base.h
@@ -742,7 +742,11 @@ fd_type_pun_const( void const * p ) {
    the other logical cores sharing the same underlying physical core for
    a few clocks without yielding it to the operating system scheduler.
    Typically useful for shared memory spin polling loops, especially if
-   hyperthreading is in use. */
+   hyperthreading is in use.  IMPORTANT SAFETY TIP!  This might act as a
+   FD_COMPILER_MFENCE on some combinations of toolchains and targets
+   (e.g. gcc documents that __builtin_ia32_pause also does a compiler
+   memory) but this should not be relied upon for portable code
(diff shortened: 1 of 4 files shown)


GaneshRapolu:
This change seems to address both of the concerns I mentioned in the linked issue:
 - add embedded compiler fencing in fd_tpool_wait
 - add compiler fencing for observing EXEC -> reading task args in fd_tpool_private_worker

so I am happy with this. Rest of changes seem sane as well from a brief skim. 

There is the rat's nest of exactly when in a multi process shared memory environment volatile and/or atomics need to be used to avoid data race UB (vs relying on compiler fences alone to constrain compiler behavior), but that is well beyond my ken. Will leave it to your judgement.

jumpsiegel (approved):

PR #3767, opened 2024-12-20
net: use gossip ip as net tile ip

anwayde:
When peers send packets, they set the destination to our gossip IP. If the gossip IP is different than the interface IP we determine using `listen_address` (for example, if the interface has multiple IPs which is the case for double zero), the incoming packets never gets routed to FD.


ripatel-fd:
Is this a double zero related fix? If so, I suggest closing this. 
My net tile rewrite natively supports ingress targeting any assigned IP (weak host model) so this isn't needed.

anwayde:
It is helpful for double zero but in general this just seems the correct behavior anyways than what it is right now.

ripatel-fd:
> It is helpful for double zero but in general this just seems the correct behavior anyways than what it is right now.

@anwayde I see. Is anything broken at the moment that is fixed by this PR? In case not, I'm hesitant to do this change and risk a regression

anwayde:
Ok, since this isn't rushed, closing in favor of the rewrite that's coming soon

PR #3784, opened 2024-12-24
flamenco, blockstore: deserialize per-batch and allow trailing bytes

yufeng-jump:
Agave has this notion of a batch of data shreds.  The presence of one of the COMPLETE flags in a data shred marks the end of a batch. Moreover, bincode deserialization of microblocks is performed on a per-batch basis.  Precisely a single array of microblocks is expected to be deserialized from a batch.  Trailing bytes in each batch are ignored by default.

Block parsers in both the runtime and the blockstore lack the notion of a batch, and consequently do not properly handle blocks with trailing bytes, such as
https://explorer.solana.com/block/308445708
which has 7335 trailing bytes in the last batch.

### DIFF
--- a/src/flamenco/runtime/fd_blockstore.c
+++ b/src/flamenco/runtime/fd_blockstore.c
@@ -460,34 +460,46 @@ fd_blockstore_scan_block( fd_blockstore_t * blockstore, ulong slot, fd_block_t *
   fd_block_txn_t txns[MAX_TXNS];
   ulong              txns_cnt = 0;
 
-  uchar * data = fd_wksp_laddr_fast( fd_blockstore_wksp( blockstore ), block->data_gaddr );
-  ulong   sz   = block->data_sz;
-  FD_LOG_DEBUG(( "scanning slot %lu, ptr %p, sz %lu", slot, (void *)data, sz ));
+  /*
+   * Agave decodes precisely one array of microblocks from each batch.
+   * As of bincode version 1.3.3, the default deserializer used when
+   * decoding a batch in the blockstore allows for trailing bytes to be
+   * ignored.
+   * https://github.com/anza-xyz/agave/blob/v2.1.0/ledger/src/blockstore.rs#L3764
+   */
+  uchar allow_trailing = 1UL;
 
-  ulong blockoff = 0;
-  while( blockoff < sz ) {
-    if( blockoff + sizeof( ulong ) > sz ) FD_LOG_ERR(( "premature end of block" ));
-    ulong mcount = FD_LOAD( ulong, (const uchar *)data + blockoff );
+  uchar const * data = fd_blockstore_block_data_laddr( blockstore, block );
+  FD_LOG_DEBUG(( "scanning slot %lu, ptr %p, sz %lu", slot, (void *)data, block->data_sz ));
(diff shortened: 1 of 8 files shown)


src/ballet/shred/fd_shred.h
```diff
@@ -142,6 +142,9 @@ typedef uchar fd_shred_merkle_t[FD_SHRED_MERKLE_NODE_SZ];
 /* Maximum number of data shreds in a slot, also maximum number of parity shreds in a slot */
 #define FD_SHRED_MAX_PER_SLOT (1 << 15UL) /* 32,768 shreds */
 
+/* Maximum number of entry batches per block.  The agave deshredder performs bincode deserialization on a per-batch basis. */
+#define FD_SHRED_MAX_BATCHES_PER_BLOCK (64UL)
+
```
topointon-jump, line 147: > The agave deshredder performs bincode deserialization on a per-batch basis.

From my reading of the code, it first deshreds the individual shreds to reconstruct the original buffer, and then bincode deserializes that buffer _once_ to reconstruct the entries. So we only need to worry about trailing bytes for the single vector of entries, not each batch in the shred.

I could be wrong 😄  where in the code do you see this repeated deserialization?

lidatong:
not sure the blockstore / deshredder need to know about trailing bytes -- seems that concept should be in the parsing in `fd_runtime`?

also there are 64 ticks but i believe there can be a lot more than 64 entry batches. if every entry batch has a single entry with a single txn

lidatong:
it makes sense to add the notion of entry batch offsets somewhere in the blockstore code though to support replaying entries vs. entire blocks. we should discuss

src/ballet/shred/fd_shred.h
lidatong: i believe yufeng is saying the same thing as you, which is deserializing a vector of entries = entry batch once

here he's defining that there are up to 64 entry batches in a block. but i believe that constant is wrong. see my comment below

yufeng-jump: Think it's flattened after each individual range gets a deserialize() call?  It's very confusing, but a range corresponds to a batch, and I think deshred() also gets called more than once.

https://github.com/anza-xyz/agave/blob/v2.1.0/ledger/src/blockstore.rs#L3744

topointon-jump (commented):
Am looking at the iteration logic in the live case now, will have a fix up soon

src/flamenco/runtime/fd_blockstore.c
```diff
@@ -535,6 +547,18 @@ fd_blockstore_scan_block( fd_blockstore_t * blockstore, ulong slot, fd_block_t *
         blockoff += pay_sz;
       }
     }
+    if( FD_UNLIKELY( blockoff > batch_end_off ) ) {
+      FD_LOG_ERR(( "parser error: shouldn't have been allowed to read past batch boundary" ));
```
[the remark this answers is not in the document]
yufeng-jump: There are a handful of FD_LOG_ERR calls in scan_block() that shouldn't crash the validator.  That can be a separate followup PR.

topointon-jump (approved):

PR #3785, opened 2024-12-24
check_acct_cnt: add a check for zero accts in a instruction

jumpsiegel:
A lot of this PR is white space edits.  The only real change is at line 1198 where I check for zero accounts 
to match the solana behavior.

### DIFF
--- a/src/flamenco/runtime/fd_executor.c
+++ b/src/flamenco/runtime/fd_executor.c
@@ -318,17 +318,17 @@ fd_executor_verify_precompiles( fd_exec_txn_ctx_t * txn_ctx ) {
 }
 
 
-/* Only accounts in the transaction account keys that are owned by one of the four 
-   loaders (bpf v1, v2, v3, v4) are iterated over in Agave's replenish_program_cache() 
-   function to be loaded into the program cache. An account may be in the program cache 
-   iff the owners match one of the four loaders since `filter_executable_program_accounts()` 
+/* Only accounts in the transaction account keys that are owned by one of the four
+   loaders (bpf v1, v2, v3, v4) are iterated over in Agave's replenish_program_cache()
+   function to be loaded into the program cache. An account may be in the program cache
+   iff the owners match one of the four loaders since `filter_executable_program_accounts()`
    filters out all other accounts here:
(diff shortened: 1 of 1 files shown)


src/flamenco/runtime/fd_executor.c
```diff
@@ -1184,89 +1184,84 @@ fd_instr_stack_pop( fd_exec_txn_ctx_t *       txn_ctx,
...
-    } else {
-      /* if txn_ctx->exec_err is not set, it indicates an instruction error */
-      if( !txn_ctx->exec_err ) {
-        FD_TXN_ERR_FOR_LOG_INSTR( txn_ctx, exec_result, txn_ctx->instr_err_idx );
+
+      // https://github.com/anza-xyz/agave/blob/b1acfbde5bd78488641e7dca3cafc07bef48b87b/program-runtime/src/invoke_context.rs#L251
+      //
+      // the get_last_program_key() throws the error that causes the UNSUPPORTED_PROGRAM_ID
+      int err = ( instr->acct_cnt == 0 ) ? FD_EXECUTOR_INSTR_ERR_UNSUPPORTED_PROGRAM_ID : fd_instr_stack_push( txn_ctx, instr );
+      if( FD_UNLIKELY( err ) ) {
```
topointon-jump, line 1199: From what I can tell, `InstructionContext.get_last_program_key` uses the _transaction's_ program indicies list. This is given to the instruction context at instruction preperation time, but is ultimately populated when the accounts for the _transaction_ are loaded: https://github.com/anza-xyz/agave/blob/e063c19025d65870a513fcc010c47685dc989d52/svm/src/account_loader.rs#L476

This uses each instruction's `program_id_index` (https://github.com/anza-xyz/agave/blob/e063c19025d65870a513fcc010c47685dc989d52/svm-transaction/src/instruction.rs#L9), _not_ the list of instruction accounts (https://github.com/anza-xyz/agave/blob/e063c19025d65870a513fcc010c47685dc989d52/svm-transaction/src/instruction.rs#L11).

So this check needs to be done at the transaction level, using each instruction's `program_id` field and checking that this corresponds to a _transaction account_.

kbhargava-jump:
fixed in https://github.com/firedancer-io/firedancer/pull/3916

PR #3800, opened 2024-12-29
Decouple fd_quic from Ethernet layer

### LINKED ISSUE #3798: Remove Ethernet references from fd_quic
fd_quic doesn't need to be aware of layer 2 details

ripatel-fd:
Closes #3798

### DIFF
--- a/src/waltz/quic/tests/fd_quic_test_helpers.c
+++ b/src/waltz/quic/tests/fd_quic_test_helpers.c
@@ -18,8 +18,6 @@
 
 FILE * fd_quic_test_pcap;
 
-/* Mac address counter, incremented for each new QUIC */
-static ulong test_mac_addr_seq = 0x0A0000000000;
 /* IP address counter, incremented for each new QUIC */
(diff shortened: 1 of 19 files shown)


src/waltz/quic/tests/fd_quic_test_helpers.c
```diff
@@ -582,16 +633,14 @@ int
...
-  (void)opt_batch_idx; /* FIXME fd_aio compliance */
-
-  fd_quic_netem_t * mitm_ctx = (fd_quic_netem_t *)ctx;
+  fd_quic_netem_t * mitm_ctx = ctx;
 
   /* go packet by packet */
   for( ulong j = 0UL; j < batch_cnt; ++j ) {
     /* generate a random number and compare with threshold, and either pass thru or drop */
-    static FD_TL uint seed = 0;
+    static FD_TL uint seed = 0; /* FIXME use fd_log_tid */
```
nbridge-jump, line 643: I think tests should be deterministic. You could allow the caller to override the seed, and give the user the ability to run with different seeds, but in CI, for example, you want to be able to run locally exactly what ran in CI

nbridge-jump (approved):

PR #3803, opened 2024-12-30
build.sh: default to Linux only machine types

ripatel-fd:
Fixes bug where CI tries to build macOS and FreeBSD builds on
a Linux environment without cross-compile infra

ShellCheck fixes:
- SC2045: Iterating over ls output is fragile. Use globs
- SC2145: Argument mixes string and array. Use * or separate argument.

### DIFF
--- a/contrib/build.sh
+++ b/contrib/build.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+shopt -s nullglob
+
 # help message
 help() {
   echo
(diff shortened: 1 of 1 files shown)


ripatel-fd:
@anwayde Could you review this script and test it? On my box, the compilers are under `/opt/rh` not `/opt/gcc` so I can't test it myself.

anwayde (approved):

PR #3811, opened 2024-12-30
Reduce fdctl config interface name lengths

ripatel-fd:
Fixes a false positive stringop-truncation lint


mmcgee-jump:
Does this represent some known constant we can refer to?

mmcgee-jump (approved):

ripatel-fd:
> Does this represent some known constant we can refer to?

IF_NAMESIZE

PR #3820, opened 2025-01-02
flamenco: fix get_minimum_stake_delegation logic

two-heart:
cf. the agave logic:

```rust
/// The minimum stake amount that can be delegated, in lamports.
/// NOTE: This is also used to calculate the minimum balance of a delegated stake account,
/// which is the rent exempt reserve _plus_ the minimum stake delegation.
#[inline(always)]
pub fn get_minimum_delegation(feature_set: &FeatureSet) -> u64 {
    if feature_set.is_active(&feature_set::stake_raise_minimum_delegation_to_1_sol::id()) {
        const MINIMUM_DELEGATION_SOL: u64 = 1;
        MINIMUM_DELEGATION_SOL * LAMPORTS_PER_SOL
    } else {
        1
    }
}
```

### DIFF
--- a/src/flamenco/rewards/fd_rewards.c
+++ b/src/flamenco/rewards/fd_rewards.c
@@ -315,7 +315,7 @@ get_minimum_stake_delegation( fd_exec_slot_ctx_t * slot_ctx ) {
         return 0UL;
     }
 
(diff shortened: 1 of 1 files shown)


topointon-jump (approved):
Thank-you!

PR #3823, opened 2025-01-02
flamenco, vm: add check_account_info_pointer cpi security checks

topointon-jump:
We are missing the `check_account_info_pointer` CPI checks in direct mapping, which check that the vaddrs provided by the user correspond to the correct locations.

### DIFF
--- a/src/flamenco/runtime/program/fd_bpf_loader_serialization.c
+++ b/src/flamenco/runtime/program/fd_bpf_loader_serialization.c
@@ -109,9 +109,9 @@ write_account( fd_exec_instr_ctx_t *     instr_ctx,
 
     /* Update the mapping from instruction account index to memory region index.
        This is an optimization to avoid redundant lookups to find accounts. */
-    acc_region_metas[instr_acc_idx] = (fd_vm_acc_region_meta_t){ .region_idx          = *input_mem_regions_cnt,
-                                                                 .has_data_region     = !!dlen,
-                                                                 .has_resizing_region = (uchar)is_aligned };
+    acc_region_metas[instr_acc_idx].region_idx          = *input_mem_regions_cnt;
+    acc_region_metas[instr_acc_idx].has_data_region     = !!dlen;
+    acc_region_metas[instr_acc_idx].has_resizing_region = (uchar)is_aligned;
 
     if( dlen ) {
       new_input_mem_region( input_mem_regions, input_mem_regions_cnt, data, dlen, is_writable );
@@ -214,6 +214,17 @@ fd_bpf_loader_input_serialize_aligned( fd_exec_instr_ctx_t       ctx,
       FD_STORE( uchar, serialized_params, (uchar)dup_acc_idx[acc_idx] );
       serialized_params += sizeof(ulong);
     } else {
+      /* Calculate and store the start of the actual metadata region for this account,
+         excluding any duplicate account markers at the beginning.
+
+         We use this later for retrieving the serialized values later in the CPI security checks. */
+      ulong metadata_region_offset_with_dups = *input_mem_regions_cnt==0UL ? 0UL : 
+        input_mem_regions[ *input_mem_regions_cnt-1U ].vaddr_offset +
+        input_mem_regions[ *input_mem_regions_cnt-1U ].region_sz;
+
+      acc_region_metas[i].metadata_region_offset = metadata_region_offset_with_dups +
+        (ulong)(serialized_params - curr_serialized_params_start);
+
       FD_STORE( uchar, serialized_params, FD_NON_DUP_MARKER );
       serialized_params += sizeof(uchar);
 
@@ -484,6 +495,17 @@ fd_bpf_loader_input_serialize_unaligned( fd_exec_instr_ctx_t       ctx,
       FD_STORE( uchar, serialized_params, (uchar)dup_acc_idx[acc_idx] );
       serialized_params += sizeof(uchar);
     } else {
+      /* Calculate and store the start of the actual metadata region for this account,
+         excluding any duplicate account markers at the beginning.
+
+         We use this later for retrieving the serialized values later in the CPI security checks. */
+      ulong metadata_region_offset_with_dups = *input_mem_regions_cnt==0UL ? 0UL : 
+        input_mem_regions[ *input_mem_regions_cnt-1U ].vaddr_offset +
+        input_mem_regions[ *input_mem_regions_cnt-1U ].region_sz;
+
+      acc_region_metas[i].metadata_region_offset = metadata_region_offset_with_dups +
+        (ulong)(serialized_params - curr_serialized_params_start);
+
       FD_STORE( uchar, serialized_params, FD_NON_DUP_MARKER );
       serialized_params += sizeof(uchar);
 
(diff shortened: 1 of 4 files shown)


src/flamenco/runtime/program/fd_bpf_loader_serialization.c
```diff
@@ -241,14 +241,30 @@ fd_bpf_loader_input_serialize_aligned( fd_exec_instr_ctx_t       ctx,
 
       fd_pubkey_t key = *acc;
       FD_STORE( fd_pubkey_t, serialized_params, key );
+
+      /* Calculate the start of the metadata region */
```
ibhatt-jumptrading, line 245: maybe add a little explanation of why we need to do this calculation. May not be totally obvious to someone not familiar with this

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
```diff
@@ -241,14 +241,30 @@ fd_bpf_loader_input_serialize_aligned( fd_exec_instr_ctx_t       ctx,
...
+      ulong metadata_region_vaddr = FD_VM_MEM_MAP_INPUT_REGION_START +
```
ibhatt-jumptrading, line 246: I think it would be more straightforward to put this calculation in write_account() since then you won't need to index by a cnt-1 and have to manually do the offset yourself.

ibhatt-jumptrading: Actually, metadata_region_vaddr == INPUT_REGION_START + vaddr_offset so I am not sure if we need to do this calculation at all

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
```diff
@@ -241,14 +241,30 @@ fd_bpf_loader_input_serialize_aligned( fd_exec_instr_ctx_t       ctx,
...
+          ( *input_mem_regions_cnt==0UL ? 0UL : (
+            input_mem_regions[ *input_mem_regions_cnt-1U ].vaddr_offset +
+             input_mem_regions[ *input_mem_regions_cnt-1U ].region_sz ) );
+
+      /* Store the vaddrs of the serialized pubkey, owner and lamports for CPI security checks later.
+      
+      We calculate them in the serialization code for several reasons:
```
ibhatt-jumptrading, line 253: Can you explain why these things are true as a result of this precomputation

ibhatt-jumptrading: Also, I don't know if we need to compute this at all because you could just #define offsets for pubkey owner and lamports into the metadata and use those to grab the vaddr as long as you have the vaddr offset.

Maybe define small static inline functions that just takes in the vaddr offset for a region and gives you the vaddr. These should basically be "free"/O(1)"as you're just adding ulongs

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
```diff
@@ -515,6 +539,13 @@ fd_bpf_loader_input_serialize_unaligned( fd_exec_instr_ctx_t       ctx,
 
       fd_pubkey_t owner = *(fd_pubkey_t *)&metadata->info.owner;
       FD_STORE( fd_pubkey_t, serialized_params, owner );
+
```
ibhatt-jumptrading, line 542: don't think this is needed as described abvove

src/flamenco/vm/fd_vm.h
```diff
@@ -29,14 +29,29 @@ struct __attribute__((aligned(8UL))) fd_vm_input_region {
 };
 typedef struct fd_vm_input_region fd_vm_input_region_t;
 
+/* Absolute vaddr pointers to the various fields of a serialized account,
```
ibhatt-jumptrading, line 32: Think we could eliminate this as a result

src/flamenco/vm/syscall/fd_vm_syscall_cpi.c
```diff
@@ -516,9 +516,13 @@ fd_vm_syscall_cpi_check_authorized_program( fd_pubkey_t const *        program_i
   FD_VM_MEM_HADDR_LD( vm, acc_meta->pubkey_addr, alignof(uchar), sizeof(fd_pubkey_t) )
 
 /* VM_SYSCALL_CPI_ACC_INFO_T accessors */
```
ibhatt-jumptrading, line 518: These then should reference the helpers that I described above

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
```diff
@@ -324,14 +324,30 @@ VM_SYSCALL_CPI_TRANSLATE_AND_UPDATE_ACCOUNTS_FUNC(
...
       if (instruction_accounts[i].is_writable) {
-        out_callee_indices[*out_len] = instruction_accounts[i].index_in_caller;
+        out_callee_indices[*out_len] = index_in_caller;
         out_caller_indices[*out_len] = j;
         (*out_len)++;
       }
       found = 1;
 
       if ( vm->direct_mapping ) {
+        /* Check that the account info pointers given by the user correspond to the correct locations
```
ibhatt-jumptrading, line 336: do you need to add the corresponding log message here

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
```diff
@@ -324,14 +324,30 @@ VM_SYSCALL_CPI_TRANSLATE_AND_UPDATE_ACCOUNTS_FUNC(
...
+           in the serialized account metadata
+           
+           https://github.com/anza-xyz/agave/blob/c79c3c9e67274594ed4f43ed8386e9ddc60a99e9/programs/bpf_loader/src/syscalls/cpi.rs#L116 */
+        fd_vm_acc_region_meta_t acc_region_meta = vm->acc_region_metas[index_in_caller];
+        fd_vm_serialized_account_metadata_t serialized_acc_meta = acc_region_meta.serialized_acc_metadata;
+        if ( FD_UNLIKELY(( account_infos[j].pubkey_addr != serialized_acc_meta.key_vaddr )) ) {
```
ibhatt-jumptrading, line 342: nit: spacing and extra parens. You did for the below checks too, just writing this one comment though

src/flamenco/vm/syscall/fd_vm_syscall_cpi_common.c
```diff
@@ -324,14 +324,30 @@ VM_SYSCALL_CPI_TRANSLATE_AND_UPDATE_ACCOUNTS_FUNC(
...
         out_caller_indices[*out_len] = j;
         (*out_len)++;
       }
       found = 1;
 
       if ( vm->direct_mapping ) {
+        /* Check that the account info pointers given by the user correspond to the correct locations
+           in the serialized account metadata
+           
+           https://github.com/anza-xyz/agave/blob/c79c3c9e67274594ed4f43ed8386e9ddc60a99e9/programs/bpf_loader/src/syscalls/cpi.rs#L116 */
```
ibhatt-jumptrading, line 339: small detail but I think the permalinks that directly reference the version as opposed to the commit sha are cleaner + more informative since it's a hint as to how old this code is and it might warn you that it's outdated and needs to be changed

src/flamenco/vm/syscall/fd_vm_syscall_cpi.c
topointon-jump: No, these are to pull out the user-provided vaddrs (in the CPI account metas). They are laid out differently in memory to the vm serialized paramaters, so we can't use the same accessors unfortunately.

ibhatt-jumptrading (approved):
Nice!

PR #3825, opened 2025-01-02
flamenco: fix strict aliasing violations for loader v4

mjain-jump:


src/flamenco/runtime/program/fd_loader_v4_program.c
```diff
@@ -36,8 +36,7 @@ fd_loader_v4_set_state( fd_exec_instr_ctx_t *  instr_ctx,
     return FD_EXECUTOR_INSTR_ERR_ACC_DATA_TOO_SMALL;
   }
 
-  fd_loader_v4_state_t * out_state = (fd_loader_v4_state_t *)data;
-  *out_state = *state;
+  fd_memcpy( data, state, LOADER_V4_PROGRAM_DATA_OFFSET );
```
lidatong, line 39: you should `fd_type_pun( ... )` these instead. then you can do the cast and avoid the copy

as an aside, use `memcpy` instead of `fd_memcpy` when the size of the copy is statically known

src/flamenco/runtime/program/fd_loader_v4_program.c
```diff
@@ -54,7 +53,8 @@ fd_loader_v4_get_state( fd_borrowed_account_t const * program,
   if( FD_UNLIKELY( program->const_meta->dlen<LOADER_V4_PROGRAM_DATA_OFFSET ) ) {
     return FD_EXECUTOR_INSTR_ERR_ACC_DATA_TOO_SMALL;
   }
-  *state = *( (fd_loader_v4_state_t *) program->const_data );
+
+  fd_memcpy( state, program->const_data, LOADER_V4_PROGRAM_DATA_OFFSET );
```
lidatong, line 57: same as above

lidatong (changes requested):

lidatong (approved):

PR #3829, opened 2025-01-02
flamenco: implement verify_ticks

yufeng-jump:


src/flamenco/runtime/fd_runtime.c
```diff
@@ -827,25 +827,70 @@ fd_runtime_block_prepare( fd_blockstore_t * blockstore,
   return 0;
 }
 
-// TODO: this function doesnt do anything!
+/*
+ * https://github.com/anza-xyz/agave/blob/v2.1.0/ledger/src/blockstore_processor.rs#L1096
```
ibhatt-jumptrading, line 831: nit but can you do the block comments without the * on each new line

src/flamenco/runtime/fd_runtime.c
```diff
@@ -827,25 +827,70 @@ fd_runtime_block_prepare( fd_blockstore_t * blockstore,
...
+ */
 int
 fd_runtime_block_verify_ticks( fd_block_info_t const * block_info,
                                ulong                   tick_height,
-                               ulong                   max_tick_height ) {
-  (void)tick_height; (void)max_tick_height;
-  ulong tick_count = 0UL;
+                               ulong                   max_tick_height,
+                               ulong                   hashes_per_tick ) {
+  ulong tick_count      = 0UL;
```
ibhatt-jumptrading, line 839: nit: align

src/flamenco/runtime/fd_runtime.c
```diff
@@ -827,25 +827,70 @@ fd_runtime_block_prepare( fd_blockstore_t * blockstore,
   return 0;
 }
```
ibhatt-jumptrading, line 829: Maybe it makes sense for this function to be in fd_blockstore... Not sure though

src/flamenco/runtime/fd_runtime.c
```diff
@@ -827,25 +827,70 @@ fd_runtime_block_prepare( fd_blockstore_t * blockstore,
...
+  if( FD_UNLIKELY( next_tick_height < max_tick_height ) ) {
+    FD_LOG_WARNING(( "Too few ticks" ));
+    return FD_BLOCK_ERR_TOO_FEW_TICKS;
+  }
+  if( FD_UNLIKELY( has_trailing_entry ) ) {
+    FD_LOG_WARNING(( "Did not end with a tick" ));
+    return FD_BLOCK_ERR_TRAILING_ENTRY;
+  }
+
+  /* Not returning FD_BLOCK_ERR_INVALID_LAST_TICK because we assume the
```
ibhatt-jumptrading, line 882: Can you explain why this exists in agave. I am pretty sure that in agave, slot_full is always true

src/flamenco/runtime/fd_runtime.c
```diff
@@ -827,25 +827,70 @@ fd_runtime_block_prepare( fd_blockstore_t * blockstore,
   return 0;
```
ibhatt-jumptrading, line 827: Is how we calculate ticks correct? The +=64 feels super hacky and I'm not sure if that's correct. IIRC agave calculates it as microblocks are processed/executed

src/flamenco/runtime/fd_runtime.c
```diff
@@ -827,25 +827,70 @@ fd_runtime_block_prepare( fd_blockstore_t * blockstore,
...
-                               ulong                   max_tick_height ) {
-  (void)tick_height; (void)max_tick_height;
-  ulong tick_count = 0UL;
+                               ulong                   max_tick_height,
+                               ulong                   hashes_per_tick ) {
+  ulong tick_count      = 0UL;
+  ulong tick_hash_count = 0UL;
+  uchar has_trailing_entry      = 0U;
+  uchar invalid_tick_hash_count = 0U;
   for( ulong i = 0UL; i < block_info->microblock_batch_cnt; i++ ) {
```
ibhatt-jumptrading, line 843: It might be better to split out this calculation into its own function and explain that it combines and precomputes some of the checks like last_trailing_entry and verify_tick_hash_count. This will make this function cleaner and it'll look more the agave analog.

src/flamenco/runtime/fd_runtime.c
yufeng-jump: I don't think we need this complexity.  My best guess is that they had/have fantasies about replaying blocks piecemeal and thus verifying ticks before a block is full.  We don't have a streaming execution model yet, so all our blocks are either just full, or they are not gonna be replayed at all.  When we do eventually move to a streaming execution model, we can always update this function to support streaming tick verification, or just do tick verification at the end of a block.

src/flamenco/runtime/fd_runtime.c
yufeng-jump: I'm in favor

src/app/fdctl/run/tiles/fd_replay.c
```diff
@@ -998,16 +998,20 @@ prepare_new_block_execution( fd_replay_tile_ctx_t * ctx,
   FD_TEST( fork == child );
```
ibhatt-jumptrading, line 999: It doesn't look like we are verifying ticks in the live case at all

ibhatt-jumptrading (approved):

PR #3830, opened 2025-01-02
Complete the implementation of wen-restart

yhzhangjump:
Fix a few final TODOs for wen-restart.

1. Generate a snapshot at the end of wen-restart; Insert a hard fork and calculate a new shred version.
2. Fix tower checkpoint by checkpointing all the voted slots from the last voted slot back to the ghost root; Some of these slots may not be in the tower.
3. Improve the code style.


yhzhangjump:
Asking @alpeng-jump for review as well since Alex also knows wen-restart.

PR #3734, opened 2024-12-18
blockstore: fix and document locking

two-heart:

PR #3833, opened 2025-01-03
fix(disco, store): fix and document locking

topointon-jump:
https://github.com/firedancer-io/firedancer/pull/3734 caused an issue in live, because there was a path where the read lock was never being released. The PR was reverted.

This PR restores the original PR #3734, with a fix to handle the above bug.

Tested locally by running against testnet repeatedly.


ibhatt-jumptrading (approved):
could you just squash these commits

two-heart (approved):

ibhatt-jumptrading (approved):

PR #3867, opened 2025-01-06
 ledger: checkpoint on mismatch

kbhargava-jump:


src/app/ledger/main.c
```diff
@@ -172,9 +173,7 @@ fd_create_snapshot_task( void FD_PARAM_UNUSED *tpool,
   fd_snapshot_create_new_snapshot( snapshot_ctx,
                                    &ledger_args->last_snapshot_hash,
                                    &ledger_args->last_snapshot_cap );
-  if( FD_UNLIKELY( err ) ) {
```
kbhargava-jump, line 175: @ibhatt-jumptrading think this should be removed? we have a check on `err` above when it was last set

ibhatt-jumptrading: yes

src/app/ledger/main.c
```diff
@@ -463,6 +462,18 @@ runtime_replay( fd_ledger_args_t * ledger_args ) {
       if( ledger_args->checkpt_mismatch ) {
         fd_runtime_checkpt( ledger_args->capture_ctx, ledger_args->slot_ctx, ULONG_MAX );
       }
+      if ( ledger_args->snapshot_mismatch ) {
```
ibhatt-jumptrading, line 465: code style, please take time to fix it in new code.

ibhatt-jumptrading (approved):

PR #3878, opened 2025-01-07
flamenco, repair: cache good repair peers to file

topointon-jump:
We periodically cache the known good repair peers to a file, so that we can request from these when we restart. This reduces startup times.

### DIFF
--- a/src/app/fdctl/run/tiles/repair.seccomppolicy
+++ b/src/app/fdctl/run/tiles/repair.seccomppolicy
@@ -1,18 +1,37 @@
 # logfile_fd: It can be disabled by configuration, but typically tiles
 #             will open a log file on boot and write all messages there.
-unsigned int logfile_fd
+# good_peer_cache_file_fd: The list of known good repair peers, which we read
+#                          at boot and then periodically write to.
+unsigned int logfile_fd, unsigned int good_peer_cache_file_fd
 
 # logging: all log messages are written to a file and/or pipe
 #
 # 'WARNING' and above are written to the STDERR pipe, while all messages
 # are always written to the log file.
 #
+# good_peer_cache_file: we periodically write out the good peers to the file
+# 
 # arg 0 is the file descriptor to write to.  The boot process ensures
 # that descriptor 2 is always STDERR.
 write: (or (eq (arg 0) 2)
-           (eq (arg 0) logfile_fd))
+           (eq (arg 0) logfile_fd)
+           (eq (arg 0) good_peer_cache_file_fd))
+
+# good_peer_cache_file_fd:
+#
+# We need to seek to the start of the file when we read it in.
+lseek: (eq (arg 0) good_peer_cache_file_fd)
+
+# good_peer_cache_file_fd:
+#
+# We need to truncate the file to 0 before we start writing to it.
+ftruncate: (and (eq (arg 0) good_peer_cache_file_fd) 
+                (eq (arg 1) 0))
 
 # logging: 'WARNING' and above fsync the logfile to disk immediately
 #
 # arg 0 is the file descriptor to fsync.
 fsync: (eq (arg 0) logfile_fd)
+
+# good_peer_cache_file: we read in the good peer cache file on boot
+read: (eq (arg 0) good_peer_cache_file_fd)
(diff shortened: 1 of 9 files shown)


mmcgee-jump:
This has sandbox issues and seems like a bit of a workaround. Why is it slow to get a fresh list of repair peers on boot?

src/flamenco/repair/fd_repair.c
```diff
@@ -559,17 +568,133 @@ fd_repair_decay_stats( fd_repair_t * glob ) {
...
+}
+
+static int
+fd_read_in_good_peer_cache_file( fd_repair_t * repair ) {  
+  if ( repair->good_peer_cache_file_fd==-1 ) {
+    FD_LOG_NOTICE(( "No repair good_peer_cache_file specified, not loading cached peers" ));
+    return 0;
+  }
+
+  long seek = lseek( repair->good_peer_cache_file_fd, 0UL, SEEK_SET );
```
ibhatt-jumptrading, line 613: this breaks the sandbox. Any syscalls not in priv_init need to be part of the seccomppolicy

src/app/fdctl/run/tiles/repair.seccomppolicy
```diff
@@ -1,18 +1,26 @@
 # logfile_fd: It can be disabled by configuration, but typically tiles
```
ibhatt-jumptrading, line 1: add lseek

src/flamenco/repair/fd_repair.c
```diff
@@ -559,17 +568,133 @@ fd_repair_decay_stats( fd_repair_t * glob ) {
...
+ * read_line() reads characters one by one from 'fd' until:
+ *   - it sees a newline ('\n')
+ *   - it reaches 'max_len - 1' characters
+ *   - or EOF (read returns 0)
+ * It stores the line in 'buf' and null-terminates it.
+ *
+ * Returns the number of characters read (not counting the null terminator),
+ * or -1 on error.
+ */
+long read_line(int fd, char *buf) {
```
ibhatt-jumptrading, line 581: chat gpt code 😶‍🌫️

src/flamenco/repair/fd_repair.c
```diff
@@ -559,17 +568,133 @@ fd_repair_decay_stats( fd_repair_t * glob ) {
   }
 }
 
+/**
+ * read_line() reads characters one by one from 'fd' until:
```
ibhatt-jumptrading, line 572: Another option instead of defining a human readable file that's broken with new lines is that you define a struct of some fixed size and then pack in structs. This way you can do an fd_io_read( fd, sizeof(struct)).

src/flamenco/repair/fd_repair.c
```diff
@@ -920,6 +1049,63 @@ fd_repair_create_needed_request( fd_repair_t * glob, int type, ulong slot, uint
...
+  if ( repair->good_peer_cache_file_fd == -1 ) {
+    return 0;
+  }
+
+  if ( repair->actives_sticky_cnt == 0 ) {
+    return 0;
+  }
+
+  /* Truncate the file before we write it */
+  int err = ftruncate( repair->good_peer_cache_file_fd, 0UL );
```
ibhatt-jumptrading, line 1065: also need to add this to your seccomppolicy

src/flamenco/repair/fd_repair.c
```diff
@@ -559,17 +568,133 @@ fd_repair_decay_stats( fd_repair_t * glob ) {
   }
```
ibhatt-jumptrading, line 568: general alignment + spacing nits

src/flamenco/repair/fd_repair.c
topointon-jump: good spot ty

src/flamenco/repair/fd_repair.c
topointon-jump: I thought that it was important for an operator to be able to vet the `good_peer_cache_file` easily, before booting up and trusting these nodes.

topointon-jump:
> This has sandbox issues

Corrected the approach to set up the sandbox properly, thanks for the spot 😄 

> Why is it slow to get a fresh list of repair peers on boot?

It is slow because we don't have the stake weights until the full snapshot is loaded, and we don't have the IP addresses of the nodes until these are fetched through gossip. We are working on making both of these faster but in the meantime this cache also helps.

ibhatt-jumptrading (approved):

PR #3883, opened 2025-01-07
flamenco, fuzz: use spad debugging variants in harness setup and teardown

ravyu-jump:
Use spad debugging variants to track violations on spad memory (namely alloc overflow)

### DIFF
--- a/src/flamenco/runtime/tests/fd_exec_instr_test.c
+++ b/src/flamenco/runtime/tests/fd_exec_instr_test.c
@@ -102,7 +102,7 @@ fd_exec_instr_test_runner_new( void * mem,
 
   /* Create spad */
   runner->spad = fd_spad_join( fd_spad_new( spad_mem, FD_RUNTIME_TRANSACTION_EXECUTION_FOOTPRINT_FUZZ ) );
-  fd_spad_push( runner->spad );
+  fd_spad_push_debug( runner->spad );
   return runner;
 }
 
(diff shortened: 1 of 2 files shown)


src/util/spad/fd_spad.h
```diff
@@ -390,6 +390,11 @@ fd_spad_alloc( fd_spad_t * spad,
                ulong       sz ) {
   align = fd_ulong_if( align>0UL, align, FD_SPAD_ALLOC_ALIGN_DEFAULT ); /* typically compile time */
   ulong   off = fd_ulong_align_up( spad->mem_used, align );
+
+#if FD_SPAD_USE_HANDHOLDING
+  if( FD_UNLIKELY( off + sz > spad->mem_max ) ) FD_LOG_ERR(( "spad alloc overflow" ));
+#endif
+
```
mjain-jump, line 397: `fd_spad_alloc_debug` already uses `fd_spad_alloc_max` which seems to be doing the same thing, so i think this check is unnecessary?

src/util/spad/fd_spad.h
```diff
@@ -390,6 +390,11 @@ fd_spad_alloc( fd_spad_t * spad,
                ulong       sz ) {
   align = fd_ulong_if( align>0UL, align, FD_SPAD_ALLOC_ALIGN_DEFAULT ); /* typically compile time */
   ulong   off = fd_ulong_align_up( spad->mem_used, align );
+
+#if FD_SPAD_USE_HANDHOLDING
+  if( FD_UNLIKELY( off + sz > spad->mem_max ) ) FD_LOG_ERR(( "spad alloc overflow" ));
```
yufeng-jump, line 395: Can you log more information here, like off sz and max?

src/util/spad/fd_spad.h
mjain-jump: never mind...

yufeng-jump (approved):
I like this.  Ideally we'd have a PR that systematically adds handholding checks to fd_spad, but I'm not gonna let the perfect be the enemy of the good.

mjain-jump:
Would it be better to have a compile-time flag that uses `fd_spad_*_debug` throughout the runtime in the context of fuzzing? That way we get handholding in all allocations and not just in the harnesses.

ravyu-jump:
> Would it be better to have a compile-time flag that uses fd_spad_*_debug throughout the runtime in the context of fuzzing? That way we get handholding in all allocations and not just in the harnesses.

Yeah I was thinking about that, can probably get crafty with the linker but that should probably be a separate effort

PR #3887, opened 2025-01-08
Handholding Extras

ravyu-jump:
- `with-handholding.mk` correctly activates the handholding macros

- Use macros in spad to override default APIs with debug variants


two-heart (approved):

src/util/spad/fd_spad.h
```diff
@@ -531,6 +531,28 @@ fd_spad_private_frame_end_debug( fd_spad_t ** _spad ) {
...
+#define fd_spad_publish   fd_spad_publish_debug
+
+#undef  FD_SPAD_FRAME_BEGIN
+#undef  FD_SPAD_FRAME_END
+
+#define FD_SPAD_FRAME_BEGIN FD_SPAD_FRAME_BEGIN_DEBUG
+#define FD_SPAD_FRAME_END   FD_SPAD_FRAME_END_DEBUG
+
+#endif /* FD_SPAD_USE_HANDHOLDING */
+
```
ibhatt-jumptrading, line 554: extra space

yufeng-jump (commented):
I wonder if we shouldn't just extract the implementation into xxx_impl functions, and then the xxx_debug() variants would call the corresponding impl functions, and the non debug variants would be defined to either impl or debug, depending on whether handholding is on or not.  This way we won't have to do the undef hack, and it's abundantly clear that the non debug variants are either just impl or debug.

yufeng-jump (approved):

yufeng-jump (approved):

PR #3888, opened 2025-01-08
agave: allow specifying RPC bind address

alpeng-jump:


src/app/fdctl/config/default.toml
```diff
@@ -279,6 +279,12 @@ dynamic_port_range = "8900-9000"
     # `--private-rpc` argument.
     private = false
 
+    # The address to bind the RPC port. If no address is specified,
```
mmcgee-jump, line 282: nit: these comments are double spaced
```suggestion
    # The address to bind the RPC port.  If no address is specified,
```

mmcgee-jump (changes requested):
.

PR #3890, opened 2025-01-08
fix(ghost): voting xepoch & polish

lidatong:


src/choreo/ghost/fd_ghost.c
```diff
@@ -405,34 +409,63 @@ fd_ghost_replay_vote( fd_ghost_t * ghost, fd_voter_t * voter, ulong slot ) {
...
     #endif
     ancestor = fd_ghost_node_pool_ele( node_pool, ancestor->parent_idx );
   }
 
+}
+
+void
+fd_ghost_replay_vote( fd_ghost_t * ghost, fd_voter_t * voter, ulong slot ) {
+  replay_vote( ghost, &voter->key, voter->stake, voter->replay_vote, voter->stake, slot );
   voter->replay_vote = slot; /* update the cached replay vote slot on voter */
```
emwang-jumptrading, line 442: It seems like now `voter->replay_vote` will update to `slot` even when the `if( ( slot <= last_slot ) ) return;` condition is hit in `static void replay_vote`, but now it seems like it'll update the `voter->replay_vote` even if it doesn't update to this `slot`.

lidatong: good catch

ripatel-fd:
Closing this PR since it's quite old, feel free to reopen @lidatong

PR #3896, opened 2025-01-08
flamenco: remove unused account compute table from slot ctx

mjain-jump:


src/app/fdctl/run/tiles/fd_gossip.c
```diff
@@ -56,6 +56,21 @@
 
 static volatile ulong * fd_shred_version;
```
ibhatt-jumptrading, line 58: why is this in fd_gossip?

mjain-jump: there were some pubkey util functions used in exec slot ctx for the account compute table that shouldnt be there. the only other place its used is in a map definition in gossip, so ive just moved it there

ibhatt-jumptrading (approved):

PR #3905, opened 2025-01-09
Add ethtool-loopback stage

ripatel-fd:
Fixes Agave<>Firedancer connectivity on loopback.

Agave's QUIC client, the quinn library, uses UDP segmentation which
is incompatible with AF_XDP.  The ethtool-loopback fdctl stage is
introduced to disable the UDP segmentation feature using ethtool.

If the ethtool tx-udp-segmentation feature is enabled on the
loopback device, any sendmsg() calls that send more than one packet
via UDP_SEGMENT result in all packets getting lost.

### DIFF
--- a/src/app/fdctl/configure/ethtool-loopback.c
+++ b/src/app/fdctl/configure/ethtool-loopback.c
@@ -1,0 +1,199 @@
+/* This stage disables the "tx-udp-segmentation" offload on the loopback
+   interface.  If left enabled, AF_XDP will drop loopback UDP packets sent
+   by processes that enable TX segmentation via SOL_UDP/UDP_SEGMENT sockopt
+   or cmsg.
+
+   TLDR tx-udp-segmentation and AF_XDP are incompatible. */
+
+#include "configure.h"
+
+#include <net/if.h>
+#include <stdio.h>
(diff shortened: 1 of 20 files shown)


src/app/fdctl/configure/ethtool-loopback.c
```diff
@@ -0,0 +1,199 @@
...
+    FD_LOG_ERR(( "error configuring network device, socket(AF_INET,SOCK_DGRAM,0) failed (%i-%s)",
+                 errno, fd_io_strerror( errno ) ));
+
+  int feature_idx = find_feature_index( sock, udpseg_feature );
+  if( feature_idx<0 ) {
+    FD_LOG_INFO(( "device `lo` missing ethtool offload `" UDPSEG_FEATURE "`, ignoring" ));
+    CONFIGURE_OK(); /* returns */
+  }
+
+  _Bool udpseg_enabled = get_feature_state( sock, feature_idx );
```
mmcgee-jump, line 176: _Bool?

mmcgee-jump:
Nice! Thx for the doc updates

mmcgee-jump (approved):

PR #3907, opened 2025-01-09
fixed bugs revealed by asan and deepasan; made asan work

asiegel-jt:
important note: asan still only works under clang


config/machine/linux_clang_x86_64.mk
```diff
@@ -18,11 +18,11 @@ include config/extra/with-threads.mk
 
 CPPFLAGS+=-march=haswell -mtune=skylake
 CPPFLAGS+=-DFD_HAS_INT128=1 -DFD_HAS_DOUBLE=1 -DFD_HAS_ALLOCA=1 -DFD_HAS_X86=1 -DFD_HAS_SSE=1 -DFD_HAS_AVX=1
+LDFLAGS+=-lstdc++
```
jumpsiegel, line 21: Is there a way we can add this only when doing asan?   I know you will get pushback adding to the dependency list here...

src/flamenco/leaders/fd_leaders.c
```diff
@@ -71,7 +71,7 @@ fd_epoch_leaders_new( void                    * shmem,
   FD_TEST( laddr+fd_wsample_footprint( pub_cnt, 0 )<=(ulong)wsample_mem + fd_epoch_leaders_footprint( pub_cnt, slot_cnt ) );
 
   /* Create and seed ChaCha20Rng */
-  fd_chacha20rng_t _rng[1];
+  fd_chacha20rng_t * _rng = fd_alloca( alignof(fd_chacha20rng_t), sizeof(fd_chacha20rng_t) );;
   fd_chacha20rng_t * rng = fd_chacha20rng_join( fd_chacha20rng_new( _rng, FD_CHACHA20RNG_MODE_MOD ) );
```
jumpsiegel, line 75: We can revert this change since it was only to fix asan/gcc..

PR #3908, opened 2025-01-09
funk locking fixes

two-heart:
None of these really a problem in their respective contexts, but good to have all code do correct locking, if only for automated tooling.

### DIFF
--- a/src/flamenco/runtime/fd_acc_mgr.c
+++ b/src/flamenco/runtime/fd_acc_mgr.c
@@ -314,11 +314,10 @@ fd_acc_mgr_save_non_tpool( fd_acc_mgr_t *          acc_mgr,
   fd_funk_rec_key_t key = fd_acc_funk_key( account->pubkey );
   fd_funk_t * funk = acc_mgr->funk;
   fd_funk_rec_t * rec = (fd_funk_rec_t *)fd_funk_rec_query( funk, txn, &key );
+  fd_funk_start_write( acc_mgr->funk );
   if( rec == NULL ) {
     int err;
-    fd_funk_start_write( acc_mgr->funk );
     rec = (fd_funk_rec_t *)fd_funk_rec_insert( funk, txn, &key, &err );
-    fd_funk_end_write( acc_mgr->funk );
     if( rec == NULL ) FD_LOG_ERR(( "unable to insert a new record, error %d", err ));
   }
   account->rec = rec;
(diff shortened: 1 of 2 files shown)


mjain-jump (approved):

src/flamenco/runtime/fd_acc_mgr.c
```diff
@@ -314,11 +314,10 @@ fd_acc_mgr_save_non_tpool( fd_acc_mgr_t *          acc_mgr,
   fd_funk_rec_key_t key = fd_acc_funk_key( account->pubkey );
   fd_funk_t * funk = acc_mgr->funk;
   fd_funk_rec_t * rec = (fd_funk_rec_t *)fd_funk_rec_query( funk, txn, &key );
```
lidatong, line 316: looks like this needs a read lock

src/flamenco/runtime/fd_acc_mgr.c
```diff
@@ -330,6 +329,7 @@ fd_acc_mgr_save_non_tpool( fd_acc_mgr_t *          acc_mgr,
   if( fd_funk_val_truncate( account->rec, reclen, fd_funk_alloc( acc_mgr->funk, wksp ), wksp, &err ) == NULL ) {
     FD_LOG_ERR(( "unable to allocate account value, err %d", err ));
   }
+  fd_funk_end_write( acc_mgr->funk );
```
lidatong, line 332: not sure i'm a fan calling LOG_ERR while holding the lock ... funk is now invalid which is an issue for persistent funk

lidatong (changes requested):

src/flamenco/runtime/fd_acc_mgr.c
two-heart: I thought funk allows lock-free read operations. How do I get a read lock for funk?

src/flamenco/runtime/fd_acc_mgr.c
two-heart: There are tons of cases where we FD_LOG_ERR, while holding a funk lock: fd_runtime_process_genesis_block, fd_migrate_builtin_to_core_bpf, fd_runtime_collect_rent_accounts_prune, and many more. Getting rid of this would be a bigger change, or potentially something we want to do atexit.

topointon-jump:
Closing this as we are re-working the Funk accessors in https://github.com/firedancer-io/firedancer/pull/4416

PR #3937, opened 2025-01-13
feat(blockstore): entry batch windowing

lidatong:
Introduce entry batch support to the blockstore and replay in a backwards-compatible way. We continue to deshred and assemble entire blocks, but also add windowing metadata to support a caller copying out shreds and assembling smaller chunks (entry batches) so they don't have to wait for the whole block.

### DIFF
--- a/src/flamenco/runtime/fd_blockstore.c
+++ b/src/flamenco/runtime/fd_blockstore.c
@@ -381,7 +381,7 @@ fd_blockstore_init( fd_blockstore_t * blockstore, int fd, ulong fd_size_max, fd_
   memset( block_map_entry->child_slots, UCHAR_MAX, FD_BLOCKSTORE_CHILD_SLOT_MAX * sizeof( ulong ) );
   block_map_entry->child_slot_cnt  = 0;
 
-  block_map_entry->height          = slot_bank->block_height;
+  block_map_entry->block_height          = slot_bank->block_height;
   memcpy(&block_map_entry->block_hash, slot_bank->block_hash_queue.last_hash, sizeof(fd_hash_t));
   block_map_entry->bank_hash       = slot_bank->banks_hash;
   block_map_entry->block_hash      = slot_bank->poh;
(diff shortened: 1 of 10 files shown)


src/flamenco/runtime/fd_blockstore.c
```diff
@@ -1228,26 +1233,42 @@ fd_buf_shred_insert( fd_blockstore_t * blockstore, fd_shred_t const * shred ) {
...
+  if( FD_UNLIKELY( shred->data.flags & FD_SHRED_DATA_FLAG_DATA_COMPLETE ) ) {
+    fd_block_set_insert( block_map_entry->data_complete_idxs, shred->idx );
+
+    /* If we've contiguously received shreds through this shred marking
+       the end of the entry-batch, we can advance the data_complete_idx
+       watermark. */
+
+    if ( FD_UNLIKELY( block_map_entry->consumed_idx == shred->idx ) ) {
+      block_map_entry->data_complete_idx = shred->idx;
+    }
```
emwang-jumptrading, line 1269: If s4 was a shred w/ batch complete flag and we had received
s1 | s2 | s3| - | s5 | s6 (not complete) |
and then we recieve s4 -> then `consumed_idx` would become 6. `data_complete_idx` would need to be 4? but `consumed_idx != shred->idx`, and `data_complete_idx` wouldn't update. Maybe needs to be `>=`?

emwang-jumptrading: Although now I think there's another edge case where if s6 was also a completed batch shred `data_complete_idx` should update to s6 when s4 arrives. So might actually need to search forward in the for shreds with `data_complete` until `consumed_idx` is cleared 💭

lidatong: you're right it needs to sweep up through consumed_idx

emwang-jumptrading (dismissed):

emwang-jumptrading (approved):

emwang-jumptrading (approved):

PR #3946, opened 2025-01-14
refactor(firedancer): create and document explicit watermark fseqs

lidatong:


src/app/fdctl/run/topos/fd_firedancer.c
```diff
@@ -301,19 +299,111 @@ fd_topo_initialize( config_t * config ) {
...
+     observed >2/3 of stake has rooted. Rooting any of a given slot's
+     descendant implies rooting the slot. The finalized slot is also
+     known as the supermajority root. */
+
+  fd_topob_wksp( topo, "finalized" );
+  fd_topo_obj_t * finalized_obj = fd_topob_obj( topo, "fseq", "finalized" );
+  fd_topob_tile_uses( topo, replay_tile, finalized_obj, FD_SHMEM_JOIN_MODE_READ_WRITE );
+  FD_TEST( fd_pod_insertf_ulong( topo->props, finalized_obj->id, "finalized" ) );
+
+  /* published - this is the minimum of Firedancer's local tower root
```
mmcgee-jump, line 367: This shared state concurrency through fseqs is extremely suspicious and seems quite bug prone. Why not have the tiles pass messages via. the regular mcaches?

lidatong: whats the concern / why bug-prone? seems to me to require a lot less code than mcache message passing and seems simpler to leverage that these are all ulongs. anyone that wants to read these watermarks just joins the fseqs

mmcgee-jump: There's a variety of problems. Why is the caller using these? Now they need to poll continuously rather than being notified. What's the polling frequency? What if we want our tools to have a record of when and why it changed. What if they don't want updates to be coalesced (e.g. 19 -> 20 -> 21, but they only poll at 19 and 20, so miss doing an update at 20). etc etc, the list goes on. This fseq pattern is not a good one.

PR #3973, opened 2025-01-17
Update tuning page with 4 other tiles in fd_frankendancer.c

palmerlao:
While playing around with frankendancer layout configs, it was not clear why our core count and configured tile count were mismatching until we looked into the code to see these other tile types.


lidatong (approved):
thanks for updating the docs on this!

PR #3982, opened 2025-01-21
flamenco: mem* syscalls cannot cross account data and non-account dat…

yufeng-jump:
…a boundaries

Also add the memmove_syscall_check_reverse feature gate and add back relevant test vectors.


src/flamenco/vm/syscall/fd_vm_syscall_util.c
```diff
@@ -461,12 +461,22 @@ fd_vm_memmove( fd_vm_t * vm,
                         dst_is_input_mem_region && 
                         dst_region_idx+1UL<vm->input_mem_regions_cnt &&
                         vm->input_mem_regions[ dst_region_idx+1UL ].is_writable ) ) {
+          if( FD_UNLIKELY( vm->input_mem_regions[ dst_region_idx+1UL ].is_acct_data != vm->input_mem_regions[ dst_region_idx ].is_acct_data ) ) {
+            FD_VM_ERR_FOR_LOG_SYSCALL( vm, FD_VM_SYSCALL_ERR_INVALID_LENGTH );
+            return FD_VM_SYSCALL_ERR_SEGFAULT;
+          }
```
mjain-jump, line 467: You can probably just add this check into the if condition, and it will fall into the `else` block otherwise

mjain-jump (approved):

src/flamenco/vm/syscall/fd_vm_syscall_util.c
mjain-jump: ignore

PR #3998, opened 2025-01-22
pool: add lru accessor

mmcgee-jump:


ptaffet-jump:
I'm going to replace the one use of this feature with a dlist. These pool-based structures compose well, so there's not much advantage to adding it into the internals of pool, and right now pool doesn't care about acquired elements, so the feature feels a bit arbitrary.

PR #3999, opened 2025-01-23
map chain: add fast removal option

mmcgee-jump:


src/util/tmpl/fd_map_chain.c
```diff
@@ -727,6 +741,10 @@ MAP_(idx_insert)( MAP_(t) *   join,
 
   MAP_IDX_T * head = MAP_(private_chain)( map ) + MAP_(private_chain_idx)( &pool[ ele_idx ].MAP_KEY, map->seed, map->chain_cnt );
 
+#if MAP_OPTIMIZE_RANDOM_ACCESS_REMOVAL
+  if( FD_UNLIKELY( *head != MAP_(private_idx_null)() ) ) pool[ *head ].MAP_PREV = MAP_(private_box)( ele_idx );
```
ptaffet-jump, line 745: It seems like the more idiomatic way to check this is with `MAP_(private_idx_is_null)` instead of `==MAP_(private_idx_null)()`

ptaffet-jump (commented):
This is a good change. One nit (that appears in many places). I also kind of think it's an anti-pattern to have a bunch of test_map_chain_* files that are mostly copied and pasted, but oh well.

src/util/tmpl/fd_map_chain.c
ptaffet-jump: I fixed this for you.

nbridge-jump (approved):

PR #4001, opened 2025-01-23
handholding for tmpl data structures

two-heart:
Add debug time assertions to be used as an early bug oracle during fuzzing and other testing
* Only based on local information, so handholding can be used without any changes
* Focus on data structures currently in use Drive-by: typo fixes, make magic unique, mark unreachable code with unreachable builtin


nlgripto (dismissed):

mmcgee-jump:
I feel like if you're going to do this, you should make it more strict? Basically do a tmpl_verify after every operation to check all invariants

two-heart:
@mmcgee-jump I updated the PR to call add calls verify wherever:
(1) a verify function is available
(2) the necessary arguments are locally available
(3) the function modifies the data structure
(4) is not a wrapper around a function that already has equivalent checks

For cases that are not checked because of (2) my intention is to do a separate PR that adds verify calls from the usage side at the end of fuzz tests and other testing.

two-heart:
(rebased on main and resolved conflicts with recent fd_map_chain changes)

two-heart:
Having handholding in fd_map_giant is not practical because pretty soon the very frequently happening inserts/remove will have an overhead of 8-9s just for running verify. This is way too long to run things like ledger tests.

nlgripto (dismissed):

two-heart:
Adjusted the vote program to use the correct wrapping variants of deque push, the previous version might seem to work correctly but leaves the deque data structure in an invalid state.

two-heart:
Following Kevin's comments, I included logging services only when handholding is enabled in places where we don't do other logging. (+ removed trailing whitespace at unrelated code to make the new gh action check happy)

two-heart:
Rebased onto main to resolve map_chain merge conflict.

nlgripto (approved):

kbowers-jump (approved):

PR #4002, opened 2025-01-23
agave: allow setting accounts index and accounts hash cache paths

anwayde:


src/app/fdctl/config/default.toml
```diff
@@ -218,6 +218,19 @@ dynamic_port_range = "8900-9000"
...
     account_index_include_keys = []
 
+    # If account indexes are enabled, this is the absolute directory
+    # path to place the accounts index in.  If this is empty, it
+    # will default to the `accounts_index` subdirectory of the ledger
+    # `path` above.  This option is passed to the Agave client with
+    # the `--accounts-index-path` argument.
+    accounts_index_path = ""
+
+    # Absolute directory path to place the accounts hash cache in.  If
```
mmcgee-jump, line 228: Could you explain a bit here in the comment what the "hash cache" is?

anwayde: added

mmcgee-jump (approved):

PR #4010, opened 2025-01-23
quic: fix keys in quic-trace

0x0ece:


nbridge-jump:
This is good, but there's another instance in fd_quic_trace_handshake

0x0ece:
Good catch, fixed

nbridge-jump (approved):

Issue #3925, opened 2025-01-11
log a warning, don't exit when receiving SIGUSR1


mmcgee-jump:
Operator had their validator die, because logrotate sent a SIGUSR1. We should `LOG_WARNING(( "Firedancer received SIGUSR1 but does not support rotating the log" ));` and then swallow the signal, in the supervisor process. Maybe advise them to use `copy/truncate` instead.

jherrera-jump:
Fixed by https://github.com/firedancer-io/firedancer/pull/4018

PR #4018, opened 2025-01-24
fix(supervisor): gracefully handle SIGUSR1/2 

jherrera-jump:
Addresses https://github.com/firedancer-io/firedancer/issues/3925

### DIFF
--- a/src/app/fdctl/run/run.c
+++ b/src/app/fdctl/run/run.c
@@ -103,6 +103,12 @@ install_parent_signals( void ) {
     FD_LOG_ERR(( "sigaction(SIGTERM) failed (%i-%s)", errno, fd_io_strerror( errno ) ));
   if( FD_UNLIKELY( sigaction( SIGINT, &sa, NULL ) ) )
     FD_LOG_ERR(( "sigaction(SIGINT) failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+
+  sa.sa_handler = SIG_IGN;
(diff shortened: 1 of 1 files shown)


src/app/fdctl/run/main.seccomppolicy
```diff
@@ -29,6 +29,15 @@ write: (or (eq (arg 0) 2)
...
 fsync: (eq (arg 0) logfile_fd)
 
+# supervisor: resume supervisor execution context when returning from
+# a signal handler
+#
+# If the supervisor receives SIGUSR1 or SIGUSR2, rather than terminating
+# the supervisor process, a designated handler function executes.
+# rt_sigreturn resumes the main execution context after the signal
+# handler returns.
+rt_sigreturn
```
mmcgee-jump, line 39: I think per Ben's suggestion, it's better to just set SIG_IGN on SIGUSR1 and SIGUSR2 and have the kernel silently swallow them for us, than to enable rt_sigreturn and log the message.

jherrera-jump: Will change, I agree all the tradeoffs here (complexity, security) seem like a lot for a warning message that can probably be safely ignored.

mmcgee-jump (approved):
Nice! LGTM

mmcgee-jump (changes requested):
Oh but can you squash the commits?

ptaffet-jump (approved):

PR #4033, opened 2025-01-27
Quic: Move unfinished streams from send_streams to used_streams

akhinvasara-jumptrading:
Currently, unfinished streams in send_streams that have no data ready to transmit are simply skipped over. However, this leaves those 'stale' streams in send_streams for every iteration of generating stream frames. This change ensures that unfinished but data-less streams are moved from send_streams into used_streams.

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -3372,7 +3372,7 @@ fd_quic_gen_ping_frame( fd_quic_conn_t *     conn,
   return frame_sz;
 }
 
-static uchar *
+uchar *
 fd_quic_gen_stream_frames( fd_quic_conn_t *     conn,
                            uchar *              payload_ptr,
                            uchar *              payload_end,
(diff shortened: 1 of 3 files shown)


src/waltz/quic/fd_quic.c
```diff
@@ -3433,6 +3434,7 @@ fd_quic_gen_stream_frames( fd_quic_conn_t *     conn,
         ulong data_sz    = fd_ulong_min( data_avail, data_max );
         /* */ data_sz    = fd_ulong_min( data_sz, 0x3fffUL ); /* max 2 byte varint */
         _Bool fin        = fin_avail && data_sz==data_avail;
+        /* */ remove     = fin;
```
nbridge-jump, line 3437: The logic should be:
```
if nothing left to send, then remove from send, add to used, and clear the action flag
```
"nothing to send" means all the bytes we have, have been sent, and the FIN flag has been sent, if requested.
The corner case is where the user sends zero (additional) bytes and the FIN flag. In this case we should still send a stream frame with zero length, but an offset and the FIN flag. If it doesn't fit the stream should remain in the send list, with the upd_pkt_number and flags set to send the stream next time.

nbridge-jump (changes requested):

nbridge-jump (dismissed):

nbridge-jump (approved):

PR #4037, opened 2025-01-28
flamenco: Prevent executing precompiles if they are not native programs

### LINKED ISSUE #4040: Fix edge case when a program is owned by a precompiled program
`prevent_calling_precompiles_as_programs` is an already cleaned up feature and we currently implement this feature by executing a noop instruction execute function for those precompiled programs. However, there are edge cases where a program is owned by the precompiled program. In these cases, we should return unsupported program ID.

Related test fixture: https://github.com/firedancer-io/test-vectors/pull/80

This check should not be needed after the `move_precompile_verification_to_svm` feature is implemented.

cali-jumptrading:
We should not execute precompiled programs if they are not owned by the native loader


cali-jumptrading:
Actually this is just a testcase error.

PR #4041, opened 2025-01-28
flamenco: Skip reset of return data if executing no-op instr

cali-jumptrading:


cali-jumptrading:
superseded by https://github.com/firedancer-io/firedancer/pull/4051

PR #4045, opened 2025-01-29
Update LICENSE

maximevtush:
Updated the copyright year from 2022 to 2025 in the LICENSE file


ripatel-fd:
The year in copyright files typically means "date of first publication".

PR #4046, opened 2025-01-29
topo: detect invalid object align or footprint

ripatel-fd:


src/disco/topo/fd_pod_format.h
```diff
@@ -129,20 +129,34 @@ fd_pod_queryf_##type( uchar const * FD_RESTRICT  pod,
...
+FD_POD_IMPL( char   )
+FD_POD_IMPL( schar  )
+FD_POD_IMPL( uchar  )
+FD_POD_IMPL( float  )
 #if FD_HAS_DOUBLE
-FD_POD_IMPL( double, DOUBLE )
+FD_POD_IMPL( double )
 #endif
 
+static inline uchar const *
```
ptaffet-jump, line 146: Can you add the `__attribute__ ((format (printf, 2, 3)))` (not sure that's exactly right) declaration?

src/disco/topo/fd_pod_format.h
```diff
@@ -112,12 +112,12 @@ FD_POD_IMPL( double, DOUBLE )
    is constructed from the given format string.  Returns the query
    result on success or def on failure. */
 
-#define FD_POD_IMPL(type,TYPE)                                                  \
+#define FD_POD_IMPL(type)                                                  \
```
ptaffet-jump, line 115: nit: re-align `\`

src/disco/topo/fd_pod_format.h
```diff
@@ -14,7 +14,7 @@
 
    IMPORTANT!  THIS IS AN INVALIDATING OPERATION */
 
-#define FD_POD_IMPL(type,TYPE)                                                  \
+#define FD_POD_IMPL(type)                                                  \
```
ptaffet-jump, line 17: nit: re-align `\`

ptaffet-jump (commented):
Some nits, but mostly looks good

ptaffet-jump:
Notice that CI is failing because it's triggering your new check. Can you see if that's a real failure?

ripatel-fd:
@ptaffet-jump Thanks for the review Philip, good points. 
The CI failure was caused by the full validator having tiles with zero footprint. I excluded tiles from the 0 footprint check.

anwayde (approved):

PR #4055, opened 2025-01-30
flamenco: make transaction parsing errors return sanitize failure error

cali-jumptrading:
This only changes the test harness, so should have no effect elsewhere.


src/flamenco/runtime/tests/fd_exec_instr_test.c
```diff
@@ -781,6 +781,7 @@ _txn_context_create_and_exec( fd_exec_instr_test_runner_t *      runner,
   /* Create the raw txn (https://solana.com/docs/core/transactions#transaction-size) */
```
ibhatt-jumptrading, line 781: Generally we don't want to actually return a specific error in the harness. I think a better approach would be to return NULL here. This way these kinds of cases will never be reported by solfuzz.

cali-jumptrading: changed to null in [a1a3074](https://github.com/firedancer-io/firedancer/pull/4055/commits/a1a307466dface45d0d9268f8d3af94fd1c89dbf)

cali-jumptrading:
No longer needed

PR #4063, opened 2025-02-03
fwd geyser plugin cmdline opts along to agave

ebenali:
these are straigtforward, amended defaults toml


ripatel-fd:
Thank you for the contribution. This will probably not work during leader slots.

mmcgee-jump:
We do not support Geyser in Frankendancer, and if it does happen to work for certain case it's unintentional.

PR #4064, opened 2025-02-03
fddev pktgen

ripatel-fd:
Add packet generator tool for benchmarks

### DIFF
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -876,6 +876,12 @@ dynamic_port_range = "8900-9000"
         # handling.
         xdp_aio_depth = 256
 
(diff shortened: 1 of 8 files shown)


src/app/fdctl/config/default.toml
```diff
@@ -1382,3 +1382,7 @@ dynamic_port_range = "8900-9000"
         # duplicate transactions will not be submitted and the validator
         # with this option enabled will always be leader.
         disable_status_cache = false
+
+    # 'fddev pktgen' is an internal development tool.
```
mmcgee-jump, line 1386: Operator facing documentation! Explain the tool here!!

mmcgee-jump (dismissed):
LGTM!

PR #4071, opened 2025-02-03
flamenco: various cpi fixes

yufeng-jump:
Fix serialized owner accessor offset

Rework AccountInfo translation macros.  In particular, make sure that we mimic Agave in checking all of the following:
- Translation permission (r/w)
- Translation order
- Translation offsets, especially for the Rust boxes
- Translation length

Make sure that AccountInfo checks are gated by direct mapping feature flag in a way that matches Agave

Move AccountInfo pubkey translations to earlier in the pipeline and cache the results

Add InvalidArgument check to AccountMeta

Add more MissingAccount returns that match Agave

Signer translation should only be performed when length is non zero

Initialize instr_ctx in the instr stack

Remove translation of error codes in cpi harness

Clean up various error return handling


src/flamenco/vm/syscall/fd_vm_cpi.h
```diff
@@ -148,4 +119,37 @@ struct __attribute__((packed)) fd_vm_rust_account_info {
...
+   facilitate checks on Rust AccountInfo fields.
+ */
+
+#define FD_VM_RC_REFCELL_ALIGN (8UL)
+
+struct __attribute__((packed)) fd_vm_rc_refcell {
+  ulong strong; /* Rc */
+  ulong weak;   /* Rc */
+  ulong borrow; /* RefCell */
+  ulong T[];
```
topointon-jump, line 132: payload?

src/flamenco/vm/syscall/fd_vm_syscall_pda.c
```diff
@@ -96,7 +96,7 @@ fd_vm_translate_and_check_program_address_inputs( fd_vm_t *             vm,
                                                   ulong *               out_seed_szs,
                                                   fd_pubkey_t const * * out_program_id ) {
 
-  fd_vm_vec_t const * untranslated_seeds = FD_VM_MEM_SLICE_HADDR_LD( vm, seeds_vaddr, FD_VM_ALIGN_RUST_SLICE_U8_REF,
+  fd_vm_vec_t const * untranslated_seeds = FD_VM_MEM_SLICE_HADDR_LD( vm, seeds_vaddr, FD_VM_VEC_ALIGN,
```
topointon-jump, line 99: can we not use FD_VM_VEC_ALIGN?

topointon-jump (approved):

PR #4073, opened 2025-02-04
remove trailing whitespace from files

anwayde:


ripatel-fd:
I don't think we should do this, this will cause merge conflicts everywhere

PR #4075, opened 2025-02-04
Improve net tile TX throughput

riptl:
- **fddev pktgen**
- **Add pktgen fake_dst_ip config option**
- **Add net_xsk_send_errors metric**
- **Net tile TX coalescing**


src/app/fdctl/config/default.toml
```diff
@@ -876,6 +876,12 @@ dynamic_port_range = "8900-9000"
         # handling.
         xdp_aio_depth = 256
 
+        # When the oldest packet in a batch has been queued for more than
```
mmcgee-jump, line 879: Is this really operator configurable?

src/waltz/xdp/fd_xsk.c
```diff
@@ -726,12 +730,13 @@ fd_xsk_tx_enqueue( fd_xsk_t *            xsk,
     if( fd_xsk_tx_need_wakeup( xsk ) ) {
       if( FD_UNLIKELY( -1==sendto( xsk->xsk_fd, NULL, 0, MSG_DONTWAIT, NULL, 0 ) ) ) {
         if( FD_UNLIKELY( errno!=EAGAIN ) ) {
-          FD_LOG_WARNING(( "xsk sendto failed xsk_fd=%d (%i-%s)", xsk->xsk_fd, errno, fd_io_strerror( errno ) ));
+          *opt_perr = errno;
```
mmcgee-jump, line 733: I don't really like hiding this behind a metric. If `sendto` can't wake the driver, packets might never send, right? This is a fatal condition?

mmcgee-jump (changes requested):
comments inline

nbridge-jump (approved):

PR #4087, opened 2025-02-05
flamenco: Add asserts for error propagation checks

cali-jumptrading:
This PR checks for error propagation overwrites and existence in the runtime logic using the `FD_RUNTIME_ERR_HANDHOLDING` check. In the future, the `HANDHOLDING` checks should be consolidated into runtime and debug assertions.


src/flamenco/vm/fd_vm_private.h
```diff
@@ -204,17 +204,28 @@ FD_PROTOTYPES_BEGIN
 
 /* Log error within the instr_ctx to match Agave/Rust error. */
 
+#define FD_VM_DEBUG_TEST_ERR_EXISTS( vm )                                 \
+    FD_DEBUG_TEST( vm->instr_ctx->txn_ctx->exec_err );                    \
+    FD_DEBUG_TEST( vm->instr_ctx->txn_ctx->exec_err_kind )                \
+
+#define FD_VM_DEBUG_TEST_ERR_OVERWRITE( vm )                              \
+    FD_DEBUG_TEST( !vm->instr_ctx->txn_ctx->exec_err );                   \
+    FD_DEBUG_TEST( !vm->instr_ctx->txn_ctx->exec_err_kind )               \
```
topointon-jump, line 213: nit: could we add a comment explaining what these do/why we need them?

topointon-jump (approved):
nice! does this also get enabled in fuzzing runs?

yufeng-jump (commented):
This is great!

Rather than introducing a FD_DEBUG_TEST() and enabling in with-debug, maybe we define a new handholding macro in with-handholding and gate the err code checks behind this handholding macro.  We already enable all the handholding macros in offline replay and fuzz, so that gets us lots of coverage.

src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -249,6 +249,7 @@
   cu = fd_ulong_min( cu_req, cu );                                            \
   if( FD_UNLIKELY( err ) ) {                                                  \
     if( err==FD_VM_SYSCALL_ERR_COMPUTE_BUDGET_EXCEEDED ) cu = 0UL; /* cmov */ \
+    FD_VM_DEBUG_TEST_ERR_EXISTS( vm );                                        \
```
ravyu-jump, line 252: Won't this hit all the time, assuming syscalls (correctly?) set the `exec_err`?

ravyu-jump (changes requested):
LGTM, just one comment that might need clarification 

we also need to add the `debug` extra to the ci build targets so that we actually run these asserts in CI. Example: https://github.com/firedancer-io/firedancer/blob/6e8f845245de4f27b2fb4eefc6917733aa7aad5d/.github/workflows/tests.yml#L42

src/flamenco/vm/fd_vm_interp_core.c
ravyu-jump: nvm i mixed this up with `ERR_OVERWRITE

ravyu-jump (approved):

ravyu-jump (commented):
> Rather than introducing a FD_DEBUG_TEST() and enabling in with-debug, maybe we define a new handholding macro in with-handholding and gate the err code checks behind this handholding macro. We already enable all the handholding macros in offline replay and fuzz, so that gets us lots of coverage

Agree with macro guarding this with `FD_XYZ_HANDHOLDING` and using the `with-handholding` extra instead

src/flamenco/vm/fd_vm_interp_core.c
cali-jumptrading: No, this only triggers an assert failure if the error is not set (err is not 0 and err_kind is not 0)

src/flamenco/vm/fd_vm_private.h
cali-jumptrading: added in [a60a899](https://github.com/firedancer-io/firedancer/pull/4087/commits/a60a899abb47237ceaeed0cff077f74b5e598e07)

.github/workflows/tests.yml
```diff
@@ -76,15 +76,15 @@ jobs:
             machine: native
             label: 512G
             extras: "rpath fuzz-stubs"
-            targets: "all integration-test fdctl"
+            targets: "all integration-test fdctl handholding"
```
ravyu-jump, line 79: did u mean to add this to `extras`?

config/extra/with-handholding.mk
```diff
@@ -5,3 +5,4 @@ CPPFLAGS+=-DFD_SCRATCH_USE_HANDHOLDING=1
 CPPFLAGS+=-DFD_SPAD_USE_HANDHOLDING=1
 CPPFLAGS+=-DFD_TOWER_USE_HANDHOLDING=1
 CPPFLAGS+=-DFD_TXN_HANDHOLDING=1
+CPPFLAGS+=-DFD_ERR_HANDHOLDING=1
```
ravyu-jump, line 8: Something like `FD_RUNTIME_ERR_HANDHOLDING` might be more specific

.github/workflows/tests.yml
cali-jumptrading: oops fixed [5919c4c](https://github.com/firedancer-io/firedancer/pull/4087/commits/5919c4c952ba92c3e007fddbcd3bb6c6985365b6)

config/extra/with-handholding.mk
cali-jumptrading: yep, agreed! done in latest commit [a36bb6d](https://github.com/firedancer-io/firedancer/pull/4087/commits/a36bb6d0b158304f0b73abbf3071142521e10fa5)

ravyu-jump (approved):

ravyu-jump (approved):

PR #4098, opened 2025-02-05
flamenco: cpi fixes

yufeng-jump:
Hoist translate_instruction final checks into the right place

Move MaxInstructionTraceLengthExceeded check to right before instruction execution


kbhargava-jump (approved):
looks good, this fixed the interp diff as well :)

PR #4105, opened 2025-02-06
Remove unused tiles.net.multihome_ip_addrs config option

ripatel-fd:
This config option does nothing at all right now.
It won't be supported in the future either.


ptaffet-jump (approved):
This is only half-implemented, so I agree it shouldn't be exposed.

PR #4107, opened 2025-02-06
flamenco, fuzz: add secp256r1 program feature-gated writable checks

mjain-jump:


src/flamenco/runtime/context/fd_exec_txn_ctx.c
```diff
@@ -306,27 +306,32 @@ fd_txn_account_is_demotion( fd_exec_txn_ctx_t const * txn_ctx, int idx )
...
     return 0;
   }
 
+  /* https://github.com/anza-xyz/agave/blob/v2.1.11/sdk/program/src/message/sanitized.rs#L45 */
   if( fd_txn_account_is_demotion( txn_ctx, idx ) ) {
     return 0;
   }
 
-  return fd_txn_is_writable( txn_ctx->txn_descriptor, idx );
+  return 1;
```
cali-jumptrading, line 336: why the change here to return 1 instead of `fd_txn_writable`?

mjain-jump: the txn writable check comes first now to remove some of the extra code we had at the top of the function

cali-jumptrading: oh wait I see that you put in a check for `fd_txn_writable` at the beginning of the function. Could we keep the return `fd_txn_is_writable` and only have checks for special cases in the function?

mjain-jump: the fd txn is writable performs checks to ensure the idx value is within a correct range. if we moved the txn is writable check to the bottom then we'd have to include redundant checks for the idx since we would access txn_ctx->accounts[idx] before

src/flamenco/runtime/context/fd_exec_txn_ctx.c
```diff
@@ -306,27 +306,32 @@ fd_txn_account_is_demotion( fd_exec_txn_ctx_t const * txn_ctx, int idx )
...
   }
 
+  /* See comments in fd_system_ids.h.
+     https://github.com/anza-xyz/agave/blob/v2.1.11/sdk/program/src/message/sanitized.rs#L44 */
   if( fd_pubkey_is_active_reserved_key(&txn_ctx->accounts[idx] ) ||
-      ( FD_FEATURE_ACTIVE( txn_ctx->slot_ctx, add_new_reserved_account_keys ) && 
-                           fd_pubkey_is_pending_reserved_key( &txn_ctx->accounts[idx] ) )) {
+      ( FD_FEATURE_ACTIVE( txn_ctx->slot_ctx, add_new_reserved_account_keys ) &&
+                           fd_pubkey_is_pending_reserved_key( &txn_ctx->accounts[idx] ) ) ||
+      ( FD_FEATURE_ACTIVE( txn_ctx->slot_ctx, enable_secp256r1_precompile ) &&
```
cali-jumptrading, line 326: Could we collapse `fd_pubkey_is_secp256r1_key` into `fd_pubkey_is_pending_reserved_key`? Something like
```
int fd_pubkey_is_pending_reserved_key( fd_pubkey_t const * acct, slot_ctx ) {
    if FD_FEATURE_ACTIVE(...) &&  fd_pubkey_pending_reserved_keys_tbl_contains( acct ) {
        return 1;
    }
    else if FD_FEATURE_ACTIVE(...) && fd_pubkey_is_secp256r1_key( acct ) {
        return 1;
    }
    else {
        return 0;
    }
}
```

mjain-jump: i wanted to keep the functions agnostic of the slot context and feature set and leave the feature checking responsibility up to the caller

src/flamenco/runtime/context/fd_exec_txn_ctx.c
cali-jumptrading: ohh got it.

src/flamenco/runtime/context/fd_exec_txn_ctx.c
cali-jumptrading: Yeah makes sense, if in the future we need this in more than once place, then we should make checking if a key is a reserved key a function.

src/flamenco/runtime/test_system_ids.c
```diff
@@ -49,6 +48,12 @@ old_fd_pubkey_is_pending_reserved( fd_pubkey_t const * acct ) {
   return 0;
 }
 
+static inline int
+old_fd_pubkey_is_secp256r1( fd_pubkey_t const * acct ) {
```
cali-jumptrading, line 52: just a nit on the code format here, for spaces after parenthesis.

src/flamenco/runtime/fd_system_ids.c
```diff
@@ -113,10 +113,21 @@ const fd_pubkey_t fd_solana_stake_program_buffer_address                = { .uc
...
+}
+
+int
+fd_pubkey_is_pending_reserved_key( fd_pubkey_t const * acct ) {
+  return fd_pubkey_pending_reserved_keys_tbl_contains( acct );
+}
+
+int
+fd_pubkey_is_secp256r1_key( fd_pubkey_t const * acct ) {
+  return 0==memcmp( acct->uc, fd_solana_secp256r1_program_id.key, sizeof(fd_pubkey_t) );
```
cali-jumptrading, line 132: also a nit here to change the order of the `memcmp` and 0

cali-jumptrading (dismissed):
LGTM

src/flamenco/runtime/test_system_ids.c
```diff
@@ -2,50 +2,55 @@
...
+  if( memcmp( acct->key, fd_solana_zk_token_proof_program_id.key, sizeof(fd_pubkey_t) )==0 ) return 1;
+  if( memcmp( acct->key, fd_sysvar_epoch_rewards_id.key, sizeof(fd_pubkey_t) )==0 ) return 1;
+  if( memcmp( acct->key, fd_sysvar_last_restart_slot_id.key, sizeof(fd_pubkey_t) )==0 ) return 1;
+  if( memcmp( acct->key, fd_sysvar_owner_id.key, sizeof(fd_pubkey_t) )==0 ) return 1;
+  return 0;
+}
+
+static inline int
+old_fd_pubkey_is_secp256r1( fd_pubkey_t const * acct ) {
+  if( memcmp( acct->key, fd_solana_secp256r1_program_id.key, sizeof(fd_pubkey_t) )==0) return 1;
```
cali-jumptrading, line 53: just one more nit: the space after the 0.

cali-jumptrading (approved):

PR #4110, opened 2025-02-06
Quic: Fix NetEm Reorder, and pkt_meta->expiry margin sign

akhinvasara-jumptrading:
Fixes drops caused by the old reorder emulation by adding an extra buffer space. Respects reordering and avoids starvation via weighted coin toss. 

Also fixes a previous bug in calculating margin of safety on pkt_meta expiry.

### DIFF
--- a/src/waltz/quic/tests/fd_quic_test_helpers.c
+++ b/src/waltz/quic/tests/fd_quic_test_helpers.c
@@ -643,6 +643,7 @@ fd_quic_netem_send( void *                    ctx, /* fd_quic_net_em_t */
     static FD_TL uint seed = 0; /* FIXME use fd_log_tid */
     ulong l = fd_rng_private_expand( seed++ );
     float rnd_num = (float)l * (float)0x1p-64;
+    int weighted_tail = (int)((l&0x7)==0x7); /* 12.5% chance of being 1, else head */
 
     if( rnd_num < mitm_ctx->thresh_drop ) {
       /* dropping behaves as-if the send was successful */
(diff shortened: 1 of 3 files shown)


src/waltz/quic/tests/fd_quic_test_helpers.c
```diff
@@ -643,27 +643,32 @@ fd_quic_netem_send( void *                    ctx, /* fd_quic_net_em_t */
...
-           if we already have a reordered buffer, delay it another packet
-           else store the current packet into the reorder buffer */
-      if( mitm_ctx->reorder_sz > 0UL ) {
-        fd_aio_pkt_info_t lcl_batch[1] = { batch[j] };
-        fd_aio_send( mitm_ctx->dst, lcl_batch, 1UL, NULL, flush );
-
-        /* clear buffer */
-        mitm_ctx->reorder_sz = 0UL;
+      /* logic: if either buf free, buf it. Else, flush and replace send more recent one if head */
+      schar free = -1;
```
nbridge-jump, line 655: Since you're using this as an index, it should be ptr sized - make it a ulong.
The main use for narrow types is in structs in memory for optimizing layout. Local var like this are generally held in registers, and it will never pack different vars in separate parts of RAX, for example.
Making it a narrow type might cause the compiler to generate extra instructions for widening for the pointer arithmetic.
If there are a lot of local vars, they could spill to the stack, and there might be cause for optimization. But the general rule should be use ulong for local index vars, or vars for pointer arithmetic of any kind. Even when fetching from a narrow area it might be worth doing an explicit cast, so the code more accurately represents at first read what's happening.

nbridge-jump (approved):

PR #4111, opened 2025-02-06
Fixed "upstream" includes

dragonfrond:
Currently `util` is dependent on `ballet` and `ballet` is dependent on `disco`.

Fixed this by moving a test file into `util`, and `ballet/pack` into `disco`.


src/disco/pack/generate_delays.py
```diff
@@ -1,5 +1,5 @@
 # FIXME: This file belongs with the tile code.  Rather than clutter
-# fdctl/run/tiles with these two files, I'm putting this in ballet/pack.
+# fdctl/run/tiles with these two files, I'm putting this in disco/pack.
 # Really, all of this belongs in disco.
```
ptaffet-jump, line 3: Thanks to you, this line can now be deleted.

ptaffet-jump (commented):
Thanks! Looks good. Could you squash the commits and name it something like `pack: move to disco`

dragonfrond:
@ptaffet-jump squashed and pushed, should be clean now.

ptaffet-jump (approved):

PR #4135, opened 2025-02-07
Net tile rewrite

ripatel-fd:
- **Net tile TX coalescing**
- **Level-triggered TX batch flush**
- **Fix tiles.net.xdp_mode generic support**
- **Remove quic round_robin**
- **Add quic_retry_sent metric**
- **Rewrite XDP IO in net tile**


src/app/fdctl/run/tiles/fd_net.c
```diff
@@ -462,50 +514,338 @@ after_frag( fd_net_ctx_t *      ctx,
...
+}
+
+/* net_rx_event is called when a new XDP RX frame is available.  Calls
+   net_rx_packet, then returns the packet back to the kernel via the fill
+   ring.  */
+
+static void
+net_rx_event( fd_net_ctx_t * ctx,
+              fd_xsk_t *     xsk,
+              uint           rx_seq ) {
```
topointon-jump, line 733: minor thing - could we also pass stem to this handler? so that we can respect our consumers flow control (in archive capture mode)

if you're on board, I'll make this change in a follow-up pr

topointon-jump: see my hack here: https://github.com/firedancer-io/firedancer/commit/a889bcf4907be89e0af71ea0086f04f5792ba632#diff-5766556c14fd74914668b0e8df6dcd34184a15a9010b7fa1d8038a9391a47b6bR203-R207

topointon-jump: i'm happy to add flow control support to this tile after this re-write is merged in 😄 I have a hacked version of it in my branch

ripatel-fd:
Moving to #4141

PR #4153, opened 2025-02-09
Avoid "Could not find compiler set in environment variable CC" Error

sense112:
# Suggestion to update the cmake command to avoid C compiler errors 

While trying to setup my firedancer node I ran into a strange issue. While trying to run the `./deps.sh` command I was met with the following error message:

```
[+] Configuring secp256k1
CMake Error at /usr/share/cmake-3.22/Modules/CMakeDetermineCCompiler.cmake:49 (message):
  Could not find compiler set in environment variable CC:

  /usr/bin/gcc.
Call Stack (most recent call first):
  CMakeLists.txt:10 (project)


CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
See also "/home/ubuntu/repos/firedancer/opt/git/secp256k1/build/CMakeFiles/CMakeOutput.log".
```

While I understand this is likely due to a misconfiguration from my side installing the prior dependencies, the change in this PR to explicitly set the `C_COMPILER` during the cmake command solved my issue.

This may come in helpful to future validators.


ripatel-fd:
Thank you for the fix and sorry for the issues with the dependency. I'm going to close this since this would put a hard dependency on Clang. Firedancer can currently build with only GCC and we would like to keep that capability. 

As a workaround, we can simply not compile secp256k1. https://github.com/firedancer-io/firedancer/pull/4270
In the future, the secp256k1 dependency is likely going to be removed.

PR #4163, opened 2025-02-10
ipv4: fix fd_cstr_to_ip4_addr docs

two-heart:


src/util/net/fd_ip4.h
```diff
@@ -183,7 +183,7 @@ fd_ip4_hdr_check_fast( void const * vp_hdr ) {
 
 /* fd_cstr_to_ip4_addr parses an IPv4 address matching format
    %u.%u.%u.%u  On success stores address to out and returns 1. On fail
-   returns 0.  The given address is returned in host byte order such
+   returns 0.  The given address is returned in network byte order such
    that "1.0.0.0" => 0x01000000. */
```
ravyu-jump, line 187:
```suggestion
   that "1.0.0.0" => 0x00000001. */
```

ravyu-jump, line 187:
```suggestion
   that "1.0.0.0" => 0x00000001. */
```

ravyu-jump: Should probably be more specific: `0x00000001` is what is returned, which is `01 00 00 00` in-memory (and thus `0x01000000` in network ordering)

PR #4168, opened 2025-02-10
sandbox: allow retaining capabilities

ripatel-fd:


src/util/sandbox/fd_sandbox.c
```diff
@@ -444,7 +445,14 @@ fd_sandbox_private_drop_caps( ulong cap_last_cap ) {
 
   struct __user_cap_header_struct hdr = { _LINUX_CAPABILITY_VERSION_3, 0 };
   struct __user_cap_data_struct   data[2] = { { 0 } };
-  if( -1==syscall( SYS_capset, &hdr, data ) )                          FD_LOG_ERR(( "syscall(SYS_capset) failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+  if( -1==syscall( SYS_capget, &hdr, data ) ) FD_LOG_ERR(( "syscall(SYS_capget) failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+  data[ 0 ].permitted &= (uint)( desired_capabilities     );
+  data[ 1 ].permitted &= (uint)( desired_capabilities>>32 );
+  data[ 0 ].effective = data[ 0 ].permitted;
+  data[ 1 ].effective = data[ 1 ].permitted;
+  data[ 0 ].inheritable = 0U;
```
mmcgee-jump, line 453: These are already 0?

src/util/sandbox/fd_sandbox.c
```diff
@@ -444,7 +445,14 @@ fd_sandbox_private_drop_caps( ulong cap_last_cap ) {
 
   struct __user_cap_header_struct hdr = { _LINUX_CAPABILITY_VERSION_3, 0 };
   struct __user_cap_data_struct   data[2] = { { 0 } };
-  if( -1==syscall( SYS_capset, &hdr, data ) )                          FD_LOG_ERR(( "syscall(SYS_capset) failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+  if( -1==syscall( SYS_capget, &hdr, data ) ) FD_LOG_ERR(( "syscall(SYS_capget) failed (%i-%s)", errno, fd_io_strerror( errno ) ));
```
mmcgee-jump, line 448: Why do you capget? The point is to clear the caps, you can just set it to keep the `desired` set, no need to keep the other stuff

mmcgee-jump (changes requested):
.

src/disco/stem/fd_stem.c
```diff
@@ -554,6 +554,11 @@ STEM_(run1)( ulong                        in_cnt,
         this_in->accum[ FD_METRICS_COUNTER_LINK_OVERRUN_POLLING_COUNT_OFF ]++;
         this_in->accum[ FD_METRICS_COUNTER_LINK_OVERRUN_POLLING_FRAG_COUNT_OFF ] += (uint)(-diff);
       }
+
+#ifdef STEM_CALLBACK_AFTER_POLL_OVERRUN
```
mmcgee-jump, line 558: Wrong place, needs to be in the if statement above?

mmcgee-jump, line 558: Also need to document this, see the top of the file

mmcgee-jump (dismissed):

ripatel-fd:
Trying out a different approach

PR #4169, opened 2025-02-10
Fix bugs in test_sandbox

ripatel-fd:
- test_sandbox checks for max_mnt_namespaces==1 but fd_sandbox
  sets it to 2
- test_sandbox checks for RLIMIT_NICE=1 but fd_sandbox sets it to 1


ripatel-fd:
@mmcgee-jump I'm not sure about these fixes. Can you sanity check them?

mmcgee-jump (approved):

PR #4174, opened 2025-02-10
map_chain: fix query with random access removal

mmcgee-jump:


ptaffet-jump (dismissed):

ptaffet-jump (commented):
Wait, let me think a bit harder

ptaffet-jump (approved):

PR #4176, opened 2025-02-10
config: move firedancer specific option

mmcgee-jump:


ripatel-fd:
@mmcgee-jump I think you need to remove that from the config_parse checks too

PR #4188, opened 2025-02-11
Quic: Ack overflow probes, 10x idle_timeout buffer

akhinvasara-jumptrading:
This PR adds probe points related to ACK behavior, and a new metric for when the ACK buffer overflows.

### DIFF
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -945,7 +945,7 @@ dynamic_port_range = "8900-9000"
         idle_timeout_millis = 10000
 
         # Max delay for outgoing ACKs.
-        ack_delay_millis = 50
+        ack_delay_millis = 25
 
         # QUIC retry is a feature to combat new connection request
         # spamming.  See rfc9000 8.1.2 for more details.  This flag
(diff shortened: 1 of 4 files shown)


ripatel-fd:
@akhinvasara-jumptrading 

```
src/waltz/quic/fd_quic_ack_tx.c:28:38: error: parameter 'ack_overflow_cnt' set but not used [-Werror,-Wunused-but-set-parameter]
                 ulong             * ack_overflow_cnt ) {
```

mmcgee-jump:
Fix the commit message please

src/app/fdctl/config/default.toml
```diff
@@ -945,7 +945,7 @@ dynamic_port_range = "8900-9000"
         idle_timeout_millis = 10000
 
         # Max delay for outgoing ACKs.
-        ack_delay_millis = 50
+        ack_delay_millis = 25
```
nbridge-jump, line 948: Why are we changing the max ack delay in the default.toml?

akhinvasara-jumptrading: Not actually merging this rn, but just worth noting 25 is what the QUIC spec recommends. Directionally, this may contribute to our under-delivery issue (although unlikely)

nbridge-jump: I'm not sure we have an under-delivery issue :-)
But PRs that add tests or change testing behavior should not change production config, as a general rule.

ripatel-fd:
Closing stale PR

PR #4203, opened 2025-02-13
fix buffer overrun in disco/topo

Managarmrr:
When printing tile layout information the macro PRINTOUT tries to ensure that the buffer of size 256 is not overrun. Unfortunately due to a copy-paste bug this check is not properly done and long output will cause a buffer overflow.

This issue doesn't seem to surface when using the recommended build steps, but due to me using a custom build setup my compile flags are slightly different, causing me to run into this issue every time.


mmcgee-jump (approved):

mmcgee-jump:
Thank you for the contribution!

PR #4204, opened 2025-02-13
Support selective fdctl topo

ripatel-fd:
Allows running `fddev pktgen` and `fddev load` without Firedancer tiles


ripatel-fd:
Closing this, will reopen together with the XDP zero copy patch.

PR #4213, opened 2025-02-13
updated getting-started section to latest version

pkrasam:
made these changes
```sh [bash]
$ git clone --recurse-submodules https://github.com/firedancer-io/firedancer.git
$ cd firedancer
$ git checkout v0.305.20111 # Or the latest Frankendancer release
```


ripatel-fd:
Thank you for the contribution

PR #4217, opened 2025-02-13
fdctl: add sigusr doc

jherrera-jump:
Document https://github.com/firedancer-io/firedancer/pull/4018

### DIFF
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -100,6 +100,11 @@ dynamic_port_range = "8900-9000"
 #    NOTICE  2023-01-23 04:56:07.890123456 GMT-06 45678:45678 user:host:f0 app:thread:0 src/file.c(901)[func]: 1 is the loneliest number
 #
 # to the permanent log (log file).
+#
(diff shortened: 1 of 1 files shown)


src/app/fdctl/config/default.toml
```diff
@@ -100,6 +100,11 @@ dynamic_port_range = "8900-9000"
 #    NOTICE  2023-01-23 04:56:07.890123456 GMT-06 45678:45678 user:host:f0 app:thread:0 src/file.c(901)[func]: 1 is the loneliest number
 #
 # to the permanent log (log file).
+#
+# If you want to periodically rotate the your log file, you will not be
```
mmcgee-jump, line 104: I wouldn't address the operator directly like this, speak more illustrative / 3rd person.

> Firedancer does not support rotating the logfile in response to SIGUSR1,
> instead it will silently swallow and ignore this signal. To perform log rotation,
> it suggested to use a mechanism like the `copytruncate` directive of
> logrotate.

mmcgee-jump (changes requested):
.

mmcgee-jump (approved):

PR #4227, opened 2025-02-14
vm: check imm is not the map null key for FD_SBPF_OP_CALL_IMM

two-heart:
sbpf v0 does not validate FD_SBPF_OP_CALL_IMM


ravyu-jump (approved):
LGTM

PR #4238, opened 2025-02-16
Modularize fdctl topology

ripatel-fd:
Allow creating custom topologies


ripatel-fd:
@mmcgee-jump Abandoning this for now. Can pick this up post v0.4.

PR #4251, opened 2025-02-18
Decrease metric tile CPU usage

ripatel-fd:
Use blocking I/O to poll metric tile sockets

### DIFF
--- a/src/app/fdctl/run/tiles/fd_metric.c
+++ b/src/app/fdctl/run/tiles/fd_metric.c
@@ -1,24 +1,19 @@
-#include "../../../../disco/tiles.h"
+#include "../../version.h"
 
+#include "../../../../disco/metrics/fd_prometheus.h"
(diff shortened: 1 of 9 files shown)


src/app/fdctl/run/tiles/fd_metric.c
```diff
@@ -60,7 +60,7 @@ before_credit( fd_metric_ctx_t *   ctx,
                int *               charge_busy ) {
   (void)stem;
 
-  *charge_busy = fd_http_server_poll( ctx->metrics_server );
+  *charge_busy = fd_http_server_poll( ctx->metrics_server, -1 );
```
mmcgee-jump, line 63: You still need to heartbeat, block at most til next housekeep

mmcgee-jump (changes requested):
.

mmcgee-jump (dismissed):

kbhargava-jump (approved):

PR #4270, opened 2025-02-20
Make secp256k1 a dev-only dependency

ripatel-fd:
Operators keep having issues with the secp256k1 dependency.
Demote it to 'deps.sh +dev' only, since it's only required for
full Firedancer, but not Frankendancer.


mmcgee-jump (approved):

mmcgee-jump:
Can you do s2n-bignum as well?

ripatel-fd:
> Can you do s2n-bignum as well?

@mmcgee-jump I'd rather keep that dep in. secp256k1 is eventually going to go even for the runtime team.
Whereas s2n-bignum will stay. s2n-bignum is compiled using the FD build system so it's pretty stable.

PR #4272, opened 2025-02-20
Remove JIT compiler experiment

ripatel-fd:
ABIv1 direct mapping enforces protocol changes that severely
decrease BPF instruction per x86 cycle throughput to the point
where a JIT compiler becomes useless.

Direct mapping also took longer than expected to ship, effectively
freezing any ABIv2 work where a JIT compiler makes sense.

For now, remove the unused code.  Can be brought back in the future.


ibhatt-jumptrading (approved):

mmcgee-jump:
RIP

PR #4273, opened 2025-02-20
Remove docker-compose configs

ripatel-fd:
The images these docker-compose configs refer to don't seem to be
documented anywhere.


mmcgee-jump (approved):

ripatel-fd:
Waiting on @anwayde's review. In case there is some prod infra that depends on this (the test cluster? Monogon?)

anwayde (approved):

anwayde:
> Waiting on @anwayde's review. In case there is some prod infra that depends on this (the test cluster? Monogon?)

Nope, good to delete.

PR #4274, opened 2025-02-20
Move dev instructions to CONTRIBUTING.md

ripatel-fd:
The README should be operator oriented, not dev oriented.
Some operators seemed to blindly run 'deps.sh +dev' and 'make -j run'
thinking they are the instructions to run a production node.

Developers will know to look in CONTRIBUTING.md.


mmcgee-jump:
I think the top level README should indicate how to run a development cluster ... How about just removing the `+dev` ?

mmcgee-jump (changes requested):
.

PR #4291, opened 2025-02-25
metrics: add enum lengths

ravyu-jump:
for array sizing


ripatel-fd:
@ravyu-jump These defines are already available. They are in tile metrics, not enum metrics. E.g.

`#define FD_METRICS_COUNTER_QUIC_RECEIVED_FRAMES_CNT  (22UL)`

(Number of enum variants for the "quic_received_frames" metric)

ravyu-jump:
> These defines are already available. They are in tile metrics, not enum metrics. E.g.

Yup I noticed that. It gets a little confusing when creating cartesian products of 2 (or more) enums in the C code. Look at this change I am currently working on: https://github.com/firedancer-io/firedancer/commit/d9641fb1b39709f0e69f7449affa8e6bb64b568a#diff-5683ec9d3677dc0143727601c1977ee82972d45a4c53f892b13a7d4ca3cf257fR173

```
ulong recv_crds[FD_GOSSIP_ROUTE_COUNT][FD_METRICS_COUNTER_GOSSIP_RECEIVED_CRDS_PUSH_CNT];
```
Note that `FD_GOSSIP_ROUTE` is {push, pull}. So this is somewhat misleading because the second dimension is technically sized on the number of elements in the `CrdsValue` enum, not the `ReceivedCrdsPush` metric, even though the latter is just an instantiation of the `CrdsValue` enum.  Maybe I'm misunderstanding the prometheus spec, but the above doesn't seem right to me.

```
ulong recv_crds[FD_GOSSIP_ROUTE_COUNT][FD_METRICS_ENUM_CRDS_VALUE_CNT];
```

Seems cleaner to me

PR #4303, opened 2025-02-25
gui: add validator icons/names

jherrera-jump:
- Add link for validator info messages
- fix message size calculation in plugin
- send info messages as they are found in accounts db (rather than all at one at the end)

agave [diff](https://github.com/firedancer-io/agave/compare/5f63d808b5..jherrera/gui-fetch-peer-info-2~7)

### DIFF
--- a/src/app/fdctl/run/tiles/fd_plugin.c
+++ b/src/app/fdctl/run/tiles/fd_plugin.c
@@ -13,6 +13,7 @@
 #define IN_KIND_STARTP (5)
 #define IN_KIND_VOTEL  (6)
 #define IN_KIND_BUNDLE (7)
+#define IN_KIND_VALCFG (8)
 
 typedef struct {
   fd_wksp_t * mem;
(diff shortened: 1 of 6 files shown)


src/disco/gui/fd_gui.c
```diff
@@ -751,27 +751,13 @@ fd_gui_handle_validator_info_update( fd_gui_t *    gui,
   ulong updated[ 40200 ] = {0};
 
   ulong removed_cnt = 0UL;
-  fd_pubkey_t removed[ 40200 ] = {0};
+  /* Unlike gossip or vote account updates, validator info messages come
```
jherrera-jump, line 754: Not sure whether I should try to keep this in. I opted to just remove it since on the off chance the accounts index is enabled and someone removes their name/icon from the chain, I feel like we wouldn't need to have that update on the gui immediately. I could keep it and remove sending-while scanning, but validators would need to wait 40min to see info

src/app/fdctl/run/tiles/fd_plugin.c
```diff
@@ -57,7 +58,7 @@ during_frag( fd_plugin_ctx_t * ctx,
   ulong * dst = (ulong *)fd_chunk_to_laddr( ctx->out_mem, ctx->out_chunk );
 
   /* ... todo... sigh, sz is not correct since it's too big */
-  if( FD_UNLIKELY( ctx->in_kind[ in_idx ]==IN_KIND_GOSSIP ) && FD_LIKELY( sig==FD_PLUGIN_MSG_GOSSIP_UPDATE || sig==FD_PLUGIN_MSG_VALIDATOR_INFO ) ) {
+  if( FD_UNLIKELY( ctx->in_kind[ in_idx ]==IN_KIND_GOSSIP ) && FD_LIKELY( sig==FD_PLUGIN_MSG_GOSSIP_UPDATE ) ) {
```
mmcgee-jump, line 61: The `FD_UNLIKELY` && `FD_LIKELY` here is meaningless, you need to wrap the && like `FD_LIKELY( a && b)`

mmcgee-jump (approved):

PR #4304, opened 2025-02-25
Update testnet-jito.toml

ripatel-fd:
Don't use internal Jito domain name


mmcgee-jump (changes requested):
Can you test that it successfully connects to testnet?

ripatel-fd:
@mmcgee-jump I don't have bandwidth to test this at the moment. Feel free to hold for now

mmcgee-jump (approved):

PR #4343, opened 2025-02-28
removed rent collection and funky partitions

asiegel-jt:


ripatel-fd:
Some of those test ledgers probably still rely on rent collection and need to be regenerated.

jumpsiegel (dismissed):

jumpsiegel (approved):

jumpsiegel (approved):

PR #4361, opened 2025-03-03
shred: notify replay tile of completed shreds (do not merge)

emwang-jump:


src/app/fdctl/run/tiles/fd_shred.c
```diff
@@ -633,13 +640,22 @@ after_frag( fd_shred_ctx_t *    ctx,
...
-    fd_stem_publish( stem, 0UL, new_sig, fd_laddr_to_chunk( ctx->store_out_mem, s34+1UL ), sz1, 0UL, ctx->tsorig, tspub );
-  fd_stem_publish( stem, 0UL, new_sig, fd_laddr_to_chunk( ctx->store_out_mem, s34+2UL), sz2, 0UL, ctx->tsorig, tspub );
+    fd_stem_publish( stem, STORE_OUT_IDX, new_sig, fd_laddr_to_chunk( ctx->store_out_mem, s34+1UL ), sz1, 0UL, ctx->tsorig, tspub );
+  fd_stem_publish( stem, STORE_OUT_IDX, new_sig, fd_laddr_to_chunk( ctx->store_out_mem, s34+2UL), sz2, 0UL, ctx->tsorig, tspub );
   if( FD_UNLIKELY( s34[ 3 ].shred_cnt ) )
-    fd_stem_publish( stem, 0UL, new_sig, fd_laddr_to_chunk( ctx->store_out_mem, s34+3UL ), sz3, 0UL, ctx->tsorig, tspub );
-
+    fd_stem_publish( stem, STORE_OUT_IDX, new_sig, fd_laddr_to_chunk( ctx->store_out_mem, s34+3UL ), sz3, 0UL, ctx->tsorig, tspub );
+
+  /* Send to replay tile */
```
mmcgee-jump, line 650: I wouldn't add a 2nd publish here, which will break Frankendancer.

The publish is "either or" (either shred_store, or shred_replay), so you can just make sure in `unprivileged_init` you either have one `shred_store` or one `shred_replay`, and then set `STORE_OUT_IDX` (maybe rename `SHRED_OUT_IDX`?) to that link.

emwang-jump: Yep I have a change that conditions the second publish - but I think we want to keep both the shred_store and shred_replay link for firedancer for now to roll this change out slowly

lidatong: we can ifdef that link too

lidatong: actually we're changing the sig so dont think we can reuse the link like this @mmcgee-jump

mmcgee-jump: You don't want to use the ifdef, Richie is going to remove that pretty soon. The tile should work for either topology without recompilation.

mmcgee-jump: Can't they share a sig? The existing one just needs one bit for whether it's a leader shred or replay shred.

lidatong: well we need to figure out a sig that works for both then. it looks like all the FFI store needs is a single bit `sig` field

lidatong: > You don't want to use the ifdef, Richie is going to remove that pretty soon. 

why?

> The tile should work for either topology without recompilation.

i dont see how we can make this work with the shred tile

mmcgee-jump: As far as I can tell you could do this change in two lines of code by just creating a "shred_store" link in the fd_firedancer topology that goes to replay instead of store, and then updating the sig so it works for both consumers. Then it can be compiled once and linked by either library, too, and doesn't pollute the file with a bunch of ifdefs.

lidatong: yeah thats what i meant in my comment above about changing the sig of msgs on the link. but i dont see how we can modify the shred tile's stem publish or blcokstore_insert without an `ifdef`

src/disco/fd_disco_base.h
```diff
@@ -121,6 +121,29 @@ fd_disco_replay_sig( ulong slot,
 FD_FN_CONST static inline ulong fd_disco_replay_sig_flags( ulong sig ) { return (sig & 0xFFUL); }
 FD_FN_CONST static inline ulong fd_disco_replay_sig_slot( ulong sig ) { return (sig >> 8); }
 
+FD_FN_CONST static inline ulong
+fd_disco_shred_sig( ulong  slot,
```
lidatong, line 125: replay_sig

src/disco/fd_disco_base.h
```diff
@@ -121,6 +121,29 @@ fd_disco_replay_sig( ulong slot,
...
+                    ushort parent_off,
+                    uint   fec_idx,
+                    uchar  shred_flags,
+                    int    is_turbine ) {
+  /*
+     |  32 LSB of slot  | 8 LSB of parent_off | 15 LSB of fec_idx | 8 bits of data flags | 1 bit turbine/repair flag |
+     | slot[32:63]      | parent_off[24:31]   | shred_idx[9:23]     | shred_flags[8:1]     | is_turbine[0:0]           |
+   */
+  return ((slot & 0xFFFFFFFF) << 32 )
+         | ((ulong)(parent_off & 0xFF) << 24 )
```
lidatong, line 135: overflow check might be able to do something fancy with fd_bits

src/app/fdctl/run/tiles/fd_shred.c
mmcgee-jump: Well at a minimum it can be an if rather than an ifdef. The tiles are library code built in disco, so it doesn't make sense to push application knowledge (frankendancer vs full firedancer) into there.

Basically you should just build this as generic library functionality in the tile, probably I would add an argument to the tile topo or whatever `ulong blockstore_obj_id` and if this is not ULONG_MAX, you store shreds directly into there before broadcasting. Nothing else really needs to change except the signature?

PR #4394, opened 2025-03-05
Unify vm error with agave [wip]

yhzhangjump:


src/flamenco/vm/syscall/fd_vm_syscall_util.c
```diff
@@ -458,9 +458,7 @@ fd_vm_memmove( fd_vm_t * vm,
             - The next input memory region is writable
            Fail otherwise. */
         if( FD_LIKELY( !reverse &&
-                        dst_is_input_mem_region &&
-                        dst_region_idx+1UL<vm->input_mem_regions_cnt &&
-                        vm->input_mem_regions[ dst_region_idx+1UL ].is_writable ) ) {
+                        dst_region_idx+1UL<vm->input_mem_regions_cnt ) ) {
```
topointon-jump, line 461: Are you sure this logic is correct? This logic is important for ensuring that writes don't span more than a single account region.

topointon-jump (changes requested):
I'm not sure that this logic is correct - do you mind explaining why you're changing this, adding a link to the corresponding Agave code and adding the fixture to CI?

src/flamenco/vm/syscall/fd_vm_syscall_util.c
yhzhangjump: The `dst_bytes_rem_in_cur_region==0UL` check before this modification is ensuring "writes don't span more than a single account region", as you suggested.
The 2 specific checks involved here are about whether the region is (1) writable or (2) being the input region (instead of heap, etc.).
It seems to me that these 2 checks are missing in Agave, leading to different VM return error.

yhzhangjump:
The issue is correctly fixed in PR #4418

PR #4403, opened 2025-03-05
gui: only update peers on the leader schedule

jherrera-jump:
On "slow 4g", cuts load times from ~1min to ~30sec

### DIFF
--- a/book/api/websocket.md
+++ b/book/api/websocket.md
@@ -782,13 +782,15 @@ completed, rather than rooted, but no speculative epoch information is
 published until the epoch is finalized by rooting the slot.
 
 ### peers
-Information about validator peers from the cluster. Peer data is sourced
-from gossip, the accounts database, and the on-chain configuration
-program. All peer information is authenticated meaning it can only be
-reported from the holder of the private key, however not all peer data
-is validated or checked for correctness. In particular, data from the
-gossip network and the config program is self reported by the validator
-and could be empty, corrupt, filled with garbage, or malicious.
+Information about validator peers from the cluster. Only validator peers
+with leader slots on for the current or subsequent epoch will sent on
+on this message channel. Peer data is sourced from gossip, the accounts
+database, and the on-chain configuration program. All peer information
+is authenticated meaning it can only be reported from the holder of the
+private key, however not all peer data is validated or checked for
+correctness. In particular, data from the gossip network and the config
+program is self reported by the validator and could be empty, corrupt,
+filled with garbage, or malicious.
 
 Peer information is keyed by the validator identity key. Multiple vote
 accounts could in theory use the same identity keypair, although it is
--- a/src/disco/gui/fd_gui.c
+++ b/src/disco/gui/fd_gui.c
@@ -8,6 +8,77 @@
 #include "../../ballet/json/cJSON.h"
 #include "../../flamenco/genesis/fd_genesis_cluster.h"
 
+#define SORT_NAME        fd_sort_gui_gossip_peer
+#define SORT_KEY_T       fd_gui_gossip_peer_t
+#define SORT_BEFORE(a,b) memcmp(a.pubkey, b.pubkey, 32UL)<0
+#include "../../util/tmpl/fd_sort.c"
+
+#define SORT_NAME        fd_sort_gui_vote_account
+#define SORT_KEY_T       fd_gui_vote_account_t
+#define SORT_BEFORE(a,b) memcmp(a.vote_account, b.vote_account, 32UL)<0
+#include "../../util/tmpl/fd_sort.c"
+
+#define SORT_NAME        fd_sort_gui_validator_info
+#define SORT_KEY_T       fd_gui_validator_info_t
+#define SORT_BEFORE(a,b) memcmp(a.pubkey, b.pubkey, 32UL)<0
+#include "../../util/tmpl/fd_sort.c"
+
+#define SORT_NAME        fd_sort_gui_lsched
+#define SORT_KEY_T       fd_pubkey_t
+#define SORT_BEFORE(a,b) memcmp(&a, &b, 32UL)<0
+#include "../../util/tmpl/fd_sort.c"
+
+ulong
+fd_gui_gossip_contains( fd_gui_t * gui,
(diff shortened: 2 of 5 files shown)


src/disco/gui/fd_gui_printf.c
```diff
@@ -809,6 +822,7 @@ fd_gui_printf_peers_gossip_update( fd_gui_t *          gui,
     jsonp_open_object( gui, "value" );
       jsonp_open_array( gui, "add" );
         for( ulong i=0UL; i<added_cnt; i++ ) {
+          if( FD_LIKELY( !fd_gui_leader_schedule_contains( gui, gui->gossip.peers[ added[ i ] ].pubkey->uc ) ) ) continue;
```
mmcgee-jump, line 825: I don't think this is quite right, because a node might enter the leader schedule dynamically, in which case you need to publish an "add".

jherrera-jump: oh yea good catch. I'll have to broadcast a bunch of adds in `fd_gui_handle_leader_schedule` at epoch boundaries.

src/disco/gui/fd_gui.c
```diff
@@ -946,6 +959,28 @@ fd_gui_handle_leader_schedule( fd_gui_t *    gui,
   ulong slot_cnt            = msg[ 3 ];
   ulong excluded_stake      = msg[ 4 ];
 
+  fd_stake_weight_t const * stake = (fd_stake_weight_t const *)fd_type_pun_const( msg+5UL );
+
+  for ( ulong i=0UL; i<staked_cnt; i++ ) {
+    if( FD_UNLIKELY( ULONG_MAX==fd_gui_leader_schedule_contains( gui, stake[ i ].key.uc, 0 ) ) ) {
```
mmcgee-jump, line 965: Still don't think this is right, because the leader might have already been on the prior schedule?

src/disco/gui/fd_gui_printf.c
```diff
@@ -819,19 +803,22 @@ fd_gui_printf_peers_gossip_update( fd_gui_t *          gui,
 
       jsonp_open_array( gui, "update" );
         for( ulong i=0UL; i<added_cnt; i++ ) {
+          if( FD_LIKELY( ULONG_MAX==fd_gui_cur_and_next_lsched_contains( gui, gui->gossip.peers[ added[ i ] ].pubkey->uc ) ) ) continue;
```
mmcgee-jump, line 806: This is gonna be pretty slow, I think? 864000*1500 memcmps

jherrera-jump: Switched lookups to use binary search

src/disco/gui/fd_gui.c
```diff
@@ -8,6 +8,104 @@
...
+
+   Peer arrays must be sorted any time they are updated, otherwise these
+   functions may return incorrect results. */
+
+static ulong
+fd_gui_gossip_contains( fd_gui_t *    gui,
+                        uchar const * pubkey ) {
+  fd_gui_gossip_peer_t query;
+  fd_memcpy( query.pubkey, pubkey, 32UL );
+  ulong pubkey_idx = fd_sort_gui_gossip_peer_search_geq( gui->gossip.peers, gui->gossip.peer_cnt, query );
```
mmcgee-jump, line 44: I'm pretty sure this `search_geq` function is bugged. E.g. it fails

```  do {
    float sorted[2];
    sorted[0] = 9.0;
    sorted[1] = 15.0;
    FD_TEST( sort_up_search_geq( sorted, 2, 20.0 )==2UL );
  } while(0);```

Comment (jherrera-jump) [reply src/disco/gui/fd_gui.c:44]: I think it just has the slightly unconvential behavior of returning the 0th index when the query is less than all elements in the array and returning the last index (cnt-1) when query is greater than all elements in the array.

The the second bit was missing from the [spec](https://github.com/firedancer-io/firedancer/blob/dcce6d9e6e72edc7ccabe1b7d5c167134aca462e/src/util/tmpl/fd_sort.c#L71) (I added it), but we did [test](https://github.com/firedancer-io/firedancer/blob/dcce6d9e6e72edc7ccabe1b7d5c167134aca462e/src/util/tmpl/test_sort.c#L211) for it.

I think in these cases it shouldn't matter then since I'm checking if the index returned is a match or not right after.

File: book/api/websocket.md [hunk 5]
```diff
@@ -908,6 +910,26 @@ full and includes this node itself, nodes with a different
 `shred_version`, nodes publishing corrupt or bad information, and so
 on.
 
+#### `peers.rpc_count`
```
mmcgee-jump, line 913: Can you describe what an `rpc` peer is defined as? It should be any unstaked node that's listening to gossip.

src/disco/gui/fd_gui.c
```diff
@@ -124,7 +223,7 @@ fd_gui_join( void * shmem ) {
 void
 fd_gui_set_identity( fd_gui_t *    gui,
```
mmcgee-jump, line 224: You might need to publish the peer now, right? If it's "added" ?

src/disco/gui/fd_gui.c
```diff
@@ -177,9 +286,34 @@ fd_gui_ws_open( fd_gui_t * gui,
     }
   }
 
-  /* Print peers last because it's the largest message and would
-     block other information. */
-  fd_gui_printf_peers_all( gui );
+  ulong action[ 40200 ];
+  fd_pubkey_t peers[ 40200 ];
```
mmcgee-jump, line 290: I think the peers array here has to be 120601 to prevent corruption? Because the staked nodes could all complete change each of the 3 epochs (plus one for this node?)

mmcgee-jump: Actually should be 150001 for 50,000 staked in each epoch?

src/disco/gui/fd_gui.c
mmcgee-jump: I think you also need corresponding logic to "remove" the peer if the identity key changes.

mmcgee-jump (changes requested):
.

PR #4409, opened 2025-03-06
shred: increae link size to store to prevent backpressure

mmcgee-jump:


ripatel-fd:
Doesn't seem like we need this since we've been running for 3 months without this fix

PR #4411, opened 2025-03-06
Flamenco, Gossip: epoch slot fixes

ravyu-jump:
Couple of independent fixes to improve Gossip performance

- [flamenco, gossip: bump CRDS value expiration time](https://github.com/firedancer-io/firedancer/commit/e25aea7d32c756d6e5767c0d118722e05b2e1abc)
  - CRDS values were initially discarded after 1 min. While this was fine for regularly updated entries (vote, contact info), epoch slots suffered as they aren't updated often (we see roughly 500 new entries in testnet per minute). This meant the table discarded useful epoch slot values too early, and therefore each pull request constructed bloom filters without these entries, causing a replenish spam from other validators. By extending the expiry window from 1 minute to 1 hour (Agave's equivalent is roughly 1 epoch, which we haven't tested), we hold on to these epoch slot entries longer and therefore reduce replenishes from pull requests.
  - Worth noting that we increase the expiry window for ALL CRDS values, which might be excessive. But given that Agave holds everything for much longer, this probably works fine for us. Running a node for longer than the expiry window (1hr) showed no table full errors either. 
  - We should explore holding older entries in a smarter manner. Do we really want to hold a full value for that long? Agave has a concept of a "purge" table that only stores the hashes of the CRDS value. We might want to adopt something like this if memory usage becomes an issue. 
- [flamenco, gossip: skip sigverify on epoch slot msgs](https://github.com/firedancer-io/firedancer/commit/e310969f86d43c29637a6a11f4a23c5ca57c86ab) skips verifying epoch slots, which significantly helps with the gossip message burst in start-up.

### DIFF
--- a/src/flamenco/gossip/fd_gossip.c
+++ b/src/flamenco/gossip/fd_gossip.c
@@ -1445,8 +1445,16 @@ fd_gossip_recv_crds_array( fd_gossip_t * glob, const fd_gossip_peer_addr_t * fro
       }
 
       glob->recv_nondup_cnt++;
-      /* Sigverify */
-      if( fd_crds_sigverify( tmp->encoded_val, tmp->encoded_len, tmp->pubkey ) ) {
+      /* Sigverify step
+         Skip verifying epoch slots because they:
+          - are not used anywhere within the client
+            - still store them in table for forwarding +
+              prune message construction
+          - represent a significant portion of inbound CRDS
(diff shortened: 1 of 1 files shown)


topointon-jump (approved):
🔥

ravyu-jump:
Need to update comments with documentation, will do so in a separate PR.

PR #4415, opened 2025-03-06
codeql, chore: detect used parameters marked unused

two-heart:
CodeQL query to detect such parameters + fix current offenses.
Afaik gcc/clang don't optimize based on parameters marked as unused, so this is not a bug fix, but just for the hygine of the codebase. We want to avoid a user of an API thinking: "oh, I can pass a wrong value for this argument, because it is unused anyway.", but it actually isn't.


ravyu-jump (approved):
this is great, thank you!

PR #4456, opened 2025-03-10
Explicitly make stack non-executable in asm files

ripatel-fd:


two-heart (approved):
Tank you for finding and fixing

PR #4466, opened 2025-03-11
fdctl: improve layering

mmcgee-jump:


ripatel-fd:
This looks fun to review

ripatel-fd:
Please update doc/organization.txt

lidatong:
why separate discof / discoh? isnt the idea it's just one lib and the topo in `app` decides what's the app specific?

PR #4474, opened 2025-03-11
Fix full Firedancer tile UDP logic

ripatel-fd:


src/util/net/fd_net_headers.h
```diff
@@ -5,11 +5,14 @@
 #include "fd_eth.h"
 
 typedef union {
-  struct __attribute__((packed)) {
+  struct {
```
jumpsiegel, line 8: why did we remove this packed?  Doesn't that change the semantics ?

ripatel-fd:
@jumpsiegel Sorry this PR is junk. I'll reopen a proper fix with explanations.

PR #4483, opened 2025-03-11
txncache: improvements

yufeng-jump:
- Support worst case transaction mix under mainnet conditions, including nonce transactions.
- Implement race free reprobing.
- Adjust struct layout for better cache line behavior: up to 43% speedup for queries.
- Make the size of txncache configurable.
- Add support for multiple key sizes.
- Remove support for serializing the txncache into a snapshot.
- Better support for key slicing after loading from an Agave snapshot.
- Remove redundant read lock acquisition in executor root slot check.
- More tests, including one that tries to simulate production op mixes.
- Add metrics.

### DIFF
--- a/src/flamenco/runtime/fd_executor.c
+++ b/src/flamenco/runtime/fd_executor.c
@@ -304,12 +304,20 @@ fd_validate_fee_payer( fd_txn_account_t * account,
 
 static int
 status_check_tower( ulong slot, void * _ctx ) {
+  /* TODO remove these after we've figured out the sensible order of these checks. */
+  static ulong check_tower_same_slot = 0UL;
+  static ulong check_tower_rooted    = 0UL;
+  static ulong check_tower_history   = 0UL;
+
   fd_exec_txn_ctx_t * ctx = (fd_exec_txn_ctx_t *)_ctx;
+
   if( slot==ctx->slot ) {
+    FD_ATOMIC_FETCH_AND_ADD( &check_tower_same_slot, 1UL );
     return 1;
   }
 
-  if( fd_txncache_is_rooted_slot( ctx->status_cache, slot ) ) {
+  if( fd_txncache_is_rooted_slot_locked( ctx->txncache, slot ) ) {
+    FD_ATOMIC_FETCH_AND_ADD( &check_tower_rooted, 1UL );
     return 1;
   }
 
@@ -320,9 +328,16 @@ status_check_tower( ulong slot, void * _ctx ) {
     FD_LOG_ERR(( "Unable to read and decode slot history sysvar" ));
   }
 
+  /* Agave checks against ancestors, we check against slot_history.  On
+     the Agave side, both ancestors and slot_history are updated during
+     new_from_parent().  The two checks should be equivalent, as long as
+     our status cache stores the same set of transactions as Agave, and
+     unrooted slots in the status cache are present in both ancestors
+     and slot_history. */
   if( fd_sysvar_slot_history_find_slot( slot_history,
                                         slot,
                                         ctx->runtime_pub_wksp ) == FD_SLOT_HISTORY_SLOT_FOUND ) {
+    FD_ATOMIC_FETCH_AND_ADD( &check_tower_history, 1UL );
     return 1;
   }
 
@@ -330,9 +345,9 @@ status_check_tower( ulong slot, void * _ctx ) {
 }
 
 static int
-fd_executor_check_status_cache( fd_exec_txn_ctx_t * txn_ctx ) {
+fd_executor_check_txncache( fd_exec_txn_ctx_t * txn_ctx ) {
 
-  if( FD_UNLIKELY( !txn_ctx->status_cache ) ) {
+  if( FD_UNLIKELY( !txn_ctx->txncache ) ) {
     return FD_RUNTIME_EXECUTE_SUCCESS;
   }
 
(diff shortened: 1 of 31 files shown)


mmcgee-jump (changes requested):
Just flagging for review, will take a look soon

src/flamenco/runtime/fd_txncache.c
```diff
@@ -983,19 +1129,28 @@ fd_txncache_query_batch( fd_txncache_t *             tc,
     }
 
     ulong txnhash_offset = blockcache->txnhash_offset;
-    ulong head_hash = FD_LOAD( ulong, query->txnhash+txnhash_offset ) % FD_TXNCACHE_BLOCKCACHE_MAP_CNT;
-    for( uint head=blockcache->heads[ head_hash ]; head!=UINT_MAX; head=txnpages[ head/FD_TXNCACHE_TXNS_PER_PAGE ].txns[ head%FD_TXNCACHE_TXNS_PER_PAGE ]->blockcache_next ) {
+    if( FD_UNLIKELY( txnhash_offset) ) {
+      ulong max_key_idx = fd_ulong_sat_sub( query->key_sz, FD_TXNCACHE_KEY_SIZE+1UL );
```
cali-jumptrading, line 1133: maybe some sanity checks on key_sz. Or... if key size is inherently a property of the `txncache`, then maybe it doesn't make sense for the user to supply the key size when calling `txncache` APIs. Rather, the user could supply the key sizes for txn hash and block hash on `txncache` creation.

cali-jumptrading (commented):
Was mostly reviewing this PR to learn more about the status cache

src/flamenco/runtime/fd_executor.c
```diff
@@ -259,20 +259,38 @@ fd_validate_fee_payer( fd_txn_account_t * account,
 
 static int
 status_check_tower( ulong slot, void * _ctx ) {
```
topointon-jump, line 261: Note for future follow-up - after the runtime re-architecture, it might be worth exploring if we can make the status cache a serial data structure which is only accessed from a single tile, and remove all these locks and atomic operations.

src/flamenco/runtime/fd_txncache.c
```diff
@@ -28,44 +27,62 @@
 
 #define FD_TXNCACHE_SLOTCACHE_MAP_CNT (1024UL)
 
-/* Value for an empty blockcache `max_slot` or empty slotcache
-  `slot` entry. When the entries are set to this value, we can insert
-  to the entry, but stop iterating while running queries. */
+/* The number of unique entries in the hash lookup table for each
+   (slot) in the slotcache.
+   FIXME This is going to fill up fast if there are lots of nonces in a slot
```
topointon-jump, line 32: We need to address this - can do this in a follow-up PR

src/flamenco/runtime/fd_txncache.c
```diff
@@ -969,33 +1127,50 @@ fd_txncache_query_batch( fd_txncache_t *             tc,
...
     }
   }
 
   fd_rwlock_unread( tc->lock );
+
+  FD_ATOMIC_FETCH_AND_ADD( &tc->query_present, query_present );
+  FD_ATOMIC_FETCH_AND_ADD( &tc->query_absent,  query_absent  );
 }
 
 int
```
topointon-jump, line 1176: We should delete `fd_txncache_snapshot` and all the accompanying code/data structures, this is never used and `fd_txncache_get_entries` is used instead.

src/flamenco/runtime/fd_txncache.c
```diff
@@ -469,38 +500,122 @@ fd_txncache_delete( void * shtc ) {
...
+                    *(txn->result) ));
+  } else {
+    FD_LOG_NOTICE(( "%s: blockhash %s, slot %lu, key %s, res %u",
+                    prefix,
+                    blockhash,
+                    txn->slot,
+                    "(unsupported key length)",
+                    *(txn->result) ));
   }
+}
```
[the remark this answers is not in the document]
yufeng-jump: This is basically an irrecoverable error.  The validator is gonna crash if it can't insert.  So I think NOTICE is fine here.

src/flamenco/runtime/fd_txncache.c
```diff
@@ -553,20 +672,32 @@ static void
...
+    if( FD_UNLIKELY( root_slots[ idx ]==slot ) ) return; /* Slot already registered. */
+    if( FD_UNLIKELY( root_slots[ idx ]>slot ) ) break;   /* Would like to insert at idx if pushing right, and idx-1 if evicting left. */
   }
 
   if( FD_UNLIKELY( tc->root_slots_cnt>=tc->root_slots_max ) ) {
     if( FD_LIKELY( idx ) ) {
+      long start = fd_log_wallclock();
       fd_txncache_purge_slot( tc, root_slots[ 0 ] );
+      long end = fd_log_wallclock();
+      FD_LOG_INFO(( "registering slot %lu at idx %lu triggered a purge of %ld nanos", slot, idx, end-start ));
```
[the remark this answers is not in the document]
yufeng-jump: Purging happens very infrequently.  We do wallclock calls even in pack, and they aren't all that expensive at the frequency we call it.  I can see the argument for doing a metrics histogram rather than a log info here.  That cleanup can be a followup PR.

src/flamenco/runtime/fd_txncache.c
yufeng-jump: Deleted.

PR #4486, opened 2025-03-11
flamenco: Vote lockout should return maximum lockout when overflow is detected

cali-jumptrading:
Although the confirmation count should never be greater than 32, we should return the maximum lockout rather than 0 when the confirmation count is greater than 32. This allows votes to be properly expired.


ibhatt-jumptrading (dismissed):

mjain-jump (approved):
great ty

PR #4488, opened 2025-03-12
Add 'fddev gossip' command

ripatel-fd:
- Adds a standalone gossip run mode using the production gossip tile
- Remove 'fddev spy' (broken)
- Error fd_keyload_load if called with an empty path
- Allow links without a producer or a consumer
- Fix IPv4 options handling in eqvoc tile
- Prevent startup if gossip tile is configured with invalid IP/port
- Error fd_xdp_install if called without listen ports


src/app/fddev/commands/gossip_standalone.c
```diff
@@ -0,0 +1,198 @@
...
+  fd_topob_wksp( topo, "sign_gossip"  );
+  fd_topob_link( topo, "sign_gossip", "sign_gossip", 128UL,   64UL, 1UL );
+  fd_topob_tile_out( topo, "sign", 0UL, "sign_gossip", 0UL );
+
+  fd_topob_wksp( topo, "gossip_net" );
+  fd_topob_link( topo, "voter_gossip", "gossip", 128UL, 1UL, 1UL );
+  fd_topob_link( topo, "rstart_gossi", "gossip", 128UL, 1UL, 1UL );
+  fd_topob_link( topo, "gossip_net", "gossip_net", config->tiles.net.send_buffer_size, FD_NET_MTU, 1UL );
+  fd_topob_link( topo, "crds_shred", "gossip", 128UL, 8UL + 40200UL * 38UL, 1UL );
+  fd_topob_link( topo, "gossip_repai", "gossip", 128UL, 40200UL * 38UL, 1UL );
```
mmcgee-jump, line 68: What do you need all these links for?

mmcgee-jump: All of the links are optional ... you just need to update the tile to not expect them. You can see. e.g. fd_poh or other tiles for an example of how this is done.

mmcgee-jump (changes requested):
.

ravyu-jump (dismissed):
LGTM, just a cosmetic nit

src/app/fddev/commands/gossip_standalone.c
```diff
@@ -0,0 +1,153 @@
...
+  (void)pargc; (void)pargv; (void)args;
+}
+
+void
+gossip_cmd_fn( args_t *   args,
+               config_t * config ) {
+  gossip_topo( config );
+  fd_topo_t * topo = &config->topo;
+
+  if( FD_LIKELY( !args->dev.no_configure ) ) {
```
mmcgee-jump, line 118: Wrong struct? Should be `args->gossip.no_confiure`

src/app/fddev/commands/gossip_standalone.c
```diff
@@ -0,0 +1,153 @@
...
+    configure_stage( &fd_cfg_stage_ethtool_gro,      CONFIGURE_CMD_INIT, config );
+  }
+
+  initialize_workspaces( config );
+  initialize_stacks( config );
+  if( 0==strcmp( config->development.net.provider, "xdp" ) ) {
+    fd_topo_install_xdp( topo );
+  }
+  fd_topo_join_workspaces( topo, FD_SHMEM_JOIN_MODE_READ_WRITE );
+  fd_topo_run_single_process( topo, 2, config->uid, config->gid, fdctl_tile_run, NULL );
```
mmcgee-jump, line 131: Can you add a TODO to run this sandboxed / multiprocess.

src/app/fddev/commands/gossip_standalone.c
```diff
@@ -0,0 +1,153 @@
...
+
+  for(;;) pause();
+}
+
+static void
+configure_stage_perm( configure_stage_t const * stage,
+                      fd_cap_chk_t *            chk,
+                      config_t const *          config ) {
+  int enabled = !stage->enabled || stage->enabled( config );
+  if( enabled && stage->check( config ).result != CONFIGURE_OK )
```
mmcgee-jump, line 141: This isn't quite right. The command might have to fini first, which will require `fini_perm`. I think if result is PARTIALLY_CONFIGURED it will fini, before doing init.

mmcgee-jump (changes requested):
nice!

ravyu-jump (approved):

PR #4503, opened 2025-03-12
fuzz, flamenco: block-level dumping + replay (prep for fuzzing)

mjain-jump:


jumpsiegel (dismissed):

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1321,22 +1330,6 @@ fd_runtime_block_execute_prepare( fd_exec_slot_ctx_t * slot_ctx,
...
-  }
-
-  /* Load sysvars into cache */
-  if( FD_UNLIKELY( result = fd_runtime_sysvar_cache_load( slot_ctx, runtime_spad ) ) ) {
-    /* non-zero error */
-    return result;
-  }
-
-  return FD_RUNTIME_EXECUTE_SUCCESS;
 }
```
topointon-jump, line 1334: Need to make sure this doesn't break the live client, as `fd_runtime_block_execute_prepare` is used by the replay tile.

@kbhargava-jump is removing the ledger tool replay in favour of shredcap->replay tile, but for the meantime we have to be careful we don't break the replay tile.

src/flamenco/runtime/fd_runtime.c
```diff
@@ -3836,16 +3837,13 @@ fd_runtime_block_execute_tpool( fd_exec_slot_ctx_t *    slot_ctx,
...
-int
+void
 fd_runtime_block_pre_execute_process_new_epoch( fd_exec_slot_ctx_t * slot_ctx,
                                                 fd_tpool_t *         tpool,
                                                 fd_spad_t * *        exec_spads,
                                                 ulong                exec_spad_cnt,
                                                 fd_spad_t *          runtime_spad ) {
 
-  /* Update block height. */
-  slot_ctx->slot_bank.block_height += 1UL;
```
topointon-jump, line 3847: We need to make sure this doesn't break any callers of `fd_runtime_block_pre_execute_process_new_epoch`

topointon-jump (approved):

PR #4509, opened 2025-03-13
pack: fix treap and prq sizes

two-heart:
Previously, the underlying pool size for the treaps and the priority queue was larger than the max passed to these data structures at initialization. This leads to out of bounds accesses (treap/prq contract violation).
The problem can be observed by checking if the treap is still valid after inserting a transaction and then running pack unit tests.
```diff
@@ -1288,6 +1288,7 @@ fd_pack_insert_txn_fini( fd_pack_t  * pack,
     return replaces ? FD_PACK_INSERT_ACCEPT_VOTE_REPLACE : FD_PACK_INSERT_ACCEPT_VOTE_ADD;
   } else {
     treap_ele_insert( insert_into,         ord, pack->pool );
+    FD_TEST(treap_verify( insert_into, pack->pool )!= -1);
     return replaces ? FD_PACK_INSERT_ACCEPT_NONVOTE_REPLACE : FD_PACK_INSERT_ACCEPT_NONVOTE_ADD;
   }
 }

```

Drive-by: typo fixes

### DIFF
--- a/src/disco/pack/fd_pack.c
+++ b/src/disco/pack/fd_pack.c
@@ -266,7 +266,7 @@ typedef struct fd_pack_bitset_acct_mapping fd_pack_bitset_acct_mapping_t;
      could be because of a bug in the code that generated the
      initializer bundle, a lack of fee payer balance, or an expired
      blockhash.
-   * READY: the most recently scheduled initialzation bundle succeeded
+   * READY: the most recently scheduled initialization bundle succeeded
      and normal bundles can be scheduled in this slot. */
 #define FD_PACK_IB_STATE_NOT_INITIALIZED 0
 #define FD_PACK_IB_STATE_PENDING         1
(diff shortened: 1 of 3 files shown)


src/disco/pack/fd_pack.c
```diff
@@ -670,7 +670,7 @@ fd_pack_footprint( ulong                    pack_depth,
   l = FD_LAYOUT_APPEND( l, FD_PACK_ALIGN,       sizeof(fd_pack_t)                               );
   l = FD_LAYOUT_APPEND( l, trp_pool_align (),   trp_pool_footprint ( pack_depth+extra_depth   ) ); /* pool           */
   l = FD_LAYOUT_APPEND( l, penalty_map_align(), penalty_map_footprint( lg_penalty_trp         ) ); /* penalty_treaps */
-  l = FD_LAYOUT_APPEND( l, expq_align     (),   expq_footprint     ( pack_depth               ) ); /* expiration prq */
+  l = FD_LAYOUT_APPEND( l, expq_align     (),   expq_footprint     ( pack_depth+extra_depth   ) ); /* expiration prq */
```
ptaffet-jump, line 673: The original value of this is correct

ptaffet-jump (changes requested):
You are correct that `ele_max` for a treap is documented as the max index, not the maximum number of elements that can be in the treap at a given time. That is not true for `fd_prq` though. Thanks for all the typo fixes.

src/disco/pack/fd_pack.c
```diff
@@ -754,9 +754,9 @@ fd_pack_new( void                   * mem,
 
   penalty_map_new( _penalty_map, lg_penalty_trp );
 
-  treap_new( (void*)pack->pending,         pack_depth );
-  treap_new( (void*)pack->pending_votes,   pack_depth );
-  treap_new( (void*)pack->pending_bundles, pack_depth );
+  treap_new( (void*)pack->pending,         pack_depth+extra_depth );
```
ptaffet-jump, line 757:
```suggestion
  /* These treaps can have at most pack_depth elements at any moment, but they come from a pool of size pack_depth+extra_depth. */
  treap_new( (void*)pack->pending,         pack_depth+extra_depth );
```

ptaffet-jump (approved):

PR #4523, opened 2025-03-13
Produce chained merkle shreds in Frankendancer

0x0ece:
This adds chained merkle shreds to Frankendancer (but not Firedancer yet), with tests that we can successfully decode chained merkle shreds that we produce.

After this change, Frankendancer will only send chained merkle shreds.


src/disco/shred/fd_shredder.c
```diff
@@ -144,38 +148,60 @@ fd_shredder_next_fec_set( fd_shredder_t * shredder,
 
   if( FD_UNLIKELY( (offset==entry_sz) ) ) return NULL;
 
+  /* Set the shred type */
+
+  int   block_complete          = shredder->meta.block_complete;
+  uchar is_chained              = chained_merkle_root!=NULL;
```
ptaffet-jump, line 154: booleans like these are typically ints. If the compiler complains about ulong + ulong*int though and not about ulong+ulong*uchar, I'm okay with leaving them as is

src/disco/shred/fd_shredder.c
```diff
@@ -248,20 +284,30 @@ fd_shredder_next_fec_set( fd_shredder_t * shredder,
 
     uchar * merkle = data_shreds[ i ] + fd_shred_merkle_off( shred );
     fd_bmtree_get_proof( bmtree, merkle, i );
+
+    if( FD_UNLIKELY( is_resigned ) ) {
```
ptaffet-jump, line 288: The correct behavior is for the leader to zero out the resigned field and not to copy the shred signature? You should add a comment to that effect.

src/disco/shred/fd_shredder.h
```diff
@@ -109,38 +112,99 @@ void *          fd_shredder_delete( void *          mem      );
...
+   -                   payload_sz:  859             (merkle tree height: 8)
+   note: payload_sz is the chained payload_sz - 64 bytes (for signature). */
+
+#define FD_SHREDDER_NORMAL_FEC_SET_PAYLOAD_SZ   (31840UL)
+#define FD_SHREDDER_CHAINED_FEC_SET_PAYLOAD_SZ  (30816UL) /* -32 bytes * 32 shreds */
+#define FD_SHREDDER_RESIGNED_FEC_SET_PAYLOAD_SZ (28768UL) /* -64 bytes * 32 shreds */
 
 FD_FN_CONST static inline ulong
-fd_shredder_count_fec_sets(      ulong sz_bytes ) {
+fd_shredder_count_fec_sets(      ulong sz_bytes, ulong type ) {
```
ptaffet-jump, line 148: `type` should be described somewhere in the comment

src/disco/shred/fd_shredder.h
```diff
@@ -109,38 +112,99 @@ void *          fd_shredder_delete( void *          mem      );
...
   /* if sz_bytes < 2*31840, we make 1 FEC set.  If sz_bytes is a
      multiple of 31840, we make exactly sz_bytes/31840 sets.  Otherwise,
      we make floor(sz_bytes/31840)-1 normal set + one odd-sized set.
      These cases can be simplified to make it branchless: */
```
ptaffet-jump, line 152: It's definitely not branchless anymore :(

src/disco/shred/test_shredder.c
```diff
@@ -307,22 +308,276 @@ test_shredder_count( void ) {
...
 
-    FD_TEST( data_shred_cnt  ==fd_shredder_count_data_shreds  ( sz ) );
-    FD_TEST( parity_shred_cnt==fd_shredder_count_parity_shreds( sz ) );
+    FD_TEST( data_shred_cnt  ==fd_shredder_count_data_shreds  ( sz, FD_SHRED_TYPE_MERKLE_DATA ) );
+    FD_TEST( parity_shred_cnt==fd_shredder_count_parity_shreds( sz, FD_SHRED_TYPE_MERKLE_DATA ) );
   }
 }
 
+static void
+test_shredder_count_chained( void ) {
```
ptaffet-jump, line 328: There's probably a way to eliminate all the duplicate code between the `test_shredder_cnt*` functions if you'd like

src/discoh/poh/fd_poh_tile.c
```diff
@@ -1224,6 +1230,13 @@ fd_ext_poh_reset( ulong         completed_bank_slot, /* The slot that successful
 
   memcpy( ctx->reset_hash, reset_blockhash, 32UL );
   memcpy( ctx->hash, reset_blockhash, 32UL );
+  if( FD_LIKELY( parent_block_id!=NULL ) ) {
+    FD_LOG_INFO(( "fd_ext_poh_reset(block_id!=null,reset_slot=%lu)", completed_bank_slot ));
```
ptaffet-jump, line 1234: I think there's already a log line printed on reset, though it might be the other reset. This looks to me like debug logging that doesn't belong

src/discoh/poh/fd_poh_tile.c
```diff
@@ -1347,6 +1360,14 @@ publish_tick( fd_poh_ctx_t *      ctx,
   ulong slot = fd_ulong_if( meta->block_complete, ctx->slot-1UL, ctx->slot );
   meta->parent_offset = 1UL+slot-ctx->reset_slot;
 
+  meta->parent_block_id_valid = ctx->parent_slot == (slot-meta->parent_offset);
+  if( FD_LIKELY( meta->parent_block_id_valid ) ) {
+    FD_LOG_INFO(( "sending tick slot=%lu parent=%lu block_id!=null", slot, slot-meta->parent_offset ));
```
ptaffet-jump, line 1365: Same with this

src/discoh/poh/fd_poh_tile.c
```diff
@@ -1347,6 +1360,14 @@ publish_tick( fd_poh_ctx_t *      ctx,
   ulong slot = fd_ulong_if( meta->block_complete, ctx->slot-1UL, ctx->slot );
   meta->parent_offset = 1UL+slot-ctx->reset_slot;
 
+  meta->parent_block_id_valid = ctx->parent_slot == (slot-meta->parent_offset);
```
ptaffet-jump, line 1363: Could you add a comment explaining this line?

ptaffet-jump (commented):
Sorry it took me so long to get this reviewed. A few small things. One note about the submodule: except for changes to files that are entirely new in our patchset, we try to add a `// FIREDANCER: ` comment before each change with explanation of why it is needed.

src/disco/shred/fd_shred_tile.c
```diff
@@ -208,6 +224,8 @@ typedef struct {
       uchar raw[ 63679UL ]; /* The largest that fits in 1 FEC set */
```
ptaffet-jump, line 224: I think this needs to be updated as well, right? And the watermark variable.

0x0ece: It should eventually be decreased, right? So, I don't believe we can change this now.
Chained will be used in Frankendancer but not in Firedancer yet, so we still have to support the older Merkle shreds. (But maybe I have it the other way around.)

For the watermark I agree we could tune it, tbh I don't think it's currently tuned well on the mainnet traffic. Maybe we should tune it as we do the fixed 64 shreds?

src/disco/shred/fd_shred_tile.c
```diff
@@ -408,14 +441,22 @@ during_frag( fd_shred_ctx_t * ctx,
...
-      if( FD_UNLIKELY( ctx->batch_cnt%ctx->round_robin_cnt==ctx->round_robin_id ) ) {
+      if( SHOULD_PROCESS_THESE_SHREDS ) {
         /* If it's our turn, shred this batch. FD_UNLIKELY because shred tile cnt generally >= 2 */
         ulong batch_sz = sizeof(ulong)+ctx->pending_batch.pos;
 
         /* We sized this so it fits in one FEC set */
         long shredding_timing =  -fd_tickcount();
 
-        if( FD_UNLIKELY( entry_meta->block_complete && batch_sz < FD_SHREDDER_NORMAL_FEC_SET_PAYLOAD_SZ ) ) {
+#if FD_HAS_NO_AGAVE
```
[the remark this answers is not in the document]
0x0ece: I've been recommended to use some runtime checks based on the topology, e.g. if ctx->blockstore that's used elsewhere. I made a macro for this. I didn't propagate it outside my changes because I'd like to remove the differences asap.

src/disco/shred/fd_shred_tile.c
jvarela-jump: The raw buffer size has been kept as is (which corresponds to the max buffer size for normal fec_sets).
The watermark, however, has been adjusted to follow (chained+)resigned fec_sets, in order to comply the current logic around "last_in_batch".

src/disco/shred/fd_shredder.h
```diff
@@ -92,7 +90,8 @@ void *          fd_shredder_delete( void *          mem      );
 
 /* fd_shredder_count_{data_shreds, parity_shreds, fec_sets}: returns the
    number of data shreds, parity shreds, or FEC sets (respectively)
-   required to send an entry batch of size sz_bytes bytes.  For data and
+   required to send an entry batch of size `sz_bytes` bytes, with shreds
+   of type `type` (merkle, chained merkle, or resigned).  For data and
```
ptaffet-jump, line 94:
```suggestion
   of type `type`. `type` must be one of FD_SHRED_TYPE_MERKLE_{DATA, DATA_CHAINED, DATA_CHAINED_RESIGNED}.  For data and
```

src/discoh/poh/fd_poh_tile.c
```diff
@@ -1886,6 +1905,18 @@ publish_microblock( fd_poh_ctx_t *      ctx,
   meta->reference_tick = (ctx->hashcnt/ctx->hashcnt_per_tick) % ctx->ticks_per_slot;
   meta->block_complete = !ctx->hashcnt;
 
+  /* From poh_reset we received the block_id for ctx->parent_slot.
+     Now we're telling shred tile to build on parent: (slot-meta->parent_offset).
+     The block_id that we're passing is valid iff the two are the same,
+     i.e. ctx->parent_slot == (slot-meta->parent_offset). */
+  meta->parent_block_id_valid = ctx->parent_slot == (slot-meta->parent_offset);
```
ptaffet-jump, line 1912: Github re-ordered the comments, so I think what I said got a bit jumbled. No need to repeat the whole comment here. You can just say something like `/* see above */`

src/discoh/poh/fd_poh_tile.c
```diff
@@ -1886,6 +1905,18 @@ publish_microblock( fd_poh_ctx_t *      ctx,
...
+  if( FD_LIKELY( meta->parent_block_id_valid ) ) {
+    FD_LOG_INFO(( "sending microblock slot=%lu parent=%lu block_id!=null", slot, slot-meta->parent_offset ));
```
ptaffet-jump, line 1914: What I meant to say was that this line is too much logging. Even in the error case (down two lines), I'd get rid of it. Our logging budget is O(1) per slot, and this is more than that. If you want to increment a counter, okay, but I don't think that's necessary because you already have the other metric.

ptaffet-jump (commented):
Okay, I think the last two comments

src/discoh/poh/fd_poh_tile.c
jvarela-jump: done

src/discoh/poh/fd_poh_tile.c
jvarela-jump: The invalid_block_id_cnt metric has already been added to the shred tile (in this PR).

PR #4527, opened 2025-03-13
runtime: add function and unit test for detecting txn read/write conflicts

yhzhangjump:


src/flamenco/runtime/fd_runtime.c
```diff
@@ -1315,6 +1315,55 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+fd_acct_addr_hash( fd_acct_addr_t const * key, ulong seed ) {
+  return fd_hash( seed, key->b, FD_TXN_ACCT_ADDR_SZ );
+}
+
+#define MAP_NAME     fd_txn_writes
+#define MAP_KEY_T    fd_acct_addr_t
+#define MAP_KEY_EQ   fd_acct_addr_eq
+#define MAP_KEY_HASH fd_acct_addr_hash
+#define MAP_T        fd_txn_writes_t
+#include "../../util/tmpl/fd_map_giant.c"
```
ptaffet-jump, line 1333: Do you really want a map_giant here and not a map_dynamic or just map?

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1315,6 +1315,55 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+  for( ulong i=0; i<txn_cnt; i++ ) {
+    fd_txn_p_t *                   _txn = txns + i;
+    fd_txn_t *                      txn = TXN(_txn);
+    void *                      payload = _txn->payload;
+    fd_acct_addr_t const * account_keys = fd_txn_get_acct_addrs( txn, payload );
+
+    for( fd_txn_acct_iter_t i=fd_txn_acct_iter_init( txn, FD_TXN_ACCT_CAT_WRITABLE );
+         i!=fd_txn_acct_iter_end();
+         i=fd_txn_acct_iter_next( i ) ) {
+      fd_acct_addr_t const * writable_acc = &account_keys[ fd_txn_acct_iter_idx( i ) ];
```
ptaffet-jump, line 1351: This doesn't handle address lookup table accounts properly and is going to lead to out-of-bounds accesses

yhzhangjump: Ah, thanks, I didn't know about this. I will treat lookup table accounts carefully after I finish some simple unit tests.

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1315,6 +1315,135 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+
+      fd_acct_addr_t * writable_accts = (fd_acct_addr_t *) fd_type_pun((uchar *)val + addr_lut->writable_off);
+      for( uchar k=0; k<addr_lut->writable_cnt; k++ ) {
+        fd_txn_writes_t * ele = fd_txn_writes_query( acct_map, writable_accts[k], NULL );
+        if( FD_LIKELY( ele ) ) fd_txn_writes_remove( acct_map, ele );
+      }
+    }
+  }
+  return err;
+}
```
lidatong, line 1445: instead of doing it like this and re-parsing and re-querying everything you can just clear the map at the beginning of the function

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1315,6 +1315,135 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+
+    /* Iterate all writable accounts within the txn binary */
+    for( fd_txn_acct_iter_t j=fd_txn_acct_iter_init( TXN(txn), FD_TXN_ACCT_CAT_WRITABLE );
+         j!=fd_txn_acct_iter_end();
+         j=fd_txn_acct_iter_next( j ) ) {
+      fd_acct_addr_t * writable_acc = &account_keys[ fd_txn_acct_iter_idx( j ) ];
+
+      if( FD_UNLIKELY( fd_txn_writes_query( acct_map, *writable_acc, NULL ) ) ) {
+        err = FD_TXN_CONFLICT_WRITE_WRITE;
+        goto cleanup;
```
lidatong, line 1355: the control flow is confusing to have both `err` that is eventually returned and `goto` -- can you avoid using `goto` and just return? we generally avoid using `goto` (the exception is a stacked `goto` pattern for initialization failures)

instead you can just clear the map at the start of the function

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1315,6 +1315,135 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+#define MAP_MEMOIZE           0
+
+#include "../../util/tmpl/fd_map_dynamic.c"
+
+ulong
+fd_runtime_microblock_verify_read_write_conflicts( fd_txn_p_t *      txns,
+                                                   ulong             txn_cnt,
+                                                   fd_txn_writes_t * acct_map,
+                                                   fd_funk_t *       funk,
+                                                   fd_funk_txn_t *   funk_txn ) {
```
lidatong, line 1339: return the conflicting accounts optionally `conflict1_opt` `conflict2_opt`

src/flamenco/runtime/fd_runtime.h
```diff
@@ -385,6 +385,28 @@ fd_runtime_microblock_verify_ticks( fd_exec_slot_ctx_t *        slot_ctx,
...
+  fd_acct_addr_t key;
+  ulong next;
+};
+typedef struct fd_txn_writes fd_txn_writes_t;
+#define FD_TXN_CONFLICT_MAP_SEED      (0UL)
+
+#define FD_TXN_CONFLICT_OK            (0UL)
+#define FD_TXN_CONFLICT_READ_WRITE    (1UL)
+#define FD_TXN_CONFLICT_WRITE_WRITE   (2UL)
+#define FD_TXN_CONFLICT_MAP_TOO_SMALL (3UL)
```
lidatong, line 401: can you change this to follow this convention (use `int`) and negative err values

`FD_RUNTIME_SUCCESS                     ( 0)`
`FD_RUNTIME_ERR_TXN_CONFLICT_RW (-1)`

src/flamenco/runtime/fd_runtime.c
yhzhangjump: This has been fixed. Thanks again.

src/flamenco/runtime/fd_runtime.c
yhzhangjump: This has also been fixed. Thanks again.

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1238,6 +1238,110 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+
+#include "../../util/tmpl/fd_map_dynamic.c"
+
+int
+fd_runtime_microblock_verify_read_write_conflicts( fd_txn_p_t *      txns,
+                                                   ulong             txn_cnt,
+                                                   fd_txn_writes_t * acct_map,
+                                                   fd_funk_t *       funk,
+                                                   fd_funk_txn_t *   funk_txn,
+                                                   fd_acct_addr_t ** conflict_opt ) {
```
lidatong, line 1263: the lifetime looks suspicious. it's tied to either the txn or the acct lookup. can you just memcpy into an `fd_acct_addr_t *`?

lidatong: alternatively, you can document the lifetime semantics of the returned value

agave
yufeng-jump, line 1: Don't think the revert here is intentional?

yufeng-jump (changes requested):
nit: the convention is to prefix commit messages with the submodule they are touching and for this one it probably makes sense to do something like runtime: ...

agave
yhzhangjump: Wait what you mean? BTW, I have changed the PR title.

yufeng-jump: The PR seems to have reverted some changes in the agave submodule that I think are GUI related?

yufeng-jump: Thanks.  Sorry to nit again but could you change the commit message too?

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1238,6 +1238,110 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+#define MAP_KEY_EQUAL(k0,k1)  (0==memcmp((k0).b,(k1).b,32))
+#define MAP_KEY_HASH(key)     fd_hash( FD_TXN_CONFLICT_MAP_SEED, key.b, 32 )
+#define MAP_KEY_INVAL(k)      (0==memcmp(&fd_acct_addr_null, (k).b, 32))
+#define MAP_KEY_EQUAL_IS_SLOW 0
+#define MAP_MEMOIZE           0
+
+#include "../../util/tmpl/fd_map_dynamic.c"
+
+int
+fd_runtime_microblock_verify_read_write_conflicts( fd_txn_p_t *      txns,
```
topointon-jump, line 1258: What does Agave do if there are conflicts? Could you add a permalink to their source code, to make sure we match their behaviour?

yhzhangjump: I will. BTW, this PR is trying to fix this issue opened by Liam&Arayman: https://github.com/firedancer-io/firedancer/issues/1840

src/flamenco/runtime/test_txn_rw_conflicts.c
```diff
@@ -0,0 +1,230 @@
+#include "../../util/fd_util_base.h"
+#include "../fd_flamenco_base.h"
+#include "fd_runtime.h"
+#include "../../disco/pack/fd_microblock.h"
```
lidatong, line 4: remove, you're not using and is a circ dep

agave
lidatong: can you also address the submodule @yhzhangjump ?

yhzhangjump: Yes of course, I have removed modifications to `agave` in this PR. Sorry for the typo.

ptaffet-jump (changes requested):
See comments inline

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1238,6 +1238,110 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
   return FD_BLOCK_OK;
 }
 
+/* MAP_KEY_NULL cannot be all 0s because it is the address of the Solana transfer system program */
+static fd_acct_addr_t fd_acct_addr_null = {.b={0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF}};
```
ptaffet-jump, line 1242: I know the `NULL` key makes this kind of thing hard, but can't a malicious validator produce a block that uses this pubkey. I don't think it's safe to say this pubkey will never be added to the map.

src/flamenco/runtime/fd_runtime.c
yhzhangjump: I added pointers to Agave in `fd_runtime.h` in this PR. Please let me know whether that works.

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1238,6 +1238,110 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+                                                   fd_txn_writes_t * acct_map,
+                                                   fd_funk_t *       funk,
+                                                   fd_funk_txn_t *   funk_txn,
+                                                   fd_acct_addr_t *  conflict_opt ) {
+  fd_txn_writes_clear( acct_map );
+
+  /* Insert all writable accounts into acct_map and detect write-write conflicts */
+  for( ulong i=0; i<txn_cnt; i++ ) {
+    fd_txn_p_t *              txn = txns + i;
+    fd_acct_addr_t * account_keys = (fd_acct_addr_t *)fd_type_pun( (void*)fd_txn_get_acct_addrs( TXN(txn), txn->payload ) );
```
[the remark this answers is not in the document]
yhzhangjump: Very good question. The issue has 2 parts:
(1) `map_giant` has the `MAP_(query_const)` function taking a const-type key for query, but `map_dynamic` does not have this function. 
(2) the `fd_txn_get_acct_addrs` interface function from `fd_txn.h` returns `const fd_acct_addr_t *`.

I am certainly happy to avoid casting here if possible. Let me know any suggestions.

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1238,6 +1238,110 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+    /* Iterate all writable accounts in address lookup tables and insert into acct_map */
+    fd_txn_acct_addr_lut_t const * addr_luts = fd_txn_get_address_tables_const( TXN(txn) );
+    for( uchar j=0; j<TXN(txn)->addr_table_lookup_cnt; j++ ) {
+      fd_txn_acct_addr_lut_t const * addr_lut = &addr_luts[j];
+      fd_pubkey_t const *        addr_lut_acc = (fd_pubkey_t *)((uchar *)txn->payload + addr_lut->addr_off );
+      fd_funk_rec_key_t          funk_rec_key = fd_acc_funk_key( addr_lut_acc );
+      fd_funk_rec_t const *      addr_lut_rec = fd_funk_rec_query( funk, funk_txn, &funk_rec_key );
+      void *                              val = fd_funk_val( addr_lut_rec, fd_funk_wksp( funk ) );
+
+      fd_acct_addr_t * writable_accts = (fd_acct_addr_t *) fd_type_pun((uchar *)val + addr_lut->writable_off);
```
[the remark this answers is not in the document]
yhzhangjump: You're totally right. I will correct this by referencing the use of ALT in function `fd_executor_setup_accessed_accounts_for_txn`.

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1238,6 +1238,110 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+         j!=fd_txn_acct_iter_end();
+         j=fd_txn_acct_iter_next( j ) ) {
+      fd_acct_addr_t * writable_acc = &account_keys[ fd_txn_acct_iter_idx( j ) ];
+
+      if( FD_UNLIKELY( fd_txn_writes_query( acct_map, *writable_acc, NULL ) ) ) {
+        if( FD_LIKELY( conflict_opt ) ) *conflict_opt = *writable_acc;
+        return FD_RUNTIME_ERR_TXN_CONFLICT_WW;
+      }
+      if ( FD_LIKELY( !fd_txn_writes_insert( acct_map, *writable_acc ) ) )
+        return FD_RUNTIME_ERR_TXN_CONFLICT_OOO;
```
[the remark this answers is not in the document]
yhzhangjump: Yes, in theory impossible. Oops, I meant OOM out-of-memory. 😅

src/flamenco/runtime/fd_runtime.c
yhzhangjump: Yeah, I was discussing with Charlie about which NULL should we use. We can certainly not use the all 0s because that is the address of the transfer system program. Let me think a bit more.

src/flamenco/runtime/fd_runtime.c
topointon-jump: Comment looks good, thanks!! It doesn't seem like `fd_runtime_microblock_verify_read_write_conflicts` is used in our runtime code anywhere, though. Is this intentional?

yhzhangjump: Thank you! The plan is to add this check in replay tile code after verifying ticks. In Agave code, the `confirm_slot_entries` function in `blockstore_processor.rs` first verifies ticks and then verifies txn conflicts.

src/flamenco/runtime/fd_runtime.h
```diff
@@ -399,6 +399,78 @@ fd_runtime_microblock_verify_ticks( fd_exec_slot_ctx_t *        slot_ctx,
...
+   (6) bank.try_lock_accounts is called at the start of queue_batches_with_lock_retry
+   (7) this try_lock_accounts eventually calls another try_lock_accounts,
+       (see https://github.com/anza-xyz/agave/blob/v2.2.3/accounts-db/src/account_locks.rs#L24),
+       which acquires the locks and returns TransactionError::AccountInUse if read-write or write-write conflict is detected
+
+   txns is an array containing txn_cnt transactions in fd_txn_p_t type; acct_map and acct_map_bitvec are used to
+   detect the conflicts; acc_mgr and funk_txn are used to query Solana accounts in funk for address lookup tables;
+   slot and slot_hashes are needed for checking certain bounds in the address lookup table system program; lastly,
+   if out_conflict_opt is not NULL and a conflict is detected, out_conflict_opt will hold the conflicting address.
+ */
```
lidatong, line 425: thanks for documenting. can you format this to wrap at column 72?

src/flamenco/runtime/test_txn_rw_conflicts.c
```diff
@@ -0,0 +1,239 @@
...
+  /* Specify conflict type in this unit test                            */
+  /**********************************************************************/
+
+#define TEST_NO_CONFLICT
+//#define TEST_READ_WRITE
+//#define TEST_WRITE_WRITE
+
+#define ALT_TEST_NO_CONFLICT
+//#define ALT_TEST_READ_WRITE
+//#define ALT_TEST_WRITE_WRITE
```
lidatong, line 24: this is a little strange. can you put each test case in its own function and the unit test just runs every test case by calling every function? and share any common setup in a helper function

see `test_ghost.c` `test_eqvoc.c` and `test_fec_resolver.c` for examples of this

yhzhangjump: Thank you for the comments. I will address these later today after the work on archiver mini-topo.

src/flamenco/runtime/fd_runtime.c
yhzhangjump: I have added special checks for the sentinel when using this map_dynamic. In my unit-test, I have also added txns that use this sentinel as writable or read-only accounts. Thanks again for pointing this out.

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1238,6 +1239,225 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+        return FD_RUNTIME_TXN_ERR_INVALID_ADDRESS_LOOKUP_TABLE_INDEX;
+      }
+      readonly_lut_accs[readonly_lut_accs_cnt++] = lookup_addrs[readonly_lut_idxs[j]];
+    }
+  }
+
+  if( FD_LIKELY( out_accts_cnt_opt ) ) {
+    *out_accts_cnt_opt += writable_lut_accs_cnt + readonly_lut_accs_cnt;
+  }
+  fd_memcpy( &out_accts_alt[writable_lut_accs_cnt], readonly_lut_accs, readonly_lut_accs_cnt * sizeof(fd_acct_addr_t) );
```
ptaffet-jump, line 1348: You can save this memcpy by scanning through the ALTs and adding up the number of writable accounts before you start the main loop

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1238,6 +1239,225 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+                                                   fd_slot_hash_t *         slot_hashes,
+                                                   fd_acct_addr_t *         out_conflict_opt ) {
+  /* Clear all the entries in acct_map and acct_map_bitvec */
+  for( ulong idx=fd_txn_writes_bitvec_iter_init( acct_map_bitvec );
+       !fd_txn_writes_bitvec_iter_done( idx );
+       idx=fd_txn_writes_bitvec_iter_next( acct_map_bitvec, idx ) ) {
+    fd_txn_writes_t * entry = acct_map+idx;
+    fd_txn_writes_remove( acct_map, entry );
+  }
+  fd_txn_writes_bitvec_null( acct_map_bitvec );
```
ptaffet-jump, line 1370: Hmm okay, I know you added this in response to some of my previous comments, but I think this is not a good idea. 

First, there's a correctness issue: map_dynamic remove can move elements. I know that makes map_dynamic a bit difficult to use, but you can get around it by being very careful. 
For example, suppose A and B both hash to bucket 1. Then when A is inserted, it'll go to bucket 1, inserting 1 into the bitset.  When B is inserted, it'll go to the next bucket in probe sequence, which is bucket 2, inserting 2 into the bitset. When you delete from bucket 1, A will be deleted, and then the map will fix the probe sequence by moving B to bucket 1. Then you'll try to delete bucket 2, which is empty. Element B will stay in the map.

Second, this is pretty bad for performance. Scanning through and writing 8 MB is not very cache friendly.

Here are three ideas:

1. Just call `map_clear`. This function is slow and has similar cache thrashing issues.

1. Iterate through all the accounts in all the transactions once, inserting them into the map. Once you know whether there's a conflict, then you can iterate again, deleting them from the map. This is a bit of a pain because you have to load the address lookup tables twice.

1. The approach I use in pack (which is not officially supported but works) to solve similar problems is to have a paired map and array of inserted elements.  Iterate once, and when you insert a new element (when it's not a read-read), you append it to the list. Then when you want to clear the map, you iterate either call `map_clear` if the map is pretty full, or you iterate through the list and remove each.  If you are careful to iterate in reverse insertion order, you can do it by slot_idx or pointer. If you don't want to risk it, then you can just make the inserted list a list of keys instead.  Going back to the previous example, you can see that if you knew B was inserted after A, and you deleted B first, then it would be correct.

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1238,6 +1239,225 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+        continue;
+      }
+
+      /* writable_acc is not fd_acct_addr_null */
+      if( FD_UNLIKELY( fd_txn_writes_query( acct_map, writable_acc, NULL ) ) ) {
+        if( FD_LIKELY( out_conflict_opt ) ) *out_conflict_opt = writable_acc;
+        return FD_RUNTIME_TXN_ERR_WRITE_WRITE_CONFLICT;
+      }
+      fd_txn_writes_t * entry = fd_txn_writes_insert( acct_map, writable_acc );
+      if ( FD_LIKELY( !entry ) ) return FD_RUNTIME_TXN_ERR_OUT_OF_MEMORY_WHEN_DETECTING_CONFLICTS;
```
ptaffet-jump, line 1413: Is this possible? How do you expect the caller to handle it. Can we size things so that it's not possible?

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1238,6 +1239,225 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+  }
+
+  /* Query all read-only accounts from acct_map and detect read-write conflicts */
+  for( ulong i=0; i<txn_cnt; i++ ) {
+    fd_txn_p_t *                 txn = txns + i;
+    const fd_acct_addr_t * accts_imm = fd_txn_get_acct_addrs( TXN(txn), txn->payload );
+    ulong              accts_imm_cnt = fd_txn_account_cnt( TXN(txn), FD_TXN_ACCT_CAT_IMM );
+    fd_acct_addr_t         accts_alt[ FD_TXN_ADDR_TABLE_LOOKUP_MAX ];
+
+    int err = fd_runtime_load_txn_address_lookup_tables( TXN(txn),
```
ptaffet-jump, line 1425: I don't love the fact that you are expanding each ALT twice

src/flamenco/runtime/fd_runtime.h
```diff
@@ -399,6 +399,91 @@ fd_runtime_microblock_verify_ticks( fd_exec_slot_ctx_t *        slot_ctx,
...
+   acct_map and acct_map_bitvec are used to detect the conflicts; acc_mgr
+   and funk_txn are used to query Solana accounts for the address lookup
+   tables; slot and slot_hashes are needed for checking certain bounds in
+   the address lookup table system program; lastly, if out_conflict_opt
+   is not NULL and a conflict is detected, out_conflict_opt will hold the
+   conflicting address when this function returns.
+ */
+struct fd_txn_writes {
+  fd_acct_addr_t key;
+  ulong next;
```
ptaffet-jump, line 441: What is this next field? It is not used, right?

src/flamenco/runtime/fd_runtime.c
ptaffet-jump: I guess another idea is you could use one of the other map types with different rules.

ptaffet-jump (changes requested):
Sorry, still one subtle correctness issue that's going to give you rare issues that make you pull your hair out.

I also just want to make sure: are you aware of write lock demotion behavior in Agave? Is that supposed to factor in here?

src/flamenco/runtime/fd_runtime.h
yhzhangjump: Good catch. I guess it's leftover from map_giant.

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1238,6 +1239,212 @@ fd_runtime_block_verify_ticks( fd_blockstore_t * blockstore,
...
+      } else {
+        fd_txn_writes_t * entry = fd_txn_writes_insert( acct_map, readonly_acc );
+        entry->writable         = 0;
+      }
+    }
+  }
+
+  /* Clear all the entries inserted into acct_map */
+  for( ulong i=0; i<curr_idx; i++ ) {
+    fd_txn_writes_t * found = fd_txn_writes_query( acct_map, acct_arr[i], NULL );
```
ptaffet-jump, line 1442: This could query for the sentinel. You should either have a comment saying why that's okay or a check to prevent it

src/flamenco/runtime/fd_runtime.h
```diff
@@ -399,6 +399,87 @@ fd_runtime_microblock_verify_ticks( fd_exec_slot_ctx_t *        slot_ctx,
                                     ulong              max_tick_height,
                                     ulong              hashes_per_tick );
 
+/*
+   fd_runtime_microblock_verify_read_write_conflicts verifies that a
+   list of txns (e.g., those in a microblock) do not have read-write
+   or write-write conflits; the function requires fd_txn_writes and
+   fd_txn_writes_bitvec, and the usage can be found in the unit test
```
ptaffet-jump, line 406: The comment needs to be updated for the change from bitvec to arr

src/flamenco/runtime/fd_runtime_err.h
```diff
@@ -43,5 +43,7 @@
 #define FD_RUNTIME_TXN_ERR_PROGRAM_EXECUTION_TEMPORARILY_RESTRICTED  -36
 #define FD_RUNTIME_TXN_ERR_UNBALANCED_TRANSACTION                    -37
 #define FD_RUNTIME_TXN_ERR_PROGRAM_CACHE_HIT_MAX_LIMIT               -38
+#define FD_RUNTIME_TXN_ERR_READ_WRITE_CONFLICT                       -39
+#define FD_RUNTIME_TXN_ERR_WRITE_WRITE_CONFLICT                      -40
```
ptaffet-jump, line 47: Is it okay to just add errors here? I think these errors come from Agave

yhzhangjump:
I have addressed all the comments by far, as well as Philip's comment on write lock demotion. I have tested write lock demotion in my unit test as well. I added the `fd_txn_account_is_writable_idx_flat` function on the side of `fd_txn_account_is_writable_idx`, and hope this is fine. Let me know if there are more issues and I'm more than happy to fix.

ptaffet-jump (dismissed):
Okay, a few more thoughts, but I'm also inclined to just let you be done with this

src/flamenco/runtime/context/fd_exec_txn_ctx.c
```diff
@@ -322,25 +322,31 @@ fd_exec_txn_ctx_reset_return_data( fd_exec_txn_ctx_t * txn_ctx ) {
...
+
+uint
+fd_txn_account_has_bpf_loader_upgradeable( const fd_pubkey_t * account_keys,
+                                           const ulong         accounts_cnt ) {
+  for( ulong j = 0; j < accounts_cnt; j++ ) {
+    const fd_pubkey_t * acc = &account_keys[j];
     if ( memcmp( acc->uc, fd_solana_bpf_loader_upgradeable_program_id.key, sizeof(fd_pubkey_t) ) == 0 ) {
-      bpf_upgradeable_in_txn = 1U;
+      return 1U;
       break;
```
ptaffet-jump, line 346:
```suggestion

```

yhzhangjump: good catch. removed.

ptaffet-jump (approved):

PR #4533, opened 2025-03-14
replay: track shred idxs to trigger slice replay

emwang-jump:


src/discof/replay/fd_replay.c
```diff
@@ -100,3 +100,14 @@ fd_replay_delete( void * replay ) {
 
   return replay;
 }
+
+// void
+// fd_replay_idxs_insert( fd_replay_t * replay, ulong slot, uint start_idx, uint end_idx ) {
```
KVolodin, line 105: Did this comment end up here by accident?

PR #4542, opened 2025-03-17
repair_metrics: Simple metrics of the repair tile

jumpsiegel:
This is the first pass of the repair metrics.  After this, we want to tease out the metrics that tell us how successful we are actually being repairing missing data but that is a later PR


src/disco/metrics/metrics.xml
```diff
@@ -595,6 +595,32 @@ metric introduced.
...
+    <counter name="RecvServCorruptPkt"        clickhouse_exclude="true"                                 summary="How many corrupt server packets have we received" />
+    <counter name="RecvServInvalidSignature"  clickhouse_exclude="true"                                 summary="How many invalid signatures have we received" />
+    <counter name="RecvServFullPingTable"     clickhouse_exclude="true"                                 summary="Is our ping table full and causing packet drops" />
+    <counter name="RecvServPktTypes"          clickhouse_exclude="true"  enum="RepairServPktTypes"      summary="Server messages received" />
+    <counter name="RecvPktCorruptedMsg"       clickhouse_exclude="true"                                 summary="How many corrupt messages have we received" />
+    <counter name="SendPktCnt"                clickhouse_exclude="true"                                 summary="How many packets have sent" />
+    <counter name="SentPktTypes"              clickhouse_exclude="true"  enum="RepairSentRequestTypes"  summary="What types of client messages are we sending" />
```
topointon-jump, line 622: nit, but isn't `SendPktCnt` redundant? since we can just sum up all the `SentPktTypes`

src/flamenco/repair/fd_repair.c
```diff
@@ -1417,3 +1438,8 @@ fd_repair_recv_serv_packet( fd_repair_t *                 glob,
   } FD_SCRATCH_SCOPE_END;
   return 0;
 }
+
+fd_repair_metrics_t *
+fd_repair_get_metrics( fd_repair_t * repair ) {
+  return &repair->metrics;
+}
```
topointon-jump, line 1445: Don't we need to acquire the repair lock whilst we access these metrics?

src/disco/metrics/metrics.xml
jumpsiegel: I liked the double check to make sure all the cases were covered in the version that is broken out

src/flamenco/repair/fd_repair.c
jumpsiegel: No since there no pointers or complex data structures.  I would rather have a metrics snapshot be inconsistent then the metrics system lock out actual work being done

topointon-jump (approved):

PR #4561, opened 2025-03-18
feat(shred,repair): shred publish to repair

lidatong:
This PR changes the shred tile and topo such that shred now publishes to repair (Firedancer-only). The idea is the repair tile will now fan-in the shred tiles, and will be a single tile. It will handle merkle chaining validation (which requires serial ordering). It will also know how to repair late shreds inside in-progress FECs, and do this intelligently (only repair shreds in a FEC set that aren't redundant with what's already been received). Finally, repair will also receive and recover duplicate block issues, subsuming / removing the need for the separate eqvoc tile.

The shred_repair sig is now encoded as two different types (shred header or FEC set complete). Previously this was encoded in a single bit on a common message type.

Now the shred tile publishes different message types indicated in the sig. It only publishes to the link if it exists, and only writes to the blockstore if it exists. This simplifies the logic in the shred tile publishing.

Closes #4528


src/disco/fd_disco_base.h
```diff
@@ -123,29 +133,60 @@ fd_disco_replay_old_sig( ulong slot,
...
-     | slot[32,63]    | shred_idx[17,32]     | fec_idx[2,16]    | is_parity[1]                  | is_complete[0]                       | */
-
-  ulong slot_ul        = fd_ulong_min( (ulong)slot,        (ulong)UINT_MAX              );
-  ulong shred_idx_ul   = fd_ulong_min( (ulong)shred_idx,   (ulong)FD_SHRED_MAX_PER_SLOT );
-  ulong fec_set_idx_ul = fd_ulong_min( (ulong)fec_set_idx, (ulong)FD_SHRED_MAX_PER_SLOT );
-  ulong is_code_ul     = (ulong)is_code;
-  ulong completes_ul   = (ulong)completes;
-  return slot_ul << 32 | shred_idx_ul << 17 | fec_set_idx_ul << 2 | is_code_ul << 1 | completes_ul;
+fd_disco_shred_repair_sig( int type, int is_code_or_data_completes, ulong slot, uint fec_set_idx, uint shred_idx_or_data_cnt_or_parent_off ) {
+  ulong type_ul                                = (ulong)type;
```
KVolodin, line 167: To prevent unexpected behavior, ensure that 'type' is either FD_DISCO_SHRED_REPAIR_SIG_TYPE_HDR or FD_DISCO_SHRED_REPAIR_SIG_TYPE_FEC.

`assert(type == FD_DISCO_SHRED_REPAIR_SIG_TYPE_HDR || type == FD_DISCO_SHRED_REPAIR_SIG_TYPE_FEC);`

PR #4565, opened 2025-03-19
Increase file descriptor rlimit of sock tile

ripatel-fd:


ripatel-fd:
~~Seems like we don't need this~~

src/disco/net/sock/fd_sock_tile.c
```diff
@@ -579,6 +579,13 @@ metrics_write( fd_sock_tile_t * ctx ) {
   FD_MCNT_SET( SOCK, TX_DROP_CNT,       ctx->metrics.tx_drop_cnt      );
 }
 
+static ulong
+rlimit_file_cnt( fd_topo_t const *      topo,
+                 fd_topo_tile_t const * tile ) {
+  fd_sock_tile_t * ctx = fd_topo_obj_laddr( topo, tile->tile_obj_id );
+  return RX_SOCK_FD_MIN + ctx->sock_cnt;
```
jherrera-jump, line 586:
```suggestion
  return ctx->sock_cnt;
```

jherrera-jump (approved):

src/disco/net/sock/fd_sock_tile.c
jherrera-jump: nvm, this is wrong

PR #4599, opened 2025-03-21
tiles: sanity check size values

two-heart:


mmcgee-jump:
I think adding this here gives a misleading impression that it secures the tile from OOB reads/writes (which is what this line does), rather than just checks/enforces a non-security sensitive invariant. I kind of prefer not to add these checks (it's like peeing in the ocean, should we add 1000s of other checks too, like that slot is not 0, or not ULONG_MAX, or that we became leader for a slot we are leader for, etc, etc, etc, .... ?)

two-heart:
Fair point :+1:

Issue #4504, opened 2025-03-12
Fix GUI tip calculation rounding


mmcgee-jump:
/* This logic is done this way, rather than tips * 95 / 100 to match the order of
            operations and arimetic precendence in the tip payment program itself. */
         let tips = tip_accounts_after_tx.saturating_sub(tip_accounts_before_tx);
         let tips = tips - tips
             .checked_mul(5)
             .unwrap()
             .checked_div(100)
             .unwrap();

This rounding should be done on the total tip value in the block, not on each individual transaction tip. This is causing a small error accumulation in the frontend.

PR #4603, opened 2025-03-21
gui: add metrics for txn chart

jherrera-jump:
Adds per-transaction metadata to the websocket api

fixes https://github.com/firedancer-io/firedancer/issues/4504

agave [diff](https://github.com/firedancer-io/agave/compare/15cbd2b6cf99e2..jherrera/gui-txn-metrics~4)


book/api/websocket.md
```diff
@@ -1214,7 +1220,7 @@ new validator identity.
...
 | tile_primary_metric | `TilePrimaryMetric\|null` | If the slot is not `mine`, will be `null`. Otherwise, max value of per-tile-type primary metrics since the end of the prior leader slot |
 | tile_timers         | `TsTileTimers[]\|null`    | If the slot is not `mine`, will be `null`. Otherwise, an array of `TsTileTimers` samples from the slot, sorted earliest to latest. We store this information for the most recently completed 4096 leader slots. This will be `null` for leader slots before that |
-| compute_units       | `ComputeUnits\|null`      | If the slot is not `mine`, will be `null`. Otherwise, a listing of the timestamps when the compute units watermark of the slot changed, and how many banks were active |
+| transactions        | `Transactions\|null`      | If the slot is not `mine`, will be `null`. Otherwise, metrics for the transactions in this slot. Arrays have a seperate entry for each transaction that was packed in this slot, and are ordered in the same order the transactions were packed. |
```
mmcgee-jump, line 1223: Extreme style nits for operator facing documentation: no period at end of these docs
```suggestion
| transactions        | `Transactions\|null`      | If the slot is not `mine`, will be `null`. Otherwise, metrics for the transactions in this slot. Arrays have a seperate entry for each transaction that was packed in this slot, and are ordered in the same order the transactions appear in the block |
```

book/api/websocket.md
```diff
@@ -1267,12 +1273,67 @@ new validator identity.
...
-| active_bank_count             | `number[]`    | An array, the same length as `compute_unit_timestamps_nanos`. For each timestamp in the above array, this array contains an entry of how many banks were active at that timestamp |
+**`Transactions`**
+| Field                             | Type        | Description |
+|-----------------------------------|-------------|-------------|
+| max_compute_units                 | `number`    | The maximum number of compute units that can be packed into the slot |
+| start_timestamp_nanos             | `string`    | A UNIX timestamp in nanoseconds, representing the time that we started packing transactions into the slot |
+| target_end_timestamp_nanos        | `string`    | A UNIX timestamp in nanoseconds, representing the target time in nanoeconds that we should stop packing transactions for the slot. Transactions might still finish executing after this end time, if they started executing before it and ran over the deadline |
```
mmcgee-jump, line 1281:
```suggestion
| target_end_timestamp_nanos        | `string`    | A UNIX timestamp in nanoseconds, representing the target time in nanoeconds that we should stop packing transactions for the slot. Transactions might still finish executing after this end time, if they started executing before it and ran over the deadline. In rare cases, transactions may also appear to begin after this timestamp due to slight clock drift between execution cores |
```

book/api/websocket.md
```diff
@@ -1267,12 +1273,67 @@ new validator identity.
...
+| txn_start_timestamps_nanos        | `string[]`  | An array of UNIX timestamps, in nanoseconds. `txn_start_timestamps_nanos[i]` is approximatley the time when the microblock for the `i`-th transaction in the slot began executing. |
```
mmcgee-jump, line 1282: Why is this approximate, and not exact?

book/api/websocket.md
```diff
@@ -1267,12 +1273,67 @@ new validator identity.
...
+| txn_stop_timestamps_nanos         | `string[]`  | An array of UNIX timestamps, in nanoseconds. `txn_stop_timestamps_nanos[i]` is approximatley the time when the microblock for the `i`-th transaction in the slot completed executing. |
+| txn_compute_units_requested       | `number[]`  | `txn_compute_units_requested[i]` is the requested (via computeBudgetInstruction) or default compute units for the `i`-th transaction in the slot. |
+| txn_compute_units_estimated       | `number[]`  | `txn_compute_units_estimated[i]` is the estimated (by pack) cost for the `i`-th transaction in the slot, as determined by its payload contents and compute budget. |
+| txn_compute_units_rebated         | `number[]`  | `txn_compute_units_rebated[i]` is the difference between the cost estimate and the actual cost (which is strictly less than or equal to the estimate) for the `i`-th transaction in the slot. |
```
mmcgee-jump, line 1286: I would replace this with `compute_units_consumed`, and the developer can derive the rebate if they want. Most developers are never going to care about the rebate, but reading "consumed" is very clear.

book/api/websocket.md
```diff
@@ -1267,12 +1273,67 @@ new validator identity.
...
+| txn_priority_fee                  | `string[]`  | `txn_priority_fee[i]` is the priority fee in lamports for the `i`-th transaction in the slot. |
+| txn_tips                          | `string[]`  | `txn_tips[i]` is the total tip in lamports for the `i`-th transaction in the slot. Non-bundle transactions have a tip of 0. |
```
mmcgee-jump, line 1288: If I was building an app with this field, I wouldn't understand what it means without looking at the code. Does this include the Jito commission? Is it just the part going to the validator, or does it include the staker tip portion also? Is it a specific tip program / account (Jito), or consolidated across what might be multiple?

book/api/websocket.md
```diff
@@ -1267,12 +1273,67 @@ new validator identity.
...
+WouldExceedAccountDataTotalLimit        | 30   | Deprecated. |
+DuplicateInstruction                    | 31   | Contains duplicate instructions. |
+InsufficientFundsForRent                | 32   | Deprecated. |
+MaxLoadedAccountsDataSizeExceeded       | 33   | Retreived accounts data size exceeds the limit imposed for this transaction. |
+InvalidLoadedAccountsDataSizeLimit      | 34   | Requested an invalid data size (i.e. 0). |
+ResanitizationNeeded                    | 35   | Sanitized transaction differed before/after feature activiation. Needs to be resanitized. |
+ProgramExecutionTemporarilyRestricted   | 36   | Execution of a program referenced by this transaciton is restricted. |
+UnbalancedTransaction                   | 37   | The total accounts balance before the transaction does not equal the total balance after. |
+ProgramCacheHitMaxLimit                 | 38   | The program cache allocated for transaction batch for this transaction hit its load limit. |
+CommitCancelled                         | 39   | This transaction was part of a bundle that failed. |
```
mmcgee-jump, line 1339: Is this `39` correct? I think we have our own bundle failed error code (`BundlePeer`)

book/api/websocket.md
```diff
@@ -1267,12 +1273,67 @@ new validator identity.
...
+| txn_compute_units_estimated       | `number[]`  | `txn_compute_units_estimated[i]` is the estimated (by pack) cost for the `i`-th transaction in the slot, as determined by its payload contents and compute budget. |
+| txn_compute_units_rebated         | `number[]`  | `txn_compute_units_rebated[i]` is the difference between the cost estimate and the actual cost (which is strictly less than or equal to the estimate) for the `i`-th transaction in the slot. |
+| txn_priority_fee                  | `string[]`  | `txn_priority_fee[i]` is the priority fee in lamports for the `i`-th transaction in the slot. |
+| txn_tips                          | `string[]`  | `txn_tips[i]` is the total tip in lamports for the `i`-th transaction in the slot. Non-bundle transactions have a tip of 0. |
+| txn_error_code                    | `number[]`  | `txn_error_code[i]` is the error code that explains the failure for the `i`-th transaction in the slot. If the error code is 0, then the transaction succeeded. |
```
mmcgee-jump, line 1289: No period after description

src/disco/gui/fd_gui.c
```diff
@@ -1618,96 +1620,125 @@ fd_gui_plugin_message( fd_gui_t *    gui,
...
   fd_gui_slot_t * slot = gui->slots[ _slot % FD_GUI_SLOTS_CNT ];
-  if( FD_UNLIKELY( slot->slot!=_slot ) ) fd_gui_clear_slot( gui, _slot, ULONG_MAX );
 
-  if( FD_UNLIKELY( !slot->cus.has_offset[ 64UL ] ) ) slot->cus.start_offset[ 64UL ] = gui->cus.offset;
-  slot->cus.has_offset[ 64UL ] = 1;
-  if( FD_UNLIKELY( slot->cus.reference_ticks==LONG_MAX ) ) slot->cus.reference_ticks = tickcount;
+  if( FD_UNLIKELY( slot->txs.start_offset==ULONG_MAX ) ) slot->txs.start_offset = pack_txn_idx;
+  else                                                   slot->txs.start_offset = fd_ulong_min( slot->txs.start_offset, pack_txn_idx );
+
+  gui->pack_txn_idx = fd_ulong_max( gui->pack_txn_idx, pack_txn_idx );
```
mmcgee-jump, line 1679: You have a strong invariant here that must be true, so I'd assert it

```suggestion
  FD_TEST( pack_txn_idx==gui->pack_txn_idx+1UL );
  gui->pack_txn_idx++;
```

src/disco/gui/fd_gui.c
```diff
@@ -1618,96 +1620,125 @@ fd_gui_plugin_message( fd_gui_t *    gui,
...
+                                   int          is_bundle ) {
+  fd_gui_init_slot_txns( gui, tickcount, _slot );
   fd_gui_slot_t * slot = gui->slots[ _slot % FD_GUI_SLOTS_CNT ];
-  if( FD_UNLIKELY( slot->slot!=_slot ) ) fd_gui_clear_slot( gui, _slot, ULONG_MAX );
 
-  if( FD_UNLIKELY( !slot->cus.has_offset[ 64UL ] ) ) slot->cus.start_offset[ 64UL ] = gui->cus.offset;
-  slot->cus.has_offset[ 64UL ] = 1;
-  if( FD_UNLIKELY( slot->cus.reference_ticks==LONG_MAX ) ) slot->cus.reference_ticks = tickcount;
+  if( FD_UNLIKELY( slot->txs.start_offset==ULONG_MAX ) ) slot->txs.start_offset = pack_txn_idx;
+  else                                                   slot->txs.start_offset = fd_ulong_min( slot->txs.start_offset, pack_txn_idx );
```
mmcgee-jump, line 1677: I don't think you need to do anything in the else case? You will never see decreasing pack_txn_idx. I would just assert that invariant

src/disco/gui/fd_gui.c
```diff
@@ -1618,96 +1620,125 @@ fd_gui_plugin_message( fd_gui_t *    gui,
...
+    txn_entry->timestamp_delta_end_nanos = (int)((double)(tickcount - slot->txs.reference_ticks) / fd_tempo_tick_per_ns( NULL ));
+    txn_entry->txn_load_end_pct          = txn_load_end_pct;
+    txn_entry->txn_exec_end_pct          = txn_exec_end_pct;
+    txn_entry->tips                      = tips;
+    txn_entry->flags                    |= FD_GUI_TXN_FLAGS_ENDED;
   }
 
-  slot->cus.end_offset[ bank_idx ] = gui->cus.offset;
-  slot->cus.end_microblocks = (ushort)(slot->cus.end_microblocks + txn_cnt);
+  if( FD_UNLIKELY( 1UL!=txn_cnt ) ) FD_LOG_WARNING(( "gui expects 1 txn per microblock from bank, found %lu", txn_cnt ));
```
mmcgee-jump, line 1742: Strongly assert invariants like this, either LOG_ERR or FD_TEST

src/disco/gui/fd_gui.h
```diff
@@ -223,6 +236,33 @@ struct fd_gui_slot {
...
+  uint compute_units_requested     : 21; /* <= 1.4M */
+  uint compute_units_estimated     : 21; /* <= 1.4M */
+  uint compute_units_rebated       : 21; /* <= 1.4M */
+  uint flags                       :  4; /* assigned with the FD_GUI_TXN_FLAGS_* macros */
+  uint bank_idx                    :  6; /* in [0, 64) */
+  uint error_code                  :  6; /* in [0, 64) */
+  int timestamp_delta_start_nanos;
+  int timestamp_delta_end_nanos;
+
+  /* txn_{}_pct is used as a fraction of the total microblock
```
mmcgee-jump, line 251: Howcome you did this this way, rather than just having exact timestamps?

mmcgee-jump (changes requested):
.

book/api/websocket.md
jherrera-jump: Since there is some time between this timestamp (when pack constructs constructs the frag and sends it off to bank) and when the microblock actually begins to execute in the recieved bank, though I guess it really depends on what "start executing" means here. Ill change it to something more exact/descriptive.

book/api/websocket.md
jherrera-jump: I'm using the agave error codes here. This one is based on [this](https://github.com/firedancer-io/agave/blob/bcec2ca2e0218ca5470a2a5360ca520dc44898a7/core/src/banking_stage/committer.rs#L146)

src/disco/gui/fd_gui.c
jherrera-jump: I though stem could reorder frags on a link? If the order of two microblocks gets inverted wouldn't this "break" the invariant?

Also, `pack_txn_idx` might increment by more than 1 for bundles. Since this variable is used to detect if a past slot is overwritten I should probably be doing

```
gui->pack_txn_idx = fd_ulong_max( gui->pack_txn_idx, pack_txn_idx+txn_cnt-1UL )
```

src/disco/gui/fd_gui.c
jherrera-jump: Correct me if I'm wrong, but I though frags could be re-ordered in stem?

src/disco/gui/fd_gui.h
jherrera-jump: trying to keep the struct small and figured accuracy wasn't critical here, but happy to do it the other way if you don't think its an issue.

book/api/websocket.md
```diff
@@ -1267,12 +1274,67 @@ new validator identity.
...
+| txn_compute_units_consumed        | `number[]`  | `txn_compute_units_consumed[i]` is the actual post-execution cost of `i`-th transaction in the slot.  While some transactions costs are known from the transaction payload itself (such as the cost incurred by the amount of instruction data), other costs (like execution costs or the cost due to loaded on-chain account data) are a function of the state of the blockchain at the time of execution. This value represents the actual cost after a transaction is executed.  Consensus requires that all validators agree on this value for a given transaction in a slot.  This value will be 0 for transactions in a failed bundle, since these transactions are never included in the block.  Regular failed transactions may still consume compute units if they are included in the block |
```
mmcgee-jump, line 1286: > This value will be 0 for transactions in a failed bundle
I'm not sure on this API design, because you've hid information. Now there's no way to see consumed CUs of a bundle txn that failed. If you just showed it, the caller can not count it if it sees it's a failed bundle.

book/api/websocket.md
```diff
@@ -1267,12 +1274,67 @@ new validator identity.
...
+| txn_priority_fee                  | `string[]`  | `txn_priority_fee[i]` is the priority fee in lamports for the `i`-th transaction in the slot.  The priority fee is computed by multiplying the requested execution cost (derived from a provided computeBudget instruction, or from a protocol defined default) by the compute unit price (derived from a seperate computeBudget instruction).  Note that this array includes the priority fee parsed from failed transactions as well, even though those transactions do not pay this fee |
```
mmcgee-jump, line 1287: Failed transactions do pay the priority fee?

book/api/websocket.md
```diff
@@ -1267,12 +1274,67 @@ new validator identity.
...
+| txn_exec_end_timstamps_nanos      | `string[]`  | An array of UNIX timestamps, in nanoseconds. `txn_exec_end_timstamps_nanos[i]` is the time when the `i`-th transaction in the slot finished executing in the SVM. At this point, the execution result is known and the transaction is ready to be committed to the block. |
+| txn_microblock_id                 | `string[]`  | `txn_microblock_id[i]` is the index of the microblock for the `i`-th transaction in the slot.  Microblocks are collections of 1+ transactions.  All of the transactions from a bundle share the same microblock. Microblock ids are monotonically increasing in the order they appear in the block and start at 0 for each slot |
+
+These are the possible error codes that might be included in `txn_error_code` and their meanings.
+
+| Code Name                             | Code | Description |
+|---------------------------------------|------|-------------|
+Success                                 | 0    | The transaction successfully executed |
```
mmcgee-jump, line 1301: I think the table formatting should be
```suggestion
| Success                                 | 0    | The transaction successfully executed |
```

book/api/websocket.md
jherrera-jump: good point, will change

book/api/websocket.md
jherrera-jump: I probably meant transactions not included in the block (e.g. failed txns from a bundle). Will change/clarify.

book/api/websocket.md
```diff
@@ -981,7 +981,8 @@ initially replay one but the cluster votes on the other one.
...
 | vote_transactions    | `number\|null` | Total number of vote transactions in the block. Will always be less than or equal to `transactions`. The number of non-vote transactions is given by `transactions - vote_transactions`
 | failed_transactions  | `number\|null` | Total number of failed transactions (vote and non-vote) in the block. Failed transactions are those which are included in the block and were charged fees, but failed to execute successfully. This is different from dropped transactions which do not pay fees and are not included in the block |
-| compute_units        | `number\|null` | Total number of compute units used by the slot |
+| max_compute_units                 | `number`    | The maximum number of compute units that can be packed into the slot.  This limit is one of many consensus-critical limits defined by the solana protocol, and helps keeps blocks small enough for validators consume them quickly.  It may grow occasionally via on-chain feature activations |
```
mmcgee-jump, line 984: nit: alignment

mmcgee-jump, line 984: Extreme nit but it might be good to rename these to `cost_units` to align with the new RPC field. Up to you though, 
just consider it, maybe it's less clear that way.

src/disco/gui/fd_gui.h
mmcgee-jump: Hmm. I guess it's fine for now, but I suspect this will probably need to be changed in future

src/discoh/bank/fd_bank_tile.c
```diff
@@ -304,7 +308,16 @@ handle_microblock( fd_bank_ctx_t *     ctx,
      it shards / scales horizontally here, while PoH does not. */
   fd_microblock_trailer_t * trailer = (fd_microblock_trailer_t *)( dst + txn_cnt*sizeof(fd_txn_p_t) );
   hash_transactions( ctx->bmtree, (fd_txn_p_t*)dst, txn_cnt, trailer->hash );
-  trailer->is_last_in_bundle = 1;
+  trailer->pack_txn_idx = ctx->_txn_idx;
+  trailer->tips = 0UL;
+
+  long tickcount    = fd_tickcount();
+  long ts_end_ns    = fd_log_wallclock();
```
mmcgee-jump, line 315: Same as below, no `wallclock` on execution path

mmcgee-jump: This logic is generally too complicated to do here. Any time you are doing divisions or wall clock stuff, that really is more of a view layer thing for the GUI / RPC / whatever that's formatting for display.

mmcgee-jump (changes requested):
One more :)

mmcgee-jump (approved):

PR #4612, opened 2025-03-24
repair: remove const for data that is written to

two-heart:
There is a write to data pointed to by const pointers: https://github.com/firedancer-io/firedancer/blob/2acde0dea0b955586fe8cffcedeacffdaf06e39b/src/flamenco/repair/fd_repair.c#L1365
Because of the cast, the compiler doesn't complain, but afaik it is still UB.


topointon-jump:
@two-heart could this be a CodeQL rule?

two-heart:
I'll see if I can model this in a CodeQL query. But in any case, I should build some static analysis that attempt finding (most) such cases.

topointon-jump (dismissed):

PR #4615, opened 2025-03-24
feat(repair): implement fec_set_chainer for repair tile

lidatong:
This provides APIs for chaining FEC sets.  Every FEC set has a parent (the immediately preceding FEC set in the slot or parent slot) and children (immediately succeeding FEC set(s) for the same slot or child slots).

The purpose of the chainer is connect FEC set parents with their children as they are received asynchronously out-of-order from the network via Turbine and Repair, and to both validate and deliver these FEC sets in-order to the caling application.

Because of forks, FEC sets can have multiple children, but this will only be true across slots (ie. the parent and child must be different slots). There is a protocol violation called equivocation (also known as "duplicates") that can break this invariant.  Equivocation can also result in other anomalies such as keying collisions (doc later). The chainer makes a best-effort attempt to detect and validate equivocation.  Examples include checking merkle roots chain correctly and checking FEC sets do not overlap.  Not all cases of equivocation can be detected by the chainer however, as not all the necessary information is available at this stage in the validator pipeline. Ultimately, if there is equivocation, it is the responsibility of the consensus module to finalize handling.

### DIFF
--- a/src/discof/repair/Local.mk
+++ b/src/discof/repair/Local.mk
@@ -1,3 +1,5 @@
 ifdef FD_HAS_SSE
 $(call add-objs,fd_repair_tile,fd_discof)
+$(call add-objs,fd_fec_chainer,fd_discof)
+$(call make-unit-test,test_fec_chainer,test_fec_chainer,fd_discof fd_flamenco fd_ballet fd_util)
 endif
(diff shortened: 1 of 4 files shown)


emwang-jump (approved):

ptaffet-jump (approved):
Stamping for Emily

PR #4624, opened 2025-03-24
fuzz, flamenco: stop uninitialized executable accounts from being included

mjain-jump:


topointon-jump (commented):
Could we add a fixture and a permalink/explanation comment? 🙏

mjain-jump:
> Could we add a fixture and a permalink/explanation comment? 🙏

There’s no fixture for this - it was incorrect behavior to load in accounts blindly without checking the error code

src/flamenco/runtime/fd_executor.c
```diff
@@ -1285,7 +1285,12 @@ fd_executor_setup_borrowed_accounts_for_txn( fd_exec_txn_ctx_t * txn_ctx ) {
       }
 
       fd_pubkey_t * programdata_acc = &program_loader_state->inner.program.programdata_address;
-      fd_txn_account_create_from_funk( &txn_ctx->executable_accounts[j], programdata_acc, txn_ctx->acc_mgr, txn_ctx->funk_txn );
+      if( FD_UNLIKELY( fd_txn_account_create_from_funk( &txn_ctx->executable_accounts[j],
```
cali-jumptrading, line 1288: I was confused why we weren't checking the error here before as well... Could you add a comment to explain why it's ok to continue when an executable account is loaded with an error?

kbhargava-jump (approved):

PR #4629, opened 2025-03-25
add an additional check on the upper bound of the shred34 shred count

marcograss:
shred34 can contain up to 34 shreds, so a check was added to enforce this


ptaffet-jump (commented):
Thanks. Could you make it a single `&` here? No need for short circuit behavior on this conditional statement.

marcograss:
> Thanks. Could you make it a single `&` here? No need for short circuit behavior on this conditional statement.

Thanks for the feedback I updated the PR

ptaffet-jump (approved):

PR #4631, opened 2025-03-25
replay/store: replay in-order batches from storei with after_credit execution

emwang-jump:
- simulates intended repair tile behavior post-refactor using the store tile to deliver batches in order. Should be able to plug `STORE_IN_IDX` with `REPAIR_IN_IDX`
- patches the "Unable to select fork" err.


lidatong (approved):
This is a hotfix for the current "legacy" replay pipeline -- a lot of this code is actively being replaced / rewritten

PR #4632, opened 2025-03-25
para_pool_fix: bug fix for race condition in para_pool_acquire

jumpsiegel:


jumpsiegel:
@kbowers-jump provided this fix...

topointon-jump (commented):
Shall we add this to release too?

topointon-jump (approved):

ripatel-fd:
@jumpsiegel We should probably update the tests to maliciously cause this edge case

PR #4639, opened 2025-03-26
fix typos in README.md

xiaobei0715:


ripatel-fd:
image

ripatel-fd:
@xiaobei0715 CI is complaining about the trailing whitespace in the file. It was there before, but the check runs whenever a file is touched. Could you remove it? E.g. with `git stripspace`

PR #4643, opened 2025-03-26
quic: switch service queue to single heap

akhinvasara-jumptrading:
Closing for now

This PR swaps the service queues for a single heap, where each connection can exist at most once. The earliest event that's after 'now' will have its corresponding connection serviced. This leaves the jumble of conn_service, conn_tx, etc unchanged. 

This fixes some correctness issues by allowing for dynamic timeouts. For example, before this PR, a connection with future retransmits may not get scheduled for service. This change lets us set a timer for the earliest retx timer that has not yet fired.


ripatel-fd:
I really like this approach. Good work.

PR #4646, opened 2025-03-27
feat(shred): add temporary force_complete API to fec resolver

lidatong:


src/disco/shred/fd_fec_resolver.h
```diff
@@ -195,6 +195,49 @@ int fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
...
+   free list.  Caller should account for this ensure they only
+   force_complete when they are certain last shred is the in fact the
+   last shred, or the consequence is the entire FEC might be incorrectly
+   discarded too early.
+
+   The last_shred is used to derive the data_shred_cnt which is
+   otherwise only available after receiving a coding shred.  It is an
+   error to force completion of a FEC set that has already received at
+   least one parity shred. */
+
```
ptaffet-jump, line 235: Maybe add a word here about the lifetime of out_fec_set. It should be the same complicated lifetime policy as described by add_shred

ptaffet-jump: If you'd like, you can also feel free to make another return value, like `NOT_COMPLETABLE` or something for some of the cases

ptaffet-jump (approved):

src/disco/shred/fd_fec_resolver.h
lidatong: > If you'd like, you can also feel free to make another return value, like `NOT_COMPLETABLE` or something for some of the cases

i was thinking about this, and it might make sense to use this error type to distinguish rejected with just the shred discarded vs. rejected where the partial FEC is removed as well, and to plumb this error type in add_shred as well?

"rejected with eviction" (_EVICTED?). i can do this in a separate PR

ptaffet-jump (approved):

PR #4653, opened 2025-03-27
flamenco, runtime: change rent_fresh_accounts to be a list

topointon-jump:
Previously `rent_fresh_accounts` was the size of the number of partitions, which meant a lot of memory was consumed and the `slot_bank` serialization/deserialization was too slow. This PR makes `rent_fresh_accounts` simply be a vector, and optimizes for the common case where we have no fresh accounts.


ibhatt-jumptrading (dismissed):

ibhatt-jumptrading (approved):
🚀

PR #4671, opened 2025-03-28
flamenco: clean up acc mgr APIs

cali-jumptrading:


src/flamenco/runtime/fd_acc_mgr.h
```diff
@@ -170,58 +138,62 @@ fd_acc_exists( fd_account_meta_t const * m ) {
      record which has an active modify_data handle, etc.)
 
    It is always wrong to cast return value to a non-const pointer.
-   Instead, use fd_acc_mgr_modify_raw to acquire a mutable handle.
+   Instead, use fd_funk_get_acc_meta_mutable to acquire a mutable handle.
 
    if txn_out is supplied (non-null), the txn the key was found in
    is returned. If *txn_out == NULL, the key was found in the root
    context */
```
ibhatt-jumptrading, line 145: I would make it very clear in the documentation or via renaming the function that the acc mgr returns the start of the funk records and it is laid out s.t. the account data directly follows the account meta

src/flamenco/runtime/fd_acc_mgr.h
```diff
@@ -170,58 +138,62 @@ fd_acc_exists( fd_account_meta_t const * m ) {
...
-void
-fd_acc_mgr_unlock( fd_acc_mgr_t * acc_mgr );
-
-/* fd_acc_mgr_set_slots_per_epoch updates the slots_per_epoch setting
-   and re-calculates the partition width.  No-op unless 'slots_per_epoch' changes. */
-
-void
-fd_acc_mgr_set_slots_per_epoch( fd_exec_slot_ctx_t * slot_ctx,
-                                ulong                slots_per_epoch );
+fd_funk_get_acc_meta_readonly( fd_funk_t *            funk,
```
topointon-jump, line 148: We probably want to make it explicit that this is only safe if it is guarenteed that no other access will modify that account's record at the same time (this is true in the exec tiles and in the bpf cache scan for example).

src/flamenco/runtime/fd_txn_account.c
```diff
@@ -86,19 +137,156 @@ fd_txn_account_make_mutable( fd_txn_account_t * acct,
...
+    return FD_ACC_MGR_ERR_WRONG_MAGIC;
+  }
 
-  return fd_acc_mgr_view( acc_mgr, funk_txn, acc_pubkey, acct_ptr );
+  /* exec tile should never call this function, so the global addresses of
+     meta and data should never be used. Instead, populate the prepared_rec
+     field so that any created records can be published with fd_txn_account_mutable_fini. */
+  acct->prepared_rec = prepare;
+  fd_txn_account_setup_mutable( acct, pubkey, meta );
+
```
topointon-jump, line 219: we could maybe add a handholding mode that tries to dereference this pointer?

src/flamenco/runtime/fd_acc_mgr.h
cali-jumptrading: makes sense. I think this is already mentioned in the documentation here: https://github.com/firedancer-io/firedancer/blob/0bf13680cfe815b6224734ab7436701773448d3c/src/flamenco/runtime/fd_acc_mgr.h#L128

src/flamenco/runtime/fd_acc_mgr.h
```diff
@@ -170,58 +138,62 @@ fd_acc_exists( fd_account_meta_t const * m ) {
...
+     fd_funk_get_acc_meta_readonly.
+   - IMPORTANT:  Return value may point to the same memory region as a
+     previous calls to fd_funk_get_acc_meta_readonly or fd_funk_get_acc_meta_mutable do,
+     for the same funk rec (account/txn pair).  fd_funk_acc_mgr APIs only promises
+     that account handles requested for different funk txns will not
+     alias. Generally, for each funk txn, the user should only ever
+     access the latest handle returned by view/modify.
+
+   Caller must eventually commit funk record.  During replay, this is
+   done automatically by slot freeze. */
```
topointon-jump, line 186: Could we make it clear that we must always call `fd_txn_account_init_from_funk_mutable` to persist the change.

src/flamenco/runtime/fd_txn_account.c
cali-jumptrading: hmm... good idea, but we would need to write to the meta / record object for the exec tile to segfault here... What would we write?

src/flamenco/runtime/fd_acc_mgr.h
cali-jumptrading: I added a note at the bottom of the header documentation for both `fd_funk_get_acc_meta_readonly` and `fd_funk_get_acc_meta_mutable`.

src/flamenco/runtime/fd_acc_mgr.h
```diff
@@ -111,50 +100,29 @@ fd_funk_key_to_acc( fd_funk_rec_key_t const * id ) {
...
-   deleted.  Handles NULL input safely.  Returns 0 if the account was
-   deleted (zero lamports, empty data, zero owner).  Otherwise, returns
-   1. */
-
-static inline int
-fd_acc_exists( fd_account_meta_t const * m ) {
-
-  if( !m ) return 0;
+/* The following fd_funk_acc_mgr APIs translate between the runtime account DB abstraction
+   and the actual funk database.  Also manages rent collection.
```
topointon-jump, line 106: Does it manage rent collection? I can't see any rent collection code in here, that has all been moved to the runtime.

cali-jumptrading: ohh this is probably because the account manager object used to store `partition_width` and `slots_per_epoch`.

cali-jumptrading: I'll remove that comment.

src/flamenco/runtime/fd_txn_account.c
topointon-jump: you could copy the first byte back into itself ;)

cali-jumptrading: was just thinking that myself! I'll go ahead and add it.

src/flamenco/runtime/fd_acc_mgr.h
cali-jumptrading: Done

src/flamenco/runtime/fd_txn_account.c
cali-jumptrading: Done

src/flamenco/runtime/fd_txn_account.c
```diff
@@ -86,19 +137,164 @@ fd_txn_account_make_mutable( fd_txn_account_t * acct,
...
+  fd_funk_rec_query_t query[1];
+
+  fd_funk_rec_key_t key = fd_funk_acc_key( acct->pubkey );
+  fd_funk_rec_t *   rec = (fd_funk_rec_t *)fd_funk_rec_query_try( funk, txn, &key, query );
+
+  /* publish the record if the record is not in the current funk transaction
+     and there exists a record in preparation in the fd_txn_account_t object */
+  if( rec==NULL && acct->prepared_rec.rec!=NULL ) {
+    fd_funk_rec_publish( &acct->prepared_rec );
+  }
```
topointon-jump, line 299: Don't we need to publish the record all the time? This is `fd_funk_rec_publish`, not `fd_funk_txn_publish` - all it does it add the record into the rec map, not publishing the funk transaction.

cali-jumptrading: well the record could already be in funk and we wouldn't want to overwrite it right? As I understand it, we should only publish the record if the the record is a new record made by `fd_txn_account_init_from_funk_mutable` (account did not exist in the funk txn).

cali-jumptrading: Added the key check!

topointon-jump: for now, we can check to see if the prepared record's pubkey is the same as the transactions account

src/flamenco/runtime/fd_txn_account.c
```diff
@@ -86,19 +137,173 @@ fd_txn_account_make_mutable( fd_txn_account_t * acct,
...
+  fd_funk_rec_key_t key = fd_funk_acc_key( acct->pubkey );
+  fd_funk_rec_t *   rec = (fd_funk_rec_t *)fd_funk_rec_query_try( funk, txn, &key, query );
+
+  /* publish the record if the record is not in the current funk transaction
+     and there exists a record in preparation in the fd_txn_account_t object */
+  if( rec==NULL && acct->prepared_rec.rec!=NULL ) {
+
+    /* Ensure that the prepared record key still matches our key.
+       Otherwise, the prepared record is invalid. */
+    if( FD_UNLIKELY( memcmp( acct->prepared_rec.rec->pair.key,
```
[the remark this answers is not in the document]
topointon-jump: I was wondering if a better check would be a version number? but this is probably fine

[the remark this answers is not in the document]
topointon-jump: we should also check that
```
  rec->val_sz      == 0U;
  rec->val_max     == 0U;
  rec->val_gaddr   == 0UL;
```
(which is the case if `fd_funk_val_flush` is called)

topointon-jump (approved):

PR #4682, opened 2025-03-31
pack: exempt votes and bundles from pacing, schedule votes first

ptaffet-jump:
The policy is now:

**Bank 0 when pacing says it should be inactive (the first X ms of the block)**:
1.   If there's a vote, schedule it and return.
 1.  If there's a bundle (almost certainly schedulable), schedule it and return.

**Bank 0 when pacing says it should be active (the rest of the block)**:
   1. If there's a vote, schedule it and return.
  1.  If there's a bundle that can be scheduled, schedule it and return.
  1. If there weren't any bundles, or the next one doesn't fit in the rest of the block, try to schedule a normal transaction
  Note: if there was a bundle but it couldn't be scheduled because of account conflicts, don't schedule anything.

**Bank >0 when pacing says it should be inactive**:
  1.  If there's a vote, schedule it and return

**Bank >0 when pacing says it should be active**:
  1. If there's a vote, schedule it and return
  1. Try to schedule a normal transaction

### DIFF
--- a/src/disco/pack/fd_pack.c
+++ b/src/disco/pack/fd_pack.c
@@ -2252,19 +2252,8 @@ fd_pack_schedule_next_microblock( fd_pack_t *  pack,
                                   ulong        total_cus,
                                   float        vote_fraction,
                                   ulong        bank_tile,
+                                  int          schedule_flags,
                                   fd_txn_p_t * out ) {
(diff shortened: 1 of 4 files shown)


ptaffet-jump:
Superseded by #4892

PR #4683, opened 2025-04-01
blockstore: check that shred_max is power of two

two-heart:
nit, but it is easy to imagine an operator putting in a value that is not a power of two, which would lead to fd_buf_shred_map_footprint being 0. We want a meaningful error message in this case.

### DIFF
--- a/src/flamenco/runtime/fd_blockstore.c
+++ b/src/flamenco/runtime/fd_blockstore.c
@@ -42,6 +42,11 @@ fd_blockstore_new( void * shmem,
     return NULL;
   }
 
+  if( FD_UNLIKELY( !fd_ulong_is_pow2( shred_max ) ) ) {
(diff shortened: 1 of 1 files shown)


src/flamenco/runtime/fd_blockstore.c
```diff
@@ -42,6 +42,11 @@ fd_blockstore_new( void * shmem,
     return NULL;
   }
 
+  if( FD_UNLIKELY( !fd_ulong_is_pow2( shred_max ) ) ) {
+    FD_LOG_WARNING(( "blockstore implementation requires shred_max to be a power of two" ));
```
jvarela-jump, line 46: as discussed with @lidatong, it may be a better approach to round up the value of shred_max to the nearest power of 2 in that case, still issue the warning but continue.

jvarela-jump (changes requested):

jvarela-jump (approved):

jvarela-jump (approved):

PR #4686, opened 2025-04-01
net: record the src IP address when running on localhost

yhzhangjump:
When running FD with a local Agave, repair requests are not sent from FD because (1) the repair tile does not specify the src IP; (2) the net tile does not record the local IP. Therefore, repair requests are dropped by the xdp net tile.


yhzhangjump:
Richie will take over this effort.

PR #4687, opened 2025-04-01
flamenco: Add low-level txn account vtable helpers

cali-jumptrading:
Hides internal private state of `fd_txn_account_t`, specifically the `const_meta`, `meta`, `const_data`, `data`, `const_rec`, `rec`, `recnt_excl`, and `meta/data gaddr` fields. Makes these fields accessble through function pointer tables to control access based on whether the account is mutable or readonly.

Agave uses a similar readable/writable account API implementation here: https://github.com/anza-xyz/agave/blob/v2.1.14/sdk/account/src/lib.rs#L188

### DIFF
--- a/src/flamenco/runtime/fd_txn_account.c
+++ b/src/flamenco/runtime/fd_txn_account.c
@@ -1,5 +1,4 @@
 #include "fd_txn_account.h"
-#include "fd_acc_mgr.h"
 #include "fd_runtime.h"
 
 fd_txn_account_t *
@@ -17,14 +16,18 @@ fd_txn_account_init( void * ptr ) {
   memset( ptr, 0, FD_TXN_ACCOUNT_FOOTPRINT );
 
   fd_txn_account_t * ret = (fd_txn_account_t *)ptr;
-  ret->const_data        = NULL;
-  ret->const_meta        = NULL;
-  ret->meta              = NULL;
-  ret->data              = NULL;
-  ret->meta_gaddr        = 0UL;
-  ret->data_gaddr        = 0UL;
-  ret->starting_dlen     = ULONG_MAX;
-  ret->starting_lamports = ULONG_MAX;
+  ret->private_state.meta_gaddr = 0UL;
+  ret->private_state.data_gaddr = 0UL;
+  ret->starting_dlen            = ULONG_MAX;
+  ret->starting_lamports        = ULONG_MAX;
+
+  ret->private_state.const_data = NULL;
+  ret->private_state.const_meta = NULL;
+  ret->private_state.meta       = NULL;
+  ret->private_state.data       = NULL;
+
+  /* Defaults to writable vtable */
+  ret->vt                       = &fd_txn_account_writable_vtable;
 
   FD_COMPILER_MFENCE();
   ret->magic = FD_TXN_ACCOUNT_MAGIC;
@@ -37,8 +40,8 @@ fd_txn_account_init( void * ptr ) {
    default values for the txn account */
 void
 fd_txn_account_setup_common( fd_txn_account_t * acct ) {
-  fd_account_meta_t const * meta = acct->const_meta ?
-                                   acct->const_meta : acct->meta;
+  fd_account_meta_t const * meta = acct->private_state.const_meta ?
+                                   acct->private_state.const_meta : acct->private_state.meta;
 
   if( ULONG_MAX == acct->starting_dlen ) {
     acct->starting_dlen = meta->dlen;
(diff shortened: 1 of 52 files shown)


src/flamenco/runtime/fd_txn_account.h
```diff
@@ -2,97 +2,76 @@
...
-  fd_account_meta_t         * meta;
-  uchar                     * data;
-  fd_funk_rec_t             * rec;
+  /* The size of the private state memory region is determined by
+     sizeof(fd_txn_account_private_state_t) */
+  uchar                           private_state[ FD_TXN_ACCOUNT_PRIVATE_STATE_SIZE ];
 
-  ulong                       meta_gaddr;
-  ulong                       data_gaddr;
+  ulong                           meta_gaddr;
```
ibhatt-jumptrading, line 22: we should probably encapsulate the meta/data gaddrs as well and expose getters/setters for those

src/flamenco/runtime/fd_txn_account.h
```diff
@@ -2,97 +2,76 @@
...
-#include "../../funk/fd_funk_rec.h"
 #include "program/fd_program_util.h"
+#include "fd_txn_account_vtable.h"
 
 struct fd_acc_mgr;
 typedef struct fd_acc_mgr fd_acc_mgr_t;
 
-struct __attribute__((aligned(8UL))) fd_txn_account {
-  ulong                       magic;
+#define FD_TXN_ACCOUNT_PRIVATE_STATE_SIZE 56
```
ibhatt-jumptrading, line 11: This is the size of the meta right? could you ifdef it as the sizeof(fd_account_meta) in case the meta changes

src/flamenco/runtime/fd_txn_account.h
```diff
@@ -2,97 +2,76 @@
...
 
-  fd_pubkey_t                 pubkey[1];
+struct __attribute__((aligned(8UL))) fd_txn_account {
```
ibhatt-jumptrading, line 13: just curious: why an align of 8

src/flamenco/runtime/fd_txn_account.c
```diff
@@ -50,31 +76,69 @@ fd_txn_account_setup_common( fd_txn_account_t * acct ) {
 }
 
 void
-fd_txn_account_setup_sentinel_meta( fd_txn_account_t * acct,
-                                    fd_spad_t *        spad,
-                                    fd_wksp_t *        spad_wksp ) {
+fd_txn_account_init_from_meta_and_data_mutable( fd_txn_account_t *  acct,
+                                                fd_account_meta_t * meta,
+                                                uchar *             data ) {
+  fd_txn_account_private_state_t * private_state = fd_type_pun( &acct->private_state );
```
ibhatt-jumptrading, line 82: nits: can you fix vertical align in this file

src/flamenco/runtime/fd_txn_account.h
ibhatt-jumptrading: We really only need getters. I believe the caller will never explictly need to call it

src/flamenco/runtime/fd_txn_account.c
```diff
@@ -321,3 +380,481 @@ fd_txn_account_mutable_fini( fd_txn_account_t * acct,
...
+void
+fd_txn_account_set_executable_readonly( fd_txn_account_t * acct FD_PARAM_UNUSED,
+                                        int                is_executable FD_PARAM_UNUSED ) {
+    FD_LOG_ERR(( "cannot set executable in a readonly account!" ));
+}
+
+void
+fd_txn_account_set_owner_readonly( fd_txn_account_t * acct FD_PARAM_UNUSED,
+                                   fd_pubkey_t const * owner FD_PARAM_UNUSED ) {
+    FD_LOG_ERR(( "cannot set owner in a readonly account!" ));
```
ibhatt-jumptrading, line 655: nit: indents seem off here

src/flamenco/runtime/fd_txn_account.c
```diff
@@ -1,6 +1,26 @@
 #include "fd_txn_account.h"
 #include "fd_acc_mgr.h"
 #include "fd_runtime.h"
+#include <assert.h>
+
+struct __attribute__((aligned(8UL))) fd_txn_account_private_state {
+  fd_account_meta_t const   *     const_meta;
```
ibhatt-jumptrading, line 7: nit: const keyword and pointers dont need veritcal align

src/flamenco/runtime/fd_txn_account.c
```diff
@@ -1,6 +1,26 @@
 #include "fd_txn_account.h"
 #include "fd_acc_mgr.h"
 #include "fd_runtime.h"
+#include <assert.h>
```
ibhatt-jumptrading, line 4: do we need this?

src/flamenco/runtime/fd_txn_account.c
```diff
@@ -321,3 +380,481 @@ fd_txn_account_mutable_fini( fd_txn_account_t * acct,
...
+void
+fd_txn_account_set_mutable( fd_txn_account_t * acct ) {
+  fd_txn_account_private_state_t * private_state = fd_type_pun( &acct->private_state );
+  private_state->meta = (void *)private_state->const_meta;
+  private_state->data = (void *)private_state->const_data;
+  private_state->rec  = (void *)private_state->const_rec;
+  acct->vt            = &fd_txn_account_writable_vtable;
+}
+
+const fd_txn_account_vtable_t
```
ibhatt-jumptrading, line 774: could we use a macro here so we don't have to manually write out two vtables?

ibhatt-jumptrading (commented):
mostly lgtm, just a few minor comments.

src/flamenco/runtime/fd_txn_account.c
cali-jumptrading: This is for the static assert here: https://github.com/firedancer-io/firedancer/blob/cb0ea39ba122a7901a35f3cbba3ceecace8f7b00/src/flamenco/runtime/fd_txn_account.c#L22

cali-jumptrading: I thought it best to use `static_assert` so we don't ever get mismatched sizes between the private state memory region and the internal private state struct.

src/flamenco/runtime/fd_txn_account.h
cali-jumptrading: I think it was there to begin with. So for the `fd_txn_account_private_state_t` struct I did the same.

src/flamenco/runtime/fd_txn_account.c
cali-jumptrading: Done

src/flamenco/runtime/fd_txn_account.c
cali-jumptrading: Fixed

src/flamenco/runtime/fd_txn_account.c
cali-jumptrading: Ok so after discussing with @ptaffet-jump, I decided to just go with a private header file. If a user really wants to directly access those fields they can, they just gotta go through the private header.

src/flamenco/runtime/fd_txn_account.c
cali-jumptrading: fixed

src/flamenco/runtime/fd_txn_account.h
cali-jumptrading: Done, though I don't see any place in the codebase that uses those gaddrs right now.

src/flamenco/runtime/fd_txn_account.c
cali-jumptrading: Not sure if a macro would work here... because some functions are shared between the two vtables.

cali-jumptrading: Actually maybe it will work, it'll just be really long. let me see if I can come up with something.

cali-jumptrading: Done

ibhatt-jumptrading (approved):

PR #4701, opened 2025-04-02
flamenco: fix bpf loader new authority key err mismatch

cali-jumptrading:
Agave did not error out immediately when obtaining the new authority key while we were returning the error. See https://github.com/anza-xyz/agave/blob/v2.1.14/programs/bpf_loader/src/lib.rs#L901-L906


mjain-jump (approved):
Looks good, thanks!

PR #4708, opened 2025-04-03
archiver: playback frags to storei and reproduce the behavior of replay

yhzhangjump:


src/disco/archiver/fd_archiver_writer.c
```diff
@@ -118,7 +116,7 @@ privileged_init( fd_topo_t *      topo,
     FD_SCRATCH_ALLOC_APPEND( l, fd_alloc_align(), fd_alloc_footprint() );
     FD_SCRATCH_ALLOC_FINI( l, scratch_align() );
 
-    tile->archiver.archive_fd = open( tile->archiver.archive_path, O_RDWR | O_CREAT | O_DIRECT, 0666 );
```
yhzhangjump, line 121: @topointon-jump the `O_DIRECT` is here and please let me know how I should fix this.

topointon-jump: we need to use direct IO so that we write at line rate. this means the flushes have to be page aligned. you can track the offset and flush at the page boundaries.

yhzhangjump: I have added `O_DIRECT` back and the trade-off is that I am now using 2MB (huge page) as the IO write buffer size instead of 1GB (gigantic page). This seems fine. I have also renamed `playback` to `shred_playback`.

topointon-jump: why can't we use gigantic? is it because it takes a while to fill up and actually do the flush?

yhzhangjump: I tried to run archiver against testnet (2 shred tiles) and the archive file increases at a rate of ~7MB/s. I guess we could think about how often we would like to write the mem buffer into disk, instead of having to make the mem buffer either 2MB or 1GB?

topointon-jump: Writing to disk has a cost (I found that it wouldn't keep up when I wrote every 2MB), so yes, I agree we need to choose a number somewhere in-between those two. Maybe writing every 2 minutes is reasonable?

src/disco/archiver/fd_archiver_playback.c
```diff
@@ -169,23 +177,82 @@ unprivileged_init( fd_topo_t *      topo,
...
+after_frag( fd_archiver_playback_tile_ctx_t * ctx,
+            ulong                             in_idx,
+            ulong                             seq    FD_PARAM_UNUSED,
+            ulong                             sig    FD_PARAM_UNUSED,
+            ulong                             sz     FD_PARAM_UNUSED,
+            ulong                             tsorig FD_PARAM_UNUSED,
+            ulong                             tspub  FD_PARAM_UNUSED,
+            fd_stem_context_t *               stem   FD_PARAM_UNUSED ) {
+  if( FD_UNLIKELY( in_idx!=0 ) ) FD_LOG_ERR(( "Playback seems corrupted." ));
+  ctx->notified = 1;
```
lidatong, line 227: where is this actually get sent in `storei` tile?

src/discof/store/fd_storei_tile.c
```diff
@@ -205,6 +217,19 @@ during_frag( fd_store_tile_ctx_t * ctx,
...
+    uchar * dst = fd_chunk_to_laddr( ctx->archive_out_mem, ctx->archive_out_chunk );
+    if( in_idx==REPAIR_IN_IDX ) {
+      ctx->archive_publish=1;
+      fd_memcpy( dst, fd_chunk_to_laddr_const( ctx->repair_in_mem, chunk ), sz );
+    } else if( in_idx>=NON_SHRED_LINKS ) {
+      ctx->archive_publish=1;
+      fd_store_in_ctx_t * shred_in = &ctx->shred_in[ in_idx-NON_SHRED_LINKS ];
+      fd_memcpy( dst, fd_chunk_to_laddr_const( shred_in->mem, chunk ), sz );
+    }
+  }
```
lidatong, line 231: instead of adding conditional / copying here, do the copy into the archiver feeder out buffer in after_frag before you publish

src/discof/store/fd_storei_tile.c
```diff
@@ -258,6 +283,15 @@ after_frag( fd_store_tile_ctx_t * ctx,
     return;
   }
 
+  if( FD_UNLIKELY( ctx->archive_publish ) ) {
+    ulong new_sig = ( in_idx==REPAIR_IN_IDX )? FD_ARCHIVER_SIG_MARK_REPAIR(sig) :
+                                               FD_ARCHIVER_SIG_MARK_SHRED(sig);
+    fd_mcache_publish( ctx->archive_out_mcache, ctx->archive_out_depth, ctx->archive_out_seq, new_sig, ctx->archive_out_chunk,
```
lidatong, line 289: use stem publish

src/app/shared/fd_config_parse.c
```diff
@@ -422,7 +422,8 @@ fdctl_pod_to_cfg( config_t * config,
   CFG_POP      ( cstr,   tiles.restart.wen_restart_coordinator            );
   CFG_POP      ( cstr,   tiles.restart.genesis_hash                       );
 
-  CFG_POP      ( bool,   tiles.archiver.playback                          );
+  CFG_POP      ( bool,   tiles.archiver.enabled                           );
+  CFG_POP      ( cstr,   tiles.archiver.archive_path                      );
```
lidatong, line 426: archiver_path

src/discof/store/fd_storei_tile.c
```diff
@@ -726,6 +760,24 @@ unprivileged_init( fd_topo_t *      topo,
...
+  /* Set up archiver output */
+  if( FD_UNLIKELY( tile->out_cnt==4 &&
+                   ( strcmp( topo->links[ tile->out_link_id[ ARCHIVE_OUT_IDX ] ].name, "store_feeder" ) ||
+                     strcmp( topo->links[ tile->out_link_id[ ARCHIVE_OUT_IDX ] ].name, "storei_notif" ) ) ) ) {
+    FD_LOG_NOTICE(( "storei tile has output link %s", topo->links[ tile->out_link_id[ ARCHIVE_OUT_IDX ] ].name ));
+    fd_topo_link_t * archive_out = &topo->links[ tile->out_link_id[ ARCHIVE_OUT_IDX ] ];
+    ctx->archive_out_mcache = archive_out->mcache;
+    ctx->archive_out_sync   = fd_mcache_seq_laddr( ctx->archive_out_mcache );
+    ctx->archive_out_depth  = fd_mcache_depth( ctx->archive_out_mcache );
+    ctx->archive_out_seq    = fd_mcache_seq_query( ctx->archive_out_sync );
```
lidatong, line 772: you dont need to do this if you use stem

src/app/firedancer-dev/commands/archive_shred_playback.c
```diff
@@ -0,0 +1,411 @@
...
+  fd_topob_wksp( topo, "shred_storei" );
+  fd_topob_wksp( topo, "repair_store" );
+  if( FD_LIKELY( wait_for_notif )) fd_topob_wksp( topo, "storei_notif" );
+  fd_topob_wksp( topo, "stake_out"    );
+  fd_topob_wksp( topo, "store_replay" );
+  /*             topo,  link_name,      wksp_name,     depth,         mtu,                    burst */
+  fd_topob_link( topo, "shred_storei", "shred_storei", 65536UL,       4UL*FD_SHRED_STORE_MTU, 4UL+config->tiles.shred.max_pending_shred_sets );
+  fd_topob_link( topo, "repair_store", "repair_store", 1024UL*1024UL, FD_SHRED_MAX_SZ,        128UL                                          );
+  if( FD_LIKELY( wait_for_notif ))
+    fd_topob_link( topo, "storei_notif", "storei_notif", 65536UL,       4UL*FD_SHRED_STORE_MTU, 4UL+config->tiles.shred.max_pending_shred_sets );
```
lidatong, line 258: make this non-optional

src/app/firedancer-dev/commands/archive_shred_playback.c
```diff
@@ -0,0 +1,411 @@
...
+   1. read an archive file and reproduce the frags into the storei tile
+   2. reproduce the behavior of the replay tile (e.g., the set of forks)
+
+   The smaller topology is:
+             shred_storei          store_replay       replay_exec
+   playback--------------->storei-------------->replay------------>exec
+     ^      \         ---->   |  <-----       /
+     |       \_______/        |        \_____/
+     |      repair_store      |       stake_out
+     |_storei_notif(optional)_|
```
lidatong, line 11: you don't need an optional link like this given this is already its own topo

src/discof/store/fd_storei_tile.c
```diff
@@ -726,6 +760,24 @@ unprivileged_init( fd_topo_t *      topo,
   ctx->restart_out_wmark  = fd_dcache_compact_wmark ( ctx->restart_out_mem, restart_out->dcache, restart_out->mtu );
   ctx->restart_out_chunk  = ctx->restart_out_chunk0;
 
+  /* Set up archiver output */
+  if( FD_UNLIKELY( tile->out_cnt==4 &&
+                   ( strcmp( topo->links[ tile->out_link_id[ ARCHIVE_OUT_IDX ] ].name, "store_feeder" ) ||
+                     strcmp( topo->links[ tile->out_link_id[ ARCHIVE_OUT_IDX ] ].name, "storei_notif" ) ) ) ) {
```
lidatong, line 766: is this actually optional if youre checking for it here?

lidatong (approved):

PR #4711, opened 2025-04-03
repair/shred: force complete FEC sets without coding shred info

emwang-jump:
Needed in order to swap to new architecture. Repair shreds only arrive as data shreds, but fec_resolver cannot complete FEC sets without at least 1 coding shred. fec_repair employs a fec_set_idx lookahead to find the upper bound of a FEC set without data_cnt information, and forces completion of that FEC set in the fec_resolver.


src/disco/shred/fd_fec_resolver.h
```diff
@@ -195,6 +198,17 @@ int fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
                                fd_shred_t const   * * out_shred,
                                fd_bmtree_node_t     * out_merkle_root );
 
+int
+fd_fec_resolver_done_contains( fd_fec_resolver_t * resolver,
+                               fd_ed25519_sig_t  * signature );
+
+/* fd_fec_resolver_last_shred_query returns the last shred in the FEC */
+int
```
ptaffet-jump, line 206: These need comments

src/disco/shred/fd_fec_resolver.h
```diff
@@ -195,6 +198,17 @@ int fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
...
+fd_fec_resolver_last_shred_query( fd_fec_resolver_t * resolver,
+                                  fd_ed25519_sig_t  * signature,
+                                  uint                last_shred_idx,
+                                  fd_shred_t        * out_last_shred );
```
ptaffet-jump, line 210: Since memory lifetimes are kind of strange in `fd_fec_resolver`, definitely mention it is copied to the region of memory pointed to by out_last_shred

src/disco/shred/fd_fec_resolver.h
```diff
@@ -195,6 +198,17 @@ int fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
...
                                fd_bmtree_node_t     * out_merkle_root );
 
+int
+fd_fec_resolver_done_contains( fd_fec_resolver_t * resolver,
+                               fd_ed25519_sig_t  * signature );
+
+/* fd_fec_resolver_last_shred_query returns the last shred in the FEC */
+int
+fd_fec_resolver_last_shred_query( fd_fec_resolver_t * resolver,
+                                  fd_ed25519_sig_t  * signature,
```
ptaffet-jump, line 208: Here and in the previous function, I think `signature` should be `const`, no?

src/disco/shred/fd_fec_resolver.c
```diff
@@ -699,7 +697,33 @@ int fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
...
+                                  uint                last_shred_idx,
+                                  fd_shred_t        * out_last_shred ) {
+  wrapped_sig_t * w_sig = (wrapped_sig_t *)signature;
+  if( FD_UNLIKELY( ctx_map_key_inval( *w_sig ) ) ) return FD_FEC_RESOLVER_SHRED_REJECTED;
+
+  set_ctx_t * ctx = ctx_map_query( resolver->curr_map, *w_sig, NULL );
+  if( FD_UNLIKELY( !ctx ) ) return FD_FEC_RESOLVER_SHRED_REJECTED;
+
+  fd_fec_set_t     * set        = ctx->set;
+  fd_shred_t const * data_shred = (fd_shred_t const *)fd_type_pun_const( set->data_shreds[ last_shred_idx ] );
```
ptaffet-jump, line 720: If there's no validation that `last_shred_idx` has been received, that definitely needs to be mentioned in the comment about the function

src/disco/shred/fd_fec_resolver.c
```diff
@@ -782,6 +808,15 @@ fd_fec_resolver_force_complete( fd_fec_resolver_t *   resolver,
...
      because it is by definition the full set of rcvd data shreds. */
 
+  set_ctx_t * done_ll_sentinel = resolver->done_ll_sentinel;
+  set_ctx_t * curr_map         = resolver->curr_map;
+  set_ctx_t * done_map         = resolver->done_map;
+  ulong       done_depth       = resolver->done_depth;
+
+  ctx_ll_insert( done_ll_sentinel, ctx_map_insert( done_map, ctx->sig ) );
+  if( FD_UNLIKELY( ctx_map_key_cnt( done_map ) > done_depth ) ) ctx_map_remove( done_map, ctx_ll_remove( done_ll_sentinel->prev ) );
+  ctx_map_remove( curr_map, ctx_ll_remove( ctx ) );
```
ptaffet-jump, line 818: See the comment in `add_shred`: `ctx_map_remove` can actually change the contents of `*ctx`, so the accesses below are possibly incorrect.

src/disco/shred/fd_shred_tile.c
```diff
@@ -321,6 +323,15 @@ during_frag( fd_shred_ctx_t * ctx,
 
   ctx->tsorig = fd_frag_meta_ts_comp( fd_tickcount() );
 
+  if( FD_UNLIKELY( ctx->in_kind[ in_idx ]==IN_KIND_REPAIR ) ) {
+    if( FD_UNLIKELY( chunk<ctx->in[ in_idx ].chunk0 || chunk>ctx->in[ in_idx ].wmark ) )
+    FD_LOG_ERR(( "chunk %lu %lu corrupt, not in range [%lu,%lu]", chunk, sz,
+                ctx->in[ in_idx ].chunk0, ctx->in[ in_idx ].wmark ));
+
+    uchar const * dcache_entry = fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk );
+    fd_memcpy( ctx->shred_buffer, dcache_entry, sz);
```
ptaffet-jump, line 332:
```suggestion
    fd_memcpy( ctx->shred_buffer, dcache_entry, sz );
```

src/disco/shred/fd_shred_tile.c
```diff
@@ -553,14 +564,45 @@ after_frag( fd_shred_ctx_t *    ctx,
     return;
   }
 
+  if( FD_UNLIKELY( ctx->in_kind[ in_idx ]==IN_KIND_REPAIR ) ) {
+    fd_ed25519_sig_t * shred_sig = (fd_ed25519_sig_t *)fd_type_pun( ctx->shred_buffer );
+    if( FD_UNLIKELY( fd_fec_resolver_done_contains( ctx->resolver, shred_sig ) ) ) {
+      /* This is a FEC completion message from the repair tile.  We need
+         to make sure that we don't force complete something that's just
+         been completed. */
+      FD_LOG_INFO(( "FEC completion message from repair tile, but shred tile %lu has just completed it.", ctx->round_robin_id ));
```
ptaffet-jump, line 573: This should be a metric `MCNT_INC`, not an FD_LOG_INFO

src/disco/shred/fd_shred_tile.c
```diff
@@ -553,14 +564,45 @@ after_frag( fd_shred_ctx_t *    ctx,
...
+      return;
+    } /* FIXME: Is there a chance something will be freed so fast that the sig lives in free_list? */
+
+    FD_LOG_INFO(( "FEC completion message from repair tile, and shred tile %lu has not completed it yet.", ctx->round_robin_id ));
```
ptaffet-jump, line 577: Same, you get O(1) log per slot in the expected case, so this shouldn't be here

src/disco/shred/fd_shred_tile.c
```diff
@@ -553,14 +564,45 @@ after_frag( fd_shred_ctx_t *    ctx,
...
+
+    uint last_idx = fd_disco_repair_shred_sig_last_shred_idx( sig );
+    fd_shred_t out_last_shred[1] = { 0 };
+    int rv = fd_fec_resolver_last_shred_query( ctx->resolver, shred_sig, last_idx, out_last_shred );
+    if( FD_UNLIKELY( rv != FD_FEC_RESOLVER_SHRED_OKAY ) ) {
+      FD_LOG_ERR(( "Shred tile %lu does not have the shred stored in fec_resolver", ctx->round_robin_id ));
```
ptaffet-jump, line 583: I think you need to be able to handle this more gracefully. We can brainstorm how. Otherwise, isn't it possible (although maybe a stretch) for the following situation to occur:
1. Shred tile gets a bunch of data shreds from repair responses.
2. Repair tile tries to force complete it, so it sends the frag
3. Tons and tons of shreds come in before the shred tile can read the frag
4. By the time the shred tile gets the frag, the FEC set has been spilled

src/disco/shred/fd_shred_tile.c
```diff
@@ -553,14 +564,45 @@ after_frag( fd_shred_ctx_t *    ctx,
...
+    }
+
+    fd_fec_set_t const * out_fec_set[1];
+    rv = fd_fec_resolver_force_complete( ctx->resolver, out_last_shred, out_fec_set );
+    if( rv != FD_FEC_RESOLVER_SHRED_COMPLETES )
+      FD_LOG_ERR(( "Shred tile %lu cannot force complete the slot %lu fec_set_idx %u", ctx->round_robin_id, out_last_shred->slot, out_last_shred->fec_set_idx ));
```
ptaffet-jump, line 589: I think this error is also worth thinking about. In what cases can it occur? The `if` should also have `FD_UNLIKELY`

src/disco/shred/fd_shred_tile.c
```diff
@@ -596,14 +637,24 @@ after_frag( fd_shred_ctx_t *    ctx,
...
 
-      if( FD_LIKELY( ctx->blockstore && rv==FD_FEC_RESOLVER_SHRED_OKAY ) ) { /* optimize for the compiler - branch predictor will still be correct */
-        uchar * buf = fd_chunk_to_laddr( ctx->replay_out_mem, ctx->replay_out_chunk );
-        ulong   sz  = fd_shred_header_sz( shred->variant );
-        fd_memcpy( buf, shred, sz );
-        ulong tspub       = fd_frag_meta_ts_comp( fd_tickcount() );
-        ulong replay_sig  = fd_disco_shred_replay_sig( shred->slot, shred->idx, shred->fec_set_idx, fd_shred_is_code( fd_shred_type( shred->variant ) ), 0 );
-        fd_stem_publish( stem, REPLAY_OUT_IDX, replay_sig, ctx->replay_out_chunk, sz, 0UL, ctx->tsorig, tspub );
-        ctx->replay_out_chunk = fd_dcache_compact_next( ctx->replay_out_chunk, sz, ctx->replay_out_chunk0, ctx->replay_out_wmark );
+      if( FD_LIKELY( ctx->repair_out_idx!=ULONG_MAX ) ) { /* firedancer topo compiler hint */
```
ptaffet-jump, line 640:
```suggestion
      if( FD_LIKELY( ctx->repair_out_idx!=ULONG_MAX ) ) { /* basically: if this isn't frankendancer */
```

ptaffet-jump: Or not sure exactly what your message in the comment meant

ptaffet-jump: "Optimize for full firedancer. The branch predictor will learn quickly if this is frankendancer instead"?

src/disco/shred/fd_shred_tile.c
```diff
@@ -623,10 +674,18 @@ after_frag( fd_shred_ctx_t *    ctx,
...
   fd_shred34_t * s34 = ctx->shred34 + 4UL*ctx->send_fec_set_idx;
 
-  s34[ 0 ].shred_cnt =                         fd_ulong_min( set->data_shred_cnt,   34UL );
-  s34[ 1 ].shred_cnt = set->data_shred_cnt   - fd_ulong_min( set->data_shred_cnt,   34UL );
-  s34[ 2 ].shred_cnt =                         fd_ulong_min( set->parity_shred_cnt, 34UL );
-  s34[ 3 ].shred_cnt = set->parity_shred_cnt - fd_ulong_min( set->parity_shred_cnt, 34UL );
+  ulong data_shred_cnt   = set->data_shred_cnt;
+  ulong parity_shred_cnt = set->parity_shred_cnt;
+  if( FD_UNLIKELY( data_shred_cnt == SHRED_CNT_NOT_SET || parity_shred_cnt == SHRED_CNT_NOT_SET ) ) {
+    FD_TEST( ctx->in_kind[ in_idx ]==IN_KIND_REPAIR ); // we MUST be in a force_completes situation
```
ptaffet-jump, line 680: The comment shouldn't be a `//` comment, and maybe it should go before the `if` or on the line right after by itself

ptaffet-jump (commented):
I just looked at the shred stuff. I can review the rest later, but here's enough to get you started.

src/disco/fd_disco_base.h
```diff
@@ -126,29 +136,101 @@ fd_disco_replay_old_sig( ulong slot,
...
+
+   There are two types of messages on the shred_repair link.  The first
+   type is a generic shred message. The second is a FEC set completion
+   message. Since we have run out of bits, the reciever must look at the
+   sz of the dcache entry to determine which type of message it is.
+
+   For the first message type (SHRED):
+
+   The first bit [63] describes whether this shred contains either a
+   batch complete or data complete flag.
```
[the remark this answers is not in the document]
emwang-jump: Sorry, will make it more clear. It means that if the shred completes a batch or slot, then this field is 1. If it's not completing anything, it's 0. Basically `shred.flags & (DATA_COMPLETE | SLOT_COMPLETE)`

src/disco/fd_disco_base.h
```diff
@@ -126,29 +136,101 @@ fd_disco_replay_old_sig( ulong slot,
...
+   worst case there is an independent FEC set for every shred, which
+   results in at most 2^15 FEC sets per slot.
+
+   The next bit [15] describes whether it is a coding shred (is_code).
+   If is_code = 0, the sig describes a data shred, and the last 15 bits
+   [0, 14] encode the shred_idx.  If is_code = 1, the sig describes a
+   coding shred, and the last 15 bits encode the data_cnt.
+
+   The frag will contain the full shred header of the last data shred in
+   the FEC set.
```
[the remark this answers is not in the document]
emwang-jump: No you're right, this message is sent on each shred, and the frag is the header for that shred. Think this comment was an artifact of a previous iteration on the shred msg

src/app/firedancer/topology.c
```diff
@@ -169,12 +169,13 @@ fd_topo_initialize( config_t * config ) {
...
   fd_topob_wksp( topo, "net_voter"  );
 
+  fd_topob_wksp( topo, "repair_shred" );
   fd_topob_wksp( topo, "quic_verify"  );
   fd_topob_wksp( topo, "verify_dedup" );
   fd_topob_wksp( topo, "dedup_pack"   );
 
   fd_topob_wksp( topo, "shred_storei" );
-  fd_topob_wksp( topo, "shred_replay" );
+  fd_topob_wksp( topo, "shred_repair" );
```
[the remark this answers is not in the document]
emwang-jump: What do you mean by combining them? or do you mean repair publishes to one `repair_shred` link instead of n `repair_shred` links and the all shred tiles read that one link?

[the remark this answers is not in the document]
ptaffet-jump: Sorry, I meant keep separate links but in the same workspace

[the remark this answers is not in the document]
emwang-jump: ok this makes sense, i've condensed to use only `shred_repair` - we can also remove the repair_shred link entirely after fixed fec sets since the repair_shred link is used exclusively for force completing

src/disco/shred/fd_shred_tile.c
emwang-jump: I think in this case we would just warn, but not err, and shred tile would just end up restarting that FEC set

src/disco/shred/fd_shred_tile.c
```diff
@@ -576,13 +587,83 @@ after_frag( fd_shred_ctx_t *    ctx,
...
+    if( FD_UNLIKELY( fd_fec_resolver_done_contains( ctx->resolver, shred_sig ) ) ) {
+      /* This is a FEC completion message from the repair tile.  We need
+         to make sure that we don't force complete something that's just
+         been completed. */
+      FD_MCNT_INC( SHRED, FORCE_COMPLETE_FAILURE, 1UL );
+      return;
+    }
+
+    uint last_idx = fd_disco_repair_shred_sig_last_shred_idx( sig );
+    fd_shred_t out_last_shred[1] = { 0 };
```
ptaffet-jump, line 602: This only reserves enough room for the shred header!

src/disco/shred/fd_shred_tile.c
```diff
@@ -576,13 +587,83 @@ after_frag( fd_shred_ctx_t *    ctx,
...
+    int rv = fd_fec_resolver_shred_query( ctx->resolver, shred_sig, last_idx, out_last_shred );
+    if( FD_UNLIKELY( rv != FD_FEC_RESOLVER_SHRED_OKAY ) ) {
+      /* We will hit this case if FEC is no longer in done map or in the curr_map*/
```
ptaffet-jump, line 605: Also if the shred is the sentinel/inval, which is fine

src/disco/shred/fd_shred_tile.c
```diff
@@ -328,6 +330,15 @@ during_frag( fd_shred_ctx_t * ctx,
 
   ctx->tsorig = fd_frag_meta_ts_comp( fd_tickcount() );
 
+  if( FD_UNLIKELY( ctx->in_kind[ in_idx ]==IN_KIND_REPAIR ) ) {
+    if( FD_UNLIKELY( chunk<ctx->in[ in_idx ].chunk0 || chunk>ctx->in[ in_idx ].wmark ) )
+    FD_LOG_ERR(( "chunk %lu %lu corrupt, not in range [%lu,%lu]", chunk, sz,
+                ctx->in[ in_idx ].chunk0, ctx->in[ in_idx ].wmark ));
+
+    uchar const * dcache_entry = fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk );
+    fd_memcpy( ctx->shred_buffer, dcache_entry, sz );
```
ptaffet-jump, line 339: Do you want similar round-robin logic as in the `IN_KIND_NET` case, or does the repair tile handle which one it goes to?

src/disco/shred/fd_shred_tile.c
```diff
@@ -576,13 +587,83 @@ after_frag( fd_shred_ctx_t *    ctx,
...
+    }
+
+    uint last_idx = fd_disco_repair_shred_sig_last_shred_idx( sig );
+    fd_shred_t out_last_shred[1] = { 0 };
+    int rv = fd_fec_resolver_shred_query( ctx->resolver, shred_sig, last_idx, out_last_shred );
+    if( FD_UNLIKELY( rv != FD_FEC_RESOLVER_SHRED_OKAY ) ) {
+      /* We will hit this case if FEC is no longer in done map or in the curr_map*/
+      FD_MCNT_INC( SHRED, FORCE_COMPLETE_FAILURE, 1UL );
+      return;
+      /* FIXME: Is there a chance something will be freed so fast that the sig lives in free_list? */
```
ptaffet-jump, line 608: Make this comment a bit more clear. There's no need to hide it anymore. How about something like:
```There's something of a race condition here.  It's possible (but very unlikely) that between when the repair tile observed the FEC set needed to be force completed and now, the FEC set was completed or spilled and then so many additional FEC sets were completed that it fell off the end of the done list.  In that case `fd_fec_resolver_done_contains` would have returned false, but `fd_fec_resolver_shred_query` will not return OKAY, which means we'll end up in this block of code.  If the FEC set was competed, then there's nothing we need to do.  If it was spilled, then we'll need to re-repair all the shreds in the FEC set, but it's not fatal.```
(I'm assuming that last part is true)

File: src/disco/shred/fd_shred_tile.c [hunk 20]
```diff
@@ -576,13 +587,83 @@ after_frag( fd_shred_ctx_t *    ctx,
...
+      FD_MCNT_INC( SHRED, FORCE_COMPLETE_FAILURE, 1UL );
+      return;
+    }
+    FD_MCNT_INC( SHRED, FORCE_COMPLETE_SUCCESS, 1UL );
+    FD_TEST( ctx->fec_sets <= *out_fec_set );
+    ctx->send_fec_set_idx = (ulong)(*out_fec_set - ctx->fec_sets);
+    ctx->shredded_txn_cnt = 0UL;
+  }
+
+  if( FD_UNLIKELY( ctx->in_kind[ in_idx ]==IN_KIND_REPAIR ) ) {
```
ptaffet-jump, line 624: This block is a duplicate, right?

src/disco/shred/fd_shred_tile.c
```diff
@@ -678,10 +768,22 @@ after_frag( fd_shred_ctx_t *    ctx,
...
+  ulong parity_shred_cnt = set->parity_shred_cnt;
+  if( FD_UNLIKELY( data_shred_cnt == SHRED_CNT_NOT_SET || parity_shred_cnt == SHRED_CNT_NOT_SET ) ) {
+    /* At this point we are in a FD_FEC_RESOLVER_SHRED_COMPLETES, but
+       shred_cnts have not been set, implying we have yet to recieve any
+       coding shreds. Thus, we must have arrived here via a force
+       completes situation. */
+    FD_TEST( ctx->in_kind[ in_idx ]==IN_KIND_REPAIR );
+    data_shred_cnt   = fd_disco_repair_shred_sig_last_shred_idx( sig ) + 1;
+    parity_shred_cnt = 0UL;
+  }
```
ptaffet-jump, line 781: fd_fec_resolver_force_complete should handle this. Unless there's a strong reason otherwise, it should populate `set->data_shred_cnt` and `set->parity_shred_cnt`.

src/disco/shred/fd_shred_tile.c
```diff
@@ -698,31 +800,41 @@ after_frag( fd_shred_ctx_t *    ctx,
...
+     intentional: shreds are inserted into the blockstore before
+     notifying repair. This is because the replay tile is downstream of
+     repair, and replay assumes the shreds are already in the blockstore
+     when repair notifies it that the FEC set is complete, and we don't
+     know whether shred will finish inserting into blockstore first or
+     repair will finish validating the FEC set first. */
+
+    fd_shred_t const * last = (fd_shred_t const *)fd_type_pun_const( set->data_shreds[ data_shred_cnt - 1 ] );
+
+    /* Copy the last shred and merkle root of the FEC set into the frag. */
```
ptaffet-jump, line 829:
```suggestion
    /* Copy the last data shred's header and merkle root of the FEC set into the frag. */
```

src/ballet/bmtree/fd_bmtree.h
```diff
@@ -149,9 +149,10 @@ static uchar const fd_bmtree_node_prefix[32UL] __attribute__((aligned(32))) = "\
 struct __attribute__((packed)) fd_bmtree_node {
   uchar hash[ 32 ]; /* Last bytes may not be meaningful */
 };
-
 typedef struct fd_bmtree_node fd_bmtree_node_t;
 
+FD_STATIC_ASSERT( sizeof(fd_bmtree_node_t) == 32, update FD_SHRED_MERKLE_ROOT_SZ );
```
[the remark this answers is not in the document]
emwang-jump: unfortunately get a circular inclusion if I try to use `FD_SHRED_MERKLE_ROOT_SZ`, and I assume we want to keep that in `fd_shred.h`?

src/disco/shred/fd_shred_tile.c
emwang-jump: The repair tile handles the proper routing. We discussed doing one link similar to `IN_KIND_NET`, but the change seemed minor enough and we may not even keep this link after fixed FEC sets.

src/disco/topo/fd_topo.c
```diff
@@ -317,7 +317,7 @@ fd_topo_mem_sz_string( ulong sz, char out[static 24] ) {
 void
 fd_topo_print_log( int         stdout,
                    fd_topo_t * topo ) {
-  char message[ 16UL*4096UL ] = {0}; /* Same as FD_LOG_BUF_SZ */
+  char message[ 20UL*4096UL ] = {0}; /* Same as FD_LOG_BUF_SZ */
```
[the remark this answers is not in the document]
emwang-jump: oh yup changing this back

src/ballet/bmtree/fd_bmtree.h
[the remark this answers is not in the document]
ptaffet-jump: What about moving the assertion to fd_shred?

ptaffet-jump (dismissed):
Great, thanks! Please squash before merging, but I think that should not unstamp the PR.

ptaffet-jump (approved):
Restamping. Note, I didn't review the stuff in discof/repair because Charlie already has

PR #4716, opened 2025-04-04
fdctl: substitute {user} and {home} in ledger.accounts_path

nk4rter:
`{user}` and `{home}` variables are substituted in a lot of paths, but not `[ledger.accounts_path]` which can lead to some confusion. This PR adds the corresponding `replace` calls, and updates the documentation in default config.


mmcgee-jump (approved):

mmcgee-jump:
Nice! Thank you for the contribution

PR #4717, opened 2025-04-04
runtime: fixing global types decode/encode

ibhatt-jumptrading:
1. make it so global types decode supports nested local pointers
2. remove global types that don't need to be global
3. add global encodes

### DIFF
--- a/src/flamenco/types/fd_types_custom.c
+++ b/src/flamenco/types/fd_types_custom.c
@@ -35,6 +35,22 @@ fd_flamenco_txn_decode_footprint_inner( fd_bincode_decode_ctx_t * ctx, ulong * t
   return 0;
 }
 
+int FD_FN_UNUSED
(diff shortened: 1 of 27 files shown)


src/flamenco/types/fd_types_custom.c
```diff
@@ -35,6 +35,22 @@ fd_flamenco_txn_decode_footprint_inner( fd_bincode_decode_ctx_t * ctx, ulong * t
   return 0;
 }
 
+int FD_FN_UNUSED
+fd_flamenco_txn_encode_global( fd_flamenco_txn_t const * self,
+                               fd_bincode_encode_ctx_t * ctx ) {
+  (void)self;
+  (void)ctx;
+  FD_LOG_ERR(( "only exists for testing" ));
```
jumpsiegel, line 43: So, do we check this in?

jumpsiegel (approved):

PR #4718, opened 2025-04-04
flamenco: disable apply_cost_tracker_during_replay in offline-replay

kbhargava-jump:


topointon-jump (commented):
I don't think we should merge this otherwise offline will behave differently to live.

kbhargava-jump:
> I don't think we should merge this otherwise offline will behave differently to live.

alright, we should just remember to add the ledger in CI once its fixed and for now will run offline with this patch

PR #4724, opened 2025-04-07
quic: heap-based svc queue

akhinvasara-jumptrading:
This PR swaps the service queues for a single heap, where each connection can exist at most once. The earliest event that's after 'now' will have its corresponding connection serviced. This leaves the jumble of conn_service, conn_tx, etc unchanged.

This fixes some correctness issues by allowing for dynamic timeouts. For example, before this PR, a connection with future retransmits may not get scheduled for service. This change lets us set a timer for the earliest retx timer that has not yet fired.

Replaces https://github.com/firedancer-io/firedancer/pull/4643 for cleaner commit and branch history


src/waltz/quic/fd_quic_svc_q.h
```diff
@@ -0,0 +1,92 @@
...
+
+/* fd_quic_svc_cancel removes a connection from the service queue */
+void
+fd_quic_svc_cancel( fd_quic_svc_timers_t * timers,
+                    fd_quic_conn_t       * conn );
+
+/* fd_quic_svc_timers_next returns next event. If 'pop' is true,
+   the event is popped from the queue. It will be < now!
+   If pop is false, event won't be popped and may be in the future.
+   Returns NULL conn if queue empty. */
```
nbridge-jump, line 79: Might want to extend this comment to explain the return when events are enqueued in the future, none have triggered yet, but pop=true, since the return value of NULL doesn't imply nothing is queued in this case.

nbridge-jump (approved):

nbridge-jump (approved):

PR #4729, opened 2025-04-07
jsiegel/stubgen_enum: emitter was wrong for compact discriminants

jumpsiegel:


kbhargava-jump (approved):

mmcgee-jump:
commit message like `fd_types: emitter was wrong for compact discriminants` would be nice :)

Issue #4736, opened 2025-04-08
[BUG] Parameter Error in eval_bit_and Function in generate_filters.py


eternalsakura13:
**Severity**: Medium

**Description**:
In generate_filters.py, the eval_bit_and function contains a parameter error when calling the eval_ function:

```python
def eval_bit_and(filt, op1, op2, label_t, label_f):
    op1_type, op2_type = type(op1), type(op2)

    if op1_type is not tuple and op2_type is not tuple:
        # handle the case where both values are immediate
        raise("unsupported")

    elif op1_type is tuple and op2_type is not tuple:
        # eval op1 and do operation with op2 imm
        eval_(op1, filt, 0, 0)
        # ...

    elif op2_type is tuple and op1_type is not tuple:
        # eval op2 and do operation with op1 imm
        eval_(op2, 0, 0)
        # ...
```

In the `eval_(op2, 0, 0)` call, the second parameter should be filt (the list storing generated BPF instructions), not 0.

**Impact**:
When processing expressions like `(bit-and <expression> <immediate>)` or `(bit-and <immediate> <expression>)`, a runtime error will occur, interrupting the compilation process and preventing the generation of correct seccomp BPF rules.

**Fix Recommendation**:
Correct the parameter error and exception throwing method in the eval_bit_and function:

```python
def eval_bit_and(filt, op1, op2, label_t, label_f):
    op1_type, op2_type = type(op1), type(op2)

    if op1_type is not tuple and op2_type is not tuple:
        raise Exception("unsupported")

    elif op1_type is tuple and op2_type is not tuple:
        eval_(op1, filt, None, None)
        # ...

    elif op2_type is tuple and op1_type is not tuple:
        eval_(op2, filt, None, None)
        # ...
```

**Credit**: Nan from Zellic

ripatel-fd:
Great find, thank you. I removed the eval_bit_and logic from the seccomp compiler. We might bring it back if we actually need it, but removing helps prevent further regressions for now.

PR #4737, opened 2025-04-08
Fix incorrect parameter in `eval_bit_and` function when calling `eval_` in generate_filters.py

eternalsakura13:
**Description**  
In the `eval_bit_and` function within `generate_filters.py`, the second parameter for `eval_` was mistakenly set to `0` instead of `filt` in the branch where `op2` is a tuple but `op1` is not. This caused incorrect behavior (or a potential error) when generating BPF instructions, as `eval_` was not receiving the correct list reference to append instructions to.

```python
# Original (incorrect)
elif op2_type is tuple and op1_type is not tuple:
    eval_(op2, 0, 0)

# Fixed
elif op2_type is tuple and op1_type is not tuple:
    eval_(op2, filt, 0, 0)
```

### DIFF
--- a/contrib/codegen/generate_filters.py
+++ b/contrib/codegen/generate_filters.py
@@ -240,7 +240,7 @@ def eval_bit_and(filt, op1, op2, label_t, label_f):
 
     elif op2_type is tuple and op1_type is not tuple:
         # eval op2 and do operation with op1 imm
-        eval_(op2, 0, 0)
+        eval_(op2, filt, 0, 0)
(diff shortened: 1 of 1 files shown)


nan-zellic:
Closes #4736

two-heart (approved):
Thank you

ripatel-fd:
Thank you for the contribution. We're going to remove any unused paths of the sandbox code generator and bring them back together with unit tests.

Issue #4738, opened 2025-04-08
[BUG] Seccomp Policy Configuration Error in fd_batch_tile


eternalsakura13:
**Severity**: Medium

**Description**:
In the fd_batch_tile.seccomppolicy file, the policy for the readlink system call is incorrectly configured as:
```
readlink:
```
This configuration format is incorrect (with a colon at the end of the line and nothing after it), causing generate_filters.py to parse it as "having an expression but it's empty", ultimately resulting in readlink calls being denied rather than allowed as expected.

This can be confirmed from the generated BPF:
```
BPF_JUMP( BPF_JMP | BPF_JEQ | BPF_K, SYS_readlink, /* check_readlink */ 39, 0 ),
...
//  check_readlink:
//  RET_KILL_PROCESS:
BPF_STMT( BPF_RET | BPF_K, SECCOMP_RET_KILL_PROCESS ),
```

**Issue Details**:
In the produce_snapshot function, the program needs to call readlink("/proc/self/fd/<fd>") to get the actual path of the current snapshot file:
```c
char proc_filename[ FD_SNAPSHOT_DIR_MAX ];
char prev_filename[ FD_SNAPSHOT_DIR_MAX ];

snprintf( proc_filename, FD_SNAPSHOT_DIR_MAX, "/proc/self/fd/%d", is_incremental ? ctx->incremental_snapshot_fd : ctx->full_snapshot_fd );
long len = readlink( proc_filename, prev_filename, FD_SNAPSHOT_DIR_MAX );
if( FD_UNLIKELY( len<=0L ) ) {
  FD_LOG_ERR(( "Failed to readlink the snapshot file" ));
}
```

Due to the seccomp policy configuration error, the readlink call here will be intercepted by seccomp and trigger the "Failed to readlink the snapshot file" error.

**Impact**:
Any functionality requiring readlink will fail, especially in scenarios in the produce_snapshot function that depend on readlink("/proc/self/fd/...") to get the snapshot file path, causing the entire snapshot process to fail.

**Fix Recommendation**:
Change the incorrect configuration:
```
readlink:
```
to:
```
readlink
```
or, if parameter restrictions are needed:
```
readlink: ( condition expression )
```

**Credit**: Nan from Zellic

ripatel-fd:
Thank you for the report. Anything under src/discof is part of "full Firedancer" which is still in development. The sandbox policies for those components should be ignored, they don't work. This will be fixed eventually of course.

[...]

PR #4739, opened 2025-04-08
fix(seccomp): remove trailing colon in readlink policy to allow readlink calls

eternalsakura13:
In fd_batch_tile.seccomppolicy, the readlink: entry is incorrectly configured with a trailing colon and no follow-up expression. This causes generate_filters.py to treat the line as having an empty expression, ultimately denying readlink system calls rather than allowing them.
```
# Original (incorrect)
readlink:

# Correct (remove or properly define the policy)
readlink
# or
readlink: (conditional expressions if needed)
```
As a result, any function—such as produce_snapshot—that relies on readlink("/proc/self/fd/...") will fail, causing the snapshot process to abort with a “Failed to readlink the snapshot file” error.

### DIFF
--- a/src/discof/batch/fd_batch_tile.seccomppolicy
+++ b/src/discof/batch/fd_batch_tile.seccomppolicy
@@ -72,4 +72,4 @@ read: (or (eq (arg 0) tmp_fd)
           (eq (arg 0) tmp_inc_fd))
 
 # snapshot
-readlink:
+readlink


nan-zellic:
Closes #4738

mmcgee-jump:
Thank you for the contribution! Nice find!

ripatel-fd:
Thank you for the contribution. We deleted the code you fixed by now, so I'm closing this.

PR #4741, opened 2025-04-08
feat(repair): implement blk repair structure and API

lidatong:
Blk repair is an API for repairing blocks as they are discovered from
the cluster via Turbine or Gossip.  Shreds (from Turbine) and votes
(from Gossip) inform blk repair that a block with the given slot they
are associated with exists.  Blk repair ensures that this block is
received in its entirety by requesting repairs for missing shreds for
the block.

Like other fork-aware structures, blk repair maintains a tree that
records the ancestry of slots.  It also maintains a frontier, which
models the leaves of the tree ie. the oldest (in ancestry) blocks
that still need to be repaired (across multiple forks).

Blk repair constructs the ancestry tree backwards, and then repairs
the tree forwards (using BFS).


src/discof/repair/fd_blk_repair.c
```diff
@@ -0,0 +1,486 @@
...
+/* link ele to the tree via its sibling. */
+
+static void
+link_sibling( fd_blk_repair_t * blk_repair, fd_blk_ele_t * sibling, fd_blk_ele_t * ele ) {
+  fd_blk_ele_t * pool        = fd_blk_pool( blk_repair );
+  ulong          null        = fd_blk_pool_idx_null( pool );
+  while( FD_UNLIKELY( sibling->sibling != null ) ) {
+    sibling = fd_blk_pool_ele( pool, sibling->sibling );
+  }
+  while( FD_UNLIKELY( sibling->sibling != null )) sibling = fd_blk_pool_ele( pool, sibling->sibling );
```
emwang-jump, line 219: repeated twice?

Issue #4740, opened 2025-04-08
[BUG] Security Policy Failure Due to Multiple read Rules Overwriting in fd_rpcserv_tile


nan-zellic:
**Severity: High**

**Description**  
In the `fd_rpcserv_tile.seccomppolicy` file, the `read` system call is defined multiple times, causing the first-defined security rules to be entirely overwritten by later ones. Specifically, the first rule allows `read` on any file descriptor except for three specific ones, while the second rule permits `read` only on `blockstore_fd`. Because `generate_filters.py` processes system call rules using a dictionary-like structure, any subsequent definition of the same system call overwrites the previous definition. As a result, only the second rule remains in effect: `(eq (arg 0) blockstore_fd)`.

---

**Issue Details**  
`fd_rpcserv_tile.seccomppolicy` currently defines two `read` rules:

```bash
read: (not (or (eq (arg 0) 2)
               (eq (arg 0) logfile_fd)
               (eq (arg 0) rpcserv_socket_fd)))

# blockstore: read archival file
read: (eq (arg 0) blockstore_fd)
```

- The first rule’s intent is to allow reading from any file descriptor **except** descriptors 2, `logfile_fd`, and `rpcserv_socket_fd`.  
- The second rule’s intent is to allow reading from `blockstore_fd`.  

However, because the second rule overwrites the first, the final (and only) effective rule becomes:

```bash
read: (eq (arg 0) blockstore_fd)
```

```
//  check_read:
    /* load syscall argument 0 in accumulator */
    BPF_STMT( BPF_LD | BPF_W | BPF_ABS, offsetof(struct seccomp_data, args[0])),
    BPF_JUMP( BPF_JMP | BPF_JEQ | BPF_K, blockstore_fd, /* RET_ALLOW */ 17, /* RET_KILL_PROCESS */ 16 ),
```

---

**Impact**  
This overwriting issue significantly distorts the intended security boundary of `fd_rpcserv_tile` in two ways:

1. **Functional Impact**: Most `read` calls that should be allowed—on all file descriptors **other than** the three specifically disallowed—are now incorrectly **rejected**.  
2. **Security Impact**: If the first rule was intended to deny reads only on a small set of descriptors (2, `logfile_fd`, `rpcserv_socket_fd`), then overwriting it with the second rule results in a **much narrower allowance** (only `blockstore_fd` is allowed). This can cause unexpected behavior and potential failures in parts of the application that rely on reading from other file descriptors.

---

[...]

PR #4742, opened 2025-04-08
fix(seccomp): remove the narrower read rule to keep the broader intent

eternalsakura13:
**Description**  

**What’s the issue?**  
- In `fd_rpcserv_tile.seccomppolicy`, two `read` rules were defined:
  1. A broader rule allowing reads for all file descriptors except `2`, `logfile_fd`, and `rpcserv_socket_fd`.
  2. A narrower rule permitting reads only from `blockstore_fd`.
- Because `generate_filters.py` overwrites earlier rules with later ones, the second rule effectively replaced the first, causing all `read` operations on other FDs to be denied.

**Why it matters**  
- The first rule was meant to allow a wide range of valid `read` operations and only restrict a few specific descriptors.  
- Overwriting it with the second rule severely limited allowed `read` calls, breaking the intended security boundary and potentially hindering functionality.

**What changed?**  
- Removed the second `read` rule (that only allows `blockstore_fd`), ensuring the broader first rule remains effective.

**Impact**  
- With the narrower rule removed, the security policy reverts to allowing reads on all file descriptors except `2`, `logfile_fd`, and `rpcserv_socket_fd`.  
- This aligns with the original intent and should resolve issues where valid reads were previously denied.

### DIFF
--- a/src/discof/rpc/fd_rpcserv_tile.seccomppolicy
+++ b/src/discof/rpc/fd_rpcserv_tile.seccomppolicy
@@ -68,8 +68,5 @@ close: (not (or (eq (arg 0) 2)
 # arg 2 is the timeout.
 poll: (eq (arg 2) 0)
 
-# blockstore: read archival file
-read: (eq (arg 0) blockstore_fd)
-
 # blockstore: lseek archival file
 lseek: (eq (arg 0) blockstore_fd)


nan-zellic:
Closes #4740

ripatel-fd:
Thank you for the contribution. I'm closing this for now, since the rpcserv design and sandbox policy need a redesign overall. That code is currently not mature and not in a production code path.

ripatel-fd:
(Reopened since on second thought, this change is obviously good to go in as-is)

nan-zellic:
I am very glad that you like this fix.

PR #4744, opened 2025-04-08
fix(seccomp): remove the narrower read rule to keep the broader intent

ripatel-fd:
This is the same as PR #4742, but also checking in the changes to the generated code.

### DIFF
--- a/src/discof/rpc/fd_rpcserv_tile.seccomppolicy
+++ b/src/discof/rpc/fd_rpcserv_tile.seccomppolicy
@@ -68,8 +68,5 @@ close: (not (or (eq (arg 0) 2)
 # arg 2 is the timeout.
 poll: (eq (arg 2) 0)
 
-# blockstore: read archival file
-read: (eq (arg 0) blockstore_fd)
-
 # blockstore: lseek archival file
 lseek: (eq (arg 0) blockstore_fd)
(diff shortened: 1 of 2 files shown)


src/discof/rpc/fd_rpcserv_tile.seccomppolicy
```diff
@@ -68,8 +68,5 @@ close: (not (or (eq (arg 0) 2)
 # arg 2 is the timeout.
 poll: (eq (arg 2) 0)
 
-# blockstore: read archival file
```
mmcgee-jump, line 71: Don't you need to add `blockstore_fd` to the above `read` rule as well then?

mmcgee-jump (changes requested):
.

src/discof/rpc/fd_rpcserv_tile.seccomppolicy
eternalsakura13: If blockstore_fd is not one of the three disallowed file descriptors, then it’s already implicitly allowed by the first rule. In that case, you can simply keep the first rule.

https://github.com/firedancer-io/firedancer/issues/4740

ripatel-fd:
@asiegel-jt is reviving the rpcserv tile, so I'm closing this

PR #4782, opened 2025-04-11
flamenco, funk: remove tombstones

topointon-jump:
Now that we do not support snapshots, we can remove tombstones from Funk. This has several benefits: simplifying the LRU logic, saving space and simplifying the Funk logic.


topointon-jump:
Closing as https://github.com/firedancer-io/firedancer/pull/5581 addresses it

PR #4791, opened 2025-04-15
flamenco: support just-in-time snapshot download and validation

yufeng-jump:


src/app/firedancer/config/default.toml
```diff
@@ -13,6 +13,55 @@
...
+        #
+        # If you have a full snapshot pre-downloaded somewhere,
+        # say by a script, you may specify the path to that file in
+        # "snapshot".
+        #
+        # Two special values are recognized here.  One is "funk", and the
+        # other is "wksp:".  Setting either will cause Firedancer to pick
+        # up its state from Funk, rather than going through the full
+        # snapshot deserialization process.
+        #
```
topointon-jump, line 26: I don't think we should overload this config option. Instead we should have a seperate funk checkpoint config option.

src/app/firedancer/config/default.toml
```diff
@@ -13,6 +13,55 @@
...
+        # small chance that by the time the validator downloads the
+        # latest incremental snapshot, it no longer matches the full
+        # snapshot.  In such cases, the validator will terminate boot so
+        # the full snapshot may be re-downloaded.
+        #
+        # Only the HTTP protocol is supported at the moment for snapshot
+        # download.  So URLs in "incremental_url" should be of the format
+        # "http://..."
+        incremental = ""
+        incremental_url = ""
```
topointon-jump, line 63: Instead of having both of these, it might be cleaner to just distinguish based on the prefix (`file` vs `http`).

yufeng-jump: This was a request from @lidatong and I kind of agree that it's more robust to not have to distinguish based on string parsing.

src/app/firedancer/config/default.toml
yufeng-jump: We can do that in a followup PR.  This is simply documenting current implicit behavior.  Long term we'd have config options for booting from snapshot, booting from genesis, and booting from checkpoint.

yufeng-jump: I added a fixme.

topointon-jump (approved):

src/app/firedancer/topology.c
```diff
@@ -731,9 +732,55 @@ fd_topo_initialize( config_t * config ) {
       }
       strncpy( tile->replay.genesis, config->tiles.replay.genesis, sizeof(tile->replay.genesis) );
 
-      strncpy( tile->replay.incremental, config->tiles.replay.incremental, sizeof(tile->replay.incremental) );
+      uchar incremental_is_file, incremental_is_url;
```
ibhatt-jumptrading, line 735: nit: could we perhaps move all of the snapshot topo stuff to a helper

src/disco/topo/fd_topo.h
```diff
@@ -286,6 +286,9 @@ typedef struct {
 
       /* not specified in TOML */
```
ibhatt-jumptrading, line 288: nit: vertical align

src/discof/replay/fd_replay_tile.c
```diff
@@ -192,6 +192,8 @@ struct fd_replay_tile_ctx {
   char const * genesis;
   char const * incremental;
   char const * snapshot;
+  int incremental_src_type;
```
ibhatt-jumptrading, line 195: align

src/discof/replay/fd_replay_tile.c
```diff
@@ -2065,19 +2067,18 @@ read_snapshot( void *              _ctx,
        stake weights. After this, repair will kick off concurrently with loading
        the rest of the snapshots. */
 
-    /* TODO: enable snapshot verification for all 3 snapshot loads.
```
ibhatt-jumptrading, line 2068: Why was this comment deleted

ibhatt-jumptrading (approved):

PR #4545, opened 2025-03-17
flamenco: full-client rocksdb ingestion support

kbhargava-jump:

PR #4797, opened 2025-04-16
archiver: add rocksdb to playback tile

yhzhangjump:
This PR incorporates Kunal's #4545 into the separate topology for the `sim` command.

### DIFF
--- a/src/disco/archiver/fd_archiver_playback.c
+++ b/src/disco/archiver/fd_archiver_playback.c
@@ -14,6 +14,8 @@
 #include <linux/if_xdp.h>
 #include "generated/archiver_playback_seccomp.h"
 #include "../../disco/topo/fd_pod_format.h"
+#include "../../flamenco/runtime/fd_rocksdb.h"
+#include "../../discof/geyser/fd_replay_notif.h"
 /* The archiver playback tile consumes from the archive file, adds artificial delay
 to reproduce exactly the timing from the capture, and forwards these fragments to the
 receiver tiles (shred/quic/gossip/repair).
@@ -51,6 +53,16 @@ struct fd_archiver_playback_tile_ctx {
   fd_io_buffered_istream_t istream;
   uchar *                  istream_buf;
 
+  ulong                  use_rocksdb;
+  fd_rocksdb_t           rocksdb;
+  fd_rocksdb_root_iter_t rocksdb_root_iter;
+  fd_slot_meta_t         rocksdb_slot_meta;
+  ulong                  rocksdb_curr_idx;
+  ulong                  rocksdb_end_idx;
+  ulong                  rocksdb_end_slot;
+  uchar *                rocksdb_bank_hash;
+  ulong                  replay_end_slot;
+
   fd_archiver_playback_stats_t stats;
 
   double tick_per_ns;
@@ -59,6 +71,11 @@ struct fd_archiver_playback_tile_ctx {
   ulong now;
   ulong need_notify;
   ulong notified;
+
+  // Replay tile input
+  fd_wksp_t * replay_in_mem;
+  ulong       replay_in_chunk0;
+  ulong       replay_in_wmark;
 
   fd_archiver_playback_out_ctx_t out[ 32 ];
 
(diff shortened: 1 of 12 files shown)


kbhargava-jump (approved):

src/discof/replay/fd_replay_tile.c
```diff
@@ -330,6 +331,9 @@ struct fd_replay_tile_ctx {
 
   int blocked_on_mblock; /* Flag used for synchronizing on mblock boundaries. */
 
+  /* Replay end slot */
+  ulong replay_end_slot;
```
kbhargava-jump, line 335: we have this here right now due to shredcap replay not using playback right now

src/discof/store/fd_store.c
```diff
@@ -193,7 +193,7 @@ fd_store_slot_prepare( fd_store_t *   store,
   if( FD_UNLIKELY( !fd_uchar_extract_bit( parent_flags, FD_BLOCK_FLAG_PROCESSED ) ) ) {
     rc = FD_STORE_SLOT_PREPARE_NEED_PARENT_EXEC;
     // FD_LOG_WARNING(("NEED PARENT EXEC %lu %lu", slot, parent_slot));
-    if( FD_UNLIKELY( !fd_uchar_extract_bit( parent_flags, FD_BLOCK_FLAG_REPLAYING ) ) ) {
+    if( FD_UNLIKELY( !fd_uchar_extract_bit( flags, FD_BLOCK_FLAG_REPLAYING ) ) ) {
```
kbhargava-jump, line 196: what is the reason for this change?

src/disco/archiver/fd_archiver_playback.c
```diff
@@ -209,17 +255,190 @@ during_housekeeping( fd_archiver_playback_tile_ctx_t * ctx ) {
...
+        &err );
+      if( FD_UNLIKELY( err || vallen==0 ) ) {
+        FD_LOG_ERR(( "Failed at reading bank hash for slot%lu from rocksdb", slot ));
+      }
+      fd_bincode_decode_ctx_t decode = {
+        .data    = res,
+        .dataend = res + vallen
+      };
+      ulong total_sz = 0UL;
+      int decode_err = fd_frozen_hash_versioned_decode_footprint( &decode, &total_sz );
```
lidatong, line 297: why is all this happening in `during_frag` instead of `after_frag`?

src/disco/archiver/fd_archiver_playback.c
```diff
@@ -158,18 +179,42 @@ unprivileged_init( fd_topo_t *      topo,
   }
   ctx->valloc = fd_alloc_virtual( ctx->alloc );
 
-  /* Allocate output buffer */
-  ctx->istream_buf = fd_valloc_malloc( ctx->valloc, 4096, FD_ARCHIVE_PLAYBACK_BUFFER_SZ );
-  if( FD_UNLIKELY( !ctx->istream_buf ) ) {
-    FD_LOG_ERR(( "failed to allocate input buffer" ));
-  }
+  if( strncmp( tile->archiver.archiver_path, "rocksdb:", 8 )==0 ) {
```
lidatong, line 182: make this a separate parameter vs. doing this same hack that replay does

src/disco/archiver/fd_archiver_playback.c
```diff
@@ -133,9 +150,13 @@ privileged_init( fd_topo_t *      topo,
     FD_SCRATCH_ALLOC_APPEND( l, fd_alloc_align(), fd_alloc_footprint() );
     FD_SCRATCH_ALLOC_FINI( l, scratch_align() );
 
-    tile->archiver.archive_fd = open( tile->archiver.archiver_path, O_RDONLY | O_DIRECT, 0666 );
-    if ( FD_UNLIKELY( tile->archiver.archive_fd == -1 ) ) {
-      FD_LOG_ERR(( "failed to open archive file %s %d %d %s", tile->archiver.archiver_path, tile->archiver.archive_fd, errno, strerror(errno) ));
+    if( strncmp( tile->archiver.archiver_path, "rocksdb:", 8 )==0 ) {
```
lidatong, line 153: see below comment

lidatong (commented):
im not sure `if/else` the `sim` command to incorporate rocksdb stuff is the right approach vs. making a separate rocksdb topo

it's not really an archiver / playback pattern here. it's just reading in shreds from a db

can keep this pr as is for now to focus on the consensus testing effort, but should be migrated to a separate topo (should be straightforward to do given the established pattern for making the `sim` topo already)

src/discof/store/fd_store.c
yhzhangjump: Suppose `slot` is the last slot in rocksdb. The problem was that `slot` was first added to `pending_slots`, removed, and never added back so that the storei tile never sends `slot` to replay tile (i.e., `slot` will never be replayed). My change may not be correct, but I was trying to do something like add `slot` back when its parent has not been processed.

yhzhangjump:
> im not sure `if/else` the `sim` command to incorporate rocksdb stuff is the right approach vs. making a separate rocksdb topo
> 
> it's not really an archiver / playback pattern here. it's just reading in shreds from a db
> 
> can keep this pr as is for now to focus on the consensus testing effort, but should be migrated to a separate topo (should be straightforward to do given the established pattern for making the `sim` topo already)

Yes I agree that we can leave this PR here for now since Kunal can already use this code for his testing purposes. We can decide what's the best way of doing it later after doing more about testing consensus.

yhzhangjump:
This PR is obsolete and it's replaced by PR 5025.

PR #4802, opened 2025-04-17
flamenco, precompile: clean up precompile execution

mjain-jump:


0x0ece (dismissed):

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1749,6 +1749,10 @@ fd_runtime_pre_execute_check( fd_execute_txn_task_info_t * task_info,
                               general transaction execution
 
   */
+
+  /* Set up instr infos from the txn descriptor. No Agave equivalent to this function. */
+  fd_executor_setup_instr_infos_from_txn_instrs( txn_ctx );
```
cali-jumptrading, line 1754: Perhaps we can setup the instr infos along with the `txn accounts` in a helper function? Something like `fd_runtime_setup_txn_ctx`. I think this would make the logic look cleaner.

src/flamenco/runtime/program/fd_precompiles.h
```diff
@@ -2,49 +2,27 @@
...
 
 int
-fd_precompile_secp256k1_verify( fd_exec_txn_ctx_t *     txn_ctx,
-                                fd_txn_instr_t const *  instr );
-
-/* fd_precompile_secp256r1_execute is the svm instruction processing entrypoint
-   for the Secp256r1 precompile. It is a wrapper around fd_precompile_secp256r1_verify. */
-
-int
-fd_precompile_secp256r1_execute( fd_exec_instr_ctx_t * ctx );
```
cali-jumptrading, line 40: 🎉

src/flamenco/runtime/fd_executor.c
```diff
@@ -1080,14 +1087,23 @@ fd_execute_instr( fd_exec_txn_ctx_t * txn_ctx,
...
+        fd_log_collector_program_invoke( ctx );
+        instr_exec_result = native_prog_fn( ctx );
+      } else {
+        instr_exec_result = FD_EXECUTOR_INSTR_ERR_UNSUPPORTED_PROGRAM_ID;
+      }
     } else {
-      instr_exec_result = FD_EXECUTOR_INSTR_ERR_UNSUPPORTED_PROGRAM_ID;
+      /* Only execute the precompile if `move_precompile_verification_to_svm` feature is active */
+      fd_log_collector_program_invoke( ctx );
+      if( FD_FEATURE_ACTIVE( txn_ctx->slot, txn_ctx->features, move_precompile_verification_to_svm ) ) {
```
cali-jumptrading, line 1102: I get why this else case is needed but it also clutters the code. I would prefer adding this as
```
fd_exec_instr_fn_t native_prog_fn = fd_executor_lookup_native_precompile_program( &txn_ctx->accounts[ instr->program_id ] );
if( native_prog_fn ) {
  /* A helper function to execute precompiles would be nice here, so we can encapsulate the logic of checking the move_precompile_verification_to_svm feature and invoke the log collector and set the exec result */
} else {
  int err = fd_executor_lookup_native_program( &txn_ctx->accounts[ instr->program_id ], txn_ctx, &native_prog_fn );
  /* We could also have a hlper function to execute a native program here, encapsulating the logic of invoking the log collector, checking if the native prog fn is null to set the exec result correctly... etc. */
}

File: src/flamenco/runtime/fd_executor.c [hunk 4]
```diff
@@ -81,25 +81,25 @@ typedef struct fd_native_prog_info fd_native_prog_info_t;
...
 
-#define PERFECT_HASH( u ) (((MAP_PERFECT_HASH_C*(u))>>28)&0xFU)
+#define PERFECT_HASH( u ) (((MAP_PERFECT_HASH_C*(u))>>30)&0x3U)
 
 #define MAP_PERFECT_HASH_PP( a00,a01,a02,a03,a04,a05,a06,a07,a08,a09,a10,a11,a12,a13,a14,a15, \
                              a16,a17,a18,a19,a20,a21,a22,a23,a24,a25,a26,a27,a28,a29,a30,a31) \
-                                          PERFECT_HASH( (a08 | (a09<<8) | (a10<<16) | (a11<<24)) )
-#define MAP_PERFECT_HASH_R( ptr ) PERFECT_HASH( fd_uint_load_4( (uchar const *)ptr + 8UL ) )
+                                          PERFECT_HASH( (a00 | (a01<<8)) )
+#define MAP_PERFECT_HASH_R( ptr ) PERFECT_HASH( fd_uint_load_2( (uchar const *)ptr ) )
```
cali-jumptrading, line 98: I am not familiar with setting the values for MAP_PERFECT_HASH here... just curious, could you explain why you changed these values?

cali-jumptrading (commented):
This is a nice cleanup, thanks!

src/flamenco/runtime/fd_executor.c
mjain-jump: I shrunk the hash table which requires a new constant multiplier value (generated via a python script) and a slightly different hash function. Check out the docs in `fd_map_perfect.c` they explain everything more in depth

src/flamenco/runtime/fd_executor.c
mjain-jump: I agree, but once `move_precompile_verification_to_svm` is cleaned up (its already activated on all clusters), this logic will be a lot simpler because we can stick the precompile programs into the native program lookup map and condense this logic even further.

src/flamenco/runtime/fd_executor.c
cali-jumptrading: ohh I did not realize these numbers were generated. Thanks!

src/flamenco/runtime/fd_executor.c
cali-jumptrading: Ah I see... I would still prefer encapsulating some of the logic execute a native program into a helper function. I'm ok leaving the precompile program execution the way it is because it'll be removed at some point anyways, like you mentioned.

src/flamenco/runtime/fd_executor.c
```diff
@@ -1065,28 +1075,42 @@ fd_execute_instr( fd_exec_txn_ctx_t * txn_ctx,
...
-      fd_exec_txn_ctx_reset_return_data( txn_ctx );
+    if( FD_UNLIKELY( err ) ) {
+      FD_TXN_PREPARE_ERR_OVERWRITE( txn_ctx );
+      FD_TXN_ERR_FOR_LOG_INSTR( txn_ctx, err, txn_ctx->instr_err_idx );
+      return err;
     }
 
-    if( FD_LIKELY( native_prog_fn != NULL ) ) {
-      /* Log program invokation (internally caches program_id base58) */
+    if( FD_LIKELY( native_prog_fn!=NULL ) ) {
```
cali-jumptrading, line 1093: 🎉 nice I like this approach

src/flamenco/runtime/fd_executor.h
```diff
@@ -41,15 +41,6 @@ get_transaction_account_lock_limit( fd_exec_txn_ctx_t const * txn_ctx ) {
 
 typedef int (* fd_exec_instr_fn_t)( fd_exec_instr_ctx_t * ctx );
 
-/* fd_executor_lookup_native_program returns the appropriate instruction
```
cali-jumptrading, line 44: only thing I have is we should probably copy the header comments for this function and paste it above its definition in `fd_executor.c` so we keep the comments, also adjust it for the modified signature.

mjain-jump: fixed

cali-jumptrading (approved):

PR #4820, opened 2025-04-21
Fix: update fd_log.c to improve logging behavior

### LINKED ISSUE #4705: Wrong version number in log file
The default log file path starts with `/tmp/fd-0.0.0-`

I guess we never noticed, but the 0.0.0 refers to a version number declared in fd_util_base.h.
Should be removed and replaced with fdctl's `version.h`.

aannnguyen:
- Ensured consistency with other logging components
Fix #4705

### DIFF
--- a/src/util/log/fd_log.c
+++ b/src/util/log/fd_log.c
@@ -21,7 +21,7 @@
 #define _GNU_SOURCE
 
 #include "fd_log.h"
-
+#include "../../app/fdctl/version.h"
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdarg.h>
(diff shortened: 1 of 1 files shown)


hongancalif:
@ripatel-fd Please help me review this, thanks

ripatel-fd:
Hello @aannnguyen, thank you for the contribution. Unfortunately the fix isn't quite as simple. fd_log is part of the base utilities, so the lowest layer of the dependency tree. fdctl is the root of the dependency tree. Having the base utils depend on the root would be a circular dependency and probably cause failures. I'm going to close this for now as we plan to fix this soon.

hongancalif:
Hi @ripatel-fd, thanks for your reply. I just wondering if there is any other issue that need help at the moment?

PR #4853, opened 2025-04-23
funk: clean up join API

### LINKED ISSUE #4847: [Snapshot Performance] Improve funk join pattern

ripatel-fd:
Adds a 'join' struct containing local address space pointers to
sub objects.  Improves insert performance by about 20ns per record.

Reverts 'ele_max' to ulong to avoid integer truncation issues.

Closes #4847

### DIFF
--- a/src/app/ledger/main.c
+++ b/src/app/ledger/main.c
@@ -22,7 +22,7 @@ struct fd_ledger_args {
   fd_wksp_t *           status_cache_wksp;       /* wksp for status cache. */
   fd_blockstore_t       blockstore_ljoin;
   fd_blockstore_t *     blockstore;              /* blockstore for replay */
-  fd_funk_t *           funk;                    /* handle to funk */
+  fd_funk_t             funk[1];                 /* handle to funk */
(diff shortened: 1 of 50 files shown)


asiegel-jt:
I'm not sure about the value in this change. Accessing a relative pointer is about as fast as an absolute pointer, and the compiler should be optimizing this case correctly. You're making a very large change for a relatively trivial gain.

asiegel-jt:
The spirit behind this change makes sense from an aesthetic point of view though. It does make the API a bit less confusing. You need to fix the unit tests however.

ripatel-fd:
> I'm not sure about the value in this change. Accessing a relative pointer is about as fast as an absolute pointer, and the compiler should be optimizing this case correctly. You're making a very large change for a relatively trivial gain.

@asiegel-jt This change yields a significant performance improvement (about 20ns per record written to funk). The problem was that all these join external function calls stall the CPU's pipeline quite often. The join struct eliminates those join calls.

jumpsiegel:
I think alex said he would approve once the tests pass

topointon-jump (approved):

PR #4854, opened 2025-04-23
funk: change lock in rec_hard_remove

cali-jumptrading:


src/funk/fd_funk_rec.c
```diff
@@ -329,7 +329,7 @@ fd_funk_rec_hard_remove( fd_funk_t *               funk,
   }
   fd_funk_rec_key_copy( pair->key, key );
 
-  fd_funk_rec_pool_lock( &rec_pool, 1 );
+  while( FD_ATOMIC_CAS( &txn->lock, 0, 1 ) ) FD_SPIN_PAUSE();
```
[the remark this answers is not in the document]
cali-jumptrading: yeah... forgot that txn can be null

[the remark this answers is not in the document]
cali-jumptrading: should be fine now!

src/funk/fd_funk_rec.c
```diff
@@ -331,7 +331,14 @@ fd_funk_rec_hard_remove( fd_funk_t *               funk,
...
 
-  fd_funk_rec_pool_lock( &rec_pool, 1 );
+  uchar * lock = NULL;
+  if( txn==NULL ) {
+    lock = &funk->lock;
+  } else {
+    lock = &txn->lock;
+  }
+
+  while( FD_ATOMIC_CAS( lock, 0, 1 ) ) FD_SPIN_PAUSE();
```
topointon-jump, line 341: we need to make sure the lock is released in all code paths from `fd_funk_rec_hard_remove`

cali-jumptrading: nice catch, done

src/funk/fd_funk_rec.c
```diff
@@ -331,15 +331,22 @@ fd_funk_rec_hard_remove( fd_funk_t *               funk,
...
```
topointon-jump, line 341: not a blocker but might be nice to have some convenience functions for getting and acquiring the lock. it's super useful in GDB to see that threads are stuck in a `lock_acquire` call, for example

topointon-jump (approved):

src/funk/fd_funk_rec.c
cali-jumptrading: we could make a convenient and even simpler utility like `fd_mutex.h` for this.

PR #4855, opened 2025-04-23
flamenco, funk: remove funk_rec_hard_remove

topointon-jump:
This is an anti-pattern, leads to a memory leak and should be removed.

Addresses https://github.com/firedancer-io/firedancer/issues/4856


topointon-jump:
Closing as this will be addressed by https://github.com/firedancer-io/firedancer/pull/5581

PR #4862, opened 2025-04-23
funk, util: Add mutex utility and change all instances of rec pool locks to mutex locks in the funk/txn object

cali-jumptrading:
Concurrency smoke test in progress

### DIFF
--- a/src/flamenco/runtime/fd_hashes.c
+++ b/src/flamenco/runtime/fd_hashes.c
@@ -941,6 +941,8 @@ fd_accounts_hash( fd_funk_t *             funk,
   fd_wksp_t * wksp            = fd_funk_wksp( funk );
   fd_funk_rec_pool_t rec_pool = fd_funk_rec_pool( funk, wksp );
   fd_funk_rec_pool_lock( &rec_pool, 1 );
+  FD_MUTEX_GUARD_BEGIN( &funk->mutex ) {
+
   fd_funk_txn_start_read( funk );
 
   if( fd_exec_para_cb_is_single_threaded( exec_para_ctx ) ) {
@@ -950,9 +952,9 @@ fd_accounts_hash( fd_funk_t *             funk,
     if ( NULL != lt_hash )
       fd_lthash_zero( &lthash_values[0] );
 
-    fd_pubkey_hash_pair_t * pairs             = fd_spad_alloc( runtime_spad,
-                                                               FD_PUBKEY_HASH_PAIR_ALIGN,
-                                                               funk->rec_max * sizeof(fd_pubkey_hash_pair_t) );
+    fd_pubkey_hash_pair_t * pairs = fd_spad_alloc( runtime_spad,
+                                                   FD_PUBKEY_HASH_PAIR_ALIGN,
+                                                   funk->rec_max * sizeof(fd_pubkey_hash_pair_t) );
 
     fd_accounts_sorted_subrange_gather( funk, 0, 1, &num_pairs, lthash_values, pairs, features );
     if( FD_UNLIKELY( !pairs ) ) {
@@ -966,7 +968,7 @@ fd_accounts_hash( fd_funk_t *             funk,
 
   } else {
     /* First calculate how big the list needs to be sized out to be, bump
-       allocate the size of the array then caclulate the hash. */
+      allocate the size of the array then caclulate the hash. */
 
     fd_subrange_task_info_t task_info = {
       .features      = features,
@@ -992,8 +994,10 @@ fd_accounts_hash( fd_funk_t *             funk,
     FD_LOG_NOTICE(( "accounts_hash %s", FD_BASE58_ENC_32_ALLOCA( accounts_hash->hash ) ));
   }
 
-  fd_funk_rec_pool_unlock( &rec_pool );
   fd_funk_txn_end_read( funk );
+
+  } FD_MUTEX_GUARD_END;
+  fd_funk_rec_pool_unlock( &rec_pool );
 
   return 0;
 }
--- a/src/funk/fd_funk.h
+++ b/src/funk/fd_funk.h
@@ -252,8 +252,15 @@ struct __attribute__((aligned(FD_FUNK_ALIGN))) fd_funk_private {
      the cgroup_idx to give the funk), inferring the wksp, cgroup from
      that and allocating exclusively from that? */
 
-  ulong alloc_gaddr; /* Non-zero wksp gaddr with tag wksp tag */
-  uchar lock;        /* lock for synchronizing modifications to funk object */
+  ulong      alloc_gaddr; /* Non-zero wksp gaddr with tag wksp tag */
+  fd_mutex_t mutex;       /* IMPORTANT: Protects the following fields:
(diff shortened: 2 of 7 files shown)


src/flamenco/runtime/fd_hashes.c
```diff
@@ -938,9 +938,8 @@ fd_accounts_hash( fd_funk_t *             funk,
   /* FIXME: this is not the correct lock to use, although in reality this is fine as we never modify
      accounts at the same time as hashing them. Once the hashing has been moved into tiles, we need to
      change it so that we partition by hash chains, and acquire the lock on the individual hash chains. */
-  fd_wksp_t * wksp            = fd_funk_wksp( funk );
-  fd_funk_rec_pool_t rec_pool = fd_funk_rec_pool( funk, wksp );
-  fd_funk_rec_pool_lock( &rec_pool, 1 );
```
topointon-jump, line 943: We should keep this lock as well. Basically we want exclusive access to Funk and all it's objects inside this function. (This function will be removed soon, this is just to avoid race conditions in the meantime)

src/funk/fd_funk_txn.h
```diff
@@ -38,17 +39,22 @@ struct __attribute__((aligned(FD_FUNK_TXN_ALIGN))) fd_funk_txn_private {
...
+  uint              child_head_cidx;   /* "                              oldest   child      "                             childless */
+  uint              child_tail_cidx;   /* "                              youngest child                                    childless */
+  uint              sibling_prev_cidx; /* "                              older sibling                                     oldest sibling */
+  uint              sibling_next_cidx; /* "                              younger sibling                                   youngest sibling */
+  uint              stack_cidx;        /* Internal use by funk */
+  ulong             tag;               /* Internal use by funk */
+
+  uint              rec_head_idx;      /* Record map index of the first record, FD_FUNK_REC_IDX_NULL if none (from oldest to youngest) */
+  uint              rec_tail_idx;      /* "                       last          " */
+  fd_mutex_t        mutex;             /* Protects the following fields:
```
topointon-jump, line 52:
```suggestion
  fd_mutex_t        mutex;             /* IMPORTANT: Protects the following fields:
```

src/funk/fd_funk.h
```diff
@@ -252,8 +252,15 @@ struct __attribute__((aligned(FD_FUNK_ALIGN))) fd_funk_private {
      the cgroup_idx to give the funk), inferring the wksp, cgroup from
      that and allocating exclusively from that? */
 
-  ulong alloc_gaddr; /* Non-zero wksp gaddr with tag wksp tag */
-  uchar lock;        /* lock for synchronizing modifications to funk object */
+  ulong      alloc_gaddr; /* Non-zero wksp gaddr with tag wksp tag */
+  fd_mutex_t mutex;       /* Protects the following fields:
```
topointon-jump, line 256:
```suggestion
  fd_mutex_t mutex;       /* IMPORTANT: Protects the following fields:
```

src/funk/fd_funk_rec.h
```diff
@@ -53,6 +53,8 @@ struct __attribute__((aligned(FD_FUNK_REC_ALIGN))) fd_funk_rec {
   /* These fields are managed by funk.  TODO: Consider using record
      index compression here (much more debatable than in txn itself). */
 
+  /* NOTE: These fields are protected by either the mutex in the txn object
+     or the mutex in the funk object. */
```
topointon-jump, line 57: Can we document which mutex is used in which case (if we are in the root txn or not).

src/util/fd_mutex.h
```diff
@@ -0,0 +1,47 @@
...
+};
+typedef struct fd_mutex fd_mutex_t;
+
+FD_PROTOTYPES_BEGIN
+
+static inline void
+fd_mutex_lock( fd_mutex_t * mutex ) {
+#if FD_HAS_THREADS
+  for(;;) {
+    ushort value = mutex->lock;
```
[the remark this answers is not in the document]
cali-jumptrading: yeah nice catch

src/flamenco/runtime/fd_hashes.c
cali-jumptrading: Could we use the global funk lock for this purpose then? something like `fd_funk_start_write`?

src/funk/fd_funk_txn.h
cali-jumptrading: done

src/funk/fd_funk_rec.h
cali-jumptrading: done

src/flamenco/runtime/fd_hashes.c
cali-jumptrading: resolved

src/util/fd_mutex.h
```diff
@@ -0,0 +1,49 @@
...
+};
+typedef struct fd_mutex fd_mutex_t;
+
+FD_PROTOTYPES_BEGIN
+
+static inline void
+fd_mutex_lock( fd_mutex_t * mutex ) {
+#if FD_HAS_THREADS
+  for(;;) {
+    uchar value = mutex->lock;
```
yufeng-jump, line 19:
```suggestion
    uchar value = FD_VOLATILE_CONST(mutex->lock);
```

src/util/fd_mutex.h
```diff
@@ -0,0 +1,49 @@
...
+    FD_SPIN_PAUSE();
+  }
+#else
+  mutex->lock = 1;
+#endif
+}
+
+static inline void
+fd_mutex_unlock( fd_mutex_t * mutex ) {
+  FD_VOLATILE( mutex->lock ) = 0;
```
yufeng-jump, line 32:
```suggestion
  FD_COMPILER_MFENCE();
  FD_VOLATILE( mutex->lock ) = 0;
```

topointon-jump, line 32:
```suggestion
  FD_COMPILER_MFENCE();
  FD_VOLATILE( mutex->lock ) = 0;
```

We need a memory fence before the write to 0 here, so that other threads see the update to the critical section after they see that the lock is free. On x86 architectures this is fine as they have strong ordering guarantees but we should probably put this memory fence in to be safe. See `fd_rwlock.h` as an example.

cali-jumptrading: Done

cali-jumptrading: Done

src/util/fd_mutex.h
cali-jumptrading: done

ripatel-fd (changes requested):
I'm not a big fan of this, to be honest... The old code was more readable and has more obvious locking semantics. If we add a library it should probably have dedicated unit tests and ample documentation in which contexts this mutex is useful. 

Since mutex/rwlock are naive implementations of the concept, they behave quite badly in high-contention cases, or cases where a lock is taken for an extended amount of time. So they are not a very good 'generic'/reusable implementation of a mutex.

ripatel-fd:
No longer needed with new database design

PR #4864, opened 2025-04-23
fixes to get fd bench to boot

ravyu-jump:


src/discof/replay/fd_replay_tile.c
```diff
@@ -2367,6 +2370,33 @@ init_snapshot( fd_replay_tile_ctx_t * ctx,
 
   send_exec_epoch_msg( ctx, stem, ctx->slot_ctx );
 
+  /* Publish slot notifs */
```
topointon-jump, line 2373: Could we add a FIXME to ensure init_after_snapshot works in the genesis case in a follow-up PR?

src/flamenco/runtime/fd_hashes.c
```diff
@@ -649,6 +649,46 @@ fd_update_hash_bank_tpool( fd_exec_slot_ctx_t * slot_ctx,
                                         runtime_spad );
 }
 
+int
+fd_update_hash_bank(  fd_exec_slot_ctx_t * slot_ctx,
+                      fd_capture_ctx_t *   capture_ctx,
+                      fd_hash_t *          hash,
+                      ulong                signature_cnt,
+                      fd_spad_t *          runtime_spad ) {
```
topointon-jump, line 657: could we merge this with the tpool version above, so we don't have duplicate code

topointon-jump (approved):

PR #4869, opened 2025-04-24
snapshot: non-temporal memcpy

### LINKED ISSUE #4815: [Snapshot Performance] Investigate non-temporal memory copies
When loading a snapshot, a large amount of cycles and cache space is spent on copying around account data bytes. Account data is only used after the snapshot is already loaded, so it should be evicted from cache ASAP. 

Ideally, only account metadata and the funk index should be kept in caches.

Using _mm512_stream_si512 (non-temporal store) when moving account data bytes should streamline large (>4kB) account data copies, and reduce the amount of valuable metadata evicted from caches.

An example of this can be found in fd_pack:fd_pack_schedule_impl.

ripatel-fd:
Closes #4815

### DIFF
--- a/src/flamenco/snapshot/fd_snapshot_restore.c
+++ b/src/flamenco/snapshot/fd_snapshot_restore.c
@@ -10,6 +10,10 @@
 #include <string.h>      /* strncmp */
 #include <sys/random.h>  /* getrandom */
 
+#if FD_HAS_AVX512 || FD_HAS_AVX
+#include <immintrin.h>
+#endif
+
 /* Snapshot Restore Buffer Handling ***********************************/
 
 static void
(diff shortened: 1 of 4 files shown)


src/flamenco/snapshot/fd_snapshot_restore.c
```diff
@@ -582,6 +586,72 @@ fd_snapshot_read_account_hdr_chunk( fd_snapshot_restore_t * restore,
...
+  ulong head_sz   = dst_align - (ulong)dst;
+  fd_memcpy( dst, src, head_sz );
+  src += head_sz;
+  dst += head_sz;
+  sz  -= head_sz;
+
+  /* NT copy */
+  while( sz>=512 ) {
+#  if FD_HAS_AVX512
+    _mm512_stream_si512( (void *)( dst+  0UL ), _mm512_loadu_si512( (void const *)( src+  0UL ) ) );
```
asiegel-jt, line 612: It's a bit faster to do the stores a few instructions after the loads. The secret is interleaving. This keeps the memory pipeline full.

jumpsiegel: How deep is the speculative execution buffer?

PR #4884, opened 2025-04-25
agave: update to 2.2.12

ptaffet-jump:
I basically hit the same minor issues as Anway:

> Rebase to v2.2.7 was mostly clean, it needed a slight modification to the verify precompiles commit.


anwayde (approved):
LGTM. I did notice though that we use `solana_sdk::feature_set` in `firedancer_poh_recorder.rs` whereas most of the agave codebase uses `agave_feature_set` - not introduced in this PR but just wanted to note in case we should also use the latter.

PR #4885, opened 2025-04-25
Snapshot Loader 2.0

ripatel-fd:


src/discof/restore/fd_filerd_tile.c
```diff
@@ -0,0 +1,347 @@
...
+  return sizeof(fd_filerd_tile_t);
+}
+
+static void
+privileged_init( fd_topo_t *      topo,
+                 fd_topo_tile_t * tile ) {
+  fd_filerd_tile_t * ctx = fd_topo_obj_laddr( topo, tile->tile_obj_id );
+  fd_memset( ctx, 0, sizeof(fd_filerd_tile_t) );
+
+  if( FD_UNLIKELY( tile->in_cnt !=0UL ) ) FD_LOG_ERR(( "tile `" NAME "` has %lu ins, expected 0",  tile->in_cnt  ));
```
mmcgee-jump, line 39: These belong in `unprivileged_init`. `privileged_init` is only for the minimally scoped, auditable set of stuff that _must_ happen unsandboxed

src/discof/restore/fd_filerd_tile.c
```diff
@@ -0,0 +1,347 @@
...
+  if( FD_UNLIKELY( tile->out_cnt!=1UL ) ) FD_LOG_ERR(( "tile `" NAME "` has %lu outs, expected 1", tile->out_cnt ));
+
+  ctx->fd = open( tile->filerd.file_path, O_RDONLY|O_CLOEXEC );
+  if( FD_UNLIKELY( ctx->fd<0 ) ) FD_LOG_ERR(( "open() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
```
mmcgee-jump, line 43: This is going to be a very frequently hit operator facing error. Please add a long error message describing the problem when errno is ENOENT.

src/app/firedancer-dev/main.c
```diff
@@ -95,9 +95,7 @@ extern fd_topo_run_tile_t fd_tile_archiver_feeder;
 extern fd_topo_run_tile_t fd_tile_archiver_writer;
 extern fd_topo_run_tile_t fd_tile_archiver_playback;
 
-extern fd_topo_run_tile_t fd_tile_bencho;
-extern fd_topo_run_tile_t fd_tile_benchg;
-extern fd_topo_run_tile_t fd_tile_benchs;
+extern fd_topo_run_tile_t fd_tile_snapshot_restore_FileRd;
```
mmcgee-jump, line 98: Can you just name these the same as everything else? Pick something 6 characters, so it can be the same name across everywhere (tile code, GUI, monitor, ...)

mmcgee-jump: The hardest part is naming things :)

PR #4886, opened 2025-04-25
repair: refactor to remove callbacks

emwang-jump:
pushed directly to repair-replay-link

### DIFF
--- a/src/discof/repair/fd_repair_tile.c
+++ b/src/discof/repair/fd_repair_tile.c
@@ -12,6 +12,7 @@
 #include "../../disco/fd_disco.h"
 #include "../../disco/keyguard/fd_keyload.h"
 #include "../../disco/keyguard/fd_keyguard_client.h"
+#include "../../disco/keyguard/fd_keyguard.h"
 #include "../../disco/net/fd_net_tile.h"
 #include "../../disco/shred/fd_stake_ci.h"
 #include "../../disco/topo/fd_pod_format.h"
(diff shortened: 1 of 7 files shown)


src/discof/repair/fd_repair_tile.c
```diff
@@ -355,12 +356,265 @@ repair_shred_deliver( fd_shred_t const *            shred,
...
+  repair_signer( repair_tile_ctx, pong->signature.uc, pre_image, FD_PING_PRE_IMAGE_SZ, FD_KEYGUARD_SIGN_TYPE_SHA256_ED25519 );
+
+  fd_bincode_encode_ctx_t ctx;
+  uchar buf[1024];
+  ctx.data = buf;
+  ctx.dataend = buf + sizeof(buf);
+  FD_TEST(0 == fd_repair_protocol_encode(&protocol, &ctx));
+  ulong buflen = (ulong)((uchar*)ctx.data - buf);
+
+  repair_send_intake_packet( buf, buflen, peer_addr, self_ip4_addr, repair_tile_ctx );
```
lidatong, line 388: what i actually was thinking is this would just fill the packet inside a caller provided `buf` arg

```c
fd_repair_handle_ping( fd_repair_tile_ctx_t *  repair_tile_ctx,
                       fd_repair_t *                 glob,
                       fd_gossip_ping_t const *      ping,
                       fd_gossip_peer_addr_t const * peer_addr,
                       uint                          self_ip4_addr,
                       uchar * buf,
                       ulong * buf_sz ) {
```

and then the caller would be responsible for calling `repair_send_intake_packet` (which can probably be deleted and just directly call `send_packet`)

as opposed to invoking the function itself

src/discof/repair/fd_repair_tile.c
```diff
@@ -355,12 +356,265 @@ repair_shred_deliver( fd_shred_t const *            shred,
...
+      active->avg_reps++;
+      active->avg_lat += glob->now - val->when;
+    }
+
+    fd_shred_t const * shred = fd_shred_parse(msg, shredlen);
+
+    if( shred == NULL ) {
+      FD_LOG_WARNING(("invalid shread"));
+    } else {
+      repair_shred_deliver(shred, shredlen, src_addr, &val->id, repair_tile_ctx);
```
lidatong, line 460: similar to above, i think a cleaner API would be to not actually call `repair_shred_deliver`

(also we can probably clean this up. the mutation of active table should be managed by caller separately vs. inside this function. but this can happen in a later PR)

src/discof/repair/fd_repair_tile.c
```diff
@@ -355,12 +356,265 @@ repair_shred_deliver( fd_shred_t const *            shred,
...
+  repair_signer( repair_tile_ctx, sig.uc, buf, buflen, FD_KEYGUARD_SIGN_TYPE_ED25519 );
+
+  /* Reintroduce the signature */
+
+  buf    -= 64UL;
+  buflen += 64UL;
+  fd_memcpy( buf + 4U, &sig, 64U );
+
+  uint src_ip4_addr = 0U; /* unknown */
+  repair_send_intake_packet( buf, buflen, addr, src_ip4_addr, repair_tile_ctx );
```
lidatong, line 512: same comment as above

src/discof/repair/fd_repair_tile.c
```diff
@@ -434,6 +688,266 @@ during_frag( fd_repair_tile_ctx_t * ctx,
...
+  repair_signer( repair_tile_ctx, ping->signature.uc, pre_image, FD_PING_PRE_IMAGE_SZ, FD_KEYGUARD_SIGN_TYPE_SHA256_ED25519 );
+
+  fd_bincode_encode_ctx_t ctx;
+  uchar buf[1024];
+  ctx.data = buf;
+  ctx.dataend = buf + sizeof(buf);
+  FD_TEST(0 == fd_repair_response_encode(&gmsg, &ctx));
+  ulong buflen = (ulong)((uchar*)ctx.data - buf);
+
+  repair_send_serve_packet( buf, buflen, dst_addr, src_ip4_addr, repair_tile_ctx );
```
lidatong, line 717: same

src/discof/repair/fd_repair_tile.c
```diff
@@ -434,6 +688,266 @@ during_frag( fd_repair_tile_ctx_t * ctx,
...
+          glob->metrics.recv_serv_full_ping_table++;
+          return 0;
+        }
+        val = fd_pinged_table_insert( glob->pinged, peer_addr );
+        for ( ulong i = 0; i < FD_HASH_FOOTPRINT / sizeof(ulong); i++ )
+          val->token.ul[i] = fd_rng_ulong(glob->rng);
+      }
+      fd_hash_copy( &val->id, &header->sender );
+      val->good = 0;
+      fd_repair_send_ping( repair_tile_ctx, glob, peer_addr, self_ip4_addr, val );
```
lidatong, line 904: same, but dont worry about changing this function since we don't need it right now

lidatong (commented):
generally my idea here is to make fd_repair more library vs. "framework-y" by filling values that are returned to the caller

benefits include makes it easier to unit test, reduces nesting, separates concerns in the function

lidatong (approved):

PR #4888, opened 2025-04-28
Parallel quick sort

kbowers-jump:
By popular demand, started to install high performance parallel sorting algorithms.  This PR generalizes the existing dual pivot in-place quick sort to be NUMA aware cache oblivious data adaptive thread parallel. Like the other thread parallel unit tests, the parallel sort unit test will use as many tiles are given to the test via the command line.

Additionally:

- Minor reorganization of the sort template to follow the layout used by more recent parallel algorithm templates.  No external impact from a user point of view but should make it easier to add additional parallel sorting algos in the near future.

- Drive by documentation cleanups and replicated documentation elimination the sort template.


asiegel-jt (approved):
I approved, but I question the need for this template. We are moving away from tpool.

PR #4917, opened 2025-04-29
x25519: fix avx512 impl should reduce x1

0x0ece:
Good finding by https://hackmd.io/@kUwtgwrmSiarE1gdXx1Unw/HJsDxKE1ex

However, I don't believe the explanation is correct. The issue is not in the loop, but in the input that needs to be reduced.


src/ballet/ed25519/fd_x25519.c
```diff
@@ -48,7 +48,7 @@ fd_x25519_montgomery_ladder( fd_f25519_t *       x2,
   fd_f25519_set( x2, fd_f25519_one );
   fd_f25519_set( z2, fd_f25519_zero );
 
-  /* use fd_f25519_add to reduce x1 mod p. it's prob unnecessary but not worth the risk. */
```
0x0ece, line 51: It turns out this was necessary after all :)

src/ballet/ed25519/fd_x25519.c
```diff
@@ -155,7 +158,7 @@ fd_x25519_montgomery_ladder( fd_f25519_t *       x2,
     FD_R43X6_QUAD_LANE_SUB_FAST( P, P, 0,1,0,1, P, Q ); // D = x_3 - z_3,            P  = A  |B  |C    |D,     in u45|s44|u45|s44
     FD_R43X6_QUAD_PERMUTE      ( Q, 0,1,1,0, P );       // BB = B^2,                 P  = A  |B  |B    |A,     in u44|u44|u44|u44
     FD_R43X6_QUAD_MUL_FAST     ( P, P, Q );             // DA = D * A,               P  = AA |BB |CB   |DA,    in u62|u62|u62|u62
-    FD_R43X6_QUAD_FOLD_SIGNED  ( P, P );                // DA = D * A,               P  = AA |BB |CB   |DA,    in u44|u44|u44|u44
+    FD_R43X6_QUAD_FOLD_UNSIGNED( P, P );                // DA = D * A,               P  = AA |BB |CB   |DA,    in u44|u44|u44|u44
```
0x0ece, line 161: For the longest time I couldn't understand why I needed fold_signed after mul. That didn't make any sense, but we were failing one of the wycheproofs tests. Well, the issue was reducing the input. Once the input is reduced, the loop works fine. Bonus: we get a slight perf improvement :)

PR #4921, opened 2025-04-29
quic_tls_hs:  simplify data buf

akhinvasara-jumptrading:
Currently, the data buffer to manage outgoing tls hs data is circular. However,


nbridge-jump (approved):
LGMT

PR #4952, opened 2025-04-30
Ignore Agave C dependency code for DoubleDelete lint

ripatel-fd:


nlgripto:
doesn't this do the same thing? https://github.com/firedancer-io/firedancer/blob/main/contrib/codeql/nightly/filter.qll

nlgripto:
cc @two-heart it looks like we're still getting double delete alerts for agave files

PR #4953, opened 2025-04-30
fuzz, cpi: remove stubbed target for CPI syscalls

mjain-jump:


nlgripto:
https://github.com/firedancer-io/firedancer/blob/main/.github/workflows/clusterfuzz.yml#L55-L56 can you remove this if that's part of the plan

ravyu-jump (dismissed):

nlgripto:
there will also be required changes in solfuzz, so let's merge those first in order to avoid shipping a broken bundle

ravyu-jump (approved):

PR #4967, opened 2025-05-01
quic - fix tls_hs related segfault

nbridge-jump:
Traced segfault in syslog to this location. Ran overnight.

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -3993,9 +3993,10 @@ fd_quic_conn_free( fd_quic_t *      quic,
     }
   }
 
-  /* free tls-hs */
-  fd_quic_tls_hs_delete( conn->tls_hs );
   if( conn->tls_hs ) {
+    /* free tls-hs */
+    fd_quic_tls_hs_delete( conn->tls_hs );
+
     /* Remove the handshake from the cache before releasing it */
     fd_quic_tls_hs_cache_ele_remove( &state->hs_cache, conn->tls_hs, state->hs_pool);
     fd_quic_tls_hs_pool_ele_release( state->hs_pool, conn->tls_hs );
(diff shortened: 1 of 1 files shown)


ripatel-fd:
@akhinvasara-jumptrading Would be cool to add a regression test to exercise whatever caused the crash

src/waltz/quic/fd_quic.c
```diff
@@ -3993,9 +3993,10 @@ fd_quic_conn_free( fd_quic_t *      quic,
     }
   }
 
-  /* free tls-hs */
-  fd_quic_tls_hs_delete( conn->tls_hs );
   if( conn->tls_hs ) {
+    /* free tls-hs */
+    fd_quic_tls_hs_delete( conn->tls_hs );
```
akhinvasara-jumptrading, line 3998: tls_hs_delete already null-checks, but agreed doesn't hurt to check here
If caller always passes non-null, could remove that check from tls_hs_delete

akhinvasara-jumptrading (approved):

ripatel-fd:
@akhinvasara-jumptrading Is this related https://github.com/firedancer-io/firedancer/security/code-scanning/73 ?

src/waltz/quic/fd_quic.c
nbridge-jump: I'm not concerned about checking it twice in this case. Having the checks in both places is somewhat better from a security/robustness/future proofing perspective.

nbridge-jump:
> @akhinvasara-jumptrading Would be cool to add a regression test to exercise whatever caused the crash

I agree with this. We need to recreate it. IMHO it's likely occurring due to retransmissions.

akhinvasara-jumptrading:
> > @akhinvasara-jumptrading Would be cool to add a regression test to exercise whatever caused the crash
> 
> I agree with this. We need to recreate it. IMHO it's likely occurring due to retransmissions.

Agreed. My best guess right now is we send handshake done frame twice, so there are two pkt_meta for it. They both get acked/reclaimed, and the second time we segfault. Repeating from slack, but I flattened without guaranteeing imdempotence. 

Edit: This is quite unlikely, the state machine should prevent explicitly sending twice (retx is OK, but that would delete the first pkt_meta).

akhinvasara-jumptrading:
> @akhinvasara-jumptrading Is this related [Double delete](/firedancer-io/firedancer/security/code-scanning/73) ?

Topically related, yes. But p sure it's trippin, that's not a real issue. That delete() function it's worried about just no-ops rn anyway.

PR #4970, opened 2025-05-01
refactor(shred,storei): remove storei, replace dcache w/ wksp obj

lidatong:
This PR removes the `storei` tile from the firedancer topo.

Significantly, this requires replacing the `shred_store` dcache in the `shred` tile which houses the FEC sets. This was replaced with a new shared wksp object `fec_sets`. This conveniently allows `shred` and `repair` tiles to join the same shmem region, so repair can directly observe which FEC sets need repair (and request repairs for them). Currently, repair tile is maintaining an entirely separate copy of what the in-progress FEC sets are.


src/app/firedancer/topology.c
```diff
@@ -313,7 +311,6 @@ fd_topo_initialize( config_t * config ) {
 
   /**/                 fd_topob_link( topo, "stake_out",    "stake_out",    128UL,                                    40UL + 40200UL * 40UL,         1UL );
   /* See long comment in fd_shred.c for an explanation about the size of this dcache. */
```
mmcgee-jump, line 313: Also the comment

lidatong:
@ripatel-fd @mmcgee-jump can i get a re-review
@ptaffet-jump can you also take a look -- mainly fd_shred_tile and topo-related changes to that?

lidatong:
added description

two-heart (changes requested):

emwang-jump (dismissed):

PR #4975, opened 2025-05-01
types: remove redundant 'destroy' API

ripatel-fd:
These days, all bincode data structures are decoded into a single
contiguous memory region.  Therefore, 'destroy' doesn't do anything
except zero memory.


mmcgee-jump (approved):

cali-jumptrading (commented):
Isn't a memset not enough, because some types have pointers to dynamically allocated memory, like vectors?

ripatel-fd:
> Isn't a memset not enough, because some types have pointers to dynamically allocated memory, like vectors?

@cali-jumptrading The types decode API writes all output into the same contiguous memory region. This memory region is owned by the caller, not the types API itself. The types API currently doesn't free any memory (no calls to fd_spad_pop or fd_valloc_free). Therefore, this change doesn't have an impact with correct usage. 

The only meaningful difference is that this change removes an API to clean up dangling pointers before memory is freed. So, it removes a layer of defense against use-after-frees. Practically, we barely used destroy API calls at all, so I don't think it makes a difference in practice. Another way to achieve the same UAF defense is to zero memory when exiting a scratch/spad frame.

cali-jumptrading:
> > Isn't a memset not enough, because some types have pointers to dynamically allocated memory, like vectors?
> 
> @cali-jumptrading The types decode API writes all output into the same contiguous memory region. This memory region is owned by the caller, not the types API itself. The types API currently doesn't free any memory (no calls to fd_spad_pop or fd_valloc_free). Therefore, this change doesn't have an impact with correct usage.
> 
> The only meaningful difference is that this change removes an API to clean up dangling pointers before memory is freed. So, it removes a layer of defense against use-after-frees. Practically, we barely used destroy API calls at all, so I don't think it makes a difference in practice. Another way to achieve the same UAF defense is to zero memory when exiting a scratch/spad frame.

Oh makes sense now. Though I think the memsets feel a bit hacky... perhaps we could keep the destroy API but simplify it to memset the sizeof the type?

ripatel-fd:
> Oh makes sense now. Though I think the memsets feel a bit hacky... perhaps we could keep the destroy API but simplify it to memset the sizeof the type?

@cali-jumptrading The memsets are only symbolic, they don't serve any purpose. Semantically, some piece of logic declares that values cease to exist, and the underlying memory is now free. The content of the free memory is arbitrary- Since it's free. 

Whether the user writes `/* foo is now free */`, or `fd_foo_delete( foo )`, `foo = (foo_t){0}`, or `memset( foo, 0, sizeof(foo) )` doesn't really matter.

mmcgee-jump (approved):

PR #4976, opened 2025-05-01
config: add 'firedancer' to 'make all' target

ripatel-fd:


nlgripto (approved):

mmcgee-jump:
Add `firedancer-dev` also please

mmcgee-jump (changes requested):
.

ripatel-fd:
@mmcgee-jump firedancer-dev is already part of make all.

mmcgee-jump (approved):

PR #4990, opened 2025-05-02
[v0.4 backport] topo: fix workspace NUMA affinity

### LINKED ISSUE #4986: Wrong NUMA assignments
I realized than some workspaces were assigned to NUMA node 0 on 0.411.20121 even if I pinned no tiles on the corresponding CPUs.
This seems to be related to `found_lazy` initialized to 1 instead of 0 in `initialize_numa_assignments` causing the search for the proper tile to stop too early. On master it's still the case.
Tell me if you'd like a PR.

ripatel-fd:
Fixes an issue that results in workspace affinity to NUMA node 0
even if all user tiles are on a different NUMA node.

Backport of 0b8aa0e86d03ca23cb21384cb096a7151b56712b #4903 

Closes #4986


mmcgee-jump:
Backport to v0.5, v0.4 is closed for PRs :)

PR #5010, opened 2025-05-02
choreo: consensus unit tests

kbhargava-jump:


src/choreo/test_choreo.c
```diff
@@ -0,0 +1,1163 @@
...
+#include "../flamenco/runtime/program/fd_vote_program.h"
+#include "../flamenco/runtime/program/fd_vote_program.c"
+#include "../flamenco/runtime/fd_txn_account.c"
+#include "../funk/fd_funk_filemap.h"
+#include "../util/wksp/fd_wksp.h"
+
+
+fd_tower_t *
+make_tower(fd_wksp_t * wksp, size_t n, ...) {
+  void * tower_mem = fd_wksp_alloc_laddr( wksp, fd_tower_align(), fd_tower_footprint(), 6UL );
```
lidatong, line 13: make the mem passed in, so caller is responsible for both allocing and freeing. otherwise caller has to remember to free something they didn't allocate

alternatively you can just put the mem in global scope similar to how kevin does it here
https://github.com/firedancer-io/firedancer/blob/main/src/util/tmpl/test_treap.c#L58-L60

so you don't have to pass it around

src/choreo/test_choreo.c
```diff
@@ -0,0 +1,1163 @@
...
+  }
+  va_end(args);
+
+  return tower;
+}
+
+#define TOWER(wksp, ...) \
+  make_tower(wksp, \
+    (sizeof((ulong[]){__VA_ARGS__})/sizeof(ulong)), \
+    __VA_ARGS__)
```
lidatong, line 45: just put this all on one line

src/choreo/test_choreo.c
```diff
@@ -0,0 +1,1163 @@
...
+
+ulong
+add_tower_vote(fd_tower_t * tower, ulong slot) {
+  ulong root = fd_tower_vote(tower, slot);
+  return root;
+}
+
+#define ADD_TOWER_VOTE(tower, slot) \
+  add_tower_vote(tower, slot)
```
lidatong, line 54: this macro is prob unnecessary

src/choreo/tower/test_tower.c
```diff
@@ -81,10 +81,183 @@ test_tower_vote( void ) {
...
+}
+
+void
+test_tower_agave( void ) {
+  /* This function contains the 7 unit tests from tower_vote_state.rs in Agave. */
+#define INIT_EMPTY_TOWER \
+   void * tower_mem   = fd_wksp_alloc_laddr( wksp, fd_tower_align(), fd_tower_footprint(), 1UL ); \
+   fd_tower_t * tower = fd_tower_join( fd_tower_new( tower_mem ) ); \
+   ulong root         = ULONG_MAX; \
+   (void)root;
```
lidatong, line 115: redundant with the `TOWER` macro from earlier?

src/choreo/tower/test_tower.c
```diff
@@ -81,10 +81,183 @@ test_tower_vote( void ) {
...
+
+#define UPDATE_TOWER(s) \
+  do{ \
+    ulong new_root=fd_tower_vote( tower, s ); \
+    if( new_root!=ULONG_MAX ) root=new_root; \
+  } while(0);
```
lidatong, line 121: unnecessary, as with `ADD_TOWER_VOTE` above

src/choreo/test_choreo.c
```diff
@@ -0,0 +1,1163 @@
...
+[331233129, 331233223]
+                └── [331233228, 331233303]
+                                ├── [331233304, 331233307]
+                                                └── [331233310, 331233319]
+                                                                ├── [331233324]
+                                                                └── [331233320, 331233323]
+                                                                                └── [331233326, 331233435]
+                                                                                                └── [331233440, 331233470]
+                                └── [331233308]
+*/
```
lidatong, line 955: test case looks good, but can you also document what you're testing in this case?

src/choreo/test_choreo.c
```diff
@@ -0,0 +1,1163 @@
...
+  } \
+  fd_landed_vote_t * landed_votes_##slot_num##_##voter_num = mock_landed_votes( wksp, tower_height_##slot_num##_##voter_num, vote_tower_##slot_num##_##voter_num, (uint)vote_slot_idx_##slot_num##_##voter_num ); \
+  fd_funk_txn_xid_t xid_##slot_num##_##voter_num; \
+  xid_##slot_num##_##voter_num.ul[0] = xid_##slot_num##_##voter_num.ul[1] = slot_num; \
+  fd_funk_txn_map_query_t query_##slot_num##_##voter_num[1]; \
+  fd_funk_txn_map_t txn_map_##slot_num##_##voter_num = fd_funk_txn_map( funk, fd_funk_wksp( funk ) ); \
+  FD_TEST( fd_funk_txn_map_query_try( &txn_map_##slot_num##_##voter_num, &xid_##slot_num##_##voter_num, NULL, query_##slot_num##_##voter_num, 0 )==FD_MAP_SUCCESS ); \
+  FD_TEST( query_##slot_num##_##voter_num->ele ); \
+  fd_funk_txn_t * funk_txn_##slot_num##_##voter_num = query_##slot_num##_##voter_num->ele; \
+  set_vote_account_tower( wksp, funk, funk_txn_##slot_num##_##voter_num, &voter[voter_num], landed_votes_##slot_num##_##voter_num );
```
lidatong, line 194: i get what you're addressing my feedback about helper macros, but this one is hard to read. a cleaner way to do this is to write a generic conversion function `tower_to_vote_state`. you can see `fd_tower_from_vote_acc` to see how it converts the other way

and so you always setup everything in terms of `TOWER(1, 2, 3, 4)` etc. then you call that helper to get a `fd_vote_state_t`.

you'll still need a helper function to put it in `funk`

lidatong (commented):
test cases look like a good start, left some comments mainly around API design

src/choreo/test_choreo.c
yhzhangjump: I have transformed this part into function `test_vote_switch_check_testnet_forks`, which tests that a voter voted on a stale minority fork should be able to switch to a recent/majority fork. In addition to this function, I have also documented `test_vote_lockout_check`, `test_vote_threshold_check`, `test_vote_switch_check` and `test_vote_switch_check_demo_forks`. The last one is using the fork structure from your demo slides.

For unit tests replicated from Agave, I am not documenting carefully, but I indeed replicated the comments in Agave which should explain what the tests are doing.

src/choreo/test_choreo.c
```diff
@@ -0,0 +1,2820 @@
...
+    fd_tower_t * scratch = fd_wksp_alloc_laddr( wksp, fd_tower_align(), fd_tower_footprint(), 5UL );
+    FD_TEST( 0==fd_tower_threshold_check( towers[3], epoch, funk, query->ele, THRESHOLD_DEPTH+1, scratch ) );
+  }
+
+  FD_LOG_NOTICE(( "Pass Agave test check_vote_threshold_forks" ));
+  fd_funk_close_file( &funk_close_args );
+}
+
+void
+test_agave_check_vote_threshold_deep_and_shallow_below_threshold( fd_wksp_t * wksp ) {
```
kbhargava-jump: this test case can be removed. we do not care about deep and shallow thresholds as they are just used for logging in agave

yhzhangjump: Cool, I removed the shallow check in this function and renamed the function to `test_agave_check_vote_threshold_deep_below_threshold` since the `deep` part of the test is still useful -- it is testing with threshold depth 8.

src/choreo/test_choreo.c
```diff
@@ -0,0 +1,2845 @@
...
+    fork->slot       = FRONTIER; \
+    fork->lock       = 0; \
+    fork->end_idx    = UINT_MAX; \
+    fork->slot_ctx   = slot_ctx; \
+    if( FD_UNLIKELY( !fd_fork_frontier_ele_insert( forks->frontier, fork, forks->pool ) ) ) { \
+      FD_LOG_ERR( ( "Failed to insert frontier=%lu into forks", FRONTIER ) ); \
+    } \
+  } while(0);
+
+/* This update_ghost function is copied from fd_tower_tile.c */
```
kbhargava-jump, line 326: any reason we have to copy this over? will make it hard to maintain to copies of the same function

yhzhangjump: Previously, the function is defined in `fd_forks.c` and declared in `fd_forks.h`, so I can simply use this function by including `fd_forks.h` and then linking `fd_choreo` in `src/choreo/Local.mk`.

I'm not sure how to do that after this function is moved to a static function within `fd_tower_tile.c`. Given that it is static in `fd_tower_tile.c`, it cannot be linked into `test_choreo`.

yhzhang0128:
Please make sure to polish and merge this PR. It does some good testing for the current consensus code. There's another issue when calling the threshold check in `fd_tower.c`, which I have explained to Charlie.

ripatel-fd:
Closing stale PR (feel free to reopen)

PR #3916, opened 2025-01-09
flamenco: correct program_indices handling

kbhargava-jump:

PR #5026, opened 2025-05-05
flamenco: remove unused code when pushing instr stack

anarcheuz:
if the vector of account indices is empty then `fd_exec_txn_ctx_get_key_of_account_at_index` should fail first.

Ran a pass of local fuzzing didn't seem to trigger a mismatch.

### DIFF
--- a/src/flamenco/runtime/fd_executor.c
+++ b/src/flamenco/runtime/fd_executor.c
@@ -958,8 +958,7 @@ fd_instr_stack_push( fd_exec_txn_ctx_t *     txn_ctx,
   int err = fd_exec_txn_ctx_get_key_of_account_at_index( txn_ctx,
                                                          instr->program_id,
                                                          &program_id_pubkey );
-  if( FD_UNLIKELY( err ||
-                   !memcmp( program_id_pubkey->key, fd_solana_native_loader_id.key, sizeof(fd_pubkey_t) ) ) ) {
+  if( FD_UNLIKELY( err ) ) {
     return FD_EXECUTOR_INSTR_ERR_UNSUPPORTED_PROGRAM_ID;
   }
 


src/flamenco/runtime/fd_executor.c
```diff
@@ -958,8 +958,7 @@ fd_instr_stack_push( fd_exec_txn_ctx_t *     txn_ctx,
   int err = fd_exec_txn_ctx_get_key_of_account_at_index( txn_ctx,
                                                          instr->program_id,
                                                          &program_id_pubkey );
-  if( FD_UNLIKELY( err ||
-                   !memcmp( program_id_pubkey->key, fd_solana_native_loader_id.key, sizeof(fd_pubkey_t) ) ) ) {
```
mjain-jump, line 962: There was a really weird edge case that existed where if the invoked program was the native loader, Agave would fail out in here. The runtime code has changed quite a bit since then so I don't know if that's still the case. Do all the current test vectors still pass?
/

anarcheuz: I ran a fuzzer pass on it and it didn't seem to trigger a mismatch, I was thinking if the vector of account indices was empty then `fd_exec_txn_ctx_get_key_of_account_at_index` would also fail no?

src/flamenco/runtime/fd_executor.c
```diff
@@ -958,8 +958,7 @@ fd_instr_stack_push( fd_exec_txn_ctx_t *     txn_ctx,
...
+  if( FD_UNLIKELY( err ) ) {
```
topointon-jump, line 961: Are we sure this is correct? @kbhargava-jump

kbhargava-jump: looking into right now

kbhargava-jump: this is when i added it in: https://github.com/firedancer-io/firedancer/pull/3916 which shows the fixture. will talk to @mjain-jump on the account loading rewrite that has happened since then

anarcheuz:
yeah if the program id is the native loader, the list of executable programs is empty (not the list of accounts).

PR #5031, opened 2025-05-06
Update SECURITY.md

hongancalif:


ripatel-fd:
Thanks @aannnguyen, this is a good point.

@claudijd @nlgripto @two-heart Considering there are random other blockchains that use Firedancer's code now, we should probably ask third parties to report _any_ security vulnerability as a private security advisory, not a public issue. (Even those affecting code that isn't in production on Solana.). I couldn't care less about those projects, but we still have a duty to do responsible disclosure.

two-heart (approved):

ripatel-fd:
Oh, sorry @anwayde, I didn't realize the base branch wasn't main.

ripatel-fd:
Reverted this change on branch `ande/test`

PR #5047, opened 2025-05-07
firedancer-dev: replay a slot only after its parent's bank hash matches

yhzhangjump:


src/disco/archiver/fd_archiver_backtest.c
```diff
@@ -131,6 +117,82 @@ rocksdb_inspect( fd_archiver_backtest_tile_ctx_t * ctx,
...
+rocksdb_get_shred( fd_archiver_backtest_tile_ctx_t * ctx,
+                   ulong                           * out_sz ) {
+  if( ctx->rocksdb_curr_idx==ctx->rocksdb_end_idx ) {
+    if( FD_UNLIKELY( fd_rocksdb_root_iter_next( &ctx->rocksdb_root_iter, &ctx->rocksdb_slot_meta, ctx->valloc ) ) ) return NULL;
+    if( FD_UNLIKELY( fd_rocksdb_get_meta( &ctx->rocksdb, ctx->rocksdb_slot_meta.slot, &ctx->rocksdb_slot_meta, ctx->valloc ) ) ) return NULL;
+    ctx->rocksdb_curr_idx = 0;
+    ctx->rocksdb_end_idx  = ctx->rocksdb_slot_meta.received;
+  }
+  ulong slot                = ctx->rocksdb_slot_meta.slot;
+  rocksdb_iterator_t * iter = rocksdb_create_iterator_cf(ctx->rocksdb.db, ctx->rocksdb.ro, ctx->rocksdb.cf_handles[FD_ROCKSDB_CFIDX_DATA_SHRED]);
```
lidatong, line 131: this doesn't look right... you're creating an iterator for every shred every time

lidatong: it looks like you copy-pasted parts of the rocksdb code but then restructured a loop around the wrong part

can you change this to just copy `fd_rocksdb_import_block_blockstore`? see fd_rocksdb.c

lidatong: you'll need to pass in slot meta which you can grab `fd_rocksdb_root_iter_seek( &iter, &rocks_db, slot, &slot_meta, valloc )`

see the example code in `fd_blockstore_tool.c`

https://github.com/firedancer-io/firedancer/blob/main/src/flamenco/runtime/fd_blockstore_tool.c#L189-L199

lidatong (changes requested):

lidatong:
can you also rename everything to `backtest`? currently there's a `backtest` cmd but `rocksdb` wksp and `arch_b` tile

src/disco/archiver/fd_archiver_backtest.c
yhzhangjump: Sure, I just pushed again and call `rocksdb_create_iterator_cf` only once at the beginning.

This could improve performance, but it does not change the behavior. No matter whether we create the iterator once or many times, the `rocksdb_iter_seek` a few lines after this line will update the iterator to the right (slot, idx).

src/flamenco/runtime/tests/run_ledger_backtest.sh
```diff
@@ -149,7 +149,7 @@ echo "
      affinity = \"auto\"
      bank_tile_count = 1
      shred_tile_count = 4
-     exec_tile_count = 8
+     exec_tile_count = 4
```
yhzhangjump, line 152: We find an issue when running backtest with ledger `v208-zk-sdk-ledger-no-rent`. If we set `exec_tile_count = 8` here, the code will be trapped into an infinite loop here: https://github.com/firedancer-io/firedancer/blob/main/src/flamenco/runtime/program/fd_bpf_program_util.c#L459-L464

This is a separate issue which Kunal is looking at.

topointon-jump (commented):
Thanks for changing this 🙌

src/flamenco/runtime/tests/run_ledger_backtest.sh
lidatong: this is reproducible? this funk code was recently changed by @topointon-jump and @ripatel-fd

yhzhangjump: Yes, this happens deterministically on my box and Kunal's box. All the other 82 tests in `run_backtest_tests_all.sh` are fine.

yhzhangjump: Cool, this is solved and it is an issue in exec tiles -- Ishan will fix it.

lidatong (approved):

PR #5060, opened 2025-05-09
resurrect rpc service

asiegel-jt:


src/discof/batch/fd_batch_tile.c
```diff
@@ -66,7 +66,7 @@ FD_FN_PURE static inline ulong
 scratch_footprint( fd_topo_tile_t const * tile FD_PARAM_UNUSED ) {
   ulong l = FD_LAYOUT_INIT;
   l = FD_LAYOUT_APPEND( l, alignof(fd_snapshot_tile_ctx_t), sizeof(fd_snapshot_tile_ctx_t) );
-  l = FD_LAYOUT_APPEND( l, fd_spad_align(), fd_ulong_align_up( MEM_FOOTPRINT, fd_spad_align() ) );
+  l = FD_LAYOUT_APPEND( l, fd_spad_align(), fd_spad_footprint( MEM_FOOTPRINT ) );
```
jumpsiegel, line 69: richie made stuff like this configurable in the toml... Kind of annoying to pass it all the way through but might be worth while making it configurable in case we guess wrong...

PR #5064, opened 2025-05-09
Sender: Lands UDP votes

akhinvasara-jumptrading:
This PR cleans up the sender tile such that we can land votes. As of now, they're sent over UDP to the advertised udp vote port. This has been tested against jump's private test cluster. 

This is not the final state of sender...there is ongoing work to add quic support and clean up the data structures used. Landing this PR will help further code evolve around this plumbing and for continuous testing with vote=true. 

Will create a separate PR to rename everything from voter/sender to send, a verb consistent with the rest of the tiles


src/discof/poh/fd_pohi_tile.c
```diff
@@ -435,7 +435,7 @@ fd_pohi_unprivileged_init( fd_topo_t *      topo,
 }
 
 /* One tick, one microblock, and one leader update. */
-#define STEM_BURST (3UL)
+#define STEM_BURST (4UL)
```
mmcgee-jump, line 438: Why did this change? Was it wrong before?

src/app/firedancer/topology.c
```diff
@@ -220,7 +220,7 @@ fd_topo_initialize( config_t * config ) {
   fd_topob_wksp( topo, "net_gossip" );
   fd_topob_wksp( topo, "net_repair" );
   fd_topob_wksp( topo, "net_quic"   );
-  fd_topob_wksp( topo, "net_voter"  );
+  fd_topob_wksp( topo, "sender_net" );
```
mmcgee-jump, line 223: Just call this `net_sender` same as the others

src/app/firedancer/topology.c
```diff
@@ -262,10 +262,9 @@ fd_topo_initialize( config_t * config ) {
   fd_topob_wksp( topo, "bank_busy"    );
   fd_topob_wksp( topo, "root_slot"    );
   fd_topob_wksp( topo, "pack_replay"  );
-  fd_topob_wksp( topo, "replay_voter" );
-  fd_topob_wksp( topo, "gossip_voter" );
-  fd_topob_wksp( topo, "voter_gossip" );
-  fd_topob_wksp( topo, "voter_dedup"  );
+  fd_topob_wksp( topo, "replay_send"  );
+  fd_topob_wksp( topo, "gossip_send"  );
+  fd_topob_wksp( topo, "sender_spray" );
```
mmcgee-jump, line 267: What is "spray"? Looks like this is just votes to gossip?

src/discof/sender/fd_sender_tile.c
```diff
@@ -117,69 +101,69 @@ scratch_align( void ) {
...
+get_current_leader_tpu_vote_contact( fd_sender_tile_ctx_t *       ctx,
+                                     ulong                       poh_slot,
                                      fd_shred_dest_weighted_t ** out_dest ) {
-  ulong poh_slot = fd_fseq_query( ctx->poh_slot );
-  if( poh_slot==ULONG_MAX ) { return -1; }
 
   fd_epoch_leaders_t const * lsched = fd_stake_ci_get_lsched_for_slot( ctx->stake_ci, poh_slot );
-  if( FD_UNLIKELY( !lsched      ) ) { return -1; }
+  if( FD_UNLIKELY( !lsched      ) ) {
+    ctx->metrics.leader_sched_not_found++;
```
mmcgee-jump, line 153: Don't love logging metrics in this function. Just return an error code, and have the parent track the result codes in an enum is a little cleaner.

```<enum name="GetLeaderResult">
  <Success>
  <ErrNoScheduleForSlot>
  <ErrNoContactInfo>
</enum>```

File: src/discof/sender/fd_sender_tile.c [hunk 5]
```diff
@@ -117,69 +101,69 @@ scratch_align( void ) {
...
+    return -1;
+  }
 
   fd_pubkey_t const * slot_leader = fd_epoch_leaders_get( lsched, poh_slot );
-  if( FD_UNLIKELY( !slot_leader ) ) { return -1 ; } /* Count this as bad slot too */
+  if( FD_UNLIKELY( !slot_leader ) ) {
+    ctx->metrics.leader_not_found++;
```
mmcgee-jump, line 159: I believe per contract, this cannot fail so you can FD_LOG_ERR here and abort.

src/discof/sender/fd_sender_tile.c
```diff
@@ -117,69 +101,69 @@ scratch_align( void ) {
...
+    return -1;
+  }
 
   fd_shred_dest_t * sdest = fd_stake_ci_get_sdest_for_slot( ctx->stake_ci, poh_slot );
   fd_shred_dest_idx_t sdest_idx = fd_shred_dest_pubkey_to_idx( sdest, slot_leader );
   if( FD_UNLIKELY( sdest_idx==FD_SHRED_DEST_NO_DEST ) ) {
+    ctx->metrics.leader_contact_not_found++;
```
mmcgee-jump, line 166: I'd also check in this fn that the ip address and port are nonzero, otherwise it's an err (nonroutable) as well.

src/discof/sender/fd_sender_tile.c
```diff
@@ -273,50 +261,55 @@ after_frag( fd_sender_tile_ctx_t * ctx,
   (void)tspub;
   (void)stem;
 
-  if( FD_UNLIKELY( in_idx==ctx->contact_in_idx ) ) {
+
+  if( in_idx>=IN_LINK_CNT ) {
```
mmcgee-jump, line 265: Why would this ever happen? Assert that in_cnt<IN_LINK_CNT during initialization, if needed. But this IN_LINK_CNT thing is too rigid on the topology imo. Just have the tile accept any in_links as long and assign them in_kink like the others I think is more flexible.

src/discof/sender/fd_sender_tile.c
```diff
@@ -273,50 +261,55 @@ after_frag( fd_sender_tile_ctx_t * ctx,
...
+    FD_LOG_WARNING(( "in_idx %lu >= IN_LINK_CNT %lu", in_idx, IN_LINK_CNT ));
+    return;
+  }
+
+  fd_sender_link_in_t * in_link = &ctx->in_links[ in_idx ];
+  ulong                kind    = in_link->kind;
+
+  if( FD_UNLIKELY( kind==IN_KIND_GOSSIP ) ) {
     finalize_new_cluster_contact_info( ctx );
+    ctx->metrics.contact_update_indiv_cnt += ctx->new_dest_cnt;
```
mmcgee-jump, line 275: You do not need metrics for this, there are already standardized metrics on the links tracking packets etc.

src/discof/sender/fd_sender_tile.c
```diff
@@ -273,50 +261,55 @@ after_frag( fd_sender_tile_ctx_t * ctx,
...
-  if( FD_UNLIKELY( in_idx==ctx->replay_in_idx ) ) {
+  if( FD_UNLIKELY( kind==IN_KIND_REPLAY ) ) {
     fd_txn_p_t * txn = (fd_txn_p_t *)fd_type_pun(ctx->txn_buf);
 
     /* sign the txn */
     uchar * signature = txn->payload + TXN(txn)->signature_off;
     uchar * message   = txn->payload + TXN(txn)->message_off;
     ulong message_sz  = txn->payload_sz - TXN(txn)->message_off;
     fd_keyguard_client_sign( ctx->keyguard_client, signature, message, message_sz, FD_KEYGUARD_SIGN_TYPE_ED25519 );
+    ctx->metrics.txns_signed++;
```
mmcgee-jump, line 293: You don't need this either, the enum you can add (per above) for `get_current_leader_tpu_vote_contact` will have this count

src/discof/sender/fd_sender_tile.c
```diff
@@ -273,50 +261,55 @@ after_frag( fd_sender_tile_ctx_t * ctx,
...
-    ctx->dedup_out_chunk = fd_dcache_compact_next( ctx->dedup_out_chunk, txn->payload_sz, ctx->dedup_out_chunk0,
-        ctx->dedup_out_wmark );
+    /* send to gossip and dedup */
+    fd_sender_link_out_t * gossip_dedup_out = ctx->gossip_dedup_out;
+    uchar * msg_to_gossip = fd_chunk_to_laddr( gossip_dedup_out->mem, gossip_dedup_out->chunk );
+    fd_memcpy( msg_to_gossip, txn->payload, txn->payload_sz );
+    fd_stem_publish( stem, gossip_dedup_out->idx, 1UL, gossip_dedup_out->chunk, txn->payload_sz, 0UL, 0, 0 );
+    gossip_dedup_out->chunk = fd_dcache_compact_next( gossip_dedup_out->chunk, txn->payload_sz, gossip_dedup_out->chunk0,
+        gossip_dedup_out->wmark );
+    ctx->metrics.txns_sprayed++;
```
mmcgee-jump, line 312: Really don't like the term "sprayed" here, and you also don't need this metric, links already track the count of frags sent across them.

mmcgee-jump (changes requested):
Thanks! Some good improvements

src/discof/repair/fd_repair_tile.c
```diff
@@ -1439,6 +1444,13 @@ unprivileged_init( fd_topo_t *      topo,
 
   fd_repair_update_addr( ctx->repair, &ctx->repair_intake_addr, &ctx->repair_serve_addr );
 
+  /* TODO: this is a hack so replay generates votes.
+     Once we develop a better notion of 'caught up',
+     remove all uses of current_slot in this file */
+  ulong poh_slot_obj_id = fd_pod_query_ulong( topo->props, "poh_slot", ULONG_MAX );
+  FD_TEST( poh_slot_obj_id!=ULONG_MAX );
+  ctx->current_slot = fd_fseq_join( fd_topo_obj_laddr( topo, poh_slot_obj_id ) );
```
lidatong, line 1452: can you rename this to `ctx->first_turbine_slot`?

src/discof/repair/fd_repair_tile.c
```diff
@@ -920,6 +922,9 @@ after_frag( fd_repair_tile_ctx_t * ctx,
     }
 
     fd_shred_t * shred = (fd_shred_t *)fd_type_pun( ctx->buffer );
+    if( fd_fseq_query( ctx->current_slot ) == ULONG_MAX ) {
+      fd_fseq_update( ctx->current_slot, shred->slot );
+    }
```
lidatong, line 927: also leave a FIXME for the hack here

src/discof/repair/fd_repair_tile.c
```diff
@@ -1439,6 +1444,13 @@ unprivileged_init( fd_topo_t *      topo,
 
   fd_repair_update_addr( ctx->repair, &ctx->repair_intake_addr, &ctx->repair_serve_addr );
 
+  /* TODO: this is a hack so replay generates votes.
+     Once we develop a better notion of 'caught up',
+     remove all uses of current_slot in this file */
```
lidatong, line 1449:
```suggestion
  /* TODO: this is a hack to set the first turbine slot so that replay knows when we have caught up. This is important for voting, because you need have "read-back" your latest landed tower before you can vote. Using an fseq is a temporary hack, and this will be replaced with the standard frag-signaling pattern across tiles with a separate consensus tile. */
```

lidatong: (reformat to 72 cols)

src/app/firedancer/topology.c
akhinvasara-jumptrading: consistency > correctness ? 
We'd prefer for the others to also be swapped, right? But leaving as is to avoid UX problems? 
I'll add a comment explaining that. You still think we'd rather have them be consistent?

src/app/firedancer/topology.c
akhinvasara-jumptrading: and dedup! maybe others later
basically a general sender_out ... maybe that's a better name ?
'spray' meant to evoke this catch-all, non-targeted send

src/discof/poh/fd_pohi_tile.c
akhinvasara-jumptrading: good catch, thanks
was in prep for using msg passing instead of shared var for poh_slot, but ended up kicking that can down the road. Reverting for now.

src/discof/sender/fd_sender_tile.c
```diff
@@ -117,69 +101,69 @@ scratch_align( void ) {
...
+  fd_stem_publish( stem, net_out_link->idx, sig, net_out_link->chunk, packet_sz, 0UL, tsorig, tspub );
+  net_out_link->chunk = fd_dcache_compact_next( net_out_link->chunk, packet_sz, net_out_link->chunk0, net_out_link->wmark );
 }
 
+
 static int
-get_current_leader_tpu_vote_contact( fd_sender_tile_ctx_t *      ctx,
+get_current_leader_tpu_vote_contact( fd_sender_tile_ctx_t *       ctx,
+                                     ulong                       poh_slot,
                                      fd_shred_dest_weighted_t ** out_dest ) {
```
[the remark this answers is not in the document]
akhinvasara-jumptrading: yeah agreed, actually refactoring that whole stake_ci tracking stuff along with the quic changes so will just change that there
but yeah this is p 🤮

src/discof/sender/fd_sender_tile.c
akhinvasara-jumptrading: This will also change soon, so going to leave as is for now

src/discof/sender/fd_sender_tile.c
akhinvasara-jumptrading: this will also change with quic PR, and ill incorporate all this feedback in that

src/app/firedancer-dev/config/default.toml
```diff
@@ -32,7 +32,7 @@
         prometheus_listen_address = "0.0.0.0"
         prometheus_listen_port = 7999
 [consensus]
-    vote = false
+    vote = true
     expected_shred_version = 64475
```
jumpsiegel: Lets leave this default as false

src/app/firedancer-dev/config/private.toml
```diff
@@ -26,7 +26,7 @@
         prometheus_listen_address = "0.0.0.0"
         prometheus_listen_port = 7999
 [consensus]
-    vote = false
+    vote = true
     expected_shred_version = 20896
```
jumpsiegel: same here... i don't want to change the default to true yet

src/disco/metrics/generate/types.py
```diff
@@ -25,7 +25,8 @@ class Tile(Enum):
     GOSSIP = 19
     NETLNK = 20
     SOCK = 21,
-    REPAIR = 22
+    REPAIR = 22,
+    SENDER = 23
```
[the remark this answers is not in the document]
jumpsiegel: Can you make an issue out of this and lets do this in a different PR since we have branches off this branch so doing this rename will cause a merge disaster...

src/disco/metrics/metrics.xml
```diff
@@ -210,6 +210,13 @@ metric introduced.
     <counter name="PktRetransmissions" summary="Number of QUIC packets that retransmitted." />
 </tile>
 
+<tile name="sender">
```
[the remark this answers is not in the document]
jumpsiegel: we will do this rename as a different PR to avoid a downstream merge disaster with other branches off this branch

src/discof/replay/fd_replay_tile.c
```diff
@@ -1029,9 +1028,12 @@ send_tower_sync( fd_replay_tile_ctx_t * ctx ) {
...
   fd_hash_t vote_block_hash[1] = { 0 };
-  int err = fd_blockstore_bank_hash_query( ctx->blockstore, vote_slot, vote_bank_hash );
-  if( err ) FD_LOG_ERR(( "invariant violation: missing bank hash for tower vote" ));
-  err = fd_blockstore_block_hash_query( ctx->blockstore, vote_slot, vote_block_hash );
+
+  /* guaranteed to be on frontier from caller check */
+  fd_fork_t const * fork = fd_forks_query_const( ctx->forks, vote_slot );
+  fd_memcpy( vote_bank_hash, &fork->slot_ctx->slot_bank.banks_hash, sizeof(fd_hash_t) );
+
+  int err = fd_blockstore_block_hash_query( ctx->blockstore, vote_slot, vote_block_hash );
```
jumpsiegel, line 1036: How much does voting performance matter... ie, should we generate the vote as soon as we have the bank hash verses here?  ie, what is the cause/effect?  No change required here but lets think about changing the flow of operations...

jumpsiegel (changes requested):

src/disco/metrics/generate/types.py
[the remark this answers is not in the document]
ripatel-fd: @jumpsiegel sounds good

PR #5087, opened 2025-05-12
sysvar: always use fd_sysvar_slot_hashes_join

two-heart:
one branch reimplemented fd_sysvar_slot_hashes_join, which caused a CodeQL static analysis finding.


ripatel-fd:
cc @ibhatt-jumptrading

PR #5088, opened 2025-05-12
Unmap MAP_SHARED stacks on fork()

ripatel-fd:
Firedancer threads don't support fork(). If a user calls fork() regardless, wild memory corruption ensues, crashing both parent and child. This PR keeps the parent alive and crashes the child reliably.

- [x] Test regular `fddev`
- [x] Test `fddev --no-sandbox`
- [x] Test `fddev --no-sandbox --no-clone`
- [x] Test fd_tile APIs

### DIFF
--- a/src/disco/topo/fd_topo_run.c
+++ b/src/disco/topo/fd_topo_run.c
@@ -147,6 +147,8 @@ typedef struct {
   uint               gid;
   int *              done_futex;
   volatile int       copied;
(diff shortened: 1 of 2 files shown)


two-heart (dismissed):

two-heart (approved):

ripatel-fd:
Fyi @mmcgee-jump merging this

PR #5107, opened 2025-05-13
Use 4K page stacks for MSan

ripatel-fd:
Don't attempt to use MAP_HUGETLB stack memory on ASan and MSan. 
Fixes random ENOMEM failures with MSan when creating guard pages.

### DIFF
--- a/src/disco/topo/fd_topo_run.c
+++ b/src/disco/topo/fd_topo_run.c
@@ -184,8 +184,12 @@ fd_topo_tile_stack_join_anon( void ) {
   int   prot  = PROT_READ|PROT_WRITE;
   int   flags = MAP_PRIVATE|MAP_ANONYMOUS|MAP_STACK;
 
-  uchar * stack = mmap( NULL, sz, prot, flags|MAP_HUGETLB, -1, 0 );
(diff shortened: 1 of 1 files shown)


src/disco/topo/fd_topo_run.c
```diff
@@ -184,8 +184,12 @@ fd_topo_tile_stack_join_anon( void ) {
...
   int   flags = MAP_PRIVATE|MAP_ANONYMOUS|MAP_STACK;
 
-  uchar * stack = mmap( NULL, sz, prot, flags|MAP_HUGETLB, -1, 0 );
-  if( FD_UNLIKELY( stack==MAP_FAILED ) ) {
+  uchar * stack = NULL;
+#if !FD_HAS_ASAN && !FD_HAS_MSAN
+  stack = mmap( NULL, sz, prot, flags|MAP_HUGETLB, -1, 0 );
+#endif
+
+  if( !stack ) {
```
mmcgee-jump, line 192: Isn't this bugged? You should check `if( stack==MAP_FAILED )` and initialize it to MAP_FAILED instead

mmcgee-jump (changes requested):
.

mmcgee-jump (approved):

PR #5120, opened 2025-05-14
fuzz, vm: push instruction context before executing vm interp

mjain-jump:


src/flamenco/runtime/tests/harness/fd_vm_harness.c
```diff
@@ -363,6 +372,10 @@ do{
   *output = effects;
   fd_runtime_fuzz_instr_ctx_destroy( runner, instr_ctx );
   return actual_end - (ulong)output_buf;
+
+error:
```
topointon-jump, line 376: nit: I don't love this goto pattern - better to return from a function

topointon-jump (approved):

topointon-jump (approved):

PR #5122, opened 2025-05-14
sock - log & continue in the event of error returned from sendmmsg

nbridge-jump:
sendmmsg has a hard-to-determine list of transient errors that should not terminate the validator.
As such, this PR logs the errors and continues.


ripatel-fd (approved):

ripatel-fd:
Taking this over

PR #5129, opened 2025-05-15
vm: fix sol_log decl/def parameter name mismatch

two-heart:


ripatel-fd:
@two-heart Fixed in https://github.com/firedancer-io/firedancer/pull/5172

PR #5149, opened 2025-05-16
removed legacy transaction map

asiegel-jt:


jumpsiegel (approved):

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1835,10 +1561,6 @@ fd_runtime_finalize_txn( fd_exec_slot_ctx_t *         slot_ctx,
   fd_exec_txn_ctx_t * txn_ctx      = task_info->txn_ctx;
   int                 exec_txn_err = task_info->exec_res;
 
-  /* For ledgers that contain txn status, decode and write out for solcap */
-  if( capture_ctx != NULL && capture_ctx->capture && capture_ctx->capture_txns && slot_ctx->slot_bank.slot>=capture_ctx->solcap_start_slot ) {
-    fd_runtime_write_transaction_status( capture_ctx, slot_ctx, txn_ctx, exec_txn_err );
-  }
```
topointon-jump, line 1841: We need this functionality to debug BHM with solcap. Could we not remove this until we have it somewhere else?

topointon-jump (dismissed):

src/flamenco/runtime/fd_runtime.c
asiegel-jt: I don't think this feature works anyway. It depends on metadata being stored in the transaction map, which we generally don't do.

topointon-jump: We do when we run the ledger tests. These are now being moved to the `backtest` code path (which runs the ledger test blocks through the tile code), so we will need to make sure we write the transaction status here too.

cc @yhzhangjump @kbhargava-jump

topointon-jump (approved):

PR #5158, opened 2025-05-20
runtime: remove unused acc_mgr fields

ripatel-fd:
Remove 'magic' and 'hlen' fields from acc_mgr.


topointon-jump (commented):
Let's chat offline, I would prefer to keep `magic` for now.

ripatel-fd:
Revisiting later

PR #5160, opened 2025-05-21
Fix SIGABRT on unknown subcommand

ripatel-fd:


src/app/shared_dev/boot/fd_dev_boot.c
```diff
@@ -97,7 +97,10 @@ fd_dev_main( int          argc,
     }
   }
 
-  if( FD_UNLIKELY( !action ) ) FD_LOG_ERR(( "unknown subcommand `%s`", action_name ));
+  if( FD_UNLIKELY( !action ) ) {
```
mmcgee-jump, line 100: Why? This is not SIGABRT, FD_LOG_ERR is also just an exit(1) ?

mmcgee-jump (changes requested):
.

mmcgee-jump (approved):

PR #5182, opened 2025-05-23
Add 'make check' target

ripatel-fd:
Check all compile units for warnings in under 3 seconds.


anwayde (approved):

ripatel-fd:
@kbowers-jump I'm merging this for now so we can improve our CI. Happy to belt-sand later.

PR #5183, opened 2025-05-23
Separating out some basic config, pod and file infra.

thartley-jumptrading:


mmcgee-jump:
Agree with Richie. Was the shared lib not working for some reason?

ripatel-fd:
Thanks for the contribution @thartley-jumptrading. I took the idea further here: https://github.com/firedancer-io/firedancer/pull/5197

PR #5196, opened 2025-05-27
resurrect rpc service as separate process

asiegel-jt:


src/app/firedancer/topology.c
```diff
@@ -734,11 +734,12 @@ fd_topo_initialize( config_t * config ) {
     /**/ fd_topob_tile_in( topo, "arch_w", 0UL, "metric_in", "arch_f2w", 0UL, FD_TOPOB_RELIABLE, FD_TOPOB_POLLED );
   }
 
-  if( enable_rpc ) {
-    fd_topob_wksp( topo, "replay_notif" );
-    /**/                 fd_topob_link( topo, "replay_notif", "replay_notif", FD_REPLAY_NOTIF_DEPTH,                    FD_REPLAY_NOTIF_MTU,           1UL   );
-    /**/                 fd_topob_tile_out( topo, "replay",  0UL,                       "replay_notif",  0UL                                                  );
+  fd_topob_wksp( topo, "replay_notif" );
+  /* We may be notifying an external service, so always publish on this link. */
```
mmcgee-jump, line 738: This is what the `enable_rpc` flag is for. I don't see why this needs to be changed? The external link won't exist if they aren't doing RPC, which makes it zero cost.

asiegel-jt: i'm supporting an external rpc service which doesn't run as a tile, as well as external geyser like stuff

mmcgee-jump: haven't we agreed a long time ago that this stuff needs to be tiles? external process is not a good design for a bunch of reasons

asiegel-jt: adding optional tiles is too hard to expect our customers to always do it

asiegel-jt: you can run rpc as a tile if you want. both options are supported.

mmcgee-jump: Both options are supported if you leave publishing behind a configuration flag. Why are you forcing non-RPC users to publish RPC information?

PR #5211, opened 2025-05-30
Cleanup stake_ci uses, factor out multi-epoch leader

akhinvasara-jumptrading:
- Clean up several 'uses' of stake_ci where it's not actually used.
- Implement fd_multi_epoch_leaders_t to handle stake updates and store/query leader schedules for multiple (currently assumed 2) epochs.
- Replace stake_ci with the new multi_epoch_leaders structure in the handful of cases we just care about leader schedule info, and not shred destinations
- minor refactors to use the new fd_stake_weight_msg_t struct to encode/decode stake update headers

To minimize blast radius, shred's fd_stake_ci is left unchanged at the cost of some code duplication, which ideally would be avoided by having stake_ci use the new multi_epoch_leader

Each consumer is modified in a separate commit, for easy reverts if needed. Review best done commit by commit, focusing on 60a75b921a71a9ae8cbbe7fb9b762fc99063bc0e.

### DIFF
--- a/src/disco/shred/fd_stake_ci.c
+++ b/src/disco/shred/fd_stake_ci.c
@@ -44,27 +44,19 @@ void * fd_stake_ci_delete( void          * mem  ) { return mem;          }
 
 
 void
-fd_stake_ci_stake_msg_init( fd_stake_ci_t * info,
-                            uchar const   * new_message ) {
-  ulong const * hdr = fd_type_pun_const( new_message );
-
-  ulong epoch               = hdr[ 0 ];
-  ulong staked_cnt          = hdr[ 1 ];
-  ulong start_slot          = hdr[ 2 ];
-  ulong slot_cnt            = hdr[ 3 ];
-  ulong excluded_stake      = hdr[ 4 ];
-
-  if( FD_UNLIKELY( staked_cnt > MAX_SHRED_DESTS ) )
+fd_stake_ci_stake_msg_init( fd_stake_ci_t               * info,
+                            fd_stake_weight_msg_t const * msg ) {
+  if( FD_UNLIKELY( msg->staked_cnt > MAX_SHRED_DESTS ) )
     FD_LOG_ERR(( "The stakes -> Firedancer splice sent a malformed update with %lu stakes in it,"
-                 " but the maximum allowed is %lu", staked_cnt, MAX_SHRED_DESTS ));
+                 " but the maximum allowed is %lu", msg->staked_cnt, MAX_SHRED_DESTS ));
 
-  info->scratch->epoch          = epoch;
-  info->scratch->start_slot     = start_slot;
-  info->scratch->slot_cnt       = slot_cnt;
-  info->scratch->staked_cnt     = staked_cnt;
-  info->scratch->excluded_stake = excluded_stake;
+  info->scratch->epoch          = msg->epoch;
+  info->scratch->start_slot     = msg->start_slot;
+  info->scratch->slot_cnt       = msg->slot_cnt;
+  info->scratch->staked_cnt     = msg->staked_cnt;
+  info->scratch->excluded_stake = msg->excluded_stake;
 
-  fd_memcpy( info->stake_weight, hdr+5UL, sizeof(fd_stake_weight_t)*staked_cnt );
+  fd_memcpy( info->stake_weight, msg->weights, msg->staked_cnt*sizeof(fd_stake_weight_t) );
 }
 
 static inline void
--- a/src/discof/poh/fd_poh_tile.c
+++ b/src/discof/poh/fd_poh_tile.c
@@ -315,12 +315,11 @@
 #include "../../disco/metrics/fd_metrics.h"
 #include "../../util/pod/fd_pod_format.h"
 #include "../../disco/shred/fd_shredder.h"
-#include "../../disco/shred/fd_stake_ci.h"
 #include "../../disco/keyguard/fd_keyload.h"
 #include "../../disco/keyguard/fd_keyswitch.h"
 #include "../../disco/metrics/generated/fd_metrics_poh.h"
 #include "../../disco/plugin/fd_plugin.h"
-#include "../../flamenco/leaders/fd_leaders.h"
+#include "../../flamenco/leaders/fd_multi_epoch_leaders.h"
 
 #include <string.h>
 
@@ -501,7 +500,7 @@ typedef struct {
 
   fd_sha256_t * sha256;
 
-  fd_stake_ci_t * stake_ci;
+  fd_multi_epoch_leaders_t * mleaders;
 
   /* The last sequence number of an outgoing fragment to the shred tile,
      or ULONG max if no such fragment.  See fd_keyswitch.h for details
@@ -548,6 +547,8 @@ typedef struct {
 
   ulong parent_slot;
   uchar parent_block_id[ 32 ];
+
+  uchar __attribute__((aligned(FD_MULTI_EPOCH_LEADERS_ALIGN))) mleaders_mem[ FD_MULTI_EPOCH_LEADERS_FOOTPRINT ];
 } fd_poh_ctx_t;
 
 /* The PoH recorder is implemented in Firedancer but for now needs to
@@ -1136,19 +1137,7 @@ next_leader_slot( fd_poh_ctx_t * ctx ) {
   /* If we have published anything in a particular slot, then we
      should never become leader for that slot again. */
   ulong min_leader_slot = fd_ulong_max( ctx->slot, fd_ulong_if( ctx->highwater_leader_slot==ULONG_MAX, 0UL, ctx->highwater_leader_slot ) );
-
-  for(;;) {
-    fd_epoch_leaders_t * leaders = fd_stake_ci_get_lsched_for_slot( ctx->stake_ci, min_leader_slot ); /* Safe to call from Rust */
-    if( FD_UNLIKELY( !leaders ) ) break;
-
-    while( min_leader_slot<(leaders->slot0+leaders->slot_cnt) ) {
-      fd_pubkey_t const * leader = fd_epoch_leaders_get( leaders, min_leader_slot ); /* Safe to call from Rust */
-      if( FD_UNLIKELY( !memcmp( leader->key, ctx->identity_key.key, 32UL ) ) ) return min_leader_slot;
-      min_leader_slot++;
-    }
-  }
-
-  return ULONG_MAX;
+  return fd_multi_epoch_leaders_get_next_slot( ctx->mleaders, min_leader_slot, &ctx->identity_key );
 }
 
 extern int
@@ -1180,7 +1169,6 @@ maybe_change_identity( fd_poh_ctx_t * ctx,
     }
 
     memcpy( ctx->identity_key.uc, ctx->keyswitch->bytes+32UL, 32UL );
-    fd_stake_ci_set_identity( ctx->stake_ci, &ctx->identity_key );
 
     /* When we switch key, we might have ticked part way through a slot
        that we are now leader in.  This violates the contract of the
@@ -1350,15 +1338,12 @@ fd_ext_poh_get_leader_after_n_slots( ulong n,
                                      uchar out_pubkey[ static 32 ] ) {
   fd_poh_ctx_t * ctx = fd_ext_poh_write_lock();
   ulong slot = ctx->slot + n;
-  fd_epoch_leaders_t * leaders = fd_stake_ci_get_lsched_for_slot( ctx->stake_ci, slot ); /* Safe to call from Rust */
+  fd_pubkey_t const * leader = fd_multi_epoch_leaders_get_leader_for_slot( ctx->mleaders, slot );
 
   int copied = 0;
-  if( FD_LIKELY( leaders ) ) {
-    fd_pubkey_t const * leader = fd_epoch_leaders_get( leaders, slot ); /* Safe to call from Rust */
-    if( FD_LIKELY( leader ) ) {
-      memcpy( out_pubkey, leader, 32UL );
-      copied = 1;
-    }
+  if( FD_LIKELY( leader ) ) {
+    memcpy( out_pubkey, leader, 32UL );
+    copied = 1;
   }
   fd_ext_poh_write_unlock();
   return copied;
@@ -1374,7 +1359,6 @@ scratch_footprint( fd_topo_tile_t const * tile ) {
   (void)tile;
   ulong l = FD_LAYOUT_INIT;
   l = FD_LAYOUT_APPEND( l, alignof( fd_poh_ctx_t ), sizeof( fd_poh_ctx_t ) );
-  l = FD_LAYOUT_APPEND( l, fd_stake_ci_align(), fd_stake_ci_footprint() );
   l = FD_LAYOUT_APPEND( l, FD_SHA256_ALIGN, FD_SHA256_FOOTPRINT );
   return FD_LAYOUT_FINI( l, scratch_align() );
 }
@@ -1821,7 +1805,7 @@ during_frag( fd_poh_ctx_t * ctx,
             ctx->in[ in_idx ].chunk0, ctx->in[ in_idx ].wmark ));
 
     uchar const * dcache_entry = fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk );
-    fd_stake_ci_stake_msg_init( ctx->stake_ci, dcache_entry );
+    fd_multi_epoch_leaders_stake_msg_init( ctx->mleaders, fd_type_pun_const( dcache_entry ) );
     return;
   }
 
@@ -1957,7 +1941,7 @@ after_frag( fd_poh_ctx_t *      ctx,
   if( FD_UNLIKELY( ctx->skip_frag ) ) return;
 
   if( FD_UNLIKELY( ctx->in_kind[ in_idx ]==IN_KIND_STAKE ) ) {
-    fd_stake_ci_stake_msg_fini( ctx->stake_ci );
+    fd_multi_epoch_leaders_stake_msg_fini( ctx->mleaders );
     /* It might seem like we do not need to do state transitions in and
        out of being the leader here, since leader schedule updates are
        always one epoch in advance (whether we are leader or not would
@@ -2215,7 +2199,6 @@ unprivileged_init( fd_topo_t *      topo,
 
   FD_SCRATCH_ALLOC_INIT( l, scratch );
   fd_poh_ctx_t * ctx = FD_SCRATCH_ALLOC_APPEND( l, alignof( fd_poh_ctx_t ), sizeof( fd_poh_ctx_t ) );
-  void * stake_ci = FD_SCRATCH_ALLOC_APPEND( l, fd_stake_ci_align(),              fd_stake_ci_footprint()            );
   void * sha256   = FD_SCRATCH_ALLOC_APPEND( l, FD_SHA256_ALIGN,                  FD_SHA256_FOOTPRINT                );
 
 #define NONNULL( x ) (__extension__({                                        \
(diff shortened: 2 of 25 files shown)


src/flamenco/repair/fd_repair.c
```diff
@@ -28,6 +28,7 @@ fd_repair_new ( void * shmem, ulong seed ) {
   shm = FD_SCRATCH_ALLOC_APPEND( l, fd_pinged_table_align(), fd_pinged_table_footprint(FD_REPAIR_PINGED_MAX) );
   glob->pinged = fd_pinged_table_join(fd_pinged_table_new(shm, FD_REPAIR_PINGED_MAX, seed));
   glob->stake_weights = FD_SCRATCH_ALLOC_APPEND( l, alignof(fd_stake_weight_t), FD_STAKE_WEIGHTS_MAX * sizeof(fd_stake_weight_t) );
+  glob->stake_weights_temp = FD_SCRATCH_ALLOC_APPEND( l, alignof(fd_stake_weight_t), FD_STAKE_WEIGHTS_MAX * sizeof(fd_stake_weight_t) );
   glob->stake_weights_cnt = 0;
```
ptaffet-jump, line 31: Not strictly necessary, but set `stake_weights_temp_cnt` to 0 as well

src/flamenco/repair/fd_repair.c
```diff
@@ -686,19 +687,24 @@ void fd_repair_add_sticky( fd_repair_t * glob, fd_pubkey_t const * id ) {
...
-  fd_memset( repair->stake_weights, 0, FD_STAKE_WEIGHTS_MAX * sizeof(fd_stake_weight_t) );
-  fd_memcpy( repair->stake_weights, stake_weights, stake_weights_cnt * sizeof(fd_stake_weight_t) );
-  repair->stake_weights_cnt = stake_weights_cnt;
+  fd_memcpy( repair->stake_weights_temp, stake_weights, stake_weights_cnt * sizeof(fd_stake_weight_t) );
+  repair->stake_weights_temp_cnt = stake_weights_cnt;
+}
+
+void
+fd_repair_set_stake_weights_fini( fd_repair_t * repair ) {
+  fd_memcpy( repair->stake_weights, repair->stake_weights_temp, repair->stake_weights_temp_cnt * sizeof(fd_stake_weight_t) );
```
ptaffet-jump: It's a bit of a shame this takes two memcpys, but I suppose it is infrequent

src/discof/eqvoc/fd_eqvoc_tile.c
```diff
@@ -60,39 +55,19 @@ scratch_footprint( fd_topo_tile_t const * tile FD_PARAM_UNUSED ) {
...
-  }
+handle_new_cluster_contact_info( fd_eqvoc_tile_ctx_t * ctx FD_PARAM_UNUSED, uchar const * buf FD_PARAM_UNUSED, ulong buf_sz FD_PARAM_UNUSED ) {
+  /* FIXME - do something useful */
 }
 
 static inline void
-finalize_new_cluster_contact_info( fd_eqvoc_tile_ctx_t * ctx ) {
-  fd_stake_ci_dest_add_fini( ctx->stake_ci, ctx->new_dest_cnt );
+finalize_new_cluster_contact_info( fd_eqvoc_tile_ctx_t * ctx FD_PARAM_UNUSED ) {
+  /* FIXME - do something useful */
```
ptaffet-jump: Talk to @lidatong and see if these functions can just be deleted

src/flamenco/leaders/fd_leaders.h
```diff
@@ -14,7 +14,7 @@
 
      slots_per_epoch = sched_cnt * slots_per_rotation
 
-   The leader can only change between rotations.  An example leader
+   The leader can only change between rotations.  An example leaderasdf
```
ptaffet-jump: Typo

src/flamenco/leaders/fd_multi_epoch_leaders.h
```diff
@@ -0,0 +1,175 @@
...
+   and querying to find the leader for a given slot.  While maintaining
+   the leader schedule for the current epoch i, you can also prepare the
+   schedule for epoch i+1 and send to the next epoch's leader as you
+   approach the boundary. */
+
+typedef uchar __attribute__((aligned(FD_EPOCH_LEADERS_ALIGN)))
+    _lsched_t[FD_EPOCH_LEADERS_FOOTPRINT(MAX_STAKED_LEADERS, MAX_SLOTS_PER_EPOCH)];
+
+#define MULTI_EPOCH_LEADERS_EPOCH_CNT (2UL)
+struct fd_multi_epoch_leaders {
```
ptaffet-jump: This struct is intended to be private, right? The name should reflect it, or at the very least, there should be a comment about it

src/flamenco/leaders/fd_multi_epoch_leaders.h
```diff
@@ -0,0 +1,175 @@
...
+
+/* fd_multi_epoch_leaders_delete unformats a memory region and returns owner-
+   ship back to the caller. */
+
+void *
+fd_multi_epoch_leaders_delete( void * shleaders );
+
+/* ///////////////////////////////////////////////////*/
+/* ///////   LEADER INFO GETTER FUNCTIONS     /////// */
+/* ///////////////////////////////////////////////////*/
```
ptaffet-jump: Sections like this are a bit non-standard. If you need a divider, I'd try something like 

> /**********************************************************************/

src/flamenco/leaders/fd_multi_epoch_leaders.h
```diff
@@ -0,0 +1,175 @@
...
+   excluded_stake for that epoch, instead of returning the correct value
+   (which is not known), returns a pointer to a pubkey with value
+   FD_INDETERMINATE_LEADER. */
+
+FD_FN_PURE static inline fd_pubkey_t const *
+fd_multi_epoch_leaders_get_leader_for_slot( fd_multi_epoch_leaders_t const * leaders,
+                                            ulong                            slot ) {
+  fd_pubkey_t const * even_leader_option = fd_ptr_if( leaders->init_done[0], fd_epoch_leaders_get( leaders->lsched[ 0 ], slot ), NULL );
+  fd_pubkey_t const * odd_leader_option  = fd_ptr_if( leaders->init_done[1], fd_epoch_leaders_get( leaders->lsched[ 1 ], slot ), NULL );
+  return fd_ptr_if( !even_leader_option, odd_leader_option, even_leader_option );
```
ptaffet-jump: You defined `MULTI_EPOCH_LEADERS_EPOCH_CNT` which suggests that you can change that value and the struct will adapt, but this implementation limits it to 2. You can put a static assert here that it is 2 with a note that the implementation needs to be adjusted otherwise.

src/flamenco/leaders/fd_multi_epoch_leaders.c
```diff
@@ -0,0 +1,125 @@
...
+  return fd_ptr_if( !!even_lsched, even_lsched, odd_lsched );
+}
+
+static inline ulong
+fd_multi_epoch_leaders_get_epoch_idx( fd_multi_epoch_leaders_t const * leaders,
+                                      ulong                            slot ) {
+  fd_epoch_leaders_t const * even_lsched = leaders->lsched[0];
+  fd_epoch_leaders_t const * odd_lsched  = leaders->lsched[1];
+
+  ulong even_match = fd_ulong_if( even_lsched->slot0<=slot && slot<even_lsched->slot0+even_lsched->slot_cnt, 0UL, ULONG_MAX );
```
ptaffet-jump: This one doesn't check `init_done`

ptaffet-jump: Is it possible to factor this kind of logic out so you don't keep repeating it?

src/flamenco/leaders/fd_multi_epoch_leaders.h
ptaffet-jump: It's not great that this code is essentially duplicated in the `.c` file. Is this performance critical such that the function call overhead makes it necessary for this to be inline-able?

src/flamenco/leaders/fd_multi_epoch_leaders.h
```diff
@@ -0,0 +1,175 @@
...
+   NULL if not tracked by mleaders. */
+
+FD_FN_PURE fd_epoch_leaders_t const *
+fd_multi_epoch_leaders_get_lsched_for_epoch( fd_multi_epoch_leaders_t const * mleaders, ulong epoch );
+FD_FN_PURE fd_epoch_leaders_t const *
+fd_multi_epoch_leaders_get_lsched_for_slot(  fd_multi_epoch_leaders_t const * mleaders, ulong slot  );
+
+/* fd_multi_epoch_leaders_get_next_slot returns the first slot (on or after
+   start_slot) that 'leader' will be leader. It only checks the epoch containing
+   start_slot. If it can't find one, returns ULONG_MAX. */
```
ptaffet-jump: Nits, but I wouldn't put "on or after start_slot" in parentheses.  It's actually a critical part of the operation of this function.  And then "If it can't find one" should be explained. Why wouldn't it be able to find one? I can think of three cases.

src/flamenco/leaders/fd_multi_epoch_leaders.c
```diff
@@ -0,0 +1,125 @@
...
+  /* Find epoch containing start_slot */
+  ulong epoch_idx = fd_multi_epoch_leaders_get_epoch_idx( leaders, start_slot );
+  if( FD_UNLIKELY( epoch_idx == ULONG_MAX ) ) return ULONG_MAX;
+
+  /* Find the leader in the epoch */
+  fd_epoch_leaders_t const * epoch_lsched = leaders->lsched[ epoch_idx ];
+  ulong                      slot0        = epoch_lsched->slot0;
+  ulong                      slot_end     = slot0 + epoch_lsched->slot_cnt;
+
+  for( ulong slot = start_slot; slot < slot_end; slot++ ) {
```
ptaffet-jump: The formatting here and elsewhere is not quite right.  See this random example from some of Kevin's code for details: https://github.com/firedancer-io/firedancer/blob/2e7dd29aa2912347bb1d1f4a9d25a1849333a30a/src/tango/dcache/test_dcache.c#L83

src/flamenco/leaders/fd_multi_epoch_leaders.c
```diff
@@ -0,0 +1,125 @@
...
+  ulong odd_match  = fd_ulong_if( odd_lsched->slot0<=slot && slot<odd_lsched->slot0+odd_lsched->slot_cnt, 1UL, ULONG_MAX );
+
+  return fd_ulong_if( even_match!=ULONG_MAX, even_match, odd_match );
+}
+
+fd_epoch_leaders_t const *
+fd_multi_epoch_leaders_get_lsched_for_slot( fd_multi_epoch_leaders_t const * mleaders,
+                                             ulong                            slot ) {
+  const ulong epoch_idx = fd_multi_epoch_leaders_get_epoch_idx( mleaders, slot );
+  if( FD_UNLIKELY( epoch_idx == ULONG_MAX ) ) return NULL;
```
ptaffet-jump: See formatting note below and https://github.com/firedancer-io/firedancer/blob/2e7dd29aa2912347bb1d1f4a9d25a1849333a30a/src/tango/tempo/fd_tempo.c#L153

src/flamenco/leaders/test_multi_leaders.c
```diff
@@ -0,0 +1,285 @@
...
+  fd_multi_epoch_leaders_t * mleaders = fd_multi_epoch_leaders_join( fd_multi_epoch_leaders_new( mleaders_mem ) );
+
+  fd_multi_epoch_leaders_stake_msg_init( mleaders, generate_stake_msg( stake_msg, 0UL, "ABC" ) );
+  fd_multi_epoch_leaders_stake_msg_fini( mleaders );
+
+  /* Test finding next slot for each leader */
+  fd_pubkey_t test_key;
+  memset( test_key.uc, 'A', sizeof(fd_pubkey_t) );
+  ulong next_slot = fd_multi_epoch_leaders_get_next_slot( mleaders, 0UL, &test_key );
+  FD_TEST( next_slot != ULONG_MAX );
```
ptaffet-jump: You should test it respects start_slot and that the returned slot actually has the right leader

src/flamenco/leaders/fd_multi_epoch_leaders.h
```diff
@@ -0,0 +1,175 @@
...
+void
+fd_multi_epoch_leaders_stake_msg_fini( fd_multi_epoch_leaders_t * leaders );
+
+
+/* ///////////////////////////////////////////////////*/
+/* ///////     MLEADER METADATA GETTERS       /////// */
+/* ///////////////////////////////////////////////////*/
+
+/* fd_multi_epoch_leaders_get_start_{epoch,slot} returns the earliest
+   epoch/slot that the leader schedule covers. Returns ULONG_MAX
```
ptaffet-jump: You might want to explicitly call out that it's not necessarily contiguous from there? These functions just seem a bit strange to me. They smell like someone might be using it wrong.

src/disco/shred/fd_stake_ci.c
```diff
@@ -1,5 +1,6 @@
 #include "fd_stake_ci.h"
 #include "../../util/net/fd_ip4.h" /* Just for debug */
+#include "../../discof/replay/fd_exec.h"
```
ptaffet-jump: I'm not a fan of disco->disco{f/h} dependencies. I think you should move this type to fd_stake_ci.h if you want to use it like this

src/discof/poh/fd_poh_tile.c
```diff
@@ -1136,19 +1137,7 @@ next_leader_slot( fd_poh_ctx_t * ctx ) {
...
-
-    while( min_leader_slot<(leaders->slot0+leaders->slot_cnt) ) {
-      fd_pubkey_t const * leader = fd_epoch_leaders_get( leaders, min_leader_slot ); /* Safe to call from Rust */
-      if( FD_UNLIKELY( !memcmp( leader->key, ctx->identity_key.key, 32UL ) ) ) return min_leader_slot;
-      min_leader_slot++;
-    }
-  }
-
-  return ULONG_MAX;
+  return fd_multi_epoch_leaders_get_next_slot( ctx->mleaders, min_leader_slot, &ctx->identity_key );
```
ptaffet-jump, line 1140: Nice! Making the function part of `fd_multi_epoch_leaders` was a good idea. This is an improvement.

src/discof/rpcserver/fd_rpc_service.c
```diff
@@ -76,7 +76,7 @@ struct fd_rpc_global_ctx {
   fd_perf_sample_t * perf_samples;
   fd_perf_sample_t perf_sample_snapshot;
   long perf_sample_ts;
-  fd_stake_ci_t * stake_ci;
+  fd_multi_epoch_leaders_t * ml;
```
ptaffet-jump: Can you give it a better name than `m1`?

src/flamenco/repair/fd_repair.c
akhinvasara-jumptrading: removing a memcpy by just swapping the ptrs on fini :) was pretty setup to do that already. @lidatong can you just sanity check that pls, thanks!

src/flamenco/leaders/fd_multi_epoch_leaders.h
akhinvasara-jumptrading: I thought it was called in a loop somewhere, but I was wrong / moved that into this class. Moving it all to the .c file :)

src/disco/shred/fd_stake_ci.c
mmcgee-jump: Yup this is a layering violation

src/discof/poh/fd_poh_tile.c
```diff
@@ -1180,7 +1169,6 @@ maybe_change_identity( fd_poh_ctx_t * ctx,
     }
 
     memcpy( ctx->identity_key.uc, ctx->keyswitch->bytes+32UL, 32UL );
-    fd_stake_ci_set_identity( ctx->stake_ci, &ctx->identity_key );
```
mmcgee-jump, line 1183: Why is this removed?

akhinvasara-jumptrading: leader schedule computation doesn't depend on identity. stake_ci uses that for shred dest calcs, which we don't use here.  @ptaffet-jump correct me if I'm wrong ?

src/flamenco/leaders/fd_multi_epoch_leaders.h
```diff
@@ -0,0 +1,169 @@
...
+FD_FN_PURE fd_epoch_leaders_t const *
+fd_multi_epoch_leaders_get_lsched_for_epoch( fd_multi_epoch_leaders_t const * mleaders,
+                                             ulong                            epoch );
+FD_FN_PURE fd_epoch_leaders_t const *
+fd_multi_epoch_leaders_get_lsched_for_slot( fd_multi_epoch_leaders_t const * mleaders,
+                                             ulong                           slot   );
+
+/* fd_multi_epoch_leaders_get_sorted_lscheds returns up to two lscheds,
+   sorted in increasing epoch order. If we only have data for one epoch,
+   the first element will be the corresponding lsched. If no lsched data,
```
ptaffet-jump, line 114:
```suggestion
   the first element will be the corresponding lsched, and the second will be NULL. If no lsched data,
```

ptaffet-jump: This comment also needs to talk about the lifetime of the pointers

src/discof/rpcserver/fd_rpc_service.c
```diff
@@ -517,18 +517,27 @@ method_getBlockProduction(struct json_values* values, fd_rpc_ctx_t * ctx) {
...
-    if( lsched ) {
-      for ( ulong i = startslot; i <= endslot; ++i ) {
-        fd_pubkey_t const * slot_leader = fd_epoch_leaders_get( lsched, i );
+    for( ulong i=0UL; i<2UL; i++ ) {
+      const fd_epoch_leaders_t * lsched = lscheds.lscheds[i];
+      if( !lsched ) continue;
+
+      ulong const epoch_start = fd_ulong_max( startslot, lsched->slot0 );
+      ulong const epoch_end   = fd_ulong_min( endslot+1, lsched->slot0 + lsched->slot_cnt );
+      if( epoch_end <= epoch_start ) continue;
```
ptaffet-jump: The names are a bit confusing here. I'd call it `start_slot_in_epoch` or something.  Also, technically the if statement isn't necessary, but it's fine for clarity

src/flamenco/leaders/fd_multi_epoch_leaders.c
```diff
@@ -0,0 +1,151 @@
...
+  return fd_ptr_if( !!even_lsched, even_lsched, odd_lsched );
+}
+
+static inline ulong
+fd_multi_epoch_leaders_get_epoch_idx( fd_multi_epoch_leaders_t const * mleaders,
+                                      ulong                            slot ) {
+  fd_epoch_leaders_t const * even_lsched = mleaders->lsched[0];
+  fd_epoch_leaders_t const * odd_lsched  = mleaders->lsched[1];
+
+  ulong even_match = fd_ulong_if( mleaders->init_done[0] && even_lsched->slot0<=slot && slot<even_lsched->slot0+even_lsched->slot_cnt, 0UL, ULONG_MAX );
```
ptaffet-jump: Not a big deal, but using `&&` makes this actually not branchless.

src/discof/poh/fd_poh_tile.c
ptaffet-jump: Yep

src/discof/rpcserver/fd_rpc_service.c
akhinvasara-jumptrading: rm-ing it, added a comment to that effect instead

PR #5219, opened 2025-05-30
Disable bundle crank if vote account not set

ripatel-fd:


src/discof/poh/fd_poh_tile.c
```diff
@@ -2089,6 +2089,7 @@ privileged_init( fd_topo_t *      topo,
   const uchar * identity_key = fd_keyload_load( tile->poh.identity_key_path, /* pubkey only: */ 1 );
   fd_memcpy( ctx->identity_key.uc, identity_key, 32UL );
 
+  if( !tile->poh.bundle.vote_account_path[0] ) tile->poh.bundle.enabled = 0;
```
ptaffet-jump, line 2092: Can you mark it as `FD_UNLIKELY` just for clarity? Same with the one below

src/disco/pack/fd_pack_tile.c
```diff
@@ -992,6 +992,7 @@ static void
 privileged_init( fd_topo_t *      topo,
                  fd_topo_tile_t * tile ) {
   if( FD_LIKELY( !tile->pack.bundle.enabled ) ) return;
+  if( FD_UNLIKELY( !tile->pack.bundle.vote_account_path[0] ) ) return;
```
ptaffet-jump, line 995:
```suggestion
  if( FD_UNLIKELY( !tile->pack.bundle.vote_account_path[0] ) ) {
    FD_LOG_WARNING(( "disabling bundles because no vote account was specified" ));
    return;
  }
```

PR #5224, opened 2025-05-30
Send the stake state in the format the rest of the tiles are expecting

jumpsiegel:
fix poh initialization given the new tile

### DIFF
--- a/src/flamenco/runtime/fd_runtime.c
+++ b/src/flamenco/runtime/fd_runtime.c
@@ -148,6 +148,58 @@ fd_runtime_update_slots_per_epoch( fd_exec_slot_ctx_t * slot_ctx,
   fd_runtime_repartition_fresh_account_partitions( slot_ctx );
 }
 
+#define FIREDANCER_STAKE_WEIGHT_CNT (40200UL)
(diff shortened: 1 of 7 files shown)


src/flamenco/runtime/fd_runtime.c
```diff
@@ -149,6 +149,58 @@ fd_runtime_update_slots_per_epoch( fd_exec_slot_ctx_t * slot_ctx,
...
+    stake_weight_cnt=FIREDANCER_STAKE_WEIGHT_CNT;
+  }
+
+  uchar *memory = fd_spad_alloc_check( runtime_spad, 1U, FIREDANCER_PACKET_SZ );
+  if( FD_UNLIKELY( NULL == memory ) ) {
+    FD_LOG_ERR(("spad allocation failure setting up a firedancer_packet"));
+  }
+
+
+  *((ulong *) &memory[0]) = epoch;
```
akhinvasara-jumptrading, line 183: @emwang-jump put this in a struct, might be cleaner to use that? src/discof/replay/fd_exec.h

jumpsiegel: I was unaware somebody had already written this code...

PR #5242, opened 2025-06-01
 quic: lazy evict from conn ID map 

ripatel-fd:
- **quic: test_quic_conformance in CI**
- **quic: lazy evict from conn ID map**

Let conn IDs for freed conns linger in the conn ID map.  Only
evict those dead conn IDs when a new conn is initialized.
If a packet arrives for those dead conn IDs, count it towards
the 'key not available' metric, instead of the 'no conn ID'
metric.

This allows the operator to better differentiate between packets
that address completely unknown conn IDs, vs packets that arrived
just after a conn timed out or closed.

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -673,6 +673,7 @@ static inline void
 fd_quic_svc_schedule1( fd_quic_conn_t * conn ) {
   fd_quic_svc_schedule( fd_quic_get_state(conn->quic)->svc_timers, conn );
 }
+
 /* validates the free conn list doesn't cycle, point nowhere, leak, or point to live conn */
 static void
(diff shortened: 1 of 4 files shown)


src/waltz/quic/fd_quic.c
akhinvasara-jumptrading, line 1: Not necessarily part of this PR, but do we want to remove the duplicate check for !conn right before calling handle_v1_one_rtt. Not correctness issue, but may be cleaner?

ripatel-fd: Sure, where would you move it to?

ripatel-fd: I see now, nevermind

ripatel-fd: Fixed

PR #5255, opened 2025-06-02
flamenco, runtime: input region should be 16-byte aligned

yufeng-jump:


mjain-jump (approved):

mjain-jump:
You'll have to regenerate the fixtures

kbhargava-jump:
mismatches from `sol_prog_vm_interp_diff` use the instruction harness and hence should not be in `vm_interp-fixtures.list`. thats probably why the fixtures are failing there

kbhargava-jump (approved):

PR #5263, opened 2025-06-04
Reverting svc queue change

akhinvasara-jumptrading:
Reverting the service queue changes.


src/waltz/quic/fd_quic_svc_q.h
ripatel-fd, line 1: Can we keep this header?

src/waltz/quic/fd_quic_common.h
ripatel-fd, line 1: Could you revert this whitespace change?

src/waltz/quic/tests/test_quic_conformance.c
ripatel-fd, line 1: IIRC: The old service queue had an invariant that all conns that aren't INVALID state are present in one of the service lists. Are we still checking this in the validate function?

akhinvasara-jumptrading: yep, bc we validate the 3 svc queues, and then check that all non INVALID have been visited
imo that validation stuff could be a bit more clean/explicit, but good for now

nbridge-jump (approved):

PR #5265, opened 2025-06-04
ci: update linux_gcc_x86_64 to use gcc 15

anwayde:
DO NOT MERGE


ripatel-fd:
> DO NOT MERGE

Why not?

anwayde:
Oh because it was broken, so I thought we should fix the build before merging the check. Nightly builds will check gcc-15 anyways. I'll re-open when it builds.

PR #5270, opened 2025-06-04
fd_txncache_set_is_constipated switch to write lock

anarcheuz:
Comment seems to indicate except for queries and insertions, other state modifications need a write lock:
```Insertion and querying only take a read lock as they can be done lockless but all other operations will take a write lock internally.```


ibhatt-jumptrading:
we are getting rid of this constipation mechanism completely CC: @yufeng-jump is rewriting the status cache

PR #5274, opened 2025-06-04
quic: fine-grained pkt_meta alloc metrics

akhinvasara-jumptrading:
This PR replaces pkt_tx_alloc_fail_cnt with an enum metric measuring success and two failure reasons. Because a given piece of outgoing data will increment failure each iteration of service until it succeeds, adding success count lets us look at the proportion of service iterations that were successful.


src/waltz/quic/fd_quic.c
```diff
@@ -3029,6 +3029,33 @@ fd_quic_tx_buffered( fd_quic_t *      quic,
...
 
+static inline int
+fd_quic_conn_can_acquire_pkt_meta( fd_quic_conn_t             * conn,
+                                   fd_quic_pkt_meta_tracker_t * tracker ) {
+  fd_quic_state_t * state = fd_quic_get_state( conn->quic );
+  fd_quic_metrics_t * metrics = &conn->quic->metrics;
+
+  #define FRAME_TX_ALLOC_SUCCESS         0
+  #define FRAME_TX_ALLOC_FAIL_EMPTY_POOL 1
+  #define FRAME_TX_ALLOC_FAIL_CONN_MAX   2
```
ripatel-fd: I would include `fd_metrics_enums.h` instead and just use the enum value. That way, you avoid dup definitions of these macros.

akhinvasara-jumptrading: yeah this is so much better, thanks!

PR #5277, opened 2025-06-04
quic - only increment pkt_number when we have committed the current packet into the tx buffer

nbridge-jump:
Current behavior causes gaps in packet numbers, leading to inefficient ack ranges, and difficulty with diagnostics.
This patch avoids most such skips.

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -3688,7 +3688,7 @@ fd_quic_conn_tx( fd_quic_t      * quic,
     /* get next packet number
        Returned to pool if not sent as gaps are harmful for ACK frame
        compression. */
-    ulong pkt_number = conn->pkt_number[pn_space]++;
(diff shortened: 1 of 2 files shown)


akhinvasara-jumptrading (approved):
Nice!

nbridge-jump:
> At risk of being annoying, can we extend the test case to make sure acks still get sent (and thus pkt num still goes up) even with no pkt meta?

Good idea

nbridge-jump:
> At risk of being annoying, can we extend the test case to make sure acks still get sent (and thus pkt num still goes up) even with no pkt meta?

I'm going to put this suggestion into another PR, so this can get merged

nbridge-jump:
> At risk of being annoying, can we extend the test case to make sure acks still get sent (and thus pkt num still goes up) even with no pkt meta?

Change of plan: it failed due to merge issue, so I added this suggestion when I fixed the merge issue.

PR #5284, opened 2025-06-05
[net 2.0] multi-queue support

ripatel-fd:


src/app/shared/commands/run/run.c
```diff
@@ -233,7 +234,7 @@ main_pid_namespace( void * _args ) {
     if( FD_UNLIKELY( close( args->closefd ) ) ) FD_LOG_ERR(( "close() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
   }
 
-  config_t const * config = args->config;
+  fd_config_t * config = args->config;
```
mmcgee-jump, line 237: Why not const?

src/app/shared/commands/run/run.c
```diff
@@ -286,7 +287,7 @@ main_pid_namespace( void * _args ) {
   if( FD_UNLIKELY( -1==save_priority && errno ) ) FD_LOG_ERR(( "getpriority() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
 
   int need_xdp = 0==strcmp( config->net.provider, "xdp" );
-  fd_xdp_fds_t xdp_fds = {0};
+  fd_xdp_multi_fds_t xdp_fds = {0};
   if( need_xdp ) {
     xdp_fds = fd_topo_install_xdp( &config->topo, config->net.bind_address_parsed );
```
mmcgee-jump: This needs to move before `fd_config_to_memfd` for the changes to flow to child processes.

mmcgee-jump (changes requested):
.

src/app/shared/commands/run/run.c
ripatel-fd: @mmcgee-jump Because it's modified to update the file descriptor numbers for XSK map and BPF prog link

src/app/fdctl/config/default.toml
```diff
@@ -904,6 +904,8 @@ dynamic_port_range = "8900-9000"
     # this should be the same as [development.netns.interface0].
     interface = ""
 
+    ethtool_queue_count = 8
```
ripatel-fd: @mmcgee-jump Can you think of a good doc comment?

ripatel-fd: - refers to the total queue count load balanced across net tiles
- each net tile can handle up to 8 queues

ripatel-fd:
Closing for now

PR #5304, opened 2025-06-06
flamenco, runtime: clean up partitioned epoch rewards feature gating

yufeng-jump:
https://github.com/anza-xyz/agave/pull/4186


kbhargava-jump:
lets see what ledgers fail and figure out if we can remove them (doing offline)

topointon-jump:
@kbhargava-jump let's not remove them, we should be able to easily re-write them. the partitioned rewards ledgers are valuable.

PR #5306, opened 2025-06-06
shred: drop unchained merkle shreds

0x0ece:


ptaffet-jump (commented):
Maybe my vscode is wrong, but I think because the rust side is using `solana_sdk::feature_set`, it hasn't picked up the re-key.  Also, could you add a metric counter for this error?

0x0ece:
I'll add the metric.

The rekey could be in a diff PR, it's just for Firedancer (and in Firedancer we still can't do chained merkle shreds, we're still awaiting for the leader pipeline).

0x0ece:
@ptaffet-jump Added the metric and rebased to main - this also updated the gate in agave. So I left the fd gate in, since now they're the same.

ptaffet-jump (approved):

PR #5309, opened 2025-06-06
simd-0178: static syscalls (update)

0x0ece:
Amending https://github.com/firedancer-io/firedancer/pull/3711

As a follow up of the discussion in https://github.com/solana-foundation/solana-improvement-documents/pull/178, we decided to no longer have an array of syscalls but keep using murmur32(syscall-name).


src/flamenco/vm/test_vm_interp.c
```diff
@@ -298,60 +298,6 @@ test_0cu_exit( void ) {
...
-    "sol_get_rent_sysvar",
-  };
-
-  FD_TEST( FD_VM_SBPF_STATIC_SYSCALLS_LIST[0]==0 );
-  for( ulong i=1; i<FD_VM_SBPF_STATIC_SYSCALLS_LIST_SZ; i++ ) {
-    const char *name = static_syscalls_from_simd[i-1];
-    uint key = fd_murmur3_32( name, strlen(name), 0 );
-    FD_TEST( FD_VM_SBPF_STATIC_SYSCALLS_LIST[i]==key );
-  }
-}
```
ripatel-fd: Please keep the murmur32 hash tests for these syscalls. In case someone breaks the hash function

0x0ece: good idea, I'll move the list inside the test (since we don't need the list anymore in the vm)

PR #5312, opened 2025-06-08
ibverbs tile

riptl:
Adds a basic ibverbs-based net tile replacement.
Not yet optimized for performance.


ripatel-fd:
Moving to https://github.com/firedancer-io/firedancer/pull/5399

PR #5325, opened 2025-06-09
fuzz, vm: add syscall protobuf dumper

mjain-jump:


src/app/ledger/main.c
```diff
@@ -664,7 +665,7 @@ fd_ledger_capture_setup( fd_ledger_args_t * args ) {
   int has_solcap           = args->capture_fpath && args->capture_fpath[0] != '\0';
   int has_checkpt          = args->checkpt_path && args->checkpt_path[0] != '\0';
   int has_checkpt_funk     = args->checkpt_funk && args->checkpt_funk[0] != '\0';
-  int has_dump_to_protobuf = args->dump_insn_to_pb || args->dump_txn_to_pb || args->dump_block_to_pb;
+  int has_dump_to_protobuf = args->dump_insn_to_pb || args->dump_txn_to_pb || args->dump_block_to_pb || args->dump_syscall_to_pb;
```
topointon-jump, line 668: Do we need to start adding these features to `backtest` instead? As we are removing ledger replay

mjain-jump: i'll add them to backtest as well in a separate pr

PR #5333, opened 2025-06-10
deps: only install llvm-toolset on rhel

ripatel-fd:
https://github.com/firedancer-io/firedancer/issues/5315

### DIFF
--- a/deps.sh
+++ b/deps.sh
@@ -162,7 +162,10 @@ check_fedora_pkgs () {
     protobuf-compiler  # Agave, solfuzz
   )
   if [[ $DEVMODE == 1 ]]; then
-    REQUIRED_RPMS+=( autoconf automake bison cmake clang flex gettext-devel gmp-devel llvm-toolset lcov )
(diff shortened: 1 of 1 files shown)


deps.sh
```diff
@@ -162,7 +162,10 @@ check_fedora_pkgs () {
     protobuf-compiler  # Agave, solfuzz
   )
   if [[ $DEVMODE == 1 ]]; then
-    REQUIRED_RPMS+=( autoconf automake bison cmake clang flex gettext-devel gmp-devel llvm-toolset lcov )
+    REQUIRED_RPMS+=( autoconf automake bison cmake clang flex gettext-devel gmp-devel lcov )
+    if [[ "${ID_LIKE:-}" == *rhel* ]]; then
```
anwayde, line 166:
```suggestion
    if [[ "${ID_LIKE:-}" != *fedora* ]]; then
```

ripatel-fd: I think ==rhel is the way to go here, actually. llvm-toolset is a RHEL specific package. And RHEL derivatives like rocky will still have the `rhel` string in ID_LIKE

PR #5342, opened 2025-06-11
missing deprecated types in gossip definitions

jumpsiegel:
missing deprecated types which allowed for FD to not correctly detect an invalid gossip message


ravyu-jump (approved):
fuzzer found this? Pretty cool

PR #5344, opened 2025-06-11
disco: shred tile fixed32 shreds per fec set

jvarela-jump:
Upgrading the shred_tile to always generate fec sets with 32 data + 32 parity (code) shreds.

### DIFF
--- a/src/disco/shred/fd_shred_tile.c
+++ b/src/disco/shred/fd_shred_tile.c
@@ -120,11 +120,58 @@ FD_STATIC_ASSERT( sizeof(fd_entry_batch_meta_t)==56UL, poh_shred_mtu );
 #define SHOULD_PROCESS_THESE_SHREDS ( ctx->batch_cnt%ctx->round_robin_cnt==ctx->round_robin_id )
 #endif
 
-/* The behavior of the shred tile is slightly different for Frankendancer vs Firedancer,
-   for example Frankendancer produces chained merkle shreds, while Firedancer doesn't yet.
-   We can check at runtime the difference by inspecting the topology. The simplest way is
-   to test if ctx->blockstore is enabled. */
+/* The behavior of the shred tile is slightly different for
+   Frankendancer vs Firedancer.  For example, Frankendancer produces
+   chained merkle shreds, while Firedancer doesn't yet.  We can check
+   at runtime the difference by inspecting the topology. The simplest
+   way is to test if ctx->blockstore is enabled. */
 #define IS_FIREDANCER ( ctx->blockstore!=NULL )
+
+/* The number of fec sets that we intend to send per batch (i.e. until
+   the watermark) is arbitrary.  The larger the value, the higher the
+   efficiency in terms of compactness, by reducing excessive padding.
+   However, the larger the value, the longer it takes before we can
+   start generating and forwarding fec sets.  The number of extra fec
+   sets (after the watermark) needs to suppport FD_POH_SHRED_MTU bytes
+   (for further details, refer to PENDING_BATCH_WMARK).  Also keep in
+   mind that PENDING_BATCH_FEC_SETS_MAX affects the value of STEM_BURST
+   (refer to the calculation of STEM_BURST for further details). */
+#define PENDING_BATCH_FEC_SETS_WMARK ( 2UL )
+#define PENDING_BATCH_FEC_SETS_EXTRA ( 2UL )
+#define PENDING_BATCH_FEC_SETS_MAX   ( PENDING_BATCH_FEC_SETS_WMARK + PENDING_BATCH_FEC_SETS_EXTRA )
+FD_STATIC_ASSERT( ( PENDING_BATCH_FEC_SETS_EXTRA * FD_SHREDDER_NORMAL_FEC_SET_PAYLOAD_SZ   ) >= ( FD_POH_SHRED_MTU ), PENDING_BATCH_FEC_SETS_EXTRA );
+FD_STATIC_ASSERT( ( PENDING_BATCH_FEC_SETS_EXTRA * FD_SHREDDER_CHAINED_FEC_SET_PAYLOAD_SZ  ) >= ( FD_POH_SHRED_MTU ), PENDING_BATCH_FEC_SETS_EXTRA );
+FD_STATIC_ASSERT( ( PENDING_BATCH_FEC_SETS_EXTRA * FD_SHREDDER_RESIGNED_FEC_SET_PAYLOAD_SZ ) >= ( FD_POH_SHRED_MTU ), PENDING_BATCH_FEC_SETS_EXTRA );
+
+/* PENDING_BATCH_WMARK: Following along the lines of dcache, batch
+   microblocks until either the slot ends or the batch would exceed
+   the corresponding watermark.  Only when the received microblock
+   completes the block is the buffer allowed to go beyond the given
+   watermark, since all remaining fec sets need to be generated and
+   forwarded at once.  This is also the reason why the extra fec
+   sets must support FD_POH_SHRED_MTU.  The watermark is relative
+   to the beginning of the payload in the pending_batch buffer (i.e.
+   excluding the 8 bytes needed for microblock_cnt).  It also needs
+   to support at least FD_POH_SHRED_MTU bytes. */
+#define PENDING_BATCH_WMARK_NORMAL   ( PENDING_BATCH_FEC_SETS_WMARK * FD_SHREDDER_NORMAL_FEC_SET_PAYLOAD_SZ   - 8UL )
+#define PENDING_BATCH_WMARK_CHAINED  ( PENDING_BATCH_FEC_SETS_WMARK * FD_SHREDDER_CHAINED_FEC_SET_PAYLOAD_SZ  - 8UL )
+#define PENDING_BATCH_WMARK_RESIGNED ( PENDING_BATCH_FEC_SETS_WMARK * FD_SHREDDER_RESIGNED_FEC_SET_PAYLOAD_SZ - 8UL )
+/* Prefer > to >= as a safety margin, to avoid any dependency on
+   how we determine if the batch would exceed the watermark. */
+FD_STATIC_ASSERT( ( PENDING_BATCH_WMARK_NORMAL   ) > ( FD_POH_SHRED_MTU ), PENDING_BATCH_FEC_SETS_EXTRA );
+FD_STATIC_ASSERT( ( PENDING_BATCH_WMARK_CHAINED  ) > ( FD_POH_SHRED_MTU ), PENDING_BATCH_FEC_SETS_EXTRA );
+FD_STATIC_ASSERT( ( PENDING_BATCH_WMARK_RESIGNED ) > ( FD_POH_SHRED_MTU ), PENDING_BATCH_FEC_SETS_EXTRA );
+
+/* There are three different raw buffer sizes depending on the fec
+   set type: normal, chained, and resigned (see fd_shredder.h for
+   further details).  In order to support all three, we allocate the
+   largest of them (i.e. normal).  TODO once normal fec sets have
+   been deprecated, replace NORMAL with CHAINED payload size in the
+   calculation below. */
+#define PENDING_BATCH_RAW_BUF_SZ     ( PENDING_BATCH_FEC_SETS_MAX * FD_SHREDDER_NORMAL_FEC_SET_PAYLOAD_SZ )
+FD_STATIC_ASSERT( ( PENDING_BATCH_RAW_BUF_SZ ) >= ( PENDING_BATCH_WMARK_NORMAL  + FD_POH_SHRED_MTU ), PENDING_BATCH_RAW_BUF_SZ );
+FD_STATIC_ASSERT( ( PENDING_BATCH_RAW_BUF_SZ ) >= ( PENDING_BATCH_WMARK_CHAINED + FD_POH_SHRED_MTU ), PENDING_BATCH_RAW_BUF_SZ );
+FD_STATIC_ASSERT( ( PENDING_BATCH_RAW_BUF_SZ ) >= ( PENDING_BATCH_WMARK_RESIGNED+ FD_POH_SHRED_MTU ), PENDING_BATCH_RAW_BUF_SZ );
 
 typedef union {
   struct {
@@ -179,13 +226,13 @@ typedef struct {
   ulong shredder_fec_set_idx;     /* In [0, shredder_max_fec_set_idx) */
   ulong shredder_max_fec_set_idx; /* exclusive */
 
-  ulong send_fec_set_idx;
+  ulong send_fec_set_idx[ PENDING_BATCH_FEC_SETS_MAX ];
+  ulong send_fec_set_cnt;
   ulong tsorig;  /* timestamp of the last packet in compressed form */
 
   /* Includes Ethernet, IP, UDP headers */
   ulong shred_buffer_sz;
   uchar shred_buffer[ FD_NET_MTU ];
-
 
   fd_shred_in_ctx_t in[ 32 ];
   int               in_kind[ 32 ];
@@ -227,18 +274,14 @@ typedef struct {
 
   struct {
     ulong txn_cnt;
-    ulong pos; /* in payload, so 0<=pos<63671 */
+    ulong pos; /* in payload, range [0, PENDING_BATCH_RAW_BUF_SZ-8UL) */
     ulong slot; /* set to 0 when pos==0 */
     union {
       struct {
         ulong microblock_cnt;
-        uchar payload[ FD_SHREDDER_NORMAL_FEC_SET_RAW_BUF_SZ - 8UL ];
+        uchar payload[ PENDING_BATCH_RAW_BUF_SZ - 8UL ];
       };
-      /* There are three different raw buffer sizes depending on the
-         fec_set type: normal, chained, and (chained+)resigned (see
-         fd_shredder.h for further details).  In order to support all
-         three, we allocate the largest of them (i.e. normal). */
-      uchar raw[ FD_SHREDDER_NORMAL_FEC_SET_RAW_BUF_SZ ];
+      uchar raw[ PENDING_BATCH_RAW_BUF_SZ ];
     };
   } pending_batch;
 
@@ -248,19 +291,6 @@ typedef struct {
 
   uchar chained_merkle_root[ FD_SHRED_MERKLE_ROOT_SZ ];
 } fd_shred_ctx_t;
-
-/* PENDING_BATCH_WMARK: Following along the lines of dcache, batch
-   microblocks until either the slot ends or we excede the watermark.
-   We know that if we're <= watermark, we can always accept a message
-   of maximum size.  The watermark is computed for (chained+)resigned
-   fec_sets, which support the smallest raw buffer size (see comments
-   inside fd_shred_ctx_t for further details).  This is a tradeoff
-   between fec_set compactness efficiency and computational complexity,
-   making normal and chained fec_sets slightly less efficient (about
-   10% and 6% respectively) in the worst-case scenario.  TODO consider
-   modifying the logic around "last_in_batch" to support multiple
-   watermarks. */
-#define PENDING_BATCH_WMARK (FD_SHREDDER_RESIGNED_FEC_SET_RAW_BUF_SZ - 8UL - FD_POH_SHRED_MTU)
 
 FD_FN_CONST static inline ulong
 scratch_align( void ) {
@@ -332,7 +362,7 @@ handle_new_cluster_contact_info( fd_shred_ctx_t * ctx,
   ctx->new_dest_cnt = dest_cnt;
 
   for( ulong i=0UL; i<dest_cnt; i++ ) {
-    memcpy( dests[i].pubkey.uc, in_dests[i].pubkey, 32UL );
+    fd_memcpy( dests[i].pubkey.uc, in_dests[i].pubkey, 32UL );
     dests[i].ip4  = in_dests[i].ip4_addr;
     dests[i].port = in_dests[i].udp_port;
   }
@@ -402,7 +432,7 @@ during_frag( fd_shred_ctx_t * ctx,
   }
 
   if( FD_UNLIKELY( ctx->in_kind[ in_idx ]==IN_KIND_POH ) ) {
-    ctx->send_fec_set_idx = ULONG_MAX;
+    ctx->send_fec_set_cnt = 0UL;
 
     if( FD_UNLIKELY( (fd_disco_poh_sig_pkt_type( sig )==POH_PKT_TYPE_FEAT_ACT_SLOT) ) ) {
       /* There is a subset of FD_SHRED_FEATURES_ACTIVATION_... slots that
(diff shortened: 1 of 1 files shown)


src/disco/shred/fd_shred_tile.c
```diff
@@ -326,7 +311,7 @@ handle_new_cluster_contact_info( fd_shred_ctx_t * ctx,
   ctx->new_dest_cnt = dest_cnt;
 
   for( ulong i=0UL; i<dest_cnt; i++ ) {
-    memcpy( dests[i].pubkey.uc, in_dests[i].pubkey, 32UL );
+    fd_memcpy( dests[i].pubkey.uc, in_dests[i].pubkey, 32UL );
```
0x0ece: I think you can leave memcpy when the sz is known. fd_memcpy does an extra sz != 0 check. Same below.

jvarela-jump: Usage of fd_memcpy and fd_memset has been adjusted accordingly (inside fd_shred_tile.c).

src/disco/shred/fd_shred_batch.h
```diff
@@ -0,0 +1,124 @@
...
+   generating and sending FEC sets.  Assuming n batches in a given
+   block, the (upper-limit) quantity of FEC sets is determined by
+   FD_SHRED_BATCH_FEC_SETS_WMARK for batch ids [0, n-2], and by
+   FD_SHRED_BATCH_FEC_SETS_MAX for batch id n-1.  This is needed
+   because when the microblock completing the block is received,
+   it must be included in the current batch irrespective of how
+   full the latter might be.  The diagrams below (not at scale)
+   show 2 FEC sets before the watermark (W), which indicates where
+   to close the current batch, as well as 2 extra FEC sets that
+   are available for the last batch in the block.
```
ptaffet-jump:
```suggestion
   data shreds and 32 parity shreds each.  Each batch is made up of
   an integer number of microblocks, which means that
   no microblock can cross the batch boundary, and that the
   last FEC set may need to be padded with 0s.  Since padding
   completes the payload of the last FEC set in the batch, and doesn't impact
   batch itself,  no FEC set can contain only padding. 
   The number of FEC sets per batch is limited only by block-level limits, so the larger the
   value, the less relative padding overhead in every batch.  However, the
   larger the value, the longer it takes before the tile can start
   generating and sending FEC sets, unless you use some strange speculative tricks, which we don't use.  Assuming n batches in a given
   block, the (upper-limit) quantity of FEC sets is determined by
   FD_SHRED_BATCH_FEC_SETS_WMARK for the first n-1 batches, and by
   FD_SHRED_BATCH_FEC_SETS_MAX for the last batch.  This is needed
   because when the microblock completing the block is received,
   we include it in the current batch irrespective of how
   full the batch might be.  The diagrams below (not at scale)
   show 2 FEC sets before the watermark (W), which indicates where
   to close the current batch, as well as 2 extra FEC sets that
   are available for the last batch in the block.
```

src/disco/shred/fd_shred_batch.h
```diff
@@ -0,0 +1,124 @@
...
+
+     Normal: (to be deprecated)
+       +------------+------------+~~~~~~~~~~~+~~~~~~~~~~~~+
+       |   FEC_set  |   FEC_set (W)    extra |      extra |
+       +------------+------------+~~~~~~~~~~~+~~~~~~~~~~~~+
+
+     Chained: (batch ids [0, n-2])
```
ptaffet-jump:
```suggestion
     Chained: (first n-1 batches)
```

src/disco/shred/fd_shred_batch.h
```diff
@@ -0,0 +1,124 @@
...
+       +----------+----------+~~~~~~~~~~+~~~~~~~~~~+
+
+   Batch ids [0, n-2] are chained, whereas batch id n-1 (the
+   last one in the block) is (chained+)resigned, with a smaller
+   payload capacity.  The transition between the two implies a
+   watermark regression, requiring FD_SHRED_BATCH_FEC_SETS_EXTRA
+   to be able to support (at least) FD_POH_SHRED_MTU bytes plus
+   the watermark regression.  This payload reduction does not
+   affect normal FEC sets, which are scheduled to be deprecated
+   at some point (TODO). */
```
ptaffet-jump:
```suggestion
   The first n-1 batches are chained, whereas the
   last one in the block is (chained+)resigned, which means each shred (and thus each FEC set) in the last batch has a smaller
   payload capacity.  In order to size the region labeled extra, we need to consider the worst case scenario: we've buffered W-1 bytes, using the chained watermark when a max-size microblock arrives, indicating it's the last in the block.  This situation can only happen when running in low-power mode, because otherwise the last microblock would be a tick, which is small, but it doesn't seem to be worth plumbing in that sort of logic just to save a little memory that won't pollute cache anyway.
   
  Note that this payload reduction does not
   affect normal FEC sets, which are scheduled to be deprecated
   at some point (TODO). */
```

src/disco/shred/fd_shred_batch.h
```diff
@@ -0,0 +1,124 @@
...
+#define FD_SHRED_BATCH_WMARK_NORMAL   ( FD_SHRED_BATCH_FEC_SETS_WMARK * FD_SHREDDER_NORMAL_FEC_SET_PAYLOAD_SZ   - 8UL )
+#define FD_SHRED_BATCH_WMARK_CHAINED  ( FD_SHRED_BATCH_FEC_SETS_WMARK * FD_SHREDDER_CHAINED_FEC_SET_PAYLOAD_SZ  - 8UL )
+#define FD_SHRED_BATCH_WMARK_RESIGNED ( FD_SHRED_BATCH_FEC_SETS_WMARK * FD_SHREDDER_RESIGNED_FEC_SET_PAYLOAD_SZ - 8UL )
+/* Validate watermark payload capacity.  It must support at least
+   FD_POH_SHRED_MTU bytes.  Prefer > to >= as a safety margin, to
+   avoid any dependency on how we determine later on if the batch
+   would exceed the watermark. */
+FD_STATIC_ASSERT( ( FD_SHRED_BATCH_WMARK_NORMAL   ) > ( FD_POH_SHRED_MTU ), FD_SHRED_BATCH_WMARK_NORMAL   );
+FD_STATIC_ASSERT( ( FD_SHRED_BATCH_WMARK_CHAINED  ) > ( FD_POH_SHRED_MTU ), FD_SHRED_BATCH_WMARK_CHAINED  );
+FD_STATIC_ASSERT( ( FD_SHRED_BATCH_WMARK_RESIGNED ) > ( FD_POH_SHRED_MTU ), FD_SHRED_BATCH_WMARK_RESIGNED );
```
ptaffet-jump, line 88: All these static asserts are great! Thanks!

src/disco/shred/fd_shred_batch.h
```diff
@@ -0,0 +1,124 @@
...
+   further details).  In order to support all three, we allocate the
+   largest of them (i.e. normal).  TODO once normal FEC sets have
+   been deprecated, replace NORMAL with CHAINED payload size in the
+   calculation of FD_SHRED_BATCH_RAW_BUF_SZ below. */
+#define FD_SHRED_BATCH_RAW_BUF_SZ     ( FD_SHRED_BATCH_FEC_SETS_MAX * FD_SHREDDER_NORMAL_FEC_SET_PAYLOAD_SZ )
+/* Validate batch raw buffer size.  Note that (chained+)resigned
+   needs to take into considerations the watermark regression. */
+FD_STATIC_ASSERT( ( FD_SHRED_BATCH_RAW_BUF_SZ ) >= ( FD_SHRED_BATCH_WMARK_NORMAL   + FD_POH_SHRED_MTU ), FD_SHRED_BATCH_RAW_BUF_SZ );
+FD_STATIC_ASSERT( ( FD_SHRED_BATCH_RAW_BUF_SZ ) >= ( FD_SHRED_BATCH_WMARK_CHAINED  + FD_POH_SHRED_MTU ), FD_SHRED_BATCH_RAW_BUF_SZ );
+FD_STATIC_ASSERT( ( FD_SHRED_BATCH_RAW_BUF_SZ ) >= ( FD_SHRED_BATCH_WMARK_RESIGNED + FD_POH_SHRED_MTU + FD_SHRED_BATCH_RESIGNED_WMARK_REGRESSION ), FD_SHRED_BATCH_RAW_BUF_SZ );
```
ptaffet-jump, line 102: Maybe also add an assertion that `FD_SHRED_BATCH_WMARK_RESIGNED + FD_SHRED_BATCH_RESIGNED_WMARK_REGRESSION  == FD_SHRED_BATCH_WMARK_CHAINED`? I tend to think about it as the chained wmark and not in terms of the regression, but they're equivalent.

src/disco/shred/fd_shred_batch.h
```diff
@@ -0,0 +1,124 @@
...
+
+/* Each block is limited to 32k parity shreds.  The number of parity
+   shreds in each FEC set is always at least as many as the number of
+   data shreds, so we do not need to consider the data shreds limit.
+   Since each FEC set is now guaranteed to contain 32 shreds (note:
+   32 data shreds and 32 parity shreds), the total maximum number of
```
ptaffet-jump:
```suggestion
   Since each FEC set is now guaranteed to contain
   32 data shreds and 32 parity shreds, the maximum number of
```

src/disco/shred/fd_shred_batch.h
```diff
@@ -0,0 +1,124 @@
...
+   FEC sets is 1024 (32k/32).  We consider the payload capacity of
+   chained FEC set as the raw capacity (which is smaller than normal
+   FEC sets), and subtract from there the overheads.  The worst case
+   padding overhead (OHEAD_PAD) can be approximated as one where the
+   last FEC set in each batch contains 1 byte of data and the rest is
```
ptaffet-jump:
```suggestion
   padding overhead (OHEAD_PAD) occurs when the
   last FEC set in each batch contains 1 byte of data and the rest is
```

src/disco/shred/fd_shred_batch.h
```diff
@@ -0,0 +1,124 @@
...
+   all padded with 0s.  There is also a watermark regression overhead
+   (OHEAD_REG) for the last batch when transitioning from chained to
+   (chained+)resigned.  The batch header is 8 bytes, and the worst
+   case header overhead (OHEAD_HDR) can be approximated as one were
```
ptaffet-jump:
```suggestion
   case header overhead (OHEAD_HDR) can be approximated as one where
```

src/disco/shred/fd_shred_batch.h
```diff
@@ -0,0 +1,124 @@
...
+/* FD_SHRED_BATCH_BLOCK_DATA_OHEAD_PAD assumes that each batch may
+   have up to 2 FEC sets before the watermark, which yields (1024/2)
+   512 batches per block. */
+FD_STATIC_ASSERT( ( FD_SHRED_BATCH_FEC_SETS_WMARK ) == ( 2UL ), FD_SHRED_BATCH_FEC_SETS_WMARK );
+#define FD_SHRED_BATCH_BLOCK_DATA_OHEAD_PAD ( 512UL * ( FD_SHREDDER_CHAINED_FEC_SET_PAYLOAD_SZ - 1UL ) )
+#define FD_SHRED_BATCH_BLOCK_DATA_OHEAD_REG ( FD_SHRED_BATCH_FEC_SETS_MAX * ( FD_SHREDDER_CHAINED_FEC_SET_PAYLOAD_SZ - FD_SHREDDER_RESIGNED_FEC_SET_PAYLOAD_SZ ) )
+#define FD_SHRED_BATCH_BLOCK_DATA_OHEAD_HDR ( 1024UL * 8UL )
+#define FD_SHRED_BATCH_BLOCK_DATA_OHEAD_ALL ( FD_SHRED_BATCH_BLOCK_DATA_OHEAD_PAD + FD_SHRED_BATCH_BLOCK_DATA_OHEAD_REG + FD_SHRED_BATCH_BLOCK_DATA_OHEAD_HDR )
+#define FD_SHRED_BATCH_BLOCK_DATA_SZ_MAX    ( FD_SHRED_BATCH_BLOCK_DATA_SZ_RAW - FD_SHRED_BATCH_BLOCK_DATA_OHEAD_ALL )
+FD_STATIC_ASSERT( ( FD_SHRED_BATCH_BLOCK_DATA_SZ_RAW ) > ( FD_SHRED_BATCH_BLOCK_DATA_OHEAD_ALL ), FD_SHRED_BATCH_BLOCK_DATA_SZ_MAX );
```
ptaffet-jump: Note to self: review these carefully

src/disco/shred/fd_shred_tile.c
```diff
@@ -480,71 +464,112 @@ during_frag( fd_shred_ctx_t * ctx,
...
-          ulong payload_for_32_shreds = FD_SHREDDER_RESIGNED_FEC_SET_PAYLOAD_SZ;
-          if( FD_LIKELY( IS_FIREDANCER ) ) {
-            chained_merkle_root = NULL;
-            payload_for_32_shreds = FD_SHREDDER_NORMAL_FEC_SET_PAYLOAD_SZ;
-          }
+      if( FD_LIKELY( process_current_batch )) {
 
-          if( FD_UNLIKELY( entry_meta->block_complete && batch_sz < payload_for_32_shreds ) ) {
+        uchar * chained_merkle_root = ctx->chained_merkle_root;
+        ulong load_for_32_shreds = FD_SHREDDER_CHAINED_FEC_SET_PAYLOAD_SZ;
```
ptaffet-jump: `load_for_32_shreds` could be computed with `pending_batch_wmark` just to simplify the code a tiny bit

src/disco/shred/fd_shred_tile.c
```diff
@@ -480,71 +464,112 @@ during_frag( fd_shred_ctx_t * ctx,
...
-              padding with 0s. Because the last FEC set is "oddly sized"
-              we only expect this code path to execute for blocks
-              containing less data than can fill 32 data shred payloads
-              (hence FD_UNLIKELY).
+        /* Batch and padding size calculation. */
+        ulong batch_sz        = sizeof(ulong) + ctx->pending_batch.pos; /* without padding */
+        ulong last_load_sz    = batch_sz;
+        while( last_load_sz  >= load_for_32_shreds ) { last_load_sz -= load_for_32_shreds; }
+        ulong padding_sz      = load_for_32_shreds - last_load_sz;
+        ulong batch_sz_padded = batch_sz + padding_sz;
```
ptaffet-jump: You can just replace all this with `batch_sz_padded = load_for_32_shreds * ( (batch_sz + load_for_32_shreds - 1 )/load_for_32_shreds)`
Then `padding_sz = batch_sz_padded - batch_sz`

src/disco/shred/fd_shred_tile.c
```diff
@@ -832,131 +872,147 @@ after_frag( fd_shred_ctx_t *    ctx,
...
+  /* Try to distribute shredded txn count across the fec sets.
+     This is an approximation, but it is acceptable. */
+  ulong shredded_txn_cnt_per_fec_set = (ulong)( ctx->shredded_txn_cnt / ctx->send_fec_set_cnt );
+  ulong shredded_txn_cnt [ FD_SHRED_BATCH_FEC_SETS_MAX ];
+  ulong shredded_txn_cnt_remain = ctx->shredded_txn_cnt;
+  for( ulong k=0; k<ctx->send_fec_set_cnt; k++ ) {
+    shredded_txn_cnt[ k ] = shredded_txn_cnt_per_fec_set;
+    shredded_txn_cnt_remain -= shredded_txn_cnt_per_fec_set;
+  }
+  shredded_txn_cnt[ ctx->send_fec_set_cnt - 1UL ] += shredded_txn_cnt_remain;
```
ptaffet-jump: You can get rid of this and just compute it one at a time inside the fset_k loop below since the logic is so simple

src/disco/shred/fd_shred_batch.h
jvarela-jump: Done.

src/disco/shred/fd_shred_tile.c
```diff
@@ -1247,7 +1305,10 @@ populate_allowed_fds( fd_topo_t const *      topo,
   return out_cnt;
 }
 
-#define STEM_BURST (5UL)
+/* Excluding net_out (where the link is unreliable), STEM_BURST
+   needs to guarantee enough credits for the worst case:
+   1<repair> + FD_SHRED_BATCH_FEC_SETS_MAX*max(1<repair>,4<store>) */
+#define STEM_BURST (1UL+FD_SHRED_BATCH_FEC_SETS_MAX*4UL)
```
ptaffet-jump: Okay, I might have made a mistake, but I think the correct value of this is `FD_SHRED_BATCH_FEC_SETS_MAX*2UL`

There are 4 cases to consider here (IN_KIND_NET/IN_KIND_POH) x (Frankendancer/Firedancer)

In the NET case, we have to send at most one microblock.  For Frankendancer, that can be 4 frags to store.  For firedancer, it's one frag for the shred to repair, and then another frag to repair for the FEC set.
In the POH case, for Frankendancer, there might be `FD_SHRED_BATCH_FEC_SETS_MAX` FEC sets, but we know they're 32:32, which means only two shred34s. For firedancer, it's  `FD_SHRED_BATCH_FEC_SETS_MAX` frags to repair (one per FEC set).

Can you double check all this logic please?

src/disco/shred/fd_shred_batch.h
```diff
@@ -0,0 +1,130 @@
...
+#define FD_SHRED_BATCH_RAW_BUF_SZ     ( FD_SHRED_BATCH_FEC_SETS_MAX * FD_SHREDDER_NORMAL_FEC_SET_PAYLOAD_SZ )
+/* Validate batch raw buffer size.  Note that (chained+)resigned
+   needs to take into considerations the watermark regression. */
+FD_STATIC_ASSERT( ( FD_SHRED_BATCH_RAW_BUF_SZ ) >= ( FD_SHRED_BATCH_WMARK_NORMAL   + FD_POH_SHRED_MTU ), FD_SHRED_BATCH_RAW_BUF_SZ );
+FD_STATIC_ASSERT( ( FD_SHRED_BATCH_RAW_BUF_SZ ) >= ( FD_SHRED_BATCH_WMARK_CHAINED  + FD_POH_SHRED_MTU ), FD_SHRED_BATCH_RAW_BUF_SZ );
+FD_STATIC_ASSERT( ( FD_SHRED_BATCH_RAW_BUF_SZ ) >= ( FD_SHRED_BATCH_WMARK_RESIGNED + FD_POH_SHRED_MTU + FD_SHRED_BATCH_RESIGNED_WMARK_REGRESSION ), FD_SHRED_BATCH_RAW_BUF_SZ );
+
+/* Each block is limited to 32k parity shreds.  The number of parity
+   shreds in each FEC set is always at least as many as the number of
+   data shreds, so we do not need to consider the data shreds limit.
```
ptaffet-jump:
```suggestion
/* Each block is limited to 32k parity shreds.
```

ptaffet-jump: Right now, we always produce the same number of data shreds and parity shreds, so we only need to consider the parity shred limit.

src/disco/shred/fd_shred_batch.h
```diff
@@ -0,0 +1,130 @@
...
+   Since each FEC set is now guaranteed to contain 32 data shreds and
+   32 parity shreds, the maximum number of FEC sets is 1024 (32k/32).
+   We consider the payload capacity of chained FEC set as the raw
+   capacity (which is smaller than normal FEC sets), and subtract from
+   there the overheads.  The worst case padding overhead (OHEAD_PAD)
+   occurs when the last FEC set in each batch contains 1 byte of data
```
ptaffet-jump: In the normal (non-last-batch) case, each batch is typically two FEC sets but can be one FEC set, but the worst case occurs when each batch has two FEC sets, of which the second is entirely padding other than a single byte.

src/disco/shred/fd_shred_batch.h
jvarela-jump: This has been simplified, and documented accordingly. Please double-check.

src/disco/shred/fd_shred_tile.c
jvarela-jump: Done.

src/disco/shred/fd_shred_tile.c
jvarela-jump: Done.

src/disco/shred/fd_shred_tile.c
jvarela-jump: Done.

src/disco/shred/fd_shred_tile.c
jvarela-jump: Yes, I double checked, and that seems correct. I have updated STEM_BURST and the explanation accordingly.

src/disco/shred/fd_shred_batch.h
jvarela-jump: Done. The documentation has been expanded as well.

src/disco/shred/fd_shred_tile.c
```diff
@@ -480,71 +464,104 @@ during_frag( fd_shred_ctx_t * ctx,
...
+        ctx->pending_batch.microblock_cnt += 1UL;
+        ctx->pending_batch.txn_cnt        += microblock->txn_cnt;
+      }
 
-          /* We sized this so it fits in one FEC set */
-          long shredding_timing =  -fd_tickcount();
+      if( FD_LIKELY( process_current_batch )) {
+        /* Batch and padding size calculation. */
+        ulong batch_sz        = sizeof(ulong) + ctx->pending_batch.pos; /* without padding */
+        ulong batch_sz_padded = load_for_32_shreds * (ulong)( ( batch_sz + load_for_32_shreds - 1 ) / load_for_32_shreds );
```
ptaffet-jump:
```suggestion
        ulong batch_sz_padded = load_for_32_shreds * ( ( batch_sz + load_for_32_shreds - 1UL ) / load_for_32_shreds );
```

ptaffet-jump: I think, right? Did the compiler complain and require the cast?

src/disco/shred/fd_shred_tile.c
```diff
@@ -832,131 +864,142 @@ after_frag( fd_shred_ctx_t *    ctx,
...
-  fd_shred34_t * s34 = ctx->shred34 + 4UL*ctx->send_fec_set_idx;
+  if( FD_UNLIKELY( ctx->send_fec_set_cnt==0UL ) ) return;
 
-  s34[ 0 ].shred_cnt =                         fd_ulong_min( set->data_shred_cnt,   34UL );
-  s34[ 1 ].shred_cnt = set->data_shred_cnt   - fd_ulong_min( set->data_shred_cnt,   34UL );
-  s34[ 2 ].shred_cnt =                         fd_ulong_min( set->parity_shred_cnt, 34UL );
-  s34[ 3 ].shred_cnt = set->parity_shred_cnt - fd_ulong_min( set->parity_shred_cnt, 34UL );
+  /* Try to distribute shredded txn count across the fec sets.
+     This is an approximation, but it is acceptable. */
+  ulong shredded_txn_cnt_per_fec_set  = (ulong)( ctx->shredded_txn_cnt / ctx->send_fec_set_cnt );
```
ptaffet-jump: Same here. When I see the `ulong` cast for something that I thought was `ulong/ulong`, it makes me wonder if maybe one the types is wrong. I'd just delete the cast.

ptaffet-jump (commented):
Looking great! Two nits, and then I'll stamp.

src/disco/shred/fd_shred_tile.c
jvarela-jump: Ok, done.

ptaffet-jump (approved):

PR #5361, opened 2025-06-13
send: quic wiring, votes landing

akhinvasara-jumptrading:
This PR updates the send tile to use quic, successfully landing votes. It merges in the quic wiring - we can then iterate on this in smaller PRs. It also cleans up the send_tile further, and provides cleaner metrics.

This includes a temp commit by @ravyu-jump to provide the correct contact info for quic TPU, a stop-gap for the ongoing gossip re-write. May be best to review changes separately for the two commits.


ravyu-jump:
LGTM on the gossip parts

jumpsiegel (dismissed):
Lets land this so that we can start to belt sand it...

PR #5364, opened 2025-06-14
deps: add auto install flag support for default case

ultrasilicon:
# Description
This PR is pretty minor, it makes running `./deps.sh` respect `FD_AUTO_INSTALL_PACKAGES` flag, so that it doesn't ask for user confirmation.

Before: 
```sh
$ FD_AUTO_INSTALL_PACKAGES=1 ./deps.sh 
[~] This will fetch, build, and install Firedancer's dependencies into /opt/firedancer/opt
[~] For help, run: ./deps.sh help

[~] Running ./deps.sh fetch check install
[?] Continue? (y/N)
```

After: 
```sh
$ FD_AUTO_INSTALL_PACKAGES=1 ./deps.sh 
[~] This will fetch, build, and install Firedancer's dependencies into /opt/firedancer/opt
[~] For help, run: ./deps.sh help

[~] Running ./deps.sh fetch check install
Submodule 'agave' (https://github.com/firedancer-io/agave.git) registered for path 'agave'
Cloning into '/home/echo/repo/firedancer/agave'...
    ^silently cotinues deps setup
```

This is particularly useful for unattended Firedancer deployment with automation tools like Ansible.

### DIFF
--- a/deps.sh
+++ b/deps.sh
@@ -542,9 +542,9 @@ index d7c8a12..8cfff9b 100644
 --- a/db/blob/blob_file_meta.h
 +++ b/db/blob/blob_file_meta.h
 @@ -5,6 +5,7 @@
- 
+
  #pragma once
- 
+
 +#include <cstdint>
  #include <cassert>
  #include <iosfwd>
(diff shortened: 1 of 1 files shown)


deps.sh
```diff
@@ -542,9 +542,9 @@ index d7c8a12..8cfff9b 100644
 --- a/db/blob/blob_file_meta.h
 +++ b/db/blob/blob_file_meta.h
 @@ -5,6 +5,7 @@
- 
+
  #pragma once
-
```
ripatel-fd: Could you undo the whitespace changes here? This is part of a patch file

ultrasilicon:
@ripatel-fd Thank you for catching that. I have updated the commit to remove deltas that removed those spaces. Please check again. :D

PR #5368, opened 2025-06-16
small changes needed for geyser

asiegel-jt:


src/app/grpc-geyser/Local.mk
```diff
@@ -0,0 +1,27 @@
+ifdef FD_GEYSER
+
+# Location of grpc installation (https://github.com/grpc/grpc)
+PKG = /data/asiegel/pkg
+
```
jumpsiegel: need to put this into deps.sh

src/app/grpc-geyser/Local.mk
```diff
@@ -0,0 +1,27 @@
+ifdef FD_GEYSER
```
jumpsiegel: Is there a EXTRAS="geyser" to bring this in?  so that it is not on by default?

asiegel-jt: you pass -DFD_GEYSER=1 to make

src/app/grpc-geyser/Local.mk
asiegel-jt: there would be conflicting versions of libssl and libcrypto

firedancer-admin:
We're gonna move 99% of this into [its own repo ](https://github.com/firedancer-io/fd-yellowstone)

jumpsiegel (approved):

PR #5373, opened 2025-06-17
funk: default config use correct key

two-heart:
The parser will otherwise just set heap_size_gib to 0.


topointon-jump (approved):
Thanks for catching this!

kbhargava-jump (approved):
thank you

PR #5382, opened 2025-06-17
snapshot: add support for custom http header

topointon-jump:
Some private RPC nodes require an extra non-standard HTTP header for token authentication. Unfortunately, this is not a standard API bearer token format, but a custom http header.

### DIFF
--- a/src/app/firedancer-dev/config/default.toml
+++ b/src/app/firedancer-dev/config/default.toml
@@ -28,6 +28,7 @@
     [tiles.replay]
         snapshot_url = "http://{validator_ip}:8899/snapshot.tar.bz2"
         incremental_url = "http://{validator_ip}:8899/incremental-snapshot.tar.bz2"
+        snapshot_additional_http_header = ""
     [tiles.metric]
         prometheus_listen_address = "0.0.0.0"
         prometheus_listen_port = 7999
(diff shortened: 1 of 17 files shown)


src/app/firedancer/config/default.toml
```diff
@@ -1081,6 +1081,14 @@ user = ""
         snapshot = ""
         snapshot_url = ""
 
+        # An additional, optional, custom HTTP header which all HTTP snapshot
+        # requests will contain.
+        #
+        # Some private RPC nodes require an extra non-standard HTTP header
+        # to pass in an API token. This is not a standard API bearer token
```
mmcgee-jump: Not quite following this comment. Does this string specify the header, or the value passed in some already known header? What value is set in there?

src/app/firedancer-dev/config/default.toml
```diff
@@ -28,6 +28,7 @@
     [tiles.replay]
         snapshot_url = "http://{validator_ip}:8899/snapshot.tar.bz2"
         incremental_url = "http://{validator_ip}:8899/incremental-snapshot.tar.bz2"
+        snapshot_custom_http_header = ""
```
ripatel-fd: This should just be `snapshot_auth_token`. It's safe to assume that the endpoint uses Bearer auth.

ripatel-fd: A lot of RPC services also support `?auth=xxx` style syntax btw.

topointon-jump:
Closing this as we just won't use this RPC node

PR #5388, opened 2025-06-18
flamenco: fix vm interp jump tables

kbhargava-jump:


yufeng-jump (approved):

kbhargava-jump:
> I think having the ifs makes the tables more readable (we could of course modify the tables).

I agree with this point, I was also thinking of just hard coding within each of the tables but having the SIMD and if statements, makes it more understandable.

PR #5391, opened 2025-06-19
firedancer-dev: update shred version for v2.3.0 cluster

alpeng-jump:


src/app/firedancer-dev/config/private.toml
```diff
@@ -27,7 +27,7 @@
         prometheus_listen_port = 7999
 [consensus]
     vote = false
-    expected_shred_version = 7388
+    expected_shred_version = 3019
```
ripatel-fd, line 30: Just pointing out that you can discover the shred version from an entrypoint. So this config default is redundant.

kbhargava-jump (approved):

PR #5394, opened 2025-06-20
Update tuning.md

Leomrqdd:
Hi,

I have seen a small discrepancy between the default.toml file provided in docs.firedancer.io that says :  How many bank tiles to run.  Should be set to 4 for perf and balanced scheduling modes. 

And the tuning part that says about bank tiles : 
Handles 20-40k TPS per tile, with diminishing returns from adding more tiles. Designed to scale out for future network conditions, but 2 tiles is enough to handle current mainnet-beta conditions.

One source says that 4 is best and the other one 2 with a diminishing returns from adding more tiles.

Hope I understand it well and it can help others. 

Capture d’écran 2025-06-20 à 17 51 47
Capture d’écran 2025-06-20 à 17 51 38

Best,

Léo


ripatel-fd:
Good point @Leomrqdd, thanks. @mmcgee-jump can you take a look at this?

Issue #5393, opened 2025-06-19
gui crash due to optimistically confirmed assumption


mmcgee-jump:
```
ERR   2025-06-18 20:30:13.492917177 GMT+00 3771788:3771821 solana:host651516:23  fd1:[group]:gui:0 src/disco/gui/fd_gui.c(1403)[fd_gui_handle_optimistically_confirmed_slot]: _slot 347673679 i 863999 we expect parent_slot 346809680 got slot->slot 347673680
ERR   2025-06-18 20:30:13.497326328 GMT+00 3771760:3771755 solana:host651516:f0  fd1:[group]:pidns src/app/shared/commands/run/run.c(419)[main_pid_namespace]: tile gui:0 exited with signal 31 (SIGSYS-Bad system call)
```

PR #5395, opened 2025-06-20
gui: fix invariant check

jherrera-jump:
resolves https://github.com/firedancer-io/firedancer/issues/5393

### DIFF
--- a/src/disco/gui/fd_gui.c
+++ b/src/disco/gui/fd_gui.c
@@ -1258,12 +1258,17 @@ fd_gui_handle_reset_slot( fd_gui_t * gui,
   ulong last_slot = _slot;
   long last_published = gui->slots[ _slot % FD_GUI_SLOTS_CNT ]->completed_time;
 
-  for( ulong i=0UL; i<fd_ulong_min( _slot+1, 750UL ); i++ ) {
+  for( ulong i=0UL; i<fd_ulong_min( _slot+1, FD_GUI_SLOT_DURATION_HISTORY_SAMPLE_CNT ); i++ ) {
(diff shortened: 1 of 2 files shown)


src/disco/gui/fd_gui.c
```diff
@@ -1377,7 +1382,12 @@ fd_gui_handle_rooted_slot( fd_gui_t * gui,
     fd_gui_slot_t * slot = gui->slots[ parent_idx ];
     if( FD_UNLIKELY( slot->slot==ULONG_MAX) ) break;
 
-    if( FD_UNLIKELY( slot->slot!=parent_slot ) ) {
```
mmcgee-jump, line 1380: This is required for correctness, otherwise we do not notify the full parentage chain of state changes correctly. I also don't see how it could happen, the comment is not specific.

The error case seems to be that, before a slot roots, the buffer wraps all the way around and the parent slot has already been overwritten by a new unrooted slot. That would require essentially `864000` slots without rooting, which does not seem reasonable?

jherrera-jump: Yeah you're right. I'll leave a comment explaining that this assumes FD_GUI_SLOTS_CNT >= the longest possible unrooted-slot chain.

jherrera-jump: Actually, I think I need to think more about this

jherrera-jump:
closing because this was definitley not the issue.

There is a bug where the gui tile doesn't recieve rooted slot updates from agave but I haven't been able to reproduce it, and it only seems to affect unstaked nodes.

PR #5396, opened 2025-06-20
quic: separate out rtt API

ripatel-fd:
- Add tests
- Remove unused sched_granularity_ticks setting
- Add RTT metrics to bundle


mmcgee-jump (approved):

src/waltz/quic/fd_quic_private.h
```diff
@@ -488,11 +464,12 @@ fd_quic_calc_expiry( fd_quic_conn_t * conn, ulong now ) {
      a timer for the PTO period as follows:
      PTO = smoothed_rtt + max(4*rttvar, kGranularity) + max_ack_delay  */
 
-  fd_quic_conn_rtt_t * rtt = conn->rtt;
+  fd_rtt_estimate_t * rtt = conn->rtt;
 
-  ulong duration = (ulong)( rtt->smoothed_rtt
-                        + fmaxf( 4.0f * rtt->var_rtt, rtt->sched_granularity_ticks )
```
akhinvasara-jumptrading, line 494: why do we no longer use sched_granularity_ticks ?

src/waltz/quic/fd_quic.c
```diff
@@ -4420,19 +4426,19 @@ fd_quic_conn_create( fd_quic_t *               quic,
...
                                                /* updated when we get transport parameters */
-  rtt->smoothed_rtt             = FD_QUIC_INITIAL_RTT_US * (float)quic->config.tick_per_us;
-  rtt->latest_rtt               = FD_QUIC_INITIAL_RTT_US * (float)quic->config.tick_per_us;
-  rtt->min_rtt                  = FD_QUIC_INITIAL_RTT_US * (float)quic->config.tick_per_us;
-  rtt->var_rtt                  = FD_QUIC_INITIAL_RTT_US * (float)quic->config.tick_per_us * 0.5f;
-  rtt->rtt_period_ticks         = FD_QUIC_RTT_PERIOD_US  * (float)quic->config.tick_per_us;
+  rtt->smoothed_rtt              = FD_QUIC_INITIAL_RTT_US * (float)quic->config.tick_per_us;
+  rtt->latest_rtt                = FD_QUIC_INITIAL_RTT_US * (float)quic->config.tick_per_us;
+  rtt->min_rtt                   = FD_QUIC_INITIAL_RTT_US * (float)quic->config.tick_per_us;
+  rtt->var_rtt                   = FD_QUIC_INITIAL_RTT_US * (float)quic->config.tick_per_us * 0.5f;
```
akhinvasara-jumptrading, line 4440: Can we use the sample API here to set these values? Unless we specifically wanted to leave is_rtt_valid false?

akhinvasara-jumptrading: Turns out we do specifically want to leave is_rtt_valid as false:
From [RFC9002](https://datatracker.ietf.org/doc/html/rfc9002#section-5.3-12)
On the first RTT sample after initialization, the estimator is reset using that sample. This ensures that the estimator retains no history of past samples.

Issue #4341, opened 2025-02-28
ibverbs tile


ripatel-fd:
Use libibverbs in 'raw packet' / Ethernet mode for high performance mlx5 operation

Low priority R&D item

ripatel-fd:
Done in https://github.com/firedancer-io/firedancer/pull/10815

PR #5399, opened 2025-06-20
InfiniBand support

ripatel-fd:
Closes https://github.com/firedancer-io/firedancer/issues/4341

### DIFF
--- a/src/disco/net/fd_net_tile_topo.c
+++ b/src/disco/net/fd_net_tile_topo.c
@@ -81,7 +81,6 @@ setup_ibeth_tile( fd_topo_t *             topo,
   fd_topob_tile_uses( topo, tile, umem_obj, FD_SHMEM_JOIN_MODE_READ_WRITE );
   fd_pod_insertf_ulong( topo->props, umem_obj->id, "net.%lu.umem", 0UL ); /* FIXME multi queue support */
 
-  tile->ibeth.umem_dcache_obj_id = umem_obj->id;
   fd_cstr_fini( fd_cstr_append_cstr_safe(
       fd_cstr_init( tile->ibeth.if_name ),
       net_cfg->interface,
--- a/src/disco/net/ibeth/fd_ibeth_tile.c
+++ b/src/disco/net/ibeth/fd_ibeth_tile.c
@@ -243,6 +243,18 @@ registered:
   if(0){}
 }
 
+static void
+rxq_assign_all( fd_ibeth_tile_t * ctx,
+                fd_topo_t *       topo,
+                fd_topo_tile_t *  tile ) {
+  rxq_assign( ctx, topo, tile, DST_PROTO_TPU_UDP,  "net_quic",   tile->ibeth.net.legacy_transaction_listen_port );
+  rxq_assign( ctx, topo, tile, DST_PROTO_TPU_QUIC, "net_quic",   tile->ibeth.net.quic_transaction_listen_port   );
+  rxq_assign( ctx, topo, tile, DST_PROTO_SHRED,    "net_shred",  tile->ibeth.net.shred_listen_port              );
+  rxq_assign( ctx, topo, tile, DST_PROTO_GOSSIP,   "net_gossip", tile->ibeth.net.gossip_listen_port             );
+  rxq_assign( ctx, topo, tile, DST_PROTO_REPAIR,   "net_shred",  tile->ibeth.net.repair_intake_listen_port      );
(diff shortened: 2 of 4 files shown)


src/disco/net/ibeth/fd_ibeth_tile.c
```diff
@@ -0,0 +1,811 @@
...
+  int const hdr_ok =
+    ( fd_ushort_bswap( l2->net_type )==FD_ETH_HDR_TYPE_IP ) &
+    ( FD_IP4_GET_VERSION( *l3 )==4 ) &
+    ( l3->protocol==FD_IP4_HDR_PROTOCOL_UDP );
+
+  ushort const net_dport = fd_ushort_bswap( l4->net_dport );
+  int out_idx = -1;
+  for( ulong i=0UL; i<FD_IBETH_UDP_PORT_MAX; i++ ) {
+    if( ctx->dst_ports[ i ]==net_dport ) {
+      out_idx = ctx->dst_out_idx[ i ];
```
0x0ece, line 623: We should centralize this code across the 3 net tile implementations (it's diverging already).
(But we could also centralize in a separated PR)

ripatel-fd: @0x0ece I know, good point. I had a separate PR to do this. But I don't want to affect the production network stack with this experimental feature yet. Let's do it in a follow up PR.

src/disco/net/fd_net_router.h
```diff
@@ -0,0 +1,151 @@
...
+
+  uint neigh_ip = next_hop->ip4_gw;
+  if( !neigh_ip ) neigh_ip = dst_ip;
+
+  fd_neigh4_hmap_query_t neigh_query[1];
+  int neigh_res = fd_neigh4_hmap_query_try( ctx->neigh4, &neigh_ip, NULL, neigh_query, 0 );
+  if( FD_UNLIKELY( neigh_res!=FD_MAP_SUCCESS ) ) {
+    /* Neighbor not found */
+    if( FD_LIKELY( ctx->neigh4_solicit->mcache ) ) {
+      fd_netlink_neigh4_solicit(
```
ripatel-fd, line 95: This will generate neighbor solicit request for interfaces that the topo cannot route to. (Missing if_idx check...)

src/disco/net/ibeth/fd_ibeth_tile.c
ripatel-fd: https://github.com/firedancer-io/firedancer/issues/5490

src/disco/net/fd_net_tile_topo.c
```diff
@@ -64,6 +64,41 @@ setup_sock_tile( fd_topo_t *             topo,
...
+  fd_topob_link( topo, "net_netlnk", "net_netlnk", 128UL, 0UL, 0UL );
+  fd_topob_tile_in(  topo, "netlnk", 0UL, "metric_in", "net_netlnk", 0UL, FD_TOPOB_UNRELIABLE, FD_TOPOB_POLLED );
+  fd_topob_tile_out( topo, "ibeth",  0UL,              "net_netlnk", 0UL );
+  fd_netlink_topo_join( topo, netlink_tile, tile );
+
+  fd_topo_obj_t * umem_obj = fd_topob_obj( topo, "dcache", "net_umem" );
+  fd_topob_tile_uses( topo, tile, umem_obj, FD_SHMEM_JOIN_MODE_READ_WRITE );
+  fd_pod_insertf_ulong( topo->props, umem_obj->id, "net.%lu.umem", 0UL ); /* FIXME multi queue support */
+
+  tile->ibeth.umem_dcache_obj_id = umem_obj->id;
```
two-heart: Duplicate assignment

src/app/shared/commands/configure/sysctl.c
```diff
@@ -177,7 +177,7 @@ check( config_t const * config ) {
     sock_params[ 1 ].value = config->net.socket.send_buffer_size;
     r = check_param_list( sock_params );
   } else {
-    FD_LOG_ERR(( "unknown net provider: %s", config->net.provider ));
+    CONFIGURE_OK();
```
two-heart, line 180: Is there a problem with checking iverbs the same way as the others?

two-heart (changes requested):

src/app/shared/commands/configure/sysctl.c
ripatel-fd: No. The check stage should get skipped by default unless specifically requested.

PR #5407, opened 2025-06-23
txncache: fix constipation locking and initialization

two-heart:


yufeng-jump:
Closing because constipation is being torn out in #4483

PR #5408, opened 2025-06-23
txncache: move log before bailout, otherwise it is dead

two-heart:


yufeng-jump:
Closing because txncache is being rewritten in #4483

PR #5424, opened 2025-06-24
flamenco: update fd_executor_validate_account_locks agave perma link

two-heart:
the previous one pointed to an old implemenation with different logic, which confused me


src/flamenco/runtime/fd_executor.c
```diff
@@ -643,7 +643,7 @@ fd_executor_validate_account_locks( fd_exec_txn_ctx_t const * txn_ctx ) {
     }
   }
 
-  /* https://github.com/anza-xyz/agave/blob/ced98f1ebe73f7e9691308afa757323003ff744f/sdk/src/transaction/sanitized.rs#L286-L288 */
+  /* https://github.com/anza-xyz/agave/blob/838c1952595809a31520ff1603a13f2c9123aa51/accounts-db/src/account_locks.rs#L126-L128 */
```
kbhargava-jump: since we are changing this, can we please use a tagged version instead of a commit

two-heart: yeah, sure

PR #5458, opened 2025-06-25
bank, gui: Use correct err for bundle peer transactions

jherrera-jump:


jherrera-jump:
Change will be included in https://github.com/firedancer-io/firedancer/pull/5948

PR #5459, opened 2025-06-25
quic: probe gen_ack_frames

akhinvasara-jumptrading:
This PR adds a probe in gen_ack frames that reports how many ack ranges were sent, whether the buffer was fully flushed, and the enc level.


src/waltz/quic/fd_quic_ack_tx.c
```diff
@@ -137,5 +139,8 @@ fd_quic_gen_ack_frames( fd_quic_ack_gen_t * gen,
     FD_ACK_DEBUG( FD_LOG_DEBUG(( "Not all ACK frames were flushed" )); )
   }
 
+  const int flushed = !gen->is_elicited;
```
ripatel-fd:
```suggestion
  int const flushed = !gen->is_elicited;
```
type always first

PR #5462, opened 2025-06-26
flamenco: add missing err check in extend_lookup_table

two-heart:


two-heart:
@mjain-jump backtest ci is flaky (observed the same for unrelated doc change PRs)

PR #5464, opened 2025-06-26
Fix broken link to getting-started.md in system-xdp documentation

VolodymyrBg:
Updated the link to the getting-started guide in doc/system-xdp.md to point to the correct location at ../book/guide/getting-started.md. This resolves a broken link error and ensures users can access the required setup instructions.


ripatel-fd:
Thank you for the contribution. This documentation page is outdated, so we're removing it entirely instead: https://github.com/firedancer-io/firedancer/pull/5484

PR #5466, opened 2025-06-26
backtest: add capture section to config for solcap and protobufs

kbhargava-jump:


src/discof/exec/fd_exec_tile.c
```diff
@@ -140,6 +140,8 @@ struct fd_exec_tile_ctx {
 
   /* Pairs len is the number of accounts to hash. */
   ulong                 pairs_len;
+
+  fd_capture_ctx_t *    capture_ctx;
```
kbhargava-jump, line 144: temporary, will remove from exec tile in a later PR

PR #5480, opened 2025-06-27
runtime: fix restrict violation in fd_log_address_type

topointon-jump:
We were allocating both `out` and `base_addr` from the same `spad` allocator, and then passing in `base_addr` to `snprintf` which was writing back into `out`. This meant the compiler couldn't guarantee that `out` and `base_addr` didn't overlap. To fix this we use a stack-allocated buffer for `base_addr`. For correctness we also change fd_log_address_type to be a macro.

Fixes some GCC builds.


lidatong (approved):

mjain-jump (approved):
thank you!

PR #5481, opened 2025-06-27
quic: Resurrect prq svc queue 

akhinvasara-jumptrading:
This PR re-applies the previously reverted commit to use a prq as the underlying data structure for the quic service queue. It also fixes the associated bugs that caused the initial reversion. This has been tested on both testnet and mainnet validators.

The first commit will be merged separately in https://github.com/firedancer-io/firedancer/pull/6231.


src/waltz/quic/fd_quic.c
```diff
@@ -2020,7 +2020,6 @@ fd_quic_lazy_ack_pkt( fd_quic_t *           quic,
     ( ( pkt->enc_level == fd_quic_enc_level_initial_id   ) |
       ( pkt->enc_level == fd_quic_enc_level_handshake_id ) );
   if( ack_sz_threshold_hit | force_instant_ack ) {
-    conn->unacked_sz = 0UL;
```
akhinvasara-jumptrading, line 2132: This is already reset after the acks are actually sent out, and shouldn't be here.

src/waltz/quic/fd_quic.c
```diff
@@ -2462,6 +2467,11 @@ fd_quic_process_packet( fd_quic_t * quic,
       }
 
       rc = fd_quic_process_quic_packet_v1( quic, &pkt, cur_ptr, cur_sz );
+      if( FD_UNLIKELY( fd_quic_svc_cnt_events( state->svc_timers ) != quic->metrics.conn_active_cnt ) ) {
+        FD_LOG_WARNING(( "only %lu out of %lu connections are in timer",
```
akhinvasara-jumptrading: Would love some feedback on these checks/logs - it's a good invariant to 'test' live, and fairly cheap, but probs not super useful for operators?

src/waltz/quic/fd_quic.c
```diff
@@ -3573,7 +3593,8 @@ fd_quic_conn_tx( fd_quic_t      * quic,
   ulong now = fd_quic_get_state( quic )->now;
 
   /* initialize expiry and tx_time */
-  fd_quic_pkt_meta_t pkt_meta_tmpl[1] = {{.expiry = now+500000000UL, .tx_time = now}};
+  ulong const retx_timeout = 1250UL*1000000UL; /* 1250M ticks ~ 500 ms?*/
```
akhinvasara-jumptrading: Depends on the tick rate. This should probably be replaced soon by the PTO stuff anyway, so thought simpler/safer for now to just hardcode ticks instead of throwing the conversion in there?? Open to changing that also!

src/waltz/quic/fd_quic.c
```diff
@@ -3951,6 +3971,7 @@ fd_quic_conn_service( fd_quic_t * quic, fd_quic_conn_t * conn, ulong now ) {
     case FD_QUIC_CONN_STATE_INVALID:
       /* fall thru */
     default:
+      FD_LOG_ERR(( "invalid conn state %u", conn->state ));
```
akhinvasara-jumptrading: Is it better to just crash here if quic is in some really weird state that should never happen? assuming yes, lmk if anyone disagrees.

ripatel-fd: Yes, but I'd crash with FD_LOG_CRIT so we get a core dump

ripatel-fd:
@akhinvasara-jumptrading Kevin is working on a new fast nanosecond-based clock source. This allows us to retire fd_tickcount(). We might want to wait for that feature to arrive before merging this, because it eliminates the time scaling class of bugs. (No strong opinion on this though. If you think it's a good idea to merge this now, we shall do so.)

src/waltz/quic/fd_quic.c
ripatel-fd: If an invariant is broken, we should always FD_LOG_CRIT (produces a core dump / easy to catch with a debugger).
FD_LOG_WARNING logs are diagnostics for operators. The operator would have no idea what to do with that log entry though.

src/waltz/quic/fd_quic.c
ripatel-fd: Uhh don't hardcode tickrate just use fd_tempo_tick_per_ns if needed

src/waltz/quic/fd_quic.c
ripatel-fd: This is sensitive code, if this is just a drive-by change unrelated to the PR, please ensure that this is already exhaustively tested.

akhinvasara-jumptrading: undoing for now

PR #5483, opened 2025-06-28
Cooperative scheduling

bzhangabylov-jump:
This PR implements cooperative scheduling for Firedancer tiles, allowing them to sleep when idle instead of busy-spinning to improve CPU utilization and power efficiency without losing latency when there is work to do.

### Changes:
- Added futex-based wait/wake mechanism to stem framework.

**Wake Mechanism (Producer Side):**

- Tiles determine which consumers use cooperative scheduling and track them via wake_out_cnt and wake_out arrays in the STEM_run function. Every tile implements this just in case it has consumers that futex_wait.
- Wake operations are amortized by adding them to the housekeeping range for consistent wake calls
- Producers must increment futex_flag when publishing (automatic with fd_stem_publish)

**Wait Mechanism (Consumer Side)**
- After FD_STEM_SPIN_THRESHOLD iterations without consuming work, tiles call futex_waitv to sleep
- futex_waitv waits on futex flags in mcache headers across all input links
- Tiles wake when:
- Any producer calls futex_wake (with INT_MAX to wake all waiters)
- The sleep timeout expires (typically set to housekeeping deadline)
- Custom deadline from get_tile_deadline callback (tile can set this to 0 to signal that it should not be slept)

**Configuration**
- Enable by adding the following to the config:
[power]
    low_power_mode = true
- This flag propagates through: config -> topology -> tile -> sleeps argument in STEM_run
- STEM_CAN_SLEEP macro in the tile defines whether the tile can be toggled to have cooperative scheduling.

**Supported Tiles**
- verify, dedup, bank, shred, sign, quick, resolv, store, pack, poh, plugin
- TODO: net, gui
- feature not turned on, but blocking logic added to: bundle, metric


src/disco/stem/fd_stem.c
```diff
@@ -154,6 +154,14 @@
 #error "STEM_BURST must be defined"
 #endif
 
+#ifndef STEM_ALWAYS_SPINNING
+#define STEM_ALWAYS_SPINNING 1
+#endif
+
+#ifndef STEM_TIME_BEFORE_WAKE
+#define STEM_TIME_BEFORE_WAKE (5000000000L) // 5 seconds
```
mmcgee-jump: Way too long (like many orders of magnitude). We should probably switch this to a loop spin counter as well (how many spins without publishing on that link, before we wake).

ripatel-fd:
Some very good changes in there, if you're not going to PR them yourself, I will definitely cherry-pick some stuff. Thanks for the great work @bzhangabylov-jump

PR #5487, opened 2025-06-28
Temporarily remove shredcap

ripatel-fd:


ripatel-fd:
Shredcap will only be removed once we have a pcap based replacement.

PR #5491, opened 2025-06-30
runtime: add missing unlock in fd_banks_get_bank

two-heart:


src/flamenco/runtime/fd_bank.c
```diff
@@ -445,6 +445,7 @@ fd_banks_get_bank( fd_banks_t * banks, ulong slot ) {
   bank = fd_banks_map_ele_query( bank_map, &slot, NULL, bank_pool );
   if( FD_UNLIKELY( !bank ) ) {
     FD_LOG_WARNING(( "Failed to get bank" ));
+    fd_rwlock_unread( &banks->rwlock );
```
lidatong, line 448: wait shouldnt the caller be retaining the read lock here? how does the pointer stay valid on return -- its not a copy?

lidatong: github's PR UI is idiotic and doesn't let me tag other lines, but referring to line 452 not this

ibhatt-jumptrading:
I spoke with charlie, the caller does indeed need locks when they use use a bank. You don't want a bank you are using to get published away

ibhatt-jumptrading:
Pr #5622 fixes this. Thanks @lidatong for finding this

PR #5494, opened 2025-06-30
added purify funktion to funk

asiegel-jt:


src/funk/fd_funk_rec.h
```diff
@@ -56,10 +56,6 @@ struct __attribute__((aligned(FD_FUNK_REC_ALIGN))) fd_funk_rec {
   uint  prev_idx;  /* Record map index of previous record in its transaction */
   uint  next_idx;  /* Record map index of next record in its transaction */
 
-  /* UNUSED: reserved for future use by the Funk LRU. Do not remove or modify. */
-  uint  accounts_lru_prev_idx; /* Record map idx of the next record in the accounts LRU dlist */
-  uint  accounts_lru_next_idx; /* Record map idx of the prev record in the accounts LRU dlist */
```
ripatel-fd, line 61: Maybe we should leave this in considering the sticky tape on it?

src/funk/run_reconnect.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+rm -f /mnt/.fd/.gigantic/reconnect_test /tmp/core.*
+echo "/tmp/core.%e.%p.%t" | sudo tee /proc/sys/kernel/core_pattern
```
ripatel-fd, line 4: Why do we need to change the core dump pattern?

jumpsiegel: because the default coredump pattern on my box and on alex's box was filtering out the cores

src/funk/fd_funk_rec.h
jumpsiegel: @topointon-jump ,  thoughts?

src/funk/fd_funk_txn.c
```diff
@@ -463,6 +463,12 @@ fd_funk_txn_update( fd_funk_t *                  funk,
   fd_funk_rec_pool_t * rec_pool = funk->rec_pool;
   fd_funk_txn_pool_t * txn_pool = funk->txn_pool;
 
+  int critical = (_dst_rec_head_idx == &funk->shmem->rec_head_idx);
+  if (critical) {
+    /* Smash the magic number to indicate that purify should fail */
+    funk->shmem->magic = FD_FUNK_MAGIC+1;
```
jumpsiegel: Could you make this "magic" be macro's or a function?  we might decide how to mark the critical section differently later

jumpsiegel (approved):

PR #5500, opened 2025-06-30
pack: pack is responsible for ending leader slots

mmcgee-jump:
In full Firedancer the bank forks objects are implicitly refcounted by the leader pipeline, so they are acquired by pack and released by poh, because of this pack must "flush" the pipeline and give poh a way to guarantee the bank fork is not in use when it releases, which it can do by exactly sequencing the done packing messaage (must arrive after the bank microblocks).

### DIFF
--- a/src/disco/gui/fd_gui_tile.c
+++ b/src/disco/gui/fd_gui_tile.c
@@ -43,7 +43,8 @@ extern uint  const fdctl_commit_ref;
 #define IN_KIND_PLUGIN    (0UL)
 #define IN_KIND_POH_PACK  (1UL)
 #define IN_KIND_PACK_BANK (2UL)
-#define IN_KIND_BANK_POH  (3UL)
+#define IN_KIND_PACK_POH  (3UL)
+#define IN_KIND_BANK_POH  (4UL)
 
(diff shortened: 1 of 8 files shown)


src/disco/gui/fd_gui_tile.c
```diff
@@ -511,6 +512,7 @@ unprivileged_init( fd_topo_t *      topo,
     if( FD_LIKELY( !strcmp( link->name, "plugin_out"     ) ) ) ctx->in_kind[ i ] = IN_KIND_PLUGIN;
     else if( FD_LIKELY( !strcmp( link->name, "poh_pack"  ) ) ) ctx->in_kind[ i ] = IN_KIND_POH_PACK;
     else if( FD_LIKELY( !strcmp( link->name, "pack_bank" ) ) ) ctx->in_kind[ i ] = IN_KIND_PACK_BANK;
+    else if( FD_LIKELY( !strcmp( link->name, "pack_poh" ) ) )  ctx->in_kind[ i ] = IN_KIND_PACK_POH;
```
ptaffet-jump, line 515:
```suggestion
    else if( FD_LIKELY( !strcmp( link->name, "pack_poh"  ) ) ) ctx->in_kind[ i ] = IN_KIND_PACK_POH;
```

PR #5503, opened 2025-07-01
chore: fix some minor issues in the comments

sellskin:
fix some minor issues in the comments


src/ballet/nanopb/pb_common.c
```diff
@@ -145,7 +145,7 @@ static void advance_iterator(pb_field_iter_t *iter)
 
         /* Add to fields.
          * The cast to pb_size_t is needed to avoid -Wconversion warning.
-         * Because the data is is constants from generator, there is no danger of overflow.
```
two-heart: We generally don't want to modify vendored files if not necessary.

two-heart (changes requested):

src/ballet/nanopb/pb_common.c
sellskin: Thanks. Reverted. 

Please review again.

PR #5506, opened 2025-07-01
Sysvar cache v2

ripatel-fd:
Centralize all sysvar accesses through a central write-through cache. 
The cache is position-independent, flat, self-contained, and thread-safe.
So it can be copied or mapped into different processes arbitrarily. 
Removes some bank fields that are redundantly stored as sysvars.


src/flamenco/runtime/sysvar/fd_sysvar_cache.h
```diff
@@ -0,0 +1,370 @@
...
+
+uchar const *
+fd_sysvar_cache_data_query(
+    fd_sysvar_cache_t const * sysvar_cache,
+    void const *              address, /* 32 bytes */
+    ulong *                   psz
+);
+
+/* fd_sysvar_cache_recover rebuilds the sysvar cache from the account
+   database.  Logs warnings in case sysvar datas fail to deserialize.
```
mmcgee-jump: Return int? Caller is going to want to terminate

ripatel-fd:
Turns out in order to comply with consensus, you have to support conflicting sysvar accounts and bank fields. This behavior is found on a real cluster

PR #5524, opened 2025-07-02
runtime: side effect free max calc

two-heart:
Compare with agave: https://github.com/anza-xyz/agave/blob/v2.0.1/programs/vote/src/vote_state/mod.rs#L1127

### DIFF
--- a/src/flamenco/runtime/program/fd_vote_program.c
+++ b/src/flamenco/runtime/program/fd_vote_program.c
@@ -1853,20 +1853,20 @@ process_vote_with_account( fd_borrowed_account_t *       vote_account,
       return FD_EXECUTOR_INSTR_ERR_CUSTOM_ERR;
     }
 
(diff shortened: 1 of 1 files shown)


two-heart:
@lidatong please review if I am correct here

src/flamenco/runtime/program/fd_vote_program.c
```diff
@@ -1853,20 +1853,20 @@ process_vote_with_account( fd_borrowed_account_t *       vote_account,
       return FD_EXECUTOR_INSTR_ERR_CUSTOM_ERR;
     }
 
-    ulong * max = deq_ulong_peek_head( vote->slots ) ? deq_ulong_peek_head( vote->slots ) : NULL;
```
ripatel-fd, line 1856: deque_peek_head is documented as never returning NULL until someone randomly hacked it ... Could you use the proper accessor? `!deque_empty` while you are at it. Thank you!

PR #5530, opened 2025-07-02
gui: add bounds check to peer messages

jherrera-jump:
waiting for testnet to realive to test this
 
resolves https://github.com/firedancer-io/auditor-internal/issues/303

agave [diff](https://github.com/firedancer-io/agave/compare/04b8dfe993b358194cf77ccaee1c69aad7dc74fb..jherrera/gui-valinfo-bounds-check)

### DIFF
--- a/src/disco/gui/fd_gui.c
+++ b/src/disco/gui/fd_gui.c
@@ -546,19 +546,23 @@ fd_gui_poll( fd_gui_t * gui ) {
 static void
 fd_gui_handle_gossip_update( fd_gui_t *    gui,
                              uchar const * msg ) {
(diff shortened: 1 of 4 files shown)


agave
```diff
@@ -1 +1 @@
-Subproject commit 04b8dfe993b358194cf77ccaee1c69aad7dc74fb
```
ripatel-fd, line 1: Can you revert out the submodule change?

ripatel-fd: (Or is this part of the fix?)

jherrera-jump: its a redundant bounds check on the agave side, so can be reverted, but is there a reason to not include it?

ripatel-fd: Ok makes sense to me, nevermind then. I thought you might have inadvertently changed the submodule.

PR #5544, opened 2025-07-03
runtime: fix deadlock in fd_update_epoch_stakes

two-heart:


src/flamenco/runtime/fd_runtime.c
```diff
@@ -1713,14 +1713,11 @@ fd_update_epoch_stakes( fd_exec_slot_ctx_t * slot_ctx ) {
                    fd_vote_accounts_pair_global_t_map_footprint( 50000UL ) +
                    4000 * 50000UL;
 
-  fd_vote_accounts_global_t const * next_epoch_stakes = fd_bank_next_epoch_stakes_locking_query( slot_ctx->bank );
-
-  fd_vote_accounts_global_t * epoch_stakes = fd_bank_epoch_stakes_locking_modify( slot_ctx->bank );
+  fd_vote_accounts_global_t       * epoch_stakes      = fd_bank_epoch_stakes_locking_modify( slot_ctx->bank );
+  fd_vote_accounts_global_t const * next_epoch_stakes = fd_bank_next_epoch_stakes_unsafe_query( slot_ctx->bank );
```
ripatel-fd, line 1717: Why are you not taking a read lock on the "next_epoch_stakes" field? This seems wrong?

ripatel-fd:
@two-heart I would love to understand more about this bug though, thanks for reporting. How can we reproduce the deadlock?

Issue #5098, opened 2025-05-12
Allow overriding config options via command-line


ripatel-fd:
```
fddev --hugetlbfs.max_page_size huge ...
```

beesyst:
Yo, I’ve sent in two PRs for this issue and both just got closed with zero real feedback (honestly, it felt like talking to a bot). If there’s something wrong, just say it straight up - I’d be glad to fix it. I’m doing this to help out and spending my own time. A little respect would be cool. Thx.

ripatel-fd:
@beesyst We don't accept community contributions for Firedancer as a rule of thumb. I'm sorry about your experience so far, we should have made this more clear. Unfortunately, it typically takes us less time to implement a small feature ourselves than to work with the community to review it for various reasons (an admittedly very bespoke code style and OOP pattern that is not well documented). Because we don't accept community contributions, we haven't been writing proper issue descriptions (everyone in the team sort of knows what to do...). In case you want to contribute anyways, please reach out to me and we'll find a way. Thanks again.

beesyst:
@ripatel-fd Got it, totally my bad for not asking about the rules up front. Hung around for almost a week here - wasn’t super productive and yeah, there’s a bit of a sour aftertaste, but honestly, it was a valuable experience for me. Thanks for that. Wishing you all the best, time for me to move on!

PR #5526, opened 2025-07-02
fix: replace funk reference in syscall handler with sysvar cache look…

beesyst:
### Summary
Refactored the `get_sysvar` syscall handler to use the sysvar cache instead of direct funk access, ensuring SVM remains pure and decoupled from storage/database APIs.

### Scope
- Affected files: `fd_vm_syscall_runtime.c`, `fd_bank_abi.c`
- All other funk/database references untouched

### Reason
Fixes a rare layering violation (see #5030), keeping SVM logic isolated from db/storage code.

### Testing
- Local build and all tests pass
- Verified all `get_sysvar` calls use the cache
- Ran clang-format / linter

PR #5546, opened 2025-07-03
Feature/cmdline overrides

beesyst:
**Summary**  
Added support for overriding config options via command-line flags, as discussed in #5098. Now options like `--hugetlbfs.max_page_size huge` work as expected.

**Scope**  
- Affected file: `src/app/shared/boot/fd_boot.c`  

**Reason**  
Implements feature requested in issue to allow easier configuration overrides from the CLI.

**Testing**  
Full validator run was not possible due to resource limits (my dev machine couldn’t handle the full launch),  
**BUT:**  
- Wrote a minimal local test to call `apply_cmdline_overrides()` directly with sample argv/argc.
- Output confirmed override logic works: `hugetlbfs.max_page_size = huge`
- Manual code review also confirmed behavior.

Screenshot_3

### DIFF
--- a/src/app/shared/boot/fd_boot.c
+++ b/src/app/shared/boot/fd_boot.c
@@ -10,6 +10,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <sys/mman.h>
+#include <stdio.h>
 
 extern action_t * ACTIONS[];
 extern fd_topo_run_tile_t * TILES[];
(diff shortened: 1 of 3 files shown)


mmcgee-jump:
Not sure why there are 3 unrelated commits in this PR. But please follow FD style when making changes, and resubmit.

beesyst:
> Not sure why there are 3 unrelated commits in this PR. But please follow FD style when making changes, and resubmit.

Yep, for some reason, one commit from the previous issue got into #5526. And two extra files along with it. I wanted to delete them, but you already closed it. Created a new one.

ripatel-fd:
Hi @beesyst, thanks again for your contribution. I suggest we first chat about which issues can be fixed at the moment, which ones we have bandwidth to review, and how to approach them, so we don't waste your time. We haven't accepted external contributions except for typo fixes so far (mostly because not many people submit changes). So, we don't have great issue descriptions. For this PR for example, I was hoping for a more general approach that hooks into the TOML config loader, instead of hardcoding individual config entries. We are also a bit cautious with AI generated code, since it is quick to generate, but it takes a long time to review for errors.

PR #5550, opened 2025-07-04
fix: add timeout to set-identity to avoid infinite loop

crypto-volk:
During an identity key switch, the process got stuck in a 100% CPU spin loop while waiting for a tile to complete its keyswitch. To prevent this, replaced FD_SPIN_PAUSE() with a FD_PAUSE_WITH_TIMEOUT() macro that triggers an error after a ~3s timeout. This avoids indefinite hangs if a tile fails to respond. Timeout is now applied to all blocking states in poll_keyswitch().

### DIFF
--- a/src/app/shared/commands/set_identity.c
+++ b/src/app/shared/commands/set_identity.c
@@ -139,6 +139,14 @@ find_keyswitch( fd_topo_t const * topo,
   return keyswitch;
 }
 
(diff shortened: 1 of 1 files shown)


mmcgee-jump:
It's not really valid to time these out, better to figure out why the tile is not serving the request correctly. But also please use an actual timer, e.g. `fd_log_wallclock` if you're going to do this. I'll close for now but feel free to reopen.

PR #5554, opened 2025-07-06
Standardize Dcache Bounds

mulupuruvikas:
Added dcache bounds checking to src/tango/dcache:fd_dcache_bounds_check. Added this bounds check to fd_stem STEM_CALLBACK_DURING_FRAG macro, and since that is the touchpoint for all demux and dispatching of tiles, I removed the bounds check from all other during_frag occurrences. Unit tests included.

### DIFF
--- a/src/disco/archiver/fd_archiver_feeder.c
+++ b/src/disco/archiver/fd_archiver_feeder.c
@@ -125,9 +125,6 @@ during_frag( fd_archiver_feeder_tile_ctx_t * ctx,
              ulong                           sz,
              ulong                           ctl FD_PARAM_UNUSED ) {
   /* TODO: filter by signature in before_credit */
-  if( FD_UNLIKELY( chunk<ctx->in[ in_idx ].chunk0 || chunk>ctx->in[ in_idx ].wmark ) ) {
-    FD_LOG_ERR(( "chunk %lu %lu corrupt, not in range [%lu,%lu]", chunk, sz, ctx->in[ in_idx ].chunk0, ctx->in[ in_idx ].wmark ));
-  }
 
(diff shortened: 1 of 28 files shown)


mulupuruvikas:
@ripatel-fd Sorry for the late revision, I was re-reading the Solana whitepaper and going through Firedancer architecture/design principles to understand the project more thoroughly.

I made the following changes:
-removed bounds check from `fd_net_rx_translate_frag` so that all the bounds checks are done in the `STEM_CALLBACK_DURING_FRAG`. The bounds checks should happen only once on every frag since they are all going through the stem dispatcher before they are accessed during runtime? (correct me if I'm wrong here)
-added a mtu, burst, and compact value to `fd_stem_tile_in`  struct. The burst is nonzero if the fragment is allocated using the burst strategy. The compact is 1 if the frag is allocated using the compact strategy and 0 if frag is allocated using burst strategy. If the link MTU was originally 0, I know that the frag is allocated using net rx strategy (either XDP program in NIC ->user space memory, or gossiping/shred), and I set the MTU in `fd_stem_tile_in` to FD_NET_MTU
-updated these members in `STEM_(run1)`

Please let me know if there are any gaps in my thought process above. I'd really like to have a solid understanding of this architecture so I can improve my efficiency in contributions in the future.

Also, I haven't run the full build/test just yet, but I will before completing the PR. I just want to make sure that my thoughts are on the right track so far.

ripatel-fd:
@mulupuruvikas I'm closing this now since upstream has diverged quite far from this PR. I don't think we're able to standardize dcache bounds checking at the moment. Thank you for the contribution though, this is an interesting approach.

PR #5555, opened 2025-07-06
Add disclaimer about external contributions

ripatel-fd:


ripatel-fd:
Need to think about wording ... closing

PR #5562, opened 2025-07-07
zk-sdk: add c_max_proof to the sigma proof transcript

0x0ece:
See https://github.com/anza-xyz/agave/pull/6523/files#diff-2d6b8b55bf791cc2df6dc74b3ec5d258889993107436265228aeebd8d75ca229


riptl (commented):
Approving

PR #5572, opened 2025-07-07
feat(store): merkle-chaining and equivocation aware store for dispatcher

lidatong:


src/disco/shred/fd_shred_tile.c
```diff
@@ -898,9 +903,37 @@ after_frag( fd_shred_ctx_t *    ctx,
...
+
+      fd_hash_t merkle_root;
+      memcpy( merkle_root.uc, out_merkle_root.hash, FD_SHRED_MERKLE_ROOT_SZ );
+      fd_hash_t chained_merkle_root;
+      memcpy( chained_merkle_root.uc, (uchar *)last + fd_shred_chain_off( last->variant ), FD_SHRED_MERKLE_ROOT_SZ );
+      ulong off = 0;
+      for( ulong i=0UL; i<set->data_shred_cnt; i++ ) {
+        fd_shred_t const * data_shred = (fd_shred_t const *)fd_type_pun_const( set->data_shreds[ i ] );
+        ulong sz = fd_shred_payload_sz( data_shred );
+        memcpy( ctx->store_buffer + off, data_shred, sz );
```
ptaffet-jump: You need the payload offset! This copies the shred header, no?

src/disco/shred/fd_shred_tile.c
```diff
@@ -898,9 +903,37 @@ after_frag( fd_shred_ctx_t *    ctx,
...
+        off += sz;
+      }
+      fd_rwlock_write( &ctx->store->lock );
```
ptaffet-jump: This can never be contended, right? A comment to that effect would be good if true

src/discof/store/fd_store.c
```diff
@@ -0,0 +1,263 @@
...
+  if( FD_UNLIKELY( !footprint ) ) {
+    FD_LOG_WARNING(( "bad fec_max (%lu)", fec_max ));
+    return NULL;
+  }
+
+  fd_wksp_t * wksp = fd_wksp_containing( shmem );
+  if( FD_UNLIKELY( !wksp ) ) {
+    FD_LOG_WARNING(( "shmem must be part of a workspace" ));
+    return NULL;
+  }
```
ptaffet-jump, line 27: I think this is a bit gross. I think using offsets from `shmem` is better. Then you don't care where the memory came from, which makes testing easier

src/discof/store/fd_store.h
```diff
@@ -0,0 +1,324 @@
...
+   EQUIVOCATION
+
+   There is a protocol violation called equivocation (also known as
+   "duplicates") that results in two or more blocks for the same slot.
+   For example, there is a block for slot 100 and another block for slot
+   100'.  The actual conflict occurs at the shred level (eg. a leader
+   produces two conflicting shreds, for example two conflicting payloads
+   for slot 100 idx 1), but the result will be two different FEC sets
+   (recall every shred in a FEC set has the same signature that would
+   change if the shred's payload changed).
```
ptaffet-jump:
```suggestion
   change if any of the shreds' payload changed).
```

src/discof/store/fd_store.h
```diff
@@ -0,0 +1,324 @@
...
+  /* Metadata */
+
+  ulong slot;        /* slot of the FEC set */
+  ulong parent_off;  /* offset to parent slot, [0,USHORT_MAX-1] */
+  ulong fec_set_idx; /* fec_set_idx of the FEC set, [0,32768-1] */
+
+  /* Data */
+
+  uchar data[FD_STORE_DATA_MAX]; /* FEC set payload = coalesced data shreds (byte array) */
+  ulong data_sz;                 /* FIXME fixed-32. sz of the FEC set payload, guaranteed < FD_STORE_DATA_MAX */
```
ptaffet-jump: I think it's better for cache performance to flip the order of these members. Otherwise, in the common case that data is much smaller than the max, and there's a bunch of stuff at the end of it you never read, you have another cache miss to load data_sz.

src/discof/store/fd_store.h
```diff
@@ -0,0 +1,324 @@
...
+void *
+fd_store_delete( void * store );
+
+/* fd_store_init initializes a store.  Assumes store is a valid local
+   join and no one else is joined.  root is the initial root store will
+   use.  This is the snapshot slot if booting from a snapshot, 0 if the
+   genesis slot.
+
+   In general, this should be called by the same process that formatted
+   store's memory, ie. the caller of fd_store_new. */
```
ptaffet-jump: This feels like it belongs in the `new` function, but maybe it can't?

src/discof/store/fd_store.h
```diff
@@ -0,0 +1,324 @@
...
+
+/* fd_store_insert inserts a new FEC set keyed by merkle.  Returns the
+   newly inserted fd_store_fec_t.  Copies data and data_sz into its
+   corresponding fields and copies at most FD_STORE_DATA_MAX bytes from
+   data (if handholding is enabled, it will abort the caller with a
+   descriptive error message if data_sz is too large).
+
+   Assumes the store has been initialized and has space for another
+   element, and that the parent element (keyed by chained_merkle_root)
+   has already been inserted.  Does additional checks when handholding
```
ptaffet-jump: Wait, doesn't shred call this without meeting this requirement?

ptaffet-jump (commented):
I didn't finish reviewing, but a few things to think about.  Also, how is repair going to be served if we're just storing the payload?

src/disco/shred/fd_shred_tile.c
lidatong: good catch

src/disco/shred/fd_shred_tile.c
lidatong: it can be contended by Replay tile taking out a write to publish or Replay tile reading out of Store

sorry, the wiring code with the tiles is incomplete. i just realized i should have separated into a different PR and made this just the store API

src/discof/store/fd_store.c
lidatong: hm thats a good point. i was thinking the store would always come with a wksp but that would be more flexible

src/discof/store/fd_store.h
lidatong: interesting, thanks!

src/discof/store/fd_store.h
lidatong: yeah, unfortunately `new` is being called in the wksp object machinery and at that point there is no snapshot slot to load yet

src/discof/store/fd_store.h
lidatong: yeah i realized the same while stubbing out the store insert in shred_tile -- i was thinking either repair tile would have to insert (after chainer) or reconstruct the ancestry

src/disco/shred/fd_shred_tile.c
lidatong: added some documentation for the concurrency model we brainstormed

src/discof/store/fd_store.h
lidatong: went with second option -- added a `fd_store_link`

src/discof/store/fd_store.c
```diff
@@ -0,0 +1,198 @@
...
+  ulong            null = fd_store_pool_idx_null( pool );
+  fd_store_fec_t * fec  = fd_store_pool_ele_acquire( pool );
+  fec->key              = *merkle_root;
+  fec->next             = null;
+  fec->parent           = null;
+  fec->child            = null;
+  fec->sibling          = null;
+  fec->data_sz          = data_sz;
+  memcpy( fec->data, data, data_sz );
+  fd_store_map_ele_insert( fd_store_map( store ), fec, fd_store_pool( store ) );
```
ptaffet-jump, line 127:
```suggestion
  fd_store_map_ele_insert( fd_store_map( store ), fec, pool );
```
right?

ptaffet-jump (approved):

src/disco/shred/fd_shred_tile.c
lidatong: ripping out all the tiles integration (which is incomplete) into a separate PR because it's becoming a big change

src/discof/store/fd_store.c
lidatong: will make this change in a separate PR

src/discof/store/fd_store.h
lidatong: got rid of `init` and made `insert` set it as root. the linking of parent->child happens later anyways

PR #5583, opened 2025-07-08
remove legacy wait duration ticks

cavemanloverboy:
`wait_duration`/`wait_duration_ticks` and its use in this particular branch

```c
/* Do I have enough transactions and/or have I waited enough time? */
  if( FD_UNLIKELY( (ulong)(now-ctx->last_successful_insert) <
  ctx->wait_duration_ticks[ fd_ulong_min( fd_pack_avail_txn_cnt( ctx->pack ), MAX_TXN_PER_MICROBLOCK ) ] ) ) {
  update_metric_state( ctx, now, FD_PACK_METRIC_STATE_TRANSACTIONS, 0 );
  return;
}
```
in the pack tile's `after_credit` routine is a sort of legacy pacing strategy that is both ineffective and interferes with existing strategies. Consider the following for each of the following strategies:

1. `FD_PACK_STRATEGY_PERF`: In this strategy, we are trying to pack transaction as fast as possible and should not wait for transactions in any case
2. `FD_PACK_STRATEGY_BALANCED`: in this strategy, we already have `fd_pacer` that is limiting scheduling and do not want to wait more time if we are below are target cus when we have a few transactions buffered.
3. `FD_PACK_STRATEGY_BUNDLE`: in this strategy, we should not wait around votes, we always schedule bundles in FIFO, and have extremely limited time to pack nonvote nonbundle transactions and do not want to wait in the case we have few transactions in the buffer.

### DIFF
--- a/src/disco/pack/fd_pack_tile.c
+++ b/src/disco/pack/fd_pack_tile.c
@@ -60,22 +60,6 @@ const float VOTE_FRACTION = 0.75f; /* TODO: Is this the right value? */
 #define EFFECTIVE_TXN_PER_MICROBLOCK MAX_TXN_PER_MICROBLOCK
 #endif
 
-/* There's overhead associated with each microblock the bank tile tries
-   to execute it, so the optimal strategy is not to produce a microblock
-   with a single transaction as soon as we receive it.  Basically, if we
-   have less than 31 transactions, we want to wait a little to see if we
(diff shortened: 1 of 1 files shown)


ptaffet-jump:
Sorry for taking so long to get to this.  #7286 does basically the same thing, so I'm going to close this.

PR #5584, opened 2025-07-08
shred_dest: variable contact info count

akhinvasara-jumptrading:
UPDATE:
Letting this sit as a draft because upcoming changes to gossip mean the send_tile doesn't need to use stake_ci. If there are no consumers who require multiple contact info, it's not worth modifying stake_ci. 

This PR extends stake_ci (and therefore shred_dest) to track an arbitrary number of ip<>port pairs. It lets us use stake_ci in places where we want to track multiple ip/ports per validator (such as the send_tile) without rewriting all the messy logic. 

This is really a band-aid for timeline purposes. We should eventually refactor this to make stake_ci an underlying data source that just joins arbitrary data (including contact info) with stake info by pubkey. Shred_dest should depend on this stake_ci, rather than the other way around. Leader schedule computation should maybe also be factored out?

### DIFF
--- a/src/disco/shred/fd_stake_ci.h
+++ b/src/disco/shred/fd_stake_ci.h
@@ -15,16 +15,6 @@
 #include "fd_shred_dest.h"
 #include "../../flamenco/leaders/fd_leaders.h"
 
-#define MAX_SHRED_DESTS             MAX_STAKED_LEADERS
-/* staked+unstaked <= MAX_SHRED_DESTS implies
-   MAX_SHRED_DEST_FOOTPRINT>=fd_shred_dest_footprint( staked, unstaked )
-   This is asserted in the tests.  The size of fd_shred_dest_t, varies
-   based on FD_SHA256_BATCH_FOOTPRINT, which depends on the compiler
-   settings. */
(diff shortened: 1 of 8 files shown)


src/disco/shred/fd_stake_ci.h
```diff
@@ -16,15 +16,6 @@
 #include "../../flamenco/leaders/fd_leaders.h"
 
 #define MAX_SHRED_DESTS             MAX_STAKED_LEADERS
-/* staked+unstaked <= MAX_SHRED_DESTS implies
-   MAX_SHRED_DEST_FOOTPRINT>=fd_shred_dest_footprint( staked, unstaked )
-   This is asserted in the tests.  The size of fd_shred_dest_t, varies
-   based on FD_SHA256_BATCH_FOOTPRINT, which depends on the compiler
-   settings. */
-#define MAX_SHRED_DEST_FOOTPRINT (8386688UL + sizeof(fd_shred_dest_t))
```
akhinvasara-jumptrading, line 24: I left this 8386688UL assertion in the test with user_ci_cnt == 1. Is it worth recomputing the max_shred_dest_footprint with the max user_ci_cnt? MAX_SHRED_DEST_FOOTPRINT isn't used anywhere anymore.

PR #5587, opened 2025-07-08
fix: filter out cloned routes when loading routing table

jkbpvsc:
- Filter out routes with `RTM_F_CLONED` flag
- Immediately filter out routes with non target `table_id` before reading `RTA_TABLE` if `rtm_table` is not `RT_TABLE_UNSPEC`

This fixes the `netlnk:0 src/waltz/ip/fd_fib4_netlink.c(227): Routing table is too small! `ip route show table 254` returned 2815 entries, which exceeds the configured maximum of 128` bug.


ripatel-fd:
Thank you for the contribution 🎉  This bug was driving me crazy.

PR #5589, opened 2025-07-08
Large route table with hmap

hewang-jump:


src/app/fdctl/config/default.toml
```diff
@@ -1016,6 +1016,11 @@ dynamic_port_range = "8900-9000"
         # number of routes per table does not exceed 16.
         max_routes = 128
 
+        # The maximum number of routes per route hmap. Since these routes use
+        # /32 netmasks, destination IPs can serve directly as lookup keys for
+        # the next hop.
+        max_routes_peers = 128
```
ripatel-fd: We need to reword this to be more operator-friendly.
We can omit the fact that it uses a hashmap, since that's an implementation detail.

The default should also be really large. Here's a suggestion:

```suggestion
        # This setting configures the maximum number of IPv4 routes with a /32 netmask
        # per route table.  This setting is relevant  for Double Zero accelerated networking, 
        # which uses one /32 route per peer on the network.  The number of /32 routes
        # should not exceed 100,000.
        max_peer_routes = 8192
```

For `max_routes`, please amend the documentation to say, "excluding /32 routes, see below"

src/app/firedancer/config/default.toml
```diff
@@ -742,6 +742,11 @@ user = ""
         # number of routes per table does not exceed 16.
         max_routes = 128
 
+        # The maximum number of routes per route hmap. Since these routes use
+        # /32 netmasks, destination IPs can serve directly as lookup keys for
+        # the next hop.
+        max_routes_peers = 128
```
ripatel-fd: see above comment

src/app/shared/fd_obj_callbacks.c
```diff
@@ -230,7 +230,7 @@ fd_topo_obj_callbacks_t fd_obj_cb_neigh4_hmap = {
 static ulong
 fib4_footprint( fd_topo_t const *     topo,
                 fd_topo_obj_t const * obj ) {
-  return fd_fib4_footprint( VAL("route_max") );
+  return fd_fib4_footprint( VAL("route_max"), VAL("route_max_peer") );
```
ripatel-fd:
```suggestion
  return fd_fib4_footprint( VAL("route_max"), VAL("route_peer_max") );
```

we usually put `max` last in naming

src/disco/netlink/fd_netlink_tile.h
```diff
@@ -35,6 +35,7 @@ void
 fd_netlink_topo_create( fd_topo_tile_t * netlink_tile,
                         fd_topo_t *      topo,
                         ulong            netlnk_max_routes,
+                        ulong            netlnk_max_routes_peers,
```
ripatel-fd:
```suggestion
                        ulong            netlnk_max_peer_routes,
```

src/waltz/ip/fd_fib4.c
```diff
@@ -40,18 +45,30 @@ fd_fib4_new( void * mem,
   }
 
   FD_SCRATCH_ALLOC_INIT( l, mem );
-  fd_fib4_t *     fib4 = FD_SCRATCH_ALLOC_APPEND( l, alignof(fd_fib4_t),     sizeof(fd_fib4_t)               );
-  fd_fib4_key_t * keys = FD_SCRATCH_ALLOC_APPEND( l, alignof(fd_fib4_key_t), route_max*sizeof(fd_fib4_key_t) );
-  fd_fib4_hop_t * vals = FD_SCRATCH_ALLOC_APPEND( l, alignof(fd_fib4_hop_t), route_max*sizeof(fd_fib4_hop_t) );
+  fd_fib4_t *     fib4     = FD_SCRATCH_ALLOC_APPEND( l, alignof(fd_fib4_t),     sizeof(fd_fib4_t)               );
+  fd_fib4_key_t * keys     = FD_SCRATCH_ALLOC_APPEND( l, alignof(fd_fib4_key_t), route_max*sizeof(fd_fib4_key_t) );
+  fd_fib4_hop_t * vals     = FD_SCRATCH_ALLOC_APPEND( l, alignof(fd_fib4_hop_t), route_max*sizeof(fd_fib4_hop_t) );
+  void * fib4_hmap_mem     = FD_SCRATCH_ALLOC_APPEND( l, fd_fib4_hmap_align(), fd_fib4_hmap_footprint( route_max_peer, FIB4_HMAP_LOCK_CNT, FIB4_HMAP_PROBE_MAX ) );
```
ripatel-fd: if route_max_peer==0, does this crash?

hewang-jump: the hmap_new will fail later on. I will add a check for that.

src/waltz/ip/fd_fib4.h
```diff
@@ -107,6 +130,18 @@ fd_fib4_append( fd_fib4_t * fib,
...
+   hmap. Assume the netmask for the ip4_dst entry is 32, and the ip4_dst
+   entry is not already present in fib->hmap, and ip4_dst is not zero.
+   The insertion to fib->hmap is blocking. Return FD_MAP_SUCCESS on success,
+   FD_MAP_ERR_FULL if the hmap is full.
+*/
+
+int
+fd_fib4_hmap_insert( fd_fib4_t * fib,
+                     uint ip4_dst,
+                     fd_fib4_hop_t hop );
```
ripatel-fd: This seems like a private API. Can you move it to fd_fib4_private.h?

src/waltz/ip/fd_fib4.h
```diff
@@ -137,16 +172,21 @@ fd_fib4_hop_or( fd_fib4_hop_t const * left,
 FD_FN_PURE ulong
 fd_fib4_max( fd_fib4_t const * fib );
 
-/* fd_fib4_cnt returns the number of routes in the table. */
+/* fd_fib4_max returns the max number of routes in the hmap. */
+
+FD_FN_PURE ulong
+fd_fib4_hmap_max( fd_fib4_t const * fib );
```
ripatel-fd: In the public API, I suggest following the documentation and naming as in the user-facing config file.

So, I would call this `fd_fib4_peer_max` and document it as "max number of /32 routes (backed by a hashmap)"

src/waltz/ip/fd_fib4.h
```diff
@@ -137,16 +172,21 @@ fd_fib4_hop_or( fd_fib4_hop_t const * left,
...
 FD_FN_PURE ulong
 fd_fib4_cnt( fd_fib4_t const * fib );
 
-/* fd_fib4_free_cnt returns the number of fd_fib4_append calls that are
+/* fd_fib4_table_free_cnt returns the number of fd_fib4_append calls that are
    guaranteed to succeed. */
 
 ulong
-fd_fib4_free_cnt( fd_fib4_t const * fib );
+fd_fib4_table_free_cnt( fd_fib4_t const * fib );
```
ripatel-fd: I suggest the following APIs (first one returns total free count, second one returns hashmap free count)

```
fd_fib4_free_cnt( fd_fib4_t const * fib );

fd_fib4_peer_free_cnt( fd_fib4_t const * fib );
```

src/waltz/ip/fd_fib4.h
hewang-jump: the fd_fib4_netlink.c uses this API in fd_fib4_netlink_translate. That file only includes the fd_fib4.h tho.

src/waltz/ip/fd_fib4.h
hewang-jump: the fd_fib4_table_free_cnt was used in fd_fib4_netlink.c before fd_fib4_append. Should I do the following:
 'fd_fib4_free_cnt( fd_fib4_t const * fib ); // for total free cnt '
`fd_fib4_peer_free_cnt( ... fib) // for hmap`
`fd_fib4_table_free_cnt( ... fib) // for table`

src/waltz/ip/fd_fib4.h
```diff
@@ -57,6 +57,25 @@ struct __attribute__((aligned(16))) fd_fib4_hop {
...
+
+struct __attribute__((aligned(16))) fd_fib4_hmap_entry {
+  uint dst_addr; /* prefix bits, little endian (low bits outside of mask are undefined) */
+  fd_fib4_hop_t next_hop;
+};
+
+typedef struct fd_fib4_hmap_entry fd_fib4_hmap_entry_t;
+
+#define MAP_IMPL_STYLE 0
+#include "../../util/tmpl/fd_map_slot_para.c"
```
ripatel-fd: @hewang-jump Could we move this hashmap definition to fd_fib4_private.h?

src/waltz/ip/fd_fib4.h
```diff
@@ -137,16 +158,15 @@ fd_fib4_hop_or( fd_fib4_hop_t const * left,
 FD_FN_PURE ulong
 fd_fib4_max( fd_fib4_t const * fib );
 
-/* fd_fib4_cnt returns the number of routes in the table. */
+/* fd_fib4_max returns the max number of 32/ routes (backed by a hashmap). */
 
 FD_FN_PURE ulong
-fd_fib4_cnt( fd_fib4_t const * fib );
+fd_fib4_peer_max( fd_fib4_t const * fib );
```
ripatel-fd:
```suggestion
/* fd_fib4_peer_max returns the max number of /32 routes (backed by a hashmap). */

FD_FN_PURE ulong
fd_fib4_peer_max( fd_fib4_t const * fib );
```

src/waltz/ip/fd_fib4.h
```diff
@@ -137,16 +158,15 @@ fd_fib4_hop_or( fd_fib4_hop_t const * left,
...
 
-/* fd_fib4_free_cnt returns the number of fd_fib4_append calls that are
-   guaranteed to succeed. */
+/* fd_fib4_cnt returns the number of routes in the table and the hmap. */
```
ripatel-fd:
```suggestion
/* fd_fib4_cnt returns the total number of routes stored in the fib4.
   This also includes /32 routes. */
```

src/app/fdctl/config/default.toml
```diff
@@ -1013,9 +1013,23 @@ dynamic_port_range = "8900-9000"
...
+        # accelerated networking, which uses one /32 route per peer on the
+        # network. The number of /32 routes should not exceed 100,000.
+        max_peer_routes = 8192
+
+        # This configures the number of locks for concurrent access to the IPv4
+        # /32 route table (see above). Each lock protects roughly
+        # max_peer_routes/lock_cnt route entries. Higher values improve
+        # parallelism for concurrent operations but increase memory overhead.
+        # Adjust based on expected concurrency level (typically 16-1024).
+        lock_cnt_peer_routes = 32
```
ripatel-fd: This is pretty technical, so I wouldn't expose this to our users.
Could you remove this config option and instead choose a default based on the route table size in the obj_callback code?

src/waltz/ip/fd_fib4.c
```diff
@@ -76,35 +107,82 @@ fd_fib4_delete( void * mem ) {
...
-                int         prefix,
-                uint        prio ) {
+int
+fd_fib4_insert( fd_fib4_t *     fib,
+                uint            ip4_dst,
+                int             prefix,
+                uint            prio,
+                fd_fib4_hop_t * hop ) {
+
+  if( ip4_dst!=0 && prefix==32 && hop[0].ip4_src!=0 ) {
```
ripatel-fd: I think we can remove this filter: `hop[0].ip4_src!=0`

src/waltz/ip/fd_fib4.c
```diff
@@ -76,35 +107,82 @@ fd_fib4_delete( void * mem ) {
...
+    if( fd_fib4_hmap_insert( fib, ip4_dst, hop )==FD_MAP_SUCCESS ) return 1;
```
ripatel-fd, line 185: If this is not success, we should log a warning and drop the route, instead of falling through and adding the route to the main fib4 array. To maintain the invariant that the array only contains /0-31 routes and hmap only contains /32 routes

src/waltz/ip/fd_fib4.c
```diff
@@ -137,6 +218,30 @@ fd_fib4_lookup( fd_fib4_t const * fib,
...
+    fd_fib4_hmap_entry_t sentinel[1];
+    int find_err  = fd_fib4_hmap_query_try( hmap, &key, sentinel, query, 0 );
+    if( find_err==FD_MAP_SUCCESS ) {
+      fd_fib4_hmap_entry_t const * ele = fd_fib4_hmap_query_ele_const( query );
+      fd_fib4_hop_t next_hop           = ele->next_hop;                    // speculatively save the next hop
+      find_err                         = fd_fib4_hmap_query_test( query ); // test again
+      if( FD_UNLIKELY( find_err==FD_MAP_ERR_AGAIN ) ) {
+        return &fd_fib4_hop_blackhole;
+      }
+      fd_memcpy( out, &next_hop, sizeof( fd_fib4_hop_t ) );
```
ripatel-fd, line 247: This falls through if find_err returns something other than ERR_EGAIN (might not be possible today, but in the future if the hashmap implementation changes)  
I would just return the blackhole route whenever any error occurs

src/waltz/ip/fd_fib4.c
```diff
@@ -137,6 +218,30 @@ fd_fib4_lookup( fd_fib4_t const * fib,
...
+      return out;
+    } else if( FD_UNLIKELY( find_err==FD_MAP_ERR_AGAIN ) ) {
+      return &fd_fib4_hop_blackhole;
+    }
```
ripatel-fd:
```suggestion
    } else {
      return &fd_fib4_hop_blackhole;
    }
```

src/waltz/ip/fd_fib4.c
```diff
@@ -76,35 +107,82 @@ fd_fib4_delete( void * mem ) {
...
 fd_fib4_cnt( fd_fib4_t const * fib ) {
-  return fib->cnt;
+  return fib->cnt+fib->hmap_cnt;
 }
 
-ulong
-fd_fib4_free_cnt( fd_fib4_t const * fib ) {
-  if( FD_UNLIKELY( fib->cnt > fib->max ) ) FD_LOG_ERR(( "invalid fib4 state: cnt>max" ));
-  return fib->max - fib->cnt;
+int
```
ripatel-fd:
```suggestion
static int
```
Internal function, so it should be statically linked

src/waltz/ip/test_fib4.c
```diff
@@ -65,29 +425,39 @@ main( int     argc,
...
-    "broadcast 192.0.2.160 dev 6 scope link src 192.0.2.165\n"
-    "local 192.0.2.165 dev 6 scope host src 192.0.2.165\n"
-    "broadcast 192.0.2.191 dev 6 scope link src 192.0.2.165\n"
-    "broadcast 127.0.0.0 dev 1 scope link src 127.0.0.1\n"
+    "local 192.0.2.165/24 dev 6 scope host src 192.0.2.165\n"
+    "broadcast 127.0.0.0/30 dev 1 scope link src 127.0.0.1\n"
     "local 127.0.0.0/8 dev 1 scope host src 127.0.0.1\n"
-    "local 127.0.0.1 dev 1 scope host src 127.0.0.1\n"
-    "broadcast 127.0.255.255 dev 1 scope link src 127.0.0.1\n" );
+    "broadcast 127.0.255.255/30 dev 1 scope link src 127.0.0.1\n" );
```
ripatel-fd: Could you modify this to also print the hmap entries? You can use the _iter API of the hashmap for this

src/waltz/ip/fd_fib4.c
```diff
@@ -280,6 +395,28 @@ fd_fib4_fprintf( fd_fib4_t const * fib,
...
+  ulong lock_cnt  = fd_fib4_hmap_get_lock_cnt( elem_max );
+  ulong ignored[ fd_fib4_hmap_lock_max( ) ];
+  FD_TEST( fd_fib4_hmap_lock_range( hmap, 0, lock_cnt, FD_MAP_FLAG_BLOCKING | FD_MAP_FLAG_RDONLY, ignored )==FD_MAP_SUCCESS );
+
+  // loop through the hmap elements
+  for( ulong i=0; i<elem_max; i++ ) {
+    if( elems[i].dst_addr!=0 ) {
+      fd_fib4_key_t key;
+      key.addr = fd_uint_bswap( elems[i].dst_addr );
+      key.mask = 31;
```
ripatel-fd, line 412: Wouldn't the mask be 32?

PR #5588, opened 2025-07-08
GRE bound check

hewang-jump:

PR #5590, opened 2025-07-09
net: move comparison into FD_UNLIKELY scope

two-heart:


ripatel-fd:
Duplicate of https://github.com/firedancer-io/firedancer/pull/5588 (Thanks for the CodeQL query though)

PR #5596, opened 2025-07-09
snapshots: clean shutdown

mmcgee-jump:


src/discof/restore/fd_snaprd_tile.c
```diff
@@ -750,9 +748,10 @@ unprivileged_init( fd_topo_t *      topo,
 #define STEM_CALLBACK_CONTEXT_TYPE  fd_snaprd_tile_t
 #define STEM_CALLBACK_CONTEXT_ALIGN alignof(fd_snaprd_tile_t)
 
-#define STEM_CALLBACK_METRICS_WRITE metrics_write
-#define STEM_CALLBACK_DURING_FRAG   during_frag
-#define STEM_CALLBACK_AFTER_CREDIT  after_credit
+#define STEM_CALLBACK_SHOULD_SHUTDOWN should_shutdown
```
ripatel-fd, line 758: Why not just make this a bit in `fd_stem_context_t`?

src/disco/topo/fd_topo.h
```diff
@@ -108,6 +108,7 @@ struct fd_topo_tile {
   char  name[ 7UL ];            /* The name of this tile.  There can be multiple of each tile name in a topology. */
   ulong kind_id;                /* The ID of this tile within its name.  If there are n tile of a particular name, they have IDs [0, N).  The pair (name, kind_id) uniquely identifies a tile, as does "id" on its own. */
   int   is_agave;               /* If the tile needs to run in the Agave (Anza) address space or not. */
+  int   allow_shutdown;         /* If the tile is allowed to shutdown gracefully.  If false, when the tile exits it will tear down the entire applicatiom. */
```
cali-jumptrading:
```suggestion
  int   allow_shutdown;         /* If the tile is allowed to shutdown gracefully.  If false, when the tile exits it will tear down the entire application. */
```

src/disco/topo/fd_topo_run.c
```diff
@@ -136,7 +136,9 @@ fd_topo_run_tile( fd_topo_t *          topo,
     tile_run->unprivileged_init( topo, tile );
 
   tile_run->run( topo, tile );
-  FD_LOG_ERR(( "tile run loop returned" ));
+  if( FD_UNLIKELY( !tile->allow_shutdown ) ) FD_LOG_ERR(( "tile %s:%lu run loop returned", tile->name, tile->kind_id ));
```
cali-jumptrading, line 139: perhaps log a NOTICE if the `allow_shutdown` is true?

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: +1

src/disco/topo/fd_topo_run.c
mmcgee-jump: The supervisor process is going to log instead

src/discof/restore/fd_snaprd_tile.c
mmcgee-jump: Eh, it's callback based right now, and stem context is for providing context to publish, not return into. If you feel strongly I'll do it but seems minor.

ripatel-fd: hmm, I do feel strongly about this ... Any "should_shutdown" callback that is more complex than reading a single variable is probably bugged. So, might as well just hardcode it

PR #5597, opened 2025-07-09
gui: don't error on conn close, properly reset http stage

jherrera-jump:


src/waltz/http/fd_http_server.c
```diff
@@ -1099,10 +1099,14 @@ fd_http_server_ws_send( fd_http_server_t * http,
      and any connections that were previously using that allotted space
      are closed.  There is a small chance that ws_conn_id is one of
      those connections, and has therefore already been closed. */
-  if( FD_LIKELY( http->pollfds[ http->max_conns+ws_conn_id ].fd==-1 ) ) return -1;
+  if( FD_LIKELY( http->pollfds[ http->max_conns+ws_conn_id ].fd==-1 ) ) {
+    http->stage_len = 0;
```
mmcgee-jump, line 1103: Why do we need to zero the staging len if the connection is closed?

jherrera-jump: If the staging area keeps whatever was there it will get pre-pended to the next broadcast message. Allen was seeing messages like this
```
{"topic":"summary","key":"ping","id":1,"value": null}{"topic":"summary","key":"live_tile_timers","value":[-1.00,0.69,-1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,1.00,0.60,0.77,0.66,0.96,1.00,0.00,0.00,1.00,0.97,0.00]}
```
which were getting rejected by the frontend due to being invalid json. I think this is the reason.

mmcgee-jump (approved):

PR #5605, opened 2025-07-09
snapshots: configure stage

cali-jumptrading:
Add a configure stage for snapshots in firedancer-dev which configures the snapshots path from the config [paths.snapshots]

### DIFF
--- a/book/guide/initializing.md
+++ b/book/guide/initializing.md
@@ -30,9 +30,9 @@ where `mode` is one of:
  - `fini` Unconfigure (reverse) the stage if it is reversible.
 
 `stage` can be one or more of `hugetlbfs`, `sysctl`, `hyperthreads`,
-`ethtool-channels`, `ethtool-gro`, `ethtool-loopback` and these stages
-are described below. You can also use the stage `all` which will
-configure everything.
+`ethtool-channels`, `ethtool-gro`, `ethtool-loopback`, and `snapshots`
+and these stages are described below. You can also use the stage `all`
+which will configure everything.
 
 Stages have different privilege requirements, which you can see by
 trying to run the stage without privileges. The `check` mode never
@@ -208,3 +208,16 @@ Firedancer. It has no dependencies on any other stage.
 
 Changing device settings with `ethtool-loopback` requires root privileges,
 and cannot be performed with capabilities.
+
+## snapshots
+When starting up, validators must load a snapshot to catch up to the
+current state of the blockchain. Snapshots are downloaded from other
+validator peers in the cluster and are stored to a snapshots directory.
+
+In init, the snapshots configure phase will create the snapshots
+directory if it does not exist. In fini, the snapshots configure phase
+will remove the snapshots directory recursively.
+
+::: tip NOTE
+
+The snapshots configure phase is only enabled in the Firedancer binary.
--- a/src/app/firedancer-dev/main.c
+++ b/src/app/firedancer-dev/main.c
@@ -68,6 +68,7 @@ configure_stage_t * STAGES[] = {
   &fd_cfg_stage_ethtool_loopback,
   &fd_cfg_stage_keys,
   &fd_cfg_stage_genesis,
+  &fd_cfg_stage_snapshots,
   NULL,
 };
 
--- a/src/app/firedancer/main.c
+++ b/src/app/firedancer/main.c
@@ -48,6 +48,7 @@ fd_topo_obj_callbacks_t * CALLBACKS[] = {
   &fd_obj_cb_bank_hash_cmp,
   NULL,
 };
+
 configure_stage_t * STAGES[] = {
   &fd_cfg_stage_hugetlbfs,
   &fd_cfg_stage_sysctl,
@@ -55,6 +56,7 @@ configure_stage_t * STAGES[] = {
   &fd_cfg_stage_ethtool_channels,
   &fd_cfg_stage_ethtool_gro,
   &fd_cfg_stage_ethtool_loopback,
+  &fd_cfg_stage_snapshots,
   NULL,
 };
 
--- a/src/app/shared/Local.mk
+++ b/src/app/shared/Local.mk
@@ -26,6 +26,7 @@ $(call add-objs,commands/configure/ethtool-loopback,fdctl_shared)
 $(call add-objs,commands/configure/hugetlbfs,fdctl_shared)
 $(call add-objs,commands/configure/hyperthreads,fdctl_shared)
 $(call add-objs,commands/configure/sysctl,fdctl_shared)
+$(call add-objs,commands/configure/snapshots,fdctl_shared)
 $(call add-objs,commands/monitor/monitor commands/monitor/helper,fdctl_shared)
 $(call add-objs,commands/run/run commands/run/run1,fdctl_shared)
 
--- a/src/app/shared/commands/configure/configure.h
+++ b/src/app/shared/commands/configure/configure.h
@@ -71,6 +71,7 @@ extern configure_stage_t fd_cfg_stage_hyperthreads;
 extern configure_stage_t fd_cfg_stage_ethtool_channels;
 extern configure_stage_t fd_cfg_stage_ethtool_gro;
 extern configure_stage_t fd_cfg_stage_ethtool_loopback;
+extern configure_stage_t fd_cfg_stage_snapshots;
 
 extern configure_stage_t * STAGES[];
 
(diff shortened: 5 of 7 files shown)


src/app/shared_dev/commands/configure/snapshots.c
```diff
@@ -0,0 +1,33 @@
...
+}
+
+configure_stage_t fd_cfg_stage_snapshots = {
+  .name            = "snapshots",
+  .always_recreate = 1,
+  .enabled         = NULL,
+  .init_perm       = NULL,
+  .fini_perm       = NULL,
+  .init            = init,
+  .fini            = NULL,
```
topointon-jump: We should probably have a fini here which removes the directory

src/app/shared_dev/commands/configure/snapshots.c
```diff
@@ -0,0 +1,33 @@
...
+  if( FD_UNLIKELY( stat( config->paths.snapshots, &st ) && errno==ENOENT ) )
+    NOT_CONFIGURED( "`%s` does not exist", config->paths.snapshots );
+
+  CHECK( check_dir( config->paths.snapshots, config->uid, config->gid, S_IFDIR | S_IRUSR | S_IWUSR | S_IXUSR ) );
+  CONFIGURE_OK();
+}
+
+configure_stage_t fd_cfg_stage_snapshots = {
+  .name            = "snapshots",
+  .always_recreate = 1,
```
ripatel-fd: Doesn't this mean mkdir_all runs even if check returns ok? Seems unnecessary, but I might be missing something.

src/app/shared_dev/commands/configure/snapshots.c
```diff
@@ -0,0 +1,33 @@
...
+  .enabled         = NULL,
+  .init_perm       = NULL,
+  .fini_perm       = NULL,
```
ripatel-fd: IMO it's cleaner to omit these NULL things

src/app/shared_dev/commands/configure/snapshots.c
ripatel-fd: Hmm good point. I think a fini subcommand for snapshot specifically would be good.
But I don't want this thing to nuke data when an operator runs `fdctl configure fini all`

cali-jumptrading: we would want to validators to keep snapshots around in case they need to quickly restart after shutting down right? So we don't want to nuke the directory here.

src/app/shared_dev/commands/configure/snapshots.c
```diff
@@ -0,0 +1,33 @@
...
+
+#include "../../../platform/fd_file_util.h"
+
+#include <errno.h>
+#include <sys/stat.h>
+
+static void
+init( config_t const * config ) {
+  if( FD_UNLIKELY( -1==fd_file_util_mkdir_all( config->paths.snapshots, config->uid, config->gid ) ) )
+    FD_LOG_ERR(( "could not create identity directory `%s` (%i-%s)", config->paths.snapshots, errno, fd_io_strerror( errno ) ));
```
mmcgee-jump:
```suggestion
    FD_LOG_ERR(( "could not create snapshots directory `%s` (%i-%s)", config->paths.snapshots, errno, fd_io_strerror( errno ) ));
```

src/app/shared_dev/commands/configure/snapshots.c
mmcgee-jump: This should be 0

src/app/shared_dev/commands/configure/snapshots.c
mmcgee-jump: Yes, please add a fini stage

src/app/firedancer-dev/main.c
```diff
@@ -54,6 +54,7 @@ extern configure_stage_t fd_cfg_stage_kill;
 extern configure_stage_t fd_cfg_stage_netns;
 extern configure_stage_t fd_cfg_stage_genesis;
 extern configure_stage_t fd_cfg_stage_keys;
+extern configure_stage_t fd_cfg_stage_snapshots;
```
mmcgee-jump: Please update the documentation in the `book/` regarding the configure stages

mmcgee-jump (changes requested):
.

src/app/firedancer-dev/main.c
```diff
@@ -66,6 +67,7 @@ configure_stage_t * STAGES[] = {
   &fd_cfg_stage_ethtool_loopback,
   &fd_cfg_stage_keys,
   &fd_cfg_stage_genesis,
+  &fd_cfg_stage_snapshots,
```
mmcgee-jump, line 71: Please add to `firedancer/main.c` as well

src/app/firedancer-dev/main.c
ripatel-fd: We need to do this in a way that keeps Frankendancer and full Firedancer apart

src/app/shared_dev/commands/configure/snapshots.c
cali-jumptrading: Added a fini stage

src/app/shared_dev/commands/configure/snapshots.c
cali-jumptrading: omitted NULLs

src/app/firedancer-dev/main.c
cali-jumptrading: added to firedancer/main.c

src/app/shared_dev/commands/configure/snapshots.c
```diff
@@ -0,0 +1,38 @@
...
+  CHECK( check_dir( config->paths.snapshots, config->uid, config->gid, S_IFDIR | S_IRUSR | S_IWUSR | S_IXUSR ) );
+  CONFIGURE_OK();
+}
+
+static void
+fini( config_t const * config,
+      int              pre_init ) {
+  (void)pre_init;
+  if( -1==fd_file_util_rmtree( config->paths.snapshots, 1 ) )
+   FD_LOG_ERR(( "rmtree `%s` failed (%i-%s)", config->paths.snapshots, errno, fd_io_strerror( errno ) ));
```
mmcgee-jump: nit :) 

```suggestion
    FD_LOG_ERR(( "rmtree `%s` failed (%i-%s)", config->paths.snapshots, errno, fd_io_strerror( errno ) ));
```

src/app/firedancer/main.c
```diff
@@ -48,13 +48,25 @@ fd_topo_obj_callbacks_t * CALLBACKS[] = {
...
+extern configure_stage_t fd_cfg_stage_snapshots;
+
 configure_stage_t * STAGES[] = {
   &fd_cfg_stage_hugetlbfs,
   &fd_cfg_stage_sysctl,
   &fd_cfg_stage_hyperthreads,
   &fd_cfg_stage_ethtool_channels,
   &fd_cfg_stage_ethtool_gro,
   &fd_cfg_stage_ethtool_loopback,
+  &fd_cfg_stage_kill,
```
mmcgee-jump: These shouldn't be here, the development stages do not get included or run in the production validator binary.

mmcgee-jump (changes requested):
.

src/app/firedancer/main.c
cali-jumptrading: oh... what constitutes a development stage? Confused which stages are development and which are for production.

book/guide/initializing.md
```diff
@@ -208,3 +208,22 @@ Firedancer. It has no dependencies on any other stage.
 
 Changing device settings with `ethtool-loopback` requires root privileges,
 and cannot be performed with capabilities.
+
+## genesis
+All validators must agree on the initial state of the blockchain before
```
cali-jumptrading: Please suggest improvements to the documentation! I don't completely understand the genesis configure phase and might be glossing over some things.

src/app/firedancer/main.c
mmcgee-jump: The ones in the `dev` or `shared_dev` directory are development only. For an example, `kill` is responsible for terminating any already running `fdctl` instances, but we definitely do not want to do that from the production command (it would terminate running validators).

mmcgee-jump: I don't even think this will compile correctly, as those stages are not linked to the production binary

mmcgee-jump (dismissed):

PR #5608, opened 2025-07-10
snapshots: simplify snaprd tile

mmcgee-jump:


src/discof/restore/fd_snaprd_tile.c
```diff
@@ -598,169 +308,278 @@ after_credit( fd_snaprd_tile_t *  ctx,
...
+    *incremental_slot = slot;
+  } else {
+    *incremental_slot = ULONG_MAX;
+  }
+
+  ptr = next + 1;
+  next = strchr( ptr, '.' );
+  if( FD_UNLIKELY( !next ) ) return -1;
+
+  if( !strncmp( next, ".tar.zst", 8UL ) ) return -1;
```
two-heart, line 423: inverted?

two-heart (changes requested):

src/discof/restore/utils/fd_ssping.c
```diff
@@ -0,0 +1,503 @@
...
+  peer->fd.idx = idx;
+
+  ssping->fds_len--;
+}
+
+void
+fd_ssping_remove( fd_ssping_t * ssping,
+                  fd_ip4_port_t addr ) {
+  fd_ssping_peer_t * peer = peer_map_ele_query( ssping->map, &addr, NULL, ssping->pool );
+  FD_TEST( peer );
```
two-heart, line 254: Does not match the header comment:
> If the peer is not tracked, this is a no-op.

two-heart (changes requested):

cali-jumptrading (commented):
I think this would be better as incremental, separate PRs.

PR #5613, opened 2025-07-10
snapshots: correct stem burst value for snaprd

cali-jumptrading:
snaprd can publish up to two frags in after credit. Once when publishing the contents of the snapshot, and once when sending a control message to downstream tiles.


ripatel-fd:
@cali-jumptrading Idea: add a "handholding" check to fd_stem that logs if the STEM_BURST limit is exceeded after an after_credit callback.

PR #5625, opened 2025-07-10
runtime, replay: refactor replay execution state management

yufeng-jump:
- Fix a bug which caused exec tiles to be stranded in certain states forever on fork switches.
- Fix a race condition which caused potentially conflicting transactions to be dispatched for parallel execution.
- Documentation.


yufeng-jump:
#5688

PR #5626, opened 2025-07-11
send: quic signs tls CV with identity key, target tpu_vote_quic port

akhinvasara-jumptrading:
This PR uses a staked quic connection to send txns, by signing the tls CertificateVerify with the validator's identity key. We can now target the tpu quic vote port specifically. 

A follow-up PR will make the signing asynchronous, and we will target multiple ports once the gossip re-write is complete.


src/waltz/tls/fd_tls.h
```diff
@@ -65,6 +65,32 @@
...
+  { 0xe3, 0xb0, 0xc4, 0x42, 0x98, 0xfc, 0x1c, 0x14,
+    0x9a, 0xfb, 0xf4, 0xc8, 0x99, 0x6f, 0xb9, 0x24,
+    0x27, 0xae, 0x41, 0xe4, 0x64, 0x9b, 0x93, 0x4c,
+    0xa4, 0x95, 0x99, 0x1b, 0x78, 0x52, 0xb8, 0x55 };
+
+static uchar const handshake_derived[ 32 ] =
+  { 0x6f, 0x26, 0x15, 0xa1, 0x08, 0xc7, 0x02, 0xc5,
+    0x67, 0x8f, 0x54, 0xfc, 0x9d, 0xba, 0xb6, 0x97,
+    0x16, 0xc0, 0x76, 0x18, 0x9c, 0x48, 0x25, 0x0c,
+    0xeb, 0xea, 0xc3, 0x57, 0x6c, 0x36, 0x11, 0xba };
```
ripatel-fd: There's no point in moving these to the public API, could you move them back?

src/waltz/tls/fd_tls.h
```diff
@@ -65,6 +65,32 @@
...
 
+
+/* Pre-generated keys *********************************************************/
+
+static char const fd_tls13_cli_sign_prefix[ 98 ] =
+  "                                "  /* 32 spaces */
+  "                                "  /* 32 spaces */
+  "TLS 1.3, client CertificateVerify";
+
+static char const fd_tls13_srv_sign_prefix[ 98 ] =
```
ripatel-fd: This would define the symbol in every compile unit that includes fd_tls.h which isn't ideal. I suggest making this externally linked if you want to reuse the symbol

mmcgee-jump:
keyguard piece looks ok 👍

jherrera-jump (approved):
LGTM

PR #5628, opened 2025-07-11
snapshots: tweak control flow

mmcgee-jump:


src/disco/stem/fd_stem.c
```diff
@@ -641,6 +641,17 @@ STEM_(run1)( ulong                        in_cnt,
       continue;
     }
 
+#ifdef STEM_CALLBACK_RETURNABLE_FRAG
```
cali-jumptrading, line 669: Thinking more about this:
- This RETURNABLE_FRAG construct is actually just a hack on top of an inflexible fd_stem that wasn't designed to allow returnable frags. Now that you've added this RETURNABLE_FRAG construct, it creates confusion between RETURNABLE_FRAG and AFTER_FRAG. Should users use RETURNABLE_FRAG, AFTER_FRAG, or both? What happens if both RETURNABLE_FRAG and AFTER_FRAG are defined in the same tile?

Please document this RETURNABLE_FRAG construct at the very least, if you still want to merge the RETURNABLE_FRAG construct in. Alternatively, you could just make AFTER_FRAG have the ability to return whether the frag was consumed and avoid the confusion of yet another stem callback.

src/discof/restore/Local.mk
```diff
@@ -11,9 +11,6 @@ ifdef FD_HAS_INT128
...
 endif
-$(call add-objs,utils/fd_snapshot_reader,fd_discof)
-$(call add-objs,utils/fd_snapshot_file,fd_discof)
-$(call add-objs,utils/fd_snapshot_archive,fd_discof)
-$(call add-objs,utils/fd_snapshot_httpdl,fd_discof)
-$(call add-objs,utils/fd_snapshot_peers_manager,fd_discof)
-$(call add-objs,utils/fd_icmp_ping,fd_discof)
+$(call add-objs,utils/fd_ssping,fd_discof)
+$(call add-objs,utils/fd_sshttp,fd_discof)
+$(call add-objs,utils/fd_ssarchive,fd_discof)
```
cali-jumptrading, line 16: remove ssarchive if you're not going to use it

src/discof/restore/fd_snapdc_tile.c
```diff
@@ -63,245 +47,198 @@ struct fd_snapdc_tile {
...
+  FD_TEST( chunk>=ctx->in.chunk0 && chunk<=ctx->in.wmark && sz<=ctx->in.mtu && sz>=ctx->data_pos );
+
+  uchar const * data = fd_chunk_to_laddr_const( ctx->in.wksp, chunk );
+
+  uchar const * in  = data+ctx->data_pos;
+  uchar * out = fd_chunk_to_laddr( ctx->out.wksp, ctx->out.chunk );
+  ulong in_consumed = 0UL, out_produced = 0UL;
+  ulong error = ZSTD_decompressStream_simpleArgs( ctx->zstd,
+                                                  out,
+                                                  USHORT_MAX,
```
cali-jumptrading: Use the zstd link's MTU value here instead.

src/discof/restore/fd_snapdc_tile.c
```diff
@@ -63,245 +47,198 @@ struct fd_snapdc_tile {
...
+      ctx->state = FD_SNAPDC_STATE_SHUTDOWN;
       break;
-    }
     default:
-      FD_LOG_ERR(( "snapdc: unexpected sig %lu", sig ));
+      FD_LOG_ERR(( "unexpected control sig %lu", sig ));
+      return;
   }
+
+  fd_stem_publish( stem, 0UL, sig, ctx->out.chunk, 0UL, 0UL, 0UL, 0UL );
```
cali-jumptrading: It's quite confusing to have this reset logic at the end of this function. It would be much more readable if this was an appropriately named function that is called for every state that requires a reset.

src/discof/restore/fd_snapdc_tile.c
```diff
@@ -1,58 +1,42 @@
...
-   receives a tagged end-of-message notification from snaprd. */
-#define FD_SNAPDC_STATE_DONE                      (3)
+#define FD_SNAPDC_STATE_DECOMPRESSING (0) /* We are in the process of decompressing a valid stream */
+#define FD_SNAPDC_STATE_FINISHING     (1) /* The frame has been fully decompressed, we are waiting to make sure the snapshot has no more data */
+#define FD_SNAPDC_STATE_MALFORMED     (2) /* The decompression stream is malformed, we are waiting for a reset notification */
+#define FD_SNAPDC_STATE_DONE          (3) /* The decompression stream is done, the tile is waiting for a shutdown message */
+#define FD_SNAPDC_STATE_SHUTDOWN      (4) /* The tile is done, been told to shut down, and has likely already exited */
 
 struct fd_snapdc_tile {
+  int full;
```
cali-jumptrading, line 28: In my opinion, using a boolean flag to indicate whether the tile is decompressing the full or incremental snapshot is not clean and detracts from readability. Now a reader has to understand that most state is contained by the state machine, but there is additional state not documented by the state machine (ctx->full). Additionally, it would be helpful for the UI to know which stream the decompressor tile thinks it's decompressing.

src/discof/restore/fd_snapdc_tile.c
```diff
@@ -63,245 +47,198 @@ struct fd_snapdc_tile {
...
-        decompressed_bytes,
-        0UL,
-        0UL,
-        0UL );
-      ctx->out.chunk = fd_dcache_compact_next( ctx->out.chunk,
-                                               decompressed_bytes,
-                                               ctx->out.chunk0,
-                                               ctx->out.wmark );
-      num_produced_frags++;
+  if( FD_UNLIKELY( !error ) ) {
```
cali-jumptrading, line 217: please document this error handling case. Why does this error handling case occur after publishing a data frag and accumulating metrics?

src/discof/restore/fd_snapdc_tile.c
```diff
@@ -345,16 +284,15 @@ unprivileged_init( fd_topo_t *      topo,
                  (ulong)scratch + scratch_footprint( tile ) ));
 }
 
-#define STEM_BURST SNAPDC_BURST_MAX
-#define STEM_LAZY  1e3L
+#define STEM_BURST 1UL
```
cali-jumptrading: This might not always be 1. If while decompressing the zstd stream, you publish a valid data fragment and then hit a transition_malformed, you would publish two frags. Please update STEM_BURST or update the decompressing logic accordingly.

Since you are publishing to two different consumers, it might be ok, but given that `fd_stem` is only designed to handle backpressure for one producer, it might make fd_stem's credits available go below 0 (which actually just overflows)

This is why `STEM_BURST` is a horrible abstraction. It is not only hard to reason about, it is also just plain wrong.

src/discof/restore/fd_snapdc_tile.c
```diff
@@ -63,245 +47,198 @@ struct fd_snapdc_tile {
...
-                       0UL,
-                       0UL,
-                       0UL,
-                       0UL,
-                       0UL );
-      ctx->shutdown = 1;
+    case FD_SNAPSHOT_MSG_CTRL_RESET_FULL:
+      ctx->state = FD_SNAPDC_STATE_DECOMPRESSING;
+      ctx->full = 1;
+      ctx->metrics.full.compressed_bytes_read   = 0UL;
```
cali-jumptrading, line 116: can we just wrap this duplicated reset logic into a reset function?

src/discof/restore/fd_snapin_tile.c
```diff
@@ -463,73 +267,162 @@ handle_data_frag( fd_snapin_tile_t * ctx,
...
+  fd_topo_wksp_t const * in_wksp = &topo->workspaces[ topo->objs[ in_link->dcache_obj_id ].wksp_id ];
+  ctx->in.wksp                   = in_wksp->wksp;;
+  ctx->in.chunk0                 = fd_dcache_compact_chunk0( ctx->in.wksp, in_link->dcache );
+  ctx->in.wmark                  = fd_dcache_compact_wmark( ctx->in.wksp, in_link->dcache, in_link->mtu );
+  ctx->in.mtu                    = in_link->mtu;
 }
 
-#define STEM_BURST                  1UL
-#define STEM_LAZY                   1e3L
+#define STEM_BURST 1UL
```
cali-jumptrading: Similar comment as the snapdc tile. If you think this is ok, please document why.

Also, there is a case where you might publish more than one frag: when sending the manifest and the snapshot done message to the replay tile.

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -229,356 +75,157 @@ struct fd_snaprd_tile {
...
-  } else {
-    FD_LOG_ERR(( "snaprd: unexpected state %d for retry", ctx->state ));
+  if( FD_UNLIKELY( !result ) ) {
+    switch( ctx->state ) {
+      case FD_SNAPRD_STATE_READING_INCREMENTAL_FILE:
+        fd_stem_publish( stem, 0UL, FD_SNAPSHOT_MSG_CTRL_DONE, 0UL, 0UL, 0UL, 0UL, 0UL );
+        ctx->state = FD_SNAPRD_STATE_FLUSHING_INCREMENTAL_FILE;
+        break;
+      case FD_SNAPRD_STATE_READING_FULL_FILE:
+        if( FD_LIKELY( ctx->config.incremental_snapshot_fetch ) ) {
```
cali-jumptrading, line 153: After reading a full snapshot from disk, snaprd should be able to download an incremental snapshot. I don't see the logic to handle this. If you are not handling this particular case, please add it in. The state machine has already grown so complicated that we shouldn't leave out any state transitions right now.

src/discof/restore/utils/fd_ssarchive.h
```diff
@@ -0,0 +1,26 @@
...
+   gives the highest incremental slot.  The incremental slot is set to
+   ULONG_MAX if the highest pair is just a full snapshot alone, or if
+   incremental_snapshot is 0.
+
+   Returns -1 on failure, and 0 on success. */
+
+int
+fd_ssarchive_latest_pair( char const * directory,
+                          int          incremental_snapshot,
+                          ulong *      full_slot,
```
cali-jumptrading, line 19: I would probably consolidate these return parameters into a struct of some sorts. Not required for this PR though

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -587,178 +234,254 @@ after_credit( fd_snaprd_tile_t *  ctx,
...
-        fd_snapshot_peers_manager_reset_pings( ctx->peers_manager );
+      fd_ip4_port_t best = fd_ssping_best( ctx->ssping );
+      if( FD_UNLIKELY( !best.l ) ) {
+        ctx->state = FD_SNAPRD_STATE_WAITING_FOR_PEERS;
+        break;
+      }
 
-        ctx->state = fd_snaprd_shared_state_transition( ctx );
+      ulong highest_cluster_slot = 0UL; /* TODO: Implement */
+      if( FD_LIKELY( ctx->local.full_snapshot_slot!=ULONG_MAX && ctx->local.full_snapshot_slot>=fd_ulong_sat_sub( highest_cluster_slot, ctx->config.maximum_local_snapshot_age ) ) ) {
```
cali-jumptrading, line 270: we should use the incremental snapshot slot todetermine the snapshot age. See `get_highest_local_snapshot_hash` in agave.

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -229,356 +75,157 @@ struct fd_snaprd_tile {
...
+          }
+          ctx->state = FD_SNAPRD_STATE_FLUSHING_FULL_HTTP;
+          break;
+        default:
+          break;
+      }
+      break;
+    }
+    case FD_SSHTTP_ADVANCE_DATA: {
+      switch( ctx->state ) {
```
cali-jumptrading, line 210: having switch statements within switch statements in my opinion is not readable nor maintainable. It is hard to read and prone to errors. Changing the state machine requires tracking down every single swtich statement and failure to track down every switch statement results in bugs.

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -229,356 +75,157 @@ struct fd_snaprd_tile {
...
-    case FD_SNAPRD_STATE_COLLECTING_RESPONSES_INCREMENTAL:
-      return FD_SNAPRD_STATE_INCREMENTAL_DOWNLOAD;
-    default: {
-      FD_LOG_ERR(( "snaprd: unexpected state %d for shared state transition", ctx->state ));
-      return -1; /* error */
+      fd_stem_publish( stem, 0UL, FD_SNAPSHOT_MSG_DATA, ctx->out.chunk, data_len, 0UL, 0UL, 0UL );
+      ctx->out.chunk = fd_dcache_compact_next( ctx->out.chunk, data_len, ctx->out.chunk0, ctx->out.wmark );
+      break;
     }
+    default: break;
```
cali-jumptrading: FD_LOG_ERR here if the returned state is not expected

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -587,178 +234,254 @@ after_credit( fd_snaprd_tile_t *  ctx,
...
+    case FD_SNAPRD_STATE_FLUSHING_INCREMENTAL_HTTP:
+      if( FD_UNLIKELY( ctx->ack_cnt<2UL ) ) break;
+      ctx->ack_cnt = 0UL;
 
-/* SnapRd Tile Functions **********************************************/
+      ctx->state = FD_SNAPRD_STATE_SHUTDOWN;
+      fd_stem_publish( stem, 0UL, FD_SNAPSHOT_MSG_CTRL_SHUTDOWN, 0UL, 0UL, 0UL, 0UL, 0UL );
+      break;
+    case FD_SNAPRD_STATE_FLUSHING_FULL_FILE:
+      if( FD_UNLIKELY( ctx->ack_cnt<2UL ) ) break;
```
cali-jumptrading: Can we not use magic constant numbers everywhere? This is a common source of bugs and can easily be resolved by defining these constants in a common header file. The number of tiles in the snapshot loader is not set in stone right now and might be changed in the future.

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -21,203 +22,48 @@
...
-
-/* fd_snaprd_download_pair indicates whether
-   the full and incremental snapshots need to be downloaded or not. */
-struct fd_snaprd_download_pair {
-  int full;
-  int incremental;
-};
-
-typedef struct fd_snaprd_download_pair fd_snaprd_download_pair_t;
+#define FD_SNAPRD_STATE_WAITING_FOR_PEERS         ( 0) /* Waiting for first peer to arrive from gossip to download from */
```
cali-jumptrading, line 25: The snaprd tile's state machine has become quite complex. Please document the state transitions here and in the state machine in after_credit. What are the conditions for each state transition?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -229,356 +75,157 @@ struct fd_snaprd_tile {
...
-                    FD_SNAPSHOT_MSG_CTRL_FULL_DONE,
-                    ctx->out.chunk,
-                    0UL,
-                    0UL,
-                    0UL,
-                    0UL );
-}
+  FD_TEST( ctx->state==FD_SNAPRD_STATE_READING_INCREMENTAL_FILE || ctx->state==FD_SNAPRD_STATE_READING_FULL_FILE );
+  int full = ctx->state==FD_SNAPRD_STATE_READING_FULL_FILE;
+  long result = read( full ? ctx->local.full_snapshot_fd : ctx->local.incremental_snapshot_fd , out, ctx->out.mtu );
```
cali-jumptrading, line 131: ok this is pretty ugly. just use a switch statement here between the full reading state and incremental reading state. Maybe put the call to `read` and its error handling into a function so you can just write a simple switch statement.

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -0,0 +1,146 @@
...
+   incremental, the incremental slot will be set to ULONG_MAX, otherwise
+   it is set to the incremental slot number.  On success, the full slot
+   and the snapshot hash are always set.  The hash will be the base58
+   decoded hash. */
+
+int
+fd_ssarchive_parse( char const * filename,
+                    ulong *      full_slot,
+                    ulong *      incremental_slot,
+                    uchar        hash[ static 32UL ] );
```
cali-jumptrading, line 25: Ok so this is another example of using magic constant numbers. If possible, please find a definition for the size of a hash or explain why we don't have one and leave a TODO.

src/discof/restore/fd_snapin_tile.c
```diff
@@ -463,73 +267,162 @@ handle_data_frag( fd_snapin_tile_t * ctx,
...
 
-  /* handle frag */
-  if( FD_UNLIKELY( sz==0 ) ) handle_control_frag( ctx, stem, sig );
-  else                       handle_data_frag( ctx, ctx->in._chunk, sz );
+  FD_TEST( ctx->state!=FD_SNAPIN_STATE_SHUTDOWN );
+
+  if( FD_UNLIKELY( sig==FD_SNAPSHOT_MSG_DATA ) ) handle_data_frag( ctx, chunk, sz, stem );
+  else                                           handle_control_frag( ctx, stem, sig  );
+
+  return 0;
```
cali-jumptrading, line 360: ok so you used `RETURNABLE_FRAG` but always return 0, so it's bascially the same as AFTER_FRAG. Just use AFTER_FRAG here.

src/discof/restore/fd_snapin_tile.c
```diff
@@ -1,143 +1,125 @@
...
-  /* Input */
-  struct {
-    fd_wksp_t *  wksp;
-    ulong        chunk0;
-    ulong        wmark;
-    ulong        _chunk;
-  } in;
-
-  /* State machine */
+  int full;
```
cali-jumptrading, line 29: Same as the snapdc tile, I think this additional state should just be in the state machine.

cali-jumptrading (commented):
First pass over this PR. I haven't taken a close look at the pinger yet, but I think its solid. 

If you decide not to address these comments, please explain why.

src/discof/restore/Local.mk
cali-jumptrading: oh oops, got confused by the branch. ignore

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -229,356 +75,157 @@ struct fd_snaprd_tile {
...
+      break;
+    }
+    case FD_SSHTTP_ADVANCE_DONE: {
+      switch( ctx->state ) {
+        case FD_SNAPRD_STATE_READING_INCREMENTAL_HTTP:
+          fd_stem_publish( stem, 0UL, FD_SNAPSHOT_MSG_CTRL_DONE, 0UL, 0UL, 0UL, 0UL, 0UL );
+          ctx->state = FD_SNAPRD_STATE_FLUSHING_INCREMENTAL_HTTP;
+          break;
+        case FD_SNAPRD_STATE_READING_FULL_HTTP:
+          if( FD_LIKELY( ctx->config.incremental_snapshot_fetch ) ) {
```
cali-jumptrading, line 197: This logic to figure out which control mesage to send is duplicated across both `read_file_data` and `read_http_data`, as well as the logic to accumulate metrics. This is an example of fragile code that is hard to read, hard to maintain, prone to errors, and hard to scale.

I will probably fix this later, but in my opinion, if we keep writing code like this, we'll have to spend a lot more time in the future fixing bugs and redesigning code.

src/discof/restore/fd_snapdc_tile.c
```diff
@@ -1,58 +1,42 @@
...
 
 struct fd_snapdc_tile {
+  int full;
   int state;
-  int shutdown;
 
-  fd_zstd_dstream_t * dstream;
+  ZSTD_DCtx * zstd;
+
+  ulong data_pos;
```
cali-jumptrading: Could we store data_pos with the in struct? It's part of the receiver state so it makes sense to me to store it there.

src/discof/restore/utils/fd_ssping.c
```diff
@@ -0,0 +1,503 @@
...
+#include "../../../util/tmpl/fd_treap.c"
+
+#define DLIST_NAME  deadline_list
+#define DLIST_ELE_T fd_ssping_peer_t
+#define DLIST_PREV  deadline.prev
+#define DLIST_NEXT  deadline.next
+
+#include "../../../util/tmpl/fd_dlist.c"
+
+struct fd_ssping_private {
```
cali-jumptrading, line 98: Would be nice to have more documentation in this file as well to explain each state.... not required for this PR.

I like this queue design a lot, so much more cleaner than what I came up with.

src/discof/restore/utils/fd_ssping.c
```diff
@@ -0,0 +1,503 @@
...
+  deadline_list_t *  pinged;
+  deadline_list_t *  valid;
+  deadline_list_t *  refreshing;
+  deadline_list_t *  invalid;
+
+  ulong           fds_len;
+  struct pollfd * fds;
+  ulong *         fds_idx;
+
+  ulong              magic; /* ==FD_SSPING_MAGIC */
```
cali-jumptrading, line 113: alignment nit

src/discof/restore/utils/fd_ssping.c
```diff
@@ -0,0 +1,503 @@
...
+void
+fd_ssping_advance( fd_ssping_t * ssping,
+                   long          now ) {
+  while( !deadline_list_is_empty( ssping->unpinged, ssping->pool ) ) {
+    fd_ssping_peer_t * peer = deadline_list_ele_pop_head( ssping->unpinged, ssping->pool );
+
+    FD_LOG_INFO(( "pinging " FD_IP4_ADDR_FMT ":%hu", FD_IP4_ADDR_FMT_ARGS( peer->addr.addr ), peer->addr.port ));
+    int result = peer_connect( ssping, peer );
+    if( FD_UNLIKELY( -1==result ) ) {
+      peer->state = PEER_STATE_INVALID;
```
cali-jumptrading, line 428: alignment nit

cali-jumptrading (commented):
Generally, not a fan of naming things `sspinger` and `ssarchive`. In my opinion, we should just be explicit to avoid confusion.

src/discof/restore/fd_snapdc_tile.c
mmcgee-jump: This simplifies the state machine logic, since we don't have to duplicate many of the states. The UI already knows the snapshot being decompressed by the snaprd tile metrics.

src/discof/restore/fd_snapdc_tile.c
mmcgee-jump: All of the control messages require a reset. I can put it at the top if you prefer?

src/discof/restore/fd_snapdc_tile.c
mmcgee-jump: The ordering of these three doesn't really matter, except I prefer the metrics to increment after reading the frag (`read` counter is post-hoc).

src/discof/restore/fd_snaprd_tile.c
mmcgee-jump: This is the only line of code that will ever use this value, it doesn't make sense to move it somewhere else. It will also not cause bugs: if the value is wrong the application hangs at boot, making it immediately recognizable and unshippable.

src/discof/restore/utils/fd_ssping.c
mmcgee-jump: Not a fan of aligning in this case

src/discof/restore/fd_snaprd_tile.c
mmcgee-jump: This doesn't bother me. These two readers are very simple, right next to one another so easily discoverable, and trying to tie the logic together is going to introduce unnecessary coupling between them.

src/discof/restore/fd_snapin_tile.c
mmcgee-jump: I used it to get the `chunk` and `stem` at the same time, which we discourage with most other tiles, since they should be written to get ovverrun (that wouldn't make sense here).

src/discof/restore/fd_snapdc_tile.c
mmcgee-jump: Prefer not to here, very localized duplication which keeps the state machine clear and readable.

src/discof/restore/utils/fd_ssarchive.c
mmcgee-jump: The size of a sha256 hash is obvious and assumed throughout the codebase. It will never change.

src/discof/restore/fd_snapin_tile.c
mmcgee-jump: Tend to disagree, there are two duplicated state machines here, with trivial differences (whether it transitions out to done, or incremental), so better to have represent that model in the code (one state machine, and one flag of which one of the two it is).

src/discof/restore/fd_snapdc_tile.c
cali-jumptrading: Maybe just adding a comment is fine. Something like `all control messages go through a reset`?

src/discof/restore/utils/fd_ssarchive.c
cali-jumptrading: Ok, I'd still prefer to use a defined constant because I've been underfire before for not using defined constants in Jump Classic. But if you think this is ok, that is fine.

src/discof/restore/fd_snapdc_tile.c
cali-jumptrading: Ok, I still disagree here because I think it's cleaner to express all state in the state machine, rather than trying to simplify the state machine. But will leave it up to you.

src/discof/restore/fd_snapdc_tile.c
cali-jumptrading: ok, I would still prefer de-duplicating here, as this doesn't involve introducing an abstraction.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: in my mind, the nested ternary expression is hard to read and easy to miss when scanning the file. If you feel this is fine, feel free to ignore.

src/discof/restore/fd_snapin_tile.c
cali-jumptrading: ok I still disagree but if you think it's fine, feel free to ignore.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: ok I still disagree but I do agree that it's hard to unify the duplicated code and my original attempt was certaintly a very messy one.

cali-jumptrading (approved):

PR #5631, opened 2025-07-11
snapshots: send manifest

cali-jumptrading:
make the snapin tile send the manifest every time it parses one.

### DIFF
--- a/src/app/firedancer-dev/commands/backtest.c
+++ b/src/app/firedancer-dev/commands/backtest.c
@@ -311,7 +311,7 @@ backtest_topo( config_t * config ) {
 
   /* Replay decoded manifest dcache topo obj */
   fd_topo_obj_t * replay_manifest_dcache = fd_topob_obj( topo, "dcache", "replay_manif" );
-  fd_pod_insertf_ulong( topo->props, 2UL << 30UL, "obj.%lu.data_sz", replay_manifest_dcache->id );
+  fd_pod_insertf_ulong( topo->props, (4 * 1UL << 30UL /* gib */), "obj.%lu.data_sz", replay_manifest_dcache->id );
   fd_pod_insert_ulong(  topo->props, "manifest_dcache", replay_manifest_dcache->id );
 
   fd_topob_tile_uses( topo, snapin_tile, funk_obj, FD_SHMEM_JOIN_MODE_READ_WRITE );
--- a/src/app/firedancer-dev/commands/snapshot_load.c
+++ b/src/app/firedancer-dev/commands/snapshot_load.c
@@ -46,7 +46,7 @@ snapshot_load_topo( config_t *     config,
      manifest */
   fd_topob_wksp( topo, "replay_manif" );
   fd_topo_obj_t * replay_manifest_dcache = fd_topob_obj( topo, "dcache", "replay_manif" );
-  fd_pod_insertf_ulong( topo->props, 1UL << 30UL, "obj.%lu.data_sz", replay_manifest_dcache->id );
+  fd_pod_insertf_ulong( topo->props, (4 * 1UL << 30UL /* gib */), "obj.%lu.data_sz", replay_manifest_dcache->id );
   fd_pod_insert_ulong(  topo->props, "manifest_dcache", replay_manifest_dcache->id );
 
   /* read() tile */
--- a/src/app/firedancer/topology.c
+++ b/src/app/firedancer/topology.c
@@ -381,7 +381,7 @@ fd_topo_initialize( config_t * config ) {
 
   /* Replay decoded manifest dcache topo obj */
   fd_topo_obj_t * replay_manifest_dcache = fd_topob_obj( topo, "dcache", "replay_manif" );
-  fd_pod_insertf_ulong( topo->props, 2UL << 30UL, "obj.%lu.data_sz", replay_manifest_dcache->id );
+  fd_pod_insertf_ulong( topo->props, (4 * 1UL << 30UL /* gib */), "obj.%lu.data_sz", replay_manifest_dcache->id );
   fd_pod_insert_ulong(  topo->props, "manifest_dcache", replay_manifest_dcache->id );
 
   ushort parsed_tile_to_cpu[ FD_TILE_MAX ];
--- a/src/discof/replay/fd_replay_tile.c
+++ b/src/discof/replay/fd_replay_tile.c
@@ -656,6 +656,7 @@ restore_slot_ctx( fd_replay_tile_ctx_t * ctx,
 
   fd_solana_manifest_global_t * manifest_global
     = (fd_solana_manifest_global_t *)fd_chunk_to_laddr( fd_wksp_containing( ctx->manifest_dcache ), chunk );
+
   fd_exec_slot_ctx_t * recovered_slot_ctx = fd_exec_slot_ctx_recover( ctx->slot_ctx,
                                                                       manifest_global,
                                                                       ctx->runtime_spad );
@@ -985,6 +986,7 @@ on_snapshot_message( fd_replay_tile_ctx_t * ctx,
     }
     case FD_FULL_SNAPSHOT_MANIFEST_EXTERNAL:
     case FD_INCREMENTAL_SNAPSHOT_MANIFEST_EXTERNAL: {
+      FD_LOG_NOTICE(( "Received decoded global snapshot manifest message" ));
       /* We may either receive a full snapshot manifest or an
          incremental snapshot manifest.  Note that this external message
          id is only used temporarily because replay cannot yet receive
--- a/src/discof/restore/fd_snapin_tile.c
+++ b/src/discof/restore/fd_snapin_tile.c
@@ -38,14 +38,20 @@ struct fd_snapin_tile {
      manifest. */
   fd_lthash_value_t lthash;
 
+  /* stem pointer for snapshot parser callbacks */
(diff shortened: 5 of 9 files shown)


ibhatt-jumptrading (commented):
As discussed in pm, needs this so we don't try to clone the same bank twice
> 
> /* If the bank already exists, that means that we have already
>      restored the bank for this slot and we can return out. */
>   fd_bank_t * bank = fd_banks_get_bank( slot_ctx->banks, manifest->bank.slot );
>   if( FD_UNLIKELY( !!bank ) ) {
>     FD_LOG_NOTICE(( "The bank for slot %lu already exists. Skipping manifest restore", manifest->bank.slot ));
>     slot_ctx->bank = bank;
>     return slot_ctx;
>   }

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -302,7 +302,7 @@ backtest_topo( config_t * config ) {
 
   /* Replay decoded manifest dcache topo obj */
   fd_topo_obj_t * replay_manifest_dcache = fd_topob_obj( topo, "dcache", "replay_manif" );
-  fd_pod_insertf_ulong( topo->props, 2UL << 30UL, "obj.%lu.data_sz", replay_manifest_dcache->id );
+  fd_pod_insertf_ulong( topo->props, 4UL << 30UL, "obj.%lu.data_sz", replay_manifest_dcache->id );
```
lidatong: changing the shift base is a little strange. is this just approximating the manifest size because of variable length structures inside it? is there a hard bound we can derive from max unrooted banks, etc. ?

src/discof/replay/fd_replay_tile.c
```diff
@@ -656,6 +656,15 @@ restore_slot_ctx( fd_replay_tile_ctx_t * ctx,
...
   fd_solana_manifest_global_t * manifest_global
     = (fd_solana_manifest_global_t *)fd_chunk_to_laddr( fd_wksp_containing( ctx->manifest_dcache ), chunk );
+
+  /* If the bank already exists, that means that we have already
+    restored the bank for this slot. */
+  fd_bank_t * bank = fd_banks_get_bank( ctx->slot_ctx->banks, manifest_global->bank.slot );
+  if( FD_UNLIKELY( !!bank ) ) {
+    FD_LOG_NOTICE(( "The bank for slot %lu already exists, skipping slot context restore", manifest_global->bank.slot ));
+    return;
+  }
```
lidatong: hm - is the bank entirely derived from manifest? so multiple banks can get restored if a snapshot gets cancelled but manifest is already sent?

src/discof/restore/utils/fd_snapshot_parser.h
```diff
@@ -52,29 +52,36 @@ typedef struct fd_snapshot_accv_map fd_snapshot_accv_map_t;
...
+struct fd_stem_context;
+typedef struct fd_stem_context fd_stem_context_t;
+
 struct fd_snapshot_parser;
 typedef struct fd_snapshot_parser fd_snapshot_parser_t;
 
 typedef void
 (* fd_snapshot_parser_process_manifest_fn_t)( fd_snapshot_parser_t *        parser,
                                               void *                        _ctx,
+                                              fd_stem_context_t *           stem,
```
lidatong: seems like this callback API already supports a generic ctx parameter. can the stem be passed in via that instead? otherwise this API won't work outside a stem

src/app/firedancer-dev/commands/backtest.c
cali-jumptrading: well, in a normal snapshot load, we would expect two manifests. A mainnet snapshot manifest can be 2GB right now. I think we can define for ourselves a hard bound but I don't think a hard bound exists on Agave's side.

src/discof/replay/fd_replay_tile.c
cali-jumptrading: working that out with @ibhatt-jumptrading right now. I think he's going to make an API to properly clear a bank. We decided to always restore the latest manifest, so we won't skip restoring the manifest if the bank already exists, we'll just clear the bank and restore it again.

src/discof/restore/utils/fd_snapshot_parser.h
cali-jumptrading: hmm... good point. Initially I felt it was weird to store stem inside the context, as the stem context is created on the fly in `fd_stem` and might not always be valid.... but I also dislike this approach of passing stem as a parameter.

src/app/firedancer-dev/commands/backtest.c
lidatong: yeah, ultimately we do pick a somewhat arbitrary bound (eg. "this bound lets us run for 10 minutes without rooting a slot before OOM"). i think our goal is to unify that same bound across everything -- think it would apply to the manifest structures too

if it turns out to be too small for just the manifest, then there probably was a serious misestimation of a relevant bound on our part (or something went wrong with the cluster and it's still producing snapshots)

cali-jumptrading: I think I upped this bound because publishing to an external dcache has no credits tracking. So I want the buffer to be big enough to store at least two manifests. This external dcache should go away when the bank no longer relies on `fd_types` so it just needs to be big enough for now to allow us to buffer 2 manifests if needed.

src/discof/restore/utils/fd_snapshot_parser.h
cali-jumptrading: updated to use stem pointer in context instead.

src/app/firedancer-dev/commands/backtest.c
lidatong: gotcha -- minor nit but `1 << 32` will be more readable

cali-jumptrading: oh I see... what if I just did `4 * 1024 * 1024 * 1024`?

src/discof/restore/fd_snapin_tile.c
```diff
@@ -208,8 +263,9 @@ snapshot_copy_acc_data( fd_snapshot_parser_t * parser FD_PARAM_UNUSED,
 }
 
 static void
-snapshot_reset_acc_data( fd_snapshot_parser_t * parser FD_PARAM_UNUSED,
+snapshot_reset_acc_data( fd_snapshot_parser_t * parser,
                          void *                 _ctx ) {
+  (void)parser;
```
lidatong, line 250: why `FD_PARAM_UNUSED` to `(void)`?

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -302,7 +302,7 @@ backtest_topo( config_t * config ) {
 
   /* Replay decoded manifest dcache topo obj */
   fd_topo_obj_t * replay_manifest_dcache = fd_topob_obj( topo, "dcache", "replay_manif" );
-  fd_pod_insertf_ulong( topo->props, 2UL << 30UL, "obj.%lu.data_sz", replay_manifest_dcache->id );
+  fd_pod_insertf_ulong( topo->props, (4 * 1UL << 30UL), "obj.%lu.data_sz", replay_manifest_dcache->id );
```
lidatong:
```suggestion
  fd_pod_insertf_ulong( topo->props, 4 * (1UL << 30UL /* gib */), "obj.%lu.data_sz", replay_manifest_dcache->id );
```

src/app/firedancer-dev/commands/snapshot_load.c
```diff
@@ -46,7 +46,7 @@ snapshot_load_topo( config_t *     config,
      manifest */
   fd_topob_wksp( topo, "replay_manif" );
   fd_topo_obj_t * replay_manifest_dcache = fd_topob_obj( topo, "dcache", "replay_manif" );
-  fd_pod_insertf_ulong( topo->props, 1UL << 30UL, "obj.%lu.data_sz", replay_manifest_dcache->id );
+  fd_pod_insertf_ulong( topo->props, (4 * 1UL << 30UL), "obj.%lu.data_sz", replay_manifest_dcache->id );
```
lidatong:
```suggestion
  fd_pod_insertf_ulong( topo->props, 4 * (1UL << 30UL /* gib */),, "obj.%lu.data_sz", replay_manifest_dcache->id );
```

src/app/firedancer/topology.c
```diff
@@ -377,7 +377,7 @@ fd_topo_initialize( config_t * config ) {
 
   /* Replay decoded manifest dcache topo obj */
   fd_topo_obj_t * replay_manifest_dcache = fd_topob_obj( topo, "dcache", "replay_manif" );
-  fd_pod_insertf_ulong( topo->props, 2UL << 30UL, "obj.%lu.data_sz", replay_manifest_dcache->id );
+  fd_pod_insertf_ulong( topo->props, (4 * 1UL << 30UL), "obj.%lu.data_sz", replay_manifest_dcache->id );
```
lidatong:
```suggestion
  fd_pod_insertf_ulong( topo->props, 4 * (1UL << 30UL /* gib */),, "obj.%lu.data_sz", replay_manifest_dcache->id );
```

lidatong (dismissed):

src/discof/restore/fd_snapin_tile.c
cali-jumptrading: I just thought being consistent with using (void) was better. let me know if you disagree

PR #5632, opened 2025-07-11
zen5 machine target on gcc 14+

nlgripto:


SEJeff:
> To which I then upgrade to gcc 14, which does not actually support `-mtune=znver5`

We definitely want gcc 15 for [all](https://gcc.gnu.org/pipermail/gcc-cvs/2024-September/408729.html) [of](https://gcc.gnu.org/pipermail/gcc-cvs/2024-September/408732.html) [the](https://gcc.gnu.org/pipermail/gcc-cvs/2024-September/408735.html) [zen5](https://gcc.gnu.org/pipermail/gcc-cvs/2024-September/408740.html) [patches](https://gcc.gnu.org/pipermail/gcc-cvs/2024-September/408742.html).

Here are some [generic benchmarks](https://www.phoronix.com/review/gcc-15-amd-zen5) using the phoronix test suite.

nlgripto:
~~yep, havent finished updating the PR yet~~

config/machine/linux_gcc_zen5.mk
```diff
@@ -0,0 +1,49 @@
...
+include config/extra/with-security.mk
+include config/extra/with-brutality.mk
+include config/extra/with-optimization.mk
+include config/extra/with-threads.mk
+
+# GCC 8 (Firedancer's minimum supported GCC version) only supports znver1.
+# note: I dont agree with this behavior but inheriting it from zen4.mk.
+ifeq ($(shell $(CC) -dumpversion),8)
+CPPFLAGS+=-march=znver1 -mtune=znver1
+else
```
SEJeff: Should we not just barf here with "wrong gcc for this target" instead of lies @ripatel-fd? Otherwise the node operator thinks they have an optimized build when they in fact do not.

ripatel-fd: I agree with @SEJeff. Remove the GCC version 8 check and just keep the below rule that bans anything below GCC 15.

PR #5634, opened 2025-07-11
quic: Improve loss detection with time threshold 

akhinvasara-jumptrading:
This is very much a draft. Tests are yet unwritten. Intended to obtain feedback. 

This PR improves the ability to detect losses, particulary with few packets on a connection. In addition to the packet skip threshold, it uses a time threshold to mark packets as lost.


akhinvasara-jumptrading:
> lgtm, but we never found the source of the spamming, did we?

No, we didn't. Was planning to let these changes soak on mainnet to monitor before merging, to see if we still see that. All after merging the service queue heap to avoid hiding bugs here.

src/waltz/quic/tests/fd_quic_test_helpers.c
```diff
@@ -399,32 +399,63 @@ fd_quic_udpsock_service( fd_quic_udpsock_t const * udpsock ) {
...
                     ulong                     batch_cnt,
                     ulong *                   opt_batch_idx FD_PARAM_UNUSED,
                     int                       flush ) {
   fd_quic_netem_t * mitm_ctx = ctx;
+  ulong drop_sequence = mitm_ctx->drop_sequence;
+
+  long now  = *(mitm_ctx->now_ptr);
+       now += mitm_ctx->one_way_latency; /* updates self now_ptr */
+
+  /* Horrible hack to update dst state now, bc aio send/rcv don't support directly */
```
akhinvasara-jumptrading, line 439: @nbridge-jump @ripatel-fd any ideas that are less terrible ? Not the end of the world bc just in the test helpers, but still 🤮

ripatel-fd: Seems fine to me. It's test code. I wouldn't worry about it. If you want to get rid of the cached QUIC now timestamp, feel free to do so. Also feel free to introduce timestamps into the aio API.

akhinvasara-jumptrading: cool cool ill leave as is for now, ty!

nbridge-jump (approved):

ripatel-fd (dismissed):

PR #5639, opened 2025-07-12
pack: use stable query for noncemap

ripatel-fd:


ripatel-fd:
@mmcgee-jump @ptaffet-jump this is deployed to v0.6 (it is slower but safer), but I don't think we should have this on main.

PR #5647, opened 2025-07-14
bundle: use fd_clock, re-enable HTTP/2 keepalives

ripatel-fd:


ripatel-fd:
Replaced by https://github.com/firedancer-io/firedancer/pull/5670

PR #5656, opened 2025-07-14
ci: clean up all hugetlbfs mounts before running CI tests

topointon-jump:
Ensures all other hugetlbfs mounts are cleaned up before running CI tests.


jherrera-jump (approved):
lgtm

PR #5657, opened 2025-07-15
ci: bump hugepages reservation

topointon-jump:
We were not reserving enough hugepages for the test vectors CI step.


akhinvasara-jumptrading (approved):
LGTM (but not pow2 :()

PR #5665, opened 2025-07-15
replay: patch for replay state machine

ibhatt-jumptrading:
this pr probably doesn't solve all of the issues but mostly unblocks us as testnet is able to stay up and and replay through forks

### DIFF
--- a/src/discof/replay/fd_replay_tile.c
+++ b/src/discof/replay/fd_replay_tile.c
@@ -890,6 +890,15 @@ init_after_snapshot( fd_replay_tile_ctx_t * ctx,
     fd_fseq_update( ctx->published_wmark, root );
   }
 
+  /* Now that the snapshot(s) are done loading, we can mark all of the
+     exec tiles as ready. */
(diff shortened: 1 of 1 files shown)


lidatong (approved):

src/discof/replay/fd_replay_tile.c
```diff
@@ -1687,7 +1693,24 @@ after_credit( fd_replay_tile_ctx_t * ctx,
...
+      for( ulong i=0UL; i<ctx->exec_cnt; i++ ) {
+        if( ctx->exec_ready[ i ] == EXEC_TXN_READY ) {
+          free_cnt++;
+        }
+      }
+      if( free_cnt==ctx->exec_cnt ) {
+        break;
+      }
+      handle_writer_state_updates( ctx );
+    }
```
lidatong, line 1713: are there any vulnerabilities here where we can get hung by a malicious leader stopping midway through sending a block?

ibhatt-jumptrading: they would have to somehow send a transaction which infinitely loops which ideally can't happen in the executor

PR #5666, opened 2025-07-15
Akhin/0.6 svc q heap

akhinvasara-jumptrading:
Not a real PR, just to run CI :)


ripatel-fd:
I understand it's not a real PR. Just pointing out that I think this feature is a bit too big to backport to v0.6, would prefer putting this into a v0.7 release off main.

PR #5668, opened 2025-07-16
flamenco,vm: treat lddw as branching instruction for CU count

ravyu-jump:
Agave's JIT treats the lddw (0x18) instruction as a branching instruction, which performs a CU count update prior to executing the instruction. The linear segment is reset AFTER the instruction is executed. With how the FD VM's lddw implementation handles this reset, we don't need to account for the extra instruction in lddw. lddw is the only multiword instruction in the SBPF ISA, which eliminates the need for `ic_correction`


ripatel-fd (changes requested):
Blocking as per discussion on Slack

PR #5670, opened 2025-07-16
bundle: improve keepalive logic

ripatel-fd:
- Moves keepalive logic to a separate API
- Fixes a HTTP/2 PING flood bug
- Clears RTT Prometheus metrics when the connection breaks
- Switch bundle tile from fd_tickcount to fd_log_wallclock

### DIFF
--- a/src/disco/bundle/fd_keepalive.h
+++ b/src/disco/bundle/fd_keepalive.h
@@ -1,0 +1,94 @@
+#ifndef HEADER_fd_src_disco_bundle_fd_keepalive_h
+#define HEADER_fd_src_disco_bundle_fd_keepalive_h
+
+/* fd_keepalive.h provides an API to periodically generate keepalive
+   events.  The general usage is as follows:
+   - Configure a target keepalive interval and timeout
+   - Periodically check back whether a keepalive is due, or whether a
+     keepalive timed out
+   - Notify the object whenever a keepalive was sent or a keepalive ACK
+     was received. */
+
+#include "../../util/rng/fd_rng.h"
+
+struct fd_keepalive {
+  long ts_next_tx;   /* Timestamp when to send next ping */
+  long ts_deadline;  /* Timestamp by which ping ACK is expected to be received */
+  long interval;
+  long timeout;
+
+  long ts_last_tx;   /* Timestamp of last ping sent */
+  long ts_last_rx;   /* Timestamp of last ping ACK received */
+
+  uint inflight : 1;
+};
+
+typedef struct fd_keepalive fd_keepalive_t;
+
+FD_PROTOTYPES_BEGIN
+
+static inline long
+fd_keepalive_interval_reload( fd_rng_t * rng,
+                              long       interval ) {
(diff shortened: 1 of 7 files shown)


ripatel-fd:
@mmcgee-jump or @ptaffet-jump could either of you do a thorough review? The purpose of this change is to fix our keepalive logic. The bundle tile is supposed to send a HTTP/2 PING frame to the Jito bundle edge server every 5 seconds or so randomized. Only one PING should be inflight at a time, timeouts (5 second max) trigger a conn reset, and RTT is tracked.

I tested the following scenarios with fddev against a fake bundle server behave correctly:
- User sets keepalive_interval_millis to zero (disables keepalives)
- Multiple nominal ping round trips
- Conn resets due to bundle server closing it
- PING times out
- Prometheus metrics updated correctly

src/disco/bundle/fd_keepalive.h
```diff
@@ -0,0 +1,86 @@
...
+  uint inflight : 1;
+};
+
+typedef struct fd_keepalive fd_keepalive_t;
+
+FD_PROTOTYPES_BEGIN
+
+static inline long
+fd_keepalive_interval_reload( fd_rng_t * rng,
+                              long       interval ) {
```
ptaffet-jump, line 27: Document that `interval>1`

src/disco/bundle/fd_keepalive.h
```diff
@@ -0,0 +1,86 @@
...
+fd_keepalive_init( fd_keepalive_t * ka,
+                   fd_rng_t *       rng,
+                   long             interval,
+                   long             timeout,
+                   long             now ) {
+  memset( ka, 0, sizeof(fd_keepalive_t) );
+  if( FD_UNLIKELY( interval<2L || !timeout ) ) return NULL;
+  ka->interval = interval;
+  ka->timeout  = timeout;
+  if( FD_UNLIKELY( !ka->interval ) ) return NULL;
```
ptaffet-jump: Isn't this one dead code? `ka->interval == 0` implies `interval==0` which implies `interval<2L`

ripatel-fd: Yes that is dead code, good catch.

src/disco/bundle/fd_keepalive.h
```diff
@@ -0,0 +1,86 @@
...
+  ka->ts_next_tx = now + (long)fd_keepalive_interval_reload( rng, ka->interval );
+  return ka;
+}
+
+static inline int
+fd_keepalive_should_tx( fd_keepalive_t const * ka,
+                        long                   now ) {
+  return (!!ka->interval) & (ka->ts_next_tx <= now) & (!ka->inflight);
```
ptaffet-jump, line 58: Similarly, what is the `!!ka->interval` supposed to check?

ripatel-fd: @ptaffet-jump If keepalive is disabled, interval==0. (But the struct might still exist in a zerod out state)

PR #5673, opened 2025-07-16
Remove wildcard fixtures from CI

ripatel-fd:


jumpsiegel:
Is there a reason you are removing these tests?  Is it just the mechanism of using find instead of a fixed list?

ripatel-fd:
@jumpsiegel Yes, those vectors should be added using a text list, not a wildcard. CI was broken for many hours because of this wildcard. Someone added new vectors and then all CI checks started failing.

kbhargava-jump:
this change makes sense, but lets wait for @0x0ece to make the correct change where we have text lists of the fixture files

kbhargava-jump:
lets close for https://github.com/firedancer-io/firedancer/pull/5677 once someone looks into minimizing

PR #5676, opened 2025-07-16
sandbox: do not reset RLIMIT_CORE when dumpable

alpeng-jump:


mmcgee-jump:
Why is this needed? IIRC we haven't had issues with making core dumps without these changes?

src/app/fdctl/commands/run_agave.c
```diff
@@ -262,7 +262,7 @@ run_agave_cmd_fn( args_t *   args FD_PARAM_UNUSED,
 
   /* Also clone Agave into PID namespaces so it cannot signal
      other tile or the parent. */
-  int flags = config->development.sandbox ? CLONE_NEWPID : 0;
+  int flags = ( config->development.sandbox || !config->development.core_dump ) ? CLONE_NEWPID : 0;
```
mmcgee-jump: Do you need this? Agave will dump anyway, right?

src/app/shared/commands/run/run.c
```diff
@@ -242,7 +242,7 @@ main_pid_namespace( void * _args ) {
   fd_log_private_stack_discover( FD_TILE_PRIVATE_STACK_SZ,
                                  &fd_tile_private_stack0, &fd_tile_private_stack1 );
 
-  if( FD_UNLIKELY( !config->development.sandbox ) ) {
+  if( FD_UNLIKELY( !config->development.sandbox || config->development.core_dump ) ) {
```
mmcgee-jump: I don't think you need this, the SUBREPEAER stuff is slightly buggy (it doesn't catch all exits for some reason) and was for test/fddev only. But your pidns process is still `init` so I think it can reap as normal.

alpeng-jump:
Reverting `CLONE_NEWPID` for tiles since having them run as init is fine for core dumps - https://elixir.bootlin.com/linux/v6.11/source/kernel/signal.c#L1351. We do need to leave the rlimit though. More observations explained in slack.

src/util/sandbox/fd_sandbox.c
```diff
@@ -386,7 +386,8 @@ struct rlimit_setting {
 void
 fd_sandbox_private_set_rlimits( ulong rlimit_file_cnt,
                                 ulong rlimit_address_space,
-                                ulong rlimit_data ) {
+                                ulong rlimit_data,
+                                int dumpable ) {
```
mmcgee-jump:
```suggestion
                                int    dumpable ) {
```

src/util/sandbox/fd_sandbox.c
```diff
@@ -425,6 +426,7 @@ fd_sandbox_private_set_rlimits( ulong rlimit_file_cnt,
   };
 
   for( ulong i=0UL; i<sizeof(rlimits)/sizeof(rlimits[ 0 ]); i++ ) {
+    if( dumpable && rlimits[i].resource == RLIMIT_CORE ) continue;
```
mmcgee-jump:
```suggestion
    if( dumpable && rlimits[i].resource==RLIMIT_CORE ) continue;

```

(add newline after)

mmcgee-jump (changes requested):
just some nits, but strict about sandbox code

mmcgee-jump (approved):

PR #5677, opened 2025-07-16
ci: update CI with interp lists

kbhargava-jump:


ripatel-fd:
@0x0ece @ravyu-jump Let's run libFuzzer minimize to cut down on the size of this corpus. This is excessive

PR #5680, opened 2025-07-16
configure: allow shmem mount path to be a symlinked directory

anwayde:


src/app/shared/commands/configure/hugetlbfs.c
```diff
@@ -338,9 +332,16 @@ check( config_t const * config ) {
           PARTIALLY_CONFIGURED( "mount `%s` is on unrecognized device, expected `none`", mount_path[ i ] );
         }
 
+        /* Resolve symlinks for mount_path as the mounts table has the resolved paths */
+        char resolved_mount_path[ PATH_MAX ];
+        if( FD_UNLIKELY( !realpath( mount_path[ i ], resolved_mount_path ) ) ) {
```
mmcgee-jump, line 337: Use real syscall, e.g. `readlink` here instead of libc which might do other kinds of garbage underneath us. Also, failure to resolve is an FD_LOG_ERR, not a PARTIALLY_CONFIGURED (which is going to cause us to try blow away the directory).

topointon-jump:
Let's hold off on this for now

ripatel-fd:
A hack for a hack ... Or we just allow a user to specify a custom funk backing file?

anwayde:
Seems like we don't need/want the ability to specify symlinks for mount paths. Closing for now.

ripatel-fd:
Thanks for sending this PR, it makes sense given what we already have on main. The underlying issue is the normalpages configure stage, which is a broken design (requiring different mount paths to map a single file)

PR #5687, opened 2025-07-16
firedancer-dev: repair profiler

jvarela-jump:
Adding firedancer-dev command *repair*. This is a standalone application.
The *fd_shredcap_tile* has been modified to *publish_stake_weights* from the snapshot manifest supplied by the snapshot tile(s).
This PR also includes an upgrade to the *replay* tile's *generate_stake_weight_msg*.

### DIFF
--- a/src/discof/replay/fd_replay_tile.c
+++ b/src/discof/replay/fd_replay_tile.c
@@ -311,7 +311,7 @@ publish_stake_weights( fd_replay_tile_ctx_t * ctx,
   if( epoch_stakes_root!=NULL ) {
     ulong * stake_weights_msg = fd_chunk_to_laddr( ctx->stake_out->mem, ctx->stake_out->chunk );
     ulong epoch = fd_slot_to_leader_schedule_epoch( epoch_schedule, fd_bank_slot_get( slot_ctx->bank ) );
-    ulong stake_weights_sz = generate_stake_weight_msg( slot_ctx, ctx->runtime_spad, epoch - 1, stake_weights_msg );
+    ulong stake_weights_sz = generate_stake_weight_msg( slot_ctx, ctx->runtime_spad, epoch - 1, epoch_stakes, stake_weights_msg );
     ulong stake_weights_sig = 4UL;
     fd_stem_publish( stem, 0UL, stake_weights_sig, ctx->stake_out->chunk, stake_weights_sz, 0UL, 0UL, fd_frag_meta_ts_comp( fd_tickcount() ) );
     ctx->stake_out->chunk = fd_dcache_compact_next( ctx->stake_out->chunk, stake_weights_sz, ctx->stake_out->chunk0, ctx->stake_out->wmark );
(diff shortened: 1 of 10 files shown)


src/discof/replay/fd_replay_tile.c
```diff
@@ -318,7 +318,7 @@ publish_stake_weights( fd_replay_tile_ctx_t * ctx,
   if( epoch_stakes_root!=NULL ) {
     ulong * stake_weights_msg = fd_chunk_to_laddr( ctx->stake_out->mem, ctx->stake_out->chunk );
     ulong epoch = fd_slot_to_leader_schedule_epoch( epoch_schedule, fd_bank_slot_get( slot_ctx->bank ) );
```
ibhatt-jumptrading, line 313: I think based on fd_bank.h and comments in fd_runtime.c

Assuming we are executing in epoch E
next_epoch_stakes (E - 1)
epoch_stakes (E - 2)
stakes->vote_accounts (E)

topointon-jump: That is correct 👍

topointon-jump (commented):
Stake weight changes look good to me

src/discof/replay/fd_replay_tile.c
0x0ece: That doesn't look right to me.

To compute leader schedule in epoch E, we need the list of (identity, stake) at E-1 -- and soon (vote, stake, identity). And similarly for epoch E-1 we need the data at E-2.
We do store next_epoch_stakes(E-1) and epoch_stakes(E-2).

But using vote_accounts(E) is incorrect. We need to use the correct vote accounts at epoch boundaries, that are part of the manifest/bank.epoch_stakes (or bank.versioned_epoch_stakes -- I've been told these are the same but I haven't checked yet. Agave seems to be switching to versioned in... 2.4? not sure about this either).

topointon-jump: We should use `epoch_stakes` (E-2), which is what I think `stake_weights` is supposed to represent in this message.

lidatong: maybe we can just rename this stuff to be the same and add copious documentation in this PR?

topointon-jump: There is some documentation at the top of `fd_runtime_process_new_epoch`:
```
/* Starting a new epoch.
  New epoch:        T
  Just ended epoch: T-1
  Epoch before:     T-2

  In this function:
  - stakes in T-2 (epoch_stakes) should be replaced by T-1 (next_epoch_stakes)
  - stakes at T-1 (next_epoch_stakes) should be replaced by updated stakes at T (stakes->vote_accounts)
  - leader schedule should be calculated using new T-2 stakes (epoch_stakes)

  Invariant during an epoch T:
  next_epoch_stakes holds the stakes at T-1
  epoch_stakes holds the stakes at T-2
 */
```

Agree we should make this clearer

src/discof/shredcap/fd_shredcap_tile.c
```diff
@@ -51,6 +57,19 @@ typedef union {
...
+struct fd_stake_out_link {
+  ulong       idx;
+  fd_frag_meta_t * mcache;
+  ulong *          sync;
+  ulong            depth;
+  ulong            seq;
+  fd_wksp_t * mem;
+  ulong       chunk0;
+  ulong       wmark;
+  ulong       chunk;
```
lidatong: align

src/discof/shredcap/fd_shredcap_tile.c
```diff
@@ -105,7 +137,28 @@ scratch_align( void ) {
...
+}
+
+FD_FN_CONST static inline ulong
+manifest_spad_align( void ) {
+  return 128UL;
+}
+
+FD_FN_CONST static inline ulong
+manifest_spad_footprint( void ) {
+  return 1UL*1024UL*1024UL*1024UL;
```
lidatong, line 161: these magic values seem a little strange. are they just copied over from the snapshot tile?

src/discof/shredcap/fd_shredcap_tile.c
```diff
@@ -141,6 +261,9 @@ before_frag( fd_capture_tile_ctx_t * ctx,
   if( FD_LIKELY( ctx->in_kind[ in_idx ]==NET_SHRED ) ) {
     return (fd_disco_netmux_sig_proto( sig )!=DST_PROTO_SHRED) & (fd_disco_netmux_sig_proto( sig )!=DST_PROTO_REPAIR);
   }
+  if( FD_LIKELY( ctx->in_kind[ in_idx ]==SNAP_IN ) ) {
+    return 0;
+  }
```
lidatong: doesnt seem needed given `return 0` below?

lidatong (dismissed):
nice! this will be super useful for us

jvarela-jump:
This PR has been superseded by https://github.com/firedancer-io/firedancer/pull/5807.
The version that uses the snapshot tile may be considered in the future as a _startup_ profiler.

PR #5693, opened 2025-07-16
offline-replay: update snapshot params

kbhargava-jump:


contrib/offline-replay/run_offline_replay_backtest.sh
```diff
@@ -198,6 +198,8 @@ while true; do
 
             rm -rf $TEMP_LOG && touch $TEMP_LOG && chmod 777 $TEMP_LOG
 
+            chmod -R 0700 $LEDGER_DIR
```
cali-jumptrading, line 201: technically you don't need this anymore since there isn't a fini stage anymore.

kbhargava-jump: you do, fini wont happen but you will fail on node start up with permission error

contrib/offline-replay/offline_replay.toml
```diff
@@ -43,3 +42,8 @@
     vote_account = "{vote_account_path}"
 [hugetlbfs]
     mount_path = "/data/firedancer/mnt"
+[snapshots]
```
cali-jumptrading, line 46: not sure if you need to specify snapshots path too via
```
[paths]
    snapshots = "<path>"
```

cali-jumptrading (approved):

contrib/offline-replay/offline_replay.toml
kbhargava-jump: oops thanks

cali-jumptrading (approved):

PR #5694, opened 2025-07-16
configure, ci: don't allow backtest in CI to raise reserved pages

topointon-jump:
Raising the number of reserved hugepages should only be done explicitly by the `hugepages` GitHub action.


anwayde (dismissed):

src/app/firedancer/config/default.toml
```diff
@@ -718,6 +718,20 @@ user = ""
...
+    # and hugepages-1048576kB/nr_hugepages sysfs values for
+    # each NUMA node.
+    #
+    # If this is true, the hugetlbfs configure stage
+    # will raise the number of reserved huge and gigantic pages
+    # to the amount required to run Firedancer.
+    #
+    # If this is false, then the hugetlbfs configure stage will
+    # fail if there are not enough pages reserved.
+    raise_reservations_allowed = true
```
mmcgee-jump: I'd call this `allow_hugepage_increase` and something like

```
    # Whether the hugetlbfs configure stage is allowed to increase the
    # number of huge or gigantic pages available on the system.  If not
    # allowed, and the system does not have enough huge or gigantic
    # pages available to run the application, configuration will fail
    # with an error.  If allowed, configuration will attempt to reserve
    # enough pages from the kernel that running will always succeed.
```

mmcgee-jump (changes requested):
.

PR #5696, opened 2025-07-16
snapshots: fix archive

cali-jumptrading:


src/discof/restore/utils/fd_snapshot_archive.h
```diff
@@ -0,0 +1,56 @@
+#ifndef HEADER_fd_src_discof_restore_utils_fd_snapshot_archive_h
```
mmcgee-jump, line 1: Really prefer the old filename, more consistent with fd style. E.g. see `fd_tpool`, not `fd_thread_pool`, and many other cases.

ripatel-fd: Keystrokes are cheap. Nobody but the original maintainers would know what an ssarchive is looking at a directory listing

PR #5704, opened 2025-07-17
mcache: reduce minimum depth to 1

mmcgee-jump:


ripatel-fd:
That said, if this gets merged, I would just remove the interleave feature entirely

PR #5710, opened 2025-07-17
gui: don't count invalid fee payer txns

jherrera-jump:


jherrera-jump:
There are two reasons a scheduled transaction might not end up paying fees

1. It doesn't land in the block
2. it lands in the block, but contains an invalid fee payer

this pr essentially zeros fees for these transactions in the slot progression chart.

@mmcgee-jump Do we want to zero fees for these transactions in the bank/bars chart? This would change the behavior of the "income per CU" toggle as well as what's show in the tooltip.

My view is that we should leave these fees in the banks/bars chart and instead add another indicator in the tooltip that a transaction did not pay fees. I think the bars chart should reflect the scheduler's view of the world and the scheduler doesn't actuallly know if a transaction pays fees before its executed.

PR #5714, opened 2025-07-17
snapshots: file write

cali-jumptrading:


src/discof/restore/fd_snaprd_tile.c
```diff
@@ -433,8 +489,17 @@ privileged_init( fd_topo_t *      topo,
                                                  &incremental_slot,
                                                  full_path,
                                                  incremental_path ) ) ) {
-    ctx->local.full_snapshot_slot = ULONG_MAX;
+    ctx->local.full_snapshot_slot        = ULONG_MAX;
     ctx->local.incremental_snapshot_slot = ULONG_MAX;
+
+    fd_cstr_printf_check( ctx->local.full_snapshot_path, PATH_MAX, NULL, "%s/snapshot.tar.bz2", tile->snaprd.snapshots_path );
```
mmcgee-jump:
```suggestion
    FD_TEST( fd_cstr_printf_check( ctx->local.full_snapshot_path, PATH_MAX, NULL, "%s/snapshot.tar.bz2", tile->snaprd.snapshots_path ) );
```

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -221,6 +259,24 @@ read_http_data( fd_snaprd_tile_t *  ctx,
 
       fd_stem_publish( stem, 0UL, FD_SNAPSHOT_MSG_DATA, ctx->out.chunk, data_len, 0UL, 0UL, 0UL );
       ctx->out.chunk = fd_dcache_compact_next( ctx->out.chunk, data_len, ctx->out.chunk0, ctx->out.wmark );
+
+      int full = ctx->state==FD_SNAPRD_STATE_READING_FULL_HTTP;
+      int fd = full ? ctx->local.full_snapshot_fd : ctx->local.incremental_snapshot_fd;
```
mmcgee-jump: Isn't this the snapshot we are reading from? I think you need two new file descriptors for write operations?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -54,9 +57,14 @@ struct fd_snaprd_tile {
     ulong incremental_snapshot_slot;
     int   incremental_snapshot_fd;
     char  incremental_snapshot_path[ PATH_MAX ];
+    uchar file_buffer[ SNAPRD_FILE_BUF_SZ ];
+    ulong file_buffer_pos;
+    ulong file_buffer_sz;
+    ulong file_write_pos;
   } local;
```
mmcgee-jump: Maybe rename this one `local_in` and stick your new buffer, and 2x file descriptors in `local_out` ?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -167,13 +175,32 @@ read_file_data( fd_snaprd_tile_t *  ctx,
...
+            ulong sz,
+            ulong * written_sz ) {
+  int err = fd_io_write( fd,
+                         data,
+                         sz,
+                         sz,
+                         written_sz );
+  if( FD_UNLIKELY( err ) && err!=EAGAIN ) {
+    FD_LOG_ERR(( "fd_io_write() failed (%d-%s) requested %lu bytes and wrote %lu bytes", err, fd_io_strerror( err ), sz, *written_sz ));
+    return -1;
```
mmcgee-jump: No need to return FD_LOG_ERR will exit

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -167,13 +175,32 @@ read_file_data( fd_snaprd_tile_t *  ctx,
   ctx->out.chunk = fd_dcache_compact_next( ctx->out.chunk, (ulong)result, ctx->out.chunk0, ctx->out.wmark );
 }
 
+static int
+write_file( int fd,
```
mmcgee-jump: align args

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -204,6 +232,16 @@ read_http_data( fd_snaprd_tile_t *  ctx,
         default:
           break;
       }
+
+      int full = ctx->state==FD_SNAPRD_STATE_READING_FULL_HTTP;
+      char const * old_path = full ? ctx->local.full_snapshot_path : ctx->local.incremental_snapshot_path;
+      char rename_path[ PATH_MAX ];
+      fd_cstr_printf_check( rename_path, PATH_MAX, NULL, "%s/%s", ctx->config.path, fd_sshttp_snapshot_name( ctx->sshttp ) );
```
mmcgee-jump:
```suggestion
      FD_TEST( fd_cstr_printf_check( rename_path, PATH_MAX, NULL, "%s/%s", ctx->config.path, fd_sshttp_snapshot_name( ctx->sshttp ) ) );
```

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -204,6 +232,16 @@ read_http_data( fd_snaprd_tile_t *  ctx,
...
+      int err = rename( old_path, rename_path );
```
mmcgee-jump: I would use `renameat` here at pre-open in the directory file descriptor in privileged init, that way the sandbox can allow only rename within this snapshot directory.

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -33,6 +33,8 @@ struct fd_sshttp_private {
   ulong response_len;
   char  response[ USHORT_MAX ];
 
+  char snapshot_name[ PATH_MAX ];
```
mmcgee-jump: I wouldn't reuse the name as-is, but parse it into slot/incremental_slot/hash, and pass those back instead

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -188,11 +205,22 @@ read_http_data( fd_snaprd_tile_t *  ctx,
       break;
     }
     case FD_SSHTTP_ADVANCE_DONE: {
+      int full = ctx->state==FD_SNAPRD_STATE_READING_FULL_HTTP;
+      char const * old_path = full ? ctx->local_out.full_snapshot_path : ctx->local_out.incremental_snapshot_path;
+      char rename_path[ PATH_MAX ];
+      fd_cstr_printf_check( rename_path, PATH_MAX, NULL, "%s/%s", ctx->config.path, fd_sshttp_snapshot_name( ctx->sshttp ) );
```
mmcgee-jump:
```suggestion
      FD_TEST( fd_cstr_printf_check( rename_path, PATH_MAX, NULL, "%s/%s", ctx->config.path, fd_sshttp_snapshot_name( ctx->sshttp ) ) );
```

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -36,6 +37,8 @@
 #define FD_SNAPRD_STATE_FLUSHING_INCREMENTAL_HTTP (11) /* Incremental snapshot was downloaded ok, confirm it decompressed and inserted ok */
 #define FD_SNAPRD_STATE_SHUTDOWN                  (12) /* The tile is done, and has likely already exited */
 
+#define SNAPRD_FILE_BUF_SZ 1024 * 1024UL /* 1 mib */
```
mmcgee-jump: I'd make this bigger, at least the same size as the HTTP recv() buffer size

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -312,7 +367,7 @@ after_credit( fd_snaprd_tile_t *  ctx,
         break;
       }
 
-      FD_LOG_NOTICE(( "reading incremental snapshot from local file `%s`", ctx->local.incremental_snapshot_path ));
+      FD_LOG_NOTICE(( "Reading incremental snapshot from local file `%s`", ctx->local_in.incremental_snapshot_path ));
```
mmcgee-jump: I was just in the process of switching all the boot messages to lowercase, lol, to align with fd_util and tango

cali-jumptrading: I just didn't catch this when rebasing, thanks for catching.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: Not sure I understand, which HTTP recv buffer size?

src/discof/restore/utils/fd_sshttp.c
cali-jumptrading: I would prefer to defer this to later.

mmcgee-jump: This is a security issue I don't think you should defer it. You should not write files to the system from arbitrary untrusted input. For example, the server could set the path to `../../../../../id_rsa` and give themselves SSH access to your box.

src/discof/restore/fd_snaprd_tile.c
ripatel-fd: Nah, this size is good for file I/O. If you go significantly beyond 1MB (exceeding half of L3 cache), you're just going to decrease file writing performance.

mmcgee-jump: That should be controlled at the `write(})` gruanularity level? This is determining how much can buffer from net before backpressuring, 1MB is low

ripatel-fd: @mmcgee-jump @cali-jumptrading Sorry to unresolve this but to clarify, what counts is the **virtual address range** that you pass into write(), not the size argument of write. If you have a 100MB buffer and you do 1MB chunk writes, that is way worse than calling write() on the same 1MB uchar array

mmcgee-jump: I'm dumb, why does that matter at all?

ripatel-fd: @mmcgee-jump Because when you buffer things in a 100MB array (larger than L3 cache), you are competing on DRAM bandwidth with the write() syscall when filling the local buffer array. And often, the write() syscall doesn't actually flush to disk, it just buffers in dirty cache. So, you basically cut the performance of short write bursts to dirty cache in half by having a large input buffer. We tried really hard to keep the local context of every snapshot loader tile below 6 megabyte in the "optimize snapshot performance" patch set, and that yielded significant throughput improvements.

ripatel-fd: The mental model I had is "how many round trips in and out of the L3 cache does it take to process 1 account on average". If you have a large local buffer, it is automatically +1 because you spill that local buffer L3->DRAM on local buffer write, and then you fill again DRAM->L3 on write().

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -202,6 +206,34 @@ follow_redirect( fd_sshttp_t *        http,
 
       location_len = headers[ i ].value_len;
       location = headers[ i ].value;
+
+      if( location_len>=PATH_MAX ) {
```
mmcgee-jump:
```suggestion
      if( FD_UNLIKELY( location_len>=PATH_MAX ) ) {
```

mmcgee-jump: `location_len` here doesn't include terminating NUL, so you want `>=PATH_MAX-1UL` I believe

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -202,6 +206,34 @@ follow_redirect( fd_sshttp_t *        http,
...
+        fd_sshttp_cancel( http );
+        return FD_SSHTTP_ADVANCE_ERROR;
+      }
+
+      char snapshot_name[ PATH_MAX ];
+      fd_memcpy( snapshot_name, location+1, location_len-1 );
+      snapshot_name[ location_len-1 ] = '\0';
```
mmcgee-jump: Per above, location_len doesn't include the NUL
```suggestion
      snapshot_name[ location_len ] = '\0';
```

src/discof/restore/utils/fd_sshttp.c
mmcgee-jump: Oh I see, you exclude the initial `/`

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -202,6 +206,34 @@ follow_redirect( fd_sshttp_t *        http,
...
       location_len = headers[ i ].value_len;
       location = headers[ i ].value;
+
+      if( location_len>=PATH_MAX ) {
+        fd_sshttp_cancel( http );
+        return FD_SSHTTP_ADVANCE_ERROR;
+      }
+
+      char snapshot_name[ PATH_MAX ];
+      fd_memcpy( snapshot_name, location+1, location_len-1 );
```
mmcgee-jump: Per above,
```suggestion
      fd_memcpy( snapshot_name, location+1UL, location_len-1UL );
```

mmcgee-jump: Does this handle `location_len==0UL` case? I think you need to cancel/error then as well

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -71,6 +55,17 @@ parse_filename( char *  _name,
   next = strchr( ptr, '.' );
   if( FD_UNLIKELY( !next ) ) return -1;
 
+  if( get_hash ) {
```
mmcgee-jump: Any particular reason for this arg? Might be cleaner to just always return hash

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -71,6 +55,17 @@ parse_filename( char *  _name,
...
+    ulong sz = (ulong)(next - ptr);
+
+    if( sz!=FD_BASE58_ENCODED_32_SZ-1 ) return -1;
+
+    char encoded_hash[ FD_BASE58_ENCODED_32_SZ ];
+    fd_memcpy( encoded_hash, ptr, FD_BASE58_ENCODED_32_SZ-1 );
+    encoded_hash[ FD_BASE58_ENCODED_32_SZ-1 ] = '\0';
+    fd_base58_decode_32( encoded_hash, hash );
```
mmcgee-jump: This can fail, and return error

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -380,13 +435,13 @@ after_frag( fd_snaprd_tile_t *  ctx,
       case FD_SNAPRD_STATE_READING_FULL_FILE:
       case FD_SNAPRD_STATE_FLUSHING_FULL_FILE:
       case FD_SNAPRD_STATE_FLUSHING_FULL_FILE_RESET:
-        FD_LOG_ERR(( "Error reading snapshot from local file `%s`", ctx->local.full_snapshot_path ));
+        FD_LOG_ERR(( "Error reading snapshot from local file `%s`", ctx->local_in.full_snapshot_path ));
```
mmcgee-jump: Fix up these case changes :)

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -221,6 +250,32 @@ read_http_data( fd_snaprd_tile_t *  ctx,
...
+                             write_sz,
+                             &written_sz );
+      if( FD_UNLIKELY( err ) && err!=EAGAIN ) {
+        FD_LOG_ERR(( "fd_io_write() failed (%d-%s) requested %lu bytes and wrote %lu bytes", err, fd_io_strerror( err ), write_sz, written_sz ));
+      }
+
+      ctx->local_out.file_write_pos += written_sz;
+      if( ctx->local_out.file_write_pos == ctx->local_out.file_buffer_pos &&
+          ctx->local_out.file_buffer_pos == ctx->local_out.file_buffer_sz ) {
+        ctx->local_out.file_buffer_pos = 0UL;
```
mmcgee-jump: Not a good spot for this, I'd move it to where the state machine transitions to "READING" or whatever

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -221,6 +250,32 @@ read_http_data( fd_snaprd_tile_t *  ctx,
...
+
+      ulong write_sz = ctx->local_out.file_buffer_pos - ctx->local_out.file_write_pos;
+      ulong written_sz = 0UL;
+      int err = fd_io_write( fd,
+                             ctx->local_out.file_buffer + ctx->local_out.file_write_pos,
+                             write_sz,
+                             write_sz,
+                             &written_sz );
+      if( FD_UNLIKELY( err ) && err!=EAGAIN ) {
+        FD_LOG_ERR(( "fd_io_write() failed (%d-%s) requested %lu bytes and wrote %lu bytes", err, fd_io_strerror( err ), write_sz, written_sz ));
```
mmcgee-jump: This is going to be a pretty common user facing error due `ENOSPC`. I'd write a nice long comment for that case explaining the problem.

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -173,26 +189,38 @@ read_http_data( fd_snaprd_tile_t *  ctx,
...
       break;
     }
     case FD_SSHTTP_ADVANCE_DONE: {
+      int full = ctx->state==FD_SNAPRD_STATE_READING_FULL_HTTP;
+      char const * old_path = full ? ctx->local_out.full_snapshot_path : ctx->local_out.incremental_snapshot_path;
+      char rename_path[ PATH_MAX ];
+      fd_cstr_printf_check( rename_path, PATH_MAX, NULL, "%s/%s", ctx->config.path, fd_sshttp_snapshot_name( ctx->sshttp ) );
+      int err = renameat( ctx->local_out.dir_fd, old_path, ctx->local_out.dir_fd, rename_path );
+      if( FD_UNLIKELY( err ) ) {
+        FD_LOG_WARNING(( "Failed to rename incremental snapshot file from `%s` to `%s`: %d-%s",
```
mmcgee-jump:
```suggestion
        FD_LOG_WARNING(( "Failed to rename incremental snapshot file from `%s` to `%s` (%d-%s)",
```

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -173,26 +189,38 @@ read_http_data( fd_snaprd_tile_t *  ctx,
...
       fd_stem_publish( stem, 0UL, FD_SNAPSHOT_MSG_CTRL_RESET_FULL, 0UL, 0UL, 0UL, 0UL, 0UL );
       ctx->state = FD_SNAPRD_STATE_FLUSHING_FULL_HTTP_RESET;
       ctx->deadline_nanos = now;
       break;
     }
     case FD_SSHTTP_ADVANCE_DONE: {
+      int full = ctx->state==FD_SNAPRD_STATE_READING_FULL_HTTP;
+      char const * old_path = full ? ctx->local_out.full_snapshot_path : ctx->local_out.incremental_snapshot_path;
+      char rename_path[ PATH_MAX ];
+      fd_cstr_printf_check( rename_path, PATH_MAX, NULL, "%s/%s", ctx->config.path, fd_sshttp_snapshot_name( ctx->sshttp ) );
```
mmcgee-jump:
```suggestion
      FD_TEST( fd_cstr_printf_check( rename_path, PATH_MAX, NULL, "%s/%s", ctx->config.path, fd_sshttp_snapshot_name( ctx->sshttp ) ) );
```

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -47,16 +50,29 @@ struct fd_snaprd_tile {
...
     int   full_snapshot_fd;
     char  full_snapshot_path[ PATH_MAX ];
     ulong incremental_snapshot_slot;
     int   incremental_snapshot_fd;
     char  incremental_snapshot_path[ PATH_MAX ];
-  } local;
+  } local_in;
 
   struct {
+    char path[ PATH_MAX ];
```
mmcgee-jump, line 77: Is this used?

src/discof/restore/fd_snaprd_tile.c
mmcgee-jump:
```suggestion
#define SNAPRD_FILE_BUF_SZ (1024UL*1024UL) /* 1 MiB */
```

src/discof/restore/utils/fd_sshttp.c
cali-jumptrading: yes it is handled in the lines of code above.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: yes, it is used to create the new path when renaming the downloaded snapshot file

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -201,7 +205,35 @@ follow_redirect( fd_sshttp_t *        http,
...
+      location     = headers[ i ].value;
+
+      if( FD_UNLIKELY( location_len>=PATH_MAX-1 ) ) {
+        fd_sshttp_cancel( http );
+        return FD_SSHTTP_ADVANCE_ERROR;
+      }
+
+      char snapshot_name[ PATH_MAX ];
+      fd_memcpy( snapshot_name, location+1UL, location_len-1UL );
+      snapshot_name[ location_len-1 ] = '\0';
```
mmcgee-jump:
```suggestion
      snapshot_name[ location_len-1UL ] = '\0';
```

Sorry for the nits, hopefully one time and then can be consistent from here :)

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: Can you explain? why is this not a good spot?

mmcgee-jump: `file_buffer_pos = 0` is an action associated with starting a new write, but here you set it when ending a prior write.  If you modeled as a state machine, you would zero on transition into starting a snapshot, not on completing one. It works here because of a fragile assumption that completing a snapshot precedes starting a new one (and you need additionally to zero them in unprivileged init).

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -201,7 +205,43 @@ follow_redirect( fd_sshttp_t *        http,
...
+
+      if( FD_UNLIKELY( err ) ) {
+        FD_LOG_WARNING(( "unrecognized snapshot file `%s` in redirect location header", snapshot_name ));
+        fd_sshttp_cancel( http );
+        return FD_SSHTTP_ADVANCE_ERROR;
+      }
+
+      char encoded_hash[ FD_BASE58_ENCODED_32_SZ ];
+      char * result = fd_base58_encode_32( decoded_hash, NULL, encoded_hash );
+      if( FD_UNLIKELY( !result ) ) {
```
mmcgee-jump: base58 encoding cannot fail, you don't need to check this

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -201,7 +205,43 @@ follow_redirect( fd_sshttp_t *        http,
...
+        /* This shouldn't fail because fd_ssarchive_parse_filename
+           decodes the hash. */
+        FD_LOG_WARNING(( "failed to encode hash in redirect location header" ));
+        fd_sshttp_cancel( http );
+        return FD_SSHTTP_ADVANCE_ERROR;
+      }
+
+
+      if( incremental_entry_slot!=ULONG_MAX ) {
```
mmcgee-jump:
```suggestion
      if( FD_LIKELY( incremental_entry_slot!=ULONG_MAX ) ) {
```

style consistency

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -201,7 +205,43 @@ follow_redirect( fd_sshttp_t *        http,
       }
 
       location_len = headers[ i ].value_len;
-      location = headers[ i ].value;
+      location     = headers[ i ].value;
+
+      if( FD_UNLIKELY( location_len>=PATH_MAX-1 ) ) {
```
mmcgee-jump:
```suggestion
      if( FD_UNLIKELY( location_len>=PATH_MAX-1UL ) ) {
```

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -71,6 +55,19 @@ parse_filename( char *  _name,
   next = strchr( ptr, '.' );
   if( FD_UNLIKELY( !next ) ) return -1;
 
+  ulong sz = (ulong)(next - ptr);
+
+  if ( get_hash ) {
```
mmcgee-jump: Still not sure why there's this `get_hash` arg. Can't you just always return the hash?

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -71,6 +55,19 @@ parse_filename( char *  _name,
...
+    if( sz<32UL || sz>FD_BASE58_ENCODED_32_SZ-1UL ) return -1;
```
mmcgee-jump:
```suggestion
    if( FD_UNLIKELY( sz>FD_BASE58_ENCODED_32_LEN ) ) return -1;
```

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -71,6 +55,19 @@ parse_filename( char *  _name,
...
+
+    char encoded_hash[ FD_BASE58_ENCODED_32_SZ ];
+    fd_memcpy( encoded_hash, ptr, FD_BASE58_ENCODED_32_SZ-1 );
```
mmcgee-jump:
```suggestion
    fd_memcpy( encoded_hash, ptr, sz-1UL );
```

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -71,6 +55,19 @@ parse_filename( char *  _name,
...
+    encoded_hash[ FD_BASE58_ENCODED_32_SZ-1 ] = '\0';
```
mmcgee-jump:
```suggestion
    encoded_hash[ sz ] = '\0';
```

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -173,26 +191,44 @@ read_http_data( fd_snaprd_tile_t *  ctx,
...
   uchar * out = fd_chunk_to_laddr( ctx->out.wksp, ctx->out.chunk );
 
-  ulong data_len = ctx->out.mtu;
+  if( ctx->local_out.file_write_pos==ctx->local_out.file_buffer_pos &&
+    ctx->local_out.file_buffer_pos==ctx->local_out.file_buffer_sz ) {
+    ctx->local_out.file_buffer_pos = 0UL;
+    ctx->local_out.file_write_pos  = 0UL;
+  }
+
+  ulong avail_buffer_sz = ctx->local_out.file_buffer_sz - ctx->local_out.file_buffer_pos;
```
mmcgee-jump:
```suggestion
  ulong avail_buffer_sz = ctx->local_out.file_buffer_sz-ctx->local_out.file_buffer_pos;
```

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -204,23 +240,51 @@ read_http_data( fd_snaprd_tile_t *  ctx,
...
       }
+
       break;
     }
     case FD_SSHTTP_ADVANCE_DATA: {
+      fd_stem_publish( stem, 0UL, FD_SNAPSHOT_MSG_DATA, ctx->out.chunk, data_len, 0UL, 0UL, 0UL );
+      ctx->out.chunk = fd_dcache_compact_next( ctx->out.chunk, data_len, ctx->out.chunk0, ctx->out.wmark );
+
+      FD_TEST( ctx->local_out.file_buffer_pos + data_len <= ctx->local_out.file_buffer_sz );
+      fd_memcpy( ctx->local_out.file_buffer + ctx->local_out.file_buffer_pos, out, data_len );
```
mmcgee-jump: No need to memcpy here in the typical case. You only want to memcpy the bytes left over if the `io_write` would block.

src/discof/restore/utils/fd_ssarchive.c
cali-jumptrading: removed.

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -3,6 +3,8 @@
 
 #include "../../../waltz/http/picohttpparser.h"
 #include "../../../util/log/fd_log.h"
+#include "../../../flamenco/types/fd_types_custom.h"
+#include "fd_ssarchive.h"
```
mmcgee-jump: Move this up, to be low `fd_sshttp.h`

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -433,21 +497,33 @@ privileged_init( fd_topo_t *      topo,
...
-    ctx->local.full_snapshot_slot = ULONG_MAX;
-    ctx->local.incremental_snapshot_slot = ULONG_MAX;
+    ctx->local_in.full_snapshot_slot        = ULONG_MAX;
+    ctx->local_in.incremental_snapshot_slot = ULONG_MAX;
+
+    ctx->local_out.dir_fd = open( tile->snaprd.snapshots_path, O_DIRECTORY|O_CLOEXEC );
+    if( FD_UNLIKELY( -1==ctx->local_out.dir_fd ) ) FD_LOG_ERR(( "open() failed `%s` (%i-%s)", tile->snaprd.snapshots_path, errno, fd_io_strerror( errno ) ));
+
+    FD_TEST( fd_cstr_printf_check( ctx->local_out.full_snapshot_path, PATH_MAX, NULL, "%s/snapshot.tar.bz2", tile->snaprd.snapshots_path ) );
+    ctx->local_out.full_snapshot_fd = open( ctx->local_out.full_snapshot_path, O_WRONLY|O_CREAT|O_TRUNC|O_NONBLOCK, S_IRUSR|S_IWUSR );
```
mmcgee-jump: Use `openat` on your dir_fd

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: I'd rather not optimize in this PR. Can we leave it how it is and optimize later?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -433,21 +497,33 @@ privileged_init( fd_topo_t *      topo,
...
                                                  incremental_path ) ) ) {
-    ctx->local.full_snapshot_slot = ULONG_MAX;
-    ctx->local.incremental_snapshot_slot = ULONG_MAX;
+    ctx->local_in.full_snapshot_slot        = ULONG_MAX;
+    ctx->local_in.incremental_snapshot_slot = ULONG_MAX;
+
+    ctx->local_out.dir_fd = open( tile->snaprd.snapshots_path, O_DIRECTORY|O_CLOEXEC );
+    if( FD_UNLIKELY( -1==ctx->local_out.dir_fd ) ) FD_LOG_ERR(( "open() failed `%s` (%i-%s)", tile->snaprd.snapshots_path, errno, fd_io_strerror( errno ) ));
+
+    FD_TEST( fd_cstr_printf_check( ctx->local_out.full_snapshot_path, PATH_MAX, NULL, "%s/snapshot.tar.bz2", tile->snaprd.snapshots_path ) );
```
mmcgee-jump: The in progress one shouldn't be named like this. You want to rename it later. I'd call it `full.tar.bz2.partial` or something.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: why do you prefer `openat`?
```
openat() opens a file by using the specified directory file descriptor as the starting location for the path search. It operates in the same way as open() except for the following differences.
If pathname is absolute, dirfd is ignored.
```

cali-jumptrading: I don't see the difference between `open` and `openat`, unless I add in extra logic to parse out the relative path.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: I'd rather call it something like `snapshot.tar.bz2-partial` or something.

mmcgee-jump: Sure

src/discof/restore/fd_snaprd_tile.c
mmcgee-jump: There's no TOCTOU condition, where the directory gets renamed after you open it

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -205,23 +241,51 @@ read_http_data( fd_snaprd_tile_t *  ctx,
...
+      ulong written_sz = 0UL;
+      int err = fd_io_write( fd,
+                             ctx->local_out.file_buffer+ctx->local_out.file_write_pos,
+                             write_sz,
+                             write_sz,
+                             &written_sz );
+      if( FD_UNLIKELY( err ) && err!=EAGAIN ) {
+        /* Reaching this error likely means there is no disk space
+           left. */
+        FD_LOG_ERR(( "Out of disk space when writing %lu bytes to %s with size %lu", write_sz, snapshot_path, bytes_written ));
```
mmcgee-jump: Only if `errno==ENOSPC`, otherwise just do the regular errno print

mmcgee-jump (approved):

PR #5747, opened 2025-07-19
fdctl: add get-identity command to retrieve runtime validator identity

huiskylabs:
## Add `fdctl get-identity` command

  ### Problem
  - `fdctl set-identity` can dynamically change validator identity at runtime
  - No way to query the actual running identity after a switch
  - Config files become outdated after identity changes
  - GUI tile is optional and unreliable for programmatic access
  - Tools like [SVS](https://github.com/huiskylabs/solana-validator-switch) currently rely on `solana catchup` to infer the active identity, which is indirect, slow and unreliable.

  ### Solution
  Add `fdctl get-identity` to retrieve the current runtime identity from the shred
  tile.

  ### Usage
  ```bash
  $ fdctl get-identity
  7Np41oeYqPefeNQEHSv1UDhYrehxin3NStELsSKCT4K2
```
  Use Cases

  - Pre-switch validation: Verify current identity before switching
  - Post-switch verification: Confirm new identity is active
  - Monitoring tools: Track identity changes programmatically
  - Automation: Enable scripts to handle identity rotation safely

Demonstration:

```
//current ID
solana@hui-9174F:~$ fdctl --config firedancer-config.toml get-identity
Log at "/mnt/solana/log/solana-validator.log"
8CzCcNCwg8nx3C4LfiUzanwok4EXoJUFBKQeBfHzoJza

//set-identity
solana@hui-9174F:~$ fdctl set-identity --config firedancer-config.toml /home/solana/funded-validator-keypair.json 
Log at "/mnt/solana/log/solana-validator.log"
NOTICE  07-19 02:01:37.160898 11761  f0   
main src/app/shared/commands/set_identity.c(378): 
Validator identity key switched to `Txh6mKsvA7rwesigHzALTiFggyrkGDVtMP1vicuepAs`

//get-identity reflect new id
solana@hui-9174F:~$ fdctl --config firedancer-config.toml get-identity
Log at "/mnt/solana/log/solana-validator.log"
Txh6mKsvA7rwesigHzALTiFggyrkGDVtMP1vicuepAs
```

### CODE REVIEW

**Comment by mmcgee-jump** [comment]
Wow nice, thanks for the contribution! Very helpful

File: src/app/shared/commands/get_identity.c [hunk 1]
```diff
@@ -0,0 +1,71 @@
+#include "../fd_config.h"
+#include "../fd_action.h"
+#include "../../../util/fd_util.h"
+#include "../../../ballet/base58/fd_base58.h"
+#include "../../../flamenco/types/fd_types_custom.h"
+#include <unistd.h>
+
+/* Minimal shred tile context structure to access identity_key.
```
mmcgee-jump: Not a huge fan of the redefinition, since we will forget to update this if shred tile changes. Are you able to move the shred_ctx to a header `fd_shred_tile.h` ? and just include it?

src/app/shared/commands/get_identity.c
```diff
@@ -0,0 +1,71 @@
...
+
+  /* Cast to our minimal context structure */
+  fd_shred_ctx_minimal_t const * shred_ctx = (fd_shred_ctx_minimal_t const *)shred_mem;
+  
+  /* The shred tile maintains the current identity in shred_ctx->identity_key */
+  fd_pubkey_t const * identity_key = shred_ctx->identity_key;
+
+  /* Convert to base58 and print */
+  char identity_key_str[ FD_BASE58_ENCODED_32_SZ ];
+  fd_base58_encode_32( identity_key->uc, NULL, identity_key_str );
```
mmcgee-jump: There's a very subtle edge case bug here but this can return a torn/corrupt identity key in case the shred tile is currently processing a `set-identity` command and has only overwritten half of the bytes so far. Not sure you need to fix because it will be exceedingly rare, but can you add a comment and `/* TODO ... ` about this so we remember to tidy it up in future.

src/app/shared/commands/get_identity.c
```diff
@@ -0,0 +1,71 @@
...
+  
+  FD_LOG_STDOUT(( "%s\n", identity_key_str ));
+
+  /* Leave the workspace */
```
mmcgee-jump: Don't need to cleanup as you're about to exit

src/app/shared/commands/get_identity.c
```diff
@@ -0,0 +1,71 @@
...
+  ulong shred_wksp_id = shred_obj->wksp_id;
+
+  /* Join the workspace in read-only mode */
+  fd_topo_join_workspace( &config->topo, &config->topo.workspaces[ shred_wksp_id ], FD_SHMEM_JOIN_MODE_READ_ONLY );
+  fd_topo_workspace_fill( &config->topo, &config->topo.workspaces[ shred_wksp_id ] );
+
+  /* Access the shred context through the workspace */
+  void * shred_mem = fd_topo_obj_laddr( &config->topo, shred_tile->tile_obj_id );
+  if( FD_UNLIKELY( !shred_mem ) ) {
+    fd_topo_leave_workspaces( &config->topo );
```
mmcgee-jump, line 35: No need to cleanup, FD_LOG_ERR exits the process and we prefer the kernel to cleanup instead

src/app/shared/commands/get_identity.c
```diff
@@ -0,0 +1,71 @@
...
+}
+
+action_t fd_action_get_identity = {
+  .name           = "get-identity",
+  .args           = NULL,
+  .fn             = get_identity_cmd_fn,
+  .require_config = 1,
+  .perm           = NULL,
+  .description    = "Get the current active identity of the running validator",
+};
```
mmcgee-jump: Newline at EOF

src/app/shared/commands/get_identity.h
```diff
@@ -0,0 +1,14 @@
...
+
+FD_PROTOTYPES_BEGIN
+
+void get_identity_cmd_fn( args_t * args, config_t * config );
+
+FD_PROTOTYPES_END
+
+extern action_t fd_action_get_identity;
+
+#endif /* HEADER_fd_src_app_shared_commands_get_identity_h */
```
mmcgee-jump: Newline at EOF

src/app/shared/commands/get_identity.c
```diff
@@ -0,0 +1,71 @@
...
+  /* Leave the workspace */
+  fd_topo_leave_workspaces( &config->topo );
+}
+
+action_t fd_action_get_identity = {
+  .name           = "get-identity",
+  .args           = NULL,
+  .fn             = get_identity_cmd_fn,
+  .require_config = 1,
+  .perm           = NULL,
```
mmcgee-jump: Another kind of subtle problem here, but there actually is a permission needed for this command, you might need to raise RLIMIT_MLOCK to `mlock(2)` the workspace in memory. No need to fix this either, but maybe add a comment also to fix later.

src/app/shared/commands/get_identity.c
huiskylabs: updated!

src/app/shared/commands/get_identity.c
huiskylabs: updated with `TODO`

src/app/shared/commands/get_identity.c
huiskylabs: removed

src/app/shared/commands/get_identity.c
huiskylabs: removed

src/app/shared/commands/get_identity.c
huiskylabs: Commented `TODO`

huiskylabs:
> Thank you for the contribution!

Thanks for the quick review!

src/app/shared/commands/get_identity.c
```diff
@@ -0,0 +1,66 @@
...
+
+  /* Cast to shred context structure */
+  fd_shred_ctx_t const * shred_ctx = (fd_shred_ctx_t const *)shred_mem;
+  
+  /* The shred tile maintains the current identity in shred_ctx->identity_key.
+     TODO: There's a potential edge case where this could return a torn/corrupt
+     identity key if the shred tile is currently processing a set-identity command
+     and has only overwritten half of the bytes. This is exceedingly rare but
+     should be addressed in the future with proper synchronization. */
+  fd_pubkey_t const * identity_key = shred_ctx->identity_key;
```
ripatel-fd: Let's get this right upfront rather than leaving an edge case around. You could rely on x86 TSO memory ordering to detect a torn read. 

There are two places in the shred tile where the identity gets written:
- On startup (unprivileged_init)
- When switching keys (during_housekeeping)

Let's focus on the latter. The sequence works as follows:

```
T0: state = FD_KEYSWITCH_STATE_SWITCH_PENDING
T1: copies key to tile memory
T1: state = FD_KEYSWITCH_STATE_COMPLETED
```

So, we know that the critical section starts some time after the `SWITCH_PENDING` state is observed,  and ends some time before the `COMPLETED` (or any other state) is observed. Unfortunately, keyswitch has a design flaw where two quick key switches in succession cannot be remotely observed, but we can probably assume that set-identity runs less frequently than it takes to propagate a write across the NOC.

Simplifying the above, we just need to ensure:
- The keyswitch state is the same before and after
- The keyswitch state is not SWITCH_PENDING

The following should work:

```c
/* Peek the pre-read keyswitch state */
ulong switch_state0 = fd_keyswitch_state_query( keyswitch );
/* Speculatively read current key */
fd_pubkey_t key = FD_VOLATILE_CONST( *shred_ctx->identity_key );
/* Peek the post-read keyswitch state */
ulong switch_state1 = fd_keyswitch_state_query( keyswitch );

if( FD_UNLIKELY( switch_state0!=switch_state1 ) ) ... fail ...
if( FD_UNLIKELY( switch_state0==FD_KEYSWITCH_STATE_SWITCH_PENDING ) ) ... fail ...
```

ripatel-fd: Note that `ctx->keyswitch` is in the shred tile's address space, not get-identity.

So, you need to join it from topology

```
  fd_keyswitch_t * keyswitch = fd_keyswitch_join( fd_topo_obj_laddr( topo, tile->keyswitch_obj_id ) );
  FD_TEST( ctx->keyswitch );
```

ripatel-fd: If a failure occurs, worth retrying after 10 milliseconds or so, in case the tile is stuck in the SWITCH_PENDING state for a while.

src/app/shared/commands/get_identity.c
```diff
@@ -0,0 +1,66 @@
...
+    FD_LOG_ERR(( "Shred tile object not found" ));
+  }
+  
+  /* Find the workspace containing the shred tile object */
+  fd_topo_obj_t const * shred_obj = &config->topo.objs[ shred_tile->tile_obj_id ];
+  ulong shred_wksp_id = shred_obj->wksp_id;
+
+  /* Join the workspace in read-only mode */
+  fd_topo_join_workspace( &config->topo, &config->topo.workspaces[ shred_wksp_id ], FD_SHMEM_JOIN_MODE_READ_ONLY );
+  fd_topo_workspace_fill( &config->topo, &config->topo.workspaces[ shred_wksp_id ] );
```
ripatel-fd: This `fd_topo_workspace_fill` call is unnecessary (used to populate some pointers that this piece of code doesn't use), so I would remove it

src/app/shared/commands/get_identity.c
```diff
@@ -0,0 +1,66 @@
...
+
+  /* Access the shred context through the workspace */
+  void * shred_mem = fd_topo_obj_laddr( &config->topo, shred_tile->tile_obj_id );
+  if( FD_UNLIKELY( !shred_mem ) ) {
+    fd_topo_leave_workspaces( &config->topo );
+    FD_LOG_ERR(( "Failed to access shred tile object" ));
+  }
+
+  /* Cast to shred context structure */
+  fd_shred_ctx_t const * shred_ctx = (fd_shred_ctx_t const *)shred_mem;
```
ripatel-fd: This cast is unnecessary in C17, you can just write

```c
fd_shred_ctx_t const * shred_ctx = fd_topo_obj_laddr( &config->topo, shred_tile->tile_obj_id );
```

src/app/shared/commands/get_identity.c
huiskylabs: Thanks for the detailed suggestion. 
Implemented your keyswitch state checking approach. Note that the main use case for me (svs) is calling
`get-identity` after `set-identity`, so racing is unlikely, but handled it properly as requested.

src/app/shared/commands/get_identity.c
huiskylabs: removed

src/app/shared/commands/get_identity.c
huiskylabs: updated

src/app/shared/commands/get_identity.c
ripatel-fd: Thanks for updating the PR, looks good @huiskylabs

agave
```diff
@@ -1 +1 @@
-Subproject commit 97fa8db8bb4dbaeb94b174b079abd857d779d111
```
ripatel-fd: It looks like this submodule changed by accident, can you try 

```
git checkout origin/main -- agave
git commit --amend --no-edit
```

huiskylabs: my bad, fixed

src/app/shared/commands/get_identity.c
```diff
@@ -0,0 +1,93 @@
...
+}
+
+action_t fd_action_get_identity = {
+  .name           = "get-identity",
+  .args           = NULL,
+  .fn             = get_identity_cmd_fn,
+  .require_config = 1,
+  .perm           = NULL, /* TODO: This command may require RLIMIT_MLOCK permissions
+                             to mlock(2) the workspace in memory. This should be
+                             addressed in future updates. */
```
ripatel-fd, line 93: Going to test locally what happens if the prlimit for memlock is zero 🤔

src/app/fdctl/main.c
```diff
@@ -106,6 +107,7 @@ action_t * ACTIONS[] = {
   &fd_action_mem,
   &fd_action_netconf,
   &fd_action_set_identity,
+  &fd_action_get_identity,
```
ripatel-fd, line 110: Can you also add this to fddev, please?

src/app/shared/commands/get_identity.c
ripatel-fd:
```
$ sudo prlimit --pid=$$ --memlock=0
$ build/native/gcc/bin/fdctl get-identity --config src/app/fdctl/config/default.toml 
ERR     07-21 13:32:15.567030 278153 f0   0    src/app/shared/boot/fd_boot.c(45): mlock(0x7ffff7fc0000,4096) (1-EPERM-operation not permitted); unable to lock log file shared lock in memory
```

Sure enough, it's bad ... I'll fix it in a subsequent PR

src/app/shared/commands/get_identity.c
mmcgee-jump: You still need to leave a TODO because this still has an edge case lol. If you're going to fix this just do it fully then ... have the keyswitch (or shred ctx) keep a counter of how many times it's updated the pubkey, and check that before and after as well.

src/app/fdctl/main.c
huiskylabs: done

src/app/shared/commands/get_identity.c
huiskylabs: @mmcgee-jump added a TODO here for this edge case, trying to keep this PR minimal and also the main usage should fully covered (checking identity after set-identity is done)

src/app/shared/commands/get_identity.c
huiskylabs: @mmcgee-jump the requested changes are made, mind to take another look?

huiskylabs:
@mmcgee-jump I’ve updated this as suggested, mind to take another look if it looks good now? 🙏

PR #5758, opened 2025-07-21
snapshots: fix snapin

cali-jumptrading:


src/discof/restore/fd_snapin_tile.c
```diff
@@ -226,6 +226,7 @@ handle_control_frag( fd_snapin_tile_t *  ctx,
         transition_malformed( ctx, stem );
         break;
       }
+      ctx->state = FD_SNAPIN_STATE_LOADING;
```
mmcgee-jump: Ah my bad. But I think you want this in `fd_snapshot_parser_reset` for the other cases too

cali-jumptrading: I don't think this state transition belongs in `fd_snapshot_parser_reset`. I'll add the state transition to the other cases manually

mmcgee-jump: It literally does, if you are resetting the parser it's now ready for new bytes to come in

cali-jumptrading: No this is the state transition for `snapin`. I think we should keep the state machine for snapin separate from the parser.

cali-jumptrading: Otherwise it doesn't make sense for us to have a parser abstraction. We should just bring it into the state machine otherwise.

cali-jumptrading: I think you mean this right? https://github.com/firedancer-io/firedancer/blob/main/src/discof/restore/utils/fd_snapshot_parser.h#L151

we already reset the flags in the parser. The parser should not have any knowledge of the snapin state machine.

mmcgee-jump (approved):

PR #5763, opened 2025-07-21
replay: change fd_exec_slice to a deque dynamic

kbhargava-jump:


contrib/offline-replay/offline_replay.toml
```diff
@@ -11,6 +11,7 @@
         rocksdb_path = "{ledger}/rocksdb"
     [tiles.replay]
         cluster_version = "{cluster_version}"
+        max_exec_slices = 256
```
lidatong, line 14: why is this in `contrib/offline-replay` but other stuff is in `firedancer-dev/config`

src/app/firedancer/config/default.toml
```diff
@@ -1208,6 +1208,9 @@ user = ""
     [tiles.replay]
         cluster_version =  "1.18.0"
 
+        # Maximum number of exec slices that can be buffered in the deque.
+        max_exec_slices = 4096
```
lidatong: this is too small. why did someone change it from my original 64k value?

contrib/offline-replay/offline_replay.toml
```diff
@@ -11,6 +11,7 @@
         rocksdb_path = "{ledger}/rocksdb"
     [tiles.replay]
         cluster_version = "{cluster_version}"
```
lidatong, line 13: not sure what your script is doing, but can you make these actual interpolations as specified by the default.toml like here?

https://github.com/firedancer-io/firedancer/blob/main/src/app/platform/fd_sys_util.c#L33

otherwise it's very confusing

contrib/offline-replay/offline_replay.toml
kbhargava-jump: everything in `firedancer-dev/config` is for general use while offline-replay one is not

contrib/offline-replay/offline_replay.toml
kbhargava-jump: can combine ledger backtest and offline replay to be sourced from a backtest.toml in a separate PR

PR #5769, opened 2025-07-22
add differential fuzzers for chkdup, ar, textstream, hash

marcograss:
add a differential fuzzer for the chkdup implementations, and push some other small, self contained misc fuzzers


src/util/fuzz_hash.c
```diff
@@ -0,0 +1,39 @@
...
+  if ( choice == 0 ) {
+    fd_hash( seed, content, size );
+  } else {
+    uchar *dst = malloc( size );
+    if ( dst == NULL ) {
+        return 0;
+    }
+    fd_hash_memcpy( seed, dst, content, size );
+    free( dst );
+  }
```
ripatel-fd, line 30: As far as I know, fd_hash_memcpy and fd_hash are the same algorithm. We could do both every fuzz iteration. The repeated mallocs and frees should be fast enough

nlgripto (approved):

PR #5775, opened 2025-07-22
simd-0180: vote-based leader schedule in frankendancer

0x0ece:
This is PR 1/4 + 2/4.
- [x] 1/4 Change msg on stake_out to support SIMD-0180: (vote, id, stake) instead of (id, stake). Send the correct msg from Agave to Frank. SIMD-0180 NOT YET implemented. Goal is to show that despite the many changes in many files we’re not breaking anything / all existing tests pass unmodified.
- [x] 2/4 Implement SIMD-0180 in Frank (shred, gui tiles). Based on epoch, compute leader schedule based on vote accounts vs identities. Firedancer will break when feature gets activated in testnet. Goal is to show that Agave vs Frank leader schedule is the same with and without feature, and that Turbine tree remains unchanged.
- [x] 3/4 Implement SIMD-0180 in Firedancer. https://github.com/firedancer-io/firedancer/commit/06e54304970f72e2d012cb0ec07dbfd7229f8d2f
- [ ] 4/4 Clean up / remove unnecessary code (after activation).

-----
Tests.

Most of the tests are unchanged to validate that -despite touching 1m files- the old logic is still working.

I added a few tests for the new logic.

I manually run fdctl in both testnet and mainnet, with the feature disabled and force-enabled. I printed the leader scheduler in both agave and frank and they match. I also printed the contact list in both, they match, and they don't change with/without feature. All as expected.

Finally, I run a version of testnet forcing the feature to be enabled only in frank, but not in agave. In frank, I accepted all shreds (ignoring the leader signature, as by construction I had the wrong leader schedule). I still caught up with the cluster, and I could see the new leader schedule in the gui, consistent with what I could print from the shred tile.

-----

Main issues with simd-0180:
- we need a new message in stake_out, because (Frankendancer) we need to pass (vote, id, stake) instead of (id, stake)
- we need to know the state of the feature gate at boot => easier to pass it as part of stake_out message (than to try to sync different messages)

As a result, fd_stake_weight_msg_t changes format, both its header and the internal stake weights.
- Cascading issue. In too many places the various pieces of size of fd_stake_weight_msg_t were hardcoded => moved to macros so we can update it a bit more easily in future.

At this point we'd have to modify also fd_stake_weight_t from (key, stake) to (vote_key, id_key, stake). I spent over a week and miserably failed, this struct is used everywhere.
=> Introduced a new struct fd_vote_stake_weight_t (vote_key, id_key, stake).

I added fd_vote_stake_weight_t in critical places and handled a copy fd_stake_weight_t<>fd_vote_stake_weight_t.
This is only done once/epoch (and twice at startup) so it's not the end of the world.


src/flamenco/leaders/fd_leaders.c
```diff
@@ -24,7 +24,9 @@ fd_epoch_leaders_new( void  *                   shmem,
                       ulong                     slot_cnt,
                       ulong                     pub_cnt,
                       fd_stake_weight_t const * stakes,
```
mjain-jump: is it possible to make this an `fd_vote_stake_weight_t` type? the leader schedule should perform sample based on the vote pubkeys and associated stakes, but the pubkeys in the schedule still need to be the associated node IDs.

0x0ece: I agree on the spirit, but I'll do the change in a separated PR.
I added some more context in the PR description, I need to do these changes in phases, otherwise I'll break everything.

0x0ece: done

PR #5777, opened 2025-07-22
snapshots: resolve snapshots

cali-jumptrading:
Instead of pinging each snapshot peer, we resolve the full and incremental snapshots for each peer and parse out the slot and hash for each snapshot.

### DIFF
--- a/src/discof/restore/utils/fd_sshttp.c
+++ b/src/discof/restore/utils/fd_sshttp.c
@@ -53,7 +53,7 @@ FD_FN_CONST ulong
 fd_sshttp_footprint( void ) {
   ulong l;
   l = FD_LAYOUT_INIT;
-  l = FD_LAYOUT_APPEND( l, FD_SSHTTP_ALIGN,       sizeof(fd_sshttp_t) );
+  l = FD_LAYOUT_APPEND( l, FD_SSHTTP_ALIGN, sizeof(fd_sshttp_t) );
   return FD_LAYOUT_FINI( l, FD_SSHTTP_ALIGN );
 }
 
--- a/src/discof/restore/utils/fd_ssping.c
+++ b/src/discof/restore/utils/fd_ssping.c
@@ -1,4 +1,6 @@
+#define _GNU_SOURCE
 #include "fd_ssping.h"
+#include "fd_ssresolve.h"
 
 #include "../../../util/bits/fd_bits.h"
 #include "../../../util/log/fd_log.h"
@@ -9,6 +11,7 @@
 #include <sys/socket.h>
 #include <netinet/in.h>
 #include <netinet/ip_icmp.h>
+#include <netinet/tcp.h>
 
 #define PEER_STATE_UNPINGED   0
 #define PEER_STATE_PINGED     1
@@ -16,13 +19,31 @@
 #define PEER_STATE_REFRESHING 3
 #define PEER_STATE_INVALID    4
 
-#define PEER_DEADLINE_NANOS_PING    (1L*1000L*1000L*1000L)     /* 1 second */
+#define PEER_DEADLINE_NANOS_PING    (2L*1000L*1000L*1000L)     /* 2 seconds */
 #define PEER_DEADLINE_NANOS_VALID   (2L*60L*1000L*1000L*1000L) /* 2 minutes */
 #define PEER_DEADLINE_NANOS_INVALID (5L*60L*1000L*1000L*1000L) /* 5 minutes */
 
 struct fd_ssping_peer {
-  ulong         refcnt;
-  fd_ip4_port_t addr;
+  ulong            refcnt;
+  fd_ip4_port_t    addr;
+
+  fd_ssresolve_t * full_ssresolve;
+  fd_ssresolve_t * inc_ssresolve;
+
+  struct {
+
+    struct {
+      ulong     slot;
+      fd_hash_t hash;
+    } full;
+
+    struct {
+      ulong     base_slot;
+      ulong     slot;
+      fd_hash_t hash;
+    } incremental;
+
+  } snapshot_info;
 
   struct {
     ulong next;
@@ -50,6 +71,8 @@ struct fd_ssping_peer {
   } fd;
 
   int   state;
+  ulong full_latency_nanos;
+  ulong incremental_latency_nanos;
   ulong latency_nanos;
   long  deadline_nanos;
 };
@@ -131,8 +154,13 @@ fd_ssping_footprint( ulong max_peers ) {
   l = FD_LAYOUT_APPEND( l, deadline_list_align(), deadline_list_footprint() );
   l = FD_LAYOUT_APPEND( l, deadline_list_align(), deadline_list_footprint() );
   l = FD_LAYOUT_APPEND( l, deadline_list_align(), deadline_list_footprint() );
-  l = FD_LAYOUT_APPEND( l, sizeof(struct pollfd), max_peers*sizeof(struct pollfd) );
-  l = FD_LAYOUT_APPEND( l, sizeof(ulong),         max_peers*sizeof(ulong) );
+  l = FD_LAYOUT_APPEND( l, sizeof(struct pollfd), max_peers*sizeof(struct pollfd)*2UL );
+  l = FD_LAYOUT_APPEND( l, sizeof(ulong),         max_peers*sizeof(ulong)*2UL );
+
+  for( ulong i=0UL; i<max_peers*2UL; i++ ) {
+    l = FD_LAYOUT_APPEND( l, fd_ssresolve_align(), fd_ssresolve_footprint() );
+  }
+
   return FD_LAYOUT_FINI( l, FD_SSPING_ALIGN );
 }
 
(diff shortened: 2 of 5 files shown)


src/discof/restore/utils/fd_ssresolve.h
```diff
@@ -0,0 +1,81 @@
...
+#define FD_SSRESOLVE_ADVANCE_ERROR   (-1) /* fatal error */
+#define FD_SSRESOLVE_ADVANCE_AGAIN   ( 0) /* try again */
+#define FD_SSRESOLVE_ADVANCE_PASS    ( 1) /* no-op */
+#define FD_SSRESOLVE_ADVANCE_SUCCESS ( 2) /* success */
+
+/* fd_ssresolve_advance_poll_out advances the ssresolve state machine
+   when the given socket file descriptor is ready for sending data. */
+int
+fd_ssresolve_advance_poll_out( fd_ssresolve_t * ssresolve,
+                               int              sockfd,
```
mmcgee-jump: I'd probably pass `sockfd` and `addr` during init? These seem like init time fields

src/discof/restore/utils/fd_ssresolve.h
```diff
@@ -0,0 +1,81 @@
...
+
+fd_ssresolve_t *
+fd_ssresolve_join( void * ssresolve );
+
+void
+fd_ssresolve_init( fd_ssresolve_t * ssresolve );
+
+#define FD_SSRESOLVE_ADVANCE_ERROR   (-1) /* fatal error */
+#define FD_SSRESOLVE_ADVANCE_AGAIN   ( 0) /* try again */
+#define FD_SSRESOLVE_ADVANCE_PASS    ( 1) /* no-op */
```
mmcgee-jump: Not sure what the purpose of PASS is, isn't it going to be identical to callers to AGAIN ?

src/discof/restore/utils/fd_ssresolve.h
```diff
@@ -0,0 +1,81 @@
...
+                               fd_ip4_port_t    addr );
+
+/* fd_ssresolve_advance_poll_in advances the ssresolve state machine
+   when the given socket file descriptor is ready for receiving data. */
+int
+fd_ssresolve_advance_poll_in( fd_ssresolve_t *        ssresolve,
+                              int                     sockfd,
+                              fd_ssresolve_result_t * result );
+
+int
```
mmcgee-jump, line 62: Is this needed? Caller should just see they got ADVANCE_SUCCESS or whatever, and mark a `done` variable themselves

src/discof/restore/utils/fd_ssresolve.c
```diff
@@ -0,0 +1,308 @@
...
+  }
+
+  return ssresolve;
+}
+
+void
+fd_ssresolve_init( fd_ssresolve_t * ssresolve ) {
+  ssresolve->state = FD_SSRESOLVE_STATE_FULL_REQ;
+
+  fd_memset( ssresolve->request, 0, sizeof(ssresolve->request) );
```
mmcgee-jump: No need to memset these, since the length field is zeroed

src/discof/restore/utils/fd_ssping.c
```diff
@@ -222,6 +246,13 @@ fd_ssping_add( fd_ssping_t * ssping,
     peer->refcnt = 0UL;
     peer->state  = PEER_STATE_UNPINGED;
     peer->addr   = addr;
+    peer->snapshot_info.full.slot             = ULONG_MAX;
+    peer->snapshot_info.incremental.base_slot = ULONG_MAX;
+    peer->snapshot_info.incremental.slot      = ULONG_MAX;
+    peer->full_latency_nanos        = 0UL;
+    peer->incremental_latency_nanos = 0UL;
+    fd_ssresolve_t * ssresolve = fd_ssresolve_join( fd_ssresolve_new( peer->ssresolve ) );
```
mmcgee-jump: You just want to `fd_ssresolve_init` the `ssresolve` here, you should new/join it only at boot time (e.g. during `unprivileged_init` or whatever).

Think of `new` as one time memory initialization, not repeated. Same for join, it returns a local address join, but should only be done once (for this object it's a no-op).

src/discof/restore/utils/fd_ssping.c
```diff
@@ -339,37 +370,65 @@ poll_advance( fd_ssping_t * ssping,
...
 
     if( FD_LIKELY( pfd->revents & POLLIN ) ) {
-      struct icmphdr icmp_hdr;
-      long result = recv( pfd->fd, &icmp_hdr, sizeof(icmp_hdr), 0 );
-      if( FD_UNLIKELY( -1==result && errno==EAGAIN ) ) continue;
-      else if( FD_UNLIKELY( -1==result || (ulong)result<sizeof(icmp_hdr) || icmp_hdr.type!=ICMP_ECHOREPLY ) ) {
+      fd_ssresolve_result_t resolve_result;
+      int res = fd_ssresolve_advance_poll_in( peer->ssresolve, pfd->fd, &resolve_result );
+
+      if( res==FD_SSRESOLVE_ADVANCE_ERROR ) {
```
mmcgee-jump: kbowers style is to always have an FD_LIKELY/UNLIKELY for these conditions

src/discof/restore/utils/fd_ssresolve.c
```diff
@@ -0,0 +1,308 @@
...
+#include <stdlib.h>
+#include <strings.h>
+
+#include <sys/socket.h>
+#include <netinet/tcp.h>
+#include <netinet/in.h>
+
+#define FD_SSRESOLVE_STATE_FULL_REQ  (0) /* sending request for full snapshot */
+#define FD_SSRESOLVE_STATE_FULL_RESP (1) /* receiving full snapshot response */
+#define FD_SSRESOLVE_STATE_INC_REQ   (2) /* sending request for incremental snapshot */
```
mmcgee-jump: Not sure I love the full->inc path here, because those could technically be issued concurrently. Might be better to keep ssresolve for a single fetch / HTTP redirect follow, and have two file descriptors per peer, that connect() and poll in parallel?

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -47,7 +47,7 @@ FD_FN_CONST ulong
 fd_sshttp_footprint( void ) {
   ulong l;
   l = FD_LAYOUT_INIT;
-  l = FD_LAYOUT_APPEND( l, FD_SSHTTP_ALIGN,       sizeof(fd_sshttp_t) );
+  l = FD_LAYOUT_APPEND( l, FD_SSHTTP_ALIGN, sizeof(fd_sshttp_t) );
```
mmcgee-jump, line 56: You probably want to disallow redirects now for fd_sshttp right, and just have it use the full path with slot and hash determined by the resolver? That way we know the slot is a match and won't get swapped/race conditioned out etc.

src/discof/restore/utils/fd_ssresolve.h
cali-jumptrading: The purpose of PASS is to make the caller continue with polling the state machine rather than trying again. I realized I haven't implemented PASS on the caller side though.. will fix. As in, when the caller receives PASS when polling the file descriptor to send data, it should continue on to try to receive data from the file descriptor rather than polling the next file descriptor.

src/discof/restore/utils/fd_ssresolve.h
cali-jumptrading: oh, this is needed currently because the state machine returns ADVANCE_SUCCESS whenever a state transition is complete, not when the state machine is done. But I think this could be cleaner if we just issued two requests out in parallel; then we wouldn't need this done method.

src/discof/restore/utils/fd_sshttp.c
cali-jumptrading: yeah I realized this PR is going to introduce some more edge cases in the state machine... and also needs to be integrated with snapshot hashes at some point. Thinking about this a bit more... if a peer's snapshot is suddenly changed after we resolve its snapshots but before we start downloading, this should trigger a transition to mark the peer as refreshing so its snapshots can be re-validated...

src/discof/restore/utils/fd_ssresolve.h
mmcgee-jump: The reason I'm suspicious of this is that I've never seen this in any polling state machine before, it's a new concept on a 50 year old thing. I still don't really see what the difference is between "continue with polling the state machine rather than trying again"

cali-jumptrading: um I think this is because I tried to make a simple polling state machine work with the current way ssping is set up, which uses `struct pollfd` to poll for events on file descriptors. And I wanted the no-op return value to signal to ssping to not continue to the next file descriptor, given that this simple polling state machine actually has two entry points. However, I see your point and I think we can simplify all of this by making ssping interpret `try again` as trying the next ssresolve poll entrypoint.

src/discof/restore/utils/fd_ssresolve.h
cali-jumptrading: Actually... in my opinion I think having the `fd_ssresolve_is_done` method is cleaner so that the caller doesn't have to track extra state related to the `ssresolve` state machine object. I'd rather the caller just call this exposed method to check if the state machine is done.

src/discof/restore/utils/fd_ssresolve.c
cali-jumptrading: Done, nice idea to simplify the state machine and ping peers faster.

src/discof/restore/utils/fd_ssping.c
cali-jumptrading: Did you have any concerns about how I'm storing the `ssresolve` object in `fd_ssping_peer_t`? I used `ssresolve[1]` to not have to allocate the ssresolve struct in some random scratch memory because our data structures like pool generally assume the elements are simple POD structs.

src/discof/restore/utils/fd_ssresolve.c
```diff
@@ -0,0 +1,285 @@
...
+  ssresolve->response_len = 0UL;
+}
+
+static void
+fd_ssresolve_render_req( fd_ssresolve_t * ssresolve,
+                         fd_ip4_port_t    addr ) {
+  ssresolve->request_sent = 0UL;
+  ssresolve->response_len = 0UL;
+
+  if( ssresolve->full ) {
```
mmcgee-jump:
```suggestion
  if( FD_LIKELY( ssresolve->full ) ) {
```

src/discof/restore/utils/fd_ssresolve.c
```diff
@@ -0,0 +1,285 @@
...
+  ssresolve->state        = FD_SSRESOLVE_STATE_REQ;
+  ssresolve->request_sent = 0UL;
+  ssresolve->request_len  = 0UL;
+  ssresolve->response_len = 0UL;
+}
+
+static void
+fd_ssresolve_render_req( fd_ssresolve_t * ssresolve,
+                         fd_ip4_port_t    addr ) {
+  ssresolve->request_sent = 0UL;
```
mmcgee-jump: Don't think you need either of these lines? Actually it's buggy, cause if you don't send the whole request in one send() syscall it's going to loop sending it forever

cali-jumptrading: oh right, an artifact from before the state machine was changed.

src/discof/restore/utils/fd_ssping.c
```diff
@@ -9,20 +11,41 @@
...
 #define PEER_DEADLINE_NANOS_VALID   (2L*60L*1000L*1000L*1000L) /* 2 minutes */
 #define PEER_DEADLINE_NANOS_INVALID (5L*60L*1000L*1000L*1000L) /* 5 minutes */
 
 struct fd_ssping_peer {
-  ulong         refcnt;
-  fd_ip4_port_t addr;
+  ulong            refcnt;
+  fd_ip4_port_t    addr;
+
+  /* TODO: how do I hide the ssresolve struct and not have to alloc it
```
mmcgee-jump: Make these pointers, like `fd_ssresolve_t * full`, then stick the struct on the end of FOOTPRINT and NEW, e.g. in new

```
void * _resolve_full = FD_SCRATCH_ALLOC( ssresovle_align(), ss_resolv_footprint() );
ctx->resolv_full = fd_ssresolv_join( fd_ssresolv_new( _resolv_full ) ); 
FD_TEST( ctx->resolv_full )
```

src/discof/restore/utils/fd_ssping.c
```diff
@@ -323,53 +368,81 @@ fd_ssping_invalidate( fd_ssping_t * ssping,
...
+          unping_peer( ssping, peer_pool_ele( ssping->pool, ssping->fds_idx[ i ] ), now );
+          continue;
+        } else if( FD_UNLIKELY( res==FD_SSRESOLVE_ADVANCE_AGAIN ) ) {
+          continue;
+        } else { /* FD_SSRESOLVE_ADVANCE_SUCCESS */
+          FD_TEST( peer->deadline_nanos>now );
+          if( resolve_result.base_slot==ULONG_MAX ) {
+            peer->snapshot_info.full.slot = resolve_result.slot;
+            memcpy( &peer->snapshot_info.full.hash, &resolve_result.hash, sizeof(fd_hash_t) );
+            peer->full_latency_nanos = PEER_DEADLINE_NANOS_PING - (ulong)(peer->deadline_nanos - now);
```
mmcgee-jump, line 424: At first glance I'd think we want to use time to first byte to measure latency, not time to finish the whole state machine. But I'm not sure. Any thoughts?

src/discof/restore/utils/fd_ssping.c
```diff
@@ -323,53 +368,81 @@ fd_ssping_invalidate( fd_ssping_t * ssping,
 }
 
 static inline void
-poll_advance( fd_ssping_t * ssping,
-              long          now ) {
-  if( FD_LIKELY( !ssping->fds_len ) ) return;
+poll_fds( fd_ssping_t * ssping,
+          int           full,
+          long          now ) {
+  struct pollfd * fds = full ? ssping->full_fds : ssping->incremental_fds;
```
mmcgee-jump: Why split the fds list in two? You could just have one array here and `poll` the combined list, saves a syscall also

cali-jumptrading: I wanted to keep the polling logic simple... and not have to track two fds per peer in one list. But I didn't consider the extra syscall...

src/discof/restore/utils/fd_ssping.c
cali-jumptrading: I was just thinking round trip time because it's simple to measure. Not opposed to time to first byte though.

cali-jumptrading:
completed by https://github.com/firedancer-io/firedancer/pull/6285

PR #5780, opened 2025-07-23
chore(firedancer): update tomls

lidatong:


src/app/firedancer/config/default.toml
```diff
@@ -1288,7 +1292,11 @@ user = ""
     #
     # In development, you can disable the sandbox for testing and
     # debugging with the `--no-sandbox` argument to `firedancer-dev`.
-    sandbox = true
+    sandbox = false # FIXME
+
+    # Setting this to true will disable any Agave-specific components
+    # ie. runs the full Firedancer topology.
+    no_agave = true # FIXME
```
mmcgee-jump: Not needed for full binary

src/app/firedancer/config/default.toml
```diff
@@ -223,7 +223,11 @@ user = ""
     # These entrypoints should be highly trusted, as they are used to
     # bootstrap.  A compromised entrypoint could trick the validator
     # into loading a bad snapshot, or joining a malicious cluster.
-    entrypoints = []
+    entrypoints = [
+        "entrypoint.testnet.solana.com:8001",
```
mmcgee-jump: Should not be part of default.toml

src/app/firedancer/config/default.toml
```diff
@@ -1315,7 +1323,7 @@ user = ""
     # This option can be enabled in production if a crash cannot be
     # easily reproduced, but it decreases security and should not be
     # enabled during routine running of the validator.
-    core_dump = false
+    core_dump = true # FIXME
```
mmcgee-jump: Not in `default.toml`, these are operator defaults

ripatel-fd: Core dumping on crash is a very good operator default

mmcgee-jump: No it is not, it weakens the sandbox

ripatel-fd: @mmcgee-jump How so? It wouldn't dump anything containing keys like resolv or sign tile, if we do it right

mmcgee-jump: See `fd_sandbox.c`, the dumpable bit of the process controls a few security things which are nice defense in depth

PR #5790, opened 2025-07-24
Fix metrics generation to support converters for counters

npatel-jump:
Fixed counter/gauge metrics with converter="nanoseconds" showing CPU ticks instead of nanoseconds by applying converters to all metric types.


agave
ripatel-fd, line 1: please undo this change

PR #5792, opened 2025-07-24
runtime: solcap support for solfuzz

ripatel-fd:


mjain-jump (approved):
Nice! This would be useful for debugging block fuzzing mismatches. Would need to set up a similar one for Agave though.

ripatel-fd:
It's still useful for debugging Firedancer regressions during development

PR #5794, opened 2025-07-24
Add cooperatie scheduling/low power mode for Firedancer tiles

bzhangabylov-jump:
This PR implements cooperative scheduling for Firedancer tiles, allowing them to sleep when idle instead of busy-spinning. This improves CPU utilization and power efficiency by waking up at the next housekeeping interval to process work when available.

### Changes:
* fd_stem.c: Added idle detection that tracks whether work is performed in before/after credit callbacks (via charge_busy) or if fragments are processed. When idle iterations exceed STEM_IDLE_THRESHOLD (configurable per tile), the tile sleeps until the next housekeeping deadline. Lower thresholds result in more frequent sleeping and reduced CPU utilization, but may impact throughput.
* Configuration: Added low_power_mode option under [layout] to enable power-efficient operation. This flag propagates through the stack: config → topology → tile → sleep argument in STEM_run. In topology.c (fddev), it's passed to all tiles except metric and cswtch (which sleep by default). Defaults to 0 in other topologies.
* Metrics: Added a new "sleeping" regime to track time spent sleeping. Updated the monitor screen to display this regime.
* Syscalls: Updated seccomp policies to allow sleep syscalls used by fd_log_sleep (sched_yield, clock_nanosleep). Changed fd_log_sleep to use clock_nanosleep for compatibility with modern glibc wrappers.

Extra TODO:
* Shared "Am I Leader" state to disable the feature when leader to ensure no performance impact
* Once validated for fddev, integrate and set defaults for full client (some modified functions are shared)
* Run on mainnet to measure performance and CPU/power differences vs main
* Add adaptive seccomp policy based on feature flag
* Enhanced floating support: maintain NUMA awareness, limit CPU core range


src/disco/stem/fd_stem.c
```diff
@@ -492,6 +500,20 @@ STEM_(run1)( ulong                        in_cnt,
       now = next;
     }
 
+    idle_iter_cnt++;
+
+    if ( FD_UNLIKELY( sleeps && idle_iter_cnt>STEM_IDLE_THRESHOLD ) ) {
+      long ticks_until_deadline = then - fd_tickcount();
```
mmcgee-jump: `then - now` , only one tickcount per loop

src/disco/stem/fd_stem.c
```diff
@@ -492,6 +500,20 @@ STEM_(run1)( ulong                        in_cnt,
...
+      long ns_until_deadline    = (long)((double)ticks_until_deadline / fd_tempo_tick_per_ns(NULL));
```
mmcgee-jump: No `fd_tempo_tick_per_ns` here, cache it

src/disco/metrics/metrics.xml
```diff
@@ -37,6 +37,8 @@ metric introduced.
 
     <int value="6" name="CaughtUpPostfrag" label="Caught up + Postfrag" />
     <int value="7" name="ProcessingPostfrag" label="Processing + Postfrag" />
+
+    <int value="8" name="Sleeping" label="Sleeping" />
```
mmcgee-jump, line 41: I think you gotta update the GUI / frontend to handle this too

src/disco/stem/fd_stem.c
```diff
@@ -552,6 +574,7 @@ STEM_(run1)( ulong                        in_cnt,
       long next = fd_tickcount();
       metric_regime_ticks[ 3+was_busy ] += (ulong)(next - now);
       now = next;
+      if( was_busy ) idle_iter_cnt = 0;
```
mmcgee-jump:
```suggestion
      if( FD_UNLIKELY( was_busy ) ) idle_iter_cnt = 0;
```

src/disco/stem/fd_stem.c
```diff
@@ -190,6 +190,10 @@
 #define STEM_LAZY (0L)
 #endif
 
+#ifndef STEM_IDLE_THRESHOLD
+#define STEM_IDLE_THRESHOLD (10000UL)
```
mmcgee-jump: Feels too high? How chosen? We observed between 6 and 100 in some random libraries

src/disco/sign/fd_sign_tile.seccomppolicy
```diff
@@ -16,3 +16,7 @@ write: (or (eq (arg 0) 2)
 #
 # arg 0 is the file descriptor to fsync.
 fsync: (eq (arg 0) logfile_fd)
+
+# sleep operations: fd_log_sleep uses these syscalls
```
mmcgee-jump: Not a useful comment, describe why we sleep! Who is calling fd_log_sleep ?

src/app/fdctl/topology.c
```diff
@@ -117,24 +117,24 @@ fd_topo_initialize( config_t * config ) {
...
-  /**/                 fd_topob_tile( topo, "poh",     "poh",     "metric_in",  tile_to_cpu[ topo->tile_cnt ], 1,        1 );
-  FOR(shred_tile_cnt)  fd_topob_tile( topo, "shred",   "shred",   "metric_in",  tile_to_cpu[ topo->tile_cnt ], 0,        1 );
-  /**/                 fd_topob_tile( topo, "store",   "store",   "metric_in",  tile_to_cpu[ topo->tile_cnt ], 1,        0 );
-  /**/                 fd_topob_tile( topo, "sign",    "sign",    "metric_in",  tile_to_cpu[ topo->tile_cnt ], 0,        1 );
-  /**/                 fd_topob_tile( topo, "metric",  "metric",  "metric_in",  tile_to_cpu[ topo->tile_cnt ], 0,        0 );
-  /**/                 fd_topob_tile( topo, "cswtch",  "cswtch",  "metric_in",  tile_to_cpu[ topo->tile_cnt ], 0,        0 );
+  /*                                  topo, tile_name, tile_wksp, metrics_wksp, cpu_idx,                       is_agave, uses_keyswitch,               sleeps */
+  FOR(quic_tile_cnt)   fd_topob_tile( topo, "quic",    "quic",    "metric_in",  tile_to_cpu[ topo->tile_cnt ], 0,        0,                            config->layout.low_power_mode );
```
mmcgee-jump: Doesn't look like anything changed about affinity here? Don't you need to unpin these cores? And allow Agave to use them again? And update the auto layout etc?

src/app/fdctl/config/default.toml
```diff
@@ -819,6 +819,18 @@ dynamic_port_range = "8900-9000"
     # very high TPS rates because the cluster size will be very small.
     shred_tile_count = 1
 
+    # Enable cooperative scheduling for tiles to reduce CPU usage when
+    # idle.  When enabled, tiles will sleep after detecting inactivity
+    # for a configurable number of iterations, rather than continuously
```
mmcgee-jump: Comment seems inaccurate? Where is the configurable iterations? Also don't think we should allow that to be configurable.

src/app/fdctl/config/default.toml
```diff
@@ -819,6 +819,18 @@ dynamic_port_range = "8900-9000"
...
+    # spinning for new work.  This significantly reduces power consumption
+    # and CPU usage during periods of low activity.
+    #
+    # The tradeoff is a small increase in latency when resuming from
```
mmcgee-jump: This is downplaying the tradeoff, I think. We are defaulting it to off remember. I'd try and explain the philosophy of Firedancer a bit more at the beginning, we want to use all the system resources that are available, so yielding to the kernel is not ideal.

src/disco/stem/fd_stem.c
bzhangabylov-jump: I don't think we can drop as low as 6-100 iterations - tiles wouldn't have enough time to do work before sleeping, so throughput suffers, and going lower gives very marginal CPU utilization benefits. 
My initial math was 10^9 / 81920 (TPS) / ~8 nanos per stem loop iteration (varies by tile but rough estimate) = ~1500 iterations. 10k was a safe conservative pick that gets most CPU gains without risking throughput. Just ran a few experiments that confirm this - lower values like 512/1024 give very slight CPU utilization benefit but become unstable in throughput. 2048 seems to be around the lowest bound that captures basically all CPU utilization gains while reliably maintaining 75-82k TPS. 
This can obviously be super tile-specific, which is why tiles can override the macro themselves - this is just the default value, which I'd rather keep with some slack.

src/app/fdctl/topology.c
bzhangabylov-jump: I kept low_power_mode separate from affinity controls for now, just because they are different features and someone might want to various combinations: pinned + sleeping, floating + sleeping. That said, if low_power_mode should be like a switch that potentially overrides layout affinities or just does something different for the “auto” settings, I can update it

src/app/fdctl/config/default.toml
bzhangabylov-jump: The comment refers to the adjustable macro, but it is configurable by the tile creator not the end user, so you are right, shouldn't be mentioned here

src/app/fdctl/config/default.toml
bzhangabylov-jump: updated

src/util/log/fd_log.c
```diff
@@ -489,7 +489,7 @@ fd_log_sleep( long dt ) {
   struct timespec rem[1];
   req->tv_sec  = (time_t)( ((ulong)ns_dt) / ((ulong)1e9) ); /* in [0,2^31-1] */
   req->tv_nsec = (long)  ( ((ulong)ns_dt) % ((ulong)1e9) ); /* in [0,1e9) */
-  if( FD_UNLIKELY( nanosleep( req, rem ) ) && FD_LIKELY( errno==EINTR ) ) dt += ((long)1e9)*((long)rem->tv_sec) + rem->tv_nsec;
+  if( FD_UNLIKELY( clock_nanosleep( CLOCK_REALTIME, 0, req, rem ) ) && FD_LIKELY( errno==EINTR ) ) dt += ((long)1e9)*((long)rem->tv_sec) + rem->tv_nsec;
```
mmcgee-jump: Why? This changed from CLOCK_MONOTONIC to REALTIME ?

src/disco/verify/fd_verify_tile.seccomppolicy
```diff
@@ -16,3 +16,8 @@ write: (or (eq (arg 0) 2)
 #
 # arg 0 is the file descriptor to fsync.
 fsync: (eq (arg 0) logfile_fd)
+
+# low_power_mode: stem calls fd_log_sleep to reduce CPU usage when
+#                 the tile is idle. Can be enabled by configuration.
+sched_yield
+clock_nanosleep
```
mmcgee-jump: Sounds silly but please make this maximally restrictive, aka check the arguments

src/app/fdctl/version.mk
```diff
@@ -1,3 +1,3 @@
 VERSION_MAJOR := 0
-VERSION_MINOR := 1
+VERSION_MINOR := 7
```
mmcgee-jump: Revert

src/app/shared/fd_config.h
```diff
@@ -255,6 +255,7 @@ struct fd_config {
     uint verify_tile_count;
     uint bank_tile_count;
     uint shred_tile_count;
+    int  low_power_mode;
```
0x0ece, line 256: I would call it `low_cores_mode`, low power seems weak... in fact this is prob a high power mode for low end machines :)

ripatel-fd:
This is great work. I'm closing this for now since v0.7 is feature-frozen, but we will definitely revisit.

PR #5808, opened 2025-07-25
arm64: standardize names

anwayde:
- Added a space at the end of `populate_sock_filter_policy_*` codegen function for consistency (which ended up editing a big chunk of files)
- Renamed `sock` and `xdp` seccomppolicies to `fd_sock_tile` and `fd_xdp_tile` for consistency

Successful arm64 build run: https://github.com/firedancer-io/firedancer/actions/runs/16521778788/job/46725086193


anwayde:
Change request addressed here: https://github.com/firedancer-io/firedancer/pull/5857

Rebased and new successful run: https://github.com/firedancer-io/firedancer/actions/runs/16629597937/job/47055245799

src/disco/gui/fd_gui_tile.c
ripatel-fd, line 1: @anwayde Could you replace poll() in http_server.c with the new poll wrapper? Then we can get rid of the separate arm policy for the metric and gui tiles.

anwayde: yup sure, will make another separate PR like the sock and bundle one

anwayde:
Ok after https://github.com/firedancer-io/firedancer/pull/5879, nightly arm64 builds are passing. So this PR is basically a formatting change. @ripatel-fd feel free to close it if it seems not necessary anymore.

PR #5809, opened 2025-07-25
runtime: adding epoch-boundary based bank bounds

ibhatt-jumptrading:


src/app/firedancer/config/default.toml
```diff
@@ -418,7 +418,8 @@ user = ""
         max_transactions_per_slot = 65536
         snapshot_grace_period_seconds = 409
         max_vote_accounts = 2000000
-        max_banks = 64
+        max_total_banks = 128
+        max_fork_width = 32
```
ripatel-fd, line 439: Please add some documentation what these limits mean (default.toml is a good place to do these docs)

PR #5815, opened 2025-07-26
snaprd: replace testnet peer

alpeng-jump:


ripatel-fd:
Hardcoding IPv4 addresses a terrible idea (what if this is a rented IP address that gets returned to a pool, and used by someone who isn't Solana). And even more so if it's some random person's host we don't know. And even more so considering that Solana trusts snapshots and has no defenses against corrupting some bank data structures (the validator would run fine for half a day and then calculate a malicious result, for example). All of which together can lead to hostile takeover of that IP. The only records that should be in there are DNS records of entrypoints, but not a hardcoded IP. (cc @t-nelson curious how Agave does bootstrapping of snapshots).

PR #5828, opened 2025-07-28
Add firedancer-dev backtest --solcap-path flag

ripatel-fd:


kbhargava-jump:
solcap is already enabled on backtest:
```
[capture]
    capture_start_slot = {start slot}
    solcap_capture = "{solcap path}"
 ```

ripatel-fd:
I'll probably redo the command-line interface when we have the functionality all-in-one and can get rid of the wrapper shell script that generates the TOML

PR #5834, opened 2025-07-28
Remove pack-replay link

ripatel-fd:


src/discof/replay/fd_replay_tile.c
```diff
@@ -47,7 +44,6 @@
 #define PLUGIN_PUBLISH_TIME_NS ((long)60e9)
 
 #define IN_KIND_REPAIR (0)
```
ibhatt-jumptrading, line 46: will link indexing still be correct?

ibhatt-jumptrading (approved):

src/discof/replay/fd_replay_tile.c
ripatel-fd: Yes, these are stable / don't have to be gapless

PR #5842, opened 2025-07-28
map_chain: configurable null idx

emwang-jump:


src/util/tmpl/fd_map_chain.c
```diff
@@ -819,9 +819,7 @@ MAP_(idx_insert)( MAP_(t) *   join,
   pool[ ele_idx ].MAP_PREV = MAP_(private_box)( MAP_(private_idx_null)() );
 # endif
   pool[ ele_idx ].MAP_NEXT = *head;
-# if MAP_INSERT_FENCE
-  FD_COMPILER_MFENCE();
-# endif
+  FD_COMPILER_MFENCE(); /* Useful in limited SPSC use cases that assume head points to the rest of the chain */
```
ripatel-fd: I'm a tiny bit worried of this blocking some compiler optimizations for users that don't have concurrency requirements. Would you mind spot-checking the assembly generated by `make -j asm` before and after for any of the other map_chain users?

emwang-jump: just keeping the opt-in macro for ease

PR #5854, opened 2025-07-29
ci: make jobs cleanup after themselves

anwayde:
This splits up the hugepage action so that jobs can cleanup after themselves. This also adds the ability to use different shmem paths if that ever becomes necessary. There is no change in the `allocate` behaviour as it still cleans up as before.

### DIFF
--- a/.github/actions/hugepages/action.yml
+++ b/.github/actions/hugepages/action.yml
@@ -9,19 +9,42 @@ inputs:
     description: 'Number of gigantic pages'
     required: true
     default: '64'
+  shmem_path:
+    description: 'Base path for shmem mount'
+    required: false
+    default: '/mnt/.fd'
+  action:
+    description: 'Operation to perform'
+    required: true
+    choice:
+    - allocate
+    - cleanup
+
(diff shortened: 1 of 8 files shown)


.github/actions/hugepages/action.yml
```diff
@@ -9,19 +9,42 @@ inputs:
     description: 'Number of gigantic pages'
     required: true
     default: '64'
+  shmem_path:
+    description: 'Base path for shmem mount'
+    required: false
+    default: '/mnt/.fd'
```
topointon-jump, line 15: I'm not sure this is necessary, we can just use the default mount path all the time (as we do today). Having two different mount paths was very confusing (my bad).

anwayde: We are using the default path, this just has the option (if it's ever needed). None of this changes anything or introduces new paths.

topointon-jump: Yeah I just think we should discourage people from doing this lol (but if you want to keep the config option feel free)

PR #5856, opened 2025-07-29
flamenco, program-cache: remove end-of-slot iterations

mjain-jump:


src/flamenco/runtime/program/fd_bpf_loader_program.c
```diff
@@ -2545,29 +2571,31 @@ fd_bpf_loader_program_execute( fd_exec_instr_ctx_t * ctx ) {
...
+       could also theoretically cause some currently-deployed programs
+       to fail in the future if ELF / VM checks are eventually made
+       stricter. */
+    fd_program_cache_entry_t const * cache_entry = NULL;
+    if( FD_UNLIKELY( fd_program_cache_load_entry( ctx->txn_ctx->funk,
+                                                  ctx->txn_ctx->funk_txn,
+                                                  program_id,
+                                                  &cache_entry )!=0 ) ) {
       fd_log_collector_msg_literal( ctx, "Program is not cached" );
+      FD_LOG_WARNING(("FAILED BOZOOOOO"));
```
cali-jumptrading: haha remove?

src/flamenco/runtime/program/fd_program_cache.c
```diff
@@ -1,54 +1,56 @@
...
 
   /* Failed verification flag */
-  validated_prog->failed_verification = 0;
+  cache_entry->failed_verification = 0;
+
+  /* Last slot the program was modified */
+  cache_entry->last_slot_modified = last_slot_modified;
+
+  /* Last slot verification checks were ran for this program */
+  cache_entry->last_slot_verification_ran = last_slot_verification_ran;
```
cali-jumptrading: This makes more sense to me, but up to you!
```suggestion
  cache_entry->last_slot_verified = last_slot_verified;
```

src/flamenco/runtime/program/fd_program_cache.h
```diff
@@ -0,0 +1,229 @@
...
+   reverified since the last epoch (regardless of it having been
+   modified or not), it will be reverified and updated.
+
+   If a program fails verification due to invalid ELF headers / sBPF
+   loading failures, then we update the `failed_verification` tag
+   and set `last_slot_modified` to the current slot. A future
+   program upgrade / network feature set change could make this
+   program valid again.
+
+   A key invariant with the program cache design is that it does not
```
cali-jumptrading, line 62: Is there an upper bound on the size of the program cache?

mjain-jump: nope

src/flamenco/runtime/program/fd_program_cache.c
mjain-jump: i like that a lot better. fixed

src/flamenco/runtime/context/fd_exec_txn_ctx.h
```diff
@@ -106,6 +106,13 @@ struct fd_exec_txn_ctx {
   fd_txn_account_t                executable_accounts[ MAX_TX_ACCOUNT_LOCKS ]; /* Array of BPF upgradeable loader program data accounts */
   fd_txn_account_t                accounts[ MAX_TX_ACCOUNT_LOCKS ];            /* Array of borrowed accounts accessed by this transaction. */
 
+  /* When a program is deployed or upgraded, we must queue it to be
+     updated in the program cache (if it exists already) so that
+     the cache entry's ELF / sBPF information can be updated for future
+     executions. We use a uchar array to track which accounts need
+     to be queued for reverification. */
+  uchar                           program_idx_needs_reverify[ MAX_TX_ACCOUNT_LOCKS ];
```
topointon-jump: Worth documenting the sementics of this: if program_idx_needs_reverify[x] = 1, then we will reverify the program at account x in the transaction at the start of the next slot, etc

topointon-jump: Maybe `program_acc_idx_needs_reverify` would be a better name? To make it clear the array refers to account indicies.

src/flamenco/runtime/program/fd_bpf_loader_program.c
```diff
@@ -251,6 +275,13 @@ fd_deploy_program( fd_exec_instr_ctx_t * instr_ctx,
   if( FD_UNLIKELY( validate_result!=FD_VM_SUCCESS ) ) {
     return FD_EXECUTOR_INSTR_ERR_INVALID_ACC_DATA;
   }
+
+  /* Queue the program for reverification */
```
topointon-jump, line 279: Could we document _when_ the program will be reverified?

src/flamenco/runtime/program/fd_program_cache.h
topointon-jump: we should document that the size of the program cache is bounded by the size of the funk instance

src/flamenco/runtime/program/fd_program_cache.h
```diff
@@ -0,0 +1,228 @@
...
+   evict any entries - it only grows through the lifetime of the
+   client's execution. This is meant to prevent any DOS vectors where
+   an attacker can invoke programs that repeatedly reverify and
+   evict programs. Because of this invariant, we do not touch cache
+   entries for programs which are closed, and instead let account
+   loading checks catch cases where a user tries to invoke a closed
+   program.
+
+   EDGE CASES (and how we handle them):
```
topointon-jump, line 80: Could you make sure that all of these edge cases are covered by the unit test?

src/flamenco/runtime/program/fd_program_cache.h
```diff
@@ -0,0 +1,228 @@
...
+#include "../context/fd_exec_slot_ctx.h"
+#include "../../vm/syscall/fd_vm_syscall.h"
+#include "../fd_system_ids.h"
+
+/* fd_program_cache contains the core logic for the program cache's
+   behavior, including accesses, insertions, updates, and verifies.
+   Specifically, the program cache operates in a lazy manner where
+   programs are only inserted / reverified when they are referenced
+   in a transaction.
+
```
topointon-jump, line 15: Could we also document the concurrency requirements/guarentees of the program cache? As well as where it is stored (inside Funk atm).

src/flamenco/runtime/program/fd_bpf_loader_program.c
mjain-jump: Already documented in `fd_program_cache.h`

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: done

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: done

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: @kbhargava-jump added a ledger test [here](https://github.com/firedancer-io/firedancer/pull/5874)

topointon-jump: Could we write unit tests also? 🙏

cali-jumptrading: +1

src/flamenco/runtime/context/fd_exec_txn_ctx.h
mjain-jump: yea i agree. updated

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: i already added unit tests in test_program_cache.c

src/flamenco/runtime/fd_runtime.h
```diff
@@ -562,7 +562,9 @@ fd_runtime_block_pre_execute_process_new_epoch( fd_exec_slot_ctx_t * slot_ctx,
                                                 fd_spad_t *          runtime_spad,
                                                 int *                is_epoch_boundary );
 
-/* This function is responsible for inserting fresh entries or updating existing entries in the program cache.
+/* TODO: UPDATE THIS COMMENT
```
ibhatt-jumptrading: intentional?

src/flamenco/runtime/program/fd_program_cache.h
```diff
@@ -0,0 +1,271 @@
...
+   the writable account keys and queue any programs for reverification
+   since their account data may have changed (e.g. the program was
+   upgraded, closed, etc). To mark a program for reverification, we
+   simply update the `last_slot_modified` field to the current slot.
+   When the program is next referenced in a transaction,
+   `fd_program_cache_update_program()` will check if the program was
+   modified since the last time it was reverified, and reverify
+   the program accordingly to update the cache entry's sBPF info,
+   calldests, etc. Once the program has been reverified, the entry's
+   `last_slot_verified` field will be updated to the current slot.
```
ripatel-fd: This implies that the program's cache entry is updated in-place. 

What happens with the cache entry if a block reverts by 1 slot? 
Obviously the `last_slot_modified` value would then be different (too high, since it was modified in a later slot). But how would the block revert routine know which cache entries to revert? Does it walk the transaction list of the reverted block again? 

Or is the cache entry copied into a new funk record such that the revert becomes trivial?
If it's copied into a new funk record, why is this field needed? Wouldn't the transaction XID already imply the recency of a funk cache entry?

src/flamenco/runtime/program/fd_program_cache.h
```diff
@@ -0,0 +1,271 @@
...
+   the network's active feature set may have changed, and any previously
+   {valid,invalid} programs may now be {valid,invalid}. Therefore,
+   if a program is referenced in a transaction that has not been
+   reverified since the last epoch (regardless of it having been
+   modified or not), it will be reverified and updated.
+
+   If a program fails verification due to invalid ELF headers / sBPF
+   loading failures, then we update the `failed_verification` tag,
+   set `last_slot_verified` to the current slot, and set
+   `last_slot_modified` to 0. A future program upgrade / network feature
```
ripatel-fd, line 59: Why is last_slot_modified zero instead of the current slot?

src/flamenco/runtime/program/fd_program_cache.h
```diff
@@ -0,0 +1,271 @@
...
+   set change could make this program valid again.
+
+   A key invariant with the program cache design is that it does not
+   evict any entries - it only grows through the lifetime of the
+   client's execution. This is meant to prevent any DOS vectors where
+   an attacker can invoke programs that repeatedly reverify and
+   evict programs. Because of this invariant, we do not touch cache
```
ripatel-fd: > This is meant to prevent any DOS vectors where an attacker can invoke programs that repeatedly reverify and evict programs.

I don't follow this logic. The attacker can always add entries and cause them to get verified (by spamming program deployments/retracts). In a cache with an eviction policy, this condition is survivable. In a cache that grows unbounded, you will OOM and crash. So, isn't this design more susceptible to DoS?

src/flamenco/runtime/program/fd_program_cache.h
```diff
@@ -0,0 +1,271 @@
...
+   entries for programs which are closed, and instead let account
+   loading checks catch cases where a user tries to invoke a closed
+   program.
+
+   Another key invariant is that for any given program, we will insert /
+   update its cache entry at most ONCE per slot, and this will happen
+   the FIRST time the program is referenced in a transaction before it
+   is dispatched to the exec tiles. This is important because it ensures
```
ripatel-fd, line 72: This is more specific to the replay/exec tile scheduling design than the program cache API. But I appreciate the docs nonetheless

src/flamenco/runtime/program/fd_program_cache.h
```diff
@@ -0,0 +1,271 @@
...
+           not do anything because the program does not exist in the
+           cache yet, and so there is nothing to update. The next time
+           the program is referenced in a transaction,
+           `fd_program_cache_update_program()` will see that the program
+           is missing in the cache and will perform ELF / sBPF
+           verifications and insert the entry into the cache.
+     - Deploying and invoking in the same slot
+         - BPF loader checks will fail because the program account's
+           slot value will be equal to the current slot, which
+           gets caught by the `DelayedVisibility` checks.
```
ripatel-fd, line 93: Out of curiosity, what happens if the BPF loader check breaks for some reason? Does the program cache just return an outdated record, or does something less predictable happen?

src/flamenco/runtime/program/fd_program_cache.h
```diff
@@ -0,0 +1,271 @@
...
+   - Upgrading programs
+      - Upgrading a program
```
ripatel-fd, line 95: What is the distinction between "deploying" and "upgrading" a program? 

If I close a program, delete the program account, and then redeploy the program does it count as a deploy or an upgrade? Note that according to your "closing programs" docs, the cache entry would survive the program deletion

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: Because if someone spams transactions on a program that fails verification, we would unnecessarily keep running verifications on the same program. Setting it to 0 makes it non-reverifiable ever unless we're at a new epoch or someone upgrades the program.

src/flamenco/runtime/program/fd_program_cache.h
```diff
@@ -0,0 +1,271 @@
...
+
+   Q: Is the program cache concurrency-safe?
+   A: Yes. In any given slot, the replay tile first iterates through
+      each transaction in a single-threaded manner. For each
+      transaction, it will read the accounts and will call
+      `fd_program_cache_update_program()` for each program, and then
+      dispatch the transaction to the exec tiles. This means
+      that the VERY FIRST TIME any program is referenced in ANY
+      transaction in the slot, the replay tile will update the
+      associated program cache entry BEFORE dispatching the
```
ripatel-fd: Nit but I would make this a bit less strongly worded, since the scheduling concept you described is very straightforward

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: The BPF loader program will read the slot value from the program account, not the cache record. So the cache is never actually touched here

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: This is true, but with the program cache currently in Funk there is no straightforward way to implement any kind of eviction policy (e.g. LRU). Definitely should discuss this when we move it into its own data structure

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: The record is cloned down into the current funk transaction (if it doesnt exist already) and then modified there. I added a few docs on it. It was cleaner to add a field containing the last modified slot rather than modifying / making a new Funk API to return the funk transaction's recency. But yes technically you could check the funk transaction's age to determine this, but then you still need a flag of some sort to indicate that the program cache needs to reverify the program the next time its invoked.

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: Deploys and upgrades are treated the same way - if the cache entry exists, it's queued for reverification. If it doesn't, the program cache will verify and add it to the cache the next time it's invoked. I take advantage that the program execution behavior can only change if the program is deployed or upgraded (or closed even, but BPF / account loading checks will catch cases where someone tries invoking a closed program account), so I don't handle the case where a program may be closed / retracted. Of course this is liable to DOS vectors as discussed in the comments above but we should implement an eviction policy when the cache lives outside of funk.

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: done

src/flamenco/runtime/program/fd_program_cache.h
ripatel-fd: Ok I see, so 0 is some sentinel value that prevents re-verification. Unclear to me how this works with genesis program accounts

src/flamenco/runtime/program/fd_program_cache.h
ripatel-fd: @mjain-jump Could you update the comment to remove this DoS argument?

src/flamenco/runtime/program/fd_program_cache.h
ripatel-fd: @mjain-jump Yes. But what if that mechanism breaks? Do we have layers of safety in case the BPF loader check erroneously permits an execution of a recently modified program? I'm just curious what's going to happen.

src/flamenco/runtime/program/fd_program_cache.h
ripatel-fd: @mjain-jump Could you update the docs to mention to clarify there's no distinction between "deploy" and "upgrade" then, but that it's purely about local state of the cache? Deploy and upgrade are BPF Loader interface APIs and the cache update mechanism you described seems to be orthogonal.

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: AFAIK the only genesis programs are native programs (unless I'm mistaken?), which will not live in the cache

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: done

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: If the program is freshly deployed, the cache entry wouldn't be inserted until AT MINIMUM the next slot, so the cache query would fail in the BPF loader. If the program is upgraded, then the program cache entry will be stale (containing old ELF / sBPF information) but will still execute as normal using metadata from the old cache entry.

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: Sure. I forgot to update the doc for this too, but we only queue programs for reverification upon calls to `fd_deploy_program()` which are called from program upgrades / deploys /extends / etc

src/flamenco/runtime/program/fd_program_cache.h
ibhatt-jumptrading: You can have non-native programs at genesis. This is how we did the nano-token breakpoint demo

src/flamenco/runtime/program/fd_bpf_loader_program.c
```diff
@@ -251,6 +276,10 @@ fd_deploy_program( fd_exec_instr_ctx_t * instr_ctx,
...
+  instr_ctx->txn_ctx->programs_to_reverify[instr_ctx->txn_ctx->programs_to_reverify_cnt++] = *program_key;
```
ibhatt-jumptrading, line 280: if we are deploying or upgrading a program why do we need to actually reverify the program? Don't we do all of the checks in the loader program? Couldn't we skip the verification in this case and just save the slot + bytecode

src/flamenco/runtime/program/fd_program_cache.h
```diff
@@ -0,0 +1,284 @@
...
+   set `last_slot_verified` to the current slot, and set
+   `last_slot_modified` to 0. A future program upgrade / network feature
+   set change could make this program valid again.
+
+   A key invariant with the program cache design is that it does not
+   evict any entries - it only grows through the lifetime of the
+   client's execution. Because of this invariant, we do not touch cache
+   entries for programs which are closed, and instead let account
+   loading checks catch cases where a user tries to invoke a closed
+   program.
```
ibhatt-jumptrading: to be clear, this check happens in the bpf loader though right?

ibhatt-jumptrading (commented):
one outstanding question i have is:

how are we going to cleanly handle resolving an ALUT in replay tile. Will we propagate it to the exec tiles so we don't do it twice? How does it fit into the new dispatcher?

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: It happens in both the transaction-level account loading and BPF loader program checks. I'll update this comment to make that clear

src/flamenco/runtime/program/fd_bpf_loader_program.c
mjain-jump: wouldnt that require propagating all of the ELF + sbpf program information (calldests, offsets, entry pc, etc) back up the call stack all the way to the writer tile? it would also require more special-casing since the program is only reverified the next time it is invoked (if invoked in the same epoch it was modified, then we would skip reverification, otherwise we would need to reverify anyways)

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: At genesis the cache will be empty, so when someone invokes those programs, they will be added to the cache. To trigger this edge case, someone would have to invoke a genesis BPF program at slot 0 that fails verification. In this very specific case, the program would get reverified again (since the last modified slot == last reverified slot == 0) the very next time the program gets invoked in a different slot, and then the behavior is the same as any other program after that.

src/flamenco/runtime/fd_runtime.h
mjain-jump: oops, fixed

src/flamenco/runtime/program/fd_program_cache.h
ripatel-fd: umm ... sounds good :) Could we add a unit test for this?

src/flamenco/runtime/program/fd_program_cache.h
ripatel-fd: Thanks for clarifying

src/flamenco/runtime/program/fd_program_cache.h
mjain-jump: done

yufeng-jump (approved):

PR #5857, opened 2025-07-29
use ppoll instead of poll in sock and bundle tile

anwayde:


src/util/fd_util.c
```diff
@@ -25,6 +27,21 @@ _fd_tickcount( void const * _ ) {
...
+  if( timeout<0 ) {
+    return ppoll( fds, nfds, NULL, NULL );
+  } else {
+    struct timespec ts = {
+      .tv_sec  = (long)( timeout/1000 ),
+      .tv_nsec = (long)((timeout%1000)*1000000),
+    };
+    return ppoll( fds, nfds, &ts, NULL );
+  }
+}
```
ripatel-fd, line 54: Guard this with FD_HAS_HOSTED and `defined(__linux__)`. 

like so:
- FD_HAS_HOSTED && linux -> ppoll
- FD_HAS_HOSTED && not linux -> fall back to default
- not FD_HAS_HOSTED -> don't generate function

src/util/fd_util.h
```diff
@@ -1,6 +1,8 @@
 #ifndef HEADER_fd_src_util_fd_util_h
 #define HEADER_fd_src_util_fd_util_h
 
+#include <poll.h>
```
ripatel-fd: No need to import this. You can have an incomplete type if it's only a pointer.

src/util/fd_util.h
```diff
@@ -297,6 +299,11 @@ fd_boot( int *    pargc,
 void
 fd_halt( void );
 
+int
+fd_syscall_poll( struct pollfd * fds,
+                 nfds_t          nfds,
+                 int             timeout );
```
ripatel-fd: guard with FD_HAS_HOSTED

src/util/fd_util.h
```diff
@@ -297,6 +299,11 @@ fd_boot( int *    pargc,
 void
 fd_halt( void );
 
+int
+fd_syscall_poll( struct pollfd * fds,
```
ripatel-fd, line 316: Add a comment why this is necessary. Important points to include:
- The Firedancer sandbox whitelists syscalls
- Depending on glibc version, calling the standard library `poll` function results either in the `poll` or `ppoll` syscall
- This wrapper always calls the `ppoll` syscall under the hood on Linux

ripatel-fd: Also document that this sets errno

src/util/fd_util.h
anwayde: I still needed it, otherwise build was failing

ripatel-fd: @anwayde This is needed for the function definition but not for the declaration. Please use a forward declaration to avoid this poll import. e.g. `struct pollfd;`

anwayde: ok got it, removed the import

PR #5863, opened 2025-07-30
Add staking reward events to solcap

ripatel-fd:


ripatel-fd:
Still need to extend fd_solcap_yaml

kbhargava-jump (approved):

PR #5865, opened 2025-07-30
backtest: various improvements

mmcgee-jump:
- Improves the performance of backtest by about 40%
- Fixes two memory corruption issues
- Improves default logging levels and output of the backtest tool to be readable
- Moves most of the manifest loading to be from the snapshot tile manifest, rather than the fd_types decoder, in preparation for removing the fd_types decode which will save another 20%


src/app/firedancer-dev/commands/backtest.c
```diff
@@ -323,45 +327,137 @@ backtest_topo( config_t * config ) {
...
+            (double)( acc_cnt-acc_cnt_old  )/1e6 );
+    fflush( stdout );
+    total_off_old    = total_off;
+    snaprd_backp_old = snaprd_backp;
+    snaprd_wait_old  = snaprd_wait;
+    snapdc_backp_old = snapdc_backp;
+    snapdc_wait_old  = snapdc_wait;
+    snapin_backp_old = snapin_backp;
+    snapin_wait_old  = snapin_wait;
+    acc_cnt_old      = acc_cnt;
```
ripatel-fd, line 446: Let's not copy paste this metrics code

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -323,45 +327,137 @@ backtest_topo( config_t * config ) {
...
+  return args;
+}
+
+void
+backtest_cmd_perm( args_t *         args FD_PARAM_UNUSED,
+                   fd_cap_chk_t *   chk,
+                   config_t const * config ) {
+  args_t c_args = configure_args();
+  configure_cmd_perm( &c_args, chk, config );
+  run_cmd_perm( NULL, chk, config );
```
ripatel-fd, line 360: No, this is broken. backtest SHOULD NOT have socket(2) or bpf(2) permissions.

src/disco/topo/fd_topo.c
```diff
@@ -136,7 +136,12 @@ fd_topo_wksp_new( fd_topo_t const *          topo,
 
     for( ulong j=0UL; callbacks[ j ]; j++ ) {
       if( FD_LIKELY( strcmp( callbacks[ j ]->name, obj->name ) ) ) continue;
+
+      long ts = -fd_log_wallclock();
       if( FD_LIKELY( callbacks[ j ]->new ) ) callbacks[ j ]->new( topo, obj );
+      long elapsed = fd_log_wallclock() + ts;
+      if( FD_UNLIKELY( elapsed>(1000L*1000L*100L ) ) ) FD_LOG_WARNING(( "fd_topo_wksp_new(%s) took %ld ms", obj->name, elapsed/(1000L*1000L) ));
+      if( FD_UNLIKELY( elapsed>(1000L*1000L*5L   ) ) ) FD_LOG_INFO   (( "fd_topo_wksp_new(%s) took %ld ms", obj->name, elapsed/(1000L*1000L) ));
```
ripatel-fd, line 144: Typical code style would be (long)100e6

src/disco/topo/fd_topo_run.c
```diff
@@ -318,7 +320,7 @@ run_tile_thread( fd_topo_t *         topo,
     }
   }
 
-  fd_topo_run_thread_args_t args = {
+  *args = (fd_topo_run_thread_args_t){
```
ripatel-fd:
```suggestion
  *args = (fd_topo_run_thread_args_t) {
```

src/disco/topo/fd_topob.c
```diff
@@ -359,7 +359,7 @@ fd_topob_auto_layout( fd_topo_t * topo,
     "pack",
     "bank",   /* FRANK only */
     "poh",    /* FRANK only */
-    "poh",   /* FIREDANCER only */
+    "poh",    /* FIREDANCER only */
```
ripatel-fd: Why is the same exact tile twice in this list?  Was this meant to say `pohi`?

src/discof/backtest/fd_backtest_tile.c
```diff
@@ -474,18 +488,17 @@ after_credit( ctx_t *             ctx,
...
-             ulong                             sz,
-             ulong                             ctl FD_PARAM_UNUSED ) {
-  FD_TEST( in_idx==0 );
+             ulong   in_idx,
+             ulong   seq FD_PARAM_UNUSED,
+             ulong   sig FD_PARAM_UNUSED,
+             ulong   chunk,
+             ulong   sz,
+             ulong   ctl FD_PARAM_UNUSED ) {
+  FD_TEST( !in_idx );
```
ripatel-fd, line 501:
```suggestion
  FD_TEST( in_idx==0 );
```

Revert this, it was more readable before. No need to use boolean syntax for index

src/app/firedancer-dev/commands/backtest.c
ripatel-fd: https://github.com/firedancer-io/firedancer/pull/5868

mmcgee-jump: It's a development command that's going to elevate sudo anyway. It needs all of the other permissions from run, who cares if it also has CAP_NET_RAW? It's bugged right now. Your fix which you ported out of here is not correct.

src/app/firedancer-dev/commands/backtest.c
mmcgee-jump: Doesn't matter

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -323,45 +327,137 @@ backtest_topo( config_t * config ) {
...
+    printf( "bw=%4.0f MB/s backp=(%3.0f%%,%3.0f%%,%3.0f%%) busy=(%3.0f%%,%3.0f%%,%3.0f%%) acc=%3.1f M/s\n",
+            (double)( total_off-total_off_old )/1e6,
+            ( (double)( snaprd_backp-snaprd_backp_old )*ns_per_tick )/1e7,
+            ( (double)( snapdc_backp-snapdc_backp_old )*ns_per_tick )/1e7,
+            ( (double)( snapin_backp-snapin_backp_old )*ns_per_tick )/1e7,
+            100-( ( (double)( snaprd_wait-snaprd_wait_old  )*ns_per_tick )/1e7 ),
+            100-( ( (double)( snapdc_wait-snapdc_wait_old  )*ns_per_tick )/1e7 ),
+            100-( ( (double)( snapin_wait-snapin_wait_old  )*ns_per_tick )/1e7 ),
+            (double)( acc_cnt-acc_cnt_old  )/1e6 );
+    fflush( stdout );
```
topointon-jump, line 438: we should probably remove this printf before merging

src/app/firedancer-dev/commands/backtest.c
ripatel-fd: It does, blocking

src/app/firedancer-dev/commands/backtest.c
ripatel-fd: I care because this command is essential for runtime development. This development tooling needs to be robust and usable for other core devs, like Anza. Anza is not going to touch our tooling if it randomly reconfigures their NIC when replaying a ledger locally.

src/app/firedancer-dev/commands/backtest.c
mmcgee-jump: No, it's there by design

src/app/firedancer-dev/commands/backtest.c
mmcgee-jump: What does that have to do with run permissions? It is deciding whether to elevate to sudo or not. The correct fix is to copy paste the run permission lines into here and remove like, one or two of them, but it actually will never have any effect so I didn't do it (the other permissions are more stringent than the ones which you could "remove" so it never does anything).

src/app/firedancer-dev/commands/backtest.c
mmcgee-jump: No, it doesn't. It's going to introduce unnecessary coupling between two completely different consumers of that data. I am already close to changing the way this one works to be subtly different.

src/flamenco/runtime/fd_blockstore.c
```diff
@@ -46,7 +46,7 @@ fd_blockstore_new( void * shmem,
     FD_LOG_WARNING(( "blockstore implementation requires shred_max to be a power of two, rounding it up to %lu", shred_max ));
   }
 
-  fd_memset( blockstore_shmem, 0, fd_blockstore_footprint( shred_max, block_max, idx_max ) );
+  fd_memset( blockstore_shmem, 0, sizeof( fd_blockstore_shmem_t ) );
```
mmcgee-jump, line 49: Performance tweak, saves about 1 second of memsetting on boot

src/flamenco/runtime/fd_bank.c
```diff
@@ -195,12 +195,10 @@ fd_banks_new( void * shmem, ulong max_total_banks, ulong max_fork_width ) {
 
   /* Need to layout all of the CoW pools. */
   #define HAS_COW_1_LIMIT_1(name) \
-    void * name##_pool_mem = FD_SCRATCH_ALLOC_APPEND( l, fd_bank_##name##_pool_align(), fd_bank_##name##_pool_footprint( max_fork_width ) ); \
-    memset( name##_pool_mem, 0, fd_bank_##name##_pool_footprint( max_fork_width ) );
+    void * name##_pool_mem = FD_SCRATCH_ALLOC_APPEND( l, fd_bank_##name##_pool_align(), fd_bank_##name##_pool_footprint( max_fork_width ) );
```
mmcgee-jump, line 198: Performance tweak, saves 6 seconds on boot not pointlessly zeroing memory

src/discof/backtest/fd_backtest_tile.c
```diff
@@ -633,6 +646,8 @@ after_frag( ctx_t *             ctx,
 fd_topo_run_tile_t fd_tile_backtest = {
   .name                     = "back",
   .loose_footprint          = loose_footprint,
+  .scratch_align            = scratch_align,
```
mmcgee-jump, line 649: This was buggy and would randomly corrupt

src/disco/topo/fd_topo_run.c
```diff
@@ -353,13 +353,23 @@ fd_topo_run_single_process( fd_topo_t *       topo,
...
-    run_tile_thread( topo, tile, run_tile, uid, gid, floating_cpu_set, save_priority );
+    run_tile_thread( topo, tile, run_tile, uid, gid, floating_cpu_set, save_priority, &args[ i ] );
+  }
+
+  for( ulong i=0UL; i<topo->tile_cnt; i++ ) {
+    fd_topo_tile_t * tile = &topo->tiles[ i ];
+    if( !agave && tile->is_agave ) continue;
+    if( agave==1 && !tile->is_agave ) continue;
+
+    while( !FD_VOLATILE( args[ i ].copied ) ) FD_SPIN_PAUSE();
```
mmcgee-jump, line 372: Minor perf tweak maybe ~500ms, parallel creating all the threads

src/app/shared/commands/run/run.c
```diff
@@ -653,7 +666,12 @@ initialize_stacks( config_t const * config ) {
 
     ulong sub_page_cnt[ 1 ] = { 6 };
     ulong sub_cpu_idx [ 1 ] = { stack_cpu_idx };
-    int err = fd_shmem_create_multi( name, FD_SHMEM_HUGE_PAGE_SZ, 1, sub_page_cnt, sub_cpu_idx, S_IRUSR | S_IWUSR ); /* logs details */
+    int err;
+    if( FD_UNLIKELY( update_existing ) ) {
```
mmcgee-jump, line 670: Perf tweak to reuse existing stack pages saves maybe 500ms+

src/app/firedancer-dev/commands/backtest.c
ripatel-fd: Why would we need this tool to elevate to sudo? If hugetlbfs and snapshot dirs exist and have the correct permissions, run it without sudo. This is how all other unit tests / tools outside fdctl behave too. Otherwise, run with superuser.

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -119,7 +120,10 @@ backtest_topo( config_t * config ) {
   fd_topob_wksp( topo, "snapin_rd" );
   fd_topob_wksp( topo, "snap_out" );
   fd_topob_wksp( topo, "replay_manif" );
-  fd_topob_link( topo, "snap_out", "snap_out", 2UL, 5UL*(1UL<<30UL), 1UL );
+  /* TODO: Should be depth of 1 or 2, not 4, but it causes backpressure
+     from the replay tile parsing the manifest, remove when this is
+     fixed. */
+  fd_topob_link( topo, "snap_out", "snap_out", 4UL, 5UL*(1UL<<30UL), 1UL );
```
mmcgee-jump, line 126: Perf tweak prevents backpressure on snapin for now (will be fixed later after removing the manifest decoder). Saves 5+ seconds

src/app/firedancer-dev/commands/backtest.c
mmcgee-jump: You might not have RLIMIT_MLOCK for the workspaces, for example, which it needs to elevate to raise.

mmcgee-jump: I'd agree if it was a separate test binary, but this is part of `fddev` (I don't know if that's a good idea), and it's an entire design point of `fddev` is that it elevates for you as one-shot runner

0x0ece (approved):

PR #5867, opened 2025-07-30
runtime: remove unused snapshot hasher

ripatel-fd:
This code is untested and unused


ripatel-fd:
Nvm, this needs a ledger rewrite. Which Tom is working on, I believe.

topointon-jump:
@ripatel-fd this code will be removed in my hashes pr

PR #5869, opened 2025-07-30
`_slot_end_ns` unused; now read from became_leader_t

cavemanloverboy:


src/disco/pack/fd_pack_tile.c
```diff
@@ -1,4 +1,4 @@
-#include "../tiles.h"
+end_#include "../tiles.h"
```
ripatel-fd: .end

mmcgee-jump:
Thanks! Can you squash commits?

cavemanloverboy:
> Thanks! Can you squash commits?

will do

PR #5879, opened 2025-07-30
update gui and metric tiles to use ppoll

anwayde:
and rpcserv tile as well


src/waltz/http/fd_http_server.h
```diff
@@ -423,9 +423,9 @@ int
 fd_http_server_ws_broadcast( fd_http_server_t * http );
 
 /* fd_http_server_poll needs to be continuously called in a spin loop to
-   drive the HTTP server forward.  poll_timeout is the timeout arg to
-   poll(2).  poll_timeout==0 does a non-blocking socket poll.   Returns
-   1 if there was any work to do on the HTTP server, or 0 otherwise. */
+   drive the HTTP server forward.  Setting poll_timeout==0 makes it a
+   non-blocking socket ppoll(2).   Returns 1 if there was any work to do
```
ripatel-fd, line 427: Shouldn't it return the number of file descriptors for which events are ready?

anwayde: That's the return for `fd_syscall_poll` right? This is the return value for `fd_http_server_poll`.

PR #5888, opened 2025-07-31
funk: it's data race season

yufeng-jump:
We hit race conditions during devnet and mainnet offline replay.  This commit fixes an inproper usage of lock-free query_try() in fd_funk_txn_valid().  This caused fd_funk_txn_valid() to incorrectly and transiently return false on a valid funk txn due to ERR_AGAIN.  This could happen if the funk txn being queried happens to be on the same map chain as another funk txn that is being concurrently modified, for example due to the replay tile rooting a slot in the past.  Or it could simply be due to a concurrent entity acquiring the exact same funk txn.  The latter shouldn't happen and would have been an indication of a bug at a higher level.  The former we likely did observe in devnet offline replay.  Slot 396129199 and 396129167 both land on funk txn map chain idx 30 given a funk txn map size of 32.  Slot 396129167 likely was being concurrently rooted while slot 396129199 was executing.

### DIFF
--- a/src/flamenco/Local.mk
+++ b/src/flamenco/Local.mk
@@ -5,3 +5,4 @@ $(call make-unit-test,test_flamenco,test_flamenco,fd_flamenco fd_ballet fd_util)
 $(call run-unit-test,test_flamenco)
 endif
 $(call add-hdrs,fd_rwlock.h)
+$(call add-hdrs,fd_rwlock_recursive.h)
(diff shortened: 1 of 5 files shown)


ripatel-fd:
Thanks for the patch. I hope we can remove the need for this soon enough with the new database tile.

ibhatt-jumptrading (approved):

PR #5893, opened 2025-08-01
contrib: tool for converting agave flags to FD toml

anwayde:


contrib/toml-gen.py
```diff
@@ -0,0 +1,357 @@
...
+import sys
+import toml
+import logging
+import argparse
+
+from pprint import pprint
+
+# regex patterns for parsing files
+pattern1 = r'ADD.?\(\s*"([^"]*)"\s*(?:,\s*([^)]*))?\s*\)'                           # ADD macros
+pattern2 = r'if\(\s*(?:FD_(?:LIKELY|UNLIKELY)\(\s*)?!?(.*?)(?:\s*\)\s*\)|\s*\))'    # if statements for boolean flags
```
SEJeff: By making these globals and calling `re.search()` in a for loop, you're compiling the regex over and over. There is a regex cache in python to make this less painful, but better to do:

```suggestion
pattern1 = re.compile(r'ADD.?\(\s*"([^"]*)"\s*(?:,\s*([^)]*))?\s*\)')                           # ADD macros
pattern2 = re.compile(r'if\(\s*(?:FD_(?:LIKELY|UNLIKELY)\(\s*)?!?(.*?)(?:\s*\)\s*\)|\s*\))')    # if statements for boolean flags
```

And update the code elsewhere to change `re.search()` --> `match.match()`

contrib/toml-gen.py
```diff
@@ -0,0 +1,357 @@
...
+            if not line.endswith('\\'):
+                in_cmdline = False
+
+    return cmdline
+
+# parse a agave-validator commandline to get a list of (flag, arg)
+def parse_cmdline(flags_string):
+    flags = []
+    flag  = ()
+    for line in flags_string.split('\n'):
```
SEJeff:
```suggestion
    for line in flags_string.splitlines()
```

contrib/toml-gen.py
```diff
@@ -0,0 +1,357 @@
...
+FD_DEFAULT_TOML = "src/app/fdctl/config/default.toml"
+FD_AGAVE_FILE   = "src/app/fdctl/commands/run_agave.c"
+FD_CONFIG_FILE  = "src/app/shared/fd_config_parse.c"
+
+# generated maps
+flag_config_map = {} #  --agave-flag: config.option
+config_type_map = {} # config.option: data_type
+
+#    --agave-flags                          ->      confg.option
+hardcoded_flag_config_map = {
```
SEJeff: This is effectively a constant and should be uppercased.

contrib/toml-gen.py
```diff
@@ -0,0 +1,357 @@
...
+                if len(flag) == 1 and flag[0].startswith('--'):
+                    flag += (True,)
+                if flag: flags.append(flag)
+                flag = (sword,)
+            elif not sword.startswith('--'):
+                # argument for the currently parsing flag
+                flag += (sword,)
+            else:
+                logging.error(f"unrecognized token while parsing - {sword}")
+                sys.exit(1)
```
SEJeff:
```suggestion
                raise SystemExit(f"unrecognized token while parsing - {sword}")
```

Does exactly the same thing with less code

contrib/toml-gen.py
```diff
@@ -0,0 +1,357 @@
...
+            else:
+                config_type_map[full_key] = type(value).__name__
+
+# parse FD_AGAVE_FILE lines to build flag_config_map
+def build_flag_config_map(lines):
+    for i in range(len(lines)):
+        line  = lines[i].strip()
+        # find ADD macro
+        match = re.search(pattern1, line)
+        if match:
```
SEJeff: For big indented blocks, consider inversing the logic here. More like:

```
if not match:
    continue
```

Then you can left indent the code in the entire rest of the function and have less nested logic.

contrib/toml-gen.py
```diff
@@ -0,0 +1,357 @@
...
+    arguments = argini()
+
+    logging.info("building maps")
+    generate_maps()
+
+    if arguments.detect_mismatch:
+        if hardcoded_flag_config_map != flag_config_map:
+            logging.error("detected difference between config maps")
+            # shows a mismatch in keys between the hardcoded and parsed maps
+            if len(hardcoded_flag_config_map.keys()) > len(flag_config_map.keys()):
```
SEJeff:
```suggestion
            if len(hardcoded_flag_config_map) > len(flag_config_map):
```


```python
>>> map1 = {1:1, 2:2, 3:3}
>>> len(map1.keys())
3
>>> len(map1)
3
```

PR #5895, opened 2025-08-01
agave: make sure only fdctl can be used to change identity

anwayde:
The rebase is on top of https://github.com/firedancer-io/firedancer/pull/5873, so will wait for that to get merged before merging this.


anwayde:
Included the change in https://github.com/firedancer-io/firedancer/pull/5873

PR #5896, opened 2025-08-01
repair: fix peer cold start

npatel-jump:
Ping peers as we see them to bring repair start time down + hacky 2x speed up on after_credit issue (will dissapear after I make sign tile fix soon)


src/discof/repair/fd_repair_tile.c
```diff
@@ -424,6 +399,36 @@ fd_repair_send_requests( fd_repair_tile_ctx_t *   ctx,
...
+     consider an eviction policy. */
+  for( ulong i=0UL; i<dest_cnt; i++ ) {
+   if( FD_UNLIKELY( ctx->repair->peer_cnt >= FD_ACTIVE_KEY_MAX ) ) break;// FIXME: aiming to move all peer tracking out of lib into tile, leaving like this for now
+    fd_repair_peer_addr_t repair_peer = {
+      .addr = in_dests[i].ip4_addr,
+      .port = fd_ushort_bswap( in_dests[i].udp_port ),
+    };
+    int dup = fd_repair_add_active_peer( ctx->repair, &repair_peer, in_dests[i].pubkey );
+    if( !dup ) {
+      fd_repair_send_request(ctx, ctx->stem, ctx->repair, 0, 0, 0, in_dests[i].pubkey, fd_log_wallclock());
```
lidatong, line 610: hm is this the only line changing? can you move the function back so it's a cleaner diff?

src/discof/repair/fd_repair_tile.c
```diff
@@ -974,6 +979,13 @@ after_credit( fd_repair_tile_ctx_t * ctx,
               fd_stem_context_t *    stem FD_PARAM_UNUSED,
               int *                  opt_poll_in FD_PARAM_UNUSED,
               int *                  charge_busy ) {
+
+  // temporary starving while fixing signing -- just helps development speed by a lot for right now
+  ctx->loop_idx++;
+  if ((ctx->loop_idx % 20) != 0) {
+    return;
+  }
```
lidatong: are you intentionally including in PR into main?

npatel-jump: That was my plan, it makes repair MUCH faster, but only for a couple days while I clean the rest up

So if anyone is rebasing on this, they would be able to fly through repair without any issues and increase development speed, but this will get thrown out after.

src/discof/repair/fd_repair_tile.c
npatel-jump: It was a function decleration of fd_repair_send_requests being invalid, my options were either change the ordering, or declare a prototype

lidatong: could you also add a block comment / blurb for what the send request 0, 0, 0 is doing?

src/discof/repair/fd_repair_tile.c
lidatong: discussed offline and going to remove this in lieu of a proper fix in a subsequent PR

src/discof/repair/fd_repair_tile.c
npatel-jump: added

src/discof/repair/fd_repair_tile.c
```diff
@@ -424,6 +397,40 @@ fd_repair_send_requests( fd_repair_tile_ctx_t *   ctx,
...
+    fd_repair_peer_addr_t repair_peer = {
+      .addr = in_dests[i].ip4_addr,
+      .port = fd_ushort_bswap( in_dests[i].udp_port ),
+    };
+    int dup = fd_repair_add_active_peer( ctx->repair, &repair_peer, in_dests[i].pubkey );
+    if( !dup ) {
+      /*
+      Peer warm up. Peers send ping-pong messages on first repair request. Due to a design flaw in the iterator, we wait 40ms between each peer during warm up.
+      This is fix to workaround the iterator and decrease warm up time from ~40ms * num_peers to ~ time until first turbine. (about 50 seconds decrease in startup time)
+      */
```
lidatong: nit, but can you reformat to be a block comment capped out 72 columns formatted similarly to this?

https://github.com/firedancer-io/firedancer/blob/emwang/feat/store-pool-para/src/choreo/tower/fd_tower.c#L391-L394

lidatong: also, you can more generically address what the code is doing -- maybe rephrase to "Repair involves a Ping-Pong protocol that requires 1 RTT on the first repair request, so as an optimization we immediately send a dummy Repair request upon getting a peer's contact info for the first time to pre-pay that round-trip"

src/discof/repair/fd_repair_tile.c
```diff
@@ -424,6 +397,43 @@ fd_repair_send_requests( fd_repair_tile_ctx_t *   ctx,
...
+    };
+    int dup = fd_repair_add_active_peer( ctx->repair, &repair_peer, in_dests[i].pubkey );
+    if( !dup ) {
+      /*
+      The repair process uses a Ping-Pong protocol that incurs one
+      round-trip time (RTT) for the initial repair request. To optimize
+      this, we proactively send a placeholder Repair request as soon as
+      we receive a peer's contact information for the first time,
+      effectively prepaying the RTT cost.
+      */
```
lidatong: sorry, but it still doesn't look quite right. keep it on the same line as the `/*` and align to at most 72-columns

```suggestion
/* The repair process uses a Ping-Pong protocol that incurs one
   round-trip time (RTT) for the initial repair request. To optimize
   this, we proactively send a placeholder Repair request as soon as we
   receive a peer's contact information for the first time, effectively
   prepaying the RTT cost. */
```

PR #5905, opened 2025-08-01
added sorted list template

asiegel-jt:


ripatel-fd:
@asiegel-jt What are you planning to use this template for?

PR #5892, opened 2025-07-31
snapshot: parse versioned epoch stakes

0x0ece:

PR #5907, opened 2025-08-01
firedancer-dev: repair profiler light - update

jvarela-jump:
Upgrading the repair profiler light tool.
It now uses *fd_ssmanifest_parser_t* and *fd_snapshot_manifest_t* to retrieve the required data from the manifest. 
This PR is dependent on  https://github.com/firedancer-io/firedancer/pull/5892 (which should be merged first).

### DIFF
--- a/src/discof/restore/utils/fd_ssmsg.h
+++ b/src/discof/restore/utils/fd_ssmsg.h
@@ -7,8 +7,13 @@
 #define FD_SSMSG_MANIFEST_INCREMENTAL (1) /* A snapshot manifest message from the incremental snapshot */
 #define FD_SSMSG_DONE                 (2) /* Indicates the snapshot is fully loaded and tiles are shutting down */
 
-/* TODO: Bound this correctly */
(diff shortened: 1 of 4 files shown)


src/discof/restore/utils/fd_ssmsg.h
```diff
@@ -212,7 +212,7 @@ struct fd_snapshot_manifest {
      Currently, these parameters can never change and are fixed from
      genesis onwards, although in future they may change with new
      feature flags. */
-  fd_snapshot_manifest_epoch_schedule_params_t epoch_schedule_params;
+  fd_epoch_schedule_t epoch_schedule_params;
```
lidatong: i would not change the ssmsg schema here (believe snapshot team intentionally made a separate type) and instead do a type pun cast at the usage site in publish_stake_weights

lidatong: as at first glance it looks like these types are binary compatible

lidatong (approved):

PR #5908, opened 2025-08-02
Wiredancer F2 Support

monological:
Added Wiredancer support for the aws f2 platform.


ripatel-fd:
Thank you very much for the contribution @monological. We'd love to upstream this, would you mind reopening the PR?

PR #5909, opened 2025-08-02
quic_trace: fix conn_at_idx

akhinvasara-jumptrading:
quic_trace used the quic_conn_at_idx helper, which relied on state->conn_base. But that address was in the quic_tile's address space, not translated to the local address space. This PR fixes the resulting segfault.

### DIFF
--- a/src/app/shared_dev/commands/quic_trace/fd_quic_trace_main.c
+++ b/src/app/shared_dev/commands/quic_trace/fd_quic_trace_main.c
@@ -121,10 +121,10 @@ dump_quic_config( fd_quic_config_t * config ) {
 }
 
 static char const *
-peer_cid_str( fd_quic_conn_t * conn ) {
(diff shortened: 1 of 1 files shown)


src/waltz/quic/fd_quic_private.h
```diff
@@ -156,7 +156,7 @@ FD_PROTOTYPES_BEGIN
    pointer to fd_quic_t.  Const func, guaranteed to not access memory. */
 
 FD_FN_CONST static inline fd_quic_state_t *
-fd_quic_get_state( fd_quic_t * quic ) {
+fd_quic_get_state( fd_quic_t const * quic ) {
   return (fd_quic_state_t *)( (ulong)quic + FD_QUIC_STATE_OFF );
 }
```
ripatel-fd: Const violation

PR #5911, opened 2025-08-04
store: fix under allocation due to rounding-up after footprint calcul…

two-heart:
…ation

otherwise for a non power of two fec_max, store_map will be zero, because it is used unchecked in fd_store_footprint. Also shele will be under-allocated.


lidatong:
thanks for catching - im going to go a different way here and just document that the caller must pass in a value that's pow2

two-heart:
I included that fixed in the backlinked PR. I also like the document and check way better.

PR #5918, opened 2025-08-04
shred: keep track of parent block_id when we are leader

0x0ece:
Changes:
1. Removed `bank.set_block_id()` in Agave (I previously added it when first built chained merkle shreds)
2. Added block_ids table to shred tile to keep track of leader's block_ids

Tested in lab, running 2 validators both with stake. To simulate skips, we set-identity on one of the 2 validators so that it wouldn't produce slots. We see the correct behavior of the other validator skipping and fetching the parent block id from the new table.

### DIFF
--- a/src/discof/poh/fd_poh_tile.c
+++ b/src/discof/poh/fd_poh_tile.c
@@ -1255,8 +1255,6 @@ fd_ext_poh_reset( ulong         completed_bank_slot, /* The slot that successful
   if( FD_LIKELY( parent_block_id!=NULL ) ) {
     ctx->parent_slot = completed_bank_slot;
     memcpy( ctx->parent_block_id, parent_block_id, 32UL );
-  } else {
-    FD_LOG_WARNING(( "fd_ext_poh_reset(block_id=null,reset_slot=%lu,parent_slot=%lu) - ignored", completed_bank_slot, ctx->parent_slot ));
   }
   ctx->slot         = completed_bank_slot+1UL;
   ctx->hashcnt      = 0UL;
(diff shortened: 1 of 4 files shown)


src/discof/poh/fd_poh_tile.c
```diff
@@ -1395,6 +1393,8 @@ publish_tick( fd_poh_ctx_t *      ctx,
   if( FD_LIKELY( meta->parent_block_id_valid ) ) {
     fd_memcpy( meta->parent_block_id, ctx->parent_block_id, 32UL );
   }
+  meta->parent_by_same_leader = (slot % 4) != 0
```
mmcgee-jump: Why does this need to be arbitrated by PoH? Can't you just determine this in the shred tile?

src/discoh/poh/fd_poh_tile.c
```diff
@@ -1399,6 +1397,8 @@ publish_tick( fd_poh_ctx_t *      ctx,
   if( FD_LIKELY( meta->parent_block_id_valid ) ) {
     fd_memcpy( meta->parent_block_id, ctx->parent_block_id, 32UL );
   }
+  meta->parent_by_same_leader = (slot % 4) != 0
```
mmcgee-jump: I don't like how the Agave code calls `bank.set_block_id` for our leader slots, that's violating the Agave invariant of no block id on leader slots, you should just pass it directly through to fd_ext_poh_reset as a function argument

0x0ece: Yeah I agree. Let me verify/test, but I think when I added the set_block_it I thought that it'd always be available, then discovered some cases where it wasn't.

Now I think going with the invariant "it's always available when non-leader, we won't need it when we're leader" is the best strategy, so I'll try to remove that block of code from Agave.

mmcgee-jump (approved):

Issue #5231, opened 2025-05-30
[types] Replace StaticVector, Vector, and BitVector with first class types


cali-jumptrading:

ripatel-fd:
@cali-jumptrading Looks like this was done?

cali-jumptrading:
> [@cali-jumptrading](https://github.com/cali-jumptrading) Looks like this was done?

No, this doesn't look to be done. What I meant by this issue is that we should have first class types support for `Vector`, `BitVector`, and `Maps` types, instead of approximating these types with pointers, lengths and red-black trees

moldis:
On behave of Noders team [https://noders.team] requesting permissions to raise PR

PR #5920, opened 2025-08-05
types: Replace BitVector with first class types

moldis:
Replacing BitVector with first type class, partially closing https://github.com/firedancer-io/firedancer/issues/5231


cali-jumptrading:
Hey, we're planning to redesign `fd_types` completely, so we're not accepting external changes to `fd_types` right now. Thanks for your contribution though!

PR #5921, opened 2025-08-05
store: don't ignore fd_store_map_verify ret val in fd_store_verify

two-heart:


src/disco/store/fd_store.c
```diff
@@ -281,8 +281,7 @@ fd_store_verify( fd_store_t * store ) {
     }
     ele_cnt++;
   }
-  fd_store_map_verify( map, ele_cnt, fec0 );
-  return 0;
```
emwang-jump, line 285: good catch, but looking at this realized the `ele_cnt` getting passed is also wrong, if you could add a change to the PR to do `fd_store_map_verify( map, store->fec_max, fec0 );`

emwang-jump (changes requested):

emwang-jump (approved):

PR #5923, opened 2025-08-05
offline-replay: update runtime limits

alpeng-jump:


ibhatt-jumptrading (commented):
we can probably bump down the max_total_banks in offline replay, sorry for forgetting to update this script though...

contrib/offline-replay/offline_replay.toml
```diff
@@ -22,7 +22,8 @@
 [runtime]
     heap_size_gib = {heap_size}
     [runtime.limits]
-        max_banks = 64
+        max_total_banks = 128
```
kbhargava-jump: Both of these can be set to 4

PR #5926, opened 2025-08-05
flamenco, program-cache: allocate correct size for modified funk records

mjain-jump:


src/flamenco/runtime/program/fd_program_cache.c
```diff
@@ -48,6 +67,18 @@ fd_program_cache_entry_footprint( fd_sbpf_elf_info_t const * elf_info ) {
...
 
+/* Returns the footprint of a record that failed verification. Note that
+   all other fields are undefined besides the `failed_verification`,
+   `last_slot_verified`, and `last_slot_modified` fields, so we can
+   just return the core struct's size. */
+static inline FD_FN_PURE ulong
+fd_program_cache_failed_verification_entry_footprint( void ) {
+  ulong l = FD_LAYOUT_INIT;
+  l = FD_LAYOUT_APPEND( l, alignof(fd_program_cache_entry_t), sizeof(fd_program_cache_entry_t) );
+  l = FD_LAYOUT_FINI( l, 128UL );
```
yufeng-jump: nit: define an alignment for program_cache_entry_t?  It's a little odd for the entry_t to start at 8-byte aligned, but always end at 128-byte aligned.

src/funk/fd_funk_rec.c
```diff
@@ -352,7 +352,10 @@ fd_funk_rec_try_clone_safe( fd_funk_t *               funk,
         funk, txn,key, &txn_glob, query_glob );
 
     /* If the record exists and already exists in the specified funk
-       txn, we can return successfully. */
+       txn, we can return successfully.
+
+       TODO: This should probably also check that the record has a large
+       enough size, i.e. rec_glob >= min_sz. */
```
yufeng-jump, line 358: nice catch

yufeng-jump (approved):

ripatel-fd:
When fixing bugs like this one, could you add a corresponding regression test?

src/flamenco/runtime/program/fd_program_cache.c
mjain-jump: good catch, fixed

mjain-jump:
> When fixing bugs like this one, could you add a corresponding regression test?

Done, added a unit test and verified that it fails before this PR

PR #5929, opened 2025-08-05
tower: return when trying to revote

kbhargava-jump:


lidatong:
as discussed, think there may be a deeper issue on the caller side and this patch would just bandaid over it -- would investigate that

PR #5931, opened 2025-08-05
runtime: refactor fd_stakes_t into fd_stake_delegations_t

ibhatt-jumptrading:


src/flamenco/stakes/test_stake_delegations.c
```diff
@@ -0,0 +1,120 @@
...
+int main( int argc, char ** argv ) {
+  fd_boot( &argc, &argv );
+
+  char *      _page_sz = "gigantic";
+  ulong       numa_idx = fd_shmem_numa_idx( 0 );
+  fd_wksp_t * wksp     = fd_wksp_new_anonymous( fd_cstr_to_shmem_page_sz( _page_sz ),
+                                                1UL,
+                                                fd_shmem_cpu_idx( numa_idx ),
+                                                "wksp",
+                                                0UL );
```
ripatel-fd: Nit but I would just copy paste the unit test boilerplate here. 

`test_funk_rec` is a good example. The unit test should support --page-sz and --page-cnt argumnets since the unit test runner uses this to select huge pages for all tests when 1G pages aren't available

src/flamenco/stakes/test_stake_delegations.c
```diff
@@ -0,0 +1,120 @@
...
+  void * stake_delegations_mem = fd_wksp_alloc_laddr( wksp, fd_stake_delegations_align(), fd_stake_delegations_footprint( 10UL ), 1UL );
+  FD_TEST( stake_delegations_mem );
+
+  FD_TEST( fd_stake_delegations_align()>=alignof(fd_stake_delegations_t)  );
+  FD_TEST( fd_stake_delegations_align()>=fd_stake_delegation_pool_align() );
+  FD_TEST( fd_stake_delegations_align()>=fd_stake_delegation_map_align()  );
+  FD_TEST( fd_stake_delegations_align()==FD_STAKE_DELEGATIONS_ALIGN );
+
+  ulong chain_cnt = fd_stake_delegation_map_chain_cnt_est( 10UL );
+  FD_TEST( fd_stake_delegations_footprint( 10UL ) >= fd_stake_delegations_align() + fd_stake_delegation_pool_footprint( 10UL ) + fd_stake_delegation_map_footprint( chain_cnt ) );
```
ripatel-fd: Dedup this 10UL constant to a `ulong const rec_max` at the top of the function to make it easier to change it in the future

src/flamenco/stakes/Local.mk
```diff
@@ -1,6 +1,12 @@
 ifdef FD_HAS_INT128
 $(call add-hdrs,fd_stakes.h)
 $(call add-objs,fd_stakes,fd_flamenco)
+
+$(call add-hdrs,fd_stake_delegations.h)
+$(call add-objs,fd_stake_delegations,fd_flamenco)
+$(call make-unit-test,test_stake_delegations,test_stake_delegations,fd_flamenco fd_ballet fd_util)
+$(call run-unit-test,test_stake_delegations,)
```
ripatel-fd:
```suggestion
$(call run-unit-test,test_stake_delegations)
```

src/flamenco/stakes/fd_stake_delegations.h
```diff
@@ -0,0 +1,157 @@
+#ifndef HEADER_fd_src_flamenco_stakes_fd_stake_delegations_h
+#define HEADER_fd_src_flamenco_stakes_fd_stake_delegations_h
+
+#include "../fd_flamenco_base.h"
+#include "../types/fd_types.h"
+
+FD_PROTOTYPES_BEGIN
```
ripatel-fd: PROTOTYPES only includes extern symbols (global variables, function prototypes).
So this should be moved down after the struct before the first function declaration

src/flamenco/stakes/fd_stake_delegations.h
```diff
@@ -0,0 +1,157 @@
...
+   requirements. */
+
+#define FD_STAKE_DELEGATIONS_FOOTPRINT (350000000UL)
+#define FD_STAKE_DELEGATIONS_ALIGN     (128UL)
+
+struct fd_stake_delegation {
+  ulong       next_;
+  fd_pubkey_t stake_account;
+  fd_pubkey_t vote_account;
+  ulong       stake;
```
ripatel-fd, line 38: better struct packing feng shui if you move the first two 32 byte pubkeys to the start of the struct IMO and have all the ulongs together. Doesn't matter much though

src/flamenco/stakes/fd_stake_delegations.h
```diff
@@ -0,0 +1,157 @@
...
+  double      warmup_cooldown_rate;
+};
+typedef struct fd_stake_delegation fd_stake_delegation_t;
+
+#define POOL_NAME fd_stake_delegation_pool
+#define POOL_T    fd_stake_delegation_t
+#define POOL_NEXT next_
+#include "../../util/tmpl/fd_pool.c"
+#undef POOL_NAME
+#undef POOL_T
```
ripatel-fd: pool undoes these defines for you, this is redundant

src/flamenco/stakes/fd_stake_delegations.h
```diff
@@ -0,0 +1,157 @@
...
+fd_stake_delegations_query( fd_stake_delegations_t const * stake_delegations,
+                            fd_pubkey_t const *            stake_account );
+
+
+/* fd_stake_delegations_count returns the number of stake delegations
+   in the stake delegations struct. fd_stake_delegations_t must be a
+   valid local join. */
+
+ulong
+fd_stake_delegations_count( fd_stake_delegations_t const * stake_delegations );
```
ripatel-fd:
```suggestion
fd_stake_delegations_cnt( fd_stake_delegations_t const * stake_delegations );
```

I think we usually call these "cnt", "max". But it doesn't matter. Having a max API might be useful for metrics

src/flamenco/stakes/fd_stake_delegations.h
```diff
@@ -0,0 +1,157 @@
...
+#define MAP_KEY_HASH(key,seed) (fd_hash( seed, key, sizeof(fd_pubkey_t) ))
+#define MAP_NEXT               next_
+#include "../../util/tmpl/fd_map_chain.c"
+#undef MAP_NAME
+#undef MAP_KEY_T
+#undef MAP_ELE_T
+#undef MAP_KEY
+#undef MAP_KEY_EQ
+#undef MAP_KEY_HASH
+#undef MAP_NEXT
```
ripatel-fd: remove these undefs

src/flamenco/stakes/fd_stake_delegations.h
```diff
@@ -0,0 +1,157 @@
...
+#include "../../util/tmpl/fd_pool.c"
+#undef POOL_NAME
+#undef POOL_T
+
+#define MAP_NAME               fd_stake_delegation_map
+#define MAP_KEY_T              fd_pubkey_t
+#define MAP_ELE_T              fd_stake_delegation_t
+#define MAP_KEY                stake_account
+#define MAP_KEY_EQ(k0,k1)      (!(memcmp(&(k0)->key,&(k1)->key,sizeof(fd_pubkey_t))))
+#define MAP_KEY_HASH(key,seed) (fd_hash( seed, key, sizeof(fd_pubkey_t) ))
```
ripatel-fd, line 57: fd_funk_rec_key_hash1

src/flamenco/stakes/fd_stake_delegations.h
```diff
@@ -0,0 +1,157 @@
...
+#define MAP_NEXT               next_
+#include "../../util/tmpl/fd_map_chain.c"
```
ripatel-fd, line 59: why not map_dynamic.c? It would save you from having a pool. Fine with either

src/flamenco/stakes/fd_stake_delegations.c
```diff
@@ -0,0 +1,311 @@
...
+  if( FD_UNLIKELY( idx==ULONG_MAX ) ) {
+    return;
+  }
+
+  fd_stake_delegation_pool_idx_release( stake_delegation_pool, delegation_idx );
+}
+
+fd_stake_delegation_t const *
+fd_stake_delegations_query( fd_stake_delegations_t const * stake_delegations,
+                            fd_pubkey_t const *            stake_account ) {
```
ripatel-fd: Since this is super simple and the internals are already exposed in the header, I suggest inlining this

src/flamenco/stakes/fd_stake_delegations.c
```diff
@@ -0,0 +1,311 @@
...
+
+  if( FD_UNLIKELY( !stake_delegations ) ) {
+    FD_LOG_CRIT(( "NULL stake_delegations" ));
+    return NULL;
+  }
+
+  if( FD_UNLIKELY( !stake_account ) ) {
+    FD_LOG_CRIT(( "NULL stake_account" ));
+    return NULL;
+  }
```
ripatel-fd: If you do decide to inline these, probably only generate these checks if handholding is enabled to reduce code bloat for the optimized build

src/flamenco/stakes/fd_stake_delegations.h
```diff
@@ -0,0 +1,157 @@
...
+   to the footprint of each stake_delegation * the number of total
+   stake accounts that the system will support. If there are 3M stake
+   accounts and each one is 104 bytes, then we can assume that the total
+   number is ~350MB.
+
+   TODO: This needs to be more carefully bounded where we account for
+   the overhead of the map + pool headers as well as the alignment
+   requirements. */
+
+#define FD_STAKE_DELEGATIONS_FOOTPRINT (350000000UL)
```
ripatel-fd, line 22: For magic footprint numbers like these, we should have a unit test verifying that the calculated number is correct for the claimed footprint.

But IMO there should just be a configurable "max stake delegations" setting in the config file. And the footprint should be derived from the max like with any other object

src/flamenco/stakes/fd_stake_delegations.h
ibhatt-jumptrading: no reason, just using map_chain + pool out of habit

src/flamenco/stakes/fd_stake_delegations.h
ibhatt-jumptrading: yeah agreed this needs to be configurable eventually. will address this in a follow up (left a todo)

PR #5932, opened 2025-08-05
snapshots: send highest manifest slot

cali-jumptrading:
Repair can start before the snapshot is fully loaded if the snapshot tiles send the highest manifest slot seen so far. The highest manifest slot is typically the incremental snapshot slot, but can be the full snapshot slot if incremental snapshots are disabled. The highest manifest slot is guaranteed to be monotonically increasing. It is sent by the snaprd tile when a snapshot peer is selected.

Makes use of https://github.com/firedancer-io/firedancer/pull/5777 to resolve snapshots for each peer.

### DIFF
--- a/src/discof/restore/fd_snapin_tile.c
+++ b/src/discof/restore/fd_snapin_tile.c
@@ -214,7 +214,9 @@ handle_data_frag( fd_snapin_tile_t *  ctx,
 static void
 handle_control_frag( fd_snapin_tile_t *  ctx,
                      fd_stem_context_t * stem,
-                     ulong               sig ) {
+                     ulong               sig,
+                     ulong               tsorig,
+                     ulong               tspub ) {
   switch( sig ) {
     case FD_SNAPSHOT_MSG_CTRL_RESET_FULL:
       ctx->full = 1;
@@ -257,6 +259,11 @@ handle_control_frag( fd_snapin_tile_t *  ctx,
     case FD_SNAPSHOT_MSG_CTRL_SHUTDOWN:
       ctx->state = FD_SNAPIN_STATE_SHUTDOWN;
       break;
+    case FD_SNAPSHOT_MSG_HIGHEST_MANIFEST_SLOT:
+      /* Informational control message forwarded to snap_out */
+      fd_stem_publish( stem, 0UL, FD_SSMSG_HIGHEST_MANIFEST_SLOT, 0UL, 0UL, 0UL, tsorig, tspub );
+      /* No need to send an ack for an informational control message */
+      return;
     default:
       FD_LOG_ERR(( "unexpected control sig %lu", sig ));
       return;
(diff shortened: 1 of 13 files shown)


src/discof/restore/fd_snapin_tile.c
```diff
@@ -257,6 +259,11 @@ handle_control_frag( fd_snapin_tile_t *  ctx,
     case FD_SNAPSHOT_MSG_CTRL_SHUTDOWN:
       ctx->state = FD_SNAPIN_STATE_SHUTDOWN;
       break;
+    case FD_SNAPSHOT_MSG_HIGHEST_MANIFEST_SLOT:
```
mmcgee-jump, line 262: No need to do this (pass it all the way along). Just publish it straight from `snaprd` tile

src/discof/restore/utils/fd_ssping.c
```diff
@@ -333,43 +372,71 @@ poll_advance( fd_ssping_t * ssping,
...
   for( ulong i=0UL; i<ssping->fds_len; i++ ) {
+    fd_ssping_peer_t * peer = peer_pool_ele( ssping->pool, ssping->fds_idx[ i ] );
     struct pollfd * pfd = &ssping->fds[ i ];
     if( FD_UNLIKELY( pfd->revents & (POLLERR|POLLHUP) ) ) {
       unping_peer( ssping, peer_pool_ele( ssping->pool, ssping->fds_idx[ i ] ), now );
       continue;
     }
 
-    if( FD_LIKELY( pfd->revents & POLLOUT ) ) {
-      struct icmphdr icmp_hdr = (struct icmphdr){
```
mmcgee-jump, line 343: The more I think about this the more I think I was wrong about using HTTP server ACK response time. Using ping time is actually probably correct for the pinger because we are just trying to find the server in the closest data center. Whether they had some latency spike on the HTTP server when we sent request probably doesn't matter too much

src/discof/restore/utils/fd_ssmsg.h
```diff
@@ -3,9 +3,10 @@
 
 #include "../../../flamenco/types/fd_types.h"
 
-#define FD_SSMSG_MANIFEST_FULL        (0) /* A snapshot manifest message from the full snapshot */
-#define FD_SSMSG_MANIFEST_INCREMENTAL (1) /* A snapshot manifest message from the incremental snapshot */
-#define FD_SSMSG_DONE                 (2) /* Indicates the snapshot is fully loaded and tiles are shutting down */
+#define FD_SSMSG_MANIFEST_FULL         (0) /* A snapshot manifest message from the full snapshot */
+#define FD_SSMSG_MANIFEST_INCREMENTAL  (1) /* A snapshot manifest message from the incremental snapshot */
+#define FD_SSMSG_HIGHEST_MANIFEST_SLOT (2) /* The highest manifest slot so far, guaranteed to be monotonically increasing */
```
mmcgee-jump, line 8: Comment doesn't seem right. Why would this be monotonically increasing?

mmcgee-jump: This should really be
`EXPECTED_MANIFEST_SLOT`

mmcgee-jump (changes requested):
.

cali-jumptrading:
Completed by https://github.com/firedancer-io/firedancer/pull/6285

PR #5933, opened 2025-08-05
snapshots: fuzz ssarchive parser

cali-jumptrading:


src/discof/restore/utils/fuzz_ssarchive_parser.c
```diff
@@ -0,0 +1,32 @@
...
+}
+
+int
+LLVMFuzzerTestOneInput( uchar const * data,
+                        ulong         size ) {
+  (void)size;
+  ulong full_slot;
+  ulong incremental_slot;
+  uchar decoded_hash[ FD_HASH_FOOTPRINT ];
+  fd_ssarchive_parse_filename( (char *)data, &full_slot, &incremental_slot, decoded_hash );
```
ripatel-fd:
```suggestion
  char * name_cstr = malloc( size+1 );
  assert( name_cstr );
  fd_cstr_fini( fd_cstr_append_text( fd_cstr_init( name_cstr ), (char const *)data, size );
  fd_ssarchive_parse_filename( name_cstr, &full_slot, &incremental_slot, decoded_hash );
  free( name_cstr );
```

ripatel-fd: Since this is null-terminated. You'll want to use libc malloc + free every iteration here, instead of using a fixed buffer, to get the sanitizer's OOB access detection for free.

cali-jumptrading: oh I see, thanks!

PR #5945, opened 2025-08-06
codeql: implicit void promotion lints

ripatel-fd:
Added to dev only for now since they are spammy.

Also fixed severity annotations for some low prio nightly warnings.

### DIFF
--- a/contrib/codeql/dev/ImplicitVoidPromotionAssign.ql
+++ b/contrib/codeql/dev/ImplicitVoidPromotionAssign.ql
@@ -1,0 +1,38 @@
+/**
+ * @name Implicit Void Ptr to Typed Ptr promotion in Assignment
+ * @description Implicit conversion from void* to a typed pointer in assignment expression
+ * @kind problem
+ * @problem.severity recommendation
+ * @precision low
+ * @id firedancer-io/implicit-void-promotion-assign
+ */
+
+import cpp
+
+class VoidPointer extends PointerType {
+  VoidPointer() {
+    this.getBaseType().getUnspecifiedType() instanceof VoidType
+  }
+}
+
+class NotVoidPointer extends PointerType {
+  NotVoidPointer() {
+    not this.getBaseType().getUnspecifiedType() instanceof VoidType
+  }
+}
+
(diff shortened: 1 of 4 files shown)


two-heart:
Thank you, I already found one API misuse with this query approach.
Putting it into dev sounds right to me.
How about combining both into one query, always checking for explicit conversion, using postfix cast and where possible instanceof instead of type unification as parameters, and excluding `fd_type_pun` r values?
```
/**
 * @name Implicit Void Ptr to Typed Ptr promotion
 * @description Implicit conversion from void* to a typed pointer
 * @kind problem
 * @problem.severity recommendation
 * @precision low
 * @id firedancer-io/implicit-void-promotion
 */

import cpp

class VoidPointer extends PointerType {
  VoidPointer() { this.getBaseType().getUnspecifiedType() instanceof VoidType }
}

predicate implicitVoidPromotion(Type lvalue, Type rvalue) {
  not lvalue.getUnderlyingType() instanceof VoidPointer and
  rvalue.getUnderlyingType() instanceof VoidPointer
}

predicate allowedSourceFile(File file) { file.getBaseName() != "fd_types.c" }

class BroadAssign extends Locatable {
  BroadAssign() { this instanceof Variable or this instanceof AssignExpr }

  Expr getRExpr() {
    result = this.(Variable).getInitializer().getExpr() or
    result = this.(AssignExpr).getRValue()
  }

  Type getLType() {
    result = this.(Variable).getType() or
    result = this.(AssignExpr).getLValue().getType()
  }

  Type getRType() { result = this.getRExpr().getType() }
}

from BroadAssign assign
where
  implicitVoidPromotion(assign.getLType(), assign.getRType()) and
  allowedSourceFile(assign.getLocation().getFile()) and
  not assign.getRExpr().isInMacroExpansion() and
  not assign.getRExpr().hasExplicitConversion() and
  not assign.getRExpr().(FunctionCall).getTarget().getName().matches("fd_type_pun%")
select assign, "Implicit conversion from void * to " + assign.getLType()
```

two-heart (changes requested):
.

ripatel-fd:
@two-heart Good idea. You are much better of these prompts than I am, so please feel free to reopen a cleaner version. The reason I put them into two separate prompts is that I tried to combine them but couldn't get it to work.

PR #5954, opened 2025-08-06
snapshots: lthash

### LINKED ISSUE #5719: verify snapshot lthash

cali-jumptrading:
Adds snapshot hashing tiles. They are disabled by default but enabled in CI to verify the snapshot lthash in ledgers. We plan to enable them by default once their performance can keep up with download speed.


src/flamenco/runtime/fd_hashes.c
```diff
@@ -31,6 +31,33 @@ fd_hashes_account_lthash( fd_pubkey_t const       * pubkey,
...
+
+  fd_blake3_t b3[1];
+  fd_blake3_init( b3 );
+  fd_blake3_append( b3, &lamports, sizeof( ulong ) );
+  fd_blake3_append( b3, data, data_len );
+  fd_blake3_append( b3, &executable_flag, sizeof( uchar ) );
+  fd_blake3_append( b3, owner, FD_PUBKEY_FOOTPRINT );
+  fd_blake3_append( b3, pubkey, FD_PUBKEY_FOOTPRINT );
+  fd_blake3_fini_varlen( b3, lthash_out->bytes, FD_LTHASH_LEN_BYTES );
+}
```
topointon-jump: Instead of duplicating this code could we instead call this function from `fd_hashes_account_lthash`?

cali-jumptrading: Ah good point, thanks!

src/app/firedancer/config/default.toml
```diff
@@ -698,6 +698,14 @@ user = ""
     # requests.
     sign_tile_count = 2
 
+    # How many snapshot hash tiles to run.  The snapshot hash tiles are
+    # responsible for verifying the hash of all accounts in the loaded
+    # snapshot via lthash (lattice hashing).  Currently, set to 0 by
+    # default because it is too slow to run in the full client.  TODO:
+    # enable snapshot hash tiles in the full client and update this
+    # comment.
+    hash_tile_count = 0
```
topointon-jump: Could we make it clear that if the hash tile count is 0, no hashing will be done? Wondering if we should instead have a `snapshots.verify` flag instead to enable/disable the hashing, instead of hash_tile_count = 0 => no hashing.

src/discof/restore/utils/fd_ssctrl.h
```diff
@@ -49,4 +51,61 @@
...
+/* The following message signatures define the sequence of control and
+   data messages between the snapin and snaphs tile.  The snaphs tile
+   does not participate in control messages between snaprd, snapdc, and
+   snapin.  It acts as a hashing accelerator for snapin. */
+#define FD_SNAPSHOT_HASH_MSG_RESET             (8UL)  /* Indicates snapin has a new account stream incoming */
+#define FD_SNAPSHOT_HASH_MSG_SUB               (9UL)  /* Indicates snapin has encountered a duplicate account whose hash must be subtracted */
+#define FD_SNAPSHOT_HASH_MSG_ACCOUNT_HDR       (10UL) /* Indicates snapin has encountered a new account metadata */
+#define FD_SNAPSHOT_HASH_MSG_ACCOUNT_DATA      (11UL) /* Account data that is sent as snapin processes a new account */
+#define FD_SNAPSHOT_HASH_MSG_FINI              (12UL) /* Indicates the account stream from snapin is done, awaiting hash result */
+#define FD_SNAPSHOT_HASH_MSG_RESULT            (13UL) /* Hash result ent from snaphs to snapin*/
```
topointon-jump: typo?
```suggestion
#define FD_SNAPSHOT_HASH_MSG_RESULT            (13UL) /* Hash result sent from snaphs to snapin*/
```

src/discof/restore/fd_snaphs_tile.c
```diff
@@ -0,0 +1,256 @@
...
+             ulong               seq,
+             ulong               sig,
+             ulong               chunk,
+             ulong               sz,
+             ulong               ctl ) {
+  (void)in_idx;
+  (void)seq;
+  (void)ctl;
+
+  FD_TEST( ctx->state!=FD_SNAPHS_STATE_SHUTDOWN );
```
topointon-jump: Do we want these FD_TEST's in production code?

src/discof/restore/utils/fd_ssctrl.h
```diff
@@ -49,4 +51,61 @@
...
+#define FD_SNAPSHOT_HASH_MSG_SHUTDOWN          (14UL) /* Snapin is shutting down, snaphs can shutdown too */
+
+/* Stolen from fd_runtime_public.h */
+#define FD_ACC_SZ_MAX       (10UL<<20) /* 10MiB */
+
```
topointon-jump: Can probably move this into fd_runtime_const.h and import that

src/disco/metrics/metrics.xml
```diff
@@ -928,6 +928,11 @@ metric introduced.
     <gauge name="AccountsInserted" summary="Number of accounts inserted during snpashot loading. Might decrease if snapshot load is aborted and restarted" />
 </tile>
 
+<tile name="snaphs">
+    <gauge name="State" summary="State of the tile. 0=hashing, 1=done, 2=shutdown" />
+    <gauge name="AccountsHashed" summary="Number of accounts hashed so far during snapshot loading. Might decrease if snapshot load is aborted and restarted" />
+</tile>
```
topointon-jump: Might be less confusing if this was a counter?

src/discof/restore/fd_snapin_tile.c
```diff
@@ -114,9 +169,17 @@ is_duplicate_account( fd_snapin_tile_t * ctx,
   if( FD_UNLIKELY( rec_meta ) ) {
     if( FD_LIKELY( rec_meta->slot>ctx->ssparse->accv_slot ) ) return 1;
 
-    /* TODO: Reaching here means the existing value is a duplicate
-       account.  We need to hash the existing account and subtract that
-       hash from the running lthash. */
+    if( FD_LIKELY( ctx->hash_info.enabled ) ) {
+      /* Send the existing account to the hash tile */
+      /* TODO: cache the previous lthash somewhere in funk to avoid
+         recalculating hash. */
```
topointon-jump: We considered doing this... it would mean an additional 2048 bytes per account which is probably not ideal.. this storage-compute tradeoff is something we're exploring for transaction execution also and we should revisit this after the hashing changes and database changes are done.

src/discof/restore/fd_snaphs_tile.c
cali-jumptrading: yeah it should be fine, it's in the other snapshot tiles too.

src/disco/metrics/metrics.xml
cali-jumptrading: The other snapshot tiles use gauges because the values might not always increase monotonically.

src/app/firedancer/config/default.toml
```diff
@@ -698,6 +698,14 @@ user = ""
     # requests.
     sign_tile_count = 2
 
+    # How many snapshot hash tiles to run.  The snapshot hash tiles are
```
mmcgee-jump: A few things,

- `hash_tile_count` is very vague and could refer anything. Be more specific, in this case the convention is `<tilename>_tile_count` so `snaplt_tile_count` ?
- Allowing someone to turn this off is probably not desirable, and we should just make this config production ready now. I'd suggest something like,

```suggestion
[layout]
    # How many snapshot lthash tiles to run.  Snapshot hash tiles verify
    # the contents of accounts in the loaded snapshot, in parallel.  Too
    # few and snapshot loading might be delayed.  Once the hash of the
    # accounts in a snapshot is calculated, if it does not match the
    # validator will abort with an error.
    snaplt_tile_count = 1

[development]
    [snapshots]
        # Set to true to disable verification of the lthash in the
        # loaded snapshot.  This is not safe or supported in production
        # and should only be used for development and testing purposes.
        #
        # If lthash verification is disabled, the validator will not
        # start any lthash tiles and the value of `snaplt_tile_count`
        # in the layout will be ignored.
        disable_lthash_verification = false
```

Once you have this, you can:

- Check that lthash_tile_count > 0
- Check that the development option is not enabled when running against a production cluster (same as some other options, e.g. we check this for --no-sandbox).

src/app/firedancer/topology.c
```diff
@@ -201,6 +202,7 @@ fd_topo_initialize( config_t * config ) {
   ulong bank_tile_cnt   = config->layout.bank_tile_count;
   ulong exec_tile_cnt   = config->firedancer.layout.exec_tile_count;
   ulong writer_tile_cnt = config->firedancer.layout.writer_tile_count;
+  ulong hash_tile_cnt   = config->firedancer.layout.hash_tile_count;
```
```suggestion
  ulong snaplt_tile_cnt   = config->firedancer.layout.snaplt_tile_count;
```

src/discof/restore/fd_snaphs_tile.c
```diff
@@ -0,0 +1,256 @@
...
+                                       prev_acc->hdr.executable,
+                                       prev_acc->data,
+                                       prev_acc->hdr.data_len,
+                                       prev_lthash );
+      fd_lthash_sub( &ctx->running_lthash, prev_lthash );
+      break;
+    }
+    case FD_SNAPSHOT_HASH_MSG_ACCOUNT_HDR: {
+      FD_TEST( ctx->state==FD_SNAPHS_STATE_HASHING );
+      fd_snapshot_account_t * account = (fd_snapshot_account_t *)fd_chunk_to_laddr_const( ctx->in.wksp, chunk );
```
mmcgee-jump: Could be useful to verify
```
FD_TEST( ctx->acc_data_sz==0UL );
```

src/discof/restore/fd_snaphs_tile.c
```diff
@@ -0,0 +1,256 @@
...
+      fd_memcpy( &ctx->account, account, sizeof(fd_snapshot_account_t) );
+      break;
+    }
+    case FD_SNAPSHOT_HASH_MSG_ACCOUNT_DATA: {
+      FD_TEST( ctx->state==FD_SNAPHS_STATE_HASHING );
+      fd_memcpy( ctx->acc_data + ctx->acc_data_sz,
```
mmcgee-jump: Looks like the `lthash` calculation is just a series of blake3 appends, so you can do it unbuffered / without memcpy by just streaming the account data into blake3 here?

src/discof/restore/fd_snapin_tile.c
```diff
@@ -89,17 +121,40 @@ metrics_write( fd_snapin_tile_t * ctx ) {
...
   fd_snapin_tile_t * ctx = (fd_snapin_tile_t*)_ctx;
 
   ulong sz = sizeof(fd_snapshot_manifest_t)+manifest_sz;
+
+  fd_snapshot_manifest_t * manifest = (fd_snapshot_manifest_t * )fd_chunk_to_laddr_const( ctx->manifest_out.wksp, ctx->manifest_out.chunk );
+  if( FD_LIKELY( manifest->has_accounts_lthash ) ) {
+    uchar * manifest_lthash = ctx->full ? ctx->hash_info.full_lthash.bytes : ctx->hash_info.incremental_lthash.bytes;
+    fd_memcpy( manifest_lthash, manifest->accounts_lthash, sizeof(fd_lthash_value_t) );
+  } else {
+    /* disable lthash verification if there's no lthash in the
```
mmcgee-jump: You should probably just abort with an error here, or trigger a malformed snapshot.

src/app/firedancer-dev/main.c
```diff
@@ -106,6 +106,7 @@ extern fd_topo_run_tile_t fd_tile_shredcap;
 extern fd_topo_run_tile_t fd_tile_snaprd;
 extern fd_topo_run_tile_t fd_tile_snapdc;
 extern fd_topo_run_tile_t fd_tile_snapin;
+extern fd_topo_run_tile_t fd_tile_snaphs;
```
mmcgee-jump: Also add to `main.c` in firedancer

src/discof/restore/fd_snapin_tile.c
```diff
@@ -114,9 +169,17 @@ is_duplicate_account( fd_snapin_tile_t * ctx,
   if( FD_UNLIKELY( rec_meta ) ) {
     if( FD_LIKELY( rec_meta->slot>ctx->ssparse->accv_slot ) ) return 1;
 
-    /* TODO: Reaching here means the existing value is a duplicate
-       account.  We need to hash the existing account and subtract that
-       hash from the running lthash. */
+    if( FD_LIKELY( ctx->hash_info.enabled ) ) {
+      /* Send the existing account to the hash tile */
+      /* TODO: cache the previous lthash somewhere in funk to avoid
```
mmcgee-jump: This is extremely rare and not worth the memory. I'd remove the TODO, what you have is good

src/app/firedancer/topology.c
```diff
@@ -376,11 +383,8 @@ fd_topo_initialize( config_t * config ) {
   /**/                 fd_topob_link( topo, "snap_out",     "snap_out",     2UL,                                      5UL*(1UL<<30UL),               1UL );
   /**/                 fd_topob_link( topo, "snapdc_rd",    "snapdc_rd",    128UL,                                    0UL,                           1UL );
   /**/                 fd_topob_link( topo, "snapin_rd",    "snapin_rd",    128UL,                                    0UL,                           1UL );
-
-  /* Replay decoded manifest dcache topo obj */
-  fd_topo_obj_t * replay_manifest_dcache = fd_topob_obj( topo, "dcache", "replay_manif" );
-  fd_pod_insertf_ulong( topo->props, 2UL << 30UL, "obj.%lu.data_sz", replay_manifest_dcache->id );
-  fd_pod_insert_ulong(  topo->props, "manifest_dcache", replay_manifest_dcache->id );
+  FOR(hash_tile_cnt)   fd_topob_link( topo, "snapin_hsh",   "snapin_hsh",   128UL,                                    sizeof(fd_snapshot_existing_account_t),                1UL );
```
mmcgee-jump: The convention for this kind of message passing is a single `snapin_snaph` link (not one per tile). You publish all frags on the link, and the snaphs tiles select which ones they want to process in the `before_frag` callback.

src/discof/restore/fd_snapin_tile.c
```diff
@@ -89,17 +121,40 @@ metrics_write( fd_snapin_tile_t * ctx ) {
...
+
   FD_TEST( sz<=ctx->manifest_out.mtu );
   ulong sig = ctx->full ? fd_ssmsg_sig( FD_SSMSG_MANIFEST_FULL, manifest_sz ) :
                           fd_ssmsg_sig( FD_SSMSG_MANIFEST_INCREMENTAL, manifest_sz );
   fd_stem_publish( ctx->stem, 0UL, sig, ctx->manifest_out.chunk, sz, 0UL, 0UL, 0UL );
   ctx->manifest_out.chunk = fd_dcache_compact_next( ctx->manifest_out.chunk, sz, ctx->manifest_out.chunk0, ctx->manifest_out.wmark );
+
+  if( FD_LIKELY( ctx->hash_info.enabled ) ) {
+    for( ulong i=0UL; i<ctx->hash_info.num_hash_tiles; i++ ) {
+      fd_stem_publish( ctx->stem, FD_SNAPIN_HSH_IDX + i, FD_SNAPSHOT_HASH_MSG_RESET, 0UL, 0UL, 0UL, 0UL, 0UL );
```
mmcgee-jump: See comment in topology, you just need one link here

src/discof/restore/fd_snapin_tile.c
```diff
@@ -346,6 +511,29 @@ unprivileged_init( fd_topo_t *      topo,
...
+    ctx->hash_out[ i ].chunk  = ctx->hash_out[ i ].chunk0;
+    ctx->hash_out[ i ].mtu    = hash_out_link->mtu;
+  }
+
+  fd_lthash_zero( &ctx->hash_info.full_lthash );
+  fd_lthash_zero( &ctx->hash_info.incremental_lthash );
+  fd_lthash_zero( &ctx->hash_info.result_lthash );
 }
 
 #define STEM_BURST 2UL /* For control fragments, one acknowledgement, and one malformed message */
```
mmcgee-jump: No longer correct?

src/discof/restore/fd_snapin_tile.c
```diff
@@ -237,12 +330,17 @@ handle_control_frag( fd_snapin_tile_t *  ctx,
...
+        }
+        ctx->pending_ack = 1;
+      }
       fd_snapshot_parser_reset( ctx->ssparse, fd_chunk_to_laddr( ctx->manifest_out.wksp, ctx->manifest_out.chunk ), ctx->manifest_out.mtu );
-
       fd_funk_txn_xid_t incremental_xid = fd_funk_generate_xid();
-      ctx->funk_txn = fd_funk_txn_prepare( ctx->funk, ctx->funk_txn, &incremental_xid, 0 );
-      ctx->full     = 0;
-      ctx->state    = FD_SNAPIN_STATE_LOADING;
+      ctx->funk_txn     = fd_funk_txn_prepare( ctx->funk, ctx->funk_txn, &incremental_xid, 0 );
```
mmcgee-jump: extra whitespace?

src/discof/restore/fd_snapin_tile.c
```diff
@@ -287,12 +426,32 @@ returnable_frag( fd_snapin_tile_t *  ctx,
...
+              int *               charge_busy ) {
+  (void)opt_poll_in;
+  (void)charge_busy;
+
+  if( FD_UNLIKELY( ctx->pending_ack ) ) {
+    FD_TEST( ctx->hash_info.enabled );
+    /* A pending ack is waiting for lthash results to come back from
+       the hashing tiles.  Once these hashes come back, the ack can be
+       sent. */
+    if( FD_LIKELY( ctx->hash_info.received_lthashes==ctx->hash_info.num_hash_tiles ) ) {
```
mmcgee-jump: Why in after credit, and not in after_frag when you receive the lthash?

src/discof/restore/fd_snapin_tile.c
```diff
@@ -251,17 +349,58 @@ handle_control_frag( fd_snapin_tile_t *  ctx,
...
+      /* TODO: more robust in indexing */
+      fd_lthash_value_t const * result_lthash = (fd_lthash_value_t const *)fd_chunk_to_laddr_const( ctx->hash_in[ in_idx - 1UL ].wksp, chunk );
+      fd_lthash_add( &ctx->hash_info.result_lthash, result_lthash );
+      ctx->hash_info.received_lthashes++;
+
+      if( FD_LIKELY( ctx->hash_info.received_lthashes!=ctx->hash_info.num_hash_tiles ) ) break;
+
+      fd_lthash_value_t * ref_lthash = ctx->full ? &ctx->hash_info.full_lthash : &ctx->hash_info.incremental_lthash;
+      if( FD_UNLIKELY( memcmp( ref_lthash, &ctx->hash_info.result_lthash, sizeof(fd_lthash_value_t) ) ) ) {
+        FD_LOG_WARNING(( "calculated accounts lthash %s does not match accounts lthash %s in snapshot manifest",
```
mmcgee-jump, line 361: I think you have to exit with an error here, because we already told replay the snapshot was done loading. Otherwise you need to design a reset mechanism for replay and all other tiles as well.

mmcgee-jump (changes requested):
.

src/discof/restore/fd_snapin_tile.c
cali-jumptrading: Ah yes, this is no longer correct... But also can't be changed without increasing the depth of the `snap_out` link.

src/discof/restore/fd_snapin_tile.c
cali-jumptrading: oh nice catch, I think this is ok if we move the `SSMSG_DONE` message to after the hash is validated?

src/discof/restore/fd_snapin_tile.c
cali-jumptrading: oh right, I was being dumb 🤦

src/discof/restore/fd_snapin_tile.c
cali-jumptrading: I chose to call `transition_malformed` here.

src/discof/restore/fd_snapin_tile.c
cali-jumptrading: removed todo

src/app/firedancer/topology.c
cali-jumptrading: done, but filtering is done in `during_frag`, as it is not possible in `before_frag` without tagging messages.

src/discof/restore/fd_snapin_tile.c
cali-jumptrading: done

src/discof/restore/fd_snapin_tile.c
cali-jumptrading: Perhaps it is fine to leave it at 2 for now... thinking about it, in one iteration, snapin either sends one control message to the snaplt tiles, one acknowledgement, and one malformed message, or sends one account hdr/data frag over to the snapshot tiles when processing data. So there is at most 2 frags sent per snapin out link.

src/discof/restore/fd_snaphs_tile.c
cali-jumptrading: yeah this is possible, I've added it but this might change later if we decide to batch hash accounts or something.

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -36,6 +37,12 @@ static void
 backtest_topo( config_t * config ) {
   ulong exec_tile_cnt   = config->firedancer.layout.exec_tile_count;
   ulong writer_tile_cnt = config->firedancer.layout.writer_tile_count;
+  ulong snaplt_tile_cnt = config->firedancer.layout.snaplt_tile_count;
+  int snaplt_enabled    = !config->development.snapshots.disable_lthash_verification;
+
+  if( FD_UNLIKELY( snaplt_enabled && !snaplt_tile_cnt ) ) {
```
mmcgee-jump: No need for this check here. Just `CFG_HAS_NON_ZERO` along with the other layout checks in `fd_config.c`

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -380,9 +404,13 @@ backtest_cmd_fn( args_t *   args FD_PARAM_UNUSED,
   fd_topo_tile_t * snapdc_tile = &topo->tiles[ fd_topo_find_tile( topo, "snapdc", 0UL ) ];
   fd_topo_tile_t * snapin_tile = &topo->tiles[ fd_topo_find_tile( topo, "snapin", 0UL ) ];
 
+  ulong snaplt_tile_idx        = fd_topo_find_tile( topo, "snaplt", 0UL );
+  fd_topo_tile_t * snaplt_tile = snaplt_tile_idx!=ULONG_MAX ? &topo->tiles[ snaplt_tile_idx ] : NULL;
```
mmcgee-jump: There's N of these tiles, right?  Not just 1. Might want to sum the metrics across them

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -399,8 +429,9 @@ backtest_cmd_fn( args_t *   args FD_PARAM_UNUSED,
     ulong snaprd_status = FD_VOLATILE_CONST( snaprd_metrics[ MIDX( GAUGE, TILE, STATUS ) ] );
     ulong snapdc_status = FD_VOLATILE_CONST( snapdc_metrics[ MIDX( GAUGE, TILE, STATUS ) ] );
     ulong snapin_status = FD_VOLATILE_CONST( snapin_metrics[ MIDX( GAUGE, TILE, STATUS ) ] );
+    ulong snaplt_status = snaplt_metrics ? FD_VOLATILE_CONST( snaplt_metrics[ MIDX( GAUGE, TILE, STATUS ) ] ) : 2UL;
 
-    if( FD_UNLIKELY( snaprd_status==2UL && snapdc_status==2UL && snapin_status == 2UL ) ) break;
+    if( FD_UNLIKELY( snaprd_status==2UL && snapdc_status==2UL && snapin_status == 2UL && snaplt_status==2UL ) ) break;
```
mmcgee-jump: Definitely needs to check snaplt_status of all snaplt here, not just 1

src/app/firedancer-dev/commands/snapshot_load.c
```diff
@@ -37,7 +44,7 @@ snapshot_load_topo( config_t *     config,
   static ushort tile_to_cpu[ FD_TILE_MAX ] = {0};
   if( args->snapshot_load.tile_cpus[0] ) {
     ulong cpu_cnt = fd_tile_private_cpus_parse( args->snapshot_load.tile_cpus, tile_to_cpu );
-    if( FD_UNLIKELY( cpu_cnt<4UL ) ) FD_LOG_ERR(( "--tile-cpus specifies %lu CPUs, but need at least 4", cpu_cnt ));
+    if( FD_UNLIKELY( cpu_cnt<4UL + snaplt_tile_cnt ) ) FD_LOG_ERR(( "--tile-cpus specifies %lu CPUs, but need at least %lu", cpu_cnt, 4UL + snaplt_tile_cnt ));
```
mmcgee-jump:
```suggestion
    if( FD_UNLIKELY( cpu_cnt<4UL + snaplt_tile_cnt ) ) FD_LOG_ERR(( "--tile-cpus specifies %lu CPUs, but need at least %lu", cpu_cnt, 4UL + (snaplt_enabled?snaplt_tile_cnt:0) ));
```

src/app/firedancer/topology.c
```diff
@@ -194,15 +195,22 @@ void
...
+  ulong exec_tile_cnt       = config->firedancer.layout.exec_tile_count;
+  ulong writer_tile_cnt     = config->firedancer.layout.writer_tile_count;
+  ulong snaplt_tile_cnt     = config->firedancer.layout.snaplt_tile_count;
+  ulong resolv_tile_cnt     = config->layout.resolv_tile_count;
+  ulong sign_tile_cnt       = config->firedancer.layout.sign_tile_count;
+
+  int snaplt_enabled = !config->development.snapshots.disable_lthash_verification;
+
+  if( FD_UNLIKELY( snaplt_enabled && !snaplt_tile_cnt ) ) {
+    FD_LOG_ERR(( "snapshot lthash verification enabled, but snaplt tile count is 0" ));
```
mmcgee-jump: As above, no need for this check

src/discof/restore/fd_snaplt_tile.c
```diff
@@ -0,0 +1,283 @@
...
+#define FD_SNAPLT_STATE_SHUTDOWN (3)
+
+struct fd_snaplt_tile {
+  int state;
+
+  fd_lthash_value_t running_lthash;
+
+  fd_snapshot_account_t account;
+  ulong                 acc_data_sz;
+  uchar                 acc_data[ FD_RUNTIME_ACC_SZ_MAX ];
```
mmcgee-jump: Not used?

src/discof/restore/fd_snaplt_tile.c
```diff
@@ -0,0 +1,283 @@
...
+  } out;
+
+};
+
+typedef struct fd_snaplt_tile fd_snaplt_tile_t;
+
+static int
+should_hash_account( fd_snaplt_tile_t * ctx,
+                     uchar const        account_pubkey[ static FD_HASH_FOOTPRINT ] ) {
+  if( FD_LIKELY( ctx->num_hash_tiles==1UL ) ) return 1;
```
mmcgee-jump: Don't need this line

src/flamenco/runtime/fd_hashes.c
```diff
@@ -12,21 +12,38 @@ fd_hashes_account_lthash( fd_pubkey_t const       * pubkey,
                           fd_account_meta_t const * account,
                           uchar const             * data,
                           fd_lthash_value_t       * lthash_out ) {
+  fd_hashes_account_lthash_simple( pubkey->uc,
```
mmcgee-jump, line 16: Don't think you need these changes now?

src/discof/restore/fd_snapin_tile.c
```diff
@@ -31,9 +37,17 @@ struct fd_snapin_tile {
...
-  fd_stem_context_t * stem;
+  fd_stem_context_t *    stem;
   fd_snapshot_parser_t * ssparse;
 
+  struct {
+    int               enabled;
+    ulong             received_lthashes;
+    ulong             num_hash_tiles;
+    fd_lthash_value_t manifest_lthash;
+    fd_lthash_value_t result_lthash;
```
mmcgee-jump:
```suggestion
    fd_lthash_value_t expected_lthash;
    fd_lthash_value_t calculated_lthash;
```

src/flamenco/runtime/fd_hashes.c
cali-jumptrading: I still use this function to hash a duplicate account in the `snaplt` tile

src/app/firedancer-dev/commands/backtest.c
cali-jumptrading: done

src/app/firedancer-dev/commands/backtest.c
cali-jumptrading: done

src/app/firedancer/topology.c
cali-jumptrading: done

src/discof/restore/fd_snaplt_tile.c
cali-jumptrading: removed

src/discof/restore/fd_snaplt_tile.c
cali-jumptrading: removed

src/app/firedancer-dev/commands/snapshot_load.c
cali-jumptrading: done

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -38,6 +39,8 @@ static void
 backtest_topo( config_t * config ) {
   ulong exec_tile_cnt   = config->firedancer.layout.exec_tile_count;
   ulong writer_tile_cnt = config->firedancer.layout.writer_tile_count;
+  ulong snaplt_tile_cnt = config->firedancer.layout.snaplt_tile_count;
+  int snaplt_enabled    = !config->development.snapshots.disable_lthash_verification;
```
mmcgee-jump: Prefer to describe these as positive action away from the default, so this should be `snaplt_disabled` (the same as it is in the config file)

src/app/firedancer/config/default.toml
```diff
@@ -1697,3 +1704,17 @@ user = ""
         # This option has not been tested or security audited and should
         # not currently be used in production.
         websocket_compression = false
+
+    # Development only options for snapshot laoding.
```
```suggestion
    # Development only options for snapshot loading.
```

src/app/firedancer/topology.c
```diff
@@ -211,7 +212,9 @@ fd_topo_initialize( config_t * config ) {
   ulong exec_tile_cnt   = config->firedancer.layout.exec_tile_count;
   ulong writer_tile_cnt = config->firedancer.layout.writer_tile_count;
   ulong sign_tile_cnt   = config->firedancer.layout.sign_tile_count;
+  ulong snaplt_tile_cnt = config->firedancer.layout.snaplt_tile_count;
 
+  int snaplt_enabled = !config->development.snapshots.disable_lthash_verification;\
```
mmcgee-jump: I would only enable this is `snapshots_enabled` also

src/app/firedancer/topology.c
```diff
@@ -499,6 +517,13 @@ fd_topo_initialize( config_t * config ) {
                       fd_topob_tile_in (    topo, "replay",  0UL,          "metric_in", "snap_out",     0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   }
 
+  if( FD_LIKELY( snapshots_enabled && snaplt_enabled ) ) {
+                         fd_topob_tile_out( topo, "snapin", 0UL, "snapin_lt",  0UL );
```
mmcgee-jump: alignment

src/discof/restore/fd_snapin_tile.c
```diff
@@ -89,14 +119,38 @@ metrics_write( fd_snapin_tile_t * ctx ) {
   FD_MGAUGE_SET( SNAPIN, STATE, (ulong)ctx->state );
 }
 
+static void
+transition_malformed( fd_snapin_tile_t * ctx,
```
mmcgee-jump: formatting

src/discof/restore/fd_snapin_tile.c
```diff
@@ -89,14 +119,38 @@ metrics_write( fd_snapin_tile_t * ctx ) {
...
+}
+
 static void
 manifest_cb( void * _ctx ) {
   fd_snapin_tile_t * ctx = (fd_snapin_tile_t*)_ctx;
 
+  ulong sz = sizeof(fd_snapshot_manifest_t);
+
+  if( FD_LIKELY( ctx->hash_info.enabled ) ) {
+    fd_snapshot_manifest_t * manifest = (fd_snapshot_manifest_t * )fd_chunk_to_laddr_const( ctx->manifest_out.wksp, ctx->manifest_out.chunk );
```
mmcgee-jump:
```suggestion
    fd_snapshot_manifest_t const * manifest = fd_chunk_to_laddr_const( ctx->manifest_out.wksp, ctx->manifest_out.chunk );
```

src/discof/restore/fd_snapin_tile.c
```diff
@@ -89,14 +119,38 @@ metrics_write( fd_snapin_tile_t * ctx ) {
...
+}
+
 static void
 manifest_cb( void * _ctx ) {
   fd_snapin_tile_t * ctx = (fd_snapin_tile_t*)_ctx;
 
+  ulong sz = sizeof(fd_snapshot_manifest_t);
+
+  if( FD_LIKELY( ctx->hash_info.enabled ) ) {
+    fd_snapshot_manifest_t const * manifest = (fd_snapshot_manifest_t const * )fd_chunk_to_laddr_const( ctx->manifest_out.wksp, ctx->manifest_out.chunk );
```
mmcgee-jump:
```suggestion
    fd_snapshot_manifest_t const * manifest = fd_chunk_to_laddr_const( ctx->manifest_out.wksp, ctx->manifest_out.chunk );
```

src/discof/restore/fd_snaplt_tile.c
```diff
@@ -0,0 +1,297 @@
...
+  (void)ctl;
+
+  FD_TEST( ctx->state!=FD_SNAPLT_STATE_SHUTDOWN );
+  if( FD_LIKELY( sz ) ) FD_TEST( chunk>=ctx->in.chunk0 && chunk<=ctx->in.wmark && sz<=ctx->in.mtu );
+
+  switch( sig ) {
+    case FD_SNAPSHOT_HASH_MSG_SUB: {
+      FD_TEST( ctx->state==FD_SNAPLT_STATE_HASHING );
+
+      fd_snapshot_existing_account_t * prev_acc = (fd_snapshot_existing_account_t *)fd_chunk_to_laddr_const( ctx->in.wksp, chunk );
```
mmcgee-jump:
```suggestion
      fd_snapshot_existing_account_t const * prev_acc = fd_chunk_to_laddr_const( ctx->in.wksp, chunk );
```

src/discof/restore/fd_snapin_tile.c
```diff
@@ -143,6 +216,14 @@ account_cb( void *                          _ctx,
   ctx->metrics.accounts_inserted++;
 
   if( FD_LIKELY( should_publish ) ) fd_funk_rec_publish( ctx->funk, prepare );
+
+  if( FD_LIKELY( ctx->hash_info.enabled ) ) {
+    /* send account hdr to snaplt tile */
+    fd_snapshot_account_t * account = (fd_snapshot_account_t *)fd_chunk_to_laddr( ctx->hash_out.wksp, ctx->hash_out.chunk );
```
mmcgee-jump: etc, try fix these everywhere

src/discof/restore/fd_snapin_tile.c
```diff
@@ -245,18 +339,57 @@ handle_control_frag( fd_snapin_tile_t *  ctx,
         break;
       }
 
-      if( FD_LIKELY( ctx->funk_txn ) ) fd_funk_txn_publish_into_parent( ctx->funk, ctx->funk_txn, 0 );
-      fd_stem_publish( stem, 0UL, fd_ssmsg_sig( FD_SSMSG_DONE ), 0UL, 0UL, 0UL, 0UL, 0UL );
+      if( FD_LIKELY( ctx->hash_info.enabled ) ) {
+        fd_stem_publish( stem, FD_SNAPIN_HSH_IDX, FD_SNAPSHOT_HASH_MSG_FINI, 0UL, 0UL, 0UL, 0UL, 0UL );
```
mmcgee-jump: Can't you just always pass these ctrl messages through, same way e.g. snapdc does? Also does this change STEM_BURST ?

src/discof/restore/fd_snapin_tile.c
```diff
@@ -245,18 +339,57 @@ handle_control_frag( fd_snapin_tile_t *  ctx,
...
     default:
       FD_LOG_ERR(( "unexpected control sig %lu", sig ));
       return;
   }
 
+  /* snapin waits for result lthashes to come back from the snaplt tiles.
+     Until these hashes come back, snapin cannot ack the control message
+     sent from snaprd because snaprd would advance before snapin
+     is ready to receive another snapshot byte stream or shutdown. */
+  if( FD_UNLIKELY( ctx->pending_ack ) ) return;
```
mmcgee-jump, line 382: I don't really like the separate control flow here. Why not just stick the snaplt tiles on the same ACK bus that's controlled by snaprd?

mmcgee-jump (changes requested):

src/discof/restore/fd_snapin_tile.c
cali-jumptrading: Done

src/discof/restore/fd_snapin_tile.c
cali-jumptrading: Done, I changed STEM_BURST to 3 to account for one account subtraction, one account header message, and one malformed message.

cali-jumptrading:
superseded by https://github.com/firedancer-io/firedancer/pull/7064

PR #5957, opened 2025-08-07
quic: only dealloc timed out conns when needed

akhinvasara-jumptrading:
This PR adds a new mode of connection idle timeout handling. Controlled by config->keep_timed_out, instead of fully freeing the connection, we keep it around while there is capacity. Other than the conn object itself and the tls_hs, we free all resources. This connection can be revived if it receives a new incoming packet, but we don't guarantee delivery of data that was sent before the connection timed out. This mode is turned on for the quic tile. 

It also frees pkt meta when a connection is deallocated. And goes back to removing conn ids from the map when it calls conn_free.


src/waltz/quic/fd_quic.c
```diff
@@ -4151,6 +4214,23 @@ fd_quic_conn_free( fd_quic_t *      quic,
...
+void
+fd_quic_conn_free( fd_quic_t *      quic,
+                   fd_quic_conn_t * conn ) {
+
+  if( FD_UNLIKELY( fd_quic_conn_free_resources( quic, conn ) ) ) {
+    return;
+  }
+
+  FD_COMPILER_MFENCE();
+  fd_quic_set_conn_state( conn, FD_QUIC_CONN_STATE_INVALID );
```
nbridge-jump, line 4228: These fences aren't needed. This memory isn't shared

akhinvasara-jumptrading: potentially useful for quic trace or debugging a dump? can remove if you feel strongly, but doesn't seem like it has a big perf impact here? 
(also maintaining current behavior)

nbridge-jump: Valid.
We should probably have it in fd_quic_set_conn_state

src/waltz/quic/fd_quic_conn.h
```diff
@@ -18,8 +18,10 @@
 #define FD_QUIC_CONN_STATE_ABORT              5 /* connection terminating due to error */
 #define FD_QUIC_CONN_STATE_CLOSE_PENDING      6 /* connection is closing */
 #define FD_QUIC_CONN_STATE_DEAD               7 /* connection about to be freed */
+#define FD_QUIC_CONN_STATE_TIMED_OUT          8 /* connection timed out but kept for reuse */
+#define FD_QUIC_CONN_STATE_COUNT              9
 
-FD_STATIC_ASSERT( FD_QUIC_CONN_STATE_DEAD+1 == sizeof(((fd_quic_metrics_t*)0)->conn_state_cnt)/sizeof(((fd_quic_metrics_t*)0)->conn_state_cnt[0]),
```
nbridge-jump, line 22: I prefer `(fd_quic_metrics_t){0}.conn_state_cnt`, which doesn't refer to a zero address.

nbridge-jump (approved):

nbridge-jump (approved):

PR #5962, opened 2025-08-07
flamenco, fuzz: consolidate fuzzing harnesses with production code

mjain-jump:
Removes dead fuzz-specific code that was previously used in ledger replay

### DIFF
--- a/src/flamenco/runtime/fd_runtime.c
+++ b/src/flamenco/runtime/fd_runtime.c
@@ -185,6 +185,24 @@ fd_runtime_update_leaders( fd_bank_t * bank,
 /* Various Private Runtime Helpers                                            */
 /******************************************************************************/
 
+static fd_funk_txn_t *
+fd_runtime_funk_txn_get( fd_funk_t * funk,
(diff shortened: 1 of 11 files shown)


yufeng-jump (approved):
nice cleanup

src/flamenco/runtime/fd_runtime.c
```diff
@@ -1409,6 +1394,9 @@ fd_runtime_finalize_txn( fd_funk_t *         funk,
         FD_LOG_CRIT(( "fd_runtime_finalize_txn: failed to join account at idx %u", i ));
       }
 
+      /* Reclaim any accounts that have 0-lamports */
+      fd_executor_reclaim_account( txn_ctx, &txn_ctx->accounts[i] );
```
topointon-jump, line 1398: Why do we need this special-case logic for 0-lamport accounts? What does reclaiming achieve? Wonder if we still need this..

mjain-jump: Good point. It's old code that I didn't bother touching

PR #5965, opened 2025-08-07
bundle: fix cast

jherrera-jump:
fixes unable to compile with `FD_DISABLE_OPTIMIZATION=1` on gcc 13


akhinvasara-jumptrading (approved):
noice, thanks!

Issue #5299, opened 2025-06-06
[CI] Add a builds.yml option to check only


ripatel-fd:
Full nightly builds take a long time

Would be useful to run `make check` only for the compiler matrix in `builds.yml`, which takes a couple seconds only

PR #5966, opened 2025-08-07
ci: add a make check for all machine and compilers

anwayde:
Closes https://github.com/firedancer-io/firedancer/issues/5299


.github/workflows/tests.yml
```diff
@@ -18,22 +18,13 @@ jobs:
...
         # Attach additional params to machine types
         include:
-          - test-case: linux_gcc_noarch64
-            machine: linux_gcc_noarch64
-            label: X64
-            deps-extras: "+dev"
-            targets: "check"
-            compiler: gcc
-            compiler-version: 11.4.0
           - test-case: linux_gcc_x86_64
```
ripatel-fd, line 26: remove this one

0x0ece (approved):

.github/workflows/builds.yml
```diff
@@ -139,16 +153,22 @@ jobs:
...
           # dry-run
           if [ "${{ inputs.dry_run }}" == "true" ]; then
             ARGS="$ARGS --dry-run"
           fi
-          # verbose
-          if [ "${{ inputs.verbose }}" == "true" ]; then
-            ARGS="$ARGS --verbose"
+          # check-run
+          if [ "${{ inputs.check_run }}" == "true" ]; then
+            ARGS="$ARGS --check-run --no-deps"
```
ripatel-fd, line 171: @anwayde Why do we do `--no-deps` here despite installing deps above?

.github/workflows/tests.yml
```diff
@@ -80,13 +51,6 @@ jobs:
             compiler-version: 15.0.6
             run-unit-tests: true
             run-integration-tests: true
-          - test-case: native-no-deps
```
ripatel-fd, line 83: Hmm, maybe we should bring this back. native already covers AVX512 afaik, so perhaps we run the icelake target without deps. It would be good to have at least one run that runs without dependencies but tries to link all binaries. In case we have some bug where compiling works but linking fails.

But it probably doesn't matter, if dependencies are missing the build would fail on missing includes before it tries to link.

.github/workflows/builds.yml
anwayde: It tells the `builds.sh` script to not install dependencies during the build (as we handle it in the workflow).

PR #5967, opened 2025-08-07
stake_ci: add APIs for updating or removing individual contact info entries

ravyu-jump:
As a part of GossipV2 rollout

### DIFF
--- a/src/disco/shred/fd_stake_ci.c
+++ b/src/disco/shred/fd_stake_ci.c
@@ -1,4 +1,5 @@
 #include "fd_stake_ci.h"
+#include "fd_shred_dest.h"
 #include "../../util/net/fd_ip4.h" /* Just for debug */
 
 #define SORT_NAME sort_pubkey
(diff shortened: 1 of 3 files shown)


src/disco/shred/fd_stake_ci.c
```diff
@@ -390,6 +391,103 @@ fd_stake_ci_set_identity( fd_stake_ci_t *     info,
...
+                         fd_pubkey_t const *   pubkey,
+                         uint                  ip4,
+                         ushort                port ) {
+  /* per epoch */
+  ci_dest_update_impl( info, pubkey, ip4, port, info->epoch_info + 0UL );
+  ci_dest_update_impl( info, pubkey, ip4, port, info->epoch_info + 1UL );
+}
+
+void
+fd_stake_ci_dest_remove( fd_stake_ci_t * info,
```
mmcgee-jump, line 481:
```suggestion
fd_stake_ci_dest_remove( fd_stake_ci_t *     info,
```

src/disco/shred/fd_stake_ci.c
```diff
@@ -390,6 +391,103 @@ fd_stake_ci_set_identity( fd_stake_ci_t *     info,
...
+                         fd_pubkey_t const * pubkey ) {
+  /* per epoch */
```
mmcgee-jump:
```suggestion
```

src/disco/shred/fd_stake_ci.c
```diff
@@ -390,6 +391,103 @@ fd_stake_ci_set_identity( fd_stake_ci_t *     info,
...
+  ci_dest_remove_impl( info, pubkey, info->epoch_info + 0UL );
+  ci_dest_remove_impl( info, pubkey, info->epoch_info + 1UL );
+
```
mmcgee-jump, line 485:
```suggestion
```

src/disco/shred/fd_stake_ci.c
```diff
@@ -390,6 +391,103 @@ fd_stake_ci_set_identity( fd_stake_ci_t *     info,
...
+  /* per epoch */
+  ci_dest_update_impl( info, pubkey, ip4, port, info->epoch_info + 0UL );
+  ci_dest_update_impl( info, pubkey, ip4, port, info->epoch_info + 1UL );
+}
+
+void
+fd_stake_ci_dest_remove( fd_stake_ci_t * info,
+                         fd_pubkey_t const * pubkey ) {
+  /* per epoch */
+  ci_dest_remove_impl( info, pubkey, info->epoch_info + 0UL );
```
mmcgee-jump:
```suggestion
  ci_dest_remove_impl( info, pubkey, info->epoch_info+0UL );
```

src/disco/shred/fd_stake_ci.h
```diff
@@ -152,6 +160,18 @@ void                       fd_stake_ci_stake_msg_fini( fd_stake_ci_t * info
 fd_shred_dest_weighted_t * fd_stake_ci_dest_add_init ( fd_stake_ci_t * info                                    );
 void                       fd_stake_ci_dest_add_fini ( fd_stake_ci_t * info, ulong                         cnt );
 
+/* Firedancer only:
+   The full client's Gossip update model publishes individual contact
+   info updates (update/insert or remove), which requires a different
+   set of dest_ APIs.
+
+   fd_stake_ci_dest_update updates (or adds, if necessary) a shred dest
+   entry. ip4 is in net order, port is in host order and are both
```
mmcgee-jump, line 169: I don't see why it matters if they are zero / non-zero?

src/disco/shred/fd_stake_ci.h
```diff
@@ -152,6 +160,18 @@ void                       fd_stake_ci_stake_msg_fini( fd_stake_ci_t * info
...
+   assumed to be non-zero */
```
mmcgee-jump:
```suggestion
   assumed to be non-zero. */
```

src/disco/shred/fd_stake_ci.c
```diff
@@ -390,6 +391,103 @@ fd_stake_ci_set_identity( fd_stake_ci_t *     info,
...
+  dest->ip4                       = ip4;
+  dest->port                      = port;
+}
+
+void
+ci_dest_remove_impl( fd_stake_ci_t *       info,
+                     fd_pubkey_t const *   pubkey,
+                     fd_per_epoch_info_t * ei ) {
+  fd_shred_dest_idx_t idx = fd_shred_dest_pubkey_to_idx( ei->sdest, (fd_pubkey_t *)pubkey );
+  if( idx==FD_SHRED_DEST_NO_DEST ) return;
```
mmcgee-jump:
```suggestion
  if( FD_UNLIKELY( idx==FD_SHRED_DEST_NO_DEST ) ) return;
```

src/disco/shred/fd_stake_ci.c
```diff
@@ -390,6 +391,103 @@ fd_stake_ci_set_identity( fd_stake_ci_t *     info,
...
+  }
+  /* TODO: Alternative batched copy using memcpy. Check with Philip if this is safe */
+  // fd_shred_dest_weighted_t * cur_dest = ei->sdest->all_destinations;
+  // fd_memcpy( info->shred_dest_temp, cur_dest, sizeof(fd_shred_dest_weighted_t)*(idx) );
+  // fd_memcpy( info->shred_dest_temp + idx, cur_dest + idx + 1UL, sizeof(fd_shred_dest_weighted_t)*(cur_cnt - idx - 1UL) );
+  refresh_sdest( info, info->shred_dest_temp, cur_cnt - 1UL, fd_shred_dest_cnt_staked( ei->sdest ), ei );
+}
+
+void
+fd_stake_ci_dest_update( fd_stake_ci_t *       info,
```
mmcgee-jump, line 472:
```suggestion
fd_stake_ci_dest_update( fd_stake_ci_t *    info,
```

src/disco/shred/fd_stake_ci.c
```diff
@@ -390,6 +391,103 @@ fd_stake_ci_set_identity( fd_stake_ci_t *     info,
...
+  ulong cur_cnt = fd_shred_dest_cnt_all( ei->sdest );
+  for( ulong i=0UL, j=0UL; i<cur_cnt; i++ ) {
+    if( FD_UNLIKELY( i==idx ) ) continue;
+    info->shred_dest_temp[ j++ ] = *fd_shred_dest_idx_to_dest( ei->sdest, (fd_shred_dest_idx_t) i );
+  }
+  /* TODO: Alternative batched copy using memcpy. Check with Philip if this is safe */
+  // fd_shred_dest_weighted_t * cur_dest = ei->sdest->all_destinations;
+  // fd_memcpy( info->shred_dest_temp, cur_dest, sizeof(fd_shred_dest_weighted_t)*(idx) );
+  // fd_memcpy( info->shred_dest_temp + idx, cur_dest + idx + 1UL, sizeof(fd_shred_dest_weighted_t)*(cur_cnt - idx - 1UL) );
+  refresh_sdest( info, info->shred_dest_temp, cur_cnt - 1UL, fd_shred_dest_cnt_staked( ei->sdest ), ei );
```
mmcgee-jump:
```suggestion
  refresh_sdest( info, info->shred_dest_temp, cur_cnt-1UL, fd_shred_dest_cnt_staked( ei->sdest ), ei );
```

src/disco/shred/fd_stake_ci.c
```diff
@@ -390,6 +391,103 @@ fd_stake_ci_set_identity( fd_stake_ci_t *     info,
...
+  refresh_sdest( info, info->shred_dest_temp, cur_cnt, fd_shred_dest_cnt_staked( ei->sdest ), ei );
+}
+
+void
+ci_dest_update_impl( fd_stake_ci_t *       info,
+                     fd_pubkey_t const *   pubkey,
+                     uint                  ip4,
+                     ushort                port,
+                     fd_per_epoch_info_t * ei ) {
+  fd_shred_dest_idx_t idx = fd_shred_dest_pubkey_to_idx( ei->sdest, (fd_pubkey_t *)pubkey );
```
mmcgee-jump: Can you change `fd_shred_dest_pubkey_to_idx` to take a `fd_pubkey_t const *` ?

src/disco/shred/fd_stake_ci.c
```diff
@@ -390,6 +391,103 @@ fd_stake_ci_set_identity( fd_stake_ci_t *     info,
...
+  if( FD_UNLIKELY( ei->sdest==NULL ) ) {
+    FD_LOG_ERR(( "Too many validators have higher stake than this validator.  Cannot continue." ));
+  }
+}
+
+void
+ci_dest_add_one_unstaked( fd_stake_ci_t *            info,
+                          fd_shred_dest_weighted_t * new_entry,
+                          fd_per_epoch_info_t *      ei ) {
+  if( fd_shred_dest_cnt_all( ei->sdest )+ 1UL>=MAX_SHRED_DESTS ) {
```
mmcgee-jump: Isn't this off by one? `fd_shred_dest_cnt_all( ei->sdest )>=MAX_SHRED_DESTS` would be the right bound

src/disco/shred/fd_stake_ci.c
```diff
@@ -390,6 +391,103 @@ fd_stake_ci_set_identity( fd_stake_ci_t *     info,
...
+    FD_LOG_ERR(( "too many validators in shred table to add a new validator" ));
```
mmcgee-jump: Don't log error here, this is an obvious attacker DoS. You can just not add the unstaked node

src/disco/shred/fd_stake_ci.c
```diff
@@ -390,6 +391,103 @@ fd_stake_ci_set_identity( fd_stake_ci_t *     info,
...
+               ulong                      cnt,
+               ulong                      staked_cnt,
+               fd_per_epoch_info_t *      ei ) {
+  sort_pubkey_inplace( shred_dest_temp + staked_cnt, cnt - staked_cnt );
+
+  fd_shred_dest_delete( fd_shred_dest_leave( ei->sdest ) );
+  ei->sdest = fd_shred_dest_join( fd_shred_dest_new( ei->_sdest, shred_dest_temp, cnt, ei->lsched, info->identity_key,
+                                                     ei->excluded_stake ) );
+  if( FD_UNLIKELY( ei->sdest==NULL ) ) {
+    FD_LOG_ERR(( "Too many validators have higher stake than this validator.  Cannot continue." ));
```
mmcgee-jump, line 405: Where does this message come from, what does that even mean? I think this join can never fail? You could probably just `FD_TEST( ei->sdest )` unless this message means something I don't understand

src/disco/shred/fd_stake_ci.c
```diff
@@ -1,5 +1,6 @@
 #include "fd_stake_ci.h"
 #include "../../util/net/fd_ip4.h" /* Just for debug */
+#include "fd_shred_dest.h"
```
mmcgee-jump, line 2: nit: move one line up above the `fd_ip4.h` include

src/disco/shred/fd_stake_ci.c
```diff
@@ -390,6 +391,103 @@ fd_stake_ci_set_identity( fd_stake_ci_t *     info,
   *info->identity_key = *identity_key;
 }
 
+void
+refresh_sdest( fd_stake_ci_t *            info,
```
mmcgee-jump, line 395: Can you add some tests for the add / update case?

src/disco/shred/fd_stake_ci.c
ravyu-jump: I ripped this one off from `fd_stake_ci_dest_add_fini_impl`

```c
  if( FD_UNLIKELY( ei->sdest==NULL ) ) {
    /* Happens if the identity key is not present, which can only happen
       if the current validator's stake is not in the top 40,200.  We
       could initialize ei->sdest to a dummy value, but having the wrong
       stake weights could lead to potentially slashable issues
       elsewhere (e.g. we might product a block when we're not actually
       leader).  We're just going to terminate in this case. */
    FD_LOG_ERR(( "Too many validators have higher stake than this validator.  Cannot continue." ));
  }
```

src/disco/shred/fd_stake_ci.c
ravyu-jump: Yeah makes sense, there's a bunch of similar LOG_ERRs in the functions above should probably remove them as well?

src/disco/shred/fd_stake_ci.c
ravyu-jump: why?

mmcgee-jump: It's not aligned?

src/disco/shred/fd_stake_ci.h
ravyu-jump: Zero `ip4_port_t` value is treated as a null destination. In the unstaked case we don't want to add such an entry to the shuffle, right? In the staked case I agree this does not matter

PR #5971, opened 2025-08-07
correct shred_repair mtu

akhinvasara-jumptrading:
Shred_repair mtu was too small. Debug credit @mmcgee-jump @ravyu-jump


src/disco/fd_disco_base.h
```diff
@@ -55,8 +55,8 @@
 /* FD_SHRED_REPAIR_MTU is the maximum size of a frag on the shred_repair
    link.  This is the size of a data shred header + merkle root. */
```
ravyu-jump:
```suggestion
   link.  This is the size of a data shred header + merkle root. 
   + chained merkle root*/
```

PR #5980, opened 2025-08-08
ipecho: add ipecho client, server, tile, and shred version commands

mmcgee-jump:


src/discof/ipecho/fd_ipecho_server.c
```diff
@@ -0,0 +1,335 @@
...
+  conn->response_bytes_written += (ulong)sz;
+  if( FD_UNLIKELY( conn->response_bytes_written<sizeof(conn->response_bytes) ) ) return;
+
+  close_conn( server, conn_idx, CLOSE_OK );
+}
+
+void
+fd_ipecho_server_poll( fd_ipecho_server_t * server,
+                       int *                charge_busy ) {
+  int nfds = fd_syscall_poll( server->pollfds, (uint)( server->max_connection_cnt+1UL ), 0 );
```
ripatel-fd: The IP echo server is almost always idle. Use a larger timeout here (like cswtch)

PR #5981, opened 2025-08-08
flamenco, program-cache: join syscall map

two-heart:
esp. with util/tmpl we should stick close to the API contracts for maintainability

### DIFF
--- a/src/flamenco/runtime/program/fd_program_cache.c
+++ b/src/flamenco/runtime/program/fd_program_cache.c
@@ -260,7 +260,8 @@ fd_program_cache_validate_sbpf_program( fd_exec_slot_ctx_t const * slot_ctx,
 
   /* Allocate syscalls */
 
-  fd_sbpf_syscalls_t * syscalls = fd_sbpf_syscalls_new( fd_spad_alloc( runtime_spad, fd_sbpf_syscalls_align(), fd_sbpf_syscalls_footprint() ) );
+  void               * syscalls_mem = fd_spad_alloc( runtime_spad, fd_sbpf_syscalls_align(), fd_sbpf_syscalls_footprint() );
+  fd_sbpf_syscalls_t * syscalls     = fd_sbpf_syscalls_join( fd_sbpf_syscalls_new( syscalls_mem ) );
   if( FD_UNLIKELY( !syscalls ) ) {
     FD_LOG_CRIT(( "Call to fd_sbpf_syscalls_new() failed" ));
   }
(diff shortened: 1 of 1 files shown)


src/flamenco/runtime/program/fd_program_cache.c
```diff
@@ -275,6 +276,7 @@ fd_program_cache_validate_sbpf_program( fd_exec_slot_ctx_t const * slot_ctx,
   if( FD_UNLIKELY( 0!=fd_sbpf_program_load( prog, program_data, program_data_len, syscalls, false ) ) ) {
     FD_LOG_DEBUG(( "fd_sbpf_program_load() failed: %s", fd_sbpf_strerror() ));
     cache_entry->failed_verification = 1;
+    fd_sbpf_syscalls_leave( syscalls );
```
mjain-jump, line 280: i dont think you need this

src/flamenco/runtime/program/fd_program_cache.c
```diff
@@ -316,6 +318,7 @@ fd_program_cache_validate_sbpf_program( fd_exec_slot_ctx_t const * slot_ctx,
   }
 
   int res = fd_vm_validate( vm );
+  fd_sbpf_syscalls_leave( syscalls );
```
mjain-jump, line 322: same here

src/flamenco/runtime/program/fd_program_cache.c
ripatel-fd: Well, it's good code style. Ideally, we'd destroy the object (delete, leave) when it's no longer used.

src/flamenco/runtime/program/fd_program_cache.c
```diff
@@ -260,7 +260,8 @@ fd_program_cache_validate_sbpf_program( fd_exec_slot_ctx_t const * slot_ctx,
 
   /* Allocate syscalls */
 
-  fd_sbpf_syscalls_t * syscalls = fd_sbpf_syscalls_new( fd_spad_alloc( runtime_spad, fd_sbpf_syscalls_align(), fd_sbpf_syscalls_footprint() ) );
+  void               * syscalls_mem = fd_spad_alloc( runtime_spad, fd_sbpf_syscalls_align(), fd_sbpf_syscalls_footprint() );
```
ripatel-fd:
```suggestion
  void               * syscalls_mem = fd_spad_alloc_check( runtime_spad, fd_sbpf_syscalls_align(), fd_sbpf_syscalls_footprint() );
```

There's pretty much never a reason to ignore the spad alloc return code in our runtime code. Because the only reason you'd ever omit this error checking is for extreme performance (i.e. it happens >30 million times per second)

two-heart: Agree, added

ripatel-fd:
@mjain-jump could you take another look and merge it if it looks good to you?

PR #5982, opened 2025-08-08
snapshots: add initial peer selection

alpeng-jump:


src/disco/topo/fd_topo.h
```diff
@@ -479,12 +479,13 @@ struct fd_topo_tile {
...
     struct {
       char  snapshots_path[ PATH_MAX ];
-      char  cluster[ 8UL ];
       int   incremental_snapshot_fetch;
       int   do_download;
       uint  maximum_local_snapshot_age;
       uint  minimum_download_speed_mib;
       uint  maximum_download_retry_abort;
+      ulong initial_peers_cnt;
+      fd_ip4_port_t initial_peers[ 16 ];
```
cali-jumptrading, line 488: nit: alignment, also 16UL

src/app/shared/fd_config_parse.c
```diff
@@ -5,13 +5,20 @@
...
     FD_LOG_ERR(( "[config->paths.snapshots] is too long (max %lu)", PATH_MAX-1UL ));
   }
   if( FD_UNLIKELY( config->paths.snapshots[ 0 ]!='\0' && config->paths.snapshots[ 0 ]!='/' ) ) {
     FD_LOG_ERR(( "[config->paths.snapshots] must be an absolute path and hence start with a '/'"));
   }
+  if( FD_UNLIKELY( config_f->snapshots.download && config_f->snapshots.initial_peers_cnt==0UL ) ) {
+    FD_LOG_ERR(( "[configf->snapshots.download] is true but [configf->snapshots.initial_peers] is not set. "
+                 "Must specify at least one initial peer to download from if download is enabled" ));
+  }
+  if( FD_UNLIKELY( config_f->snapshots.initial_peers_cnt!=0UL && !config_f->snapshots.download ) ) {
```
cali-jumptrading, line 18: I don't think we need this check because there might be a scenario where we download from initial peers if the snapshot on disk is corrupted.

src/app/firedancer/config/default.toml
```diff
@@ -282,7 +282,8 @@ user = ""
     # present.
     #
     # TODO: implement the "at most one downloaded " logic. "
-    download = true
+    # TODO: default to true once integrated with gossip
+    download = false
```
cali-jumptrading, line 286: we should keep this as true because the snapshot loader has no way of knowing whether the snapshots on disk are recent right now.

src/app/firedancer/topology.c
```diff
@@ -176,22 +176,70 @@ resolve_gossip_entrypoints( config_t * config ) {
...
 setup_snapshots( config_t *       config,
                  fd_topo_tile_t * tile ) {
   fd_memcpy( tile->snaprd.snapshots_path, config->paths.snapshots, PATH_MAX );
-  fd_memcpy( tile->snaprd.cluster, config->firedancer.snapshots.cluster, sizeof(tile->snaprd.cluster) );
   tile->snaprd.incremental_snapshot_fetch   = config->firedancer.snapshots.incremental_snapshots;
   tile->snaprd.do_download                  = config->firedancer.snapshots.download;
   tile->snaprd.maximum_local_snapshot_age   = config->firedancer.snapshots.maximum_local_snapshot_age;
   tile->snaprd.minimum_download_speed_mib   = config->firedancer.snapshots.minimum_download_speed_mib;
   tile->snaprd.maximum_download_retry_abort = config->firedancer.snapshots.maximum_download_retry_abort;
+  tile->snaprd.initial_peers_cnt            = fd_ulong_min( config->firedancer.snapshots.resolved_initial_peers_cnt, 16UL );
```
cali-jumptrading, line 234: Should we error out if there are more than 16 initial peers given?

src/app/firedancer/topology.c
```diff
@@ -176,22 +176,70 @@ resolve_gossip_entrypoints( config_t * config ) {
   config->gossip.resolved_entrypoints_cnt = resolved_entrypoints;
 }
 
+static int
+resolve_initial_peer( char const *    host_port,
```
cali-jumptrading, line 180: could we reuse `resolve_gossip_entrypoint` here? Or... rename it to `resolve_entrypoint`? or `resolve_peer`?

src/app/shared/fd_config_parse.c
alpeng-jump: In that scenario, snapshots.download is true

src/app/firedancer/topology.c
alpeng-jump: We should still proceed since those are valid endpoints. This mirrors gossip entrypoint's approach as well.

src/app/firedancer/topology.c
alpeng-jump: We don't do a byteswap in resolve_initial_peer. Also, the initial peer set is likely to be removed once we hook it up with gossip.

src/app/firedancer/config/default.toml
alpeng-jump: Having download as true involves setting an initial peer in the default config, which assumes a cluster. This needs to be a trusted peer too. This is partly the reason we removed hardcoded entrypoints in the first place. The user knows the target cluster and should set this field. 

Second, I believe the download option isn't actually used anywhere in snapshot code apart from the config check I added.

src/app/shared/fd_config_parse.c
cali-jumptrading: oh right, ignore me

src/app/firedancer/config/default.toml
cali-jumptrading: oh you are right, but we should check the download config option in the snapshot tiles.

PR #5983, opened 2025-08-08
flamenco: cleanup unused / dirty `fd_executor` functions

mjain-jump:


src/flamenco/runtime/program/fd_system_program.h
```diff
@@ -42,14 +47,27 @@ int fd_system_program_exec_upgrade_nonce_account   ( fd_exec_instr_ctx_t * ctx
 
 /* User APIs */
 
-/* fd_check_transaction_age returns 0 if the transactions age is
+/* `fd_check_transaction_age()` returns 0 if the transactions age is
```
ripatel-fd: Can you revert this, we don't use this doc style ... If we do hyperlink kind stuff in the future, it should be Doxygen compatible

mjain-jump: done

PR #5984, opened 2025-08-08
Remove valloc usages

ripatel-fd:
- Use spad/malloc/scratch alloc directly instead of valloc for
  various old code
- Re-enable test_types_walk
- Remove unnecessary allocs in VM test code
- Remove instruction context new/join/leave/delete boilerplate


src/flamenco/gossip/fd_gossip_spy.c
```diff
@@ -243,21 +241,22 @@ main( int     argc,
 
   config.shred_version = 4274;
 
+  void * yaml_mem = aligned_alloc( fd_flamenco_yaml_align(), fd_flamenco_yaml_footprint() );
+  FD_TEST( yaml_mem );
```
ibhatt-jumptrading, line 245: technically FD_TEST shouldn't be used outside of a test but we violated that everywhere so not a big deal

src/flamenco/gossip/fd_gossip_spy.c
```diff
@@ -277,7 +276,7 @@ main( int     argc,
   //   return 1;
   // if ( fd_gossip_add_active_peer(glob, resolve_hostport("entrypoint3.testnet.solana.com:8001", &peeraddr)) )
   //   return 1;
-  if ( fd_gossip_add_active_peer(glob, resolve_hostport(":1024", &peeraddr)) )
+  if( fd_gossip_add_active_peer(glob, resolve_hostport(":1024", &peeraddr)) )
```
ibhatt-jumptrading: fd_gossip_add_active_peer( glob, resolve_hostport( ":1024", &peeraddr ) )

jumpsiegel (dismissed):

src/flamenco/gossip/fd_gossip_spy.c
ripatel-fd: I agree ... But i think Ravyu is about to delete this file

lidatong (approved):

PR #5992, opened 2025-08-09
flame: fixed perf->flamegraph to handle terminated processes

npatel-jump:


src/app/shared_dev/commands/flame.c
```diff
@@ -58,7 +71,7 @@ flame_cmd_fn( args_t *   args,
   install_parent_signals();
 
   fd_topo_join_workspaces( &config->topo, FD_SHMEM_JOIN_MODE_READ_ONLY );
-  fd_topo_fill( &config->topo );
+  fd_topo_fill_resilient( &config->topo );
```
mmcgee-jump, line 74: Why do you need this? The workspaces should still be valid?

src/app/shared_dev/commands/flame.c
```diff
@@ -97,29 +110,47 @@ flame_cmd_fn( args_t *   args,
...
+
+    /* Skip tiles that don't have valid PID/TID */
+    if( FD_UNLIKELY( !pid || !tid || pid > INT_MAX || tid > INT_MAX ) ) {
+      FD_LOG_WARNING(( "skipping tile %s:%lu - invalid PID/TID (pid=%lu, tid=%lu)", tile->name, tile->kind_id, pid, tid ));
+      continue;
+    }
 
-    FD_TEST( pid<=INT_MAX );
     if( FD_UNLIKELY( -1==kill( (int)pid, 0 ) ) ) {
-      if( FD_UNLIKELY( errno==ESRCH ) ) FD_LOG_ERR(( "tile %s:%lu is not running", config->topo.tiles[ i ].name, config->topo.tiles[ i ].kind_id ));
```
mmcgee-jump, line 112: I feel like changing this to WARNING should be pretty much the only change needed? Why all the other pid / tid bounds checks + resilient thing etc?

npatel-jump: The _fd_topo_workspace_fill_ does FD_TEST's itself that fail (fseq checks).

For PID checks, I was getting many PID/TID's = 0, and a tid=438430872235253 for writer:1, and filtered around them so perf wouln't err due to a failed attachment.
I also think you previously had a check for it, so I included a similar filter to just continue on without them.
`FD_TEST( pid<=INT_MAX );`

mmcgee-jump: What FD_TEST would fail? I don't think that seems right

npatel-jump: It's failing on:
`main src/disco/topo/fd_topo.c(178): tile net in_link_fseq[0] = (nil)
  main src/disco/topo/fd_topo.c(179): FAIL: tile->in_link_fseq[ j ]`

I'll take a deeper look into why it might be failing

mmcgee-jump: Are you providing the config file to the flamegraph tool?

npatel-jump: Yes, but I just realized I was using the wrong one.  That makes a lot more sense. Thank you, ill make the changes so it just fixes the warnings.

PR #5996, opened 2025-08-10
quic: use correct alignment for stream map

two-heart:
+ drive-by fixes: wrong doc, wrong time unit in string, wrong branch hint

### DIFF
--- a/src/waltz/quic/fd_quic.h
+++ b/src/waltz/quic/fd_quic.h
@@ -159,7 +159,7 @@ struct __attribute__((aligned(16UL))) fd_quic_config {
 #define FD_QUIC_CONFIG_LIST(X,...) \
   X( role,                        "%d",     enum,  "enum",         __VA_ARGS__ ) \
   X( retry,                       "%d",     bool,  "bool",         __VA_ARGS__ ) \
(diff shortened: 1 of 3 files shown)


src/waltz/quic/fd_quic.h
```diff
@@ -593,8 +593,8 @@ fd_quic_get_next_wakeup( fd_quic_t * quic );
 
 /* fd_quic_service services the next QUIC connection, including stream
```
akhinvasara-jumptrading: While we're here, let's tighten up docs even more - how about this?
"fd_quic_service services the next QUIC connection at each service level, including stream transmit ops..... (same stuff as now) ... Returns the number of connections serviced, where 0 means the call did no work.

akhinvasara-jumptrading (dismissed):
Left one minor suggestion on the docs, else lgtm! Thanks for the finds!

src/waltz/quic/fd_quic.h
two-heart: Sounds good to me; updated

akhinvasara-jumptrading (approved):
👍

PR #5801, opened 2025-07-24
Improve additional shred destinations

HGuillemet:
* allow multiple destinations
* separate received and leader shreds

Fixes #5509

PR #5997, opened 2025-08-10
Fix additional shred destinations

HGuillemet:
The last commit of PR #5801 was incomplete, causing the new config directives to be ignored. Sorry about this.

Also both leader and received shreds were sent to `additional_shred_destinations_retransmit`. I believe it's better to keep both streams disjoint and only send received shreds to these destinations.


ripatel-fd:
Thank you @HGuillemet. We're going to look into this on Monday.

HGuillemet:
Apparently this fix has been forgotten for v705.

ripatel-fd:
> Apparently this fix has been forgotten for v705.

Not forgotten, we just didn't get a chance to get the fix in because we were working on other things. Thank you for the reminder.

anwayde (approved):

PR #6002, opened 2025-08-11
metrics (wip)

emwang-jump:


src/discof/forest/fd_forest.h
```diff
@@ -387,13 +396,23 @@ fd_forest_query( fd_forest_t * forest, ulong slot );
 fd_forest_blk_t *
 fd_forest_blk_insert( fd_forest_t * forest, ulong slot, ulong parent_slot );
 
+enum shred_src {
+  TURBINE   = 0,
+  REPAIR    = 1,
+  RECOVERED = 2,
+};
+typedef enum shred_src shred_src_t;
```
lidatong: avoid enum and just use multiple #defines (kevin has a rant about this)

src/discof/forest/fd_forest.h
```diff
@@ -54,12 +54,21 @@ struct __attribute__((aligned(128UL))) fd_forest_blk {
...
+  fd_forest_blk_idxs_t fecs[fd_forest_blk_idxs_word_cnt]; /* fec set idxs */
   fd_forest_blk_idxs_t idxs[fd_forest_blk_idxs_word_cnt]; /* data shred idxs */
   fd_forest_blk_idxs_t cmpl[fd_forest_blk_idxs_word_cnt]; /* last shred idx of every FEC set that has been completed by shred_tile */
 
   /* i.e. when fecs == cmpl, the slot is truly complete and everything
   is contained in fec store. Look at fec_clear for more details.*/
+
+  /* Metrics */
+
+  fd_forest_blk_idxs_t code[fd_forest_blk_idxs_word_cnt]; /* code shred idxs */
```
lidatong, line 66: is this still bounded to 32768?

src/discof/forest/fd_forest.c
```diff
@@ -448,6 +449,12 @@ acquire( fd_forest_t * forest, ulong slot, ulong parent_slot ) {
   fd_forest_blk_idxs_null( blk->idxs ); /* expensive */
   fd_forest_blk_idxs_null( blk->cmpl ); /* expensive */
 
+  fd_forest_blk_idxs_null( blk->code ); /* FIXME expensive */
+  blk->first_shred_ts = 0;
+  blk->first_req_ts   = 0;
+  blk->turbine_cnt = 0;
+  blk->repair_cnt  = 0;
```
lidatong: align

src/flamenco/repair/iptable.h
```diff
@@ -0,0 +1,2392 @@
+#pragma once
```
lidatong: hm... i think better not to commit this and instead it would be optionally available when running repair cmd (ie. repair should be able to work without this)

src/discof/forest/fd_forest.h
emwang-jump: `FD_SHRED_BLK_MAX` is also defined as the max parity shreds in a slot

src/app/firedancer-dev/commands/catchup.c
```diff
@@ -0,0 +1,199 @@
...
+    if( !swapped ) break;
+  }
+}
+
+#define MAX_WIDTH 120
+static char dashes[MAX_WIDTH + 1] = "========================================================================================================================";
+static char spaces[MAX_WIDTH + 1] = "                                                                                                                        ";
+
+static void
+print_catchup_slots( fd_wksp_t * repair_tile_wksp, fd_repair_tile_ctx_t * repair_ctx ) {
```
lidatong: so i was actually thinking this would be its own module API (fd_catchup or something like that), and the data structure would be encapsulated within that

it would like in repair tile ctx's mem -- but i like the idea of a subcommand to query it

PR #6003, opened 2025-08-11
gossip2, snaprd: plumbing for gossip2 to snaprd tile contact info updates

ravyu-jump:


src/discof/restore/fd_snaprd_tile.c
```diff
@@ -489,50 +540,82 @@ after_frag( fd_snaprd_tile_t *  ctx,
...
-           fragment, and need to wait for acknowledges to come back
-           first, to ensure there's only one control message outstanding
-           at a time. */
-        ctx->malformed = 1;
+  if( FD_UNLIKELY( ctx->in_kind[ in_idx ]==IN_KIND_GOSSIP ) ) {
+    fd_gossip_update_message_t * msg = &ctx->gossip.tmp_upd_buf;
+    switch( msg->tag ) {
+      case FD_GOSSIP_UPDATE_TAG_CONTACT_INFO: {
+          fd_contact_info_upd_t * cur      = &ctx->gossip.ci_table[ msg->contact_info.idx ];
+          fd_ip4_port_t       cur_addr = ctx->gossip.ci_table[ msg->contact_info.idx ].sockets[ FD_CONTACT_INFO_SOCKET_RPC ];
```
cali-jumptrading: nit: alignment

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -489,50 +540,82 @@ after_frag( fd_snaprd_tile_t *  ctx,
...
-                        FD_IP4_ADDR_FMT_ARGS( ctx->addr.addr ), ctx->addr.port ));
-        fd_sshttp_cancel( ctx->sshttp );
-        fd_ssping_invalidate( ctx->ssping, ctx->addr, fd_log_wallclock() );
-        /* We would like to transition to FULL_HTTP_RESET, but we can't
-           do it just yet, because we have already sent a DONE control
-           fragment, and need to wait for acknowledges to come back
-           first, to ensure there's only one control message outstanding
-           at a time. */
-        ctx->malformed = 1;
+  if( FD_UNLIKELY( ctx->in_kind[ in_idx ]==IN_KIND_GOSSIP ) ) {
```
cali-jumptrading, line 712: I kinda think you can get away with matching the `in_idx` with a predefined constant.

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -489,50 +540,82 @@ after_frag( fd_snaprd_tile_t *  ctx,
...
+    switch( msg->tag ) {
+      case FD_GOSSIP_UPDATE_TAG_CONTACT_INFO: {
+          fd_contact_info_upd_t * cur      = &ctx->gossip.ci_table[ msg->contact_info.idx ];
+          fd_ip4_port_t       cur_addr = ctx->gossip.ci_table[ msg->contact_info.idx ].sockets[ FD_CONTACT_INFO_SOCKET_RPC ];
+          if( cur_addr.l ){
+            fd_ssping_remove( ctx->ssping, cur_addr );
+          }
+          fd_contact_info_upd_t * new = msg->contact_info.contact_info;
+          fd_ip4_port_t new_addr  = new->sockets[ FD_CONTACT_INFO_SOCKET_RPC ];
+          if( new_addr.l ) {
```
cali-jumptrading: Also think we need to check if `new_addr` is equal to `cur_addr` so that we don't unnecessarily remove and add the same contact info.

This is what I have:
```
 if( cur_addr.l!=new_addr.l ) {
            fd_ssping_remove( ctx->ssping, cur_addr );

            if( new_addr.l ) {
              FD_LOG_INFO(("adding contact info for peer "FD_IP4_ADDR_FMT ":%hu ",
                              FD_IP4_ADDR_FMT_ARGS( new_addr.addr ), fd_ushort_bswap( new_addr.port ) ));
              fd_ssping_add( ctx->ssping, new_addr );
            }

        *cur = *new;
       }
 ```

File: src/discof/restore/fd_snaprd_tile.c [hunk 4]
```diff
@@ -98,6 +112,11 @@ struct fd_snaprd_tile {
     } incremental;
   } metrics;
 
+  struct {
+    fd_gossip_update_message_t tmp_upd_buf;
+    fd_contact_info_upd_t *    ci_table;
```
mmcgee-jump: Don't think this needs to be the full update? Just the IP address, right?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -489,50 +540,82 @@ after_frag( fd_snaprd_tile_t *  ctx,
...
-           first, to ensure there's only one control message outstanding
-           at a time. */
-        ctx->malformed = 1;
+  if( FD_UNLIKELY( ctx->in_kind[ in_idx ]==IN_KIND_GOSSIP ) ) {
+    fd_gossip_update_message_t * msg = &ctx->gossip.tmp_upd_buf;
+    switch( msg->tag ) {
+      case FD_GOSSIP_UPDATE_TAG_CONTACT_INFO: {
+          fd_contact_info_upd_t * cur      = &ctx->gossip.ci_table[ msg->contact_info.idx ];
+          fd_ip4_port_t       cur_addr = ctx->gossip.ci_table[ msg->contact_info.idx ].sockets[ FD_CONTACT_INFO_SOCKET_RPC ];
+          if( cur_addr.l ){
```
mmcgee-jump:
```suggestion
          if( FD_LIKELY( cur_addr.l ) ) fd_ssping_remove( ctx->ssping, cur_addr );
```

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -489,50 +540,82 @@ after_frag( fd_snaprd_tile_t *  ctx,
...
-           do it just yet, because we have already sent a DONE control
-           fragment, and need to wait for acknowledges to come back
-           first, to ensure there's only one control message outstanding
-           at a time. */
-        ctx->malformed = 1;
+  if( FD_UNLIKELY( ctx->in_kind[ in_idx ]==IN_KIND_GOSSIP ) ) {
+    fd_gossip_update_message_t * msg = &ctx->gossip.tmp_upd_buf;
+    switch( msg->tag ) {
+      case FD_GOSSIP_UPDATE_TAG_CONTACT_INFO: {
+          fd_contact_info_upd_t * cur      = &ctx->gossip.ci_table[ msg->contact_info.idx ];
```
mmcgee-jump:
```suggestion
```

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -646,27 +730,50 @@ unprivileged_init( fd_topo_t *      topo,
...
-    };
-    for( ulong i=0UL; i<2UL; i++ ) fd_ssping_add( ctx->ssping, initial_peers[ i ] );
-  } else if( FD_LIKELY( !strcmp( tile->snaprd.cluster, "private" ) ) ) {
-    fd_ip4_port_t initial_peers[ 1UL ] = {
-      { .addr = FD_IP4_ADDR( 147, 28, 185, 47 ), .port = 8899 } /* A private cluster peer */
-    };
-    for( ulong i=0UL; i<1UL; i++ ) fd_ssping_add( ctx->ssping, initial_peers[ i ] );
-  } else if (FD_LIKELY( !strcmp( tile->snaprd.cluster, "mainnet" ) ) ) {
+  ctx->gossip.ci_table = _ci_table;
+  /* zero-out memory so that we can perform null checks in after_frag */
```
mmcgee-jump, line 952: I don't think you need to do this?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -474,6 +494,37 @@ after_credit( fd_snaprd_tile_t *  ctx,
...
+during_frag( fd_snaprd_tile_t * ctx,
+             ulong              in_idx,
+             ulong              seq FD_PARAM_UNUSED,
+             ulong              sig FD_PARAM_UNUSED,
+             ulong              chunk,
+             ulong              sz,
+             ulong              ctl FD_PARAM_UNUSED) {
+  if( ctx->in_kind[ in_idx ]!= IN_KIND_GOSSIP ) return;
+
+  if( FD_UNLIKELY( chunk<ctx->gossip_in.chunk0 ||
```
mmcgee-jump, line 688: Check sz against mtu as well, see other tile examples

src/discof/restore/fd_snaprd_tile.c
ravyu-jump: Predefined based on ordering in `topology.c`? Isn't that a little fragile? Also the actual link set up in `topology.c` will come with the full gossip PR, so until then we don't have a constant (I guess a placeholder would work)

src/discof/restore/fd_snaprd_tile.c
ravyu-jump: Good point

src/discof/restore/fd_snaprd_tile.c
ravyu-jump: This would affect the `if( FD_LIKELY( !!cur_addr.l ) )` checks in after_frag for a new entry

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: yeah... but the other tiles do it too? idk though up to you.

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -474,6 +494,37 @@ after_credit( fd_snaprd_tile_t *  ctx,
...
+             ulong              sig FD_PARAM_UNUSED,
+             ulong              chunk,
+             ulong              sz,
+             ulong              ctl FD_PARAM_UNUSED) {
+  if( ctx->in_kind[ in_idx ]!=IN_KIND_GOSSIP ) return;
+
+  if( FD_UNLIKELY( chunk<ctx->gossip_in.chunk0 ||
+                   chunk>ctx->gossip_in.wmark  ||
+                   sz>sizeof(fd_gossip_update_message_t) ) ) {
+    FD_LOG_ERR(( "snaprd: unexpected chunk %lu", chunk ));
```
cali-jumptrading, line 691: Could use `FD_TEST` here for simplicity.

src/discof/restore/fd_snaprd_tile.c
mmcgee-jump: You should have issued a REMOVE first, so that line is pointless and should never be needed, right? It actually seems wrong because you might release an IP that wasn't in use from gossip

ravyu-jump: > You should have issued a REMOVE first

`CONTACT_INFO_REMOVE` is only issued if a contact info was dropped from the Gossip table, not if only its RPC socket entry is removed. The removal of a valid contact info's RPC socket is communicated via the normal `CONTACT_INFO` update message. So we'd need null checks to track transitions (no addr > some addr, some addr > no addr, old addr > new addr)

ravyu-jump:
Superseded by: https://github.com/firedancer-io/firedancer/pull/6285

PR #6005, opened 2025-08-11
quic: tx_buf_sz pow2

akhinvasara-jumptrading:
Requires tx_buf_sz for quic streams to be a pow2, so masking works correctly. 
And fixes circular buf OOB bug
found by @two-heart

### DIFF
--- a/src/waltz/quic/fd_quic_stream.c
+++ b/src/waltz/quic/fd_quic_stream.c
@@ -75,7 +75,7 @@ fd_quic_buffer_load( fd_quic_buffer_t * buf,
 
   if( mtail >= mhead ) {
     /* free space split */
-    ulong end_sz = cap - mhead;
+    ulong end_sz = cap - mtail;
     if( data_sz <= end_sz ) {
       /* consists entirely of space at end of buffer */
       fd_memcpy( data, raw + mtail, data_sz );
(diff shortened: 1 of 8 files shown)


src/waltz/quic/fd_quic_stream_pool.c
```diff
@@ -38,6 +39,7 @@ fd_quic_stream_pool_new( void * mem, ulong count, ulong tx_buf_sz ) {
   offs += fd_ulong_align_up( sizeof( fd_quic_stream_pool_t ), FD_QUIC_STREAM_POOL_ALIGN );
 
   ulong stream_foot = fd_quic_stream_footprint( tx_buf_sz );
+  if( FD_UNLIKELY( !stream_foot ) ) { FD_LOG_WARNING(( "stream_foot is 0" )); return NULL; }
```
ripatel-fd: This error message is quite confusing

src/waltz/quic/fd_quic_stream.c
```diff
@@ -97,6 +97,8 @@ fd_quic_stream_align( void );
 
 ulong
 fd_quic_stream_footprint( ulong tx_buf_sz ) {
+  if( FD_UNLIKELY( !fd_ulong_is_pow2( tx_buf_sz ) ) ) return 0UL;
```
ripatel-fd, line 100: Please add a regression test for this case

PR #6015, opened 2025-08-11
bundle: sleep while disconnected

ripatel-fd:
Saves power when disconnected from the bundle server


src/disco/bundle/fd_bundle_tile.seccomppolicy
```diff
@@ -119,3 +119,6 @@ lseek: (and (or (eq (arg 0) etc_resolv_conf)
                 (eq (arg 0) etc_hosts_fd))
             (eq (arg 1) 0)
             (eq (arg 2) "SEEK_SET"))
+
+# bundle: yield to scheduler to save power
```
mmcgee-jump, line 123: You need `sched_yield` as well to call `fd_log_sleep`

mmcgee-jump (changes requested):
.

PR #6017, opened 2025-08-12
send_test: send tile test command and topology

akhinvasara-jumptrading:
This PR creates a new firedancer-dev command to test the send tile. It will allow for rapid iteration on the quic client and vote performance, continuously testing the production send tile without depending on the rest of the full client. 

This PR also adds socket support to the send tile.

### DIFF
--- a/src/app/shared/commands/run/run.c
+++ b/src/app/shared/commands/run/run.c
@@ -898,6 +898,11 @@ run_firedancer( config_t * config,
 void
 run_cmd_fn( args_t *   args FD_PARAM_UNUSED,
             config_t * config ) {
+  #define CHECK_PORT_NON_ZERO( field ) \
+    if( FD_UNLIKELY( config->field==0 ) ) { \
+      FD_LOG_ERR(( #field " is not set in your configuration file. Please set it to a non-zero value." )); \
+    }
+
   if( FD_UNLIKELY( !config->gossip.entrypoints_cnt && !config->development.bootstrap ) )
     FD_LOG_ERR(( "No entrypoints specified in configuration file under [gossip.entrypoints], but "
                  "at least one is needed to determine how to connect to the Solana cluster. If "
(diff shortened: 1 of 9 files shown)


src/app/shared/commands/run/run.c
```diff
@@ -911,6 +916,15 @@ run_cmd_fn( args_t *   args FD_PARAM_UNUSED,
                    "empty. Please remove the empty entrypoint or set it correctly. "));
   }
 
+  CHECK_PORT_NON_ZERO( gossip.port );
+  CHECK_PORT_NON_ZERO( tiles.quic.quic_transaction_listen_port );
+  CHECK_PORT_NON_ZERO( tiles.quic.regular_transaction_listen_port );
+  CHECK_PORT_NON_ZERO( tiles.shred.shred_listen_port );
+  CHECK_PORT_NON_ZERO( tiles.metric.prometheus_listen_port );
+  CHECK_PORT_NON_ZERO( tiles.gui.gui_listen_port );
```
ripatel-fd, line 924: These are already covered by CFG_HAS_NON_ZERO in fd_config.c

src/app/shared/fd_config.c
```diff
@@ -536,16 +532,11 @@ fd_config_validate( fd_config_t const * config ) {
...
   CFG_HAS_NON_ZERO( tiles.shred.max_pending_shred_sets );
-  CFG_HAS_NON_ZERO( tiles.shred.shred_listen_port );
 
   if( config->is_firedancer ) {
     CFG_HAS_POW2( tiles.repair.slot_max );
   }
 
-  CFG_HAS_NON_ZERO( tiles.metric.prometheus_listen_port );
-
-  CFG_HAS_NON_ZERO( tiles.gui.gui_listen_port );
```
ripatel-fd, line 547: Why did you remove these?

akhinvasara-jumptrading: Those ports need to be 0 when you want to run net tile without all of its out links (e.g. for send_test)...they should not be required in the base validation function. So I moved the check to run, where they are actually necessary. Chatted with @mmcgee-jump about this.

src/app/shared/commands/run/run.c
mmcgee-jump: Yah I don't really see why these need to move? You can keep checking them in config.c, even if they are unused in the topology, they will still be set in the default.toml

mmcgee-jump: Ok this is fine

PR #6031, opened 2025-08-12
New BLAKE3 implementation

ripatel-fd:
We are so back


0x0ece:
Let's go! We should write a paper :)

ripatel-fd:
I'm going to move this to (yet) another PR. Will make the following changes:
- The existing batch scheduling algorithm will be removed
- In the short term (follow up PR), a batch API for <=1KiB inputs is provided
- Long-term a new async/streaming scheduler for large >=1KiB inputs will be provided

PR #6040, opened 2025-08-13
solfuzz: revert belt-sanding changes

mjain-jump:
sol_compat harnesses are currently broken. Going to revert these changes until they're fixed to unblock the fuzzers


ripatel-fd:
This would have been a 30 line fix ... Wtf are we doing

PR #6042, opened 2025-08-13
solfuzz belt sand

ripatel-fd:


src/flamenco/runtime/tests/fd_solfuzz.h
```diff
@@ -0,0 +1,192 @@
...
+
+int
+fd_solfuzz_vm_interp_fixture( fd_solfuzz_runner_t * runner,
+                              uchar const *         in,
+                              ulong                 in_sz );
+
+/* Bincode Encoding */
+
+ulong
+fd_runtime_fuzz_type_run( fd_solfuzz_runner_t * runner,
```
mjain-jump: `fd_solfuzz_type_run`?

src/flamenco/runtime/tests/fd_solfuzz_exec.c
```diff
@@ -0,0 +1,396 @@
...
+                               ulong                 in_sz ) {
+  // Decode fixture
+  fd_exec_test_elf_loader_fixture_t fixture[1] = {0};
+  void * res = sol_compat_decode( &fixture, in, in_sz, &fd_exec_test_elf_loader_fixture_t_msg );
+  if( !res ) {
+    FD_LOG_WARNING(( "Invalid elf_loader fixture." ));
+    return 0;
+  }
+
+  fd_spad_push( runner->spad );
```
mjain-jump, line 343: probably should favor `FD_SPAD_FRAME_PUSH/POP` patterns here

mjain-jump (approved):
few comments but lgtm otherwise

src/flamenco/runtime/tests/fd_solfuzz.h
ripatel-fd: thanks

src/flamenco/runtime/tests/fd_solfuzz_exec.c
ripatel-fd: I deliberately reverted these because I think we should prefer plain push/pop. The spad frame macro relies on C extension compiler macros that emulate C++ behavior (implicit destructors). I'd rather leave it explicit what this does. 

IMO, we should use spad frame push/pop to guard any chunk of code that has branches, but all this solfuzz stuff only calls a function

mjain-jump (approved):
lgtm!

PR #6043, opened 2025-08-13
snapshots: remove old snapshots on disk

cali-jumptrading:


src/app/shared/fd_config_parse.c
```diff
@@ -12,6 +12,12 @@ fd_config_check_configf( fd_config_t *  config,
   if( FD_UNLIKELY( config->paths.snapshots[ 0 ]!='\0' && config->paths.snapshots[ 0 ]!='/' ) ) {
     FD_LOG_ERR(( "[config->paths.snapshots] must be an absolute path and hence start with a '/'"));
   }
+  if( FD_UNLIKELY( config_f->snapshots.archive.max_full_snapshots_to_keep>16UL ) ) {
```
mmcgee-jump: Any reason this is checked / enforced? Doesn't seem like a real constraint anywhere in the system

cali-jumptrading: hmm this is arbitrary... I mean I don't see why someone would want to keep more than a few snapshots on disk... what do you think?

mmcgee-jump: Yeah, the default of 1 will prevent that, but I wouldn't add arbitrary limits like this.

cali-jumptrading: ok got it, will remove. Although, there will still be an implicit limit (I will increase it to maybe 64? ) in `fd_ssarchive_remove_old_snapshots` when finding snapshots and sorting by slot, is that alright?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -549,6 +551,11 @@ privileged_init( fd_topo_t *      topo,
   ctx->peer_selection = 1;
   ctx->state          = FD_SNAPRD_STATE_WAITING_FOR_PEERS;
 
+  /* Remove old snapshots on disk first */
```
mmcgee-jump: Personally not a fan of obvious comments like this

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -140,3 +155,76 @@ fd_ssarchive_latest_pair( char const * directory,
...
+                                   uint         max_full_snapshots_to_keep,
+                                   uint         max_incremental_snapshots_to_keep ) {
+  ulong full_snapshots_cnt        = 0UL;
+  ulong incremental_snapshots_cnt = 0UL;
+  fd_ssarchive_entry_t full_snapshots[ FD_SSARCHIVE_MAX_FULL_ENTRIES ];
+  fd_ssarchive_entry_t incremental_snapshots[ FD_SSARCHIVE_MAX_INCREMENTAL_ENTRIES ];
+
+  DIR * dir = opendir( directory );
+  if( FD_UNLIKELY( !dir ) ) {
+    if( FD_LIKELY( errno==ENOENT ) ) FD_LOG_ERR(( "opendir() failed `%s` (%i-%s)", directory, errno, fd_io_strerror( errno ) ));
```
mmcgee-jump:
```suggestion
    if( FD_LIKELY( errno!=ENOENT ) ) FD_LOG_ERR(( "opendir() failed `%s` (%i-%s)", directory, errno, fd_io_strerror( errno ) ));
```

ENOENT is fine here, right?

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -5,6 +5,21 @@
...
+  char  path[ PATH_MAX ];
+};
+typedef struct fd_ssarchive_entry fd_ssarchive_entry_t;
+
+#define SORT_NAME  sort_ssarchive_entries
+#define SORT_KEY_T fd_ssarchive_entry_t
+#define SORT_BEFORE(a,b) ( (a).slot>(b).slot )
+#include "../../../util/tmpl/fd_sort.c"
+
+#define FD_SSARCHIVE_MAX_FULL_ENTRIES        (128UL)
```
mmcgee-jump: I'd use a larger value here like 4096UL just in case

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -140,3 +155,76 @@ fd_ssarchive_latest_pair( char const * directory,
...
+    if( FD_UNLIKELY( -1==fd_ssarchive_parse_filename( entry->d_name, &entry_full_slot, &entry_incremental_slot, decoded_hash ) ) ) {
+      FD_LOG_INFO(( "unrecognized snapshot file `%s/%s` in snapshots directory", directory, entry->d_name ));
+      continue;
+    }
+
+    if( FD_LIKELY( entry_incremental_slot==ULONG_MAX ) ) {
+      FD_TEST( entry_full_slot!=ULONG_MAX );
+
+      if( FD_UNLIKELY( full_snapshots_cnt>=FD_SSARCHIVE_MAX_FULL_ENTRIES ) ) {
+        FD_LOG_ERR(( "too many full snapshots in directory `%s`", directory ));
```
mmcgee-jump: Might be over-aggressive to ERR here, is it valid to just `continue` ?

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -140,3 +155,76 @@ fd_ssarchive_latest_pair( char const * directory,
...
+      }
+
+      incremental_snapshots[ incremental_snapshots_cnt ].slot = entry_incremental_slot;
+      FD_TEST( fd_cstr_printf_check( incremental_snapshots[ incremental_snapshots_cnt ].path, PATH_MAX, NULL, "%s/%s", directory, entry->d_name ) );
+      incremental_snapshots_cnt++;
+    }
+  }
+
+  if( FD_UNLIKELY( -1==closedir( dir ) ) ) FD_LOG_ERR(( "closedir() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+  if( FD_UNLIKELY( errno && errno!=ENOENT ) ) FD_LOG_ERR(( "readdir() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
```
mmcgee-jump, line 209: My bug, but the `errno` check needs to go above `closedir` (which could reset errno to zero..) We want to check `errno` after `readdir` returns NULL, to see if the dir ended or we hit an error. Could you also update the other buggy users? There's one in `fd_file_util.c`, `hugetlbfs.c`,`run.c`, etc...

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -140,3 +155,76 @@ fd_ssarchive_latest_pair( char const * directory,
...
+
+  if( FD_LIKELY( full_snapshots_cnt>max_full_snapshots_to_keep ) ) {
+    sort_ssarchive_entries_inplace( full_snapshots, full_snapshots_cnt );
+    for( ulong i=max_full_snapshots_to_keep; i<full_snapshots_cnt; i++ ) {
+      if( FD_UNLIKELY( -1==remove( full_snapshots[ i ].path ) ) ) {
```
mmcgee-jump: Prefer `unlink` here for nitty reasons. Slightly better if there's TOCTOU problems.

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -140,3 +155,76 @@ fd_ssarchive_latest_pair( char const * directory,
...
+  if( FD_LIKELY( incremental_snapshots_cnt>max_incremental_snapshots_to_keep ) ) {
+    sort_ssarchive_entries_inplace( incremental_snapshots, incremental_snapshots_cnt );
+    for( ulong i=max_incremental_snapshots_to_keep; i<incremental_snapshots_cnt; i++ ) {
+      if( FD_UNLIKELY( -1==remove( incremental_snapshots[ i ].path ) ) ) {
+        FD_LOG_ERR(( "remove() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+      }
+    }
+  }
+
+  return;
```
mmcgee-jump:
```suggestion
```

src/app/firedancer/config/default.toml
```diff
@@ -329,6 +329,18 @@ user = ""
     # are integrated with gossip
     cluster = "testnet"
 
+    # Controls how many snapshots are kept on disk when the validator
+    # boots up.  Setting these values to zero will disable loading
+    # snapshots from disk the next time the validator is restarted.  If
+    # there are more snapshots than the values specified here, the
+    # oldest snapshots will be removed by slot.
+    [snapshots.archive]
```
mmcgee-jump: Don't love the nested `archive` here, probably clearer to just have these straight under `[snapshots]`

src/app/firedancer/config/default.toml
```diff
@@ -329,6 +329,18 @@ user = ""
     # are integrated with gossip
     cluster = "testnet"
 
+    # Controls how many snapshots are kept on disk when the validator
```
```suggestion
    # Controls how many full snapshots are allowed to be kept in the
    # snapshots directory on disk.
    #
    # Firedancer does not currently support creating snapshots, and so
    # these values are only checked once on boot.  Any excess snapshots
    # will be deleted, oldest snapshot slot first.
```

mmcgee-jump (changes requested):
.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: removed

src/discof/restore/utils/fd_ssarchive.c
cali-jumptrading: done

src/discof/restore/utils/fd_ssarchive.c
cali-jumptrading: fixed

src/app/firedancer/config/default.toml
cali-jumptrading: done

src/app/firedancer/config/default.toml
cali-jumptrading: done

src/discof/restore/utils/fd_ssarchive.c
cali-jumptrading: yeah I agree, done

src/discof/restore/utils/fd_ssarchive.c
cali-jumptrading: done

src/app/platform/fd_file_util.c
```diff
@@ -160,8 +160,8 @@ fd_file_util_rmtree( char const * path,
     }
   }
 
-  if( FD_UNLIKELY( -1==closedir( dir ) ) )            return -1;
   if( FD_UNLIKELY( errno && errno!=ENOENT ) )         return -1;
+  if( FD_UNLIKELY( -1==closedir( dir ) ) )      return -1;
```
mmcgee-jump, line 164: alignment

src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -136,7 +151,79 @@ fd_ssarchive_latest_pair( char const * directory,
...
+        FD_LOG_ERR(( "remove() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+      }
+    }
+  }
+
+  if( FD_LIKELY( incremental_snapshots_cnt>max_incremental_snapshots_to_keep ) ) {
+    sort_ssarchive_entries_inplace( incremental_snapshots, incremental_snapshots_cnt );
+    for( ulong i=max_incremental_snapshots_to_keep; i<incremental_snapshots_cnt; i++ ) {
+      if( FD_UNLIKELY( -1==unlink( incremental_snapshots[ i ].path ) ) ) {
+        FD_LOG_ERR(( "remove() failed (%i-%s)", errno, fd_io_strerror( errno ) ));
```
mmcgee-jump: unlink() failed

mmcgee-jump: Probably print the path here too, so operator can debug?

mmcgee-jump (dismissed):
couple of nits

mmcgee-jump (approved):

PR #6053, opened 2025-08-14
snapshots: http peer config

cali-jumptrading:


src/app/firedancer/config/default.toml
```diff
@@ -282,7 +282,7 @@ user = ""
     # present.
     #
     # TODO: implement the "at most one downloaded " logic. "
-    download = true
+    download = false
```
mmcgee-jump: Revert

src/app/firedancer/topology.c
```diff
@@ -132,10 +132,14 @@ resolve_address( char const * address,
...
 
 static int
-resolve_gossip_entrypoint( char const *    host_port,
-                           fd_ip4_port_t * ip4_port ) {
+resolve_peer( char const *    peer,
+              fd_ip4_port_t * ip4_port ) {
 
   /* Split host:port */
+  char const * host_port = peer;
+  if( FD_LIKELY( strncmp( peer, "http://", 7UL )==0 ) ) {
```
mmcgee-jump, line 134: I'd cover https:// also

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -646,27 +646,8 @@ unprivileged_init( fd_topo_t *      topo,
   ctx->sshttp = fd_sshttp_join( fd_sshttp_new( _sshttp ) );
   FD_TEST( ctx->sshttp );
 
-  if( FD_LIKELY( !strcmp( tile->snaprd.cluster, "testnet" ) ) ) {
```
mmcgee-jump: Are we ready to remove this, won't it break til gossip is wired?

mmcgee-jump (changes requested):
.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: Yes we can safely remove this. When running the full client, the correct peer must be specified in the config, rather than relying on these hardcoded peers.

mmcgee-jump: But the "default" config should work out of the box, will that still be true?

cali-jumptrading: oh we'll just have to set the default peer in the config to a testnet peer then right?

cali-jumptrading: oh I think I see what you mean here... setting the default peer in the config to some peer we use for development is not a default option for users. When users start the client, they probably want to download a snapshot from the gossip entrypoints (most trusted sources) by default.

cali-jumptrading: In that case, I can add the cluster option back... but it's not really a default option for users either since it just hardcodes the peers into the `snaprd` tile. what do you think? Should we just go ahead with this change and wait for gossip or bring back the cluster option (which is just a hack anyways).

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -133,7 +133,7 @@ fd_sshttp_init( fd_sshttp_t * http,
 
   struct sockaddr_in addr_in = {
     .sin_family = AF_INET,
-    .sin_port   = fd_ushort_bswap( addr.port ),
+    .sin_port   = addr.port,
```
ripatel-fd: This looks wrong. Port numbers are always little endian in Firedancer.

src/discof/restore/utils/fd_ssping.c
```diff
@@ -395,7 +395,7 @@ peer_connect( fd_ssping_t *      ssping,
 
   struct sockaddr_in addr = {
     .sin_family = AF_INET,
-    .sin_port   = fd_ushort_bswap( peer->addr.port ),
+    .sin_port   = peer->addr.port,
```
ripatel-fd: same here

src/app/firedancer/config/default.toml
```diff
@@ -324,10 +324,43 @@ user = ""
...
+        # available:
+        #
+        #  /snapshot.tar.bz2
+        #  /incremental-snapshot.tar.bz2
+        #
+        # If these peers have snapshots at the given paths, they will be
+        # considered as additional sources for downloading snapshots
+        # when enabled alongside gossip and entrypoint sources.
+        [[snapshots.sources.http]]
+        # Whether to enable fetching snapshots from a list of
```
```suggestion
            # Whether to enable fetching snapshots from a list of
```
nit: whitespace

src/app/firedancer/config/default.toml
```diff
@@ -324,10 +324,43 @@ user = ""
...
+        # HTTP peers.
+        enabled = false
+
+        # Example: url = "http://example.com:1234"
```
mmcgee-jump: Just use an example URL in the field! Since it's "disable". Then use the comment to describe the field, something like

```suggestion
        # The URL to append /snapshot.tar.bz or /incremental-snapshot.tar.bz to
        # when locating the current snapshot.
        # Example: url = "http://example.com:1234"
```

src/app/shared/fd_config.h
```diff
@@ -130,14 +131,27 @@ struct fd_configf {
   } runtime;
 
   struct {
+
+    struct {
+
+      struct {
+        ulong            peers_cnt;
```
mmcgee-jump, line 137:
```suggestion
        ulong peers_cnt;
```

mmcgee-jump (changes requested):
.

src/discof/restore/utils/fd_sshttp.c
cali-jumptrading: Let's punt this to when your fix for port endianess goes in: https://github.com/firedancer-io/firedancer/pull/6062

kbhargava-jump:
please make sure snapshots aren't deleted before re-enabling CI

PR #6057, opened 2025-08-14
chacha: fix OOB issue in test code path

ripatel-fd:
Reported-by: two-heart <12869538+two-heart@users.noreply.github.com>


ripatel-fd:
nvm, the existing code is correct

PR #6058, opened 2025-08-14
fix(repair): tile fec_sig removal and force_complete logic

lidatong:


src/discof/forest/fd_forest.c
```diff
@@ -436,10 +437,11 @@ fd_forest_data_shred_insert( fd_forest_t * forest, ulong slot, ushort parent_off
     fd_forest_ancestry_ele_insert( fd_forest_ancestry( forest ), ele, pool );
     link( forest, parent, ele );
   }
-  fd_forest_ele_idxs_insert( ele->fecs, fec_set_idx );
+  if( fec_set_idx > 0) fd_forest_ele_idxs_insert( ele->fecs, fec_set_idx - 1 );
```
ripatel-fd: Out of curiosity, what's special about FEC set 0?

lidatong: it's the first FEC set in a slot

src/discof/repair/fd_repair_tile.c
```diff
@@ -887,6 +889,8 @@ after_frag( fd_repair_tile_ctx_t * ctx,
         cmr = &fd_reasm_root( ctx->reasm )->key;
       }
       FD_TEST( fd_reasm_insert( ctx->reasm, merkle_root, cmr, shred->slot, shred->fec_set_idx, shred->data.parent_off, (ushort)(shred->idx - shred->fec_set_idx + 1), data_complete, slot_complete ) );
+      fd_fec_sig_t * fec_sig = fd_fec_sig_query( ctx->fec_sigs, (shred->slot << 32) | shred->fec_set_idx, NULL );
+      FD_TEST( fec_sig );
```
emwang-jump, line 894: if we force complete it, & remove from the fec_sig map at the point of stem publishing to shred, the fec_complete msg comes back which causes the fec_sig to get readded to the map and then this query passes. but then that means we don't actually evict anything from the fec_sig map?

emwang-jump (approved):

PR #6059, opened 2025-08-14
tower, replay: remove funk usage from tower

topointon-jump:
Removes all Funk usage from Tower. This is a naive simple solution, in which Replay sends Tower all the vote accounts listed in the vote accounts cache at the end of each slot.

Potential next steps, to be done in follow-up PRs:
- [ ] Only send vote accounts who have voted in this slot. Not clear if this is worth the complexity.
- [ ] Update the set of vote accounts in the `epoch_voters` map based on the set of vote accounts received from Replay. Currently only the stake amount is updated.


src/disco/fd_disco_base.h
```diff
@@ -67,11 +67,35 @@ FD_STATIC_ASSERT( FD_SHRED_REPAIR_MTU == 152UL , update FD_SHRED_REPAIR_MTU );
...
+   needs at the end of a slot. */
+struct __attribute__((packed)) fd_replay_out {
+  fd_hash_t                  block_id;           /* block id (last FEC set's merkle root) of the slot received from replay */
+  fd_hash_t                  parent_block_id;    /* parent block id of the slot received from replay */
+  fd_hash_t                  bank_hash;          /* bank hash of the slot received from replay */
+  fd_hash_t                  block_hash;         /* last microblock header hash of slot received from replay */
+  ulong                      has_our_vote_state; /* whether our vote state is included in this message */
+  fd_replay_out_vote_state_t our_vote_state;     /* our vote state */
+  ulong                      vote_states_cnt;    /* number of elements in the vote_states field */
+  fd_replay_out_vote_state_t vote_states[];      /* the state of all the vote accounts active in the epoch, as of the end of the slot */
```
lidatong: why special case our own?

src/app/firedancer/topology.c
```diff
@@ -337,7 +337,7 @@ fd_topo_initialize( config_t * config ) {
 
   /**/                 fd_topob_link( topo, "gossip_verif", "gossip_verif", config->tiles.verify.receive_buffer_size, FD_TPU_RAW_MTU,                1UL );
   /**/                 fd_topob_link( topo, "gossip_tower", "gossip_tower", 128UL,                                    FD_TPU_MTU,                    1UL );
-  /**/                 fd_topob_link( topo, "replay_out",   "replay_out",   128UL,                                    sizeof(fd_replay_out_t),       1UL );
+  /**/                 fd_topob_link( topo, "replay_out",   "replay_out",   64UL,                                     sizeof(fd_replay_out_t) + (sizeof(fd_replay_out_vote_state_t) * FD_RUNTIME_MAX_VOTE_ACCOUNTS),       1UL );
```
lidatong: why change to 64? 128 was a placeholder because that used to be min-depth, but really should probably be sized sensibly to be a derived bound in tandem with a starting bound like max unrooted banks

also `replay_out` was meant to be a general link multiple tiles could use. given this is very specific input to tower, it should be its own `replay_tower` link

src/choreo/tower/fd_tower.c
```diff
@@ -201,12 +201,12 @@ fd_tower_switch_check( fd_tower_t const * tower,
...
-fd_tower_threshold_check( fd_tower_t const *    tower,
-                          fd_epoch_t const *    epoch,
-                          fd_funk_t *           funk,
-                          fd_funk_txn_t const * txn,
-                          ulong                 slot,
-                          fd_tower_t *          scratch ) {
+fd_tower_threshold_check( fd_tower_t const *                 tower,
+                          fd_epoch_t *                       epoch,
+                          fd_replay_out_vote_state_t const * vote_states,
+                          ulong                              vote_states_len,
```
lidatong: this should pass in an array of `fd_tower_t` instead  so `fd_tower.h` doesn't need to know about the calling layer (replay_out and tiles) -- i didn't particularly like the old way of passing in `funk` and `txn` but it did it for performance / for the "fork" model

src/choreo/tower/fd_tower.c
```diff
@@ -231,20 +231,22 @@ fd_tower_threshold_check( fd_tower_t const *    tower,
 
   /* Iterate all the vote accounts. */
 
-  fd_voter_t const * epoch_voters = fd_epoch_voters_const( epoch );
-  for (ulong i = 0; i < fd_epoch_voters_slot_cnt( epoch_voters ); i++ ) {
-    if( FD_LIKELY( fd_epoch_voters_key_inval( epoch_voters[i].key ) ) ) continue /* most slots are empty */;
+  for (ulong i = 0; i < vote_states_len; i++ ) {
+    fd_replay_out_vote_state_t const * vote_state = &vote_states[i];
```
lidatong: see above comment

src/choreo/tower/fd_tower.c
```diff
@@ -231,20 +231,22 @@ fd_tower_threshold_check( fd_tower_t const *    tower,
...
 
-    fd_voter_t const * voter = &epoch_voters[i];
+    fd_voter_t * epoch_voters = fd_epoch_voters( epoch );
+    fd_voter_t * voter        = fd_epoch_voters_query( epoch_voters, vote_state->key, NULL );
+    if( FD_UNLIKELY( !voter ) ) {
+      /* TODO: handle this case by adding a vote account to the epoch voters */
+      FD_LOG_DEBUG(( "[%s] voter %s was not in epoch voters", __func__,
```
lidatong: what does the TODO mean? also should at least warn here, if not crit? i would expect replay to always pass in data that is sync with tower in the new copy-based model

src/choreo/tower/fd_tower.c
```diff
@@ -320,12 +322,12 @@ fd_tower_reset_slot( fd_tower_t const * tower,
...
-fd_tower_vote_slot( fd_tower_t *          tower,
-                    fd_epoch_t const *    epoch,
-                    fd_funk_t *           funk,
-                    fd_funk_txn_t const * txn,
-                    fd_ghost_t const *    ghost,
-                    fd_tower_t *          scratch ) {
+fd_tower_vote_slot( fd_tower_t *                       tower,
+                    fd_epoch_t *                       epoch,
+                    fd_replay_out_vote_state_t const * vote_states,
+                    ulong                              vote_states_len,
```
lidatong: see comment in threshold check

src/choreo/tower/fd_tower.c
```diff
@@ -436,54 +438,20 @@ fd_tower_simulate_vote( fd_tower_t const * tower, ulong slot ) {
...
-
-    if( FD_LIKELY( fd_funk_rec_query_test( &query ) == FD_FUNK_SUCCESS ) ) return 0;
-    else fd_tower_votes_remove_all( tower ); /* reset the tower and try again  */
+  /* Push all the votes from the vote state onto the tower. */
+  for( ulong i = 0; i < state->votes_cnt; i++ ) {
+    fd_tower_vote_t vote = {
+      .slot = state->votes[i].slot,
+      .conf = state->votes[i].conf
+    };
+    fd_tower_votes_push_tail( tower, vote );
```
lidatong: same point as above 2 on better API layers -- do this at the tower tile level

src/disco/fd_disco_base.h
lidatong: not a fan of flexible array members pattern -- please make sure your usage (ie. the memcpy in the replay tile) isn't introducing any UB and following all the caveats

src/discof/replay/fd_replay_tile.c
```diff
@@ -1470,6 +1472,82 @@ handle_new_slice( fd_replay_tile_ctx_t * ctx, fd_stem_context_t * stem ) {
...
+        FD_LOG_ERR(( "[%s] unknown vote state version. discriminant %u", __func__, state->discriminant ));
+        return -1;
+      }
+    }
+
+    if( FD_LIKELY( fd_funk_rec_query_test( &query ) == FD_FUNK_SUCCESS ) ) {
+      return 0;
+    }
+  }
+}
```
lidatong: is this speculative code still needed? it was this way originally for zero-copy, but really you're now memcpying everything that i think it's much better and less bug-prone to just have a very short speculate -> copy the account bytes -> parse after

src/discof/replay/fd_replay_tile.c
```diff
@@ -1513,16 +1597,62 @@ exec_slice_fini_slot( fd_replay_tile_ctx_t * ctx, fd_stem_context_t * stem ) {
...
+      fd_pubkey_t const * vote_account_pubkey = &elem->elem.key;
+      if( FD_UNLIKELY( fd_replay_out_vote_state_from_funk(
+        ctx->funk, ctx->slot_ctx->funk_txn, vote_account_pubkey, elem->elem.stake, &out->vote_states[out->vote_states_cnt++] ) ) ) {
+        FD_LOG_ERR(( "failed to get vote state for vote account %s",
+          FD_BASE58_ENC_32_ALLOCA( vote_account_pubkey->uc ) ));
+      }
+    }
+    fd_bank_next_epoch_stakes_end_locking_query( ctx->slot_ctx->bank );
+
+    ulong msg_sz = sizeof(fd_replay_out_t) + ( out->vote_states_cnt * sizeof(fd_replay_out_vote_state_t) );
```
lidatong: define a function for this

src/discof/tower/fd_tower_tile.c
```diff
@@ -62,50 +60,40 @@ typedef struct {
...
+    fd_voter_t * voter = fd_epoch_voters_query( epoch_voters, vote_state->key, NULL );
+    if( FD_UNLIKELY( !voter ) ) {
+      /* TODO: handle this case, by adding a new voter to the epoch voters */
+      FD_LOG_DEBUG(( "[%s] voter %s was not in epoch voters", __func__,
+      FD_BASE58_ENC_32_ALLOCA(&vote_state->key) ));
+      continue;
     }
 
-    // FD_LOG_NOTICE(( "querying %s %lu %lu %lu", FD_BASE58_ENC_32_ALLOCA( voter->key.uc ), voter->stake, vote, root ));
+    /* Update the voter's stake */
```
lidatong: nit but inline comments like these to the same line lower-case

src/discof/tower/fd_tower_tile.c
```diff
@@ -62,50 +60,40 @@ typedef struct {
...
+    voter->stake = vote_state->stake;
+
+    /* Skip voters with no votes */
+    if( FD_UNLIKELY( vote_state->votes_cnt == 0 ) ) continue;
```
lidatong:
```suggestion
    if( FD_UNLIKELY( vote_state->votes_cnt == 0 ) ) continue; /* skip voters with no votes */
```

src/discof/replay/fd_replay_tile.c
```diff
@@ -1513,16 +1597,62 @@ exec_slice_fini_slot( fd_replay_tile_ctx_t * ctx, fd_stem_context_t * stem ) {
...
+    fd_stem_publish(
+      stem,
+      ctx->replay_out_idx,
+      curr_slot,
+      ctx->replay_out_chunk,
+      msg_sz,
+      0UL,
+      fd_frag_meta_ts_comp( fd_tickcount() ),
+      fd_frag_meta_ts_comp( fd_tickcount() ) );
+    ctx->replay_out_chunk = fd_dcache_compact_next( ctx->replay_out_chunk, msg_sz, ctx->replay_out_chunk0, ctx->replay_out_wmark );
```
lidatong: you're only publishing a single frag?

topointon-jump: changing this

src/app/firedancer/topology.c
topointon-jump: changing this

src/disco/fd_disco_base.h
topointon-jump: changing this

src/choreo/tower/fd_tower.c
```diff
@@ -200,13 +200,18 @@ fd_tower_switch_check( fd_tower_t const * tower,
   return switch_pct > SWITCH_PCT;
 }
 
+void
+fd_tower_remove_all_votes( fd_tower_t * tower ) {
+  fd_tower_votes_remove_all( tower );
+}
```
lidatong: unnecessary API

src/choreo/tower/fd_tower.c
```diff
@@ -200,13 +200,18 @@ fd_tower_switch_check( fd_tower_t const * tower,
...
-                          fd_epoch_t const *    epoch,
-                          fd_funk_t *           funk,
-                          fd_funk_txn_t const * txn,
-                          ulong                 slot,
-                          fd_tower_t *          scratch ) {
+fd_tower_threshold_check( fd_tower_t const *   tower,
+                          fd_epoch_t *         epoch,
+                          fd_pubkey_t *        vote_keys,
+                          fd_tower_t * const * vote_towers,
+                          ulong                vote_len,
```
lidatong: use `cnt` for count of elements

src/choreo/tower/fd_tower.c
```diff
@@ -272,6 +265,19 @@ fd_tower_threshold_check( fd_tower_t const *    tower,
...
+        /* This means that the cached list of epoch voters is not in sync with the list passed
+           through from replay. This likely means that we have crossed an epoch boundary and the
+           epoch_voter list has not been updated.
+
+           TODO: update the set of account in epoch_voter's to match the list received from replay,
+                 so that epoch_voters is correct across epoch boundaries. */
+        FD_LOG_WARNING(( "[%s] voter %s was not in epoch voters", __func__,
+          FD_BASE58_ENC_32_ALLOCA(&vote_keys[i]) ));
+        continue;
+      }
```
lidatong: just log crit here while it's a bug

src/choreo/tower/fd_tower.c
```diff
@@ -436,57 +442,6 @@ fd_tower_simulate_vote( fd_tower_t const * tower, ulong slot ) {
   return simulate_vote( tower, slot );
 }
 
-int
```
lidatong, line 439: why move this out into replay tile ?

src/disco/fd_disco_base.h
```diff
@@ -67,13 +67,38 @@ FD_STATIC_ASSERT( FD_SHRED_REPAIR_MTU == 152UL , update FD_SHRED_REPAIR_MTU );
 #define FD_NETMUX_SIG_MIN_HDR_SZ    ( 42UL) /* The default header size, which means no vlan tags and no IP options. */
 #define FD_NETMUX_SIG_IGNORE_HDR_SZ (102UL) /* Outside the allowable range, but still fits in 4 bits when compressed */
 
-struct fd_replay_out {
+/* Maximum number of vote account states the Tower tile can process in one slot */
+#define FD_TOWER_MAX_VOTE_ACCOUNTS (4096UL)
```
lidatong:
```suggestion
#define FD_REPLAY_TOWER_VOTE_ACC_MAX (4096UL)
```

src/discof/replay/fd_replay_tile.c
```diff
@@ -1329,6 +1332,132 @@ handle_new_slice( fd_replay_tile_ctx_t * ctx, fd_stem_context_t * stem ) {
...
+   - pubkey:         The public key of the vote account to retrieve
+   - stake:          The stake amount associated with this vote account
+   - vote_state_out: Output structure to populate with vote state information
+
+   Failure modes:
+   - Vote account not found in funk database (returns -1)
+   - Vote state cannot be parsed from account data (returns -1)
+   - Vote state has invalid version discriminant (calls FD_LOG_ERR and returns -1)
+   - Vote count exceeds maximum of 31 (calls FD_LOG_ERR, fatal) */
+static int
```
lidatong, line 1402: can you add unit tests for these deserializers? i would just pull mainnet / testnet vote account versions. see test_voter.c

src/discof/tower/fd_tower_tile.c
```diff
@@ -62,53 +59,68 @@ typedef struct {
...
+
+  /* Push all the votes from the vote state onto the tower. */
+  for( ulong i = 0; i < state->votes_cnt; i++ ) {
+    fd_tower_vote_t vote = {
+      .slot = state->votes[i].slot,
+      .conf = state->votes[i].conf
+    };
+    fd_tower_votes_push_tail( tower, vote );
+  }
+}
```
lidatong: see comments above ie. why not just use `fd_tower_from_vote_acc` to directly deserialize the vote acc bytes?

src/discof/replay/fd_replay_tile.c
```diff
@@ -1329,6 +1332,132 @@ handle_new_slice( fd_replay_tile_ctx_t * ctx, fd_stem_context_t * stem ) {
   fd_bank_shred_cnt_set( ctx->slot_ctx->bank, fd_bank_shred_cnt_get( ctx->slot_ctx->bank ) + data_cnt );
 }
 
+/* fd_replay_out_vote_state_from_funk extracts vote state information from the funk
```
lidatong: i'm thinking about it and i'm not really sure why you need to create this intermediate struct -- can't you just pass the raw account bytes as-is (relevant offset) through the link then deserialize directly into a tower as part of the frag-reading?

PR #6067, opened 2025-08-15
Add MSan poison/unpoison calls to fd_alloc

Ravi-Kappiyoor:
Built and tested using `./deps.sh +msan fetch check install` followed by `MACHINE=linux_clang_x86_64 EXTRAS=msan make -j` and `MACHINE=linux_clang_x86_64 EXTRAS=msan make run-unit-test`. There are no new test failures when run in MSan mode after this change (i.e., if the test failed before this change, it still does, but if it passed before this change, it still passes).


jumpsiegel (dismissed):

src/util/alloc/fd_alloc.c
```diff
@@ -291,6 +291,9 @@ fd_alloc_private_inactive_stack_pop( fd_alloc_vgaddr_t * inactive_stack,
 
     /**/  top_gaddr = (ulong)fd_alloc_vgaddr_off( old );
     ulong top_ver   = (ulong)fd_alloc_vgaddr_ver( old );
+#   ifdef FD_HAS_MSAN
+    fd_msan_unpoison( &top_gaddr, sizeof( top_gaddr ) );
+#   endif
```
ripatel-fd: These ifdef guards are unnecessary .`fd_msan_unpoison` will resolve to a nop

Ravi-Kappiyoor: Removed those - was copying the `#if FD_HAS_DEEPASAN` logic in this file, but that appears to be a distinction being made between turning on ASan and turning on "deep" ASan which adds more checks.

PR #6069, opened 2025-08-15
stem: simplify busy counting

ripatel-fd:
Always count before/after_credit busy time towards PROCESSING_PREFRAG


ripatel-fd:
@mmcgee-jump didn't like this design, closing

PR #6071, opened 2025-08-15
tile unit test: generic testing framework

hewang-jump:
- Add a generic testing framework for tile-level unit test. 
- Add unit test for the pack tile. 
- Rewrite unit test for the XDP tile. 
- Add partial unit test for the shred tile

### DIFF
--- a/src/app/shared/fd_tile_unit_test.h
+++ b/src/app/shared/fd_tile_unit_test.h
@@ -1,0 +1,486 @@
+#ifndef HEADER_fd_src_app_shared_fd_tile_unit_test_h
+#define HEADER_fd_src_app_shared_fd_tile_unit_test_h
+
+#include "../../app/shared/fd_config.h"
+
+/* We want a way to test the transitions between the tile's callbacks
+   (before/after_credit,before/during/after_frag, housekeeping).  To
+   that end, we need to instantiate the tile's context (e.g.
+   fd_net_ctx_t, fd_pack_ctx_t, etc), which requires calling
+   unprivileged_init, or even privileged_init.  We therefore want to
+   bring up the part of the topology that's relevant to the
+   tile under testing, such as the upstream producer links, downstream
+   consumer links, etc.  This approach loads the config TOML and
+   invokes fd_topo_initialize from either Frankendancer or Firedancer
+   topology, so that the tile is configured the same way as it would
+   be in production running on mainnet or testnet.
+   We don't want to treat the tile as a black box, and therefore
+   cannot use fd_stem.c, so we include the following template to bring
+   some structures to the test.  This is a basic structure for tile
+   unit testing and is not meant to be an exhaustive API, but is
+   expected to evolve over time: specific tiles may require manual
+   configuration at initialization and during testing.  */
+
+/* fd_tile_unit_test_init provides the skeleton initialization steps.
+   From the three config paths, only default_topo_config_path is
+   required, whereas the other two (override_topo_config_path and
+   user_topo_config_path) are optional.  These inputs, together with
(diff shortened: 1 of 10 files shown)


src/app/shared/fd_tile_unit_test.h
```diff
@@ -0,0 +1,490 @@
...
+   ulong cons_seq;  // seq of currently processing frag. initialized to ULONG_MAX, and incremented everytime before the test loop processes a frag
+
+   int is_input_link;   // 1 the link is an input/producer/upstream link, 0 if the link is an output/consumer/downstream link
+
+   // for input/producer/upstream link
+   ulong (*publish) ( test_ctx_t *, test_link_t * );    // callback to publish a frag
+   ulong (*make_sig)( test_ctx_t *, test_link_t * );    // callback to make signature for the published frag
+   int (*bf_check)( test_ctx_t *, TEST_TILE_CTX_TYPE * );  // callback to check before_frag
+   int (*df_check)( test_ctx_t *, TEST_TILE_CTX_TYPE * );  // callback to check during_frag
+   int (*af_check)( test_ctx_t *, TEST_TILE_CTX_TYPE * );  // callback to check after_frag
```
ripatel-fd: I'd prefer just writing out the full function name so it's easier to read

src/app/shared/fd_tile_unit_test.h
```diff
@@ -0,0 +1,490 @@
...
+enum test_callback_fn_num {
+   CALLBACK_FN_BC     = 0,    // before_credit
+   CALLBACK_FN_AC     = 1,    // after_credit
+   CALLBACK_FN_BF     = 2,    // before_frag
+   CALLBACK_FN_DF     = 3,    // during_frag
+   CALLBACK_FN_AF     = 4,    // after_frag
+   CALLBACK_FN_PUB    = 5,    // publish
+   CALLBACK_FN_SIG    = 6,    // make_sign
+   CALLBACK_FN_OUTVER = 7     // output verifier
+};
```
ripatel-fd: Usually we use `#define` for enums. Using `enum` is not wrong, but it's just a Firedancer code style quirk.
Could you also rename these to be `FD_STEM_CALLBACK_{...}`?  Only stem based tiles have these callbacks

src/app/shared/fd_tile_unit_test.h
```diff
@@ -0,0 +1,490 @@
...
+
+
+/* Initialize an output/consumer/downstream test_link:
+   Find the link in the topology according link_name, and initialize
+   the test link with callbacks for frag verification.  Usually, the
+   link was added to topology by fd_topob_tile_out.  The check
+   functions should return a non-zero error code when fail and log
+   warning. */
+void
+init_test_link_out( fd_topo_t     * topo,
```
ripatel-fd: These are all externally linked functions, so they should be namespaced.

E.g. `fd_stem_test_{...}`.

src/app/shared/fd_tile_unit_test.h
hewang-jump: done.

src/app/shared/fd_tile_unit_test.h
hewang-jump: Thanks. Instead of calling it `FD_STEM_CALLBACK_{...} `, we do `FD_TILE_TEST_CALLBACK_{...}`. done.

src/app/shared/fd_tile_unit_test.h
hewang-jump: Thanks. Instead of calling it `fd_stem_test_{...}`, we do `fd_tile_test_{...}`. done

PR #6078, opened 2025-08-16
quic: memclr conn in conn_create

akhinvasara-jumptrading:
Currently, conn_create has to reset all fields of the conn struct individually. This is error prone, e.g. we were not resetting retry_src_conn_id. Given that most fields should be 0, this PR clears out the entire struct. It just saves a few individual fields that should be persistent across uses of the conn object.

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -480,12 +480,6 @@ fd_quic_init( fd_quic_t * quic ) {
     /* used for indexing */
     conn->conn_idx = (uint)j;
 
-    conn->svc_type = UINT_MAX;
-    conn->svc_next = conn->svc_prev = UINT_MAX;
-    /* start with minimum supported max datagram */
-    /* peers may allow more */
-    conn->tx_max_datagram_sz = FD_QUIC_INITIAL_PAYLOAD_SZ_MAX;
-
     /* add to free list */
     *fd_ptr_if( last!=NULL, &last->svc_next, &state->free_conn_list ) = (uint)j;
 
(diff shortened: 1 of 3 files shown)


src/waltz/quic/fd_quic.c
```diff
@@ -480,12 +480,6 @@ fd_quic_init( fd_quic_t * quic ) {
     /* used for indexing */
     conn->conn_idx = (uint)j;
 
-    conn->svc_type = UINT_MAX;
-    conn->svc_next = conn->svc_prev = UINT_MAX;
-    /* start with minimum supported max datagram */
-    /* peers may allow more */
-    conn->tx_max_datagram_sz = FD_QUIC_INITIAL_PAYLOAD_SZ_MAX;
```
ripatel-fd, line 487: Your patch no longer initializes tx_max_datagram_sz. Is setting it redundant?

src/waltz/quic/fd_quic.c
```diff
@@ -1171,6 +1165,7 @@ fd_quic_conn_new_stream( fd_quic_conn_t * conn ) {
   if( FD_UNLIKELY( !entry ) ) {
     /* return stream to pool */
     fd_quic_stream_pool_free( state->stream_pool, stream );
+    FD_LOG_INFO(( "stream map insert failed" ));
```
ripatel-fd, line 1192: INFO is rather arbitrary ... If it should never happen at runtime, this should be a CRIT. If it is expected/tolerable, this should increment a metric. If it's relevant to operators, it should be a WARNING.
I'm fine with leaving this for now, but ideally we'd just ensure that the stream map always has space and hard crash here if insertion fails.

src/waltz/quic/fd_quic.c
akhinvasara-jumptrading: yep, it's alr set in fd_quic_conn_create

src/waltz/quic/fd_quic.c
akhinvasara-jumptrading: yes, a bit arbitrary - I think at some point I'll go through and add a lot more client-side metrics. Let's leave it as is for now.

PR #6084, opened 2025-08-18
Add core-to-core RTT test

ripatel-fd:
- Add 'tile-rtt' action and 'trtt', 'echo' tiles
- Adapt fd_tile_private_cpus_parse to allow specifying a custom max tile count
- Add some missing help messages to firedancer-dev


ripatel-fd:
Let's reopen this in case it becomes useful in the future

PR #6086, opened 2025-08-18
quic: timeout state fixes

akhinvasara-jumptrading:
This PR fixes the revival logic, no longer assuming that a revived conn should be ACTIVE. It also moves the fd_quic_set_conn_state method into fd_quic.c, because the method exists to manipulate quic metrics.

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -834,6 +834,30 @@ fd_quic_log_full_hdr( fd_quic_conn_t const * conn,
   return hdr;
 }
 
+static inline void
+fd_quic_set_conn_state( fd_quic_conn_t * conn,
+                        uint             new_state ) {
+
+  uint old_state = conn->state;
+
+  FD_COMPILER_MFENCE();
+
+  int was_timed_out = !!(old_state == FD_QUIC_CONN_STATE_TIMED_OUT);
+  int freeing       = was_timed_out & !!(new_state == FD_QUIC_CONN_STATE_INVALID);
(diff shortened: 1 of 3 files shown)


src/waltz/quic/fd_quic.c
```diff
@@ -834,6 +834,26 @@ fd_quic_log_full_hdr( fd_quic_conn_t const * conn,
...
+  int reviving      = was_timed_out & !freeing & !(new_state == FD_QUIC_CONN_STATE_TIMED_OUT);
+
+  fd_quic_metrics_t * metrics = &conn->quic->metrics;
+  metrics->conn_timeout_freed_cnt   += (ulong)freeing;
+  metrics->conn_timeout_revived_cnt += (ulong)reviving;
+
+  metrics->conn_state_cnt[ old_state ]--;
+  metrics->conn_state_cnt[ new_state ]++;
+
+  fd_quic_set_conn_state( conn, new_state );
```
ripatel-fd: Is there any code path that does not go through metrics? Can we unify this wrapper and the actual function?

src/waltz/quic/fd_quic.c
```diff
@@ -5685,7 +5703,7 @@ fd_quic_handle_handshake_done_frame(
   fd_quic_tls_clear_hs_data( conn->tls_hs, fd_quic_enc_level_appdata_id );
 
   /* we shouldn't be receiving this unless handshake is complete */
-  fd_quic_set_conn_state( conn, FD_QUIC_CONN_STATE_ACTIVE );
+  fd_quic_set_conn_state_metrics( conn, FD_QUIC_CONN_STATE_ACTIVE );
```
ripatel-fd: Changing all these call sites isn't ideal IMO

PR #6095, opened 2025-08-19
runtime: remove FD_PUBKEY_SUCCESS

two-heart:
previously, FD_PUBKEY_SUCCESS was mixed with FD_EXECUTOR_INSTR_ERR_* for failiure cases. Furthermore, it runs the risk of confusion with #define FD_PUBKEY_ERR_* codes, that have a semantically different meaning (custom_err vs return code).


ripatel-fd:
fd_pubkey and the instruction executors are different APIs. I'd prefer not mixing error codes here. We can look into using enum if we have too many conversion errors.

PR #6096, opened 2025-08-19
log: fix signal lock reentrancy bug

ripatel-fd:
Fixes a bug in fd_log where a thread starts infinite looping if it
catches a signal while already holding a log lock.  This affects
all regular executables (e.g. unit tests) but not fdctl.

### DIFF
--- a/src/util/log/fd_log.c
+++ b/src/util/log/fd_log.c
@@ -1008,6 +1008,13 @@ fd_log_private_sig_abort( int         sig,
                           void *      context ) {
   (void)info; (void)context;
 
+  /* Thread could have caught signal while holding a lock.
+     Hack around this re-entrancy problem by pointing the log lock to
+     a dummy buffer. */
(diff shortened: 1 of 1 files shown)


src/util/log/fd_log.c
```diff
@@ -1008,6 +1008,13 @@ fd_log_private_sig_abort( int         sig,
                           void *      context ) {
   (void)info; (void)context;
 
+  /* Thread could have caught signal while holding a lock.
+     Hack around this re-entrancy problem by pointing the log lock to
+     a dummy buffer. */
+  int * old_lock = fd_log_private_shared_lock;
+  static FD_TL int lock = 0;
```
mmcgee-jump, line 1015: Why static?

ripatel-fd: I thought it's rather ugly to have a global pointer to stack memory. Such a setup is susceptible to stack corruption

PR #6102, opened 2025-08-19
replay, flamenco: safer publishing

yufeng-jump:
The replay tile unconditionally publishes upon receiving a new tower root.
This causes premature pruning of minority forks.  Specifically, a bank object
on a minority fork might get pruned before we've finished replaying the block
to which the bank is associated.  This causes a crash every once in a while.
We introduce reference counting to solve this issue.

This commit does not introduce logic for aborting execution of a dead block.
That will be added in a follow-up commit.

### DIFF
--- a/src/discof/replay/fd_replay_tile.c
+++ b/src/discof/replay/fd_replay_tile.c
@@ -229,9 +229,143 @@ struct fd_replay_tile_ctx {
   /* Metadata updated during execution */
 
   ulong   snapshot_slot;
-  ulong   root; /* the root slot is the most recent slot to have reached
-                   max lockout in the tower  */
-  ulong   bank_idx;
+
+  /* A note on publishing ...
+
+     The watermarks are used to publish our fork-aware structures.  For
+     example, store, banks, and txncache need to be published to release
+     resources occupied by rooted or dead blocks.  In general,
+     publishing has the effect of pruning forks in those structures,
+     indicating that it is ok to release the memory being occupied by
+     the blocks on said forks.  Tower is responsible for informing us of
+     the latest block on the consensus rooted fork.  As soon as we can,
+     we should move the published root as close as possible to the
+     latest consensus root, publishing/pruning everything on the fork
+     tree along the way.  That is, all the blocks that directly descend
+     from the current published root (inclusive) to the new published
+     root (exclusive) on the rooted fork, as well as all the minority
+     forks that branch from said blocks.
+
+     Ideally, we'd move the published root to the consensus root
+     immediately upon receiving a new consensus root.  However, that's
+     not always safe to do.  One thing we need to be careful about is
+     making sure that there are no more users/consumers of
+     soon-to-be-pruned blocks, lest a use-after-free occurs.  This can
+     be done by using a reference counter for each block.  Any
+     concurrent activity, such as transaction execution through the
+     exec-writer pipeline, should retain a refcnt on the block for as
+     long as it needs access to the shared fork-aware structures related
+     to that block.  Eventually, refcnt on a given block will drop down
+     to 0 as the block either finishes replaying or gets marked as dead,
+     and any other tile that has retained a refcnt on the block releases
+     it.  At that point, it becomes a candidate for pruning.  The key to
+     safe publishing then becomes figuring out how far we could advance
+     the published root, such that every minority fork branching off of
+     blocks in between the current published root (inclusive) and the
+     new published root (exclusive) is safe to be pruned.  This is a
+     straightforward tree traversal, where if a block B on the rooted
+     fork has refcnt 0, and all minority forks branching off of B also
+     have refcnt 0, then B is safe to be pruned.  We advance the
+     published root to the farthest consecutively prunable block on the
+     rooted fork.  Note that reasm presents the replay tile with a clean
+     view of the world where every block is chained off of a parent
+     block.  So there are no orpahned/dangling tree nodes to worry
+     about.  The world is a nice single tree as far as replay is
+     concerned.
+
+     In the following fork tree, every node is a block and the number in
+     parentheses is the refcnt on the block.  The chain marked with
+     double slashes is the rooted fork.  Suppose the published root is
+     at block P, and consensus root is at block T.  We can't publish
+     past block P because Q has refcnt 1.
+
+
+          P(0)
+        /    \\
+      Q(1)    A(0)
+            / ||  \
+        X(0) B(0)  C(0)
+       /      || \
+      Y(0)   M(0) R(0)
+            / ||   /  \
+        D(2) T(0) J(0) L(0)
+              ||
+              ..
+              ..
+              ..
+              ||
+      blocks we might be actively replaying
+
+
+     When refcnt on Q drops to 0, we would be able to advance the
+     published root to block M, because blocks P, A, and B, as well as
+     all subtrees branching off of them, have refcnt 0, and therefore
+     can be pruned.  Block M itself cannot be pruned yet because its
+     child block D has refcnt 2.  After publishing/pruning, the fork
+     tree would be:
+
+
+             M(0)
+            / ||
+        D(2) T(0)
+              ||
+              ..
+              ..
+              ..
+              ||
+      blocks we might be actively replaying
+
+
+     As a result, the shared fork-aware structures can free resources
+     for blocks P, A, B, and all subtrees branching off of them.
+
+     For the reference counting part, the replay tile is the sole entity
+     that can update the refcnt.  This ensures that all refcnt increment
+     and decrement attempts are serialized at the replay tile, and that
+     there are no racy resurrection of a soon-to-be-pruned block.  If a
+     refcnt increment request arrives after a block has been pruned,
+     replay simply rejects the request.
+
+     A note on the implementation of the above ...
+
+     Upon receiving a new consensus root, we descend down the rooted
+     fork from the current published root to the new consensus root.  On
+     each node/block of the rooted fork, we do a summation of the refcnt
+     on the block and all the minority fork blocks branching from the
+     block.  If the summation is 0, the block is safe for pruning.  We
+     advance the published root to the far end of the consecutive run of
+     0 refcnt sums originating from the current published root.  On our
+     descent down the minority forks, we also mark any block that hasn't
+     finished replaying as dead, so we don't waste time executing them.
+     No more transactions shall be dispatched for execution from dead
+     blocks.
+
+     Blocks start out with a refcnt of 0.  Other tiles may send a
+     request to the replay tile for a reference on a block.  The
+     transaction dispatcher is another source of refcnt updates.  On
+     every dispatch of a transaction for block B, we increment the
+     refcnt for B.  And on every transaction finalization, we decrement
+     the refcnt for B.  This means that whenever the refcnt on a block
+     is 0, there is no more reference on that block from the execution
+     pipeline.  While it might be tempting to simply increment the
+     refcnt once when we start replaying a block, and decrement the
+     refcnt once when we finish a block, this more fine-grained refcnt
+     update strategy allows for aborting and potentially immediate
+     pruning of blocks under interleaved block replay.  Upon receiving a
+     new consensus root, we can simply look at the refcnt on minority
+     fork blocks, and a refcnt of 0 would imply that the block is safe
+     for pruning, even if we haven't finished replaying it.  Without the
+     fine-grained refcnt, we would need to first stop dispatching from
+     the aborted block, and then wait for a full drain of the execution
+     pipeline to know for sure that there are no more in-flight
+     transactions executing on the aborted block.  Note that this will
+     allow the refcnt on any block to transiently drop down to 0.  We
+     will not mistakenly prune an actively replaying block, aka a leaf
+     node, that is chaining off of the rooted fork, because the
+     consensus root is always an ancestor of the actively replaying tip.
+     */
+  ulong consensus_root; /* The most recent block to have reached max lockout in the tower. */
+  ulong published_root; /* The most recent block to have been published, always <=consensus_root. */
 
   /* Other metadata */
 
@@ -321,8 +455,8 @@ scratch_footprint( fd_topo_tile_t const * tile FD_PARAM_UNUSED ) {
   for( ulong i = 0UL; i<FD_PACK_MAX_BANK_TILES; i++ ) {
     l = FD_LAYOUT_APPEND( l, FD_BMTREE_COMMIT_ALIGN, FD_BMTREE_COMMIT_FOOTPRINT(0) );
   }
-  l = FD_LAYOUT_APPEND( l, 128UL, FD_SLICE_MAX );
   l = FD_LAYOUT_APPEND( l, block_id_map_align(), block_id_map_footprint( fd_ulong_find_msb( fd_ulong_pow2_up( FD_BLOCK_MAX ) ) ) );
+  l = FD_LAYOUT_APPEND( l, 128UL, FD_SLICE_MAX );
   l = FD_LAYOUT_APPEND( l, fd_exec_slice_map_align(), fd_exec_slice_map_footprint( 20 ) );
   l = FD_LAYOUT_FINI  ( l, scratch_align() );
   return l;
@@ -390,14 +524,6 @@ txncache_publish( fd_replay_tile_ctx_t * ctx,
                   fd_funk_txn_t *        to_root_txn,
                   fd_funk_txn_t *        rooted_txn ) {
 
-
-  /* For the status cache, we stop rooting until the status cache has been
-     written out to the current snapshot. We also need to iterate up the
-     funk transaction tree up until the current "root" to figure out what slots
-     should be registered. This root can correspond to the latest false root if
-     one exists.  */
-
-
   if( FD_UNLIKELY( !ctx->slot_ctx->status_cache ) ) {
     return;
   }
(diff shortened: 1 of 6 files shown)


src/discof/replay/fd_replay_tile.c
```diff
@@ -884,12 +1103,26 @@ fd_replay_process_solcap_account_update( fd_replay_tile_ctx_t *
...
+      fd_refcnt_rel( refcnt );
+      /* Reference counter just decreased, so check if more slots just
+         became publishable. */
+      if( ctx->tower_wmark>ctx->publishable_wmark ) {
+        try_publish( ctx );
+      }
+      /* An exec tile just got freed up, so check if the execution
+         pipeline is fully drained.*/
+      if( ctx->drain_mode && exec_pipeline_is_drained( ctx ) ) {
+        publish( ctx, ctx->publishable_wmark );
```
cali-jumptrading: what's the difference between `try_publish` and the logic to publish here?

src/discof/replay/fd_replay_tile.c
```diff
@@ -1492,6 +1723,12 @@ exec_and_handle_slice( fd_replay_tile_ctx_t * ctx, fd_stem_context_t * stem ) {
        over all accounts a second time. */
     fd_runtime_update_program_cache( ctx->slot_ctx, &txn_p, ctx->runtime_spad );
 
+    /* At this point, we are going to send the txn down the execution
+       pipeline. Increment the refcnt so we don't prematurely prune a
+       bank that's needed by an in-flight txn. */
+    fd_refcnt_t * refcnt = fd_bank_refcnt_modify( ctx->slot_ctx->bank );
```
cali-jumptrading: why not increment ref count in the exec tile?

src/discof/replay/fd_replay_tile.c
yufeng-jump: try_publish() contains logic for checking whether there is something to publish, and whether it's safe to publish.  publish() simply calls all the publish routines on the shared data structures, assuming the slot it's given is safe to be published up to.  Perhaps a better name for publish() would be do_publish().

src/discof/replay/fd_replay_tile.c
yufeng-jump: We could.  Fine either way.  Doing it in replay in a loop will probably be fewer cache misses.

src/discof/replay/fd_replay_tile.c
cali-jumptrading: like if you already had the checks in `try_publish` couldn't you call that here instead of `publish`?

src/flamenco/fd_refcnt.h
```diff
@@ -0,0 +1,286 @@
+#ifndef HEADER_fd_src_flamenco_refcnt_h
+#define HEADER_fd_src_flamenco_refcnt_h
+
```
ibhatt-jumptrading: add unit tests

src/flamenco/runtime/fd_bank.h
```diff
@@ -143,6 +144,9 @@ FD_PROTOTYPES_BEGIN
 
 #define FD_BANKS_ITER(X)                                                                                                                                                                                                                               \
   /* type,                             name,                        footprint,                                 align,                                      CoW, limit fork width, has lock */                                                          \
+  X(fd_refcnt_t,                       refcnt,                      sizeof(fd_refcnt_t),                       alignof(fd_refcnt_t),                       0,   0,                0    )  /* Reference counter */                                      \
+  X(ulong,                             flags,                       sizeof(ulong),                             alignof(ulong),                             0,   0,                0    )  /* Flags */                                                  \
```
ibhatt-jumptrading: make it more clear what flags is (change name to more descriptive and add a better description

src/flamenco/runtime/fd_bank.h
```diff
@@ -498,35 +517,22 @@ FD_BANKS_ITER(X)
...
-static inline void
-fd_banks_lock( fd_banks_t * banks ) {
-  fd_rwlock_read( &banks->rwlock );
-}
-
-static inline void
-fd_banks_unlock( fd_banks_t * banks ) {
-  fd_rwlock_unread( &banks->rwlock );
-}
+#define FD_BANK_FLAGS_INIT     (0x00000000UL)
```
ibhatt-jumptrading: describe when/why we are in these different states

src/flamenco/runtime/fd_bank.h
```diff
@@ -143,6 +144,9 @@ FD_PROTOTYPES_BEGIN
 
 #define FD_BANKS_ITER(X)                                                                                                                                                                                                                               \
   /* type,                             name,                        footprint,                                 align,                                      CoW, limit fork width, has lock */                                                          \
+  X(fd_refcnt_t,                       refcnt,                      sizeof(fd_refcnt_t),                       alignof(fd_refcnt_t),                       0,   0,                0    )  /* Reference counter */                                      \
```
ibhatt-jumptrading: my instinct would be to maintain the refcnt in the replay tile since we need to maintain one for the funk txn as well. It would also simplify the bank if we have no notion of if the refcnt that is handled by the bank.

src/flamenco/runtime/fd_bank.c
```diff
@@ -463,6 +463,10 @@ fd_banks_init_bank( fd_banks_t * banks, ulong slot ) {
   #undef HAS_LOCK_0
   #undef HAS_LOCK_1
```
ibhatt-jumptrading, line 476: maybe the overall flow in the bank should be that 

1. when we init/clone a bank we set the flag to be == INIT
2. as mentioned below, the caller is completely responsible for handling the refcnt
3. maybe we add something in publish that checks if the ancestry from old_root-->new_root contains any dead blocks and log_crit out if that is the case

src/discof/replay/fd_replay_tile.c
```diff
@@ -248,19 +246,51 @@ struct fd_replay_tile_ctx {
...
+     that point, it becomes a candidate for pruning.  The key to safe
+     publishing then becomes figuring out how far we could advance the
+     watermark, such that every minority fork branching off of blocks in
+     between the current watermark (inclusive) and the new watermark
+     (exclusive) is safe to be pruned.  This is a straightforward
+     BFS-ish greedy descent down the fork tree.  Fortunately, reasm
+     presents the replay tile with a clean view of the world where every
+     block is chained off of a parent block.  So there are no orpahned
+     nodes to worry about.  The world is a nice single tree as far as
+     replay is concerned.
```
ibhatt-jumptrading: maybe make it a little more clear that our strategy for publishing is basically always to publish the min( tower_root, replay_root )

src/discof/replay/fd_replay_tile.c
```diff
@@ -838,6 +869,201 @@ on_snapshot_message( fd_replay_tile_ctx_t * ctx,
...
+   the banks fork tree can be pruned if
+   (1) the node itself can be pruned, and
+   (2) all subtrees (except for the one on the rooted fork) forking off
+       of the node can be pruned.
+
+   The highest publishable slot is the child slot of the highest slot on
+   the rooted fork where the above is true.
+
+   Returns 0 if no such slot can be found. */
+static ulong
```
ibhatt-jumptrading: i wonder if this is worth keying to block-id since that is what all of the data structures that we care about publishing will eventually use

src/discof/replay/fd_replay_tile.c
```diff
@@ -838,6 +869,201 @@ on_snapshot_message( fd_replay_tile_ctx_t * ctx,
...
+    return 0UL;
+  }
+
+  /* Build path from old root to target. */
+  ulong       depth      = 0UL;
+  fd_bank_t * curr       = target_bank;
+  ulong       parent_idx = fd_bank_parent_idx_get( curr );
+  fd_bank_t * parent     = fd_banks_pool_ele( bank_pool, parent_idx );
+  while( parent ) {
+    fd_bank_rooted_child_slot_set( parent, fd_bank_slot_get( curr ) );
```
ibhatt-jumptrading: it feels kind of hacky to maintain this rooted_child_slot field in the bank

src/discof/replay/fd_replay_tile.c
```diff
@@ -838,6 +869,201 @@ on_snapshot_message( fd_replay_tile_ctx_t * ctx,
...
+  fd_funk_txn_xid_t xid = { .ul = { slot, slot } };
+  funk_and_txncache_publish( ctx, slot, &xid );
+
+  fd_banks_publish( ctx->banks, slot );
+
+  fd_fseq_update( ctx->published_wmark, slot );
+}
+
+/* Advance the published watermark as much as we can. */
+static void
```
ibhatt-jumptrading: do we need drain mode? can't we just rely on how many outstanding references we have to the bank.

src/flamenco/runtime/fd_bank.h
```diff
@@ -498,35 +517,22 @@ FD_BANKS_ITER(X)
...
-// publishing issue.  Remove these once we have refcnts.
-/* fd_banks_lock() and fd_banks_unlock() are locks to be acquired and
-   freed around accessing or modifying a specific bank. This is only
-   required if there is concurrent access to a bank while operations on
-   its underlying map are being performed.
-
-   Under the hood, this is a wrapper around fd_banks_t's rwlock.
-   This is done so a caller can safely read/write a specific bank.
-   Otherwise, we run the risk of accessing/modifying a bank that may be
-   freed. This is acquiring and freeing a read lock around fd_banks_t. */
```
ibhatt-jumptrading: i think we still need these locks actually around calls to fd_banks_get_bank() and in fd_banks_publish(), but we don't need to hold the read lock for the entire duration of its use in exec or writer tile

src/discof/replay/fd_replay_tile.c
yufeng-jump: I thought about that.  I'm in favor of keying by block_id.  Charlie added a slot=>block_id mapping.  I think to key this PR by block_id, we might need a block_id=>slot reverse mapping too, because stuff like banks still key by slot, and we need a way to figure out a slot number given a block_id, in order to interact with all the other keyed by slot structures.  Once we move entirely to block_id, we should be able to get rid of these mappings.

src/flamenco/runtime/fd_bank.h
yufeng-jump: Makes sense.  We already have one in publish().  Will add to other APIs.

src/flamenco/runtime/fd_bank.h
```diff
@@ -292,14 +300,16 @@ FD_PROTOTYPES_BEGIN
...
+  #define FD_BANK_HEADER_SIZE (56UL)
 
   /* Fields used for internal pool and bank management */
   ulong             slot_;       /* slot this node is tracking, also the map key */
   ulong             next;        /* reserved for internal use by fd_pool_para, fd_map_chain_para and fd_banks_publish */
   ulong             parent_idx;  /* index of the parent in the node pool */
   ulong             child_idx;   /* index of the left-child in the node pool */
   ulong             sibling_idx; /* index of the right-sibling in the node pool */
+  ulong             flags;       /* keeps track of the state of the bank, as well as some configurations */
+  ulong             refcnt;      /* reference count on the bank */
```
ibhatt-jumptrading: can you add more details about the refcnt

src/flamenco/runtime/fd_bank.h
ibhatt-jumptrading, line 1: also can we add unit tests in test_bank around the refcnt handling

src/flamenco/runtime/fd_bank.h
```diff
@@ -758,6 +773,28 @@ fd_bank_t *
 fd_banks_rekey_root_bank( fd_banks_t * banks,
                           ulong        slot );
 
+/* Returns the highest block that can be safely published between the
```
ibhatt-jumptrading, line 779: maybe be more explicit here that something is publishable (or can be pruned) if there are no outstanding references on the bank or any of its ancestors.

also maybe not bad idea to leave a note at the bottom saying how it works ( e.g. tree traversal that is a binary summing tree)

src/flamenco/runtime/fd_bank.c
```diff
@@ -759,6 +765,10 @@ fd_banks_publish( fd_banks_t * banks, ulong slot ) {
     return NULL;
   }
 
+  if( FD_UNLIKELY( old_root->refcnt!=0UL ) ) {
```
ibhatt-jumptrading, line 768: if we try to publish something with a non-zero refcnt we should do more than just log_warning. Maybe we should exit without publishing anything or even log_crit

src/flamenco/runtime/fd_bank.c
```diff
@@ -783,6 +793,9 @@ fd_banks_publish( fd_banks_t * banks, ulong slot ) {
     while( FD_LIKELY( child ) ) {
 
       if( FD_LIKELY( child!=new_root ) ) {
+        if( FD_UNLIKELY( child->refcnt!=0UL ) ) {
```
ibhatt-jumptrading, line 796: same thing here

src/flamenco/runtime/fd_bank.c
```diff
@@ -914,3 +927,151 @@ fd_banks_rekey_root_bank( fd_banks_t * banks, ulong slot ) {
 
   return bank;
 }
+
+/* Is the fork tree starting at the given bank entirely eligible for
```
ibhatt-jumptrading, line 931: add what defines is something is allowed to be pruned

src/flamenco/runtime/fd_bank.c
```diff
@@ -914,3 +927,151 @@ fd_banks_rekey_root_bank( fd_banks_t * banks, ulong slot ) {
...
+fd_banks_subtree_can_be_pruned( fd_bank_t * bank_pool, fd_bank_t * bank ) {
+  if( FD_UNLIKELY( !bank ) ) {
+    FD_LOG_CRIT(( "invariant violation: bank is NULL" ));
+  }
+
+  if( bank->refcnt!=0UL ) {
+    return 0;
+  }
+
+  /* Recursively check all children. */
```
ibhatt-jumptrading, line 945: don't love the fact that this is recursive since it adds more function calls, i wonder if we should do all of these recursively

src/flamenco/runtime/fd_bank.c
```diff
@@ -914,3 +927,151 @@ fd_banks_rekey_root_bank( fd_banks_t * banks, ulong slot ) {
...
+  ulong child_idx = bank->child_idx;
+  while( child_idx!=fd_banks_pool_idx_null( bank_pool ) ) {
+    fd_bank_t * child = fd_banks_pool_ele( bank_pool, child_idx );
+    fd_banks_subtree_mark_dead( bank_pool, child );
+    child_idx = child->sibling_idx;
+  }
+}
+
+int
+fd_banks_publish_prepare( fd_banks_t *           banks,
```
ibhatt-jumptrading: nit: extra spaces here

src/flamenco/runtime/fd_bank.c
```diff
@@ -914,3 +927,151 @@ fd_banks_rekey_root_bank( fd_banks_t * banks, ulong slot ) {
...
+  /* Mark rooted fork. */
+  fd_bank_t * curr = target_bank;
+  fd_bank_t * prev = NULL;
+  while( curr ) {
+    curr->flags |= FD_BANK_FLAGS_ROOTED;
+    prev         = curr;
+    curr         = fd_banks_pool_ele( bank_pool, curr->parent_idx );
+  }
+
+  /* If we didn't reach the old root, target is not a descendant. */
```
ibhatt-jumptrading, line 1013: this exit condition seems not ideal since we mark a set of banks as FD_BANK_FLAGS_ROOTED even if they might not be

src/flamenco/runtime/fd_bank.c
```diff
@@ -914,3 +927,151 @@ fd_banks_rekey_root_bank( fd_banks_t * banks, ulong slot ) {
...
+  if( prev!=root ) {
+    FD_LOG_WARNING(( "target slot %lu is not a descendant of root slot %lu", target_slot, fd_bank_slot_get( root ) ));
+    return 0;
+  }
+
+  /* Now traverse from root towards target and find the highest
+     publishable block. */
+  ulong highest_publishable_block = 0UL;
```
ibhatt-jumptrading, line 1020: fix indents here

src/flamenco/runtime/fd_bank.c
```diff
@@ -914,3 +927,151 @@ fd_banks_rekey_root_bank( fd_banks_t * banks, ulong slot ) {
 
   return bank;
 }
+
+/* Is the fork tree starting at the given bank entirely eligible for
+   pruning?  Returns 1 for yes, 0 for no. */
+static int
+fd_banks_subtree_can_be_pruned( fd_bank_t * bank_pool, fd_bank_t * bank ) {
```
ibhatt-jumptrading, line 936: how do you handle the case of if something is dead but has references. Is this possible? worth documenting regardless

src/discof/replay/fd_replay_tile.c
```diff
@@ -1169,30 +1333,6 @@ handle_new_slice( fd_replay_tile_ctx_t * ctx, fd_stem_context_t * stem ) {
   int    slot_complete = slice.slot_complete;
```
ibhatt-jumptrading, line 1325: I think we are missing some handling for the case where the block is dead

src/discof/replay/fd_replay_tile.c
ibhatt-jumptrading, line 1: i would make it more explicit that the replay tile basically "owns" the refcnts on the banks and is the only caller
that should increment/decrement the refcnts

ibhatt-jumptrading (approved):

PR #6104, opened 2025-08-19
solcap: fix deduplication, solcap hash, solcap start slot bugs

kbhargava-jump:
changes made:
* deduplicate accounts when investigating solcap captures
* update to use original account hash logic for solcaps (want to match agave bank hash details file)
* update logic to account fields even when bank hashes match
* integration test that runs backtest with solcap creation and tests against agave
* fix solcap creation with capture start slot
* make dedup optional (yaml only)

### DIFF
--- a/contrib/test/run_solcap_tests.sh
+++ b/contrib/test/run_solcap_tests.sh
@@ -1,0 +1,33 @@
+DUMP_DIR=${DUMP_DIR:="./dump"}
+echo $OBJDIR
+
+LEDGER="devnet-398736132"
+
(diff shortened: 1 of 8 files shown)


contrib/test/run_solcap_tests.sh
```diff
@@ -0,0 +1,33 @@
+DUMP_DIR=${DUMP_DIR:="./dump"}
+echo $OBJDIR
+
+LEDGER="devnet-398736132"
```
topointon-jump, line 4: nit, but could we call this solcap-test-ledger?

topointon-jump (approved):

PR #6105, opened 2025-08-20
runtime: fix sysvar cache function name mismatch

two-heart:
match the name to the .h declaration


ripatel-fd:
Hmm we can probably just delete these APIs, they're unused

two-heart:
I think there is value in keeping these around, since other similar functions are used. If in the future someone needs the slot history, there it is clear that it should be obtained though this way, instead of ad-hoc pulling it out some struct.

PR #6114, opened 2025-08-20
runtime: stake delegation deltas in bank

ibhatt-jumptrading:
reduces memory footprint in the bank to ~64gb for 1024 total banks


src/flamenco/runtime/fd_bank.c
```diff
@@ -618,11 +629,111 @@ fd_banks_clone_from_parent( fd_banks_t * banks,
...
+                                       fd_bank_t *              bank,
+                                       fd_stake_delegations_t * stake_delegations ) {
+
+  /* Naively what we want to do is iterate from the old root to the new
+     root and apply the delta to the full state iteratively. */
+
+  /* First, gather all of the pool indicies that we want to apply deltas
+     for in reverse order starting from the new root. We want to exclude
+     the old root since its delta has been applied previously. */
+  ulong pool_indicies[ banks->max_total_banks ];
```
ripatel-fd, line 698: I'm not sure how many banks there can be in total, but making a stack allocated array bounded by banks seems a bit risky to me. Probably fine for now

src/flamenco/runtime/test_bank.c
```diff
@@ -13,6 +24,10 @@ main( int argc, char ** argv ) {
                                                 0UL );
   FD_TEST( wksp );
 
+  FD_LOG_WARNING(("BANK FOOTPRINT %lu", fd_banks_footprint( 1024UL, 16UL ) ));
+  FD_LOG_WARNING(("VOTE STATES FOOTPRINT %lu", fd_vote_states_footprint( FD_RUNTIME_MAX_VOTE_ACCOUNTS ) ));
+  FD_LOG_WARNING(("VOTE STATES FOOTPRINT %lu", fd_vote_states_footprint( 4096 ) ));
```
ripatel-fd, line 29: Could you remove the WARNING prints here? 
(Nice to keep test logs clean so we can grep for warnings across all tests)

src/flamenco/runtime/fd_bank.c
ibhatt-jumptrading: added a todo

src/flamenco/runtime/fd_bank.h
```diff
@@ -444,7 +491,71 @@ fd_bank_slot_get( fd_bank_t const * bank ) {
...
+
+static inline void
+fd_bank_stake_delegations_delta_end_locking_query( fd_bank_t * bank ) {
+  fd_rwlock_unread( &bank->stake_delegations_delta_lock );
+}
+
+/* fd_bank_stake_delegations_frontier_query() will return a pointer to
+   the full stake delegations for the current frontier. The caller is
+   responsible that there are no concurrent readers or writers to
+   the stake delegations returned by this function.
```
yufeng-jump, line 539: Seems like it's not just that there can be no concurrent reading/writing to the returned delegations, but that there can be no concurrent calls to this function at all, because there's only one instance of the underlying buffer?  If that's the case, might be worth adding a wlock() in this function and a query_end() wunlock() to enforce this invariant in code?  I understand that the only caller right now is the epoch processing code.  But a wlock is cheap and would be an obvious deadlock rather than bizarre memory corruption if anyone ever violates the invariant.  A wlock() also would help guard against concurrent modifications of the fork tree.

yufeng-jump (dismissed):
LGTM 🔥

yufeng-jump (approved):

PR #6123, opened 2025-08-21
vote ts in seconds

akhinvasara-jumptrading:
This PR timestamps votes in seconds, rather than ns, to match agave's behavior. The current impl means any vote account used on full firedancer is impossible to use with frank/agave. 😅


src/choreo/tower/fd_tower.c
```diff
@@ -502,7 +502,7 @@ fd_tower_to_vote_txn( fd_tower_t const *    tower,
   tower_sync.root          = root;
   tower_sync.lockouts_len  = (ushort)fd_tower_votes_cnt( tower );
   tower_sync.lockouts      = lockouts_scratch;
-  tower_sync.timestamp     = fd_log_wallclock();
+  tower_sync.timestamp     = fd_log_wallclock()/1000000000L;
```
lidatong:
```suggestion
  tower_sync.timestamp     = fd_log_wallclock() / (long)1e9; /* seconds */
```

PR #6131, opened 2025-08-22
contributing: adding rules for multi-line function argument formatting

ibhatt-jumptrading:


kbhargava-jump (approved):

CONTRIBUTING.md
```diff
@@ -101,6 +101,44 @@ abort( );  /* WRONG! */
...
+    argument_four,
+    argument_five,
+    argument_six );
+
+int result = foo( argument_one, /* WRONG! */
+                  argument_two,
+                  argument_three,
+                  argument_four,
+                  argument_five,
+                  argument_six );
```
lidatong, line 138: these are all wrong

lidatong (changes requested):

CONTRIBUTING.md
lidatong: (line breaks are context-sensitive, it depends on the calling context and logical argument grouping -- otherwise having the entire call span one line is preferred based on the original code)

PR #6133, opened 2025-08-22
shred: change fec resolver to not spill to done map

emwang-jump:
so many bugs discovered 🐛 🐛  🐛  
 🐛  `force_complete` didn't remove from curr_map , but it would release the memory back to freelist, so we could see things in the curr_map that were weird af
 🐛  no pool_release in consumed_map in forest (in both publish and in advance_consumed 💀 )
 🐛  repair.c inflight table insert doesn't return NULL on no space, and would eventually overwrite the map_private_t fields
 🐛  store lock not released in replay
 
burst value: burst for the tile is already large enough due to the max burst publishing to store in frankendancer path; link burst remains <= 2 bc if an eviction happens that means there is no fec complete msg

### DIFF
--- a/src/disco/shred/fd_fec_resolver.c
+++ b/src/disco/shred/fd_fec_resolver.c
@@ -309,19 +309,15 @@ ctx_ll_insert( set_ctx_t * p, set_ctx_t * c ) {
   return c;
 }
 
-
 int
-fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
-                           fd_shred_t const     * shred,
-                           ulong                  shred_sz,
-                           uchar const          * leader_pubkey,
-                           fd_fec_set_t const * * out_fec_set,
-                           fd_shred_t const   * * out_shred,
-                           fd_bmtree_node_t     * out_merkle_root,
-                           ulong                * out_spilled_slot,
-                           uint                 * out_spilled_fec_set_idx,
-                           uint                 * out_max_dshred_idx ) {
-
+fd_fec_resolver_add_shred( fd_fec_resolver_t         * resolver,
+                           fd_shred_t const          * shred,
+                           ulong                       shred_sz,
+                           uchar const               * leader_pubkey,
+                           fd_fec_set_t const      * * out_fec_set,
+                           fd_shred_t const        * * out_shred,
+                           fd_bmtree_node_t          * out_merkle_root,
+                           fd_fec_resolver_spilled_t * out_spilled ) {
   /* Unpack variables */
   ulong partial_depth = resolver->partial_depth;
   ulong done_depth    = resolver->done_depth;
@@ -425,25 +421,27 @@ fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
       set_ctx_t * victim_ctx = resolver->curr_ll_sentinel->prev;
 
 
-      if( FD_LIKELY( out_spilled_slot || out_spilled_fec_set_idx || out_max_dshred_idx ) ) {
+      if( FD_LIKELY( out_spilled ) ) {
         fd_fec_set_t * set = victim_ctx->set;
 
         /* Find the highest data shred received in the FEC set */
 
         ulong max_rcvd_dshred_idx = d_rcvd_last( set->data_shred_rcvd );
         fd_shred_t const * max_shred;
-        if( max_rcvd_dshred_idx == ~0UL ) {
+        if( FD_UNLIKELY( max_rcvd_dshred_idx == ~0UL ) ) {
           max_rcvd_dshred_idx = FD_SHRED_BLK_MAX; /* No data shreds received. Use a parity shred to determine the spilled slot and fec_set_idx */
           max_shred = fd_shred_parse( set->parity_shreds[ p_rcvd_first( set->parity_shred_rcvd ) ], FD_SHRED_MAX_SZ );
         } else {
           max_shred = fd_shred_parse( set->data_shreds  [ max_rcvd_dshred_idx ],                    FD_SHRED_MIN_SZ );
         }
 
-        if( FD_LIKELY( out_spilled_slot        ) ) *out_spilled_slot        = max_shred->slot;
-        if( FD_LIKELY( out_spilled_fec_set_idx ) ) *out_spilled_fec_set_idx = max_shred->fec_set_idx;
-        if( FD_LIKELY( out_max_dshred_idx      ) ) *out_max_dshred_idx      = (uint)max_rcvd_dshred_idx;
+        if( FD_LIKELY( out_spilled ) ) {
+          out_spilled->slot           = max_shred->slot;
+          out_spilled->fec_set_idx    = max_shred->fec_set_idx;
+          out_spilled->max_dshred_idx = (uint)max_rcvd_dshred_idx;
+        }
 
-        FD_LOG_INFO(("Thrashed from fec_resolver in-progress map %lu %u, data_shreds_rcvd %lu, parity_shreds_rcvd %lu, max_d_rcvd_shred_idx %lu", max_shred->slot, max_shred->fec_set_idx, d_rcvd_cnt( set->data_shred_rcvd ), p_rcvd_cnt( set->parity_shred_rcvd ), max_rcvd_dshred_idx ));
+        FD_LOG_INFO(("Spilled from fec_resolver in-progress map %lu %u, data_shreds_rcvd %lu, parity_shreds_rcvd %lu, max_d_rcvd_shred_idx %lu", max_shred->slot, max_shred->fec_set_idx, d_rcvd_cnt( set->data_shred_rcvd ), p_rcvd_cnt( set->parity_shred_rcvd ), max_rcvd_dshred_idx ));
       }
 
       freelist_push_tail( free_list,        victim_ctx->set  );
--- a/src/disco/shred/fd_fec_resolver.c
+++ b/src/disco/shred/fd_fec_resolver.c
@@ -309,13 +309,19 @@ ctx_ll_insert( set_ctx_t * p, set_ctx_t * c ) {
   return c;
 }
 
-int fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
-                               fd_shred_t const     * shred,
-                               ulong                  shred_sz,
-                               uchar const          * leader_pubkey,
-                               fd_fec_set_t const * * out_fec_set,
-                               fd_shred_t const   * * out_shred,
-                               fd_bmtree_node_t     * out_merkle_root ) {
+
+int
+fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
+                           fd_shred_t const     * shred,
+                           ulong                  shred_sz,
+                           uchar const          * leader_pubkey,
+                           fd_fec_set_t const * * out_fec_set,
+                           fd_shred_t const   * * out_shred,
+                           fd_bmtree_node_t     * out_merkle_root,
+                           ulong                * out_spilled_slot,
+                           uint                 * out_spilled_fec_set_idx,
+                           uint                 * out_max_dshred_idx ) {
+
   /* Unpack variables */
   ulong partial_depth = resolver->partial_depth;
   ulong done_depth    = resolver->done_depth;
@@ -406,7 +412,7 @@ int fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
   /* This, combined with the check on shred->code.data_cnt implies that
      shred_idx is in [0, DATA_SHREDS_MAX+PARITY_SHREDS_MAX). */
 
-  if( FD_UNLIKELY( tree_depth>FD_SHRED_MERKLE_LAYER_CNT-1UL             ) ) return FD_FEC_RESOLVER_SHRED_REJECTED;
+  if( FD_UNLIKELY( tree_depth>FD_SHRED_MERKLE_LAYER_CNT-1UL          ) ) return FD_FEC_RESOLVER_SHRED_REJECTED;
   if( FD_UNLIKELY( fd_bmtree_depth( shred_idx+1UL ) > tree_depth+1UL ) ) return FD_FEC_RESOLVER_SHRED_REJECTED;
 
   if( FD_UNLIKELY( !ctx ) ) { /* This is the first shred in the FEC set */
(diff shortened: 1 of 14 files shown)


src/disco/shred/fd_fec_resolver.h
```diff
@@ -193,14 +193,32 @@ fd_fec_resolver_set_shred_version( fd_fec_resolver_t * resolver,
...
+typedef struct fd_fec_resolver_res fd_fec_resolver_res_t;
+
+fd_fec_resolver_res_t
+fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
+                           fd_shred_t const     * shred,
+                           ulong                  shred_sz,
+                           uchar const          * leader_pubkey,
+                           fd_fec_set_t const * * out_fec_set,
+                           fd_shred_t const   * * out_shred,
+                           fd_bmtree_node_t     * out_merkle_root );
```
lidatong: out parameters instead of copying struct back (minimize api surface / cleaner api)

src/disco/shred/fd_shred_tile.c
```diff
@@ -914,6 +917,15 @@ after_frag( fd_shred_ctx_t *    ctx,
...
     }
 
+    if( FD_UNLIKELY( res.thrashed_slot > 0 && res.thrashed_fec_set_idx < FD_SHRED_BLK_MAX ) ) {
+      /* We've thrashed a FEC set in the fec_resolver. We need to let
+         repair know to clear out it's cached info for that fec set and
+         re-repair those shreds. */
+      ulong sig_ = fd_disco_shred_repair_shred_sig( 0, res.thrashed_slot, (uint)res.thrashed_fec_set_idx, 0, (uint)res.max_shred_idx );
+      fd_stem_publish( stem, ctx->repair_out_idx, sig_, ctx->repair_out_chunk, 0, 0, ctx->tsorig, ctx->tsorig );
+      ctx->repair_out_chunk = fd_dcache_compact_next( ctx->repair_out_chunk, 0, ctx->repair_out_chunk0, ctx->repair_out_wmark );
+    }
```
lidatong: only advance repair forest on fec complete instead (still keep this frag to clear forest so it will repair spilled shreds)

lidatong: also you probably mean "spilled" in various places you use "thrashed" :)

src/disco/shred/fd_fec_resolver.c
```diff
@@ -418,10 +424,27 @@ int fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
...
+          max_shred = fd_shred_parse( set->parity_shreds[ p_rcvd_first( set->parity_shred_rcvd ) ], FD_SHRED_MAX_SZ );
+        } else {
+          max_shred = fd_shred_parse( set->data_shreds  [ max_rcvd_dshred_idx ],                    FD_SHRED_MIN_SZ );
+        }
+
+        if( FD_LIKELY( out_spilled_slot        ) ) *out_spilled_slot        = max_shred->slot;
+        if( FD_LIKELY( out_spilled_fec_set_idx ) ) *out_spilled_fec_set_idx = max_shred->fec_set_idx;
+        if( FD_LIKELY( out_max_dshred_idx      ) ) *out_max_dshred_idx      = (uint)max_rcvd_dshred_idx;
+
+        FD_LOG_INFO(("Thrashed from fec_resolver in-progress map %lu %u, data_shreds_rcvd %lu, parity_shreds_rcvd %lu, max_d_rcvd_shred_idx %lu", max_shred->slot, max_shred->fec_set_idx, d_rcvd_cnt( set->data_shred_rcvd ), p_rcvd_cnt( set->parity_shred_rcvd ), max_rcvd_dshred_idx ));
```
ptaffet-jump: I think you mean `evicted` here. Thrashing is a specific pattern that results in lots of eviction.

src/disco/shred/fd_fec_resolver.c
```diff
@@ -418,10 +424,27 @@ int fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
          set to the back of the free list. */
       set_ctx_t * victim_ctx = resolver->curr_ll_sentinel->prev;
 
-      /* Add this one that we're sacrificing to the done map to
-         prevent the possibility of thrashing. */
-      ctx_ll_insert( done_ll_sentinel, ctx_map_insert( done_map, victim_ctx->sig ) );
-      if( FD_UNLIKELY( ctx_map_key_cnt( done_map ) > done_depth ) ) ctx_map_remove( done_map, ctx_ll_remove( done_ll_sentinel->prev ) );
+
+      if( FD_LIKELY( out_spilled_slot || out_spilled_fec_set_idx || out_max_dshred_idx ) ) {
```
ptaffet-jump: The doc comment for the function does not mention that these are optional. In general, it seems like maybe you changed the design and forgot to update it.

src/disco/shred/fd_fec_resolver.h
```diff
@@ -193,14 +193,27 @@ fd_fec_resolver_set_shred_version( fd_fec_resolver_t * resolver,
...
-                               fd_fec_set_t const * * out_fec_set,
-                               fd_shred_t const   * * out_shred,
-                               fd_bmtree_node_t     * out_merkle_root );
+   populates any missing shreds in the FEC set stored in out_fec_set.
+
+   Along with the return err code, if adding this shred caused an
+   incomplete FEC set to be evicted from the current map, the slot and
+   FEC set idx of the evicted FEC set will be written to spilled_slot
+   and spilled_fec_set_idx.  The max data shred idx received thus far
+   of the evicted FEC set will be written to max_shred_idx.  If no FEC
```
ptaffet-jump: Comment needs to explain what happens if the evicted FEC set only has parity shreds

src/disco/shred/fd_fec_resolver.c
```diff
@@ -813,10 +836,15 @@ fd_fec_resolver_force_complete( fd_fec_resolver_t  *  resolver,
...
   /* Reject FEC set if it didn't validate and return mem to the pools */
+  set_ctx_t * done_ll_sentinel = resolver->done_ll_sentinel;
+  set_ctx_t * curr_map         = resolver->curr_map;
+  set_ctx_t * done_map         = resolver->done_map;
+  ulong       done_depth       = resolver->done_depth;
 
   if( FD_UNLIKELY( reject ) ) {
     freelist_push_tail( resolver->free_list,        ctx->set  );
     bmtrlist_push_tail( resolver->bmtree_free_list, ctx->tree );
+    ctx_map_remove( curr_map, ctx_ll_remove( ctx ));
```
ptaffet-jump, line 845: Was this just missing before?

src/disco/shred/fd_fec_resolver.h
```diff
@@ -193,14 +193,27 @@ fd_fec_resolver_set_shred_version( fd_fec_resolver_t * resolver,
...
-                               uchar const          * leader_pubkey,
-                               fd_fec_set_t const * * out_fec_set,
-                               fd_shred_t const   * * out_shred,
-                               fd_bmtree_node_t     * out_merkle_root );
+   populates any missing shreds in the FEC set stored in out_fec_set.
+
+   Along with the return err code, if adding this shred caused an
+   incomplete FEC set to be evicted from the current map, the slot and
+   FEC set idx of the evicted FEC set will be written to spilled_slot
+   and spilled_fec_set_idx.  The max data shred idx received thus far
```
ptaffet-jump: I first thought you meant the `shred->idx` field, but it seems you mean the actual index in the FEC set? Like a number in [0, 32)?

src/disco/shred/fd_fec_resolver.h
```diff
@@ -193,14 +193,27 @@ fd_fec_resolver_set_shred_version( fd_fec_resolver_t * resolver,
...
+fd_fec_resolver_add_shred( fd_fec_resolver_t    * resolver,
+                           fd_shred_t const     * shred,
+                           ulong                  shred_sz,
+                           uchar const          * leader_pubkey,
+                           fd_fec_set_t const * * out_fec_set,
+                           fd_shred_t const   * * out_shred,
+                           fd_bmtree_node_t     * out_merkle_root,
+                           ulong                * out_spilled_slot,
+                           uint                 * out_spilled_fec_set_idx,
+                           uint                 * out_max_dshred_idx );
```
ptaffet-jump: This function is getting a bit gross.  It's probably okay, but you could also define a struct containing these values, and populate it. That might help you clarify stuff too

src/disco/shred/fd_fec_resolver.h
```diff
@@ -193,14 +193,27 @@ fd_fec_resolver_set_shred_version( fd_fec_resolver_t * resolver,
...
-                               fd_bmtree_node_t     * out_merkle_root );
+   populates any missing shreds in the FEC set stored in out_fec_set.
+
+   Along with the return err code, if adding this shred caused an
+   incomplete FEC set to be evicted from the current map, the slot and
+   FEC set idx of the evicted FEC set will be written to spilled_slot
+   and spilled_fec_set_idx.  The max data shred idx received thus far
+   of the evicted FEC set will be written to max_shred_idx.  If no FEC
+   set was evicted, spilled_slot and max_shred_idx will be 0, and
+   spilled_fec_set_idx will be FD_SHRED_BLK_MAX. */
```
ptaffet-jump: It seems that the actual implementation doesn't touch the value if nothing was evicted, which is not what this says

src/disco/shred/fd_shred_tile.c
```diff
@@ -914,6 +917,15 @@ after_frag( fd_shred_ctx_t *    ctx,
       }
     }
 
+    if( FD_UNLIKELY( spilled_slot < ULONG_MAX && spilled_fec_set_idx < FD_SHRED_BLK_MAX && spilled_max_dshred_idx < FD_SHRED_BLK_MAX ) ) {
+      /* We've spilled an in-progress FEC set in the fec_resolver. We need to let
+         repair know to clear out it's cached info for that fec set and
+         re-repair those shreds. */
+      ulong sig_ = fd_disco_shred_repair_shred_sig( 0, spilled_slot, spilled_fec_set_idx, 0, spilled_max_dshred_idx );
+      fd_stem_publish( stem, ctx->repair_out_idx, sig_, ctx->repair_out_chunk, 0, 0, ctx->tsorig, ctx->tsorig );
+      ctx->repair_out_chunk = fd_dcache_compact_next( ctx->repair_out_chunk, 0, ctx->repair_out_chunk0, ctx->repair_out_wmark );
```
ptaffet-jump: You don't need to advance the dcache if you didn't use it, right?

src/disco/shred/fd_shred_tile.c
```diff
@@ -914,6 +917,15 @@ after_frag( fd_shred_ctx_t *    ctx,
       }
     }
 
+    if( FD_UNLIKELY( spilled_slot < ULONG_MAX && spilled_fec_set_idx < FD_SHRED_BLK_MAX && spilled_max_dshred_idx < FD_SHRED_BLK_MAX ) ) {
+      /* We've spilled an in-progress FEC set in the fec_resolver. We need to let
+         repair know to clear out it's cached info for that fec set and
+         re-repair those shreds. */
+      ulong sig_ = fd_disco_shred_repair_shred_sig( 0, spilled_slot, spilled_fec_set_idx, 0, spilled_max_dshred_idx );
+      fd_stem_publish( stem, ctx->repair_out_idx, sig_, ctx->repair_out_chunk, 0, 0, ctx->tsorig, ctx->tsorig );
```
ptaffet-jump, line 923: Does this affect the burst for the tile or this link?

src/util/tmpl/fd_set.c
```diff
@@ -276,6 +276,16 @@ SET_(first)( SET_(t) const * set ) {
   return ~0UL;
 }
 
+FD_FN_PURE static inline ulong
```
ptaffet-jump, line 283: This will need documentation

src/util/tmpl/fd_map_giant.c
```diff
@@ -99,7 +99,7 @@
 
     // mymap_insert inserts the key pointed to by key into the map.
     // Returns the location in the caller's address space of the element
-    // for key in the map or NULL if there was no space in the map.
```
ptaffet-jump, line 102: This is just a drive-by fix?

src/disco/shred/fd_fec_resolver.c
ptaffet-jump: Now I see Charlie's comment. Yeah, `spilled` would be fine too

src/disco/shred/fd_shred_tile.c
emwang-jump: nope! both can remain the same

src/disco/shred/fd_fec_resolver.h
emwang-jump: yep, put this in the updated docs as well

src/disco/shred/fd_shred_tile.c
```diff
@@ -873,12 +873,10 @@ after_frag( fd_shred_ctx_t *    ctx,
 
     fd_fec_set_t const * out_fec_set[1];
     fd_shred_t const   * out_shred[1];
-    ulong spilled_slot           = ULONG_MAX;
-    uint  spilled_fec_set_idx    = FD_SHRED_BLK_MAX;
-    uint  spilled_max_dshred_idx = FD_SHRED_BLK_MAX;
+    fd_fec_resolver_spilled_t spilled_fec = { 0 };
```
ptaffet-jump, line 876: Hmm I suppose it is possible to spill a FEC set with slot==0, fec_set_idx==0, max_dshred_idx==0, right? It would require someone sending you some specially crafted shreds. Would the data structures get out of sync, or just the stats?

src/disco/shred/fd_shred_tile.c
```diff
@@ -917,13 +915,21 @@ after_frag( fd_shred_ctx_t *    ctx,
...
+      /* We've spilled an in-progress FEC set in the fec_resolver. We
+         need to let repair know to clear out it's cached info for that
+         fec set and re-repair those shreds.
+
+         Note that although there are three stem_publishes to repair
+         tile, it's impossible for all three to be sent in the same
+         after_frag, which is why shred->repair link burst is 2.  The
+         resolver only gets spilled when a new FEC set ctx is created,
+         and it's not possible for a ctx to be created and for a FEC to
+         complete all in the same fec_resolver_add_shred call.  Thus we
```
ptaffet-jump: This is only true in the fixed-32 world. Otherwise you can receive a coding shred for a new FEC set with only one data shred.

src/disco/shred/fd_shred_tile.c
emwang-jump: its a valid spill if `slot==real_slot, fec_set_idx==0 and max_dshred_idx==0`. In the case where slot==0 though, the repair data structures filter out any message where the slot is less than the root slot. currently also in shred_tile it doesn't publish the evict msg to repair if the slot == 0.

src/disco/shred/fd_shred_tile.c
emwang-jump: ah you're right 😧 ill bump it to 3 for now

src/disco/shred/fd_fec_resolver.h
lidatong: do we still need the `out_spilled_slot` and `out_spilled_idx` if we only progress on fec complete?

lidatong (approved):

PR #6135, opened 2025-08-23
runtime: simplify lthash logic

ripatel-fd:
- Remove redundant 'slot_lthash' type
- Simplify 2048->32 lthash reduction (just use prefix instead of hashing)
- Reduce alignment of lthash to 64 bytes

### DIFF
--- a/contrib/test/run_solcap_tests.sh
+++ b/contrib/test/run_solcap_tests.sh
@@ -1,7 +1,7 @@
 DUMP_DIR=${DUMP_DIR:="./dump"}
 echo $OBJDIR
 
-LEDGER="devnet-398736132"
+LEDGER="devnet-398736132-solcap"
 
 if [[ ! -e $DUMP_DIR/$LEDGER && SKIP_INGEST -eq 0 ]]; then
     echo "Downloading gs://firedancer-ci-resources/$LEDGER.tar.gz"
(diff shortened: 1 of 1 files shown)


ripatel-fd:
@kbhargava-jump Could you regenerate the solcap fixtures using this patch? This commit changes the way lthash values are converted to 32-byte hashes in solcap. Previously, we'd do `blake3(lthash([ accounts ... ]))`, now we do `lthash([ accounts ... ])[ 0..32 ]`

kbhargava-jump:
> @kbhargava-jump Could you regenerate the solcap fixtures using this patch? This commit changes the way lthash values are converted to 32-byte hashes in solcap. Previously, we'd do `blake3(lthash([ accounts ... ]))`, now we do `lthash([ accounts ... ])[ 0..32 ]`

yup, will do this week. please remind me if I forget

kbhargava-jump:
> > @kbhargava-jump Could you regenerate the solcap fixtures using this patch? This commit changes the way lthash values are converted to 32-byte hashes in solcap. Previously, we'd do `blake3(lthash([ accounts ... ]))`, now we do `lthash([ accounts ... ])[ 0..32 ]`
> 
> yup, will do this week. please remind me if I forget

done

cali-jumptrading (approved):

PR #6140, opened 2025-08-23
log: skip shm locking when writing to log file

ripatel-fd:
`write(2)` says:

   If the file was [open(2)](https://man7.org/linux/man-pages/man2/open.2.html)ed with O_APPEND, the
   file offset is first set to the end of the file before writing.
   The adjustment of the file offset and the write operation are
   performed as an atomic step.

The log file is always opened with O_APPEND, therefore the userland mutex when logging to file can be skipped.

This only works assuming `write(2)` does not do short writes.


ripatel-fd:
Marking as draft. Will merge once we have short write recovery.

PR #6143, opened 2025-08-25
gossip: fix under allocation of bloom filter

two-heart:


src/flamenco/gossip/fd_bloom.c
```diff
@@ -66,7 +66,7 @@ fd_bloom_new( void *     shmem,
 
   FD_SCRATCH_ALLOC_INIT( l, shmem );
   fd_bloom_t * bloom = FD_SCRATCH_ALLOC_APPEND( l, FD_BLOOM_ALIGN, sizeof(fd_bloom_t) );
-  void * _keys       = FD_SCRATCH_ALLOC_APPEND( l, 8UL, num_keys );
+  void * _keys       = FD_SCRATCH_ALLOC_APPEND( l, 8UL, num_keys*sizeof(ulong) );
```
ripatel-fd, line 69: I'm confused, why didn't this crash the unit test?

ravyu-jump: no idea lol...

ripatel-fd: could we add regression tests for Liam's fixes?

ripatel-fd: For the parser, probably easiest to check in fuzz vectors in corpus/fuzz_gossip_msg_parse

ravyu-jump: > could we add regression tests for Liam's fixes?

What would they look like? Just alignment and footprint checks?

ripatel-fd: @ravyu-jump Insert a bunch of stuff in a way that covers most of the internal sub-objects. Then run the test with an ASan-protected allocator (libc aligned_alloc is easiest probably). ASan should easily be able to detect stray OOB accesses

PR #6144, opened 2025-08-25
runtime: fix vote_states_iter

ripatel-fd:
- **Reuse map_chain_iter struct**
- **runtime: fix vote_states_iter**


ibhatt-jumptrading (approved):
thanks for doing this

PR #6152, opened 2025-08-25
pack, poh: fix pack ownership of slot bank

jherrera-jump:


src/discoh/poh/fd_poh_tile.c
```diff
@@ -446,6 +446,11 @@ typedef struct {
      that we know which slot the slot_done message is arriving for. */
   int slot_done;
 
+  /* Pack / Bank might still own the previous leader bank in draining
+     mode.  We keep a copy of a reference here so that we can properly
+     increment / decrement its Arc. */
+  void const * draining_leader_bank;
```
ptaffet-jump:
```suggestion
  /* Pack and bank tiles need a reference to the bank object with a slightly different lifetime than current_leader_bank, particularly when we switch forks in the middle of a leader slot.  We need to make sure we don't free the last reference to the bank while the pack or bank tiles are still using it.  The strange thing is that bank tiles have no concept of the current slot, but we know they're done with the bank object when pack's inter-slot bank draining process is complete.  Pack notifies PoH by a frag with sig==ULONG_MAX on the pack_poh link when the banks are drained, and the PoH tile must then free the reference on behalf of pack.
    pack_leader_bank is non-NULL when the reference we're holding on behalf of the pack tile is acquired, and NULL when it is not acquired. */
  void const * pack_leader_bank;
```

ptaffet-jump (approved):
Code looks great! Can we clarify the comment to make it easier for the next time we need to remember how this works?

PR #6157, opened 2025-08-25
bundle: ignore bundles >5 txns

jherrera-jump:


ripatel-fd:
@jherrera-jump What motivated this change?

PR #6158, opened 2025-08-25
ballet: fix elf section header addr check

cali-jumptrading:


src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -424,6 +429,8 @@ fd_sbpf_load_shdrs( fd_sbpf_elf_info_t *  info,
       /* Coherence check sum of section sizes */
       REQUIRE( tot_section_sz + sh_actual_size >= tot_section_sz ); /* overflow check */
       tot_section_sz += sh_actual_size;
+
+      REQUIRE( highest_addr - lowest_addr <= elf_sz ); /* addr out of bounds */
```
mjain-jump: is underflow already handled?

cali-jumptrading: I don't think the address bounds were being checked.

0x0ece (dismissed):

PR #6162, opened 2025-08-25
gossip: unit tests for bloom key under-alloc

ravyu-jump:
for #6143

### DIFF
--- a/src/flamenco/gossip/test_bloom.c
+++ b/src/flamenco/gossip/test_bloom.c
@@ -2,6 +2,7 @@
 #include "fd_bloom.h"
 
 #include <stdlib.h>
+#include <string.h>
 
 FD_STATIC_ASSERT( FD_BLOOM_ALIGN    ==64UL,  unit_test );
 FD_STATIC_ASSERT( FD_BLOOM_FOOTPRINT==128UL, unit_test );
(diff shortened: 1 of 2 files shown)


src/flamenco/gossip/test_bloom.c
```diff
@@ -59,6 +60,30 @@ test_add_contains( void ) {
...
 }
 
+/* Tests that the keys footprint is correctly sized by checking for
+   overwrites into the filter bits during initialization. Filter
+   bits OOB will be caught by UBSAN. */
+void
+test_keys_oob( void ) {
+  void * bytes = aligned_alloc( fd_bloom_align(), fd_bloom_footprint( 0.1, 16 ) );
+  FD_TEST( bytes );
+  memset( bytes, 0UL, fd_bloom_footprint( 0.1, 16 ) );
```
ripatel-fd: This memset shouldn't be required, since bloom_new should take in uninitialized memory fine

ravyu-jump: im specifically testing for out of bounds `bloom->keys[i]` acces by observing if the keys generated in `fd_bloom_initialize` overwrite the memory below it which in this case is the bloom bits region. You'd need to zero alloc to reliably test that.

We can't rely on ASAN/UBSAN here since proper keys accesses can't exceed the region of memory allocated for the bloom object (based on how their lengths are computed), they can only spill into the region directly below it. 

This test is a little fragile in that re-arranging the memory layout of the bloom object can break it. Open to suggestions

PR #6195, opened 2025-08-27
fuzz, flamenco: enable vm tracing

cali-jumptrading:


src/flamenco/runtime/tests/fd_sol_compat.c
```diff
@@ -75,12 +76,16 @@ sol_compat_init( int log_level ) {
   }
 
   ulong footprint = 6UL<<30;
+  if( FD_UNLIKELY( enable_vm_tracing ) ) {
+    footprint = 100UL<<30;
```
mjain-jump: 100gb??

src/flamenco/runtime/tests/fd_sol_compat.c
```diff
@@ -75,12 +76,16 @@ sol_compat_init( int log_level ) {
...
+  }
+
   ulong part_max  = fd_wksp_part_max_est( footprint, 64UL<<10 );
   ulong data_max  = fd_wksp_data_max_est( footprint, part_max );
   wksp = fd_wksp_demand_paged_new( "sol_compat", 42U, part_max, data_max );
   if( FD_UNLIKELY( !wksp ) ) FD_LOG_ERR(( "fd_wksp_demand_paged_new() failed" ));
 
-  runner = sol_compat_setup_runner();
+  runner = sol_compat_setup_runner( enable_vm_tracing);
```
mjain-jump:
```suggestion
  runner = sol_compat_setup_runner( enable_vm_tracing );
```

src/flamenco/runtime/tests/fd_sol_compat.c
cali-jumptrading: um yeah I guess I could try a lower number... lol

cali-jumptrading: 16GB seems to work.

src/flamenco/runtime/program/fd_bpf_loader_program.c
```diff
@@ -487,19 +487,13 @@ fd_bpf_execute( fd_exec_instr_ctx_t *            instr_ctx,
     return FD_EXECUTOR_INSTR_ERR_PROGRAM_ENVIRONMENT_SETUP_FAILURE;
   }
 
-#ifdef FD_DEBUG_SBPF_TRACES
```
ibhatt-jumptrading, line 490: we might need this in backtest too so not sure if we want to make it a fuzz_config specific thing in txn_ctx. Maybe we put this in the capture_ctx as that's a better catch-all

src/flamenco/runtime/tests/fd_txn_harness.c
```diff
@@ -423,6 +423,7 @@ fd_runtime_fuzz_txn_ctx_exec( fd_solfuzz_runner_t * runner,
   fd_exec_txn_ctx_t * txn_ctx     = fd_exec_txn_ctx_join( fd_exec_txn_ctx_new( txn_ctx_mem ), runner->spad, fd_wksp_containing( runner->spad ) );
   *txn_ctx->funk                  = *slot_ctx->funk;
```
ibhatt-jumptrading: nit: align

src/flamenco/runtime/tests/fd_sol_compat.c
```diff
@@ -60,7 +60,8 @@ sol_compat_cleanup_runner( fd_solfuzz_runner_t * runner ) {
 }
 
 void
-sol_compat_init( int log_level ) {
+sol_compat_init( int log_level,
+                 int enable_vm_tracing ) {
```
ripatel-fd: Don't break stable ABI. This should probably use env vars (sorry) to stay backwards compatible. If you insist on breaking ABI here, make a `sol_compat_init_v2` and give it an extendable option struct. e.g. like this:

```
struct options {
  ulong struct_size; /* versioning for the options struct itself */
  /* options go here */
  int enable_vm_tracing; 
};
```

If we just add more function params, it's going to crash in prod

src/flamenco/runtime/tests/fd_solfuzz.h
```diff
@@ -29,6 +29,10 @@ struct fd_solfuzz_runner {
 
   fd_solcap_writer_t * solcap;
   void *               solcap_file; /* FILE * */
+
+  struct {
+    int enable_vm_tracing;
+  } fuzz_options;
```
ripatel-fd: this anon struct seems unnecessary

src/flamenco/runtime/tests/fd_sol_compat.c
cali-jumptrading: oh... um how is this ABI currently used? just in solfuzz right? So if I change solfuzz it should be ok? I actually like the extendable option struct, so I'd probably go with that.

src/flamenco/runtime/tests/fd_solfuzz.h
cali-jumptrading: removed

src/flamenco/runtime/tests/fd_sol_compat.c
cali-jumptrading: Using environment variables now.

src/flamenco/runtime/tests/fd_sol_compat.c
```diff
@@ -67,20 +68,27 @@ sol_compat_init( int log_level ) {
...
+
   fd_log_enable_unclean_exit();
   fd_boot( &argc, &argv_ );
 
   if( FD_UNLIKELY( wksp || runner ) ) {
     FD_LOG_ERR(( "sol_compat_init() called multiple times" ));
   }
 
-  ulong footprint = 6UL<<30;
+  ulong footprint = 7UL<<30;
```
ripatel-fd, line 85: If you change this, you also have to change the test runner script in contrib/test/run_test_vectors.sh

src/flamenco/runtime/tests/fd_solfuzz.h
```diff
@@ -29,10 +29,20 @@ struct fd_solfuzz_runner {
...
+
+  int enable_vm_tracing;
 };
 
 typedef struct fd_solfuzz_runner fd_solfuzz_runner_t;
 
+/* A fd_solfuzz_runner_t object is configured with options stored in a
+   fd_solfuzz_runner_options_t object. */
+struct fd_solfuzz_runner_options {
+   int enable_vm_tracing;
```
ripatel-fd:
```suggestion
  int enable_vm_tracing;
```

src/flamenco/runtime/tests/fd_solfuzz.h
```diff
@@ -45,8 +55,9 @@ FD_PROTOTYPES_BEGIN
    demand paged memory. */
 
 fd_solfuzz_runner_t *
-fd_solfuzz_runner_new( fd_wksp_t * wksp,
-                       ulong       wksp_tag );
+fd_solfuzz_runner_new( fd_wksp_t *                 wksp,
+                       ulong                       wksp_tag,
+                       fd_solfuzz_runner_options_t options );
```
ripatel-fd: Option structs are typically const pointers. Otherwise the stack layout changes when more options are added. Doesn't matter when doing internal linkage, but it's just common practice

src/flamenco/runtime/tests/fd_vm_harness.c
```diff
@@ -217,14 +217,12 @@ do{
   }
 
   /* Setup trace */
-  const uint DUMP_TRACE = 0; // Set to 1 to dump trace to stdout
-  uint tracing_enabled = input->vm_ctx.tracing_enabled;
-  fd_vm_trace_t * trace = NULL;
-  ulong event_max = 1UL<<20;
-  ulong event_data_max = 2048UL;
+  const int enable_vm_tracing = runner->enable_vm_tracing;
+  uint tracing_enabled        = input->vm_ctx.tracing_enabled;
```
ripatel-fd: What's the difference between these two?

cali-jumptrading: don't know... I'll ask Manik

src/flamenco/runtime/tests/fd_sol_compat.c
cali-jumptrading: done

src/flamenco/runtime/tests/fd_vm_harness.c
cali-jumptrading: for now I've just removed usages of it.

mjain-jump: i think one is for capturing traces in the actual effects, and the other is just for debugging purposes

mjain-jump: i dont think we support diffing traces in the effects because theyre known to be different between agave's JIT and our interpreter

ibhatt-jumptrading: what are the differences?

src/flamenco/runtime/program/fd_bpf_loader_program.c
cali-jumptrading: We would then have to modify all the harnesses to set up a capture_ctx. Do you think it is a worthwhile investment at this point?

PR #6196, opened 2025-08-27
Revert "optionally leave timed out conns allocated"

akhinvasara-jumptrading:
This PR reverts the entire feature to leave timed out conns allocated, including the original commit and the three patches that have already followed it. Not worth the complexity / increasing the bug surface. 

The second commit preserves a few minor improvements that got dropped by the revert.


src/waltz/quic/fd_quic.c
```diff
@@ -4267,6 +4198,7 @@ fd_quic_conn_free( fd_quic_t *      quic,
   conn->svc_prev        = UINT_MAX;
   conn->svc_next        = state->free_conn_list;
   state->free_conn_list = conn->conn_idx;
+  fd_quic_set_conn_state( conn, FD_QUIC_CONN_STATE_INVALID );
```
ripatel-fd, line 4201: dup set state

PR #6201, opened 2025-08-28
pack: update skip_cnt, skip logic

cavemanloverboy:
The current `skip_cnt` causes the pack tile to skip attempting to schedule a transaction `bank_cnt/2` number of times after every scheduling attempt. in the case of high bank tile count, this can lead to a larger-than-expected delay before a transaction gets scheduled, especially since typically not many banks are actually used on average. Here we fix this issue by setting `skip_cnt` to 1 after every successful txn scheduled, regardless of bank tile count.

This has the side effect of more frequent polling, which can cause a transaction waiting for a rebate to hit `FD_PACK_SKIP_CNT` and not be considered for scheduling, so we simultaneously increase `FD_PACK_SKIP_CNT` from 5 to 50.


ptaffet-jump (dismissed):

cavemanloverboy:
test was failing because `pack_bank` link is checked after every call to `after_credit`. previously, `skip_cnt` was being set to `bank_cnt/2=1/2=0`, so it was never skipped. now, it's being set to `skip_cnt=bank_cnt+1=2` so it's actually being skipped.

src/disco/pack/test_pack_tile.c
```diff
@@ -477,6 +477,11 @@ bank_out_check( fd_tile_test_ctx_t  * test_ctx,
   if( ctx->leader_slot==ULONG_MAX ) {
     return 0;
   }
+  /* after_credit is skipped when skip_cnt>0 */
+  int maybe_scheduled_txn = ctx->skip_cnt<=0;
+  if( !maybe_scheduled_txn ) {
+    return 0;
+  }
```
ptaffet-jump, line 484: This is not great, since it prevents actually testing the result of the schedule, but right now the tile unit test framework doesn't allow for anything better

cavemanloverboy: will make separate issue

PR #6202, opened 2025-08-28
runtime: get genesis working [wip, do not review]

ibhatt-jumptrading:


jumpsiegel (approved):

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -376,73 +387,75 @@ backtest_cmd_fn( args_t *   args FD_PARAM_UNUSED,
...
-    long cur = fd_log_wallclock();
-    if( FD_UNLIKELY( cur<next ) ) {
-      long sleep_nanos = fd_long_min( 1000L*1000L, next-cur );
-      FD_TEST( !fd_sys_util_nanosleep(  (uint)(sleep_nanos/(1000L*1000L*1000L)), (uint)(sleep_nanos%(1000L*1000L*1000L)) ) );
-      continue;
+      ulong total_off    = snaprd_metrics[ MIDX( GAUGE, SNAPRD, FULL_BYTES_READ ) ] +
+                          snaprd_metrics[ MIDX( GAUGE, SNAPRD, INCREMENTAL_BYTES_READ ) ];
+      ulong snaprd_backp = snaprd_metrics[ MIDX( COUNTER, TILE, REGIME_DURATION_NANOS_BACKPRESSURE_PREFRAG ) ];
+      ulong snaprd_wait  = snaprd_metrics[ MIDX( COUNTER, TILE, REGIME_DURATION_NANOS_CAUGHT_UP_PREFRAG    ) ] +
+                          snaprd_metrics[ MIDX( COUNTER, TILE, REGIME_DURATION_NANOS_CAUGHT_UP_POSTFRAG   ) ] + snaprd_backp;
```
ripatel-fd: sorry for the nit but can you keep the original vertical alignment here

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -376,73 +387,75 @@ backtest_cmd_fn( args_t *   args FD_PARAM_UNUSED,
   fd_topo_run_single_process( &config->topo, 2, config->uid, config->gid, fdctl_tile_run );
 
   fd_topo_t * topo = &config->topo;
+  int genesis_boot = config->tiles.replay.genesis[0] != '\0';
+  if( FD_LIKELY( !genesis_boot ) ) {
+    fd_topo_tile_t * snaprd_tile = &topo->tiles[ fd_topo_find_tile( topo, "snaprd", 0UL ) ];
```
ripatel-fd, line 392: i feel like we should move this monitoring code into its own function. This !genesis_boot condition seems a bit random. There will be other backtest scenarios in the future where a snapshot is not needed, for example when persisted accounts database and bank state in workspaces is available. And we shouldn't move this whole block of code again when we introduce these conditions

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -38,6 +38,8 @@ backtest_topo( config_t * config ) {
   ulong exec_tile_cnt   = config->firedancer.layout.exec_tile_count;
   ulong writer_tile_cnt = config->firedancer.layout.writer_tile_count;
 
+  int genesis_boot = config->tiles.replay.genesis[0] != '\0';
```
ripatel-fd: Great that you're making snapshot loading optional. Frustrating that it wasn't optional from the beginning. I'd call this `int enable_snap_loader` or something in case we have other reasons unrelated to genesis why you'd want to omit the snapshot loader

src/app/firedancer/topology.c
```diff
@@ -370,11 +374,13 @@ fd_topo_initialize( config_t * config ) {
   /**/                 fd_topob_link( topo, "send_txns",    "send_txns",    128UL,                                    FD_TPU_RAW_MTU,                1UL   );
   /**/                 fd_topob_link( topo, "send_sign",    "send_sign",    128UL,                                    FD_TXN_MTU,                    1UL   );
   /**/                 fd_topob_link( topo, "sign_send",    "sign_send",    128UL,                                    sizeof(fd_ed25519_sig_t),      1UL   );
+  if( FD_LIKELY(!is_genesis ) ) {
```
ripatel-fd:
```suggestion
  if( FD_LIKELY( !is_genesis ) ) {
```

src/app/firedancer/topology.c
```diff
@@ -437,12 +443,18 @@ fd_topo_initialize( config_t * config ) {
...
-  snaprd_tile->allow_shutdown = 1;
-  fd_topo_tile_t * snapdc_tile = fd_topob_tile( topo, "snapdc", "snapdc", "metric_in", tile_to_cpu[ topo->tile_cnt ], 0, 0 );
-  snapdc_tile->allow_shutdown = 1;
-  fd_topo_tile_t * snapin_tile = fd_topob_tile( topo, "snapin", "snapin", "metric_in", tile_to_cpu[ topo->tile_cnt ], 0, 0 );
-  snapin_tile->allow_shutdown = 1;
+  fd_topo_tile_t * snaprd_tile = NULL;
+  fd_topo_tile_t * snapdc_tile = NULL;
+  fd_topo_tile_t * snapin_tile = NULL;
+
+  if( FD_LIKELY(!is_genesis ) ) {
```
ripatel-fd:
```suggestion
  if( FD_LIKELY( !is_genesis ) ) {
```

src/flamenco/runtime/fd_runtime.c
```diff
@@ -2404,117 +2457,145 @@ fd_runtime_process_genesis_block( fd_exec_slot_ctx_t * slot_ctx,
...
   }
 
+  /* If we have a genesis file, we will need to open the file and read
+     the data into a buffer. */
+
   struct stat sbuf;
-  if( FD_UNLIKELY( stat( genesis_filepath, &sbuf) < 0 ) ) {
-    FD_LOG_ERR(( "cannot open %s : %s", genesis_filepath, strerror(errno) ));
+  if( FD_UNLIKELY( stat( genesis_filepath, &sbuf)<0 ) ) {
+    FD_LOG_ERR(( "cannot open %s : %s", genesis_filepath, strerror( errno ) ));
```
ripatel-fd:
```suggestion
    FD_LOG_ERR(( "cannot open %s: %i-%s", genesis_filepath, errno, fd_io_strerror( errno ) ));
```

src/flamenco/runtime/fd_runtime.c
```diff
@@ -2404,117 +2457,145 @@ fd_runtime_process_genesis_block( fd_exec_slot_ctx_t * slot_ctx,
...
   }
   int fd = open( genesis_filepath, O_RDONLY );
-  if( FD_UNLIKELY( fd < 0 ) ) {
-    FD_LOG_ERR(("cannot open %s : %s", genesis_filepath, strerror(errno)));
+  if( FD_UNLIKELY( fd<0 ) ) {
+    FD_LOG_ERR(("cannot open %s : %s", genesis_filepath, strerror( errno ) ));
```
ripatel-fd, line 2538:
```suggestion
    FD_LOG_ERR(("cannot open %s: %i-%s", genesis_filepath, errno, fd_io_strerror( errno ) ));
```

src/flamenco/runtime/fd_runtime.c
```diff
@@ -2404,117 +2457,145 @@ fd_runtime_process_genesis_block( fd_exec_slot_ctx_t * slot_ctx,
...
+  /* If we have a genesis file, we will need to open the file and read
+     the data into a buffer. */
+
   struct stat sbuf;
-  if( FD_UNLIKELY( stat( genesis_filepath, &sbuf) < 0 ) ) {
-    FD_LOG_ERR(( "cannot open %s : %s", genesis_filepath, strerror(errno) ));
+  if( FD_UNLIKELY( stat( genesis_filepath, &sbuf)<0 ) ) {
+    FD_LOG_ERR(( "cannot open %s : %s", genesis_filepath, strerror( errno ) ));
   }
   int fd = open( genesis_filepath, O_RDONLY );
```
ripatel-fd, line 2536: This wouldn't work in a production validator because we ban `open`. Which is going to be problematic for `firedancer-dev dev`. I think one way this could work is that we open the genesis file descriptor in `privileged_init`, then do the actual read in the replay tile. Not going to block the PR over this though, can fix in a follow up PR.

src/flamenco/runtime/fd_runtime.c
```diff
@@ -2404,117 +2457,145 @@ fd_runtime_process_genesis_block( fd_exec_slot_ctx_t * slot_ctx,
...
+     copied out to an indepedent buffer. Decode the genesis block from
+     the config data. */
 
-    for( ulong i=0; i<genesis_block->accounts_len; i++ ) {
-      fd_pubkey_account_pair_t * a = &genesis_block->accounts[i];
+  int err;
+  fd_genesis_solana_t * genesis_block = fd_bincode_decode_spad(
+      genesis_solana, runtime_spad, buf, sz, &err );
+  if( FD_UNLIKELY( err!=FD_BINCODE_SUCCESS ) ) {
+    FD_LOG_ERR(( "fd_genesis_solana_decode_footprint failed (%d)", err ));
```
ripatel-fd: Not operator friendly, i would just say something like: "Failed to decode genesis manifest"

src/funk/fd_funk_base.h
```diff
@@ -300,7 +300,7 @@ fd_funk_txn_xid_eq_root( fd_funk_txn_xid_t const * x ) {
 static inline fd_funk_txn_xid_t *
 fd_funk_txn_xid_set_root( fd_funk_txn_xid_t * x ) {
   ulong * a = x->ul;
-  a[0] = 0UL; a[1] = 0UL;
+  a[0] = ULONG_MAX; a[1] = ULONG_MAX;
   return x;
 }
```
ripatel-fd, line 305: Why this drive-by change of the funk root XID? I have a feeling this is going to break stuff

src/app/firedancer-dev/commands/backtest.c
jumpsiegel: I feel a bit stronger in that I think we need to merge the backtest and live topo setup, making it properly configurable, because right now when you fix one you always have to fix the other

src/app/firedancer/topology.c
jumpsiegel: done

src/funk/fd_funk_base.h
jumpsiegel: This isn't a drive by but was a major bug fix.  slot 0 was previously the root sentinel.   I needed slot 0 to be a legit slot

jumpsiegel (dismissed):

src/flamenco/runtime/fd_runtime.c
jumpsiegel: Yeah, lets hit this up in a follow up PR..

src/flamenco/runtime/fd_runtime.c
jumpsiegel: for our grand sandbox PR.. yeah, I can see this would be a good change

jumpsiegel (approved):

jumpsiegel (approved):

kbhargava-jump (approved):

PR #6210, opened 2025-08-28
Another batch of fixes for rpc service

asiegel-jt:


src/discof/replay/fd_replay_notif.h
```diff
@@ -14,12 +14,14 @@ struct __attribute__((aligned(64UL))) fd_replay_notif_msg {
       ulong parent;
       ulong root;
       ulong slot;
+      ulong slot_idx;
```
cali-jumptrading: don't really like the name of `slot_idx`. It's not clear what this field is used for. Could we name it something like `epoch_slot_offset`?

jumpsiegel (approved):

PR #6212, opened 2025-08-28
Fix fdctl netconf

ripatel-fd:


jherrera-jump (dismissed):
Leaving this here for posterity

@jherrera-jump 
> What's the motivation for trying to optimize a print? Do we print stuff in a non-development context? I feel like it makes more sense to avoid using private methods unless it will fix a broken test or make the program faster at runtime.

@ripatel-fd 
>Good point. This is not an optimization, but rather a bug fix. fdctl netconf is a debugging command that prints out some internal tile state as human-readable data. This command maps workspaces read-only (to avoid breaking a prod node). However, it calls a function that writes and takes a lock.

jherrera-jump (approved):

PR #6219, opened 2025-08-29
chore: remove redundant word in comment

zhoujiaweii:
remove redundant word in comment


ripatel-fd:
This was deliberately written that way

PR #6231, opened 2025-09-02
quic: simplify time handling

akhinvasara-jumptrading:
This PR lets fd_quic forget about ticks and time domains, to avoid any further time domain conversion bugs. fd_quic now only speaks nanoseconds, and its callers are responsible for passing 'now' in ns (typically using fd_clock)

Primarily written by @ripatel-fd .

### DIFF
--- a/src/app/shared_dev/commands/bench/fd_benchs.c
+++ b/src/app/shared_dev/commands/bench/fd_benchs.c
@@ -30,15 +30,6 @@ quic_tx_aio_send( void *                    _ctx,
                   ulong *                   opt_batch_idx,
                   int                       flush );
 
-/* quic_now is called by the QUIC engine to get the current timestamp in
-   UNIX time.  */
-
-static ulong
-quic_now( void * ctx ) {
-  (void)ctx;
-  return (ulong)fd_log_wallclock();
-}
-
 typedef struct {
   ulong round_robin_cnt;
   ulong round_robin_id;
@@ -58,7 +49,11 @@ typedef struct {
   uint             service_ratio_idx;
   fd_aio_t         tx_aio;
 
-  // vector receive members
+  long         now;        /* current time in ns    */
+  fd_clock_t   clock[1];   /* memory for fd_clock_t */
+  long         recal_next; /* next recalibration time (ns) */
+
+  /* vector receive members */
   struct mmsghdr rx_msgs[IO_VEC_CNT];
   struct mmsghdr tx_msgs[IO_VEC_CNT];
   struct iovec   rx_iovecs[IO_VEC_CNT];
@@ -69,10 +64,13 @@ typedef struct {
   ulong tx_idx;
 
   fd_wksp_t * mem;
+
+  uchar __attribute__((aligned(FD_CLOCK_ALIGN))) clock_mem[ FD_CLOCK_FOOTPRINT ];
 } fd_benchs_ctx_t;
 
 static void
-service_quic( fd_benchs_ctx_t * ctx ) {
+service_quic( fd_benchs_ctx_t * ctx,
+              long              now ) {
 
   if( !ctx->no_quic ) {
     /* Publishes to mcache via callbacks */
@@ -119,7 +117,7 @@ service_quic( fd_benchs_ctx_t * ctx ) {
           buf[2] = (uchar)( ip_len >> 8 );
           buf[3] = (uchar)( ip_len      );
 
-          fd_quic_process_packet( ctx->quic, buf, ip_len );
+          fd_quic_process_packet( ctx->quic, buf, ip_len, now );
         }
       } else if( FD_UNLIKELY( revents & POLLERR ) ) {
         int error = 0;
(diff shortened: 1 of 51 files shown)


src/waltz/quic/fd_quic.c
```diff
@@ -4432,21 +4336,21 @@ fd_quic_conn_create( fd_quic_t *               quic,
...
+  conn->peer_ack_delay_scale     = (float)( 1UL << peer_ack_delay_exponent ) * 10e3f;
   conn->peer_max_ack_delay_ticks = 0.0f;       /* starts at zero, since peers respond immediately to */
                                                /* INITIAL and HANDSHAKE */
                                                /* updated when we get transport parameters */
-  rtt->smoothed_rtt              = FD_QUIC_INITIAL_RTT_US * (float)quic->config.tick_per_us;
-  rtt->latest_rtt                = FD_QUIC_INITIAL_RTT_US * (float)quic->config.tick_per_us;
-  rtt->min_rtt                   = FD_QUIC_INITIAL_RTT_US * (float)quic->config.tick_per_us;
-  rtt->var_rtt                   = FD_QUIC_INITIAL_RTT_US * (float)quic->config.tick_per_us * 0.5f;
-  conn->rtt_period_ticks         = FD_QUIC_RTT_PERIOD_US  * (float)quic->config.tick_per_us;
+  rtt->smoothed_rtt              = FD_QUIC_INITIAL_RTT_US * 10e3f;
```
nbridge-jump: Why are these `10e3f`?
Aren't the `rtt` members in `ns`?

src/waltz/quic/fd_quic.c
```diff
@@ -4432,21 +4336,21 @@ fd_quic_conn_create( fd_quic_t *               quic,
   fd_rtt_estimate_t * rtt = conn->rtt;
 
   ulong peer_ack_delay_exponent  = 3UL; /* by spec, default is 3 */
-  conn->peer_ack_delay_scale     = (float)( 1UL << peer_ack_delay_exponent )
-                                         * (float)quic->config.tick_per_us;
+  conn->peer_ack_delay_scale     = (float)( 1UL << peer_ack_delay_exponent ) * 10e3f;
```
nbridge-jump: Why are these `10e3f`?

src/waltz/quic/fd_quic.c
akhinvasara-jumptrading: yes good catch, thank you
should be 1e3f everywhere

src/app/shared_dev/commands/bench/fd_benchs.c
```diff
@@ -511,13 +520,22 @@ quic_tx_aio_send( void *                    _ctx,
   return 0;
 }
 
+static void
+during_housekeeping( fd_benchs_ctx_t * ctx ) {
+  if( FD_UNLIKELY( ctx->recal_next <= ctx->now ) ) {
+    ctx->recal_next = fd_clock_recal( ctx->clock, ctx->now, ctx->now );
```
nbridge-jump: recal needs the value from both the x clock and the y clock. This is using the same value for both.

src/disco/quic/fd_quic_tile.c
```diff
@@ -449,8 +444,13 @@ quic_tls_keylog( void *       _ctx,
 
 static void
 during_housekeeping( fd_quic_ctx_t * ctx ) {
+  if( FD_UNLIKELY( ctx->recal_next <= ctx->now ) ) {
+    ctx->recal_next = fd_clock_recal( ctx->clock, ctx->now, ctx->now );
```
nbridge-jump: Same here: fd_clock_recal needs values from both clocks

src/discof/send/fd_send_tile.c
```diff
@@ -191,8 +183,15 @@ quic_tx_aio_send( void *                    _ctx,
 
 static void
 during_housekeeping( fd_send_tile_ctx_t * ctx ) {
+  ctx->housekeeping_ctr++;
+
+  if( FD_UNLIKELY( ctx->recal_next <= ctx->now ) ) {
+    ctx->recal_next = fd_clock_recal( ctx->clock, ctx->now, ctx->now );
```
nbridge-jump: Same here

two-heart (changes requested):
I think this is a good change to make, reducing the chance of bugs.

With the change of next_timeout to `ulong` we should update `fd_quic_svc_timers_init_conn` for consistency, not a functional bug though.
```c
  conn->svc_meta.next_timeout = ULONG_MAX;
```
to
```c
  conn->svc_meta.next_timeout = LONG_MAX;
```
And if we want to be nitpicky, also the integer suffixes in the `test_svc_cancel` test.

akhinvasara-jumptrading:
> I think this is a good change to make, reducing the chance of bugs.
> 
> With the change of next_timeout to `ulong` we should update `fd_quic_svc_timers_init_conn` for consistency, not a functional bug though.
> 
> ```c
>   conn->svc_meta.next_timeout = ULONG_MAX;
> ```
> 
> to
> 
> ```c
>   conn->svc_meta.next_timeout = LONG_MAX;
> ```
> 
> And if we want to be nitpicky, also the integer suffixes in the `test_svc_cancel` test.

These files are not being compiled in right now, they're just orphans from the revert somehow. This is addressed in https://github.com/firedancer-io/firedancer/pull/6385.

two-heart (approved):

PR #6241, opened 2025-09-03
gossip: ERR on unknown tag from gossip tile

two-heart:


ravyu-jump:
Not exactly sure why the builtin was used, should probably defer this to Michael

ravyu-jump:
Already been merged as a part of https://github.com/firedancer-io/firedancer/pull/6305

PR #6243, opened 2025-09-03
Fix /etc/localtime seccomp crash

ripatel-fd:
Fixes an fdctl sandbox crash when logfile+stderr log levels are set
to WARNING or higher.

This is due to fd_log opening /etc/localtime on the first log
message, which has to happen _before_ entering the sandbox.

This patch always opens /etc/localtime before entering the sandbox.

Thank you to James W at Figment for reporting.


ripatel-fd:
Oh sorry, didn't see #6242

PR #6248, opened 2025-09-04
another batch of rpc server fixes

asiegel-jt:


jumpsiegel (approved):
It would be interesting to factor out all the rpc topo setup stuff into shared code we could than use in backtesting.  This way, we could bring up a backtest, pause at a slot, and then explore the state using rpc...

asiegel-jt:
> It would be interesting to factor out all the rpc topo setup stuff into shared code we could than use in backtesting. This way, we could bring up a backtest, pause at a slot, and then explore the state using rpc...

you can run the rpc service as a separate process which makes debugging easy

PR #6259, opened 2025-09-05
feat: add solcap capture for bank hash mismatch debugging

TanmayDhobale:
Enables automatic solcap generation and diff analysis when bank hash mismatches occur in backtest ledgers, providing contributors with detailed debugging information instead of minimal error messages


kbhargava-jump:
unfortunately adding solcap captures to `run_ledger_backtest.sh` and `run_nightly_backtests.sh` makes end to end ledger execution slower (example: CI becomes slower), hence why developers should manually add the 3 line config for solcap captures when seeing mismatches

kbhargava-jump:
closing as CI time increases. explained above

PR #6261, opened 2025-09-05
repair: allow multiple sign tiles

emwang-jump:


src/app/firedancer/topology.c
```diff
@@ -361,7 +361,7 @@ fd_topo_initialize( config_t * config ) {
   FOR(shred_tile_cnt)  fd_topob_link( topo, "repair_shred", "shred_repair", pending_fec_shreds_depth,                 sizeof(fd_ed25519_sig_t),      1UL );
 
   /**/                 fd_topob_link( topo, "sign_ping",    "repair_sign",  128UL,                                    sizeof(fd_ed25519_sig_t),      1UL );
-  FOR(sign_tile_cnt-1) fd_topob_link( topo, "repair_sign",  "repair_sign",  128UL,                                    2048UL,                        1UL );
+  FOR(sign_tile_cnt-1) fd_topob_link( topo, "repair_sign",  "repair_sign",  1024UL /* TODO: not sure abt this one */, 128UL,                         1UL );
```
ripatel-fd, line 364: Seems fine. I'd guess it's mainly about jitter control and smoothing out request batches. 1 sign operation takes about 7.5us, so 1024 will take the sign tile ~7.7ms to work off.
There's another more complex interaction with STEM_LAZY / flow-control credit replenishment. Basically if stem publishes to too many shallow links, then read/write performance on other unrelated links can drop. But 1024 is probably fine

ripatel-fd: (Or nevermind seems like you have a custom flow control credit mechanism here, nice)

src/app/firedancer/topology.c
```diff
@@ -948,7 +948,9 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
       tile->repair.repair_intake_listen_port = config->tiles.repair.repair_intake_listen_port;
       tile->repair.repair_serve_listen_port  = config->tiles.repair.repair_serve_listen_port;
       tile->repair.slot_max                  = config->tiles.repair.slot_max;
-      strncpy( tile->repair.good_peer_cache_file, config->tiles.repair.good_peer_cache_file, sizeof(tile->repair.good_peer_cache_file) );
+
+      tile->repair.sign_link_depth = 1024; // TODO: fix it
```
ripatel-fd, line 952: Wouldn't it be cleaner to read this via `fd_mcache_depth` from the mcache handle that the repair tile gets in unprivileged_init? Saves you from passing around this value

emwang-jump: need the depth value in `scratch_footprint`, don't think I'm able to get the fd_mcache_depth there

PR #6266, opened 2025-09-08
ghost: fix scratch LAYOUT/ALLOC_APPEND order

two-heart:
even though the resulting size is the same in this case, still better to always have it in order.

Compare with:
https://github.com/firedancer-io/firedancer/blob/566cd349e7c7d66eda77304d1d1155ba2a55d4d0/src/choreo/ghost/fd_ghost.h#L249-L254


lidatong:
closing because removing ver field in another PR

PR #6281, opened 2025-09-08
util: add macros for inheritance-like behavior in C

ptaffet-jump:


ripatel-fd:
@ptaffet-jump I ended up not using it, but appreciate the feature. Let's reopen this when we have another user for it.

PR #6285, opened 2025-09-08
snapshots: gossip integration

cali-jumptrading:
Implementation:
- A snapshot peer selector is a wrapper around a treap of peers that is continuously updated by the pinger and the http resolver.
- The cluster slot is determined by the highest slot seen from snapshot hashes and from the resolver.

Features:
- Adds option to download snapshots from trusted gossip entrypoints
- Adds option to download snapshots from arbitrary gossip peers
- Implements `maximum_local_snapshot_age` to decide whether to read from disk or download a new snapshot
- Adds the ability to retry an incremental snapshot

Performance:
- Snapshot loading speed is mostly unchanged ~22 seconds. However since we are adding a lot more continuous peer selection logic to snaprd, we might want to consider separating the peer selection logic from downloading logic in the future to avoid competing for cpu resources.
- Gossip saturation generally takes between 1-1.5 minutes.


src/discof/writer/fd_writer_tile.c
```diff
@@ -505,6 +505,7 @@ populate_allowed_fds( fd_topo_t const *      topo,
 }
 
 #define STEM_BURST (1UL)
+#define STEM_LAZY 1000L
```
mmcgee-jump: rebase

config/extra/with-handholding.mk
```diff
@@ -8,3 +8,4 @@ CPPFLAGS+=-DFD_TXN_HANDHOLDING=1
 CPPFLAGS+=-DFD_FUNK_HANDHOLDING=1
 CPPFLAGS+=-DFD_RUNTIME_ERR_HANDHOLDING=1
 CPPFLAGS+=-DFD_FOREST_USE_HANDHOLDING=1
+CPPFLAGS+=-DFD_SSPEER_SELECTOR_USE_HANDHOLDING=1
```
amass-jump: This define is only used for one line. My general feeling is that we probably don't need a huge list of specific handholding flags. Maybe we can define an overarching FD_USE_HANDHOLDING, or use some other flag within the sspeer code?

src/funk/fd_funk_txn.c
```diff
@@ -410,12 +410,14 @@ fd_funk_txn_cancel_root( fd_funk_t * funk ) {
   fd_funk_rec_map_t * rec_map = funk->rec_map;
   fd_funk_rec_pool_t * rec_pool = funk->rec_pool;
   ulong chain_cnt = fd_funk_rec_map_chain_cnt( rec_map );
+  fd_funk_rec_map_iter_t iter_next;
```
amass-jump: What are these changes? They seem unrelated

src/app/firedancer/config/default.toml
```diff
@@ -353,6 +353,21 @@ user = ""
     # you highly trust, and this is the default behavior of the
     # configuration.
     [snapshots.sources]
+        # Allow fetching of snapshots from peers specified in the gossip
```
amass-jump, line 408: Perhaps as a ~last step once the overall snapshots code is fully implemented we can come back and revisit the config structure / comments. The snapshot logic we've sketched out is pretty nuanced but I think the operators should be able to easily understand what is supposed to happen based on just skimming these comments

src/app/firedancer/config/default.toml
```diff
@@ -353,6 +353,21 @@ user = ""
     # you highly trust, and this is the default behavior of the
     # configuration.
     [snapshots.sources]
```
amass-jump, line 407: in a future PR, perhaps we should add config option(s) to allow operators to never use certain gossip peers for snapshots. For example maybe their validator keeps trying to download from a given peer and failing (which is something we should handle automatically, but still having a way for an operator to control this seems reasonable)

src/app/firedancer/topology.c
```diff
@@ -206,7 +206,13 @@ setup_snapshots( config_t *       config,
   tile->snaprd.maximum_download_retry_abort      = config->firedancer.snapshots.maximum_download_retry_abort;
   tile->snaprd.max_full_snapshots_to_keep        = config->firedancer.snapshots.max_full_snapshots_to_keep;
   tile->snaprd.max_incremental_snapshots_to_keep = config->firedancer.snapshots.max_incremental_snapshots_to_keep;
+  tile->snaprd.entrypoints_enabled               = config->firedancer.snapshots.sources.entrypoints.enabled;
+  tile->snaprd.gossip_peers_enabled              = config->firedancer.snapshots.sources.gossip.enabled;
   resolve_snapshot_peers( config, tile );
+
+  tile->snaprd.gossip_entrypoints_cnt = config->gossip.entrypoints_cnt;
+  fd_memcpy( tile->snaprd.gossip_entrypoints, config->gossip.resolved_entrypoints, tile->snaprd.gossip_entrypoints_cnt * sizeof(fd_ip4_port_t) );
+
```
amass-jump: I think this is a merge conflict, looks like this code has moved on main

src/app/firedancer/topology.c
```diff
@@ -318,8 +324,8 @@ fd_topo_initialize( config_t * config ) {
     fd_topob_wksp( topo, "snap_stream" );
     fd_topob_wksp( topo, "snap_zstd" );
     fd_topob_wksp( topo, "snap_out" );
+    fd_topob_wksp( topo, "snap_signal" );
```
amass-jump: `signal` isn't very descriptive, what is this used for? It seems like the convention would be `snaprd_repair` or something of that sort?

src/app/firedancer/topology.c
```diff
@@ -726,6 +726,7 @@ fd_topo_initialize( config_t * config ) {
     fd_topob_tile_in ( topo, "snapin", 0UL, "metric_in", "snap_stream", 0UL, FD_TOPOB_RELIABLE, FD_TOPOB_POLLED   );
     fd_topob_tile_out( topo, "snapin", 0UL, "snap_out", 0UL );
     fd_topob_tile_in ( topo, "replay", 0UL, "metric_in", "snap_out", 0UL, FD_TOPOB_RELIABLE, FD_TOPOB_POLLED );
+    fd_topob_tile_out( topo, "snaprd", 0UL, "snap_signal", 0UL );
```
amass-jump: Perhaps we should add this in a separate PR where we link snaprd and the repair tile?

src/discof/restore/fd_snapin_tile.c
```diff
@@ -28,6 +28,7 @@ struct fd_snapin_tile {
   long boot_timestamp;
 
   fd_funk_t       funk[1];
+  fd_funk_txn_t * root_funk_txn;
```
amass-jump: Does this need any updating after @ripatel-fd 's latest funk PRs?

src/funk/fd_funk_txn.c
cali-jumptrading: Ah this is to fix a bug with canceling the funk root. we cancel the funk root txn in the snapin tile when we reset the full snapshot.

config/extra/with-handholding.mk
cali-jumptrading: My thoughts exactly!

src/discof/restore/utils/fd_ssping.h
```diff
@@ -18,13 +18,21 @@
...
 #define FD_SSPING_MAGIC (0xF17EDA2CE55A1A60) /* FIREDANCE SSPING V0 */
 
 struct fd_ssping_private;
 typedef struct fd_ssping_private fd_ssping_t;
 
+typedef void
+(* fd_ssping_on_ping_fn_t)( void *        _ctx,
+                            fd_ip4_port_t addr,
+                            ulong         latency );
+
```
amass-jump, line 35: I am not sure about the callback interface, I think it can be a little hard to follow / unwieldy. 

Separately it seems like a call to `fd_ssping_advance` can result in more than just a new (addr,latency) result, it could also invalidate a given addr. We are mixing direct calls to fd_sspeer_selector_ functions with on_ping callbacks

Overall I think it would be easier to follow if fd_ssping was fully self contained. All it would expose to the client (fd_snaprd_tile) would be a list of addrs and their latest ping latency / state. For example, a failed HTTP download probably shouldn't invalidate that peer in ssping. We can keep pinging it but the logic to invalidate that peer could be downstream in snaprd / etc.

fd_ssping_advance could either fill in an array of changed peers on return, or expect the client to call some other function to query all the changed peers since the last call and do housekeeping (add/remove from selector, etc.)

Another way to think about this is we could be able to re-use fd_ssping for any other general purpose "I want to ping a bunch of peers and track their latency/state" use-case

src/discof/restore/utils/fd_http_resolver.h
```diff
@@ -0,0 +1,51 @@
...
+FD_PROTOTYPES_BEGIN
+
+FD_FN_CONST ulong
+fd_http_resolver_align( void );
+
+FD_FN_CONST ulong
+fd_http_resolver_footprint( ulong peers_cnt );
+
+void *
+fd_http_resolver_new( void *                           shmem,
```
amass-jump, line 29: It took me a long time to realize that fd_http_resolver, and fd_ssresolve, are only for the non-gossip HTTP peers (is that correct?). Additionally they are only for getting the slot number being served by that peer, but not used in the actual HTTP download itself.

For one, I think it would be nice to name these or document somehow that they are only for the non-gossip peers, as really _all_ peers are HTTP downloads.

Secondly it seems like the HTTP request to resolve the slot number and the HTTP request to do a download are identical requests? So we have a lot of duplicated code. Relatedly we should maybe be doing a HEAD request rather than a GET request.

One way we could structure the code is considering gossip & non-gossip peers as the same, the only difference being that non-gossip/ "HTTP" peers took a shortcut and we don't need to resolve what their endpoint is from gossip. Similarly the initial slot resolution and the actual HTTP download are nearly the same, with the exception that we stop once we get the redirect filename rather than continuing to do the download

src/funk/fd_funk_txn.c
cali-jumptrading: Actually I think when I rebase this will no longer be relevant lol

src/app/firedancer/topology.c
```diff
@@ -395,17 +401,13 @@ fd_topo_initialize( config_t * config ) {
   /**/                 fd_topob_link( topo, "sign_send",    "sign_send",    128UL,                                    sizeof(fd_ed25519_sig_t),      1UL   );
   if( FD_LIKELY( !disable_snap_loader ) ) {
   /**/                 fd_topob_link( topo, "snap_zstd",    "snap_zstd",    8192UL,                                   16384UL,                       1UL );
+  /**/                 fd_topob_link( topo, "snap_signal",  "snap_signal",  128UL,                                    1UL,                           1UL )->permit_no_consumers = 1; /* TODO: plumb into repair */
```
amass-jump: OOC why MTU of 1? Are we passing everything in the mcache? Could we use MTU 0?

src/discof/restore/utils/fd_ssping.c
```diff
@@ -291,7 +275,6 @@ unping_peer( fd_ssping_t *      ssping,
   if( FD_UNLIKELY( peer->state==PEER_STATE_PINGED ) ) {
     deadline_list_ele_remove( ssping->pinged, peer, ssping->pool );
   } else if( FD_UNLIKELY( peer->state==PEER_STATE_REFRESHING ) ) {
-    score_treap_ele_remove( ssping->score_treap, peer, ssping->pool );
```
amass-jump, line 294: Do we have/need any callback from here so that the sspeer_selector knows to remove this peer from the score table if we suddenly start failing to ping it?

src/discof/restore/utils/fd_ssping.c
```diff
@@ -371,18 +353,15 @@ poll_advance( fd_ssping_t * ssping,
       FD_TEST( peer->deadline_nanos>now );
       peer->latency_nanos = PEER_DEADLINE_NANOS_PING - (ulong)(peer->deadline_nanos - now);
 
-      if( FD_LIKELY( peer->state==PEER_STATE_REFRESHING ) ) {
-        score_treap_ele_remove( ssping->score_treap, peer, ssping->pool );
-      }
-
-      FD_LOG_INFO(( "pinged " FD_IP4_ADDR_FMT ":%hu in %lu ns", FD_IP4_ADDR_FMT_ARGS( peer->addr.addr ), fd_ushort_bswap( peer->addr.port ), peer->latency_nanos ));
+      FD_LOG_INFO(( "pinged " FD_IP4_ADDR_FMT ":%hu in %lu ns", FD_IP4_ADDR_FMT_ARGS( peer->addr.addr ), peer->addr.port, peer->latency_nanos ));
```
amass-jump: do we want to keep this log if there will be ~2K peers now?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -25,37 +27,75 @@
...
-#define FD_SNAPRD_STATE_SHUTDOWN                  (12) /* The tile is done, and has likely already exited */
-
-#define SNAPRD_FILE_BUF_SZ (1024UL*1024UL) /* 1 MiB */
+#define FD_SNAPRD_STATE_WAITING_FOR_PEERS               ( 0) /* Waiting for first peer to arrive from gossip to download from */
+#define FD_SNAPRD_STATE_WAITING_FOR_PEERS_INCREMENTAL   ( 1) /* Waiting for peers when attempting to download an incremental snapshot */
+#define FD_SNAPRD_STATE_COLLECTING_PEERS                ( 2) /* First peer arrived, wait a little longer to see if a better one arrives */
+#define FD_SNAPRD_STATE_COLLECTING_PEERS_INCREMENTAL    ( 3) /* Collecting peers to download an incremental snapshot */
+#define FD_SNAPRD_STATE_READING_FULL_FILE               ( 4) /* Full file looks better than peer, reading it from disk */
+#define FD_SNAPRD_STATE_FLUSHING_FULL_FILE              ( 5) /* Full file was read ok, confirm it decompressed and inserted ok */
+#define FD_SNAPRD_STATE_FLUSHING_FULL_FILE_RESET        ( 6) /* Resetting to load full snapshot from file again, confirm decompress and inserter are reset too */
```
amass-jump: Definitely not something to change in this PR, but a question: If snaprd decides to reset the whole pipeline, does it actually need to wait for the downstream tiles to ack the flush before it starts writing a new snapshot into the link? Or could it just add a "sequence" number or something to the frags to indicate to the downstream tiles that these frags are for a new snapshot and they should reset everything before processing them?

amass-jump: Also theoretically could we could support downloading full snapshots and incremental snapshots in parallel? Once we start the full download we know what the base slot number will be, right? Also something for the future

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -134,8 +191,11 @@ struct fd_snaprd_tile {
   } gossip_in;
 
   struct {
-    fd_gossip_update_message_t tmp_upd_buf;
-    fd_contact_info_t *        ci_table;
+    fd_snaprd_gossip_ci_entry_t * ci_pool;
+    gossip_ci_map_t *             ci_map;
+    fd_gossip_update_message_t    tmp_upd_buf;
+    fd_ip4_port_t                 entrypoints[ 16UL ];
```
amass-jump: GOSSIP_TILE_ENTRYPOINTS_MAX ?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -170,6 +241,14 @@ should_shutdown( fd_snaprd_tile_t * ctx ) {
   return ctx->state==FD_SNAPRD_STATE_SHUTDOWN;
 }
 
+static inline int
+is_entrypoint( fd_snaprd_tile_t * ctx, fd_ip4_port_t addr ) {
+  for( ulong i=0UL; i<ctx->gossip.entrypoints_cnt; i++ ) {
+    if( FD_UNLIKELY( ctx->gossip.entrypoints[ i ].l==addr.l ) ) return 1;
+  }
+  return 0;
+}
```
amass-jump, line 249: Is it possible to have two separate validators, each behind the same IP address but using different ports? Should we be filtering on pubkey / etc?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -185,6 +264,81 @@ metrics_write( fd_snaprd_tile_t * ctx ) {
...
+        ctx->predicted_incremental.ssinfo.incremental.slot!=best_incremental_peer.ssinfo.incremental.slot ) ) {
+        ctx->predicted_incremental.addr   = best_incremental_peer.addr;
+        ctx->predicted_incremental.ssinfo = best_incremental_peer.ssinfo;
+
+        FD_LOG_INFO(( "new best incremental peer is " FD_IP4_ADDR_FMT ":%hu with slot %lu",
+                         FD_IP4_ADDR_FMT_ARGS( best_incremental_peer.addr.addr ), best_incremental_peer.addr.port,
+                         best_incremental_peer.ssinfo.incremental.slot));
+      uint tsorig; uint tspub;
+      fd_ssmsg_slot_to_frag( best_incremental_peer.ssinfo.incremental.slot, &tsorig, &tspub );
+      fd_stem_publish( ctx->stem, 1UL, FD_SSMSG_EXPECTED_SLOT, 0UL, 0UL, 0UL, tsorig, tspub );
```
amass-jump: Do we always have credit to publish here?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -509,9 +683,12 @@ after_credit( fd_snaprd_tile_t *  ctx,
   (void)opt_poll_in;
   (void)charge_busy;
 
+  ctx->stem = stem;
```
amass-jump: Is it an existing pattern to save the stem pointer like this? AFAICT this is a temporary / stack object so should only be used from within this function call

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -698,37 +913,71 @@ after_frag( fd_snaprd_tile_t *  ctx,
...
+            entry    = gossip_ci_pool_ele( ctx->gossip.ci_pool, idx );
+            cur_addr = entry->addr;
           }
-          fd_contact_info_t * new = msg->contact_info.contact_info;
-          fd_ip4_port_t new_addr  = new->sockets[ FD_CONTACT_INFO_SOCKET_RPC ];
-          if( new_addr.l ) {
-            fd_ssping_add( ctx->ssping, new_addr );
+
+          fd_ip4_port_t new_addr = msg->contact_info.contact_info->sockets[ FD_CONTACT_INFO_SOCKET_RPC ];
+          new_addr.port          = fd_ushort_bswap( new_addr.port );
```
amass-jump, line 904: Just to confirm, snapshots are always served on the RPC ip:port?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -698,37 +913,71 @@ after_frag( fd_snaprd_tile_t *  ctx,
...
+                gossip_ci_map_idx_insert( ctx->gossip.ci_map, msg->contact_info.idx, ctx->gossip.ci_pool );
+              }
+              entry->addr = new_addr;
+            }
           }
-          *cur = *new;
+          break;
         }
-        break;
       case FD_GOSSIP_UPDATE_TAG_CONTACT_INFO_REMOVE: {
```
amass-jump, line 934: Any there any security considerations to cancelling a snapshot download from a gossip peer that gets removed from the table?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -698,37 +913,71 @@ after_frag( fd_snaprd_tile_t *  ctx,
...
+        }
         break;
+      }
     }
 
   } else {
+    if( sig!=FD_SNAPSHOT_MSG_CTRL_ACK && sig!=FD_SNAPSHOT_MSG_CTRL_MALFORMED ) {
+      FD_LOG_ERR(( "unexpected sig %lu", sig ));
+    }
     FD_TEST( sig==FD_SNAPSHOT_MSG_CTRL_ACK || sig==FD_SNAPSHOT_MSG_CTRL_MALFORMED );
```
amass-jump: ?

src/discof/restore/utils/fd_http_resolver.h
```diff
@@ -0,0 +1,51 @@
+#ifndef HEADER_fd_src_discof_restore_fd_http_resolver_h
```
amass-jump: restore_utils_fd?

src/discof/restore/utils/fd_http_resolver.c
```diff
@@ -0,0 +1,442 @@
...
+static int
+create_socket( fd_http_resolver_t *  resolver,
+               fd_ssresolve_peer_t * peer ) {
+  int sockfd = socket( PF_INET, SOCK_STREAM|SOCK_NONBLOCK, 0 );
+  if( FD_UNLIKELY( -1==sockfd ) ) FD_LOG_ERR(( "socket failed (%i-%s)", errno, strerror( errno ) ));
+
+  int optval = 1;
+  if( FD_UNLIKELY( -1==setsockopt( sockfd, SOL_TCP, TCP_NODELAY, &optval, sizeof(int) ) ) ) {
+    FD_LOG_ERR(( "setsockopt() failed (%d-%s)", errno, fd_io_strerror( errno ) ));
+  }
```
amass-jump, line 207: AFAIK this only affects outgoing packets, so probably has no real purpose here but I suppose it can't hurt

src/discof/restore/utils/fd_ssresolve.c
```diff
@@ -0,0 +1,301 @@
...
+fd_ssresolve_render_req( fd_ssresolve_t * ssresolve,
+                         fd_ip4_port_t    addr ) {
+  if( FD_LIKELY( ssresolve->full ) ) {
+    FD_TEST( fd_cstr_printf_check( ssresolve->request, sizeof(ssresolve->request), &ssresolve->request_len,
+           "GET %.*s HTTP/1.1\r\n"
+           "User-Agent: Firedancer\r\n"
+           "Accept: */*\r\n"
+           "Accept-Encoding: identity\r\n"
+           "Host: " FD_IP4_ADDR_FMT "\r\n\r\n",
+           17, "/snapshot.tar.bz2", FD_IP4_ADDR_FMT_ARGS( addr.addr ) ) );
```
amass-jump: couldn't this just go in the string above rather than being an argument?

src/discof/restore/utils/fd_sspeer_selector.c
```diff
@@ -0,0 +1,367 @@
...
+                                                                          implementation to cmp either */
+#define TREAP_IDX_T     ulong
+#define TREAP_LT        COMPARE_WORSE
+#define TREAP_PARENT    score_treap.parent
+#define TREAP_LEFT      score_treap.left
+#define TREAP_RIGHT     score_treap.right
+#define TREAP_PRIO      score_treap.prio
+#include "../../../util/tmpl/fd_treap.c"
+
+#define DEFAULT_SLOTS_BEHIND   (1000UL*1000UL)        /* 100_000 slots behind */
```
amass-jump: s/100_000/1_000_000/ ?

src/discof/restore/utils/fd_sspeer_selector.c
```diff
@@ -0,0 +1,367 @@
...
+  score_treap_t *       shadow_score_treap;
+  ulong *               peer_idx_list;
+  fd_ssinfo_t           cluster_ssinfo;
+
+  ulong                 magic; /* ==FD_SSPEER_SELECTOR_MAGIC */
+};
+
+FD_FN_CONST ulong
+fd_sspeer_selector_align( void ) {
+  return alignof(fd_sspeer_t);
```
amass-jump: s/fd_sspeer_t/fd_sspeer_selector_t/?

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: yep!

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: not an existing pattern, but we do this in the `snapin` tile as well.

src/discof/restore/utils/fd_ssresolve.c
cali-jumptrading: oh yeah nice catch!

src/discof/restore/utils/fd_sspeer_selector.c
cali-jumptrading: will fix this, thanks!

src/discof/restore/utils/fd_sspeer_selector.c
cali-jumptrading: will fix this, thanks!

src/discof/restore/utils/fd_http_resolver.h
cali-jumptrading: will fix this, thanks!

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: oh this was for debugging, will remove

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: hmm filtering on pubkey does seem like a stronger guarantee. I'll try it out!

src/discof/restore/utils/fd_ssping.c
cali-jumptrading: Ah probably not.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: ah yeah, will fix!

src/discof/restore/utils/fd_http_resolver.c
cali-jumptrading: We added TCP_NODELAY because it helped with increasing download speed.

src/app/firedancer/topology.c
cali-jumptrading: Fixed

src/discof/restore/utils/fd_ssping.c
cali-jumptrading: I had `fd_ssping_advance` take a handle to the `sspeer_selector` for this purpose... which is kinda ugly... but I don't have a better way to do it without re-writing ssping.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: done!

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: hmm both good points:
- for the first point, sequence numbers will work, but overall control flow gets a little more complicated and harder to reason about... I prefer this slower, simpler to reason about lockstem approach.
- hmm we could download in parallel... would require some architectural changes to the tiles... because we still have to process each snapshot one at a time.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: Actually... this is not possible because we don't know the pubkeys until we filter by address first...

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: you're right we might not always have credit to publish here... it's hard to define a bound for the number of frags for this link as well... I'm tempted to make this link unreliable and just let repair handle dropped frags.

cali-jumptrading: Actually it's up to repair to figure out how to consume the link, as either reliable or unreliable.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: You mean we should cancel a snapshot download from a gossip peer that gets removed from the table? I think that is fine... but bloats up the gossip remove logic a bit. Not sure we need it... if the peer goes offline the download will fail and we'll retry.

src/discof/restore/fd_snaprd_tile.c
amass-jump: Not sure I follow, is it not the case that a link is reliable if the publisher respects consumer's credits, and unreliable if it does not? How would a consumer decide whether to be reliable or unreliable?

src/discof/restore/fd_snaprd_tile.c
amass-jump: Potentially yes. In general, what actions should/could we take on contact info removal?

cali-jumptrading: well right now on contact info removal we remove the peer from the pinger and from the selector. In general we should probably cancel the download if it is occurring from the same peer. I can add that in.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: in topology.c, consumer links are specified as reliable or unreliable:
```
fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "genesi_out",   0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
```

Once repair specifies that it wants to consume as reliable, then the producer "should" respect credits. But our credit control is kinda broken with STEM_BURST anyways...

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: wait the main reason why I was kinda against this was because once we cancel the download we'll have to transition the state machine and that involves a good amount of complexity... Not sure if this complexity is worth it.

cali-jumptrading: let me try it out actually.

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -128,11 +128,12 @@ backtest_topo( config_t * config ) {
     fd_topob_wksp( topo, "snapdc_rd" );
     fd_topob_wksp( topo, "snapin_rd" );
     fd_topob_wksp( topo, "snap_out" );
-    fd_topob_wksp( topo, "replay_manif" );
+    fd_topob_wksp( topo, "snaprd_rp" );
```
amass-jump, line 125: nit: Is there a length limit on these names? AFAICT the convention for link names is `SourceTileName_DestTileName` so this should be `snaprd_repair` ?

src/app/firedancer/topology.c
```diff
@@ -485,7 +487,7 @@ fd_topo_initialize( config_t * config ) {
   /**/                 fd_topob_tile_out(   topo, "gossip", 0UL,                        "gossip_gossv", 0UL                                                );
 
   if( FD_LIKELY( snapshots_enabled ) ) {
-                      fd_topob_tile_in(     topo, "snaprd",  0UL,          "metric_in", "gossip_out",   0UL,          FD_TOPOB_UNRELIABLE, FD_TOPOB_POLLED ); /* TODO: Fix backpressure issues and change this back to reliable. */
+                      fd_topob_tile_in(     topo, "snaprd",  0UL,          "metric_in", "gossip_out",   0UL,          FD_TOPOB_UNRELIABLE, FD_TOPOB_POLLED   );
```
amass-jump: How will we recover from dropped frags? Is the gossip list constantly refreshed even if the gossip tile has already seen contactInfos/etc. for some peers? Seems like the convention is reliable links so I'm curious why we are using unreliable here

src/discof/restore/Local.mk
```diff
@@ -14,6 +14,9 @@ $(call add-objs,utils/fd_ssload,fd_discof)
 $(call add-objs,utils/fd_slot_delta_parser,fd_discof)
 $(call make-unit-test,test_ssmanifest_parser,utils/test_ssmanifest_parser,fd_discof fd_flamenco fd_ballet fd_util)
 $(call make-unit-test,test_slot_delta_parser,utils/test_slot_delta_parser,fd_discof fd_flamenco fd_ballet fd_util)
+$(call make-unit-test,test_sspeer_selector,utils/test_sspeer_selector,fd_discof fd_flamenco fd_ballet fd_util)
+$(call run-unit-test,test_slot_delta_parser)
+$(call run-unit-test,test_sspeer_selector)
```
amass-jump, line 20: Do we need a run-unit-test for ssmanifest_parser too?

src/discof/restore/fd_snaprd_tile.c
mmcgee-jump: I'm not sure if it makes sense to abort a download if the peer goes offline from gossip. If they are still serving the snapshot it should be fair to continue?

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: ok I fixed this by only publishing one expected slot message per tile run loop.

src/discof/restore/utils/fd_ssping.c
```diff
@@ -332,7 +318,9 @@ poll_advance( fd_ssping_t * ssping,
   else if( FD_UNLIKELY( -1==nfds && errno==EINTR ) ) return;
   else if( FD_UNLIKELY( -1==nfds ) ) FD_LOG_ERR(( "poll failed (%i-%s)", errno, strerror( errno ) ));
 
-  for( ulong i=0UL; i<ssping->fds_len; i++ ) {
+  ulong next_poll_idx = fd_ulong_min( ssping->poll_idx+FD_SSPING_RATE_LIMIT, ssping->fds_len );
+
+  for( ulong i=ssping->poll_idx; i<next_poll_idx; i++ ) {
```
amass-jump: nit: we are polling all ~2K fds but then only processing 64, I think we can just poll the subset directly and save extra kernel cost?

src/discof/restore/utils/fd_ssping.c
```diff
@@ -369,22 +357,20 @@ poll_advance( fd_ssping_t * ssping,
...
       deadline_list_ele_push_tail( ssping->valid, peer, ssping->pool );
-      score_treap_ele_insert( ssping->score_treap, peer, ssping->pool );
       remove_ping_fd( ssping, i );
+
+      ssping->on_ping_cb( ssping->cb_arg, peer->addr, peer->latency_nanos );
     }
   }
+
+  if( FD_UNLIKELY( next_poll_idx==ssping->fds_len ) ) ssping->poll_idx = 0UL;
+  else                                                ssping->poll_idx = next_poll_idx;
```
amass-jump: If I am not mistaken next_poll_idx can be greater than ssping->fds_len here if we unping any peers in the loop. IMO these kinds of wraparound checks should always be >= even if it shouldn't be possible to be more than ==. It's a cheap insurance policy

src/discof/restore/utils/fd_ssresolve.c
```diff
@@ -0,0 +1,301 @@
...
+                                    &message_len,
+                                    headers,
+                                    &header_cnt,
+                                    ssresolve->response_len - (ulong)read );
+  if( FD_UNLIKELY( parsed==-1 ) ) {
+    FD_LOG_WARNING(( "malformed response body" ));
+    return FD_SSRESOLVE_ADVANCE_ERROR;
+  } else if( parsed==-2 ) {
+    return FD_SSRESOLVE_ADVANCE_ERROR;
+  }
```
amass-jump, line 241: Why is this ERROR and not AGAIN? Isn't -2 a partial response, i.e. more bytes needed?

src/discof/restore/utils/fd_ssresolve.c
```diff
@@ -0,0 +1,301 @@
...
+
+  int is_redirect = (status==301) | (status==303) | (status==304) | (status==307) | (status==308);
```
amass-jump: Why not `302 Found` ooc? I feel like I've seen this one before surfin' the web

src/discof/restore/utils/fd_http_resolver.c
```diff
@@ -0,0 +1,442 @@
...
+
+  void *                           cb_arg;
+  fd_http_resolver_on_resolve_fn_t on_resolve_cb;
+
+  ulong                            magic; /* ==FD_HTTP_RESOLVER_MAGIC */
+};
+
+FD_FN_CONST ulong
+fd_http_resolver_align( void ) {
+  return alignof(fd_http_resolver_t);
```
amass-jump: should we max this w/ the align of other layout components?

src/discof/restore/utils/fd_http_resolver.c
amass-jump: That's interesting... I can't think of any reason why this should impact download speed

src/discof/restore/utils/fd_sspeer_selector.h
```diff
@@ -0,0 +1,102 @@
+#ifndef HEADER_fd_src_discof_restore_fd_sspeer_selector_h
+#define HEADER_fd_src_discof_restore_fd_sspeer_selector_h
```
amass-jump: utils

src/discof/restore/utils/fd_sspeer_selector.c
```diff
@@ -0,0 +1,382 @@
...
+FD_FN_CONST ulong
+fd_sspeer_selector_align( void ) {
+  return fd_ulong_max( alignof( fd_sspeer_selector_t), fd_ulong_max( peer_pool_align(), fd_ulong_max( score_treap_align(), alignof(ulong) ) ) );
+}
+
+FD_FN_CONST ulong
+fd_sspeer_selector_footprint( ulong max_peers ) {
+  ulong l;
+  l = FD_LAYOUT_INIT;
+  l = FD_LAYOUT_APPEND( l, fd_sspeer_selector_align(), sizeof(fd_sspeer_selector_t) );
```
amass-jump: nit: this should be alignof(fd_sspeer_selector_t), not alignof the overall layout?

src/discof/restore/utils/fd_sspeer_selector.c
```diff
@@ -0,0 +1,382 @@
...
+
+/* Calculates a score for a peer given its latency and ssinfo */
+ulong
+fd_sspeer_selector_score( fd_sspeer_selector_t * selector,
+                          ulong                  peer_latency,
+                          fd_ssinfo_t const *    ssinfo ) {
+  static const ulong slots_behind_penalty = 1000UL;
+  ulong slot                              = ULONG_MAX;
+  ulong slots_behind                      = DEFAULT_SLOTS_BEHIND;
+  peer_latency = peer_latency!=ULONG_MAX ? peer_latency : DEFAULT_PEER_LATENCY;
```
amass-jump, line 223: In a future PR I think we should make these configurable params

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -25,37 +28,82 @@
...
+#include "../../util/tmpl/fd_map_chain.c"
+
 struct fd_snaprd_tile {
-  fd_ssping_t * ssping;
-  fd_sshttp_t * sshttp;
+  fd_ssping_t *          ssping;
+  fd_sshttp_t *          sshttp;
+  fd_http_resolver_t *   ssresolver;
+  fd_sspeer_selector_t * selector;
+  fd_stem_context_t *    stem;
```
amass-jump: I think this is no longer used (which is good because it shouldn't be persisted AFAICT)

amass-jump (dismissed):

src/app/firedancer-dev/commands/backtest.c
cali-jumptrading: yeah there is a name limit, I think it's 7 or 8 characters right now.

src/app/firedancer/topology.c
cali-jumptrading: oh this was because I forgot to change this back to reliable!

src/discof/restore/Local.mk
cali-jumptrading: um... the thing is that unit test requires a file input and we never checked a manifest file in lol

src/discof/restore/utils/fd_ssresolve.c
cali-jumptrading: oh nice! I didn't realize 302 was a valid redirect code as well.

src/discof/restore/utils/fd_ssping.c
cali-jumptrading: oh nice catch! will fix

src/discof/restore/utils/fd_ssresolve.c
cali-jumptrading: oh this must have been a copy-paste error, thanks for catching!

src/discof/restore/utils/fd_ssping.c
cali-jumptrading: yeah, nice catch! I will update.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: Ah yeah good catch!

src/discof/restore/utils/fd_sspeer_selector.c
cali-jumptrading: yep I agree!

src/discof/restore/utils/fd_sspeer_selector.c
cali-jumptrading: yes, thanks!

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: that's what I was thinking...

src/app/firedancer-dev/commands/backtest.c
cali-jumptrading: oh actually the length limit is 12 characters. But I think `snaprd_repair` is still too long for that.

src/app/firedancer/topology.c
```diff
@@ -321,11 +322,12 @@ fd_topo_initialize( config_t * config ) {
   FOR(quic_tile_cnt)   fd_topob_link( topo, "quic_net",     "net_quic",     config->net.ingress_buffer_size,          FD_NET_MTU,                    1UL );
 
   if( FD_LIKELY( snapshots_enabled ) ) {
-  /**/                 fd_topob_link( topo, "snap_zstd",    "snap_zstd",    8192UL,                                   16384UL,                       1UL ); /* TODO: Rename */
-  /**/                 fd_topob_link( topo, "snap_stream",  "snap_stream",  2048UL,                                   USHORT_MAX,                    1UL ); /* TODO: Rename */
+  /**/                 fd_topob_link( topo, "snap_zstd",    "snap_zstd",    8192UL,                                   16384UL,                        1UL ); /* TODO: Rename */
```
mmcgee-jump: nit but I was pretty careful about all the alignment in here, I think this makes it a little worse

src/app/firedancer/topology.c
```diff
@@ -469,7 +471,7 @@ fd_topo_initialize( config_t * config ) {
   /**/                 fd_topob_tile_out(   topo, "gossip", 0UL,                        "gossip_gossv", 0UL                                                );
 
   if( FD_LIKELY( snapshots_enabled ) ) {
-                      fd_topob_tile_in(     topo, "snaprd",  0UL,          "metric_in", "gossip_out",   0UL,          FD_TOPOB_UNRELIABLE, FD_TOPOB_POLLED ); /* TODO: Fix backpressure issues and change this back to reliable. */
+                      fd_topob_tile_in(     topo, "snaprd",  0UL,          "metric_in", "gossip_out",   0UL,          FD_TOPOB_RELIABLE, FD_TOPOB_POLLED   );
```
mmcgee-jump: nit but please be careful with alignment in this file, else it quickly becomes unreadable

src/disco/topo/fd_topo.h
```diff
@@ -501,14 +501,20 @@ struct fd_topo_tile {
...
     struct {
-      char  snapshots_path[ PATH_MAX ];
-      int   incremental_snapshot_fetch;
-      int   do_download;
-      uint  maximum_local_snapshot_age;
-      uint  minimum_download_speed_mib;
-      uint  maximum_download_retry_abort;
-      uint  max_full_snapshots_to_keep;
-      uint  max_incremental_snapshots_to_keep;
+      char          snapshots_path[ PATH_MAX ];
```
mmcgee-jump: more style nits but I don't love moving everything way over the the right for one variable at the bottom like this, it makes it less readable imo. Just align `gossip_entrypoints_cnt` and `gossip_entrypoints` and everything else can stay left

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -25,37 +28,81 @@
...
-#define FD_SNAPRD_STATE_READING_FULL_HTTP         ( 7) /* Peer was selected, reading full snapshot from HTTP */
-#define FD_SNAPRD_STATE_FLUSHING_FULL_HTTP        ( 8) /* Full snapshot was downloaded ok, confirm it decompressed and inserted ok */
-#define FD_SNAPRD_STATE_FLUSHING_FULL_HTTP_RESET  ( 9) /* Resetting to load full snapshot from HTTP again, confirm decompress and inserter are reset too */
-#define FD_SNAPRD_STATE_READING_INCREMENTAL_HTTP  (10) /* Peer was selected, reading incremental snapshot from HTTP */
-#define FD_SNAPRD_STATE_FLUSHING_INCREMENTAL_HTTP (11) /* Incremental snapshot was downloaded ok, confirm it decompressed and inserted ok */
-#define FD_SNAPRD_STATE_SHUTDOWN                  (12) /* The tile is done, and has likely already exited */
-
-#define SNAPRD_FILE_BUF_SZ (1024UL*1024UL) /* 1 MiB */
+#define FD_SNAPRD_STATE_WAITING_FOR_PEERS               ( 0) /* Waiting for first peer to arrive from gossip to download from */
+#define FD_SNAPRD_STATE_WAITING_FOR_PEERS_INCREMENTAL   ( 1) /* Waiting for peers when attempting to download an incremental snapshot */
```
mmcgee-jump: I don't really understand why the set of states changed here? What about the gossip integration required these new states?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -360,6 +568,21 @@ drain_buffer( fd_snaprd_tile_t * ctx ) {
   else                    ctx->metrics.incremental.bytes_written += written_sz;
 }
 
+static void
+remove_temp_files( fd_snaprd_tile_t * ctx ) {
```
mmcgee-jump: Why does this exist?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -170,6 +245,60 @@ should_shutdown( fd_snaprd_tile_t * ctx ) {
...
+      }
+    }
+  }
+
+  ctx->gossip.entrypoints_received = received_gossip_entrypoints;
+  return received_gossip_entrypoints==ctx->gossip.entrypoints_cnt;
+}
+
+static int
+gossip_saturated( fd_snaprd_tile_t * ctx,
```
mmcgee-jump, line 252: I think `gossip_saturated` is a one time event, it's either been saturated or it hasn't. Having it toggle back and forth to the caller does not seem sensible

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -530,35 +773,74 @@ after_credit( fd_snaprd_tile_t *  ctx,
...
+      fd_sspeer_t best = fd_sspeer_selector_best( ctx->selector, 0, ULONG_MAX );
+      if( FD_LIKELY( best.addr.l ) ) {
+        ctx->state = FD_SNAPRD_STATE_COLLECTING_PEERS_INCREMENTAL;
+        ctx->deadline_nanos = now;
       }
       break;
     }
     case FD_SNAPRD_STATE_COLLECTING_PEERS: {
-      if( FD_UNLIKELY( now<ctx->deadline_nanos ) ) break;
+      if( FD_UNLIKELY( (!gossip_saturated( ctx, now ) || !all_entrypoints_received( ctx )) && now<ctx->deadline_nanos ) ) break;
```
mmcgee-jump, line 660: I don't love `all_entrypoints_received` here, what if one of them is offline? This deviates from Agave also?

mmcgee-jump: Oh I see, you pick if either gossip is saturated, or all entrypoints received, or timed out ... I would just remove all entrypoints received? We get those ones first since we pull request from them, but the table might not be saturated

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -530,35 +773,74 @@ after_credit( fd_snaprd_tile_t *  ctx,
...
-      if( FD_UNLIKELY( !best.l ) ) {
+      fd_sspeer_t best = fd_sspeer_selector_best( ctx->selector, 0, ULONG_MAX );
+      if( FD_UNLIKELY( !best.addr.l ) ) {
         ctx->state = FD_SNAPRD_STATE_WAITING_FOR_PEERS;
         break;
       }
 
-      ulong highest_cluster_slot = 0UL; /* TODO: Implement, using incremental snapshot slot for age */
-      if( FD_LIKELY( ctx->local_in.full_snapshot_slot!=ULONG_MAX && ctx->local_in.full_snapshot_slot>=fd_ulong_sat_sub( highest_cluster_slot, ctx->config.maximum_local_snapshot_age ) ) ) {
+      fd_ssinfo_t cluster      = fd_sspeer_selector_cluster_slot( ctx->selector );
```
mmcgee-jump, line 668: Is this Agave logic, or do they just look at highest cluster slot from entrypoints? Because any peer could just publish a maliciously high slot here and cause us to never use the disk snapshot.

src/discof/restore/utils/fd_ssmsg.h
```diff
@@ -23,6 +24,24 @@ struct epoch_credits {
 
 typedef struct epoch_credits epoch_credits_t;
 
+/* The FD_SSMSG_EXPECTED_SLOT uses the tsorig and tspub fields
```
mmcgee-jump, line 27: Why? This doesn't seem like a good scheme ...

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: They look at the highest cluster slot across all rpc peers whose snapshot hash matches the snapshot hashes published by known validators... https://github.com/anza-xyz/agave/blob/master/validator/src/bootstrap.rs#L830

Do we want to implement this idea of trusted snapshot hashes? Or only trust snapshot hashes published by entrypoints?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -511,7 +745,16 @@ after_credit( fd_snaprd_tile_t *  ctx,
...
   if( FD_LIKELY( ctx->peer_selection ) ) {
-    fd_ssping_advance( ctx->ssping, now );
+    fd_ssping_advance( ctx->ssping, now, ctx->selector );
+    fd_http_resolver_advance( ctx->ssresolver, now, ctx->selector );
+
+    /* send an expected slot message as the predicted incremental
+       could have changed as a result of the pinger, resolver, or from
+       processing gossip frags in after_frag. */
+    if( FD_LIKELY( ctx->predicted_incremental.dirty ) ) {
+      send_expected_slot( stem, ctx->predicted_incremental.slot );
```
mmcgee-jump, line 626: Can you add a metric for the predicted slot please

mmcgee-jump (approved):
few minor comments

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: At boot, we set up the local file descriptors for reading from disk and the file descriptors for writing to disk because we may actually want to download snapshots if the local snapshots are too old. If we decide to read from disk, then we'll have the partial snapshot temporary files on disk in the user's snapshots directory. It made sense to remove them when done reading from disk.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: It seems that some RPC nodes don't actually serve snapshots so sometimes I've seen errors trying to download either a full or incremental snapshot from a random gossip rpc node. If we fail while trying to download an incremental snapshot, we should retry downloading the incremental snapshot from a different peer, rather than re-downloading the full snapshot as snaprd currently does on main.

src/discof/restore/fd_snaprd_tile.c
mmcgee-jump: I don't think it makes sense to remove them. We will just overwrite them on the next run, and you would have to add a sandbox exclusion for this.

src/discof/restore/fd_snaprd_tile.c
mmcgee-jump: Ok, in future I think it would be good to separate this kind of pieces into a separate PR

cali-jumptrading: makes sense

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: sounds good, I'll remove this change.

src/discof/restore/utils/fd_ssmsg.h
cali-jumptrading: I did this to save on allocating a dcache. It's kinda hacky but it works? But if you're against this, I'll change it to be a dcache message.

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: what about the case where the user has only enabled downloading snapshots from gossip entrypoints [snapshots.sources.entrypoints]? If we don't see all entrypoints in the gossip ci table, should we allow snaprd to continue? Should we wait for gossip to saturate even though the user only cares about entrypoints?

src/discof/restore/fd_snaprd_tile.c
cali-jumptrading: This returns 1 once saturation is reached via `ctx->gossip.saturated`

src/app/firedancer/topology.c
cali-jumptrading: https://github.com/firedancer-io/firedancer/blob/main/src/app/firedancer/topology.c#L326 Actually I think it was misaligned before lol

mmcgee-jump: it's not, everything is in line with the `burst` comment at the top, except a handful of lines which spill over

cali-jumptrading: oh I didn't notice the burst alignment... I'll change it back to how it was before.

amass-jump (dismissed):

src/app/firedancer/config/default.toml
```diff
@@ -405,6 +405,21 @@ user = ""
...
+        # Allow fetching of snapshots from peers specified in the gossip
+        # entrypoints.  Gossip entrypoints must be peers that are highly
+        # trusted as they are used to bootstrap your node.
+        [snapshots.sources.entrypoints]
+            enabled = true
```
amass-jump: Perhaps we should set this to false in default.toml for now, which would give devs an equivalent experience to what we have now? Once we are confident with all the new gossip bells/whistles we can flip it on by default?

cali-jumptrading: oh yeah, let's set this to false for now.

amass-jump (dismissed):

amass-jump (approved):

PR #6286, opened 2025-09-09
gossip: add scratch regression checks

two-heart:
For more complex scratch allocations it is common practice to add assert checks, to test that a change doesn't lead to a regression

* add basic unit tests calling _new, that can be expanded later
* consistent parameter naming
* remove 0 checks, since fd_ulong_is_pow2(n) is correctly defined for n=0


ravyu-jump (commented):
LGTM, will stamp once the compile error is fixed

PR #6287, opened 2025-09-09
boot: improve boot time, logging, and diagnostics

mmcgee-jump:
About a 40% perf improvement to snapshot loading, and cleaner output

```
NOTICE  09-09 05:45:43.399533 871985 1    snapin:0 src/discof/restore/fd_snapin_tile.c(64): loaded 83.3M accounts from snapshot in 39.7 seconds
[...]
NOTICE  09-09 05:46:59.973204 872434 1    snapin:0 src/discof/restore/fd_snapin_tile.c(64): loaded 83.3M accounts from snapshot in 25.4 seconds
```


ripatel-fd:
> About a 40% perf improvement to snapshot loading

Suffix that with *for testnet ;) Mainnet loading needs more work

.vscode/tasks.json
ripatel-fd, line 1: Not sure if you meant to check this in, but format this file perhaps

src/app/firedancer/config/default.toml
```diff
@@ -281,7 +281,7 @@ user = ""
     # newly downloaded snapshots will replace ones that are already
     # present.
     #
-    # TODO: implement the "at most one downloaded " logic. "
+    # TODO: implement the "at most one downloaded" logic. "
```
ripatel-fd, line 284:
```suggestion
    # TODO: implement the "at most one downloaded" logic.
```

What does that mean?

src/disco/gui/fd_gui_tile.c
```diff
@@ -556,7 +556,7 @@ unprivileged_init( fd_topo_t *      topo,
   if( FD_UNLIKELY( scratch_top > (ulong)scratch + scratch_footprint( tile ) ) )
     FD_LOG_ERR(( "scratch overflow %lu %lu %lu", scratch_top - (ulong)scratch - scratch_footprint( tile ), scratch_top, (ulong)scratch + scratch_footprint( tile ) ));
 
-  FD_LOG_WARNING(( "GUI server listening at http://" FD_IP4_ADDR_FMT ":%u", FD_IP4_ADDR_FMT_ARGS( tile->gui.listen_addr ), tile->gui.listen_port ));
+  FD_LOG_NOTICE(( "gui server listening at http://" FD_IP4_ADDR_FMT ":%u", FD_IP4_ADDR_FMT_ARGS( tile->gui.listen_addr ), tile->gui.listen_port ));
```
ripatel-fd, line 559: ?

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -355,6 +364,70 @@ rename_snapshots( fd_snaprd_tile_t * ctx ) {
   }
 }
 
+static void
+print_diagnostics( fd_snaprd_tile_t * ctx ) {
+  switch( ctx->state ) {
+    case FD_SNAPRD_STATE_WAITING_FOR_PEERS:
```
ripatel-fd, line 370: I thought case indent goes on the same level of the switch statement since those are labels, but probably doesn't matter

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -377,12 +450,17 @@ after_credit( fd_snaprd_tile_t *  ctx,
      downstream consumers, snapdc and snapin. */
 #define NUM_SNAP_CONSUMERS (2UL)
 
+  if( FD_UNLIKELY( now>ctx->diagnostic_deadline_nanos ) ) {
+    ctx->diagnostic_deadline_nanos = now+1L*1000L*1000L*1000L;
```
ripatel-fd:
```suggestion
    ctx->diagnostic_deadline_nanos = now+(long)1e9;
```

src/app/firedancer/config/default.toml
mmcgee-jump: I think this is changing soon with Caroline PR

src/disco/gui/fd_gui_tile.c
mmcgee-jump: Switching the routine log lines to be lowercase

src/app/firedancer/topology.c
```diff
@@ -198,6 +198,7 @@ static void
 setup_snapshots( config_t *       config,
                  fd_topo_tile_t * tile ) {
   fd_memcpy( tile->snaprd.snapshots_path, config->paths.snapshots, PATH_MAX );
+  tile->snaprd.diagnostics                       = 1;
```
cali-jumptrading, line 201: Is this meant to always enable diagnostics? Or would we make this a config option in the future to disable diagnostics?

mmcgee-jump: Config option to enable or disable a specific logline is overkill. The option is just to disable it from commands/tools (e.g. `snapshot-load`) which has its own diagnostics.

PR #6290, opened 2025-09-09
ethtool: add RSS flow steering dedicated queue mode

amass-jump:


src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -126,104 +130,181 @@ init( config_t const * config ) {
...
+                     "and there must be one channel per tile. You can either use a NIC "
+                     "that supports multiple channels, or run Firedancer with only one "
+                     "net tile. You can configure Firedancer to run with only one net "
+                     "tile by setting `layout.net_tile_count` to 1 in your "
+                     "configuration file.",
+                     device, config->layout.net_tile_count ));
+      } else {
+        NOT_CONFIGURED( "device `%s` does not have right number of channels (got %u but "
+                         "expected %u)",
+                         device, channels.current, config->layout.net_tile_count );
```
ripatel-fd:
```suggestion
        NOT_CONFIGURED( "device `%s` does not have right number of channels (got %u but "
                        "expected %u)",
                        device, channels.current, config->layout.net_tile_count );
```

src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -126,104 +130,181 @@ init( config_t const * config ) {
...
 
-  if( FD_UNLIKELY( close( sock ) ) )
-    FD_LOG_ERR(( "error configuring network device, close() socket failed (%i-%s)", errno, fd_io_strerror( errno ) ));
+    uint rxfh_table[ FD_ETHTOOL_MAX_RXFH_TABLE_SIZE ] = { 0 };
+    uint rxfh_table_size;
+    FD_TEST( 0==fd_ethtool_ioctl_rxfh_get_table( &ioc, rxfh_table, &rxfh_table_size ) );
+    int rxfh_error = 0;
+    uint default_queue = 0;
+    uint configured_queue = 1; /* If properly configured the 0th queue is excluded */
+    for( uint j=0u; j<rxfh_table_size; ++j) {
```
ripatel-fd:
```suggestion
    for( uint j=0u; j<rxfh_table_size; j++ ) {
```

src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -126,104 +130,181 @@ init( config_t const * config ) {
...
+      ushort ports[ 32 ];
+      ports[ num_ports++ ] = config->tiles.shred.shred_listen_port;
+      ports[ num_ports++ ] = config->tiles.quic.quic_transaction_listen_port;
+      ports[ num_ports++ ] = config->tiles.quic.regular_transaction_listen_port;
+      if( config->is_firedancer ) {
+        ports[ num_ports++ ] = config->gossip.port;
+        ports[ num_ports++ ] = config->tiles.repair.repair_intake_listen_port;
+        ports[ num_ports++ ] = config->tiles.repair.repair_serve_listen_port;
+        ports[ num_ports++ ] = config->tiles.send.send_src_port;
+      }
```
ripatel-fd, line 227: This table is present twice in the same file. Can we dedup it?

src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -126,104 +130,181 @@ init( config_t const * config ) {
...
+  if( FD_UNLIKELY( &ioc != fd_ethtool_ioctl_init( &ioc, device ) ) )
+    FD_LOG_ERR(( "error configuring network device, unable to init ethtool ioctl" ));
+
+  /* This should happen first, otherwise changing the number of channels may fail */
+  fd_ethtool_ioctl_rxfh_set_default( &ioc );
+
+  fd_ethtool_ioctl_channels_set_num( &ioc, 0 /* max */ );
+
+  fd_ethtool_ioctl_ntuple_clear( &ioc );
+  fd_ethtool_ioctl_feature_set( &ioc, FD_ETHTOOL_FEATURE_NTUPLE, 0 );
```
ripatel-fd: Not sure we need to disable the ethtool feature outright. Since sometimes it's enabled by default and it's sort of harmless anyways if there are no flow steering rules.

src/app/shared/commands/configure/ethtool-channels.c
amass-jump: I think I will save this for when we get around to centrally defining this port list and deduping across all the other files

PR #6302, opened 2025-09-10
snapshots: fix possible assert failure when abandoning download

mmcgee-jump:


src/discof/restore/fd_snaprd_tile.c
```diff
@@ -683,6 +683,8 @@ after_frag( fd_snaprd_tile_t *  ctx,
           break;
         case FD_SNAPRD_STATE_FLUSHING_FULL_HTTP:
         case FD_SNAPRD_STATE_FLUSHING_INCREMENTAL_HTTP:
+          if( FD_UNLIKELY( ctx->malformed ) ) break;
```
cali-jumptrading, line 686: Not sure I understand why we need this?

mmcgee-jump: We can get two malformed messages at the same time, one from snapdc and one from snapin, which would cause us to invalidate the peer twice.

cali-jumptrading (approved):

PR #6303, opened 2025-09-10
snapshots: give snapin exclusive non-ht core

mmcgee-jump:


src/disco/topo/fd_topob.c
```diff
@@ -388,6 +388,7 @@ fd_topob_auto_layout( fd_topo_t * topo,
   char const * CRITICAL_TILES[] = {
     "pack",
     "poh",
+    "snapin", /* TODO: Snapshot loading speed depends on having full core */
```
ripatel-fd, line 391: uhhh ... does that fix the IRQ issue? 
I'm almost certain that snapin is currently memory latency bottlenecked and has very low IPC (i.e. the code is slow enough that it uses few hw resources)

mmcgee-jump:
About 20%

src/disco/topo/fd_topob.c
mmcgee-jump: This empirically improves speed by 15-20%, it's unrelated to the IRQ issue.

cali-jumptrading (approved):

src/disco/topo/fd_topob.c
ripatel-fd: Ok good to know, worth investigating more though

PR #6307, opened 2025-09-10
Nuke groove

ripatel-fd:


ripatel-fd:
Although this module won't be used, we decided to keep it in-tree, since it's functional

PR #6308, opened 2025-09-10
ethtool: support additional drivers for dedicated mode

amass-jump:
- Fix rxfh_get_table on ice devices (simpler ioctl bugged)
- Fix a minor bug in fini() on i40e devices
- Fix a minor bug in creating ntuple rules on bnxt devices.
- Also prevent a "not undone" error log in simple-mode fini()

### DIFF
--- a/src/app/shared/commands/configure/fd_ethtool_ioctl.h
+++ b/src/app/shared/commands/configure/fd_ethtool_ioctl.h
@@ -30,7 +30,7 @@
 
 #include "../../../../util/fd_util_base.h"
 
-#define FD_ETHTOOL_MAX_RXFH_TABLE_SIZE (2048)
+#define FD_ETHTOOL_MAX_RXFH_TABLE_SIZE (32768)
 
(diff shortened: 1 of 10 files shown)


src/app/shared/commands/configure/fd_ethtool_ioctl.h
```diff
@@ -30,7 +30,7 @@
 
 #include "../../../../util/fd_util_base.h"
 
-#define FD_ETHTOOL_MAX_RXFH_TABLE_SIZE (2048)
+#define FD_ETHTOOL_MAX_RXFH_TABLE_SIZE (4096)
```
ripatel-fd: I'd rather just make this giant (few megabytes?) since we have the memory to spare

amass-jump: Not sure about megabytes... but I will increase the size of all of these max constants

amass-jump:
Tested init/check/fini for simple & dedicated on bnxt/mlx5/ice/i40e and all are working as expected

mmcgee-jump (approved):

PR #6309, opened 2025-09-10
runtime: custom constructors for banks [wip, don't review]

ibhatt-jumptrading:
eliminates ~20mb of memcpy each time a bank is created


ibhatt-jumptrading:
fixed in another pr

PR #6312, opened 2025-09-10
ignore: test whether fixtures pass when strictly enforcing feature set

0xalpharush:


0xalpharush:
I was attempting to compare to [this](https://github.com/firedancer-io/solfuzz-agave/pull/354#issuecomment-3275858844). It would be a bit strange if the fixtures are passing for features that are not yet implemented, but it's possible the comments are out of date. 

The failure before it exited early is due to a cleaned up feature, so it didn't reach the intrs fixtures
https://github.com/firedancer-io/firedancer/blob/3e270a5bd9f23847dee240f0b5c623ac7e7ff6d5/src/flamenco/features/feature_map.json#L105

PR #6316, opened 2025-09-11
VM memory hashing support for differential fuzzing

hrvoje-osec:
Implements hashing of vm memory regions after instruction execution. Exposed `sol_compat_get_memory_log` and `sol_compat_reset_memory_log` so the fuzzing harness can use them.

`fd_txn_harness.c` will return an additional account with a static key containing hashes of vm memory in its data, this is used for differential fuzzing against agave to discover potential VM memory differences.

### DIFF
--- a/config/extra/with-fuzz.mk
+++ b/config/extra/with-fuzz.mk
@@ -2,6 +2,8 @@ include config/extra/with-handholding.mk
 
 FD_HAS_FUZZ:=1
 
+CPPFLAGS+=-DFD_HAS_FUZZ=1
+
 CPPFLAGS+=-fno-omit-frame-pointer
 CPPFLAGS+=-fsanitize=fuzzer-no-link
 CPPFLAGS+=-fsanitize-coverage=inline-8bit-counters
--- a/config/extra/with-fuzz.mk
+++ b/config/extra/with-fuzz.mk
@@ -2,8 +2,6 @@ include config/extra/with-handholding.mk
 
 FD_HAS_FUZZ:=1
 
-CPPFLAGS+=-DFD_HAS_FUZZ=1
-
 CPPFLAGS+=-fno-omit-frame-pointer
 CPPFLAGS+=-fsanitize=fuzzer-no-link
 CPPFLAGS+=-fsanitize-coverage=inline-8bit-counters
--- a/src/flamenco/runtime/program/fd_bpf_loader_program.c
+++ b/src/flamenco/runtime/program/fd_bpf_loader_program.c
@@ -92,6 +92,60 @@ calculate_heap_cost( ulong heap_size, ulong heap_cost ) {
   #undef KIBIBYTE_MUL_PAGES_SUB_1
 }
 
+#ifdef FD_HAS_FUZZ
+
+static __thread ulong MEMORY_LOG_LENGTH = 0;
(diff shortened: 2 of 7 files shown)


src/flamenco/runtime/program/fd_bpf_loader_program.c
```diff
@@ -95,6 +95,60 @@ calculate_heap_cost( ulong heap_size, ulong heap_cost ) {
   #undef KIBIBYTE_MUL_PAGES_SUB_1
 }
 
+#ifdef FD_HAS_FUZZ
+
+static __thread ulong MEMORY_LOG_LENGTH = 0;
+static __thread ulong MEMORY_LOG[MAX_MEMORY_LOG_LENGTH] = { 0 };
```
ripatel-fd:
```suggestion
static FD_TL ulong MEMORY_LOG[MAX_MEMORY_LOG_LENGTH] = { 0 };
```

src/flamenco/runtime/program/fd_bpf_loader_program.h
```diff
@@ -101,6 +101,18 @@ fd_directly_invoke_loader_v3_deploy( fd_exec_slot_ctx_t * slot_ctx,
...
 
+#ifdef FD_HAS_FUZZ
+
+#define MAX_MEMORY_LOG_LENGTH   64UL
+
+void
+sol_compat_get_memory_log( uchar * dst, ulong * dst_len );
+
+void
+sol_compat_reset_memory_log( void );
```
ripatel-fd: I'd prefer putting the memory region output hashes into the "effects" protobuf instead of adding a side channel. Makes it much easier to diff

config/extra/with-fuzz.mk
```diff
@@ -2,6 +2,8 @@ include config/extra/with-handholding.mk
 
 FD_HAS_FUZZ:=1
 
+CPPFLAGS+=-DFD_HAS_FUZZ=1
```
ripatel-fd: Let's try to avoid conditional compilation for simplicity

src/flamenco/runtime/tests/fd_txn_harness.c
```diff
@@ -610,6 +615,38 @@ fd_solfuzz_txn_run( fd_solfuzz_runner_t * runner,
...
+    if( memory_log_count>0UL ) {
+      out_acct->data =
+        FD_SCRATCH_ALLOC_APPEND( l, alignof( pb_bytes_array_t ),
+                                    PB_BYTES_ARRAY_T_ALLOCSIZE( memory_log_count*8UL ) );
+        if( FD_UNLIKELY( _l > output_end ) ) {
+          abort();
+        }
+        out_acct->data->size = (pb_size_t)memory_log_count*8UL;
+        fd_memcpy( out_acct->data->bytes, memory_log, memory_log_count*8UL );
+    }
```
ripatel-fd: Let's make this strongly typed instead of overloading accounts

ripatel-fd (changes requested):
Some thoughts on the design below. Not asking you to implement these suggestions to save you time, happy to do the work myself. Thank you for the contribution

cmoyes-jump:
Think we can consider merging this if it's opt-in only / default off. Worried about the execs/sec perf impact of introducing a bulk hash into the tight fuzzing loop.

ripatel-fd:
@mjain-jump @cmoyes-jump any interest in picking this back up?

PR #6321, opened 2025-09-11
runtime: fix solfuzz failure crash

ripatel-fd:
Instruction context lifecycle managed is not correctly designed.
If an execution failure occurs during context setup, a partially
initialized object gets generated.  When destroying this object,
null pointer dereferences etc occur.

Additionally, "funk handholding" checks are just entirely broken
and result in the opposite of what they are supposed to achieve:
Enabling 'handholding' silently *hides* bugs from the user instead
of revealing them.

This patch is not a full fix but fixes some smaller issues:

- funk: crash if txn_prepare or txn_cancel is called with obviously
  invalid inputs (instead of only crashing if handholding is NOT
  enabled)

- solfuzz: crash instead of silently failing (and segfaulting) for
  various alloc failure cases in the instruction fuzzer

- solfuzz: fix null-pointer-deref if instruction fuzz context fails
  to bootstrap early (NULL txn provided to funk_txn_cancel)

### DIFF
--- a/src/funk/test_funk_txn.c
+++ b/src/funk/test_funk_txn.c
@@ -191,9 +191,7 @@ main( int     argc,
       uint idx; RANDOM_SET_BIT_IDX( ~live_pmap );
       fd_funk_txn_t * txn = fd_funk_txn_query( &recent_xid[idx], map );
       FD_TEST( !txn );
-#ifdef FD_FUNK_HANDHOLDING
-      FD_TEST( fd_funk_txn_cancel( funk, txn, verbose )==0UL );
-#endif
+      // FD_TEST( fd_funk_txn_cancel( funk, txn, verbose )==0UL );
       break;
     }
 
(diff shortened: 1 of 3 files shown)


src/funk/test_funk_txn.c
```diff
@@ -191,9 +191,7 @@ main( int     argc,
       uint idx; RANDOM_SET_BIT_IDX( ~live_pmap );
       fd_funk_txn_t * txn = fd_funk_txn_query( &recent_xid[idx], map );
       FD_TEST( !txn );
-#ifdef FD_FUNK_HANDHOLDING
-      FD_TEST( fd_funk_txn_cancel( funk, txn, verbose )==0UL );
-#endif
+      // FD_TEST( fd_funk_txn_cancel( funk, txn, verbose )==0UL );
```
topointon-jump, line 194: did you mean to leave these in?

ripatel-fd: Yes, these crash when commented back in

PR #6324, opened 2025-09-11
fddev/dump: add realtime / continuous mode

amass-jump:
- Also fix some issues in tango.lua wireshark dissector

### DIFF
--- a/src/app/shared/fd_action.h
+++ b/src/app/shared/fd_action.h
@@ -59,8 +59,9 @@ union fdctl_args {
   } txn;
 
   struct {
-    char link_name[ 64UL ];
+    char link_name[ 128UL ];
     char pcap_path[ 256UL ];
+    int  once;
   } dump;
 
   struct {
--- a/src/app/shared_dev/commands/dump.c
+++ b/src/app/shared_dev/commands/dump.c
@@ -1,8 +1,32 @@
 #include "../../shared/fd_config.h"
 #include "../../shared/fd_action.h"
+#include "../../../disco/metrics/fd_metrics.h"
 #include "../../../util/net/fd_pcap.h"
 
+#include <errno.h>
+#include <fcntl.h>
+#include <signal.h>
 #include <stdio.h>
+#include <unistd.h>
+
+struct dump_ctx {
+  fd_io_buffered_ostream_t ostream;
+  ulong                    pending_sz;
+
+  struct link {
+    fd_topo_link_t const * topo;
+    void const *           dcache_base;
+    uint                   link_hash;
+  }                        links[ FD_TOPO_MAX_LINKS ];
+  ulong                    link_cnt;
+
+  ulong *                  metrics_base;
+  long                     next_stat_log;
+  ulong                    last_frags;
+  ulong                    last_bytes;
+  ulong                    last_overrun_frags;
+};
+typedef struct dump_ctx dump_ctx_t;
 
 void
 dump_cmd_args( int      * argc,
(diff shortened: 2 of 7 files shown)


amass-jump:
I ran this version against main in once mode and verified the pcaps are identical

```
Files branch.pcap and main.pcap are identical
```

src/app/shared_dev/commands/dump.c
```diff
@@ -66,48 +93,254 @@ dump_link( void           * out_file,
...
+     this fragment could be overrun at any point during this function.
+     So we copy the relevant data into local memory (in the buffered
+     ostream peek space) and only commit it to the file later in
+     after_frag.  after_frag is only called if the fragment was not
+     overrun while we were processing it. */
+
+  ctx->pending_sz = 0UL;
 
-    void const * buffer = fd_chunk_to_laddr_const( mem, chunk );
+  if( FD_UNLIKELY( fd_frag_meta_ctl_err( ctl ) ) ) return;
```
amass-jump: Thoughts on having this line here?

ripatel-fd: I'd rather just dump out the errored frag unless the `chunk` pointer is invalid in that case

src/app/shared_dev/commands/dump.c
```diff
@@ -66,48 +93,254 @@ dump_link( void           * out_file,
...
-      }
-      void * mem = config->topo.workspaces[ config->topo.objs[ link->dcache_obj_id ].wksp_id ].wksp;
+    fd_frag_meta_t const * mcaches[ FD_TOPO_MAX_LINKS ];
+    for( ulong i=0UL; i<ctx.link_cnt; i++ ) {
+      mcaches[ i ] = ctx.links[ i ].topo->mcache;
+    }
 
-      dump_link( out, link, mem );
+    ulong const fseq_footprint = fd_ulong_align_up( fd_fseq_footprint(), fd_fseq_align() );
+    uchar * fseq_mem = fd_alloca( FD_FSEQ_ALIGN, ctx.link_cnt * fseq_footprint );
```
ptaffet-jump: It is fine to use fd_alloca because the dump command only compiles with `FD_HAS_LINUX`, which implies `FD_HAS_ALLOCA`, but it feels a bit unnecessary to me. This could be
```C
uchar fseq_mem[ FD_TOPO_MAX_LINKS ][ FD_FSEQ_FOOTPRINT ] __attribute__((aligned(FD_FSEQ_ALIGN)));
```

src/app/shared_dev/commands/dump.c
```diff
@@ -66,48 +93,254 @@ dump_link( void           * out_file,
...
+    ulong * fseqs[ FD_TOPO_MAX_LINKS ];
+    for( ulong i=0UL; i<ctx.link_cnt; i++ ) {
+      fseqs[ i ] = fd_fseq_join( fd_fseq_new( fseq_mem + (i * fseq_footprint), 0UL ) );
+    }
+
+    fd_rng_t * rng = fd_rng_join( fd_rng_new( fd_alloca( FD_RNG_ALIGN, fd_rng_footprint() ), (uint)fd_tickcount(), 0UL ) );
```
ptaffet-jump: And for the rng, we definitely can just allocate it on the stack without an alloca. That's the typical pattern, e.g. https://github.com/firedancer-io/firedancer/blob/40204c72d7fe2772000f8ad22ec37ed082ee9006/src/app/shared/commands/monitor/monitor.c#L283-L284

src/app/shared_dev/commands/dump.c
```diff
@@ -66,48 +93,254 @@ dump_link( void           * out_file,
...
+
+    uchar * scratch = fd_alloca( FD_STEM_SCRATCH_ALIGN, stem_scratch_footprint( ctx.link_cnt, 0UL, 0UL ) );
+
+    ctx.metrics_base = fd_metrics_join( fd_metrics_new( fd_alloca( FD_METRICS_ALIGN, FD_METRICS_FOOTPRINT( ctx.link_cnt, 0UL ) ), ctx.link_cnt, 0UL ) );
```
ptaffet-jump: Eh, I guess you need these, so maybe the alloca is actually cleaner for `fseq_mem`. I'd say it's not for the rng though. Do whatever you think is best.

src/app/shared/fd_action.h
```diff
@@ -61,6 +61,7 @@ union fdctl_args {
   struct {
     char link_name[ 64UL ];
```
ptaffet-jump: Could you make this slightly larger while you're at it? 128?

ptaffet-jump (approved):
Great!

ptaffet-jump (approved):

PR #2804, opened 2024-08-27
Add singly linked list

ripatel-fd:
Moving fd_quic linked list implementations to util/tmpl.

- **util: fix typos in dlist**
- **util: add fd_list singly linked list**

PR #6325, opened 2025-09-11
util: add slist, a singly linked pool-backed list

ptaffet-jump:
Also has a drive-by fix for fd_map_chain


ripatel-fd:
for reference https://github.com/firedancer-io/firedancer/pull/2804

PR #6328, opened 2025-09-11
net: default to dedicated rss queue for firedancer

mmcgee-jump:


ripatel-fd:
Would probably update the documentation comment but I see that it's full Firedancer only

PR #6334, opened 2025-09-12
gui: sortable viewport template

jherrera-jump:


src/disco/gui/fd_gui_live_table_tmpl.c
```diff
@@ -0,0 +1,798 @@
...
+  ulong cnt = LIVE_TABLE_(ele_cnt)( join );
+  FD_TEST( LIVE_TABLE_(active_sort_key_cnt)( join ) >= 1UL );
+  FD_TEST( join->activity_timers_nanos[ LIVE_TABLE_DEFAULT_TREAP_IDX ]!=LONG_MAX );
+#endif
+  /* remove from all active treaps */
+  for( ulong i=0; i<LIVE_TABLE_MAX_SORT_KEY_CNT; i++ ) {
+    if( FD_LIKELY( join->activity_timers_nanos[ i ] == LONG_MAX ) ) continue;
+    LIVE_TABLE_(private_active_sort_key_idx) = i;
+    LIVE_TABLE_(private_treap_idx_remove)( join->treaps[ i ], pool_idx, pool );
+    join->activity_timers_nanos[ i ] = fd_log_wallclock();
```
mmcgee-jump: This activity scheme is pretty bizarre. Why does this belong in the template?

mmcgee-jump: You certainly cannot do fd_log_wallclock on every insert/delete of a high performance structure like this

src/disco/gui/fd_gui_live_table_tmpl.c
```diff
@@ -0,0 +1,798 @@
+/*
+Generate prototypes, inlines and implementations for tabular viewports
+with a bounded compile-time number of columns and a bounded run-time
+fixed row capacity.
+
+A tabular viewport stores a collections of views into some underlying
```
mmcgee-jump: I don't see code for viewports here, this just looks like the sorted table?

src/disco/gui/fd_gui_live_table_tmpl.c
```diff
@@ -0,0 +1,798 @@
...
+compression for cache and memory bandwidth efficiency, etc.
+
+Typical usage:
+
+    struct myrow {
+      ulong col1;
+      uint col2;
+
+      struct {
+        ulong parent  [ MY_TABLE_MAX_SORT_KEY_CNT ];
```
mmcgee-jump: This seems cleaner (and more cache friendly) to do as
```
struct {
  ulong parent;
  ulong left;
   ...
} treaps[ MY_TABLE_MAX_SORT_KEY_CNT ];
```

src/disco/gui/fd_gui_live_table_tmpl.c
```diff
@@ -0,0 +1,798 @@
...
+
+      } treaps; // technically LIVE_TABLE_TREAP
+      ulong sort_keys; // technically LIVE_TABLE_SORT_KEYS
+    };
+    typedef struct myrow myrow_t;
+
+    static int col1_lt( void const * a, void const * b ) { return *(ulong *)a < *(ulong *)b; }
+    static int col2_lt( void const * a, void const * b ) { return *(uint  *)a < *(uint  *)b; }
+
+    ... by setting LIVE_TABLE_GC_INTERVAL_NANOS to a non-zero value,
```
mmcgee-jump: Yeah I don't understand the purpose of this. This functionality belongs in the caller of this structure, not the template itself

src/disco/gui/fd_gui_live_table_tmpl.c
jherrera-jump: Was planning on using the fd_clock api, but did think about it and agree the activity scheme is not a good approach. Instead of adding an LRU cache I just offload the responsibility to the caller

src/disco/gui/fd_gui_live_table_tmpl.c
jherrera-jump: Yea makes sense.

My reason for doing this originally was that the template is already aware of the "age" of each sort key since this is how it maintains its LRU cache for them, so adding this was a tiny extension. But I'll be removing the "activity" scheme anyways after thinking about it.

mmcgee-jump (approved):

PR #6336, opened 2025-09-12
gossip: contact info add alpenglow socket tag

ravyu-jump:


src/flamenco/gossip/crds/fd_crds_contact_info.c
```diff
@@ -38,10 +38,7 @@ fd_crds_contact_info_init( fd_gossip_view_crds_value_t const * view,
    fd_gossip_view_socket_t const * socket_view = &ci_view->sockets[ i ];
     ushort socket_tag = socket_view->key;
     if( FD_UNLIKELY( socket_tag>FD_CONTACT_INFO_SOCKET_LAST ) ) {
-        /* FIXME: We should treat this a corrupted packet, but we
-           see a bunch of contact infos in testnet that enter this
-           branch. Should investigate before making this a return
-           instead of continue. */
+      FD_LOG_WARNING(( "Unrecognized socket tag: %hu", socket_tag ));
```
mmcgee-jump: Can you fix the underlying problem? These contact infos should not be accepted. Do we still see these warnings? And it's not valid to log warning in response to network traffic, it's a dos vector. This should be covered by metrics instead

mmcgee-jump (changes requested):
.

src/flamenco/gossip/crds/fd_crds_contact_info.c
ravyu-jump: > Can you fix the underlying problem?

The underlying problem is additions to the Contact Info spec weren't communicated to us. Rather than dropping such contact infos, we should operate in a "backwards compatible" mode that only parses the information known to us. If not Agave could release an update that adds a new socket tag and we'd be dropping contact infos for a majority of the cluster. This only works if changes to the spec are simply additions and not updates to existing tags, which Anza has been diligent about so far (since it would break older clients running on the cluster).

> These contact infos should not be accepted

Agave doesn't drop these, they also "continue" on unrecognized socket keys. I realize that we do not update the port offsets when they do, which led to incorrect port offset calculations that this PR attempted to fix. I'll update that.

> And it's not valid to log warning in response to network traffic, it's a dos vector

Ah ok, let me fix that.

ravyu-jump: > Do we still see these warnings?

Just checked, yes we still see socket tag `14` entries, no idea where they come from. Could be someone's custom implementation or someone testing something.

ravyu-jump: >Do we still see these warnings?

> Just checked, yes we still see socket tag 14 entries, no idea where they come from. Could be someone's custom implementation or someone testing something.

Update: these are from invalidator machines

src/flamenco/gossip/crds/fd_crds_contact_info.c
```diff
@@ -18,7 +18,8 @@ typedef struct fd_crds_contact_info_entry fd_crds_contact_info_entry_t;
 void
 fd_crds_contact_info_init( fd_gossip_view_crds_value_t const * view,
                            uchar const *                       payload,
-                           fd_contact_info_t *                 ci ) {
+                           fd_contact_info_t *                 ci,
+                           fd_crds_metrics_t *                 metrics ) {
```
mmcgee-jump: Don't pass the whole metrics in here, just some `ulong * out_unrecognized_socket_cnt` or whatever

src/flamenco/gossip/crds/fd_crds.h
```diff
@@ -37,6 +37,11 @@ struct fd_crds_metrics {
   ulong purged_cnt;
   ulong purged_expired_cnt;
   ulong purged_evicted_cnt;
+
+  struct {
+    ulong unrecognized_socket_tag;
+    ulong ipv6_address;
+  } ci_insert_events;
```
mmcgee-jump: No need for this imo, clearer with just one metrics struct

mmcgee-jump: Wait why are these on the CRDS in the first place? These should be on `fd_gossip` they are related to parsing, not crds

mmcgee-jump (changes requested):
.

src/disco/metrics/metrics.xml
```diff
@@ -872,6 +872,9 @@ metric introduced.
     <counter name="CrdsPurgedExpiredCount" label="The number of purged entries expired from the purged list due to age" />
     <counter name="CrdsPurgedEvictedCount" label="The number of purged entries evicted from the data store to make space for new entries" />
 
+    <counter name="CrdsContactInfoUnrecognizedSocketTags" label="The number of unrecognized socket tags seen in Contact Infos" />
```
mmcgee-jump: These aren't CRDS related so strip the prefix

src/discof/gossip/fd_gossip_tile.c
```diff
@@ -193,6 +193,7 @@ metrics_write( fd_gossip_tile_ctx_t * ctx ) {
   FD_MCNT_SET(   GOSSIP, CRDS_PURGED_EVICTED_COUNT, crds_metrics->purged_evicted_cnt );
   FD_MCNT_SET(   GOSSIP, CRDS_PURGED_EXPIRED_COUNT, crds_metrics->purged_expired_cnt );
 
+
```
mmcgee-jump: Random whitespace

src/flamenco/gossip/fd_gossip_types.h
```diff
@@ -103,7 +104,7 @@ struct fd_contact_info {
   fd_ip4_port_t sockets[ FD_CONTACT_INFO_SOCKET_LAST+1UL ];
 
   struct {
-    uchar  client; /* Any uchar in [0, 255], although typically one of FD_CONTACT_INFO_VERSION_CLIENT_* indicating the self-reported client version */
+    ushort client; /* Any uchar in [0, 255], although typically one of FD_CONTACT_INFO_VERSION_CLIENT_* indicating the self-reported client version */
```
mmcgee-jump: Why changed?

src/flamenco/gossip/fd_gossip_private.h
```diff
@@ -270,7 +264,7 @@ struct fd_gossip_view_pull_request {
   ulong mask;
   uint  mask_bits;
 
-  fd_gossip_view_crds_value_t contact_info[ 1 ];
+  fd_gossip_view_crds_value_t pr_ci[ 1 ];
```
mmcgee-jump, line 267: Why changed?

ravyu-jump: Wanted to distinguish from the nested `contact_info` field otherwise you'd see something like `contact_info->ci_view->contact_info`

src/discof/gossip/fd_gossvf_tile.c
```diff
@@ -576,78 +576,52 @@ check_duplicate_instance( fd_gossvf_tile_ctx_t *   ctx,
...
-      fd_gossip_pingreq_t * pingreq = (fd_gossip_pingreq_t*)fd_chunk_to_laddr( ctx->out->mem, ctx->out->chunk );
-      fd_memcpy( pingreq->pubkey.uc, payload+value->pubkey_off, 32UL );
-      fd_stem_publish( stem, 0UL, fd_gossvf_sig( value->contact_info->addrs[ socket->index ].ip4, fd_ushort_bswap( port ), 1 ), ctx->out->chunk, sizeof(fd_gossip_pingreq_t), 0UL, 0UL, 0UL );
-      ctx->out->chunk = fd_dcache_compact_next( ctx->out->chunk, sizeof(fd_gossip_pingreq_t), ctx->out->chunk0, ctx->out->wmark );
+ping_if_unponged_contact_info( fd_gossvf_tile_ctx_t *    ctx,
+                               fd_contact_info_t const * contact_info,
+                               fd_stem_context_t *       stem ) {
+  fd_ip4_port_t const *     gossip_addr  = &contact_info->sockets[ FD_CONTACT_INFO_SOCKET_GOSSIP ];
+
+  if( FD_UNLIKELY( gossip_addr->l==0U ) ) return 1; /* implies ipv6 address */
```
mmcgee-jump, line 601: Was this a bug, shouldn't this be returning 0?

mmcgee-jump (changes requested):
.

src/discof/gossip/fd_gossvf_tile.c
ravyu-jump: The idea is to drop/filter contact infos that are unpingable, so return non-zero here. Though ipv6 addrs are a problem for us and not Agave so that might be an issue.

src/flamenco/gossip/fd_gossip_types.h
ravyu-jump: `client` is encoded as a u16 varint

src/discof/gossip/fd_gossvf_tile.c
```diff
@@ -576,78 +576,52 @@ check_duplicate_instance( fd_gossvf_tile_ctx_t *   ctx,
...
-
-    if( FD_UNLIKELY( !is_ping_active( ctx, value, payload ) ) ) {
-      fd_gossip_pingreq_t * pingreq = (fd_gossip_pingreq_t*)fd_chunk_to_laddr( ctx->out->mem, ctx->out->chunk );
-      fd_memcpy( pingreq->pubkey.uc, payload+value->pubkey_off, 32UL );
-      fd_stem_publish( stem, 0UL, fd_gossvf_sig( value->contact_info->addrs[ socket->index ].ip4, fd_ushort_bswap( port ), 1 ), ctx->out->chunk, sizeof(fd_gossip_pingreq_t), 0UL, 0UL, 0UL );
-      ctx->out->chunk = fd_dcache_compact_next( ctx->out->chunk, sizeof(fd_gossip_pingreq_t), ctx->out->chunk0, ctx->out->wmark );
+ping_if_unponged_contact_info( fd_gossvf_tile_ctx_t *    ctx,
+                               fd_contact_info_t const * contact_info,
+                               fd_stem_context_t *       stem ) {
+  fd_ip4_port_t const *     gossip_addr  = &contact_info->sockets[ FD_CONTACT_INFO_SOCKET_GOSSIP ];
```
mmcgee-jump, line 599:
```suggestion
  fd_ip4_port_t const * gossip_addr  = &contact_info->sockets[ FD_CONTACT_INFO_SOCKET_GOSSIP ];
```

mmcgee-jump (approved):

mmcgee-jump (approved):

PR #6337, opened 2025-09-12
leader: scaffolding for full leader pipeline + topology hygiene

mmcgee-jump:


src/discof/bank/fd_bank_err.h
```diff
@@ -0,0 +1,125 @@
...
+#define FD_BANK_TXN_ERR_INVALID_ADDRESS_LOOKUP_TABLE_INDEX        -12 /* The transaction references an account offset from the ALUT which does not exist */
+#define FD_BANK_TXN_ERR_MAX_LOADED_ACCOUNTS_DATA_SIZE_EXCEEDED    -13 /* The total account data size of the loaded accounts exceeds the consensus limit */
+#define FD_BANK_TXN_ERR_DUPLICATE_INSTRUCTION                     -14 /* A compute budget program instruction was invoked more than once */
+#define FD_BANK_TXN_ERR_INVALID_LOADED_ACCOUNTS_DATA_SIZE_LIMIT   -15 /* The compute budget program was invoked and set the loaded accounts data size to zero */
+
+/* Preflight errors during replay.  These are errors in validation
+   before we begin executing the transaction, which can only occur
+   during replay, as such transactions do not make it to execution when
+   we are leader. */
+#define FD_BANK_TXN_ERR_ACCOUNT_IN_USE                            -16 /* The transaction conflicts with another transaction in the microblock. TODO: No longer possible with smart dispatcher */
```
ptaffet-jump, line 36: It should say this is not possible as of `SIMD-0083`. Technically, until the feature gate for that flag gets activated, we need to check this before the dispatcher, but I don't think we're planning on doing that

src/discof/bank/fd_bank_err.h
```diff
@@ -0,0 +1,125 @@
...
+#define FD_BANK_TXN_ERR_ACCOUNT_LOADED_TWICE                      -17 /* The transaction references the same account twice */
+#define FD_BANK_TXN_ERR_SIGNATURE_FAILURE                         -18 /* The transaction had an invalid signature */
+#define FD_BANK_TXN_ERR_TOO_MANY_ACCOUNT_LOCKS                    -19 /* The transaction references too many accounts. TODO: No longer possible with smart dispatcher */
```
ptaffet-jump, line 39: I think this too is possible, even with the new dispatcher, and we need to add a check for it.  If a transaction tries to load >64 accounts, I think this is the right error

PR #6341, opened 2025-09-12
discof,rpcserver:  put the rent epoch back into the payload

jumpsiegel:


src/discof/rpcserver/fd_block_to_json.c
```diff
@@ -894,7 +894,7 @@ fd_account_to_json( fd_webserver_t * ws,
   fd_base58_encode_32((uchar*)metadata->owner, 0, owner);
   char addr[50];
   fd_base58_encode_32(acct.uc, 0, addr);
-  fd_web_reply_sprintf(ws, "\",\"%s\"],\"executable\":%s,\"lamports\":%lu,\"owner\":\"%s\",\"address\":\"%s\",\"space\":%lu}",
+  fd_web_reply_sprintf(ws, "\",\"%s\"],\"executable\":%s,\"lamports\":%lu,\"owner\":\"%s\",\"address\":\"%s\",\"rentEpoch\":0,\"space\":%lu}",
```
ripatel-fd: shouldn't this be ULONG_MAX

jumpsiegel: Honestly, unclear.  What is less likely to cause issues with random web browsers.  I will switch it to ulong_max and test

PR #6344, opened 2025-09-12
repair: refactor and clean 🧹

emwang-jump:


src/discof/repair/fd_repair.c
```diff
@@ -175,20 +196,21 @@ fd_repair_sign_ed25519( void * ctx, fd_repair_msg_t * msg, uchar * sig_out ) {
...
     preimage_pong( &msg->pong, preimage );
     fd_sha256_hash( preimage, sizeof(FD_REPAIR_PONG_PREIMAGE_PREFIX) - 1 + sizeof(fd_hash_t), msg->pong.hash.uc );
-    fd_keyguard_client_sign( client, msg->pong.sig, preimage, fd_repair_sz( msg ) - sizeof(fd_ed25519_sig_t), FD_KEYGUARD_SIGN_TYPE_SHA256_ED25519 );
+    //fd_keyguard_client_sign( client, msg->pong.sig, preimage, fd_repair_sz( msg ) - sizeof(fd_ed25519_sig_t), FD_KEYGUARD_SIGN_TYPE_SHA256_ED25519 );
     return;
   }
   case FD_REPAIR_KIND_SHRED:
   case FD_REPAIR_KIND_HIGHEST_SHRED:
   case FD_REPAIR_KIND_ORPHAN: {
-    fd_keyguard_client_sign( client, (uchar *)msg + sizeof(uint) /* sig */, preimage_req( msg, preimage ), fd_repair_sz( msg ) - sizeof(fd_ed25519_sig_t), FD_KEYGUARD_SIGN_TYPE_ED25519 );
```
lidatong: don't forget to clean these up

src/discof/repair/fd_repair_tile.c
```diff
@@ -233,7 +264,7 @@ struct fd_repair_tile_ctx {
 
   ulong turbine_slot0;  // catchup considered complete after this slot
 };
-typedef struct fd_repair_tile_ctx fd_repair_tile_ctx_t;
+typedef struct fd_repair_tile_ctx ctx_t;
```
lidatong:
```suggestion
typedef struct ctx ctx_t;
```

src/discof/repair/fd_repair_tile.c
```diff
@@ -247,40 +278,100 @@ loose_footprint( fd_topo_tile_t const * tile FD_PARAM_UNUSED ) {
...
+  l = FD_LAYOUT_APPEND( l, fd_repair_align(),    fd_repair_footprint   ()                                       );
+  l = FD_LAYOUT_APPEND( l, fd_forest_align(),    fd_forest_footprint   ( tile->repair.slot_max )                );
+  l = FD_LAYOUT_APPEND( l, fd_policy_align(),    fd_policy_footprint   ( FD_NEEDED_KEY_MAX, FD_ACTIVE_KEY_MAX ) );
+  l = FD_LAYOUT_APPEND( l, fd_inflights_align(), fd_inflights_footprint()                                       );
+  l = FD_LAYOUT_APPEND( l, fd_fec_sig_align(),   fd_fec_sig_footprint  ( 20 )                                   );
+  l = FD_LAYOUT_APPEND( l, fd_reasm_align(),     fd_reasm_footprint    ( 1 << 20 )                              );
+  l = FD_LAYOUT_APPEND( l, fd_catchup_align(),   fd_catchup_footprint  ()                                       );
+  l = FD_LAYOUT_APPEND( l, fd_repair_pending_sign_pool_align(), fd_repair_pending_sign_pool_footprint( FD_REPAIR_PENDING_SIGN_REQ_MAX ) );
+  l = FD_LAYOUT_APPEND( l, fd_repair_pending_sign_map_align(),  fd_repair_pending_sign_map_footprint ( FD_REPAIR_PENDING_SIGN_REQ_MAX ) );
```
lidatong: this can prob just be called `signs` if it's local to the tile and doesn't need a pool

src/discof/repair/fd_repair_tile.c
```diff
@@ -247,40 +278,100 @@ loose_footprint( fd_topo_tile_t const * tile FD_PARAM_UNUSED ) {
...
+
+     We want to sign over
+     [ discriminant ] [ payload ]
+      ^                ^
+      buffer           buffer+4
+   */
+
+  fd_memcpy( buffer+64, buffer, 4 );
+  buffer += 64UL;
+  len    -= 64UL;
```
lidatong: i would re-implement and document this more clearly inside `fd_repair.c` (think you can take parts of my impl for that?)

src/discof/repair/fd_repair_tile.c
```diff
@@ -604,139 +490,106 @@ during_frag( fd_repair_tile_ctx_t * ctx,
...
+  if( FD_UNLIKELY( !repair_peer.addr || !repair_peer.port ) ) return;
+  fd_policy_peer_t const * peer = fd_policy_add_peer( ctx->policy, &contact_info->pubkey, &repair_peer );
+  if( peer ) {
+    /* The repair process uses a Ping-Pong protocol that incurs one
+       round-trip time (RTT) for the initial repair request. To optimize
+       this, we proactively send a placeholder Repair request as soon as we
+       receive a peer's contact information for the first time, effectively
+       prepaying the RTT cost. */
+    out_ctx_t * sign_out = sign_avail_credits( ctx );
+    if( FD_UNLIKELY( !sign_out ) ) return;
```
lidatong: is `return` here correct? does it need to save it?

lidatong (commented):
nice work so far. general comment is the control flow in the tile should be restructured to support the fact that sign tiles might be backed up, but we don't wanna lose the pings if no signs are available

(same thing with other request types policy is trying to make)

the `reasm` is responsible for
1. reassembling FEC sets to the correct order per fork
2. repairing missing shreds in FECs when there is late data

we don't want to block 1 because 2 is stuck waiting for sign tiles, so structure should reflect that


it's possible the new `returnable_frag` might be the right API

src/discof/repair/fd_repair_tile.c
```diff
@@ -329,261 +420,56 @@ send_packet( fd_repair_tile_ctx_t * ctx,
...
-    pending->buflen = fd_repair_sign_and_send( ctx, &protocol, &peer->addr, pending->buf, sizeof(pending->buf), 1, pending->nonce, sign_out );
+fd_repair_send_request_async( ctx_t                 * ctx,
+                              out_ctx_t             * sign_out,
+                              fd_repair_msg_t const * msg ){
+  /* Acquire and add a pending request from the pool */
+  fd_repair_pending_sign_t * pending = pending_sign_request_insert( ctx, msg /*, now, recipient */);
+  if( FD_UNLIKELY( !pending ) ) {
+    FD_LOG_INFO(( "No free pending sign requests" ));
+    return;
+  }
```
lidatong, line 434: structure so that it puts in a queue and tries again alter

src/discof/repair/fd_repair_tile.c
emwang-jump: its not necessarily critical that we send out the initial req because it initiates the ping, so the first actual req we send later would initiate the ping if we drop it here
but also switched back to using keyguard client to sign here bc it seems to overwhelm the asynchronous sign tiles initially

src/discof/repair/fd_repair_tile.c
emwang-jump: ill do this in a follow up PR

lidatong (approved):

PR #6346, opened 2025-09-12
ethtool: improve simple/dedicated mode init/check/fini interactions

amass-jump:


src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -69,51 +69,57 @@ device_read_slaves( char const * device,
...
   if( FD_UNLIKELY( ret != 0 ) ) {
     if( FD_LIKELY( ret == EBUSY ) )
-      FD_LOG_ERR(( "error configuring network device, ioctl(SIOCETHTOOL,ETHTOOL_SCHANNELS) failed (%i-%s). "
+      FD_LOG_ERR(( "error configuring network device, failed to set number of channels. "
                    "This is most commonly caused by an issue with the Intel ice driver on certain versions "
                    "of Ubuntu.  If you are using the ice driver, `sudo dmesg | grep %s` contains "
                    "messages about RDMA, and you do not need RDMA, try running `rmmod irdma` and/or "
-                   "blacklisting the irdma kernel module.",
-                   ret, fd_io_strerror( ret ), device ));
+                   "blacklisting the irdma kernel module.", device ));
```
ripatel-fd: Can we keep the errno and EBUSY in the message? Reduces operator overhead by a tiny bit. In my head, this bug is filed under "ice EBUSY bug"

src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -69,51 +69,57 @@ device_read_slaves( char const * device,
...
-      FD_TEST( 0==fd_ethtool_ioctl_ntuple_set_udp_dport( &ioc, rule_idx++, config->tiles.repair.repair_intake_listen_port, 0 ) );
-      FD_TEST( 0==fd_ethtool_ioctl_ntuple_set_udp_dport( &ioc, rule_idx++, config->tiles.repair.repair_serve_listen_port, 0 ) );
-      FD_TEST( 0==fd_ethtool_ioctl_ntuple_set_udp_dport( &ioc, rule_idx++, config->tiles.send.send_src_port, 0 ) );
+      error |= ( 0!=fd_ethtool_ioctl_ntuple_set_udp_dport( &ioc, rule_idx++, config->gossip.port, 0 ) );
+      error |= ( 0!=fd_ethtool_ioctl_ntuple_set_udp_dport( &ioc, rule_idx++, config->tiles.repair.repair_intake_listen_port, 0 ) );
+      error |= ( 0!=fd_ethtool_ioctl_ntuple_set_udp_dport( &ioc, rule_idx++, config->tiles.repair.repair_serve_listen_port, 0 ) );
+      error |= ( 0!=fd_ethtool_ioctl_ntuple_set_udp_dport( &ioc, rule_idx++, config->tiles.send.send_src_port, 0 ) );
     }
+    if( FD_UNLIKELY( error ) )
+      FD_LOG_ERR(( "error configuring network device, failed to install ntuple rules. Try `net.xdp.rss_queue_mode=\"simple\"`" ));
```
ripatel-fd: Can we make sure the following is logged?

- which step failed (RXFH or ntuple?)
- an approximate rendering of the syscall that failed (e.g. `ioctl(...,{args...}`) 
- what device this failed on

Extra step to make this super nice: 
Log the kernel version, device name, and the kernel module in use (you can get that stuff from sysfs but that's "separate PR" type stuff)

ripatel-fd: The existing code might already resolve the 3 above bullet points, just double checking

src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -160,87 +180,86 @@ check_device( char const * device,
...
     }
-    CONFIGURE_OK();
-  } else {
-    int error = 0;    /* is anything not fully configured */
-    int modified = 0; /* is anything changed from the default (fini'd) state */
+  }
 
-    if( FD_UNLIKELY( channels.current != channels.max ) ) {
+  uint rxfh_table[ FD_ETHTOOL_MAX_RXFH_TABLE_SIZE ] = { 0 };
+  uint rxfh_table_size;
```
ripatel-fd: Per convention we'd call this `ele_cnt`, but doesn't matter

src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -160,87 +180,86 @@ check_device( char const * device,
...
-    uint configured_queue = 1; /* If properly configured the 0th queue is excluded */
+    uint configured_queue = dedicated_mode ? 1 : 0;
     for( uint j=0u; j<rxfh_table_size; j++) {
       modified   |= (rxfh_table[ j ] != default_queue++);
       rxfh_error |= (rxfh_table[ j ] != configured_queue++);
       if( default_queue >= channels.current )
         default_queue = 0;
       if( configured_queue >= channels.current )
-        configured_queue = 1;
+        configured_queue = dedicated_mode ? 1 : 0;
```
ripatel-fd, line 208:
```suggestion
        configured_queue = !!dedicated_mode;
```

src/app/shared/commands/configure/ethtool-channels.c
amass-jump: The `fd_ethtool_ioctl` layer has standardized WARNING logging if any underlying `ioctl` call fails. Then the outer `ethtool-channels` code has higher-level logging if steps in general fail. These have to be put together for full context. Does this seem good to you?

For example:

```
NOTICE  09-15 16:51:12.808428 2509693 f0   main src/app/shared/commands/configure/fd_ethtool_ioctl.c(139): RUN: `ethtool --set-rxfh-indir lo default`
WARNING 09-15 16:51:12.808440 2509693 f0   main src/app/shared/commands/configure/fd_ethtool_ioctl.c(27): error configuring network device (lo), ioctl(SIOCETHTOOL,ETHTOOL_GCHANNELS) failed (95-ENOTSUP-operation not supported)
ERR     09-15 16:51:12.808711 2509693 f0   main src/app/shared/commands/configure/ethtool-channels.c(103): error configuring network device, failed to isolate queue zero. Try `net.xdp.rss_queue_mode="simple"`
```

src/app/shared/commands/configure/ethtool-channels.c
amass-jump: I wrote it this way as logically `dedicated_mode` is a flag, whereas `1 : 0` is an index / number. I generally find it to be somewhat more confusing when these are used interchangeably. I'm happy to change this practice/pattern though if this is unusual

src/app/shared/commands/configure/ethtool-channels.c
amass-jump: This will be logged in a WARNING right above this ERR by the standardized ioctl logging. Do you think we need to keep EBUSY in both?

ripatel-fd: No, sounds great

PR #6352, opened 2025-09-12
quic_trace: full client support

akhinvasara-jumptrading:
This PR extends quic_trace for the full client (particularly send_tile). It simplifies the rebasing logic and fixes various quic_trace bugs.

### DIFF
--- a/src/app/shared_dev/commands/quic_trace/fd_quic_trace_main.c
+++ b/src/app/shared_dev/commands/quic_trace/fd_quic_trace_main.c
@@ -3,41 +3,40 @@
    The goal of quic-trace is to tap QUIC traffic on a live system, which
    requires encryption keys and other annoying connection state.
 
-   quic-trace does this by tapping into the shared memory segments of an
-   fd_quic_tile running on the same host.  It does so strictly read-only
+   quic-trace does this by tapping into the shared memory segments of a
+   target tile running on the same host.  It does so strictly read-only
    to minimize impact to a production system.
 
    This file (fd_quic_trace_main.c) provides the glue code required to
-   join remote fd_quic_tile objects.
+   join remote target tile objects.
 
    fd_quic_trace_rx_tile.c provides a fd_tango consumer for incoming
    QUIC packets. */
 
 #include "fd_quic_trace.h"
 
-#include "../../../shared/fd_config.h"
 #include "../../../../disco/metrics/fd_metrics.h"
 #include "../../../../disco/quic/fd_quic_tile.h"
+#include "../../../../discof/send/fd_send_tile.h"
 #include "../../../../waltz/quic/log/fd_quic_log_user.h"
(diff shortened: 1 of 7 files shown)


src/app/shared_dev/commands/quic_trace/fd_quic_trace_main.c
```diff
@@ -383,7 +382,7 @@ quic_trace_cmd_fn( args_t *   args,
 }
 
 action_t fd_action_quic_trace = {
-  .name          = "quic-trace",
+  .name          = "quic_trace",
```
nbridge-jump: There are lots of actions with hyphens. I prefer `quic-trace`

src/app/shared_dev/commands/quic_trace/fd_quic_trace_rx_tile.c
```diff
@@ -110,41 +111,36 @@ fd_quic_trace_initial( fd_quic_trace_ctx_t * trace_ctx,
...
-            if( entry ) {
-              entry->conn_idx = conn->conn_idx;
-            }
+          /* NULL entry either implies already exists, or full */
+          if( entry ) {
+            entry->conn_idx = conn->conn_idx;
           }
         }
+      } else {
+        FD_LOG_CRIT(( "conn out of bounds in fd_quic_trace_initial for conn id %lu with key idx %u", conn_id, key_idx ));
```
nbridge-jump: At least put an override for this. This is a diagnostics tool... It's better to trace as much as it can than just crash

src/app/shared_dev/commands/quic_trace/fd_quic_trace_rx_tile.c
```diff
@@ -355,11 +348,11 @@ fd_quic_trace_1rtt( fd_quic_trace_ctx_t * trace_ctx,
               "\"flow\": \"%s\", "
               "\"trace_time\": \"%s\", "
               "\"src_ip_addr\": \"" FD_IP4_ADDR_FMT "\", "
-              "\"src_udp_port\": %u, "
+              "\"src_udp_port\": \"%u\", "
```
nbridge-jump, line 351: These are numbers. Why wrap them in quotes?

akhinvasara-jumptrading: to make this lowly err msg consistent with the presumably 'correct' version in fd_quic_pretty_print_quic_pkt
do you think I should change them both to leave off the quotes?

nbridge-jump: I didn't realize another was different. This is fine. I might take a look to make everything follow the same policy. But for now, at least this makes them consistent.

nbridge-jump (approved):

PR #6354, opened 2025-09-12
snapshots: add https

cali-jumptrading:


src/app/firedancer/topology.c
```diff
@@ -878,10 +939,20 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
     for( ulong j=0UL; j<peers_cnt; j++ ) {
       if( FD_UNLIKELY( !config->firedancer.snapshots.sources.http.peers[ j ].enabled ) ) continue;
 
-      if( FD_UNLIKELY( 0==resolve_peer( config->firedancer.snapshots.sources.http.peers[ j ].url, &tile->snaprd.http.peers[ resolved_peers_cnt ] ) ) ) {
+      fd_ip4_port_t resolved_addrs[ 8UL ]; /* TODO: this is an arbitrary bound */
+      struct addrinfo hints = { .ai_family = AF_INET, .ai_socktype = SOCK_STREAM };
```
amass-jump, line 1146: I could be wrong but given that we are setting `service` to NULL when calling getaddrinfo, and generally not using the port information (we just care about getting all ipv4 addresses), I don't think ai_socktype=SOCK_STREAM does anything and I think we can omit the hint stuff altogether

src/app/firedancer/topology.c
```diff
@@ -878,10 +939,20 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
     for( ulong j=0UL; j<peers_cnt; j++ ) {
       if( FD_UNLIKELY( !config->firedancer.snapshots.sources.http.peers[ j ].enabled ) ) continue;
 
-      if( FD_UNLIKELY( 0==resolve_peer( config->firedancer.snapshots.sources.http.peers[ j ].url, &tile->snaprd.http.peers[ resolved_peers_cnt ] ) ) ) {
+      fd_ip4_port_t resolved_addrs[ 8UL ]; /* TODO: this is an arbitrary bound */
```
amass-jump: 8 seems like a reasonable number of addrs to try for the same server. In fact maybe we should do less...

src/app/firedancer/topology.c
```diff
@@ -878,10 +939,20 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
...
+                                       tile->snaprd.http.peers[ resolved_peers_cnt ].hostname,
+                                       resolved_addrs,
+                                       8UL,
+                                       &tile->snaprd.http.peers[ resolved_peers_cnt ].is_https );
+      if( FD_UNLIKELY( 0==num_resolved ) ) {
         FD_LOG_ERR(( "failed to resolve address of [snapshots.sources.http.peers] entry \"%s\"", config->firedancer.snapshots.sources.http.peers[ j ].url ));
       } else {
-        resolved_peers_cnt++;
+        for( ulong i=0UL; i<(ulong)num_resolved; i++ ) tile->snaprd.http.peers[ resolved_peers_cnt+i ].addr = resolved_addrs[ i ];
+        resolved_peers_cnt += (ulong)num_resolved;
```
amass-jump: this only has room for 16 peers, I don't think we are checking the bounds? Plus if we can do up to 8 IPs per configured endpoint, we may want more than 16

src/app/firedancer/topology.c
```diff
@@ -145,51 +148,109 @@ resolve_address( char const * address,
...
+
+  freeaddrinfo( res );
+  return resolved;
+}
+
+static int
+resolve_peer( char const *            peer,
+              struct addrinfo const * addr_resolve_hints,
+              char const *            config_str,
+              char                    hostname[ static 256UL ],
```
amass-jump, line 254: we should probably define a hostname size constant and use it everywhere...

src/app/firedancer/topology.c
```diff
@@ -145,51 +148,109 @@ resolve_address( char const * address,
...
-  ulong fqdn_len = (ulong)( colon-host_port );
+  ulong fqdn_len = (ulong)( host_end-host_port );
   if( FD_UNLIKELY( fqdn_len>254 ) ) {
-    FD_LOG_ERR(( "invalid [gossip.entrypoints] entry \"%s\": hostname too long", host_port ));
+    FD_LOG_ERR(( "invalid [%s] entry \"%s\": hostname too long", config_str, host_port ));
   }
   fd_memcpy( fqdn, host_port, fqdn_len );
+  fd_memcpy( hostname, fqdn, fqdn_len );
   fqdn[ fqdn_len ] = '\0';
+  hostname[ fqdn_len ] = '\0';
```
amass-jump, line 285: couldn't we just use hostname rather than memcpy'ing twice?

src/app/firedancer/topology.c
```diff
@@ -145,51 +148,109 @@ resolve_address( char const * address,
...
+  if( FD_LIKELY( colon ) ) {
+    char const * port_str = host_end+1;
+    char const * endptr   = NULL;
+    ulong port = strtoul( port_str, (char **)&endptr, 10 );
+    if( FD_UNLIKELY( !endptr || !port || port>USHORT_MAX || *endptr!='\0' ) ) {
+      FD_LOG_ERR(( "invalid [%s] entry \"%s\": invalid port number", config_str, host_port ));
+    }
+    for( ulong i=0UL; i<(ulong)resolved; i++ ) ip4_port[ i ].port = (ushort)fd_ushort_bswap( (ushort)port );
+  } else if( FD_LIKELY( !colon && https ) ) {
+    for( ulong i=0UL; i<(ulong)resolved; i++ ) ip4_port->port = (ushort)fd_ushort_bswap( 443U );
```
amass-jump: ooc isn't this already returning a ushort? why cast?

src/app/firedancer/topology.c
```diff
@@ -145,51 +148,109 @@ resolve_address( char const * address,
...
+  } else {
+    for( ulong i=0UL; i<(ulong)resolved; i++ ) ip4_port->port = (ushort)fd_ushort_bswap( USHORT_MAX );
```
amass-jump: should we not FD_LOG_ERR here? isn't that what we were doing before if there was no valid port number?

src/discof/restore/utils/fd_ssping.h
```diff
@@ -53,8 +60,9 @@ fd_ssping_join( void * shping );
    being tracked, trying to add a new peer is a no-op. */
 
 void
-fd_ssping_add( fd_ssping_t * ssping,
-               fd_ip4_port_t addr );
+fd_ssping_add( fd_ssping_t *            ssping,
+               fd_ip4_port_t            addr,
+               fd_sspeer_meta_t const * meta );
```
amass-jump: this could probably just be a `void * ctx` or something like that, right? ssping doesn't need to use meta, it just needs to store it for the client?

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -26,8 +36,10 @@ struct fd_sshttp_private {
 
   int   hops;
 
-  fd_ip4_port_t addr;
-  int           sockfd;
+  fd_ip4_port_t            addr;
+  fd_sspeer_meta_t const * meta;
+  int                      is_https;
```
amass-jump: isn't is_https redundant w/ meta->is_https?

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -104,27 +126,86 @@ fd_sshttp_join( void * shhttp ) {
...
 }
 
+#if FD_HAS_OPENSSL
+static void
+http_init_ssl( fd_sshttp_t *            http,
+               fd_sspeer_meta_t const * meta ) {
+  FD_TEST( meta && meta->hostname && meta->hostname_len );
+  FD_TEST( !http->ssl_ctx && !http->ssl );
+
+  SSL_CTX * ssl_ctx = SSL_CTX_new( TLS_client_method() );
```
amass-jump: generally w/ openssl you are supposed to init the context once and then reuse it. There is also other init stuff missing, like SSL_library_init() for example. This is a pretty good example https://wiki.openssl.org/index.php/SSL/TLS_Client

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -104,27 +126,86 @@ fd_sshttp_join( void * shhttp ) {
...
+
+  http->ssl = SSL_new( http->ssl_ctx );
+  if( FD_UNLIKELY( !http->ssl ) ) {
+    FD_LOG_ERR(( "SSL_new failed" ));
+  }
+
+  uchar const alpn_protos[] = { 8, 'h', 't', 't', 'p', '/', '1', '.', '1' };
+  int alpn_res = SSL_set_alpn_protos( http->ssl, alpn_protos, sizeof(alpn_protos) );
+  if( FD_UNLIKELY( alpn_res!=0 ) ) {
+    FD_LOG_ERR(( "SSL_set_alpn_protos failed (%d)", SSL_get_error( http->ssl, alpn_res ) ));
```
amass-jump: is this required? I have never needed to do this before for basic HTTPS requests

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -36,6 +48,11 @@ struct fd_sshttp_private {
   ulong response_len;
   char  response[ USHORT_MAX ];
 
+#if FD_HAS_OPENSSL
```
amass-jump, line 62: what happens if someone builds without openssl but then adds an https endpoint to their config?

src/discof/restore/utils/fd_ssping.h
cali-jumptrading: generally not a fan of using void *, since it's unclear what type the API accepts.

src/app/firedancer/topology.c
cali-jumptrading: I needed `ai_socktype=SOCK_STREAM` to avoid getting duplicate ip addresses from getaddrinfo

src/app/firedancer/topology.c
cali-jumptrading: agreed

src/discof/restore/utils/fd_sshttp.c
cali-jumptrading: I think we don't have to since http/1.1 is the default... but it also doesn't hurt...

src/discof/restore/utils/fd_sshttp.c
cali-jumptrading: got it thanks, I wasn't sure if the SSL_ctx was supposed to be reused or not.

src/discof/restore/utils/fd_sshttp.c
cali-jumptrading: I had some FD_LOG_ERRs if `FD_HAS_OPENSSL` was not enabled but the peer was an https peer.

src/app/firedancer/topology.c
cali-jumptrading: yeah good catch!

src/app/firedancer/topology.c
cali-jumptrading: fixed! Now snaprd supports up to 32 http peers

src/app/firedancer/topology.c
cali-jumptrading: let's not introduce unecessary changes and make hostname constant change in a separate PR

src/app/firedancer/topology.c
cali-jumptrading: Ah oops, good catch!

src/app/firedancer/topology.c
cali-jumptrading: yeah FD_LOG_ERR makes more sense here!

src/discof/restore/utils/fd_sshttp.c
cali-jumptrading: Yeah it kind of is redundant, I just didn't want to keep checking whether the `fd_sspeer_meta_t` pointer was NULL or not.

src/discof/restore/fd_snapct_tile.seccomppolicy
```diff
@@ -8,13 +8,27 @@ uint logfile_fd, uint dir_fd, uint out_full_fd, uint out_inc_fd, uint ping_fd
...
+# written to include the out descriptors for the full and incremental
+# snapshots and file descriptors used by openssl.  We will
+# restrict this being called on the snapshot directory file descriptor
+# and the ping file descriptor.
+write: (not (or (eq (arg 0) dir_fd)
+                (eq (arg 0) ping_fd)))
+
+# openssl: read
+#
+# Openssl requires read to read encryption key files and certificates
```
ripatel-fd: this is not true

src/discof/restore/fd_snapct_tile.seccomppolicy
```diff
@@ -8,13 +8,27 @@ uint logfile_fd, uint dir_fd, uint out_full_fd, uint out_inc_fd, uint ping_fd
...
+# when executing the TLS handshake and when encrypting/decrypting data
+# sent over TLS.  We will restrict this being called on the stderr,
+# the log file descriptor, the snapshot directory, the full and
+# incremental snapshot files, and the ping file descriptor.
+read: (not (or (eq (arg 0) 2)
+               (eq (arg 0) logfile_fd)
+               (eq (arg 0) dir_fd)
+               (eq (arg 0) out_full_fd)
+               (eq (arg 0) out_inc_fd)
+               (eq (arg 0) ping_fd)))
```
ripatel-fd, line 30: change to using whitelist

src/discof/restore/fd_snapct_tile.c
ripatel-fd, line 1: missing SSL cert verification

cali-jumptrading: I put the cert loading and verification in `fd_openssl_tile.h` Let me know what you think! `fd_sshttp_new` and `fd_http_resolver_new` both call `load_certs`.

src/disco/bundle/fd_bundle_tile.c
```diff
@@ -246,64 +247,19 @@ fd_bundle_tile_parse_endpoint( fd_bundle_tile_t *     ctx,
...
-
-  ulong old_num = *(ulong *)( (uchar *)addr - 8UL );
-  fd_memcpy( (uchar*)new + 8, (uchar*)addr, fd_ulong_min( old_num, num ) );
-  fd_alloc_free( fd_quic_ssl_mem_function_ctx, (uchar *)addr - 8UL );
-  *(ulong *)new = num;
-  return (uchar*)new + 8UL;
-}
+#define OPENSSL_MEM_FUNCTION_CTX     fd_quic_ssl_mem_function_ctx
+#define OPENSSL_ALLOC_ERROR_CALLBACK fd_ossl_alloc_error_callback
+#include "../../waltz/openssl/fd_openssl_tile.c"
```
ripatel-fd: Could you namespace these defines by FD_? It's unlikely but technically OPENSSL_* stuff could conflict with the OpenSSL library.

src/discof/restore/utils/fd_http_resolver.c
```diff
@@ -11,20 +11,27 @@
...
+#endif
+
 #define PEER_STATE_UNRESOLVED (0)
 #define PEER_STATE_REFRESHING (1)
 #define PEER_STATE_VALID      (2)
 #define PEER_STATE_INVALID    (3)
 
 #define PEER_DEADLINE_NANOS_VALID   (5L*1000L*1000L*1000L) /* 5 seconds */
-#define PEER_DEADLINE_NANOS_RESOLVE (1L*1000L*1000L*1000L) /* 1 second */
+#define PEER_DEADLINE_NANOS_RESOLVE (2L*1000L*1000L*1000L) /* 2 second */
```
ripatel-fd:
```suggestion
#define PEER_DEADLINE_NANOS_RESOLVE (2L*1000L*1000L*1000L) /* 2 seconds */
```

src/discof/restore/utils/fd_http_resolver.c
```diff
@@ -155,6 +166,19 @@ fd_http_resolver_new( void *                           shmem,
...
 
+#if FD_HAS_OPENSSL
+  SSL_CTX * ssl_ctx = SSL_CTX_new( TLS_client_method() );
+  if( FD_UNLIKELY( !ssl_ctx ) ) {
+    FD_LOG_ERR(( "SSL_CTX_new failed" ));
+  }
+
+  SSL_CTX_up_ref( ssl_ctx );
+  resolver->ssl_ctx = ssl_ctx;
+  SSL_CTX_free( ssl_ctx );
```
ripatel-fd: This up_ref / free seems unnecessary, maybe just assign it here and add a comment `transfer ownership of ssl_ctx`?

src/discof/restore/utils/fd_http_resolver.c
```diff
@@ -155,6 +166,19 @@ fd_http_resolver_new( void *                           shmem,
   resolver->cb_arg                     = cb_arg;
   resolver->on_resolve_cb              = on_resolve_cb;
 
+#if FD_HAS_OPENSSL
+  SSL_CTX * ssl_ctx = SSL_CTX_new( TLS_client_method() );
+  if( FD_UNLIKELY( !ssl_ctx ) ) {
+    FD_LOG_ERR(( "SSL_CTX_new failed" ));
+  }
+
```
ripatel-fd, line 174: Here's a good place to pin the min TLS version to 1.2 since TLS 1.1 and below are insecure

src/discof/restore/utils/fd_http_resolver.c
```diff
@@ -76,6 +83,10 @@ struct fd_http_resolver_private {
   void *                           cb_arg;
   fd_http_resolver_on_resolve_fn_t on_resolve_cb;
 
+#if FD_HAS_OPENSSL
+  SSL_CTX* ssl_ctx;
```
ripatel-fd:
```suggestion
  SSL_CTX * ssl_ctx;
```

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -42,6 +59,11 @@ struct fd_sshttp_private {
 
   char  snapshot_name[ PATH_MAX ];
 
+#if FD_HAS_OPENSSL
+  SSL_CTX* ssl_ctx;
+  SSL*     ssl;
```
ripatel-fd:
```suggestion
  SSL_CTX * ssl_ctx;
  SSL *     ssl;
```

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -80,8 +102,24 @@ fd_sshttp_new( void * shmem ) {
...
+  sshttp->ssl_ctx = NULL;
+
+  SSL_CTX * ssl_ctx = SSL_CTX_new( TLS_client_method() );
+  if( FD_UNLIKELY( !ssl_ctx ) ) {
+    FD_LOG_ERR(( "SSL_CTX_new failed" ));
+  }
+
+  SSL_CTX_up_ref( ssl_ctx );
+  sshttp->ssl_ctx = ssl_ctx;
+  SSL_CTX_free( ssl_ctx );
```
ripatel-fd: Same as above, would just add a comment here clarifying this moves ownership

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -109,31 +147,78 @@ fd_sshttp_join( void * shhttp ) {
...
+http_init_ssl( fd_sshttp_t * http ) {
+  FD_TEST( http->hostname );
+  FD_TEST( http->ssl_ctx );
+
+  http->ssl = SSL_new( http->ssl_ctx );
+  if( FD_UNLIKELY( !http->ssl ) ) {
+    FD_LOG_ERR(( "SSL_new failed" ));
+  }
+
+  uchar const alpn_protos[] = { 8, 'h', 't', 't', 'p', '/', '1', '.', '1' };
```
ripatel-fd:
```suggestion
  static uchar const alpn_protos[] = { 8, 'h', 't', 't', 'p', '/', '1', '.', '1' };
```
I don't think SSL_set_alpn_protos keeps a reference to this string, but I'd move it to `.rodata` just to be sure

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -109,31 +147,78 @@ fd_sshttp_join( void * shhttp ) {
...
+  int alpn_res = SSL_set_alpn_protos( http->ssl, alpn_protos, sizeof(alpn_protos) );
+  if( FD_UNLIKELY( alpn_res!=0 ) ) {
+    FD_LOG_ERR(( "SSL_set_alpn_protos failed (%d)", SSL_get_error( http->ssl, alpn_res ) ));
```
ripatel-fd: This usage of SSL_get_error is not correct. SSL_get_error is for when read/write like SSL calls (like SSL_read_ex) fail and returns a code like `SSL_ERR`. But alpn_res is already in that format, so it's redundant

src/disco/bundle/fd_bundle_tile.c
cali-jumptrading: done

src/discof/restore/utils/fd_http_resolver.c
cali-jumptrading: done

src/app/firedancer/config/testnet.toml
```diff
@@ -10,7 +10,7 @@
     max_account_records = 200_000_000
 [snapshots]
     [snapshots.sources]
-        servers = [ "http://solana-testnet-rpc.jumpisolated.com:8899" ]
+        servers = [ "https://solana-testnet-tls.jumpisolated.com" ]
```
amass-jump: Should we change this back before merge? Or do you intend to keep this change?

src/app/firedancer/topology.c
```diff
@@ -214,51 +217,107 @@ resolve_address( char const * address,
...
-  char const * colon = strrchr( host_port, ':' );
-  if( FD_UNLIKELY( !colon ) ) {
-    FD_LOG_ERR(( "invalid [gossip.entrypoints] entry \"%s\": no port number", host_port ));
+  char const * colon    = strrchr( host_port, ':' );
+  char const * host_end = colon;
+  if( FD_LIKELY( FD_UNLIKELY( !colon && !https )  ) ) {
+    FD_LOG_ERR(( "invalid [%s] entry \"%s\": no port number", config_str, host_port ));
+    host_end = colon;
+  } else if( FD_LIKELY( !colon && https ) ) {
+    host_end = host_port + strnlen( host_port, 256UL-(ulong)(host_port-peer) );
```
amass-jump: Why do you limit to 256 here? Your check for fqdn_len below should handle cases where strlen(peer) is >>256

src/app/firedancer/topology.c
```diff
@@ -214,51 +217,107 @@ resolve_address( char const * address,
...
-  char const * port_str = colon+1;
-  char const * endptr   = NULL;
-  ulong port = strtoul( port_str, (char **)&endptr, 10 );
-  if( FD_UNLIKELY( !endptr || !port || port>USHORT_MAX || *endptr!='\0' ) ) {
-    FD_LOG_ERR(( "invalid [gossip.entrypoints] entry \"%s\": invalid port number", host_port ));
+  if( FD_LIKELY( colon ) ) {
+    char const * port_str = host_end+1;
+    char const * endptr   = NULL;
+    ulong port = strtoul( port_str, (char **)&endptr, 10 );
+    if( FD_UNLIKELY( !endptr || !port || port>USHORT_MAX || *endptr!='\0' ) ) {
```
amass-jump: I'm not sure that `endptr` will ever be set to NULL, but checking for `endptr==port_str` is a good check when using stroul to make sure at least 1 digit was consumed

src/app/firedancer/topology.c
```diff
@@ -1089,12 +1148,35 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
...
+                                       "snapshots.sources.servers",
+                                       tile->snapct.sources.servers[ resolved_peers_cnt ].hostname,
+                                       resolved_addrs,
+                                       MAX_RESOLVED_ADDRS,
+                                       &tile->snapct.sources.servers[ resolved_peers_cnt ].is_https );
+      if( FD_UNLIKELY( 0==num_resolved ) ) {
         FD_LOG_ERR(( "[snapshots.sources.servers[%lu] invalid (%s)", i, config->firedancer.snapshots.sources.servers[ i ] ));
+      } else {
+        if( FD_UNLIKELY( resolved_peers_cnt+(ulong)num_resolved>=FD_TOPO_SNAPSHOTS_SERVERS_MAX ) ) {
+          FD_LOG_WARNING(( "ignoring peer `%s` in [[snapshots.sources.http.peers]] because reached maximum supported peers %lu", config->firedancer.snapshots.sources.servers[ i ], FD_TOPO_SNAPSHOTS_SERVERS_MAX ));
```
amass-jump: http.peers ?

amass-jump: Also, we could remove this case entirely by allowing N*M servers in fd_topo_tile_snapct, where N is the number of server strings in fd_config (16 currently) and M is the number of addresses we allow per server (you have it set to 8, but tbh that seems overkill)

In general I think it's much more important to have coverage over all the configured servers than it is to have coverage over multiple IP addresses for the same configured server.

cali-jumptrading: so... you're ok with 32*4 maximum supported servers in snapct?

src/app/firedancer/topology.c
cali-jumptrading: what would be a reasonable limit? In fd_config.h, we limit the size of the server string names to 128 anyways: https://github.com/firedancer-io/firedancer/blob/main/src/app/shared/fd_config.h#L154

src/app/firedancer/config/testnet.toml
cali-jumptrading: changing it back

src/app/firedancer/topology.c
amass-jump: It's not clear to me why we wouldn't just use strlen() here? Maybe I'm missing something

cali-jumptrading: I thought we tried to avoid strlen because of safety? What do you think?

src/discof/restore/fd_snapct_tile.seccomppolicy
cali-jumptrading: Left it as a blacklist for now, let's revisit this with OpenSSL's BIO memory mode.

src/discof/restore/fd_snapct_tile.seccomppolicy
cali-jumptrading: You're right, fixed!

src/app/firedancer/topology.c
ripatel-fd: The caller promises that `peer` is a cstr. I think it's fair to assume in this piece of code that `peer` is zero-terminated. That said, the true limit is max FQDN length (the longest domain name is 253 characters) plus 6 chars for the port number (`:65535`) plus 1 terminating char. So, if you want to use strnlen, use at least 260 chars as a bound.

cali-jumptrading: thanks, I'll just use strlen in that case.

src/discof/restore/fd_snapld_tile.c
```diff
@@ -58,15 +65,33 @@ scratch_footprint( fd_topo_tile_t const * tile FD_PARAM_UNUSED ) {
...
+  l = FD_LAYOUT_APPEND(  l, fd_alloc_align(),           fd_alloc_footprint()     );
   return FD_LAYOUT_FINI( l, scratch_align() );
 }
 
+FD_FN_CONST static inline ulong
+loose_footprint( fd_topo_tile_t const * tile ) {
+  (void)tile;
+  /* Leftover space for OpenSSL allocations */
+  return 1UL<<26UL; /* 64 MiB */
+}
```
amass-jump, line 79: I'm not too familiar with loose_footprint, how is this used?

src/discof/restore/fd_snapld_tile.c
```diff
@@ -286,9 +329,12 @@ returnable_frag( fd_snapld_tile_t *  ctx,
                ctx->state==FD_SNAPSHOT_STATE_FINISHING  ||
                ctx->state==FD_SNAPSHOT_STATE_ERROR );
       if( FD_UNLIKELY( ctx->state!=FD_SNAPSHOT_STATE_FINISHING ) ) {
-        ctx->state = FD_SNAPSHOT_STATE_ERROR;
-        fd_stem_publish( stem, 0UL, FD_SNAPSHOT_MSG_CTRL_ERROR, 0UL, 0UL, 0UL, 0UL, 0UL );
-        return 0;
```
amass-jump, line 291: For a file load or HTTP load, receiving NEXT/DONE while not in the FINISHING state is an error... it seems like we no longer have that check. It could be an error for HTTPS too unless we are specifically doing the async shutdown

amass-jump (dismissed):

src/discof/restore/fd_snapld_tile.c
cali-jumptrading: HTTPS always does the async shutdown. But to explain a bit, I think that receiving NEXT/DONE while not in the finishing state should not be considered an error here. `snapct` doesn't actually know when `snapld` is truly done reading a snapshot. `snapct` makes an educated guess that as soon as it receives all bytes of the snapshot, that it is ok to send the `NEXT/DONE` to `snapld`.

cali-jumptrading: Unless... if `snapld` were to notify `snapct` that it was truly done reading the snapshot, then receiving a `NEXT/DONE` message when not in the finished state would be an error.

src/discof/restore/fd_snapld_tile.c
cali-jumptrading: This is used to allocate some extra memory in the tile's wksp that is outside of `scratch_footprint`. I followed the format in `fd_bundle_tile.c`

src/app/firedancer/topology.c
```diff
@@ -1080,12 +1152,31 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
...
         FD_LOG_ERR(( "[snapshots.sources.servers[%lu] invalid (%s)", i, config->firedancer.snapshots.sources.servers[ i ] ));
+      } else {
+        for( ulong i=0UL; i<(ulong)num_resolved; i++ ) tile->snapct.sources.servers[ resolved_peers_cnt+i ].addr = resolved_addrs[ i ];
+        for( ulong i=1UL; i<(ulong)num_resolved; i++ ) {
+          tile->snapct.sources.servers[ resolved_peers_cnt+i ].is_https = tile->snapct.sources.servers[ resolved_peers_cnt ].is_https;
+          fd_memcpy( tile->snapct.sources.servers[ resolved_peers_cnt+i ].hostname,
+                     tile->snapct.sources.servers[ resolved_peers_cnt ].hostname,
+                     sizeof(tile->snapct.sources.servers[ resolved_peers_cnt ].hostname) );
+        }
+        resolved_peers_cnt += (ulong)num_resolved;
```
amass-jump, line 1164: Maybe I'm missing something, but don't we need to set `servers_cnt` to the new value here so that these are actually used?

src/discof/restore/fd_snapct_tile.c
```diff
@@ -452,9 +471,10 @@ log_download( fd_snapct_tile_t * ctx,
   }
 
   for( ulong i=0UL; i<ctx->config.sources.servers_cnt; i++ ) {
-    if( addr.l==ctx->config.sources.servers[ i ].l ) {
-      FD_LOG_NOTICE(( "downloading %s snapshot at slot %lu from configured server with index %lu at http://" FD_IP4_ADDR_FMT ":%hu/%s",
+    if( addr.l==ctx->config.sources.servers[ i ].addr.l ) {
+      FD_LOG_NOTICE(( "downloading %s snapshot at slot %lu from configured server with index %lu at %s://" FD_IP4_ADDR_FMT ":%hu/%s",
                       full ? "full" : "incremental", slot, i,
+                      ctx->config.sources.servers[ i ].is_https ? "https" : "http",
                       FD_IP4_ADDR_FMT_ARGS( addr.addr ), fd_ushort_bswap( addr.port ),
```
amass-jump: nit: I would log the hostname here if it's HTTPS

src/discof/restore/fd_snapct_tile.c
```diff
@@ -1175,7 +1207,7 @@ privileged_init( fd_topo_t *      topo,
   char full_path[ PATH_MAX ] = {0};
   char incremental_path[ PATH_MAX ] = {0};
   if( FD_UNLIKELY( -1==fd_ssarchive_latest_pair( tile->snapct.snapshots_path,
-                                                 tile->snapct.incremental_snapshots,
+                                                  tile->snapct.incremental_snapshots,
```
amass-jump: 🤔

src/app/firedancer/topology.c
cali-jumptrading: good catch!

src/waltz/openssl/fd_openssl_tile.h
```diff
@@ -0,0 +1,49 @@
...
+                ulong        num,
+                char const * file,
+                int          line );
+
+/* load_certs manually loads certificates into an SSL_CTX object.  This
+   should be called right after calling SSL_CTX_new in privileged init
+   in a tile. */
+
+void
+load_certs( SSL_CTX * ssl_ctx );
```
amass-jump: nit: fd_ossl_load_certs ?

src/disco/topo/fd_topo_run.c
```diff
@@ -86,9 +88,35 @@ fd_topo_run_tile( fd_topo_t *          topo,
...
+#if FD_HAS_OPENSSL
+  /* OpenSSL's CRYPTO_set_mem_functions is a global operation so it can
+     only be called once for all threads/processes. */
+  FD_ONCE_BEGIN {
+    if( FD_UNLIKELY( !CRYPTO_set_mem_functions( crypto_malloc, crypto_realloc, crypto_free ) ) ) {
+      FD_LOG_ERR(( "CRYPTO_set_mem_functions failed" ));
+    }
+  } FD_ONCE_END;
+#endif
+
```
amass-jump: Can these not go in the priv_init of the tiles that actually use openssl, guarded by FD_ONCE ?

src/discof/restore/fd_snapct_tile.c
cali-jumptrading: done

src/discof/restore/fd_snapct_tile.c
cali-jumptrading: done

amass-jump (approved):

src/disco/topo/fd_topo_run.c
cali-jumptrading: so you'd want to copy paste this in all tiles that use openssl? Code duplication would be harder to maintain no?

src/waltz/openssl/fd_openssl_tile.h
cali-jumptrading: sure, that's a better name.

src/disco/topo/fd_topo_run.c
cali-jumptrading: Unless... I guess we could put all of this into some openssl init function...

cali-jumptrading: I actually don't like the idea of moving this code to some function... it becomes unclear where the openssl initialization and cleanup are. The cleanup still has to happen in `fd_topo_run.c` but the initialization is done in the tiles.

src/disco/bundle/fd_bundle_tile.c
```diff
@@ -247,65 +248,6 @@ fd_bundle_tile_parse_endpoint( fd_bundle_tile_t *     ctx,
...
-static FD_TL fd_alloc_t * fd_quic_ssl_mem_function_ctx = NULL;
-
-static void *
-crypto_malloc( ulong        num,
-               char const * file,
-               int          line ) {
-  (void)file; (void)line;
-  void * result = fd_alloc_malloc( fd_quic_ssl_mem_function_ctx, 16UL, num + 8UL );
-  if( FD_UNLIKELY( !result ) ) {
-    FD_MCNT_INC( BUNDLE, ERRORS_SSL_ALLOC, 1UL );
```
mmcgee-jump, line 270: Think you want to remove the metric, or if you keep you need to update it somewhere else

src/waltz/openssl/fd_openssl_tile.c
```diff
@@ -0,0 +1,161 @@
...
+
+void *
+crypto_malloc( ulong        num,
+               char const * file,
+               int          line ) {
+  (void)file;
+  (void)line;
+  void * result = fd_alloc_malloc( g_fd_ossl_alloc, 8UL, num + 8UL );
+  if( FD_UNLIKELY( !result ) ) {
+    FD_LOG_ERR(("OpenSSL allocation of %lu bytes failed at %s:%d", num, file, line));
```
mmcgee-jump: Not sure if it makes sense to ERR here? That might be a DoS vector depending on traffic pattern, and would be a change from what's in the prior impl

src/waltz/openssl/fd_openssl_tile.c
```diff
@@ -0,0 +1,161 @@
...
+    return NULL;
+  }
+  *(ulong*)result = num;
+  return (uchar*)result + 8UL;
+}
+
+void
+crypto_free( void *       addr,
```
mmcgee-jump, line 45: Would make these static and hide it all behind one `openSSL_alloc_install` function or whatever, see `cJSON_alloc_install` for example

src/waltz/openssl/fd_openssl_tile.c
cali-jumptrading: Do we have a place to define global metrics? If multiple tiles are using open SSL, we would need a global open SSL error counter metric, if we wanted to use metrics instead of FD_LOG_ERR.

src/disco/bundle/fd_bundle_tile.c
cali-jumptrading: will do

src/waltz/openssl/fd_openssl_tile.c
cali-jumptrading: got it, will do

src/waltz/openssl/fd_openssl_tile.c
ripatel-fd: @cali-jumptrading you could use thread-locals  for error counters

src/disco/topo/fd_topo_run.c
```diff
@@ -3,6 +3,7 @@
 
 #include "../metrics/fd_metrics.h"
 #include "../../util/tile/fd_tile_private.h"
+#include "../../waltz/openssl/fd_openssl_tile.h"
```
amass-jump: .

book/api/metrics-generated.md
```diff
@@ -1200,6 +1199,5 @@
 | <span class="metrics-name">tower_&#8203;hard_&#8203;forks_&#8203;seen</span> | counter | Number of hard forks we've seen (block ids with multiple candidate bank hashes) |
 | <span class="metrics-name">tower_&#8203;hard_&#8203;forks_&#8203;pruned</span> | counter | Number of hard forks (candidate bank hashes) we've pruned |
 | <span class="metrics-name">tower_&#8203;hard_&#8203;forks_&#8203;active</span> | gauge | Currently active hard forks |
-| <span class="metrics-name">tower_&#8203;hard_&#8203;forks_&#8203;max_&#8203;width</span> | gauge | The max width of hard forks (block id with most candidate bank hashes) we've ever seen |
```
amass-jump, line 1203: Is this a merge conflict of some kind?

src/discof/restore/utils/fd_http_resolver.c
```diff
@@ -240,14 +271,31 @@ peer_connect( fd_http_resolver_t *  resolver,
   resolver->fds_idx[ resolver->fds_len ] = peer_pool_idx( resolver->pool, peer );
   peer->fd.idx = resolver->fds_len;
   resolver->fds_len++;
-  fd_ssresolve_init( peer->full_ssresolve, peer->addr, resolver->fds[ peer->fd.idx ].fd, 1 );
+
+  if( FD_UNLIKELY( peer->is_https ) ) {
+#if FD_HAS_OPENSSL
+    fd_ssresolve_init_https( peer->full_ssresolve, peer->addr, resolver->fds[ peer->fd.idx ].fd, 1, peer->hostname, resolver->ssl_ctx );
+#else
+    FD_LOG_ERR(( "peer requires https but firedancer is built without openssl support" ));
```
amass-jump: nit: log the hostname and ask the operator to remove from their config?

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -413,10 +660,20 @@ fd_sshttp_advance( fd_sshttp_t * http,
...
+#endif
+    case FD_SSHTTP_STATE_REQ:           return send_request( http, now );
+    case FD_SSHTTP_STATE_RESP:          return read_response( http, data_len, data, now );
+    case FD_SSHTTP_STATE_DL:            return read_body( http, data_len, data, now );
+    case FD_SSHTTP_STATE_DONE:
+      fd_sshttp_cancel( http );
+      http->state = FD_SSHTTP_STATE_INIT;
+      return FD_SSHTTP_ADVANCE_DONE;
+      break;
+    default:                            return 0;
```
amass-jump: This should be returning one of the enum codes?

amass-jump (approved):

book/api/metrics-generated.md
cali-jumptrading: Think there's some discrepancy with metrics going on. max_width is not in metrics.xml in main, but somehow still in the generated metrics...

amass-jump (approved):

src/disco/bundle/fd_bundle_tile.c
```diff
@@ -55,6 +56,9 @@ metrics_write( fd_bundle_tile_t * ctx ) {
   FD_MCNT_SET( BUNDLE, ERRORS_PROTOBUF,        ctx->metrics.decode_fail_cnt           );
   FD_MCNT_SET( BUNDLE, ERRORS_TRANSPORT,       ctx->metrics.transport_fail_cnt        );
   FD_MCNT_SET( BUNDLE, ERRORS_NO_FEE_INFO,     ctx->metrics.missing_builder_info_fail_cnt );
+#if FD_HAS_OPENSSL
+  FD_MCNT_SET( BUNDLE, ERRORS_SSL_ALLOC,       g_fd_ossl_alloc_errors                 );
```
ripatel-fd:
```suggestion
  FD_MCNT_SET( BUNDLE, ERRORS_SSL_ALLOC,       fd_ossl_alloc_err_cnt                 );
```

Global symbols should always be namespaced by fd_ not any other prefix (sorry for the nit)

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -149,17 +237,182 @@ fd_sshttp_init( fd_sshttp_t * http,
...
+    http->deadline = now + FD_SSHTTP_DEADLINE_NANOS;
+  }
+}
+
+#if FD_HAS_OPENSSL
+static int
+http_connect_ssl( fd_sshttp_t * http,
+                  long          now ) {
+  if( FD_UNLIKELY( now>http->deadline ) ) {
+    FD_LOG_WARNING(("deadline exceeded during connect"));
```
ripatel-fd, line 254:
```suggestion
    FD_LOG_WARNING(( "deadline while connecting to HTTPS snapshot server" ));
```
Might also want to log the peer URL here

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -149,17 +237,182 @@ fd_sshttp_init( fd_sshttp_t * http,
...
+    return FD_SSHTTP_ADVANCE_ERROR;
+  }
+
+  FD_TEST( http->ssl );
+  SSL_set_fd( http->ssl, http->sockfd );
+  int ssl_err = SSL_connect( http->ssl );
+  if( FD_UNLIKELY( ssl_err!=1 ) ) {
+    int ssl_err_code = SSL_get_error( http->ssl, ssl_err );
+    if( FD_UNLIKELY( ssl_err_code!=SSL_ERROR_WANT_READ && ssl_err_code!=SSL_ERROR_WANT_WRITE ) ) {
+      FD_LOG_WARNING(( "SSL_connect failed (%d)", ssl_err ));
```
ripatel-fd, line 265: These WARNING logs should ideally be operator friendly.

e.g. `error while downloading snapshot from https://: SSL_connect failed (%i-%s)`

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -149,17 +237,182 @@ fd_sshttp_init( fd_sshttp_t * http,
...
+    FD_LOG_WARNING(("deadline exceeded during shutdown"));
+    fd_sshttp_cancel( http );
+    return FD_SSHTTP_ADVANCE_ERROR;
+  }
+
+  int res = SSL_shutdown( http->ssl );
+  if( FD_LIKELY( res<=0 ) ) {
+    int ssl_err_code = SSL_get_error( http->ssl, res );
+    if( FD_UNLIKELY( ssl_err_code!=SSL_ERROR_WANT_READ && ssl_err_code!=SSL_ERROR_WANT_WRITE && res!=0 ) ) {
+      FD_LOG_WARNING(( "SSL_shutdown failed (%d)", ssl_err_code ));
```
ripatel-fd, line 292: same here, warning log shoudl be operator friendly

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -149,17 +237,182 @@ fd_sshttp_init( fd_sshttp_t * http,
...
+static long
+http_recv_ssl( fd_sshttp_t * http,
+               void *        buf,
+               ulong         bufsz ) {
+  int read_res = SSL_read( http->ssl, buf, (int)bufsz );
+  if( FD_UNLIKELY( read_res<=0 ) ) {
+    int ssl_err = SSL_get_error( http->ssl, read_res );
+
+    if( FD_UNLIKELY( ssl_err!=SSL_ERROR_WANT_READ && ssl_err!=SSL_ERROR_WANT_WRITE ) ) {
+      FD_LOG_WARNING(( "SSL_read failed (%d)", ssl_err ));
```
ripatel-fd, line 314: operator friendly warning log

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -149,17 +237,182 @@ fd_sshttp_init( fd_sshttp_t * http,
...
+static long
+http_send_ssl( fd_sshttp_t * http,
+               void *        buf,
+               ulong         bufsz ) {
+  int write_res = SSL_write( http->ssl, buf, (int)bufsz );
+  if( FD_UNLIKELY( write_res<=0 ) ) {
+    int ssl_err = SSL_get_error( http->ssl, write_res );
+
+    if( FD_UNLIKELY( ssl_err!=SSL_ERROR_WANT_READ && ssl_err!=SSL_ERROR_WANT_WRITE ) ) {
+      FD_LOG_WARNING(( "SSL_write failed (%d)", ssl_err ));
```
ripatel-fd, line 333: operator friendly warning log

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -149,17 +237,182 @@ fd_sshttp_init( fd_sshttp_t * http,
...
+           void *        buf,
+           ulong         bufsz ) {
+#if FD_HAS_OPENSSL
+  if( FD_LIKELY( http->is_https ) ) return http_send_ssl( http, buf, bufsz );
+#endif
+
+  long sent = sendto( http->sockfd, buf, bufsz, 0, NULL, 0 );
+  if( FD_UNLIKELY( -1==sent && errno==EAGAIN ) ) return FD_SSHTTP_ADVANCE_AGAIN;
+  else if( FD_UNLIKELY( -1==sent ) ) {
+    FD_LOG_WARNING(( "sendto() failed (%d-%s)", errno, fd_io_strerror( errno ) ));
```
ripatel-fd, line 380: operator friendly warning log

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -149,17 +237,182 @@ fd_sshttp_init( fd_sshttp_t * http,
...
+  if( FD_UNLIKELY( -1==read && errno==EAGAIN ) ) {
+    if( FD_UNLIKELY( ++http->empty_recvs>8UL ) ) {
+      /* If we have gone several iterations without having any data to
+         read, sleep the thread for up to one millisecond, or until
+         the socket is readable again, whichever comes first. */
+      struct pollfd pfd = {
+        .fd = http->sockfd,
+        .events = POLLIN,
+      };
+      fd_syscall_poll( &pfd, 1 /*fds*/, 1 /*ms*/ );
```
ripatel-fd: check error code of syscall (FD_LOG_ERR if poll returns a weird errno)

src/discof/restore/utils/fd_ssresolve.c
```diff
@@ -139,13 +214,32 @@ fd_ssresolve_send_request( fd_ssresolve_t * ssresolve ) {
...
-    return FD_SSRESOLVE_ADVANCE_ERROR;
+  long sent = 0L;
+  if( FD_LIKELY( ssresolve->is_https ) ) {
+#if FD_HAS_OPENSSL
+    int write_res = SSL_write( ssresolve->ssl, ssresolve->request+ssresolve->request_sent, (int)(ssresolve->request_len-ssresolve->request_sent) );
+    if( FD_UNLIKELY( write_res<=0 ) ) {
+      int ssl_err = SSL_get_error( ssresolve->ssl, write_res );
+
+      if( FD_UNLIKELY( ssl_err!=SSL_ERROR_WANT_READ && ssl_err!=SSL_ERROR_WANT_WRITE ) ) {
+        FD_LOG_WARNING(( "SSL_write failed (%d)", ssl_err ));
```
ripatel-fd, line 228: operator friendly error log

src/discof/restore/utils/fd_ssresolve.c
```diff
@@ -202,19 +296,42 @@ fd_ssresolve_parse_redirect( fd_ssresolve_t *        ssresolve,
...
+        FD_LOG_WARNING(( "SSL_read failed (%d)", ssl_err ));
+        return FD_SSRESOLVE_ADVANCE_ERROR;
+      }
+
+      return FD_SSRESOLVE_ADVANCE_AGAIN;
+    }
+
+    read = (long)read_res;
+#else
+    FD_LOG_ERR(( "cannot use HTTPS without OpenSSL" ));
```
ripatel-fd, line 326: this build of Firedancer does not support OpenSSL, rebuild with deps.sh

or something along those lines, otherwise operator is going to be confused why OpenSSL was not fonud

src/discof/restore/utils/fd_ssresolve.c
```diff
@@ -202,19 +296,42 @@ fd_ssresolve_parse_redirect( fd_ssresolve_t *        ssresolve,
...
+#endif
+  } else {
+    read = recvfrom( ssresolve->sockfd, ssresolve->response+ssresolve->response_len, sizeof(ssresolve->response)-ssresolve->response_len, 0, NULL, NULL );
+    if( FD_UNLIKELY( -1==read && errno==EAGAIN ) ) return FD_SSRESOLVE_ADVANCE_AGAIN;
+    else if( FD_UNLIKELY( -1==read ) ) {
+      FD_LOG_WARNING(( "recvfrom() failed (%d-%s)", errno, fd_io_strerror( errno ) ));
```
ripatel-fd, line 332: operator friendly error log

src/discof/restore/utils/fd_ssresolve.c
```diff
@@ -254,10 +371,62 @@ fd_ssresolve_read_response( fd_ssresolve_t *        ssresolve,
...
+#if FD_HAS_OPENSSL
+static int
+ssresolve_connect_ssl( fd_ssresolve_t * ssresolve ) {
+  FD_TEST( ssresolve->ssl );
+  SSL_set_fd( ssresolve->ssl, ssresolve->sockfd );
+  int ssl_err = SSL_connect( ssresolve->ssl );
+  if( FD_UNLIKELY( ssl_err!=1 ) ) {
+    int ssl_err_code = SSL_get_error( ssresolve->ssl, ssl_err );
+    if( FD_UNLIKELY( ssl_err_code!=SSL_ERROR_WANT_READ && ssl_err_code!=SSL_ERROR_WANT_WRITE ) ) {
+      FD_LOG_WARNING(( "SSL_connect failed (%d)", ssl_err_code ));
```
ripatel-fd, line 383: operator friendly error log

src/discof/restore/fd_snapct_tile.c
```diff
@@ -406,7 +419,16 @@ init_load( fd_snapct_tile_t *  ctx,
   fd_ssctrl_init_t * out = fd_chunk_to_laddr( ctx->out_ld.mem, ctx->out_ld.chunk );
   out->file = file;
   out->zstd = !file || (full ? ctx->local_in.full_snapshot_zstd : ctx->local_in.incremental_snapshot_zstd);
-  if( !file ) out->addr = ctx->addr;
+  if( !file ) {
+    out->addr = ctx->addr;
+    for( ulong i=0UL; i<SERVER_PEERS_MAX; i++ ) {
+      if( FD_UNLIKELY( ctx->addr.l==ctx->config.sources.servers[ i ].addr.l ) ) {
+        fd_memcpy( out->hostname, ctx->config.sources.servers[ i ].hostname, 256UL );
```
ripatel-fd: `fd_cstr_ncpy`

ripatel-fd: also ideally use `sizeof(out->hostname)` instead of hardcoding 256

src/discof/restore/fd_snapld_tile.c
```diff
@@ -1,10 +1,13 @@
+#define _GNU_SOURCE
 #include "utils/fd_ssarchive.h"
 #include "utils/fd_ssctrl.h"
 #include "utils/fd_sshttp.h"
 
 #include "../../disco/topo/fd_topo.h"
 #include "../../disco/metrics/fd_metrics.h"
+#include "../../waltz/openssl/fd_openssl_tile.h"
 
+#include <sys/mman.h>
```
ripatel-fd:
```suggestion
#include <sys/mman.h> /* memfd_create */
```

src/waltz/openssl/fd_openssl_tile.c
```diff
@@ -0,0 +1,155 @@
...
+#include "../../util/bits/fd_bits.h"
+#include "fd_openssl_tile.h"
+
+#include <dirent.h>
+#include <unistd.h>
+#include <errno.h>
+
+/* Thread-local alloc object for each tile that uses OpenSSL. */
+FD_TL fd_alloc_t * g_fd_ossl_alloc        = NULL;
+FD_TL ulong        g_fd_ossl_alloc_errors = 0UL;
```
ripatel-fd: remove `g_` prefix, all global syms should be `fd_`

ripatel-fd (approved):

src/waltz/openssl/fd_openssl_tile.c
cali-jumptrading: done

src/discof/restore/fd_snapld_tile.c
cali-jumptrading: done

src/discof/restore/utils/fd_ssresolve.c
cali-jumptrading: I'll address this in a follow up

src/discof/restore/utils/fd_ssresolve.c
cali-jumptrading: I'll address log-related comments in a follow up

ripatel-fd (approved):

src/discof/restore/fd_snapct_tile.c
cali-jumptrading: done

src/disco/bundle/fd_bundle_tile.c
cali-jumptrading: done

src/discof/tower/fd_tower_tile.c
```diff
@@ -189,8 +189,7 @@ metrics_write( ctx_t * ctx ) {
   FD_MCNT_SET( TOWER, HARD_FORKS_SEEN,     ctx->metrics.hard_forks.seen   );
   FD_MCNT_SET( TOWER, HARD_FORKS_PRUNED,   ctx->metrics.hard_forks.pruned );
 
-  FD_MGAUGE_SET( TOWER, HARD_FORKS_ACTIVE,    ctx->metrics.hard_forks.active    );
-  FD_MGAUGE_SET( TOWER, HARD_FORKS_MAX_WIDTH, ctx->metrics.hard_forks.max_width );
```
ibhatt-jumptrading, line 193: was this intended?

cali-jumptrading: yes, charlie forgot to remove this line after he removed the metric in metrics.xml

PR #6356, opened 2025-09-12
make: add parameter to frontend build script

jherrera-jump:
failing trailing whitespace test due to codegen, will have to force-merge


mmcgee-jump:
There is a list of trailing whitespace excluded paths in CI somewhere which you need to update, maybe under `.github` or somewhere

.github/workflows/trailing_whitespace.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - name: Get changed files
         id: changes
         run: |
-          echo "all=$(git diff --name-only --diff-filter=ACMRT ${{ github.event.pull_request.base.sha }} ${{ github.sha }} | xargs)" >> $GITHUB_OUTPUT
+          echo "all=$(git diff --name-only --diff-filter=ACMRT ${{ github.event.pull_request.base.sha }} ${{ github.sha }} -- . ':!src/disco/gui/dist*' | xargs)" >> $GITHUB_OUTPUT
```
mmcgee-jump, line 16: I don't think this is the right file? Because the old files were already excluded somewhere else

mmcgee-jump (approved):

PR #6358, opened 2025-09-13
tower: minor edits for clarity

mmcgee-jump:


src/discof/tower/fd_tower_tile.h
```diff
@@ -1,3 +1,38 @@
...
+   generate an update to the vote state, and potentially advance
+   the root. */
+
+struct fd_tower_slot_done {
+  ulong reset_slot; /* The reset slot is always set and represents the currently fork to build on, it may be unchanged */
+
+  /* Sometimes, finshing replay of a slot may cause a new slot to be
+     rooted.  If this happens, new root will be 1 and both root_slot and
+     root_block_id will be set to the new root values accordingly.
+     Otherwise, new root will be 0. */
```
lidatong:
```suggestion
     Otherwise, new root will be ULONG_MAX. */
```

src/app/firedancer/config/default.toml
```diff
@@ -934,7 +934,7 @@ user = ""
         #   to route interesting packets based on UDP dst port.  This mode
         #   is considered experimental and may not work with some network
         #   device setups. This should not yet be used in production.
-        rss_queue_mode = "simple"
+        rss_queue_mode = "dedicated"
```
lidatong: why change this back again?

src/discof/replay/fd_replay_tile.c
```diff
@@ -1538,9 +1538,25 @@ during_frag( fd_replay_tile_ctx_t * ctx,
...
+      /* We have recieved a root message.  We don't want to update the
+         rooted slot and block id if we are processing the genesis
+         block. */
+      if( FD_UNLIKELY( fd_bank_slot_get( ctx->slot_ctx->bank )==0UL )) return;
+      ctx->consensus_root_slot = msg->root_slot;
+
+      block_id_map_t * block_id = block_id_map_query( ctx->block_id_map, msg->root_slot, NULL );
+      if( FD_UNLIKELY( !block_id ) ) {
+        FD_LOG_CRIT(( "invariant violation: no block id found for root slot %lu", msg->root_slot ));
+      }
```
lidatong, line 1556: use block id in your frag

src/discof/tower/fd_tower_tile.h
```diff
@@ -1,3 +1,39 @@
...
+  /* Sometimes, finshing replay of a slot may cause a new slot to be
+     rooted.  If this happens, new root will be 1 and both root_slot and
+     root_block_id will be set to the new root values accordingly.
+     Otherwise, new root will be 0 and root_slot, root_block_id have
+     undefined values. */
+  int new_root;
+  ulong root_slot;
+  fd_hash_t root_block_id;
+
+  fd_hash_t block_id;
```
lidatong:
```suggestion
  /* Sometimes, finishing replay of a slot may cause a new slot to be
     rooted.  If this happens, new root will be 1 and both root_slot and
     root_block_id will be set to the new root values accordingly.
     Otherwise, new root will be 0 and root_slot, root_block_id have
     undefined values. */

  int       new_root;
  ulong     root_slot;
  fd_hash_t root_block_id;

  fd_hash_t block_id;
```

src/discof/tower/fd_tower_tile.h
```diff
@@ -1,3 +1,42 @@
...
+  fd_hash_t root_block_id;
+
+  /* TODO: This is not currently used but probably should be? */
+  fd_hash_t block_id;
+
+  /* It currently appears to be the case that every slot completion
+     triggers a vote to be sent, so this contains the vote transaction
+     payload and size.  The vote is not yet signed. */
+  ulong vote_txn_sz;
+  uchar vote_txn[ FD_TPU_MTU ];
```
lidatong: this needs a caveat, i never implemented the "refresh last vote" logic for send tile similar to what Agave has for when you don't have anything to vote on, please add a TODO for this

original thought was send would detect its the exact same txn and decide whether to refresh it (send it out) or not, prob a smarter way to do it

src/discof/tower/fd_tower_tile.h
```diff
@@ -1,3 +1,42 @@
...
+
+  /* The slot being voted on.  This is used by the vote sending tile to
+     do some internal book-keeping. */
+  ulong vote_slot;
```
lidatong: move this up with reset slot? also the bookkeeping is targeting your leader for vote lat which is probably worth stating explicitly?

src/discof/tower/fd_tower_tile.h
```diff
@@ -1,3 +1,42 @@
...
+  ulong reset_slot;
+
+  /* Sometimes, finshing replay of a slot may cause a new slot to be
+     rooted.  If this happens, new root will be 1 and both root_slot and
+     root_block_id will be set to the new root values accordingly.
+     Otherwise, new root will be 0 and root_slot, root_block_id have
+     undefined values. */
+  int new_root;
+  ulong root_slot;
+  fd_hash_t root_block_id;
```
lidatong:
```suggestion
  /* Sometimes, finishing replay of a slot may cause a new slot to be
     rooted.  If this happens, new root will be 1 and both root_slot and
     root_block_id will be set to the new root values accordingly.
     Otherwise, new root will be 0 and root_slot, root_block_id have
     undefined values. */

  int       new_root;
  ulong     root_slot;
  fd_hash_t root_block_id;

  fd_hash_t block_id;
```

src/discof/tower/fd_tower_tile.h
```diff
@@ -1,3 +1,42 @@
+#ifndef HEADER_fd_src_discof_tower_fd_tower_tile_h
+#define HEADER_fd_src_discof_tower_fd_tower_tile_h
```
lidatong, line 2: is `.h` the new pattern for frag schemas ? they're all dumped in fd_disco_base rn

src/discof/tower/fd_tower_tile.c
```diff
@@ -1,204 +1,168 @@
...
-    fd_replay_tower_t const * replay_tower = &ctx->replay_towers[i];
+static void
+update_ghost( fd_tower_tile_t * ctx ) {
+  fd_voter_t * epoch_voters = fd_epoch_voters( ctx->epoch );
+  for( ulong i=0UL; i<ctx->replay_towers_cnt; i++ ) {
+    fd_replay_tower_t const * replay_tower = &ctx->replay_towers[ i ];
     fd_pubkey_t const *       pubkey       = &replay_tower->key;
-    fd_voter_state_t const *  voter_state  = (fd_voter_state_t *)fd_type_pun_const( replay_tower->acc );
-    fd_tower_t *              tower        = ctx->vote_towers[i];
+    fd_voter_state_t const *  voter_state  = (fd_voter_state_t *)replay_tower->acc;
```
lidatong: pretty sure you need type pun for strict aliasing ? the cast should be to `(fd_voter_state_t const *)` too (that was already a bug in orig)

src/discof/tower/fd_tower_tile.h
lidatong: ignore this i misread api at first

mmcgee-jump: This isn't right? `new root` is the int value indicating if there was a root

mmcgee-jump: I clarified the comment language a bit

lidatong: yeah i misread when i firstl ooked

src/discof/tower/fd_tower_tile.h
mmcgee-jump: It seems better than dumping stuff in disco_base

src/discof/tower/fd_tower_tile.c
mmcgee-jump: IIRC aliasing with uchar * is not a violation, there's a special carve out

src/discof/tower/fd_tower_tile.h
lidatong: sgtm

src/discof/tower/fd_tower_tile.c
lidatong: the rule is only casts to `uchar *`, the other way isn't technically allowed by the standard

lidatong (approved):

PR #6378, opened 2025-09-15
adds new HANDHOLDING for __builtin_unreachable()

marcograss:
Adds new handholding for `__builtin_unreachable()`.

Replaces `__builtin_unreachable()` with `FD_UNREACHABLE()`.

By default, this is no change: `FD_UNREACHABLE()` is defined as `__builtin_unreachable()`, so no overhead is incurred.

If `FD_USE_HANDHOLDING` is set, `FD_UNREACHABLE()` becomes an `FD_LOG_CRIT` with the code location (since `__builtin_unreachable()` carries no message or location information).

This is because hitting `__builtin_unreachable()` is undefined behavior, so it makes sense to abort directly in HANDHOLDING builds where performance is not the main objective.


two-heart (approved):
lgtm. I think this is valuable to have. Deferring to one of the code owners for final approve + merge.

marcograss:
moved from `FD_LOG_ERR` to `FD_LOG_CRIT`

mmcgee-jump:
Please do not add new `*_USE_HANDHOLDING`, just stick everything behind the same handholding flag. We would never want to mix and match options

marcograss:
@mmcgee-jump ok done, this one will be the first, we have now `FD_USE_HANDHOLDING` and we can  bring the new handholdings under this flag

PR #6385, opened 2025-09-15
quic: heap service queue

akhinvasara-jumptrading:
This PR brings back the service queue. It's been soaking on mainnet val3 and val4 for a few days. Metrics look good. 

(The first commit will be merged separately in https://github.com/firedancer-io/firedancer/pull/6231, but is included here only for ease of review).


src/waltz/quic/fd_quic.h
```diff
@@ -390,6 +373,8 @@ struct fd_quic {
   fd_quic_callbacks_t cb;      /* position-dependent,   reset on join, writable pre init  */
   fd_quic_metrics_t   metrics; /* position-independent, persistent,    read only */
 
+  long now; /* recent timestamp */
```
nbridge-jump: Is there a good reason for moving `now` from state to here?

akhinvasara-jumptrading: no.. there isn't - thanks for catching that. I've fixed this on the original 'simplify time' PR ( https://github.com/firedancer-io/firedancer/pull/6231 ). I'm marking this one as a draft until that one is merged, and then I'll rebase.

nlgripto (approved):

valerio-asymmetric (approved):
LGTM

PR #6386, opened 2025-09-16
replay: fix buffer overflow in scheduler

yufeng-jump:
tested on ledger mainnet-364974841 slot 365125180


ptaffet-jump (approved):
Stamping to unblock. Will review with all the related code later

PR #6388, opened 2025-09-16
stake link: inter-tile toc/tou and oob safety

two-heart:


mmcgee-jump (changes requested):
looks ok but can you rebase

PR #6389, opened 2025-09-16
tower: remove unused rename syscall from seccomppolicy

anwayde:


src/discof/tower/fd_tower_tile.c
```diff
@@ -206,10 +206,6 @@ replay_slot_done( fd_tower_tile_t *       ctx,
   fd_tower_sync_serde_t ser;
   fd_tower_to_tower_sync( ctx->tower, ctx->root, &slot_info->bank_hash, &slot_info->block_id, ctx->ts, &ser ); /* TODO: ctx->ts is uninitialized */
 
-  // TODO: seccomppolicy is wrong, it's allowing `rename(2)`.  Only
```
mmcgee-jump: Well, the code doesn't actually call either `rename` or `renameat` at the moment, since it's commented out. The comment is about changing the *code* (and then the seccomp policy) to use `renameat`.

anwayde: I did not see the `rename` syscall being used anywhere, did not see it commented out anywhere either ( did not see it in the definition of `fd_tower_checkpt`). The reason for this PR is that seccomppolicy breaks the nightly arm64 builds and I saw that `rename` is not currently used anywhere either. Shall I remove it entirely from the seccomppolicy and leave (or modify) the TODO comment?

mmcgee-jump: Ah yeah, just remove it entirely then

PR #6396, opened 2025-09-17
fuzz, instr: add log fuzzing

mjain-jump:


cmoyes-jump (approved):
Open to giving this a shot

PR #6400, opened 2025-09-17
Rewrite core_bpf_migration

ripatel-fd:
Gets rid of funk_txn usage to revert uncomitted changes


ripatel-fd:
Missing unit tests. This feature should never have been merged without unit tests in the first place ...

kbhargava-jump (approved):

PR #6413, opened 2025-09-17
configure: add IRQ affinity stage

amass-jump:


src/app/shared/fd_config.h
```diff
@@ -9,7 +9,6 @@
 
 #define NAME_SZ                          (256UL)
 #define AFFINITY_SZ                      (256UL)
-#define CONFIGURE_STAGE_COUNT            ( 12UL)
```
amass-jump, line 12: Note: This was already wrong AFAICT and only used in one place so I figured it'd be better to just remove it and specify a large bound in fd_action.h.

src/app/shared_dev/commands/pktgen/pktgen.c
```diff
@@ -207,6 +207,8 @@ pktgen_cmd_fn( args_t *   args FD_PARAM_UNUSED,
   configure_stage( &fd_cfg_stage_hugetlbfs,        CONFIGURE_CMD_INIT, config );
   configure_stage( &fd_cfg_stage_ethtool_channels, CONFIGURE_CMD_INIT, config );
   configure_stage( &fd_cfg_stage_ethtool_offloads, CONFIGURE_CMD_INIT, config );
+  configure_stage( &fd_cfg_stage_ethtool_loopback, CONFIGURE_CMD_INIT, config );
+  configure_stage( &fd_cfg_stage_irq_affinity,     CONFIGURE_CMD_INIT, config );
```
amass-jump, line 210: @ripatel-fd thoughts here? Wasn't sure if we wanted to keep a standard set of configure stages or if loopback was intentionally omitted. And whether irq-affinity should be included here?

src/app/fdctl/main.c
```diff
@@ -38,6 +38,7 @@ configure_stage_t * STAGES[] = {
   &fd_cfg_stage_ethtool_channels,
   &fd_cfg_stage_ethtool_offloads,
   &fd_cfg_stage_ethtool_loopback,
+  &fd_cfg_stage_irq_affinity,
```
amass-jump: @mmcgee-jump Should we leave this stage out of fdctl/fddev for now and only use it in firedancer? Then we can add it in later once it's spent some more time on main?

amass-jump:
```
WARNING 10-01 23:38:58.880846 1314696 f0   main src/app/shared/commands/configure/irq-affinity.c(281): Detected irqbalance service (/run/irqbalance/irqbalance6273.sock) but current CPU ban list does not fully isolate layout CPUs.  Proceeding but performance may be reduced.  Consider modifying IRQBALANCE_BANNED_CPUS or equivalent system setup.  Masks current (ffffff,0,0,3f8,1fff) desired (fffff,0,0,0,fffff) missing (fe000).  Lists current (0,1,2,3,4,5,6,7,8,9,10,11,12,35,36,37,38,39,40,41,128,129,130,131,132,133,134,135,136,137,138,139,140,141,142,143,144,145,146,147,148,149,150,151) desired (0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,128,129,130,131,132,133,134,135,136,137,138,139,140,141,142,143,144,145,146,147) missing (13,14,15,16,17,18,19).
```

src/app/shared_dev/commands/pktgen/pktgen.c
ripatel-fd: it doesn't make sense to run pktgen over loopback, so i would remove this. pktgen looks for an ethernet peer (ARP) but there are no peers on loopback

ripatel-fd:
The log message is cool, but I wonder if it can exceed the max log message size on high core count systems.

amass-jump:
> The log message is cool, but I wonder if it can exceed the max log message size on high core count systems.

AFAICT the max log message size is `#define FD_LOG_BUF_SZ (32UL*4096UL)`. Is there a tighter limit somewhere? We would need a few orders of magnitude more cores to hit that haha

src/app/fdctl/main.c
mmcgee-jump: Yup

src/app/shared/commands/configure/irq-affinity.c
```diff
@@ -0,0 +1,336 @@
...
+    }
+    if( FD_UNLIKELY( ((off_t)-1)==lseek( fd, 0, SEEK_SET ) ) ) {
+      FD_LOG_WARNING(( "lseek(%s) failed (%i-%s)", path, errno, fd_io_strerror( errno ) ));
+      break;
+    }
+    if( buf_sz==write( fd, mask, (ulong)buf_sz ) ) {
+      modified = 1;
+      FD_LOG_INFO(( "set IRQ affinity mask (%s) to (%.*s)", path, (int)buf_sz, mask ));
+    } else if( FD_UNLIKELY( errno != EIO )) {
+      /* EIO indicates an IRQ which is not allowed to be changed */
```
mmcgee-jump, line 169: How common is EIO, should we log an info or something in that case?

book/guide/initializing.md
```diff
@@ -202,6 +204,34 @@ Firedancer. It has no dependencies on any other stage.
...
+The Linux kernel uses IRQ (interrupt request) handlers to service many
+kinds of events, such as incoming network packets. These routines have
+a CPU affinity mask associated with them that determines the set of CPU
+cores where they may run. Other processes (including Firedancer) may be
+context switched by the IRQ handler.
+
+In order to increase performance, Firedancer attempts to isolate the CPU
+cores needed for its tiles from all IRQ affinities. This is done by
+reading and writing the `/proc/irq/<num>/smp_affinity` files. Some IRQs
+do not allow their affinities to be changed and these failures will be
```
mmcgee-jump, line 234: can you give an example of the kind of IRQ that can't be changed here? Also probably better to word as

> do not allow their affinities to be changed and these IRQs will not be changed, but a (WARNING|INFO, up to you, depending on how common this is) diagnostic message will be written to the log.

mmcgee-jump (dismissed):

mmcgee-jump:
Nice!

ripatel-fd (dismissed):

src/app/shared/commands/configure/irq-affinity.c
amass-jump: It's very common. Of the 359 IRQs on my host, 194 of them fail to be changed with EIO. It's not unexpected / not an error condition, some IRQs just can't be changed

amass-jump: On my system it was `timer` and all the `nvme` interrupts

HGuillemet:
What about setting the affinity of the NIC IRQs to the core(s) running the net tile(s) ?

amass-jump:
> What about setting the affinity of the NIC IRQs to the core(s) running the net tile(s) ?

That is the eventual plan, but for now we wanted to try a simpler change to just isolate FD cores

HGuillemet:
Ok. But theoretically,  what would be the most efficient? 
Setting the affinity to the core running the net tile, or to a dedicated core in the same die or same Numa node? Or maybe the sibling?

ripatel-fd:
> Ok. But theoretically, what would be the most efficient? Setting the affinity to the core running the net tile, or to a dedicated core in the same die or same Numa node? Or maybe the sibling?

Having net tile and ksoftirqd separated results in higher throughput per net tile, but more total CPU usage per packet.

Theoretically the most efficient setup is to not use IRQs. With NAPI busy polling, IRQs are masked, so interrupt handlers and ksoftirqd don't run at all. All processing (kernel+user) is then done on the net tile's thread. This is about twice as efficient as any setup that involves interrupts. Getting the timing and kernel config for NAPI busy polling right is very tricky though.

PR #6414, opened 2025-09-18
topob: auto layout should deprioritize HT pairs

amass-jump:
We keep the existing logic for Frakendancer and only use the new deprioritization for Firedancer. Both are changed to allocate core(s) for CRITICAL tiles before non-critical tiles.


mmcgee-jump:
Completed in another PR

PR #6418, opened 2025-09-18
runtime: add RBH to sysvar cache

ripatel-fd:
Adds back the 'recent blockhashes' sysvar to the sysvar cache.
Used by nonce instruction processors in the system program.


mmcgee-jump (approved):

ripatel-fd:
Blocked due to https://github.com/firedancer-io/test-vectors/issues/116

@mjain-jump could you give me a heads up here once the test-vectors repo is updated?

PR #6419, opened 2025-09-18
ethtool: disable tx-gre-segmentation features [WIP]

### LINKED ISSUE #6406: Disable tx-gre-segmentation feature in ethtool
Reported to cause issues with GRE traffic on some versions of Firedancer

amass-jump:
Renamed ethtool-gro into ethtool-offloads and added logic for GRE segmentation.  Also refactored ethtool-loopback to use the fd_ethtool_ioctl API.

Fixes #6406


amass-jump:
Reminder: update release notes for ethtool-gro -> ethtool-offloads

src/app/shared/commands/configure/ethtool-offloads.c
```diff
@@ -69,48 +69,29 @@ device_read_slaves( const char * device,
...
-                  device ));
-
-  /* execute command */
-  if( FD_UNLIKELY( ioctl( sock, SIOCETHTOOL, &ifr ) ) ) {
-    FD_LOG_ERR(( "configuring network device, ioctl(SIOCETHTOOL,ETHTOOL_SGRO) failed (%i-%s)",
-                 errno, fd_io_strerror( errno ) ));
-  }
+  /* turn off tx-gre-segmentation and tx-gre-csum-segmentation, which
+     have been known to cause issues with packet loss due to interactions
+     between AF_XDP and GRE. */
```
ripatel-fd: This isn't technically true. We've only seen packet corruption when sending UDP packets via sockets. But not over AF_XDP. The socket path was impacted regardless if any other user on the system was using XDP.

src/app/shared/commands/configure/ethtool-offloads.c
```diff
@@ -69,48 +69,29 @@ device_read_slaves( const char * device,
...
+  FD_TEST( 0==fd_ethtool_ioctl_feature_set( &ioc, FD_ETHTOOL_FEATURE_TXGRESEG,     0 ) );
+  FD_TEST( 0==fd_ethtool_ioctl_feature_set( &ioc, FD_ETHTOOL_FEATURE_TXGRECSUMSEG, 0 ) );
```
ripatel-fd, line 89: I don't think we needed to disable csum-segmentation too, but can't hurt I suppose

amass-jump: AFAICT the only difference between these two feature flags is whether or not the packet contains the optional checksum. It's not a "normal" checksumming offload like `rx-checksumming`.

```
[2.1](https://datatracker.ietf.org/doc/html/rfc2784#section-2.1). GRE Header

   The GRE packet header has the form:

    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
    |C|       Reserved0       | Ver |         Protocol Type         |
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
    |      Checksum (optional)      |       Reserved1 (Optional)    |
    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```

PR #6420, opened 2025-09-18
poh: fix slot_done_delay units

akhinvasara-jumptrading:


mmcgee-jump (dismissed):

src/disco/metrics/fd_prometheus.c
```diff
@@ -113,7 +113,7 @@ render_counter( fd_prom_render_t *        r,
       value = fd_metrics_convert_ticks_to_nanoseconds( value );
       break;
     case FD_METRICS_CONVERTER_SECONDS:
-      value = (ulong)(fd_metrics_convert_ticks_to_seconds( value ) * 1e9);
+      value = (ulong)fd_metrics_convert_ticks_to_seconds( value );
```
ptaffet-jump: I'm not exactly sure what the use case for a counter measured in seconds is, but you should probably add 0.5 before you cast so it rounds instead of floor

ptaffet-jump (approved):

PR #6443, opened 2025-09-19
snapshots: parse status cache

cali-jumptrading:


src/discof/restore/utils/fd_slot_delta_parser.h
```diff
@@ -0,0 +1,88 @@
...
+  struct {
+    ulong next;
+  } pool;
+
+  struct {
+    ulong next;
+    ulong prev;
+  } map;
+};
+typedef struct fd_slot_entry fd_slot_entry_t;
```
ripatel-fd, line 35: Possibly stupid suggestion: object pools with power-of-2 sized slot footprint avoid integer divides which can help with performance. So, you could use `uint` for next/prev here to get a 16 byte struct, if the status cache is guaranteed to have less than 4 billion entries ... You could also argue it doesn't matter because integer-divide-by-constant is optimized out to a large integer multiplication by the compiler

src/discof/restore/utils/fd_slot_delta_parser.h
```diff
@@ -0,0 +1,88 @@
...
+                           ulong                                   bank_slot );
+
+
+int
+fd_slot_delta_parser_consume( fd_slot_delta_parser_t * parser,
+                              uchar const *            buf,
+                              ulong                    bufsz );
+
+slot_set_t const *
+fd_slot_delta_parser_get_slot_set( fd_slot_delta_parser_t * parser );
```
ripatel-fd: In Firedancer code style we usually don't prefix getters

```suggestion
fd_slot_delta_parser_slot_set( fd_slot_delta_parser_t * parser );
```

src/discof/restore/utils/fd_slot_delta_parser.c
```diff
@@ -0,0 +1,398 @@
...
+  uchar   error;
+
+  fd_slot_delta_parser_process_entry_fn_t entry_cb;
+  void *                                  cb_arg;
+
+  fd_slot_entry_t * slot_pool;
+  slot_set_t *      slot_set;
+  ulong             slot_pool_ele_cnt;
+  fd_sstxncache_entry_t entry[1];
+};
```
ripatel-fd, line 47: Can you briefly document these state variables? e.g. what does `dst` point to. What's `borsh_io_error_len`. etc

src/flamenco/runtime/sysvar/fd_sysvar_slot_history.c
```diff
@@ -117,8 +118,8 @@ fd_sysvar_slot_history_update( fd_exec_slot_ctx_t * slot_ctx ) {
 fd_slot_history_global_t *
 fd_sysvar_slot_history_read( fd_funk_t *     funk,
                              fd_funk_txn_t * funk_txn,
-                             fd_spad_t *     spad ) {
-
+                             void *          buf,
+                             ulong           bufsz ) {
```
ripatel-fd:
```suggestion
                             uchar out_mem[ FD_SYSVAR_SLOT_HISTORY_FOOTPRINT ] ) {
```

we know the worst case footprint

src/flamenco/runtime/sysvar/fd_sysvar_slot_history.c
```diff
@@ -4,6 +4,7 @@
 #include "../fd_system_ids.h"
 #include "../fd_txn_account.h"
 #include "../../../funk/fd_funk_txn.h"
+#include <stdint.h>
```
ripatel-fd: accidental import

src/flamenco/runtime/sysvar/fd_sysvar_slot_history.c
```diff
@@ -148,12 +149,11 @@ fd_sysvar_slot_history_read( fd_funk_t *     funk,
     FD_LOG_ERR(( "fd_slot_history_decode_footprint failed" ));
   }
 
-  uchar * mem = fd_spad_alloc( spad, fd_slot_history_align(), total_sz );
-  if( !mem ) {
-    FD_LOG_ERR(( "Unable to allocate memory for slot history" ));
+  if( FD_UNLIKELY( bufsz<total_sz ) ) {
+    FD_LOG_ERR(( "Insufficient buffer size for slot history: need %lu, have %lu", total_sz, bufsz ));
```
ripatel-fd: remove this if check, just require the caller to pass in FD_SYSVAR_SLOT_HISTORY_FOOTPRINT sized buffer

ripatel-fd: actually, add a check for the input data length though. FD_SYSVAR_SLOT_HISTORY_FOOTPRINT is only valid assuming the input bincode size <= FD_SYSVAR_SLOT_HISTORY_BINCODE_SZ

ripatel-fd: If the slot history sysvar is oversized, log something like "corrupt slot history sysvar: sysvar data is too large (%lu bytes)"

ripatel-fd:
based

src/discof/restore/utils/fd_ssctrl.h
```diff
@@ -38,15 +38,19 @@
...
 
-#define FD_SNAPSHOT_MSG_DATA                   (0UL) /* Fragment represents some snapshot data */
+#define FD_SNAPSHOT_MSG_DATA                    (0UL) /* Fragment represents some snapshot data */
 
-#define FD_SNAPSHOT_MSG_CTRL_RESET_FULL        (1UL) /* Reset to start loading a fresh full snapshot */
-#define FD_SNAPSHOT_MSG_CTRL_EOF_FULL          (2UL) /* Full snapshot data is done, incremental data starting now */
-#define FD_SNAPSHOT_MSG_CTRL_RESET_INCREMENTAL (3UL) /* Incremental data being retried, start incremental over */
-#define FD_SNAPSHOT_MSG_CTRL_DONE              (4UL) /* Snapshot load is over, data is finished for this tile */
-#define FD_SNAPSHOT_MSG_CTRL_SHUTDOWN          (5UL) /* All tiles have acknowledged snapshot load is done, can now shutdown */
+#define FD_SNAPSHOT_MSG_CTRL_EXPECT_INCREMENTAL (1UL) /* Informational message, snapshot tiles should expect an incremental snapshot stream */
```
mmcgee-jump: Why is this changing?

src/app/firedancer/topology.c
```diff
@@ -956,7 +956,8 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
 
   } else if( FD_UNLIKELY( !strcmp( tile->name, "snapin" ) ) ) {
 
-    tile->snapin.funk_obj_id = fd_pod_query_ulong( config->topo.props, "funk", ULONG_MAX );
+    tile->snapin.funk_obj_id   = fd_pod_query_ulong( config->topo.props, "funk",     ULONG_MAX );
+    tile->snapin.tcache_obj_id = fd_pod_query_ulong( config->topo.props, "txncache", ULONG_MAX );
```
```suggestion
    tile->snapin.txncache_obj_id = fd_pod_query_ulong( config->topo.props, "txncache", ULONG_MAX );
```

tcache has a different meaning in fd

src/discof/restore/fd_snapin_tile.c
```diff
@@ -245,6 +301,15 @@ handle_control_frag( fd_snapin_tile_t *  ctx,
         break;
       }
 
+      slot_set_t const *         slot_set  = fd_slot_delta_parser_get_slot_set( ctx->ssparse->slot_delta_parser );
+      fd_slot_entry_t const *    slot_pool = fd_slot_delta_parser_get_slot_pool( ctx->ssparse->slot_delta_parser );
+      fd_slot_history_global_t * history   = fd_sysvar_slot_history_read( ctx->funk, ctx->funk_txn, ctx->scratch, ctx->scratch_sz );
```
mmcgee-jump: Please no

src/discof/restore/utils/fd_ssctrl.h
cali-jumptrading: We only need to parse the status cache from the incremental snapshot if we are fetching one. But the snapin tile doesn't know it's going to receive an incremental snapshot until it receives `EOF_FULL`, which is too late. If it receives `DONE` then it should have parsed the status cache from the full snapshot... what do you think?

We could always parse the status cache... just not sure what effects it'll have on the status cache during replay. Maybe it just gets a little extra bloated from stale entries in the full snapshot?

mmcgee-jump: Well, but is this control flow even possible? We don't know at the time we start the full if we will get an incremental or not.

mmcgee-jump: Let's separate out the control flow changes for another PR

src/discof/restore/fd_snapin_tile.c
```diff
@@ -89,16 +100,52 @@ metrics_write( fd_snapin_tile_t * ctx ) {
...
+  /* TODO: verify bank slot matches predicted slot */
+
   ulong sig = ctx->full ? fd_ssmsg_sig( FD_SSMSG_MANIFEST_FULL ) :
                           fd_ssmsg_sig( FD_SSMSG_MANIFEST_INCREMENTAL );
   fd_stem_publish( ctx->stem, 0UL, sig, ctx->manifest_out.chunk, sizeof(fd_snapshot_manifest_t), 0UL, 0UL, 0UL );
   ctx->manifest_out.chunk = fd_dcache_compact_next( ctx->manifest_out.chunk, sizeof(fd_snapshot_manifest_t), ctx->manifest_out.chunk0, ctx->manifest_out.wmark );
 }
 
+static void
+status_cache_cb( void *                        _ctx,
```
mmcgee-jump, line 228: Just leave this empty for now, I'll fill it with a later PR

src/discof/restore/utils/fd_ssctrl.h
cali-jumptrading: that's fine with me, so just to clarify, we'll parse both status caches (from the full and incremental) and insert all parsed entries?

mmcgee-jump: When loading an incremental, we will reset the status cache to have nothing in it

src/discof/restore/utils/fd_slot_delta_parser.c
```diff
@@ -0,0 +1,398 @@
+#include "fd_slot_delta_parser.h"
```
mmcgee-jump, line 1: Can you remove the slot_delta machinery from fd_types

cali-jumptrading: Looks like there's still a dependency here: https://github.com/firedancer-io/firedancer/blob/main/src/flamenco/runtime/fd_txncache.h#L408

src/discof/restore/utils/fd_snapshot_parser.h
```diff
@@ -113,7 +118,11 @@ fd_snapshot_parser_reset( fd_snapshot_parser_t * self,
   self->flags = 0UL;
   fd_snapshot_parser_reset_tar( self );
   fd_ssmanifest_parser_init( self->manifest_parser, (fd_snapshot_manifest_t*)manifest_buf );
-  self->manifest_done = 0;
+  /* TODO: get the predicted slot from snaprd */
```
mmcgee-jump: Todo not needed?

src/discof/restore/utils/fd_snapshot_parser.c
```diff
@@ -15,6 +15,7 @@ fd_snapshot_parser_footprint( ulong max_acc_vecs ) {
   l = FD_LAYOUT_APPEND( l, alignof(fd_snapshot_parser_t), sizeof(fd_snapshot_parser_t)                   );
   l = FD_LAYOUT_APPEND( l, fd_ssmanifest_parser_align(),  fd_ssmanifest_parser_footprint( max_acc_vecs ) );
   l = FD_LAYOUT_APPEND( l, 16UL,                          1UL<<20UL                                      );
+  l = FD_LAYOUT_APPEND( l, fd_slot_delta_parser_align(), fd_slot_delta_parser_footprint()                );
```
mmcgee-jump: Put this ahead of the 1UL<<20UL probably

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -623,6 +623,7 @@ after_credit( fd_snaprd_tile_t *  ctx,
         break;
       }
 
+      /* TODO: get the expected slot */
```
mmcgee-jump: Not needed?

src/discof/restore/fd_snapin_tile.c
```diff
@@ -75,8 +125,11 @@ static ulong
 scratch_footprint( fd_topo_tile_t const * tile ) {
   (void)tile;
   ulong l = FD_LAYOUT_INIT;
-  l = FD_LAYOUT_APPEND( l, alignof(fd_snapin_tile_t),  sizeof(fd_snapin_tile_t)                  );
-  l = FD_LAYOUT_APPEND( l, fd_snapshot_parser_align(), fd_snapshot_parser_footprint( 1UL<<24UL ) );
+  l = FD_LAYOUT_APPEND( l, alignof(fd_snapin_tile_t),      sizeof(fd_snapin_tile_t)                                     );
+  l = FD_LAYOUT_APPEND( l, fd_snapshot_parser_align(),     fd_snapshot_parser_footprint( 1UL<<24UL )                    );
+  l = FD_LAYOUT_APPEND( l, alignof(fd_sstxncache_entry_t), sizeof(fd_sstxncache_entry_t)*FD_SNAPIN_TXNCACHE_MAX_ENTRIES );
+  l = FD_LAYOUT_APPEND( l, blockhash_pool_align(),         blockhash_pool_footprint( 301UL )                            );
+  l = FD_LAYOUT_APPEND( l, blockhash_set_align(),          blockhash_set_footprint( 512UL )                             );
```
mmcgee-jump: Probably make this 1024UL so the lookups are a little faster

src/discof/restore/fd_snapin_tile.c
```diff
@@ -20,6 +22,46 @@
 #define FD_SNAPIN_STATE_MALFORMED (2) /* The snapshot is malformed, we are waiting for a reset notification */
 #define FD_SNAPIN_STATE_SHUTDOWN  (3) /* The tile is done, been told to shut down, and has likely already exited */
 
+#define FD_SNAPIN_TXNCACHE_MAX_ENTRIES (2UL<<20UL)
```
mmcgee-jump: This should be exactly `300UL*FD_PACK_MAX_TXN_PER_SLOT`

mmcgee-jump: 300 comes from `status_cache::MAX_CACHE_ENTRIES` in Agave

src/discof/restore/fd_snapin_tile.c
```diff
@@ -89,16 +142,86 @@ metrics_write( fd_snapin_tile_t * ctx ) {
...
+    fd_txncache_advance_root( ctx->txncache, id );
+  }
+
+  for( ulong i=0UL; i<ctx->txncache_entries_len; i++ ) {
+    fd_sstxncache_entry_t const * entry = &ctx->txncache_entries[i];
+    if( blockhash_set_idx_query_const( ctx->blockhash_set, (fd_blockhash_wrapper_t *)entry->blockhash, ULONG_MAX, ctx->blockhash_pool )==ULONG_MAX ) continue;
+    fd_txncache_insert( ctx->txncache, root, entry->blockhash, entry->txnhash );
+  }
+
+  /* clear the blockhash set and pool */
```
mmcgee-jump, line 190: I don't see any malformed/reset logic to clear the pool/set (which is where it should happen, not here).

src/discof/restore/fd_snapin_tile.c
```diff
@@ -89,16 +142,86 @@ metrics_write( fd_snapin_tile_t * ctx ) {
...
+    fd_blockhash_entry_t * blockhash_entry = &ctx->blockhash_pool[ ctx->blockhash_pool_ele_cnt++ ];
+    blockhash_entry->blockhash.bh = blockhashes[i].hash;
+    blockhash_set_ele_insert( ctx->blockhash_set, blockhash_entry, ctx->blockhash_pool );
+
+    fd_txncache_fork_id_t id = fd_txncache_attach_child( ctx->txncache, root, 0UL, blockhashes[i].hash );
+    fd_txncache_advance_root( ctx->txncache, id );
+  }
+
+  for( ulong i=0UL; i<ctx->txncache_entries_len; i++ ) {
+    fd_sstxncache_entry_t const * entry = &ctx->txncache_entries[i];
```
mmcgee-jump: Just comment these 3 lines for main, I'll wire in correctly

src/discof/restore/fd_snapin_tile.c
cali-jumptrading: The blockhash pool/set is only needed to make blockhash lookups fast... so it should be fine to clear it after building it and using it populate the txncache.

mmcgee-jump: But if you get a malformed snapshot and reset, it will not clear it?

mmcgee-jump (approved):

PR #6444, opened 2025-09-19
fuzz: enforce presence of sysvars in context creation

mjain-jump:


src/flamenco/runtime/tests/fd_txn_harness.c
```diff
@@ -110,70 +110,31 @@ fd_runtime_fuzz_txn_ctx_create( fd_solfuzz_runner_t *              runner,
...
 
-  /* Provide default last restart slot sysvar if not provided */
-  FD_TXN_ACCOUNT_DECL( acc );
-  int err = fd_txn_account_init_from_funk_readonly( acc, &fd_sysvar_last_restart_slot_id, funk, funk_txn );
-  if( err==FD_ACC_MGR_ERR_UNKNOWN_ACCOUNT ) {
-    fd_sysvar_last_restart_slot_init( slot_ctx );
-  }
+  fd_sol_sysvar_last_restart_slot_t last_restart_slot_[1];
+  fd_sol_sysvar_last_restart_slot_t const * last_restart_slot = fd_sysvar_last_restart_slot_read( funk, funk_txn, last_restart_slot_ );
+  FD_TEST( last_restart_slot );
```
ripatel-fd: Pretty sure this sysvar is optional (does it exist if the cluster never restarted?)

src/flamenco/runtime/tests/fd_txn_harness.c
```diff
@@ -110,70 +110,31 @@ fd_runtime_fuzz_txn_ctx_create( fd_solfuzz_runner_t *              runner,
...
-
-  /* Provide default stake history if not provided */
   fd_stake_history_t * stake_history = fd_sysvar_stake_history_read( funk, funk_txn, runner->spad );
-  if( !stake_history ) {
-    // Provide a 0-set default entry
-    fd_epoch_stake_history_entry_pair_t entry = {0};
-    fd_sysvar_stake_history_init( slot_ctx );
-    fd_sysvar_stake_history_update( slot_ctx, &entry, runner->spad );
-  }
+  FD_TEST( stake_history );
```
ripatel-fd, line 129: Does Agave panic if this is missing? Firedancer should handle missing sysvars gracefully for transaction execution, except for rent, epoch schedule, and clock.

mjain-jump: The Agave bank creates a stake history entry on initialization if it doesn't already exist, which is why it needs to be provided for the txn harness. If it's missing it causes a mismatch. Solfuzz will always provide a stake history sysvar.

src/flamenco/runtime/tests/fd_txn_harness.c
mjain-jump: The sysvar exists on mainnet, testnet, and devnet, and is also always provided by Solfuzz. I could remove the `FD_TEST` but we're not currently covering the case where the sysvar is not provided.

ripatel-fd: Yes, because those networks were restarted. But a local test validator / SVM fork might not have it, I think. 
The FD_TEST should be safe to remove.

src/flamenco/runtime/tests/fd_txn_harness.c
ripatel-fd: Ok, what about the other ones? slot_history, etc

src/flamenco/runtime/tests/fd_txn_harness.c
mjain-jump: done

src/flamenco/runtime/tests/fd_txn_harness.c
mjain-jump: Slot hashes and stake history have this as well. 
https://github.com/anza-xyz/agave/blob/v3.0.0/runtime/src/bank.rs#L1397-L1400

I'm pretty sure I ran into some Agave UB / panic in the block fuzzer as well because of poorly-sized slot history.

mjain-jump: Clock, epoch schedule, rent, RBH, and last restart slot also all get initialized / updated in the entrypoint we use for bank initialization.

https://github.com/anza-xyz/agave/blob/v3.0.0/runtime/src/bank.rs#L1173-L1177

PR #6445, opened 2025-09-20
replay: various fixes

mmcgee-jump:


src/discof/replay/fd_replay_tile.c
```diff
@@ -1653,8 +1652,14 @@ process_fec_set( fd_replay_tile_t *  ctx,
...
         FD_LOG_CRIT(( "bank for leader slot %lu not found", reasm_fec->slot ));
       }
-      FD_LOG_WARNING(("FINI LEADER BANK %lu", reasm_fec->slot));
       fini_leader_bank( ctx, bank, stem );
+
+      fd_eslot_t          eslot        = fd_eslot( reasm_fec->slot, 0UL );
+      fd_funk_txn_xid_t   xid          = { .ul = { fd_eslot_slot( eslot ), fd_eslot_slot( eslot ) } };
+      fd_funk_txn_map_t * txn_map      = fd_funk_txn_map( ctx->funk );
+      fd_funk_txn_t *     funk_txn     = fd_funk_txn_query( &xid, txn_map );
+      ctx->slot_ctx->funk_txn = funk_txn;
```
lidatong, line 1661: needs a FIXME that this is wrong -- i hacked it in here since it's also where fini bank is also already happening and funk was broken otherwise

src/disco/shred/fd_shred_tile.c
```diff
@@ -254,7 +254,7 @@ typedef struct {
   fd_shred_dest_idx_t scratchpad_dests[ FD_SHRED_DEST_MAX_FANOUT*(FD_REEDSOL_DATA_SHREDS_MAX+FD_REEDSOL_PARITY_SHREDS_MAX) ];
 
   uchar * chained_merkle_root;
-  fd_bmtree_node_t out_merkle_root;
+  fd_bmtree_node_t out_merkle_roots[ FD_SHRED_BATCH_FEC_SETS_MAX ];
```
lidatong, line 257: @0x0ece @jvarela-jump @ptaffet-jump  can you take a look? this is a somewhat ugly fix i made for the leader pipeline: we can generate (and therefore "complete") up to `k` FEC sets per tile loop, so the result is we need to store up to `k` merkle roots (`k=2` currently)

(as follower, it's at most one FEC set completing per net shred frag)

this can be removed / simplified with fixed-32

src/discof/replay/fd_replay_tile.c
```diff
@@ -471,8 +471,7 @@ fd_replay_out_vote_tower_from_funk(
     fd_funk_rec_key_t funk_key = fd_funk_acc_key( pubkey );
     fd_funk_rec_t const * rec = fd_funk_rec_query_try_global( funk, funk_txn, &funk_key, NULL, &query );
     if( FD_UNLIKELY( !rec ) ) {
-      FD_LOG_WARNING(( "vote account not found. address: %s",
-        FD_BASE58_ENC_32_ALLOCA( pubkey->uc ) ));
+      FD_LOG_WARNING(( "vote account not found. address: %s", FD_BASE58_ENC_32_ALLOCA( pubkey->uc ) ));
```
lidatong, line 474: probably should be log crit..

src/flamenco/runtime/fd_runtime.c
```diff
@@ -714,22 +714,14 @@ fd_runtime_block_execute_prepare( fd_exec_slot_ctx_t * slot_ctx,
 
   fd_bank_signature_count_set( slot_ctx->bank, 0UL );
```
ibhatt-jumptrading, line 716: why is this getting deleted

src/disco/shred/fd_shred_tile.c
```diff
@@ -1414,7 +1414,7 @@ unprivileged_init( fd_topo_t *      topo,
   ctx->shredder_max_fec_set_idx = (shred_store_mcache_depth+1UL)/2UL + 1UL;
 
   ctx->chained_merkle_root = NULL;
-  memset( &ctx->out_merkle_root, 0, sizeof(fd_bmtree_node_t) );
+  memset( ctx->out_merkle_roots, 0, sizeof(fd_bmtree_node_t) );
```
ptaffet-jump: This only zeros the first

src/flamenco/runtime/fd_runtime.c
mmcgee-jump: These are accumulating counters in Agave, even across slots. We need to do the same as some of them are exposed to RPC

src/disco/shred/fd_shred_tile.c
ptaffet-jump: This change to the shred tile seems fine to me. When we simplify everything for fixed-32, maybe we can clean this up too

PR #6450, opened 2025-09-22
codeql: add query for suspicious index max comparisons

intrigus-lgtm:
@two-heart


two-heart (approved):

contrib/codeql/nightly/SuspiciousIndexMaxComparison.ql
```diff
@@ -0,0 +1,93 @@
...
+    )
+  }
+}
+
+private string getIndexPattern() {
+  result = "%IDX%" or
+  result = "%idx%" or
+  result = "%index%" or
+  // Matches common variable names like i, j, k, l, m, n
+  result = "abcdefghijklmnopqrstuvwxyz".charAt(_)
```
two-heart, line 44: would have limited it to actually just i, j, k, but does not produce FPs rn, and we can adjust it when it does

intrigus-lgtm: 👍

PR #6453, opened 2025-09-22
build: do not omit frame pointers

mmcgee-jump:


ripatel-fd:
why?

mmcgee-jump:
Here's a pretty good summary: https://www.brendangregg.com/blog/2024-03-17/the-return-of-the-frame-pointers.html

ripatel-fd:
OK, so this change trades a bit of performance (register allocation) for better production profiling with perf. 
Have you considered Intel LBR and PT?

ripatel-fd:
cc @dpzmick probably affects your build system too

PR #6465, opened 2025-09-22
funk: improve rec_publish concurrency

ripatel-fd:
- Remove a mutex from the funk_txn and funk_shmem classes.
- Rewrite doubly linked list append in rec_publish using atomic CAS
  primitive

Aims to achieve the following:

- Concurrent / racy linked list appends to a funk transaction (rec_publish) do the correct thing
- Concurrent linked list walk over funk records (forwards or backwards) must be sound (but may skip over a record when going back-to-front due to a race between the two CAS operations)

### DIFF
--- a/src/funk/fd_funk_rec.c
+++ b/src/funk/fd_funk_rec.c
@@ -250,13 +250,11 @@ fd_funk_rec_prepare( fd_funk_t *               funk,
   if( txn == NULL ) {
     fd_funk_txn_xid_set_root( rec->pair.xid );
     rec->txn_cidx = fd_funk_txn_cidx( FD_FUNK_TXN_IDX_NULL );
-    prepare->txn_lock     = &funk->shmem->lock;
   } else {
     fd_funk_txn_xid_copy( rec->pair.xid, &txn->xid );
     rec->txn_cidx = fd_funk_txn_cidx( (ulong)( txn - funk->txn_pool->ele ) );
     prepare->rec_head_idx = &txn->rec_head_idx;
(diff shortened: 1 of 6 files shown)


src/funk/fd_funk_rec.c
```diff
@@ -266,34 +264,89 @@ fd_funk_rec_prepare( fd_funk_t *               funk,
...
+    }
+
+    if( FD_UNLIKELY( !__sync_bool_compare_and_swap( next_idx_p, FD_FUNK_REC_IDX_NULL, rec_idx ) ) ) {
+      /* Another thread beat us to the punch */
+      FD_SPIN_PAUSE();
+      continue;
     }
+
+    if( FD_UNLIKELY( !__sync_bool_compare_and_swap( rec_tail_idx, rec_prev_idx, rec_idx ) ) ) {
+      /* This CAS is guaranteed to succeed if the previous CAS passed. */
```
mmcgee-jump, line 304: Not worth another CAS just for the assertion imo

ripatel-fd: @mmcgee-jump Two back to back CAS in the non-conflicting case have pretty good latency on Zen 4 as far as I can tell. Can we keep this in for now, for robustness? And then remove it only if it becomes a performance concern? 

We'll probably occasionally need to break clean API barriers occasionally, and directly modify funk guts, to give performance critical loops better performance. And having all surrounding code be paranoid helps catch bugs quickly if that highly optimized code is broken.

PR #6486, opened 2025-09-23
log: init with true tid rather than pid

### LINKED ISSUE #3644: Invalid fd_log_tid on fdctl
```
/* fd_log_tid() returns the caller's thread group thread id.  A thread
   group thread id is intended, at a minimum, to be unique over all
   running threads in a thread group.  In simple cases, this is the
   caller's OS tid.  In general cases, this is typically something
   provided to the thread when that thread started.  This is cheap after
   the first call. */

ulong fd_log_tid( void );
```

fd_log_tid returns the same integer for different threads in a thread group when running on fdctl.

```
fd1:[group]:bank:0 src/disco/topo/fd_topo_run.c(141)[fd_topo_run_tile]: fd_log: --log-tid           1805503
fd1:[group]:bank:1 src/disco/topo/fd_topo_run.c(141)[fd_topo_run_tile]: fd_log: --log-tid           1805503
fd1:[group]:bank:2 src/disco/topo/fd_topo_run.c(141)[fd_topo_run_tile]: fd_log: --log-tid           1805503
fd1:[group]:bank:3 src/disco/topo/fd_topo_run.c(141)[fd_topo_run_tile]: fd_log: --log-tid           1805503
```

amass-jump:
Fixes #3644

### DIFF
--- a/src/disco/topo/fd_topo_run.c
+++ b/src/disco/topo/fd_topo_run.c
@@ -20,10 +20,9 @@
 static void
 initialize_logging( char const * tile_name,
                     ulong        tile_kind_id,
-                    ulong        pid,
                     ulong        tid ) {
   fd_log_cpu_set( NULL );
-  fd_log_private_tid_set( pid );
+  fd_log_private_tid_set( tid );
   char thread_name[ 20 ];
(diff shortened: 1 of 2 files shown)


amass-jump:
The only place other than logging that `fd_log_tid()` is used:

```
  59 static inline ulong
  60 fd_rwlock_recursive_tid( void ) {
  61   return (fd_log_tid()<<12)+fd_log_thread_id();
  62 }
```

PR #6501, opened 2025-09-24
gossip: fix NAT support

ripatel-fd:
Fixes a bug where the `[gossip.host]` config option is ignored


src/app/firedancer/topology.c
```diff
@@ -846,8 +846,6 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
     tile->gossip.max_entries         = config->tiles.gossip.max_entries;
     tile->gossip.boot_timestamp_nanos = config->boot_timestamp_nanos;
 
-    tile->gossip.ip_addr = config->net.ip_addr;
```
mmcgee-jump, line 849: oops :)

PR #6507, opened 2025-09-25
dump: output wallclock timestamps in realtime mode

amass-jump:
This is not perfect, as some tiles do not populate tspub at all, or use some other value than fd_tickcount(). Additionally, if the original tspub was taken more than ~2s away from when the dump tool reads it, the timestamp decompression will return an incorrect result. Regardless, this should be more useful than putting the tango seq in the timestamp field and yields the expected result for the vast majority of frags.

### DIFF
--- a/src/app/shared_dev/commands/dump.c
+++ b/src/app/shared_dev/commands/dump.c
@@ -1,6 +1,7 @@
 #include "../../shared/fd_config.h"
 #include "../../shared/fd_action.h"
 #include "../../../disco/metrics/fd_metrics.h"
+#include "../../../util/clock/fd_clock.h"
 #include "../../../util/net/fd_pcap.h"
 
 #include <errno.h>
@@ -22,9 +23,16 @@ struct dump_ctx {
 
   ulong *                  metrics_base;
   long                     next_stat_log;
-  ulong                    last_frags;
-  ulong                    last_bytes;
+  ulong                    frags;
+  ulong                    bytes;
   ulong                    last_overrun_frags;
+
+  long                     warmup_until;
+
+  fd_clock_t               clock[ 1 ];
+  fd_clock_shmem_t         clock_mem[ 1 ];
+  fd_clock_epoch_t         clock_epoch[ 1 ];
+  long                     clock_recal_next;
 };
 typedef struct dump_ctx dump_ctx_t;
 
@@ -156,7 +164,7 @@ during_frag( dump_ctx_t * ctx,
   void const *           frag_data = fd_chunk_to_laddr_const( ctx->links[ in_idx ].dcache_base, chunk );
 
   fd_pcap_pkt( fd_io_buffered_ostream_peek( &ctx->ostream ),
-               (long)seq,
+               0L,
                mline,
                sizeof(fd_frag_meta_t),
                frag_data,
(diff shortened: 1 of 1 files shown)


src/app/shared_dev/commands/dump.c
```diff
@@ -297,7 +318,12 @@ dump_cmd_fn( args_t      * args,
     ctx.metrics_base = fd_metrics_join( fd_metrics_new( fd_alloca( FD_METRICS_ALIGN, FD_METRICS_FOOTPRINT( ctx.link_cnt, 0UL ) ), ctx.link_cnt, 0UL ) );
     fd_metrics_register( ctx.metrics_base );
 
-    ctx.next_stat_log = fd_log_wallclock() + (long)1e9;
+    fd_clock_default_init( ctx.clock, ctx.clock_mem );
+    ctx.clock_recal_next = fd_clock_default_recal( ctx.clock );
+    fd_clock_epoch_init( ctx.clock_epoch, ctx.clock_mem );
+    ctx.warmup_until = fd_clock_epoch_y( ctx.clock_epoch, fd_tickcount() ) + (long)1e9;
```
ptaffet-jump, line 330: Can we calculate this properly, rather than just assuming 1 second? We want to warm up until we're processing frags newer than 2^31 rdtsc ticks before now, right? If it's too hard to calculate, it needs a comment

amass-jump: Stem has no way (currently?) of catching up to the head of an existing ring. The logic always "burns through" existing frags at startup and delivers callbacks as usual. For once=1 mode this is desired, but for once=0 mode it didn't make too much sense to me. So my main goal with the warmup period was to ensure any packet in the pcap was published _after_ the dump command was started, as would be the case for a normal tcpdump / etc.

I couldn't think of any reliable way to do this other than just assuming it would take less than a second of wallclock time to burn through all the rings (while throwing away the frags). This seemed OK given the precise start/stop time of a human-ran dump command is probably not all that critical. 

Do you have any ideas on how to do this more precisely?

src/app/shared_dev/commands/dump.c
```diff
@@ -212,7 +229,11 @@ log_metrics( dump_ctx_t * ctx ) {
 
 static void
 during_housekeeping( dump_ctx_t * ctx ) {
-  long now = fd_log_wallclock();
+  long now = fd_clock_epoch_y( ctx->clock_epoch, fd_tickcount() );
+  if( FD_UNLIKELY( now > ctx->clock_recal_next ) ) {
+    ctx->clock_recal_next = fd_clock_default_recal( ctx->clock );
+    fd_clock_epoch_refresh( ctx->clock_epoch, ctx->clock_mem );
```
ptaffet-jump, line 235: I haven't used `fd_clock` before, so I'm just reading up on it. The docs say that `fd_clock_epoch_refresh` should be called much more frequently than `recal_avg`, so I think you should just shift this up/down a few lines to happen every housekeeping.

amass-jump: A more "normal" usage of fd_clock would have one "admin" thread running the recalibration logic periodically (every 10ms or so), and all the other tile/critical threads refreshing that latest epoch. These threads have no way of knowing precisely when the epoch was last calibrated / changed, so they run the epoch_refresh routine much more frequently than the epoch will actually be changed by the admin thread

In this more unique case, there is only one thread involved and we know exactly when the epoch can/will change, so we can put the epoch_refresh right there. Does that make sense?

ptaffet-jump: Ah okay, that makes sense.

src/app/shared_dev/commands/dump.c
ptaffet-jump: Okay, yeah, good point. Can you add a comment explaining this? People have thought about adding a `caught_up` callback to stem, but it hasn't really made sense so far

PR #6508, opened 2025-09-25
gui: full client minimally working

jherrera-jump:


src/disco/gui/fd_gui.c
```diff
@@ -1477,23 +1477,19 @@ fd_gui_handle_repair_slot( fd_gui_t * gui, ulong slot, long now ) {
...
-  ulong last_landed_vote = msg[ 0 ];
-
-  ulong parent_cnt = msg[ 1 ];
-  FD_TEST( parent_cnt<4096UL );
-
-  ulong _slot = msg[ 2 ];
-
-  for( ulong i=0UL; i<parent_cnt; i++ ) {
-    ulong parent_slot = msg[2UL+i];
+fd_gui_handle_reset_slot( fd_gui_t *    gui,
```
mmcgee-jump: I don't think we need to do it here, but we can do much better at building the tree in the GUI now, with an actual tree ... since we can feed enough info from replay etc. This whole slots array parent thing was something of a hack

PR #6525, opened 2025-09-26
gui: shred vis

jherrera-jump:
- live shred data is stored in a staging ring buffer
- rooted slots shred data is but in a larger, ordered history ring buffer
- adds `slot.query_shreds` and `slot.live_shreds` with support for two event timestamps (recieved and slot complete)
- catch-up history

### DIFF
--- a/src/disco/gui/fd_gui_printf.c
+++ b/src/disco/gui/fd_gui_printf.c
@@ -236,27 +236,48 @@ fd_gui_printf_vote_distance( fd_gui_t * gui ) {
 void
 fd_gui_printf_repair_slot( fd_gui_t * gui ) {
   jsonp_open_envelope( gui->http, "summary", "repair_slot" );
-    if( FD_LIKELY( gui->summary.slots_max_repair[ 0 ].slot!=ULONG_MAX  ) ) jsonp_ulong( gui->http, "value", gui->summary.slots_max_repair[ 0 ].slot );
-    else                                                                   jsonp_null ( gui->http, "value" );
+    if( FD_LIKELY( gui->summary.slot_repair!=ULONG_MAX  ) ) jsonp_ulong( gui->http, "value", gui->summary.slot_repair );
+    else                                                    jsonp_null ( gui->http, "value" );
   jsonp_close_envelope( gui->http );
 }
 
 void
 fd_gui_printf_turbine_slot( fd_gui_t * gui ) {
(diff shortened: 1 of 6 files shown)


src/disco/gui/fd_gui_tile.c
```diff
@@ -307,8 +307,8 @@ after_frag( fd_gui_ctx_t *      ctx,
           case FD_REPAIR_KIND_PING:
           case FD_REPAIR_KIND_PONG:
           case FD_REPAIR_KIND_ORPHAN: break;
-          case FD_REPAIR_KIND_SHRED: fd_gui_handle_repair_slot( ctx->gui, msg->shred.slot, tsorig_ns ); break;
-          case FD_REPAIR_KIND_HIGHEST_SHRED: fd_gui_handle_repair_slot( ctx->gui, msg->highest_shred.slot, tsorig_ns ); break;
+          case FD_REPAIR_KIND_SHRED: { if( FD_UNLIKELY( msg->shred.slot==0 ) ) { break; } fd_gui_handle_repair_slot( ctx->gui, msg->shred.slot, tsorig_ns ); break; }
```
jherrera-jump, line 310: Is this feature or a bug? I'm seeing requests with slot==0 every now and then?

src/disco/gui/fd_gui_printf.c
```diff
@@ -236,27 +236,48 @@ fd_gui_printf_vote_distance( fd_gui_t * gui ) {
...
   jsonp_close_envelope( gui->http );
 }
 
+void
+fd_gui_printf_catch_up_history( fd_gui_t * gui ) {
+  jsonp_open_envelope( gui->http, "summary", "catch_up_history" );
+    jsonp_open_object( gui->http, "value" );
+      jsonp_open_array( gui->http, "turbine" );
+        for( ulong i=0UL; i<gui->summary.catch_up_turbine_sz; i+=2 ) {
+          for( ulong j=gui->summary.catch_up_turbine[ i ]; j<=gui->summary.catch_up_turbine[ i+1UL ]; j++ ) {
```
jherrera-jump, line 266: would you prefer the message to also have some more compact encoding instead of just a list of slots? Doesn't seem to matter too much in practice

mmcgee-jump: This is fine

PR #6526, opened 2025-09-26
replay: plumb through bad block handling from scheduler

yufeng-jump:


src/discof/replay/fd_replay_tile.c
```diff
@@ -1325,12 +1325,6 @@ replay( fd_replay_tile_t *  ctx,
     FD_TEST( ready_txn->txn_idx!=FD_SCHED_TXN_IDX_NULL );
     charge_busy = 1;
 
-    /* EXTREMELY IMPORTANT TODO, SECURITY CRITICAL: We need to ensure
```
mmcgee-jump, line 1338: Where is this checked? I don't see it added

yufeng-jump: Scheduler will mark the block dead if more than max per slot transactions are parsed out.  Happens at FEC set ingestion time, not dispatch time.

src/discof/replay/fd_sched.c
```diff
@@ -1075,7 +1101,13 @@ fd_sched_parse_txn( fd_sched_t * sched, fd_sched_block_t * block, fd_sched_alut_
...
     /* Can't parse out a full transaction. */
-    return 0;
+    return FD_SCHED_PARSER_AGAIN_LATER;
+  }
+
+  if( FD_UNLIKELY( block->txn_parsed_cnt>=FD_SCHED_MAX_TXN_PER_BLOCK ) ) {
+    /* The block contains more transactions than a valid block would.
+       Mark the block dead instead of keep processing it. */
+    return FD_SCHED_PARSER_BAD_BLOCK;
   }
```
yufeng-jump: This is where max txn per slot is enforced.

src/discof/replay/fd_replay_tile.c
yufeng-jump: Scheduler will not dispatch anything from a dead block.

src/discof/replay/fd_sched.c
mmcgee-jump: This should probably check against `FD_TXN_MAX_PER_SLOT` not the `FD_SCHED_MAX_TXN_PER_BLOCK` which is sched specific? Why is it even redefined

mmcgee-jump (approved):

yufeng-jump:
#6553 fixes the merge failures

PR #6538, opened 2025-09-29
metrics, bankf: fix metrics oob in bankf

two-heart:
bug found by @intrigus-lgtm; fixed by using the correct enum for the counter in metrics.xml and then using the generated enum values, instead of redefining them and risking drift.


src/discof/bank/fd_bank_err.h
```diff
@@ -3,83 +3,38 @@
 
 #include "../../util/log/fd_log.h"
 #include "../../flamenco/runtime/fd_runtime_err.h"
+#include "../../disco/metrics/generated/fd_metrics_enums.h"
 
-#define FD_BANK_EXECUTE_SUCCESS                                     0
-
-/* Preflight and execution errors.  This is just instruction error,
```
mmcgee-jump, line 9: Can you transpose these group comments over to the metrics.xml

mmcgee-jump (changes requested):
.

mmcgee-jump (approved):

PR #6539, opened 2025-09-29
Clone less test-vectors

ripatel-fd:


mjain-jump (approved):

nlgripto (approved):

.github/workflows/tests.yml
```diff
@@ -119,13 +119,12 @@ jobs:
...
-        if: ${{ matrix.run-unit-tests && matrix.test-case == 'native' }}
-        uses: actions/checkout@v5
-        with:
-          repository: firedancer-io/test-vectors
-          path: dump/test-vectors
-          fetch-depth: 5
+      - name: check out test-vectors (persistent)
+        if: ${{ matrix.run-unit-tests && matrix.test-case != 'native' }}
+        run: |
+          if [[ ! -d /data/svc_firedancer ]]; then git -C /data/svc_firedancer clone https://github.com/firedancer-io/test-vectors.git; fi
```
mmcgee-jump, line 125: Probably add a commit hash for the test vectors repo? this is asking for trouble

ripatel-fd: This is only run once per runner, but will fix if something breaks

PR #6543, opened 2025-09-29
metrics, bank, pack, quic: use MACRO_CNT in def

intrigus-lgtm:
Use the MACRO_CNT variables in the definition of the metrics arrays. 
This is to avoid drift in the future which could result in over/ under reads/writes.

(Apologies for all the review requests...)

### DIFF
--- a/src/discoh/bank/fd_bank_tile.c
+++ b/src/discoh/bank/fd_bank_tile.c
@@ -46,8 +46,8 @@ typedef struct {
   fd_pack_rebate_sum_t rebater[ 1 ];
 
   struct {
-    ulong txn_load_address_lookup_tables[ 6 ];
-    ulong transaction_result[ 41 ];
+    ulong txn_load_address_lookup_tables[ FD_METRICS_COUNTER_BANK_TRANSACTION_LOAD_ADDRESS_TABLES_CNT ];
+    ulong transaction_result[ FD_METRICS_COUNTER_BANK_TRANSACTION_RESULT_CNT ];
     ulong processing_failed;
     ulong fee_only;
     ulong exec_failed;
--- a/src/waltz/quic/fd_quic.h
+++ b/src/waltz/quic/fd_quic.h
@@ -83,6 +83,7 @@
 
 #include "fd_quic_common.h"
 #include "fd_quic_enum.h"
+#include "../../disco/metrics/generated/fd_metrics_quic.h"
 
 #include "../aio/fd_aio.h"
 #include "../tls/fd_tls.h"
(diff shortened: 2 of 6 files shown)


src/discoh/bank/fd_bank_tile.c
```diff
@@ -46,8 +46,8 @@ typedef struct {
   fd_pack_rebate_sum_t rebater[ 1 ];
 
   struct {
-    ulong txn_load_address_lookup_tables[ 6 ];
-    ulong transaction_result[ 41 ];
+    ulong txn_load_address_lookup_tables[ FD_METRICS_COUNTER_BANK_TRANSACTION_LOAD_ADDRESS_TABLES_CNT ];
+    ulong transaction_result[ FD_METRICS_COUNTER_BANK_TRANSACTION_RESULT_CNT ];
```
mmcgee-jump, line 50: This one is actually TRANSACTION_ERROR_CNT I believe

intrigus-lgtm: Both `FD_METRICS_COUNTER_BANK_TRANSACTION_RESULT_CNT` and `FD_METRICS_ENUM_TRANSACTION_ERROR_CNT` map to the correct value, `41`.

`transaction_result` is used in `FD_MCNT_ENUM_COPY( BANK, TRANSACTION_RESULT,  ctx->metrics.transaction_result );` so `FD_METRICS_COUNTER_BANK_TRANSACTION_RESULT_CNT` definitely aligns with that.

I do agree that it's confusing to have `..._RESULT` as well as `..._ERROR`, but that's how it's defined in [metrics.xml](https://github.com/firedancer-io/firedancer/blob/4e2a850b0d75f6de158ba3f27eb07bc2c6a0f780/src/disco/metrics/metrics.xml#L688).

I don't think it would make sense to change the enum name to `TransactionResult` as that would deviate from agave and clash with `TransactionError` from bankf tile.

Should I instead rename the field to `transaction_err` and the counter name in metrics.xml to `TransactionError`?

src/waltz/quic/fd_quic.h
```diff
@@ -313,32 +314,32 @@ union fd_quic_metrics {
...
-    ulong pkt_no_conn_cnt[4];       /* number of packets with unknown conn ID (initial, retry, hs, 1-RTT) */
-    ulong frame_tx_alloc_cnt[3];    /* number of pkt_meta alloc successes, fails for empty pool, fails at conn max */
-    ulong pkt_verneg_cnt;           /* number of QUIC version negotiation packets or packets with wrong version */
-    ulong pkt_retransmissions_cnt;  /* number of pkt_meta retries */
-    ulong initial_token_len_cnt[3]; /* number of Initial packets grouped by token length */
+    ulong pkt_net_hdr_err_cnt;                                               /* number of packets dropped due to weird IPv4/UDP headers */
+    ulong pkt_quic_hdr_err_cnt;                                              /* number of packets dropped due to weird QUIC header */
+    ulong pkt_undersz_cnt;                                                   /* number of QUIC packets dropped due to being too small */
+    ulong pkt_oversz_cnt;                                                    /* number of QUIC packets dropped due to being too large */
+    ulong pkt_decrypt_fail_cnt[ FD_METRICS_ENUM_QUIC_ENC_LEVEL_CNT ];        /* number of packets that failed decryption due to auth tag */
```
nbridge-jump, line 332: Number of encryption levels shouldn't come from metrics. Use FD_QUIC_NUM_ENC_LEVELS

nbridge-jump: Maybe we should ASSERT somewhere that they are equal

nbridge-jump: Actually - FD_METRICS_ENUM_QUIC_ENC_LEVEL_CNT is fine. Ignore me

akhinvasara-jumptrading: i like the idea of adding an ASSERT to check they are equal
would be gr8 to also ASSERT that for FD_METRICS_ENUM_QUIC_CONN_STATE_CNT and FD_QUIC_CONN_STATE_CNT (in fd_quic_conn.h)

intrigus-lgtm: Done.

PR #6545, opened 2025-09-29
replay: block idx mapping patch

ibhatt-jumptrading:


src/discof/replay/fd_replay_tile.c
```diff
@@ -1431,8 +1431,15 @@ process_fec_set( fd_replay_tile_t * ctx,
     /* If an entry already exists for this bank index in the block id
        map, we can safely remove it and replace it with the new entry.
        This is safe because we know that the old entry for this fork
-       index has already been pruned away. */
-    if( fd_block_id_map_ele_query( ctx->block_id_map, &block_id_ele->block_id, NULL, ctx->block_id_arr ) ) {
+       index has already been pruned away.
+
+       If the fork index of the query doesn't match the fork index of
+       the element, this likely means that the block id for the array
+       element was set in some previous run of the program that
```
mmcgee-jump: Why would we allow this? That's not a valid error condition.

mmcgee-jump: You should initialize these to "int initialized = 0;" and check that to prevent this. It is not valid to reference memory from a prior run, which actually could still be valid and cause even worse corruption

src/discof/replay/fd_replay_tile.c
```diff
@@ -1432,10 +1435,14 @@ process_fec_set( fd_replay_tile_t * ctx,
        map, we can safely remove it and replace it with the new entry.
        This is safe because we know that the old entry for this fork
        index has already been pruned away. */
-    if( fd_block_id_map_ele_query( ctx->block_id_map, &block_id_ele->block_id, NULL, ctx->block_id_arr ) ) {
+    if( block_id_ele->initialized && fd_block_id_map_ele_query( ctx->block_id_map, &block_id_ele->block_id, NULL, ctx->block_id_arr ) ) {
```
mmcgee-jump: FD_LIKELY

src/discof/replay/fd_replay_tile.c
```diff
@@ -2019,6 +2026,10 @@ unprivileged_init( fd_topo_t *      topo,
   ctx->block_id_map = fd_block_id_map_join( fd_block_id_map_new( block_id_map_mem, chain_cnt, 999UL ) );
   FD_TEST( ctx->block_id_map );
 
+  for( ulong i=0UL; i<tile->replay.max_live_slots; i++ ) {
+    ctx->block_id_arr[ i ].initialized = 0;
```
mmcgee-jump: We typically would reuse. e.g. slot=ULONG_MAX to save memory but doesn't really matter in this case

ibhatt-jumptrading: done

src/discof/replay/fd_replay_tile.c
ibhatt-jumptrading: done

PR #6548, opened 2025-09-29
simulate: blockstore-based backtest to support forked replay

nipatel-jump:
## New Simulate Subcommand and Feccap

The simulate subcommand provides a specialized environment similar to backtest using captured live data. It replays FEC sets from persistent storage previously captured shred information by a new feccap tile.

- **Minimal Topology**: _Uses only essential tiles (simulate, snap, replay, exec, writer, tower)_
- **Replay for Forks**: _Allows for testing over forks, and interleaved execution_

### Known Issues
~~Not being backpressured because of problems that exist in the replay tile due to TOCTOU conditions.~~ 
Fixed by Ishan: a916c48bc5e4a14dbac579e1986d2d300bde67ee


src/app/firedancer-dev/commands/simulate.c
```diff
@@ -0,0 +1,367 @@
...
+
+static void
+simulate_topo( config_t * config ) {
+  ulong exec_tile_cnt   = config->firedancer.layout.exec_tile_count;
+
+  int solcap_enabled = strlen( config->capture.solcap_capture )>0;
+
+  fd_topo_t * topo = { fd_topob_new( &config->topo, config->name ) };
+  topo->max_page_size = fd_cstr_to_shmem_page_sz( config->hugetlbfs.max_page_size );
+  topo->gigantic_page_threshold = config->hugetlbfs.gigantic_page_threshold_mib << 20;
```
topointon-jump:
```suggestion
  fd_topo_t * topo                           = { fd_topob_new( &config->topo, config->name ) };
  topo->max_page_size                 = fd_cstr_to_shmem_page_sz( config->hugetlbfs.max_page_size );
  topo->gigantic_page_threshold = config->hugetlbfs.gigantic_page_threshold_mib << 20;
```

src/discof/backtest/fd_simulate_tile.c
```diff
@@ -0,0 +1,352 @@
...
+ replay tile will then process the FECs and respond with slot completion
+ messages. These slot completion messages are used to determine when
+ the simulation has completed.
+*/
+
+void
+log_success_error( ctx_t * ctx ) {
+  printf("\033[0;32mSIMULATE: Success, replayed %lu slots and %lu forks\033[0m\n", ctx->max_slot - ctx->min_slot + 1, ctx->fork_counter_cnt );
+  exit(0);
+}
```
topointon-jump, line 104: Use of `printf` here, we should use `FD_LOG_` instead

src/discof/backtest/fd_simulate_tile.c
```diff
@@ -0,0 +1,352 @@
...
+  ctx->current_fec_idx++;
+  ctx->published_slot = fd_ulong_max( ctx->published_slot, shred->slot );
+  *charge_busy = 1;  /* Keep processing */
+}
+
+FD_FN_CONST static inline ulong
+scratch_align( void ) {
+  ulong ctx_align = alignof( ctx_t );
+  ulong pool_align = fd_fork_counter_pool_align();
+  ulong map_align = fd_fork_counter_map_align();
```
topointon-jump: nit: align

src/discof/backtest/fd_simulate_tile.c
```diff
@@ -0,0 +1,352 @@
...
+    return;
+  }
+
+  fd_shred_t * shred = (fd_shred_t *)fd_type_pun( &fec_msg.chunk );
+
+  /* Read the fd_store_fec_t struct */
+  fd_store_fec_t store_fec;
+  if( FD_UNLIKELY( fread( &store_fec, sizeof(fd_store_fec_t), 1, ctx->input_file ) != 1 ) ) {
+    FD_LOG_ERR(( "Failed to read fd_store_fec_t at index %lu", ctx->current_fec_idx ));
+    *charge_busy = 1;
```
topointon-jump: `FD_LOG_ERR` will kill the process, so the `*charge_busy = 1;` won't do anything.

src/app/firedancer-dev/commands/simulate.c
```diff
@@ -0,0 +1,367 @@
+/*
+
+The simulate command spawns a smaller topology for simulating the behavior given blockstore states.
```
topointon-jump, line 3: Maybe state explicitly here that `simulate` should be used in conjunction with `feccap`. I also wonder if we should call this `feccap_simulate` or something, `simulate` is a bit vague and we have several other tools this name could feasibly apply to.

src/discof/backtest/fd_simulate_tile.c
```diff
@@ -0,0 +1,352 @@
...
+
+  /* Insert into store following shred tile pattern */
+  long shacq_start, shacq_end, shrel_end;
+  fd_store_fec_t * fec = NULL;
+  FD_STORE_SHARED_LOCK( ctx->store, shacq_start, shacq_end, shrel_end ) {
+    fec = fd_store_insert( ctx->store, 0, (fd_hash_t *)fd_type_pun( &store_fec.key.mr ) );
+  } FD_STORE_SHARED_LOCK_END;
+
+  fd_memcpy( fec->data, store_fec.data, store_fec.data_sz );
+  fec->data_sz = store_fec.data_sz;
```
topointon-jump, line 202: Race condition: need to hold the shred lock whilst accessing the `fec` pointer. See `fd_store_insert` documentation.

src/discof/shredcap/fd_feccap_tile.c
```diff
@@ -0,0 +1,282 @@
...
+    if( fread( &ctx->fecs_count, sizeof(ulong), 1, ctx->feccap_file ) != 1 ) {
+      FD_LOG_WARNING(( "FECCAP: Failed to read existing count, starting from 0" ));
+      ctx->fecs_count = 0UL;
+    }
+    fseek( ctx->feccap_file, 0, SEEK_END );  /* Seek to end for appending */
+    FD_LOG_NOTICE(( "FECCAP: Opened existing capture file %s (current count: %lu)", feccap_file_path, ctx->fecs_count ));
+  }
+
+  /* Set global context for cleanup and register cleanup function */
+  g_feccap_ctx = ctx;
```
topointon-jump: Is this safe? How do we know that `ctx` will still be valid when `atexit` is called? If it's not valid we could get a UAF bug.

src/discof/backtest/fd_simulate_tile.c
topointon-jump: Silly question: is there a guarantee that `store_fec.data_sz` is no larger than `FD_STORE_DATA_MAX`?

src/flamenco/runtime/fd_runtime.c
```diff
@@ -969,10 +969,9 @@ fd_runtime_finalize_account( fd_funk_t *               funk,
 
 /* fd_runtime_buffer_solcap_account_update buffers an account
    update event message in the capture context, which will be
-   sent to the replay tile via the exec_replay link.
-   This buffering is done to avoid passing stem down into the runtime.
+   sent to the replay tile via the exec_replay link. This buffering is
+   done to avoid passing stem down into the runtime. */
 
-   TODO: remove this when solcap v2 is here. */
```
topointon-jump, line 757: why remove this TODO?

src/discof/shredcap/fd_feccap_tile.c
nipatel-jump: yeah this was still being tested, why i brought up the flushing issue earlier. do you have any good solutions for this?

src/discof/backtest/fd_simulate_tile.c
nipatel-jump: 1. I don't think this is true. The lock is taken on the store to produce a new pointer for the fd_store_fec_t. However, once it is produced, the only user of it is the simulate_tile until it publishes the fec message out to the replay tile. This follows the usage in fd_shred_tile:1038 

2. I think in theory it can if there is corrupted data. However, we are pulling the data off of the store and using the data sz from the shred tile notification. Therefore, both are being created by the shred tile and should be consistent.

I could be wrong about either or both of these though, let me know if you disagree.

src/discof/backtest/fd_simulate_tile.c
nipatel-jump: Use this for just clean printing, exit code should be picked up by the future CI test, rather than using the . For now I will revert it back to FD_LOG_NOTICE though

topointon-jump (commented):
We probably want to add some sort of artificial delay (similar to how the generic archiver/replay tiles work), so that the replay behaviour mirrors what happened live, but we can add this later.

PR #6561, opened 2025-09-30
replay: backpressure bug with reasm and sched

ibhatt-jumptrading:


yufeng-jump (dismissed):

src/discof/replay/fd_replay_tile.c
```diff
@@ -1326,6 +1326,9 @@ replay( fd_replay_tile_t *  ctx,
   if( FD_LIKELY( ctx->exec_ready_bitset ) ) {
     fd_sched_txn_ready_t ready_txn[ 1 ];
     if( FD_UNLIKELY( !fd_sched_txn_next_ready( ctx->sched, ready_txn ) ) ) {
+      if( FD_UNLIKELY( fd_reasm_full( ctx->reasm ) ) ) {
+        FD_LOG_ERR(( "unable to buffer new transactions and nothing available to execute.  This likely indicates that the reasm needs to be resized." ));
```
mmcgee-jump: ?? wouldn't this just indicate all exec tiles are busy?

ibhatt-jumptrading: yeah this is bugged, should check that all exec tiles are free, not just that some are free. I'm actually just going to delete this entirely since you can have a race around advancing root. If this condition is true but the root can still be advanced (some other tile still has a refcnt on a bank < consensus root or tower is in the process of sending a new consensus root then we shouldn't halt.

PR #6575, opened 2025-10-01
[WIP] sbpf: elf loader rewrite

mjain-jump:


src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -246,32 +282,21 @@ fd_sbpf_program_new( void *                     prog_mem,
...
    */
 
 int
 fd_sbpf_program_load( fd_sbpf_program_t *             prog,
                       void const *                    bin,
                       ulong                           bin_sz,
                       fd_sbpf_syscalls_t *            syscalls,
-                      fd_sbpf_loader_config_t const * config );
+                      fd_sbpf_loader_config_t const * config,
+                      ulong *                         opt_out_entrypoint );
```
0x0ece: docs? I'd personally put it inside `fd_sbpf_program_t` to avoid changing the fn signature, but no strong opinion

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -285,11 +310,12 @@ fd_sbpf_program_delete( fd_sbpf_program_t * program );
    Always returns a valid cstr, though the content is undefined in case
    the last call to `fd_sbpf_program_load` returned zero (success). */
 
-char const *
-fd_sbpf_strerror( void );
-
 /* SIMD-0189 */
 static inline int fd_sbpf_enable_stricter_elf_headers( ulong sbpf_version ) { return sbpf_version >= FD_SBPF_V3; }
+/* SIMD-0178 */
+static inline int fd_sbpf_static_syscalls            ( ulong sbpf_version ) { return sbpf_version >= FD_SBPF_V3; }
```
0x0ece: I moved `fd_sbpf_enable_stricter_elf_headers` here from `fd_vm_private.h` (and change it from macro to static inline as per richie recommendation).

Maybe we should unify/avoid duplicated definitions, for example now we have both `FD_VM_SBPF_STATIC_SYSCALLS` and `fd_sbpf_static_syscalls`.

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -852,534 +407,358 @@ fd_sbpf_load_dynamic( fd_sbpf_loader_t *         loader,
...
+      /* Bounds check before writing to the rodata. */
+      if( FD_UNLIKELY( fd_ulong_sat_add( imm_low_offset, 4UL /* BYTE_LENGTH_IMMEDIATE */ )>elf_sz ) ) {
+        return FD_SBPF_ELF_ERR_VALUE_OUT_OF_BOUNDS;
+      }
 
-  /* Offsets of implicit addend (immediate fields) */
-  ulong A_off_lo = r_offset+ 4UL;
-  ulong A_off_hi = r_offset+12UL;
+      /* Write back */
+      FD_STORE( uint, rodata+imm_low_offset, (uint)addr );
```
0x0ece: very ro-data 😆

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -1445,17 +824,588 @@ fd_sbpf_elf_peek_strict( fd_sbpf_elf_info_t *            info,
...
-                          ulong                           bin_sz,
-                          fd_sbpf_loader_config_t const * config ) {
-  // FIXME
-  fd_sbpf_elf_info_t * res = fd_sbpf_elf_peek_old( info, bin, bin_sz, config->elf_deploy_checks, config->sbpf_min_version, config->sbpf_max_version );
+fd_sbpf_lenient_elf_parse( fd_sbpf_elf_info_t * info,
+                           void const *         bin,
+                           ulong                bin_sz ) {
+
+  /* This documents the values that will be set in this function */
+  info->rodata_sz        = (uint)bin_sz; // FIXME
```
0x0ece: This is actually a FIXME that I left :D
In theory our old `rodata_sz` was smaller than the full `bin_sz` but I think that to compute it correctly we need some pieces from sbpf_load (that I didn't have while I was writing this code).

I think we should prob remove the fixme, if we can compute it correctly great, otherwise it'll be a bit bigger, not the end of the world.

0x0ece (changes requested):
Main thing I'd say is to squash all commits

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: added some docs in the header file. i think i'll leave the signature as i currently have it for now

src/ballet/sbpf/fd_sbpf_loader.c
mjain-jump: reworked it so that elf_info doesn't parse out the` `rodata_sz` because it's calculated within `parse_ro_sections`. I also renamed `rodata_footprint` to `bin_sz` as its a lot clearer and avoids confusion.

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: fixed

0x0ece (approved):

0x0ece (dismissed):

ripatel-fd:
reviewing, please don't merge yet

src/ballet/elf/fd_elf64.h
```diff
@@ -87,7 +87,7 @@ typedef struct fd_elf64_rela_ fd_elf64_rela;
 /* fd_elf64_dyn: Dynamic section entry */
 
 struct __attribute__((packed)) fd_elf64_dyn_ {
-  long d_tag;
+  ulong d_tag;
```
ripatel-fd, line 92: `Elf64_Sxword    d_tag;`
per spec this is signed, please revert

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -117,15 +145,11 @@ typedef struct fd_sbpf_syscalls fd_sbpf_syscalls_t;
    to fully load the program. */
 
 struct fd_sbpf_elf_info {
-  uint text_off;    /* File offset of .text section (overlaps rodata segment) */
-  uint text_cnt;    /* Instruction count */
-  ulong text_sz;    /* Length of text segment */
+  ulong bin_sz;   /* size of ELF binary */
```
ripatel-fd, line 148: This seems like a performance regression.
It's typical for deployers to over-allocate the programdata account. The "file size" (bin_sz here according to documentation) would have unused zero padding. This is why we derive the rodata segment size, so we can truncate those trailing zeros in the program cache.

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -134,27 +158,29 @@ struct fd_sbpf_elf_info {
...
 
+  /* Dynamic relocation table entries */
+  uint dt_reloff; /* File offset of dynamic relocation table */
+  uint dt_relsz;  /* Number of dynamic relocation table entries */
+
   uint entry_pc;  /* Program counter of entry point
                      NOTE: MIGHT BE OUT OF BOUNDS! */
 
-  /* Bitmap of sections to be loaded (LSB => MSB) */
-  ulong loaded_sections[ 1024UL ];
```
ripatel-fd, line 145: Thanks for getting rid of that

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -35,20 +36,47 @@
...
+#define FD_SBPF_ELF_ERR_WRITABLE_SECTION_NOT_SUPPORTED       (-14)
+#define FD_SBPF_ELF_ERR_ADDRESS_OUTSIDE_LOADABLE_SECTION     (-15)
+#define FD_SBPF_ELF_ERR_INVALID_VIRTUAL_ADDRESS              (-16)
+#define FD_SBPF_ELF_ERR_UNKNOWN_RELOCATION                   (-17)
+#define FD_SBPF_ELF_ERR_FAILED_TO_READ_RELOCATION_INFO       (-18)
+#define FD_SBPF_ELF_ERR_WRONG_TYPE                           (-19)
+#define FD_SBPF_ELF_ERR_UNKNOWN_SYMBOL                       (-20)
+#define FD_SBPF_ELF_ERR_VALUE_OUT_OF_BOUNDS                  (-21)
 #define FD_SBPF_ELF_ERR_UNSUPPORTED_SBPF_VERSION             (-22)
+#define FD_SBPF_ELF_ERR_INVALID_PROGRAM_HEADER               (-23)
```
ripatel-fd, line 64: Why are these negative? (not really a problem just breaks convention ... rust uses positive ints internally to represent enum variants)

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -134,27 +158,29 @@ struct fd_sbpf_elf_info {
...
   int shndx_dyn;
   int shndx_dynstr;
+  int shndx_dynsymtab; /* Section header index of the dynamic symbol table */
 
   /* Known program header indices (like shndx_*) */
   int phndx_dyn;
 
+  /* Dynamic relocation table entries */
+  uint dt_reloff; /* File offset of dynamic relocation table */
+  uint dt_relsz;  /* Number of dynamic relocation table entries */
```
ripatel-fd:
```suggestion
  uint dt_rel_off; /* File offset of dynamic relocation table */
  uint dt_rel_sz;  /* Number of dynamic relocation table entries */
```

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -164,27 +190,36 @@ struct __attribute__((aligned(32UL))) fd_sbpf_program {
...
   /* text section within rodata segment */
   ulong * text;
   ulong   text_cnt;  /* instruction count */
   ulong   text_off;  /* instruction offset for use in CALL_REG instructions */
-  ulong   text_sz;   /* size of text segment */
+  ulong   text_sz;   /* size of text segment. Guaranteed to be <= bin_sz. */
   ulong   entry_pc;  /* entrypoint PC (at text[ entry_pc - start_pc ]) ... FIXME: HMMMM ... CODE SEEMS TO USE TEXT[ ENTRY_PC ] */
 
-  /* Bit vector of valid call destinations (bit count is rodata_sz) */
+  /* Bit vector of valid call destinations (bit count is text_sz). */
```
ripatel-fd: bit count should be `text_cnt`, not `text_sz` (since you can only call 8-byte aligned, we can have an 8x denser bitset here)

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -164,27 +190,36 @@ struct __attribute__((aligned(32UL))) fd_sbpf_program {
...
   fd_sbpf_calldests_t * calldests;
 };
 typedef struct fd_sbpf_program fd_sbpf_program_t;
 
 struct fd_sbpf_loader_config {
-  int elf_deploy_checks;
+  union {
+   int elf_deploy_checks;
+   int reject_broken_elfs;
+  };
```
ripatel-fd, line 230: This union seems unnecessary, please remove this

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -164,27 +190,36 @@ struct __attribute__((aligned(32UL))) fd_sbpf_program {
...
   uint sbpf_min_version;
   uint sbpf_max_version;
-  int enable_symbol_and_section_labels;
+  int  enable_symbol_and_section_labels;
+
+  /* TODO: The optimize_rodata flag, at the time of writing this
+     comment, seems to be hardcoded to false in Agave and not fully
+     implemented in FD. If this changes, any usages of this flag should
+     be revisited and tested appropriately. */
+  int  optimize_rodata;
```
ripatel-fd: can we just remove it?

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -164,27 +190,36 @@ struct __attribute__((aligned(32UL))) fd_sbpf_program {
...
 struct fd_sbpf_loader_config {
-  int elf_deploy_checks;
+  union {
+   int elf_deploy_checks;
+   int reject_broken_elfs;
+  };
   uint sbpf_min_version;
   uint sbpf_max_version;
-  int enable_symbol_and_section_labels;
+  int  enable_symbol_and_section_labels;
```
ripatel-fd:
```suggestion
  uint  enable_symbol_and_section_labels : 1;
```

prefer flags for booleans

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -241,55 +276,66 @@ fd_sbpf_program_new( void *                     prog_mem,
...
    */
 
 int
 fd_sbpf_program_load( fd_sbpf_program_t *             prog,
                       void const *                    bin,
                       ulong                           bin_sz,
                       fd_sbpf_syscalls_t *            syscalls,
-                      fd_sbpf_loader_config_t const * config );
+                      fd_sbpf_loader_config_t const * config,
+                      ulong *                         opt_out_entrypoint );
```
ripatel-fd: Don't add a separate out argument here, just save it in sbpf_program

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -241,55 +276,66 @@ fd_sbpf_program_new( void *                     prog_mem,
...
-fd_sbpf_strerror( void );
+/* SIMD-0178 */
+static inline int fd_sbpf_static_syscalls            ( ulong sbpf_version ) { return sbpf_version>=FD_SBPF_V3; }
+static inline int fd_sbpf_enable_elf_vaddr           ( ulong sbpf_version ) { return sbpf_version!=FD_SBPF_V0; }
+static inline int fd_sbpf_reject_rodata_stack_overlap( ulong sbpf_version ) { return sbpf_version!=FD_SBPF_V0; }
 
 /* SIMD-0189 */
-static inline int fd_sbpf_enable_stricter_elf_headers( ulong sbpf_version ) { return sbpf_version >= FD_SBPF_V3; }
+static inline int fd_sbpf_enable_stricter_elf_headers( ulong sbpf_version ) { return sbpf_version>=FD_SBPF_V3; }
+static inline int fd_sbpf_enable_lower_bytecode_vaddr( ulong sbpf_version ) { return sbpf_version>=FD_SBPF_V3; }
```
ripatel-fd: don't love the naming here, would call this `fd_sbpf_opt_...` or something to clarify that these check for options.
It is unclear by the function name whether `fd_sbpf_swap_sub_reg_imm_operands` checks whether the "swap_sub_reg_imm_operands" feature is enabled, or whether this actually mutates an instruction by swapping operands ...

src/ballet/sbpf/fuzz_sbpf_loader.c
```diff
@@ -42,7 +42,7 @@ LLVMFuzzerTestOneInput( uchar const * data,
 
   /* Allocate objects */
 
-  void * rodata = malloc( info.rodata_footprint );
+  void * rodata = malloc( info.bin_sz );
```
ripatel-fd, line 45: regression ... this malloc is now a lot larger for programs with trailing zero bytes

src/ballet/sbpf/fuzz_sbpf_loader.c
```diff
@@ -55,7 +55,7 @@ LLVMFuzzerTestOneInput( uchar const * data,
     fd_sbpf_syscalls_insert( syscalls, (ulong)*x );
 
   /* Load program */
-  int res = fd_sbpf_program_load( prog, data, size, syscalls, &config );
+  int res = fd_sbpf_program_load( prog, data, size, syscalls, &config, NULL );
```
ripatel-fd: Suggest reverting this API change

src/flamenco/runtime/program/fd_program_cache.c
```diff
@@ -31,7 +31,7 @@ fd_program_cache_entry_new( void *                     mem,
   cache_entry->calldests_shmem_off = l;
 
   /* rodata backing memory */
-  l = FD_LAYOUT_APPEND( l, fd_sbpf_calldests_align(), fd_sbpf_calldests_footprint(elf_info->rodata_sz/8UL) );
+  l = FD_LAYOUT_APPEND( l, fd_sbpf_calldests_align(), fd_sbpf_calldests_footprint(elf_info->text_sz/8UL) );
```
ripatel-fd: use text_cnt instead

ripatel-fd: also, are you sure this behaves correctly if the .text section does not start at virtual address 0? is the whole bitset offset now?

src/flamenco/runtime/program/fd_program_cache.c
```diff
@@ -336,7 +336,7 @@ fd_program_cache_validate_sbpf_program( fd_bank_t *                bank,
   }
 
   /* FIXME: Super expensive memcpy. */
-  fd_memcpy( fd_program_cache_get_calldests_shmem( cache_entry ), prog->calldests_shmem, fd_sbpf_calldests_footprint( prog->rodata_sz/8UL ) );
+  fd_memcpy( fd_program_cache_get_calldests_shmem( cache_entry ), prog->calldests_shmem, fd_sbpf_calldests_footprint( prog->text_sz/8UL ) );
```
ripatel-fd: This is silly indeed, but I have a fix for it in my "fd_progcache" refactor #6608

src/flamenco/runtime/tests/fd_elf_harness.c
```diff
@@ -1,8 +1,15 @@
 #include "fd_solfuzz.h"
+#include "fd_solfuzz_private.h"
 #include "generated/elf.pb.h"
 #include "../../../ballet/sbpf/fd_sbpf_loader.h"
+#include "../program/fd_bpf_loader_program.h"
 #include "../../vm/fd_vm_base.h"
 
+#define SORT_NAME        sort_uint64_t
+#define SORT_KEY_T       uint64_t
```
ripatel-fd: use `ulong`, regrettably we don't use stdint (but let's conform to existing code style)

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -89,455 +52,53 @@ typedef union fd_sbpf_elf fd_sbpf_elf_t;
...
+   header offset and offset + size respectively. Returns 0 (None)
+   otherwise, and sets both (lo, hi) to 0 (the default values for a
+   Rust Range type).
+
+   https://github.com/anza-xyz/sbpf/blob/v0.12.2/src/elf_parser/mod.rs#L87-L93 */
+
+static uchar
+fd_shdr_get_file_range( fd_elf64_shdr const * shdr,
+                        ulong *               lo,
+                        ulong *               hi ) {
```
ripatel-fd: Don't use `uchar` as boolean types (due to messiness with implicit int conversions)
Instead, suggest using a wrapper type for compound ranges

```
struct fd_sbpf_range {
  ulong lo;
  ulong hi;
};
typedef struct fd_sbpf_range fd_sbpf_range_t;

static fd_sbpf_range_t *
fd_shdr_get_file_range( fd_elf64_shdr const * shdr,
                        fd_sbpf_range_t * range )
```

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -89,455 +52,53 @@ typedef union fd_sbpf_elf fd_sbpf_elf_t;
...
-  info->entry_pc = (uint)entry_pc;
-
-  if( (info->shndx_dynstr)>=0 ) {
-    fd_elf64_shdr const * shdr_dynstr = &shdr[ info->shndx_dynstr ];
-    ulong sh_offset = shdr_dynstr->sh_offset;
-    ulong sh_size   = shdr_dynstr->sh_size;
-    REQUIRE( (sh_offset+sh_size>=sh_offset) & (sh_offset+sh_size<=info->rodata_footprint) );
-    info->dynstr_off = (uint)sh_offset;
-    info->dynstr_sz  = (uint)sh_size;
+fd_sbpf_elf_parser_err_to_elf_err( int err ) {
```
ripatel-fd, line 91: .

ripatel-fd: also, would use a switch table here

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -572,7 +133,7 @@ fd_sbpf_program_footprint( fd_sbpf_elf_info_t const * info ) {
   }
   return FD_LAYOUT_FINI( FD_LAYOUT_APPEND( FD_LAYOUT_APPEND( FD_LAYOUT_INIT,
     alignof(fd_sbpf_program_t), sizeof(fd_sbpf_program_t) ),
-    fd_sbpf_calldests_align(), fd_sbpf_calldests_footprint( info->rodata_sz / 8UL ) ),  /* calldests bitmap */
+    fd_sbpf_calldests_align(), fd_sbpf_calldests_footprint( info->text_sz/8UL ) ),  /* calldests bitmap */
```
ripatel-fd:
```suggestion
    fd_sbpf_calldests_align(), fd_sbpf_calldests_footprint( info->text_cnt ) ),  /* calldests bitmap */
```

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -644,156 +207,150 @@ fd_sbpf_program_delete( fd_sbpf_program_t * mem ) {
...
-  ulong dt_symtab;
-
-  /* Dynamic symbols */
-  uint dynsym_off;  /* File offset of .dynsym section (0=missing) */
-  uint dynsym_cnt;  /* Symbol count */
-
-  int elf_deploy_checks;
+  ulong *              calldests;             /* owned by program */
+  fd_sbpf_syscalls_t * syscalls;              /* owned by caller */
+  ulong                registered_entrypoint; /* unset if ULONG_MAX */
```
ripatel-fd: specify in the comment that this is a program counter (if I understand correctly)

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -644,156 +207,150 @@ fd_sbpf_program_delete( fd_sbpf_program_t * mem ) {
...
-  fd_elf64_shdr const * shdrs = (fd_elf64_shdr const *)( elf->bin + elf->ehdr.e_shoff );
-  fd_elf64_phdr const * phdrs = (fd_elf64_phdr const *)( elf->bin + elf->ehdr.e_phoff );
+/* Queries a single string from a section which is marked as SHT_STRTAB.
+   Returns an ElfParserError on failure and 0 on success.
+   https://github.com/anza-xyz/sbpf/blob/v0.12.2/src/elf_parser/mod.rs#L467-L496 */
+int
+fd_sbpf_lenient_get_string_in_section( char *                string,
+                                       fd_elf64_shdr const * section_header,
+                                       uint                  offset_in_section,
+                                       ulong                 maximum_length,
```
ripatel-fd, line 237: per convention, try to keep pointer and size close together and similarly named

e.g. 

```
char * str,
ulong  str_max
```

instead of `string`, `maximum_length`. 

Also document whether this is a cstr (null terminated) or something else

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -644,156 +207,150 @@ fd_sbpf_program_delete( fd_sbpf_program_t * mem ) {
...
+                                       ulong                 bin_sz ) {
+  /* This could be checked only once outside the loop, but to keep the code the same...
+     https://github.com/anza-xyz/sbpf/blob/v0.12.2/src/elf_parser/mod.rs#L474-L476 */
+  if( FD_UNLIKELY( section_header->sh_type != FD_ELF_SHT_STRTAB ) ) {
+    return FD_SBPF_ELF_PARSER_ERR_INVALID_SECTION_HEADER;
+  }
 
-  /* Try first PT_DYNAMIC in program header table */
+  /* default to -1 because length check is done checking for 0... */
+  memset( string, -1, maximum_length );
```
ripatel-fd: This seems very messy. Why do we need this memset at all? 

If anything make if `memset( string, 0xFF, ... )` 

or use fd_msan_poison to mark it as uninitialized, such that instrumentation detects read of uninitialized data

ripatel-fd: I see, because it's used in the string matching below ... Surely there's another way to write this?

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -644,156 +207,150 @@ fd_sbpf_program_delete( fd_sbpf_program_t * mem ) {
...
-      loader->dyn_cnt = (uint)(dyn_sz / sizeof(fd_elf64_dyn));
-      return 0;
+  /* https://github.com/anza-xyz/sbpf/blob/v0.12.2/src/elf_parser/mod.rs#L486-L495 */
+  int found = 0;
+  for( ulong j=0; j<maximum_length; j++ ) {
+    if( string[ j ] == 0x00 ) {
+      found = 1;
+      break;
     }
   }
```
ripatel-fd: Use [memchr](https://man7.org/linux/man-pages/man3/memchr.3.html) (generates better code)

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -644,156 +207,150 @@ fd_sbpf_program_delete( fd_sbpf_program_t * mem ) {
...
+  if( FD_UNLIKELY( fd_sbpf_syscalls_query( loader->syscalls, pc_hash, NULL ) ) ) {
+    return FD_SBPF_ELF_ERR_SYMBOL_HASH_COLLISION;
+  }
 
-    loader->dynsym_off = (uint)sh_offset;
-    loader->dynsym_cnt = (uint)(sh_size/sizeof(fd_elf64_sym));
+  /* Insert the target PC into the calldests set if it's not the
+     entrypoint. */
+  if( FD_LIKELY( !is_entrypoint ) ) {
+    fd_sbpf_calldests_insert( loader->calldests, target_pc );
```
ripatel-fd, line 330: I'm fairly sure this is an out-of-bounds access. 

Suppose you have an ELF that creates this virtual memory layout:

```
0x0000: [ ... junk ... ]
0x8000: .text
0x8010: end of text
```

Then, you get

- `text_sz==16`
- `text_cnt==2`
- but a valid program counter of `0x1000` (0x8000 divided by 8)
- ... but calldests only fits 2 bits (well, 64 bits because of rounding up) ... 

Unless the `.text` section is forced to start at `0x0000`, but I don't think that's the case in all the ELF loaders

ripatel-fd: This is why calldests was rodata segment-sized before, since the rodata segment is forced to start at 0

src/ballet/elf/fd_elf64.h
mjain-jump:
```
#[derive(Debug, Clone)]
#[repr(C)]
pub struct Elf64Dyn {
    pub d_tag: Elf64Xword,
    pub d_val: Elf64Xword,
}
```
this is what I see in Agave's code

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: its stays conformant with our convention of using negative error code values like we do for txn / instr / syscall errors etc

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: fixed

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: they're declared as different variables in agave but maintained the same way. would you prefer to just use one of the variables? makes it slightly more confusing to read imo

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: yea, i think itll get removed soon anyways but wasnt sure at the time

src/ballet/sbpf/fuzz_sbpf_loader.c
mjain-jump: im pretty sure `rodata_footprint` was always just `bin_sz` before
https://github.com/firedancer-io/firedancer/blob/main/src/ballet/sbpf/fd_sbpf_loader.c#L486

mjain-jump: all i did here was change the name from `rodata_footprint` to `bin_sz` and kept everything else the same

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: i agree. this will also all get removed pretty soon by Agave, but can rename for now

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: removed the flag entirely because its not used in prod

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: `rodata_footprint` was just renamed to `bin_sz` so it maintains the same behavior as before.

src/ballet/elf/fd_elf64.h
ripatel-fd: Oh I see, could you add a comment that this deviates from the ELF spec but is there to match Agave?

src/ballet/sbpf/fuzz_sbpf_loader.c
ripatel-fd: I see sgtm

src/ballet/elf/fd_elf64.h
mjain-jump: done

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: fixed

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: fixed

src/ballet/sbpf/fuzz_sbpf_loader.c
mjain-jump: done

src/flamenco/runtime/program/fd_program_cache.c
mjain-jump: fixed, and yes `target_pc` is constrained between `[0, text_cnt]` (see `fd_sbpf_program_relocate`)

src/flamenco/runtime/tests/fd_elf_harness.c
mjain-jump: fixed

src/ballet/sbpf/fd_sbpf_loader.c
mjain-jump: fixed

src/ballet/sbpf/fd_sbpf_loader.c
mjain-jump: fixed

src/ballet/sbpf/fd_sbpf_loader.c
mjain-jump: `target_pc` is constrained between `[0, text_cnt]` (see `fd_sbpf_program_relocate`)

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -151,40 +174,59 @@ typedef struct fd_sbpf_elf_info fd_sbpf_elf_info_t;
...
+  ulong  rodata_sz;  /* size of read-only data */
 
   /* text section within rodata segment */
   ulong * text;
   ulong   text_cnt;  /* instruction count */
   ulong   text_off;  /* instruction offset for use in CALL_REG instructions */
-  ulong   text_sz;   /* size of text segment */
-  ulong   entry_pc;  /* entrypoint PC (at text[ entry_pc - start_pc ]) ... FIXME: HMMMM ... CODE SEEMS TO USE TEXT[ ENTRY_PC ] */
+  ulong   text_sz;   /* size of text segment. Guaranteed to be <= bin_sz. */
+  ulong   entry_pc;  /* entrypoint PC (at text[ entry_pc ]). ULONG_MAX if not set. */
```
topointon-jump, line 217: Can we document that this value can only be set once, to detect entrypoint collisions as described above.

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -241,29 +283,21 @@ fd_sbpf_program_new( void *                     prog_mem,
...
-     new_elf_parser:     true
-     enable_elf_vaddr:   false
-     reject_broken_elfs: elf_deploy_checks
-
-   For documentation on these config params, see:
-   https://github.com/anza-xyz/sbpf/blob/v0.3.0/src/vm.rs#L198
-
-   Solana/Agave equivalent:
-   https://github.com/anza-xyz/sbpf/blob/v0.8.0/src/elf.rs#L361
+   As of writing, this loader is conformant with Solana RBPF v0.12.2.
```
topointon-jump: Can we also list what SBPF versions we are conformant with? (i.e. not V3)

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -607,24 +168,27 @@ fd_sbpf_program_new( void *                     prog_mem,
   FD_SCRATCH_ALLOC_INIT( laddr, prog_mem );
   fd_sbpf_program_t * prog = FD_SCRATCH_ALLOC_APPEND( laddr, alignof(fd_sbpf_program_t), sizeof(fd_sbpf_program_t) );
 
+  /* Note that entry_pc and rodata_sz get set during the loading phase. */
```
topointon-jump, line 177: Do we still need two phases?

mjain-jump: unfortunately we do in order to allocate the calldests bitset

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: done

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -279,17 +314,34 @@ fd_sbpf_program_load( fd_sbpf_program_t *             prog,
 void *
 fd_sbpf_program_delete( fd_sbpf_program_t * program );
 
-/* fd_csv_strerror: Returns a cstr describing the source line and error
-   kind after the last call to `fd_sbpf_program_load` from the same
-   thread returned non-zero.
-   Always returns a valid cstr, though the content is undefined in case
-   the last call to `fd_sbpf_program_load` returned zero (success). */
+/* SBPF versions and features
```
topointon-jump: Can we add a big comment saying we also need to update the definitions in the vm map whenever this mapping changes?

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: done

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: done

src/ballet/sbpf/fd_sbpf_loader.c
mjain-jump: improved and added some more documentation

src/ballet/sbpf/fd_sbpf_loader.c
mjain-jump: fixed, no more memset and im using `memchr` now

src/ballet/sbpf/fd_sbpf_loader.c
mjain-jump: fixed

src/ballet/sbpf/fd_sbpf_loader.h
mjain-jump: renamed

src/ballet/sbpf/fd_sbpf_loader.c
mjain-jump: good idea, done

src/ballet/sbpf/fd_sbpf_loader.c
mjain-jump: done

CharlesMoyesFD (approved):

src/ballet/elf/fd_elf64.h
```diff
@@ -96,4 +98,3 @@ struct __attribute__((packed)) fd_elf64_dyn_ {
 typedef struct fd_elf64_dyn_ fd_elf64_dyn;
 
 #endif /* HEADER_fd_src_ballet_elf_fd_elf64_h */
```
cmoyes-jump, line 100: ultra-nit: newline

cmoyes-jump: checked offline, the UI is suppressing the extra newline

PR #6577, opened 2025-10-01
resolv: replay integration fixes

ibhatt-jumptrading:


src/discof/resolv/fd_resolv_tile.c
```diff
@@ -206,6 +206,10 @@ before_frag( fd_resolv_ctx_t * ctx,
 
   if( FD_UNLIKELY( ctx->in[in_idx].kind==IN_KIND_REPLAY ) ) return 0;
 
+  /* We are not ready to process alut lookups until we have a valid
```
mmcgee-jump: You cannot just drop TPU transactions like this without a corresponding metric and updates to all the sankey/gui code. Also this isn't the right spot for this. See `NO_BANK_DROP` for how this is done in frankendancer.

mmcgee-jump (changes requested):
.

src/discof/resolv/fd_resolv_tile.c
```diff
@@ -253,23 +253,25 @@ publish_txn( fd_resolv_ctx_t *          ctx,
   txnm->reference_slot = ctx->flushing_slot;
 
   if( FD_UNLIKELY( txnt->addr_table_adtl_cnt ) ) {
+    if( FD_UNLIKELY( !ctx->bank ) ) {
+      FD_MCNT_INC( RESOLV, NO_BANK_DROP, 1 );
```
mmcgee-jump: This is confusingly, not a real metric in this tile, which is `RESOLF` for metrics purposes, not `RESOLV`

mmcgee-jump (changes requested):
.

src/discof/resolv/fd_resolv_tile.c
ibhatt-jumptrading: adding resolf metric for this

mmcgee-jump (approved):

PR #6579, opened 2025-10-01
funk: fix txn_remove_published

ripatel-fd:
Fixes a broken map iterator and adds a unit test


ripatel-fd:
credit to @cali-jumptrading for reporting.

PR #6590, opened 2025-10-02
snapshots: fix snapshot parsing

cali-jumptrading:
Only process the manifest once both the status cache and manifest are parsed in snapin.

### DIFF
--- a/src/discof/restore/fd_snapin_tile.c
+++ b/src/discof/restore/fd_snapin_tile.c
@@ -79,7 +79,14 @@ struct fd_snapin_tile {
   fd_ssparse_t *           ssparse;
   fd_ssmanifest_parser_t * manifest_parser;
   fd_slot_delta_parser_t * slot_delta_parser;
-  ulong                    bank_slot;
+
+  struct {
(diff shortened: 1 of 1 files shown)


src/discof/restore/fd_snapin_tile.c
```diff
@@ -535,6 +541,10 @@ handle_data_frag( fd_snapin_tile_t *  ctx,
         }
         break;
       }
+      case FD_SSPARSE_ADVANCE_MANIFEST_AND_STATUS_CACHE_DONE: {
+        process_manifest( ctx );
+        break;
+      }
```
amass-jump: I'm not entirely convinced that ssparse should have this extra state, it seems like something that could be stored in the client (snapin ctx) instead. Perhaps other ssparse clients would need different extra states? Seems like you just need to know "have I seen the manifest return code and have I seen the status_cache return code

cali-jumptrading: makes sense! updated.

amass-jump (approved):

PR #6594, opened 2025-10-02
snapshots: disable peer selection when no snapshot sources are selected

cali-jumptrading:


amass-jump (commented):
I'm not convinced we need this config option. It seems like we can just use the combination of

- entrypoints disabled
- all gossip peers disabled
- http servers empty/disabled

to mean the same thing. We should support this in production anyway (the operator is saying they will provide their own snapshot on disk)

src/discof/restore/fd_snaprd_tile.c
```diff
@@ -744,6 +745,11 @@ after_credit( fd_snaprd_tile_t *  ctx,
         ctx->metrics.full.bytes_total = ctx->local_in.full_snapshot_size;
         ctx->state                    = FD_SNAPRD_STATE_READING_FULL_FILE;
       } else {
+        if( FD_UNLIKELY( !ctx->config.do_download ) ) {
+          FD_LOG_ERR(( "Local snapshot `%s` is too old and downloading new snapshots is disabled. "
+                       "Please enable downloading via [snapshots.download] and restart.", ctx->local_in.full_snapshot_path ) );
```
amass-jump, line 750: Do we have any existing precedent for LOG_ERR'ing here? This is a situation where the validator isn't expecting to make progress / start up correctly, but it's not really a "crashable" edge case... 

So is there some precedent for "alert loudly and keep running" or do we prefer to crash? Presumably most operators have their validators running under systemd or whatnot which will auto restart if it crashes/stops, which won't give them any chance to correct the config

amass-jump: Separately not sure how I feel about asking the operator to enable downloading. Maybe they just need to get a newer snapshot some other way external to our code. We should also probably log the local slot, cluster slot, and the max slot age so they can see those things

cali-jumptrading: Agree with logging the local slot, cluster slot, and the max slot age. Not sure if we have precedent for alert loudly and keep going, or crashing. We added this download config option because agave had something similar: `no_snapshot_fetch`, which disabled downloading snapshots and only loaded from local snapshots if present.

PR #6597, opened 2025-10-02
ethtool: add "auto" option to rss_queue_mode

amass-jump:


amass-jump:
Apologies for the large diff, I basically had to rewrite the stage

mmcgee-jump (approved):

PR #6598, opened 2025-10-03
runtime: simplify read-only account queries

ripatel-fd:
Removes speculative query / overrun recovery from read-only funk
queries.


ripatel-fd:
Will just rewrite this with the accdb patch

PR #6601, opened 2025-10-03
shred: exit early when inserting existing merkle root

emwang-jump:


src/disco/store/fd_store.c
```diff
@@ -147,12 +147,15 @@ fd_store_fec_t *
 fd_store_insert( fd_store_t * store,
                  ulong        part_idx,
                  fd_hash_t  * merkle_root ) {
+  if( FD_UNLIKELY( fd_store_query_const( store, merkle_root ) ) ) {
+    FD_LOG_INFO(( "merkle root %s already in store.  This could either "
```
mmcgee-jump: Please do not log a message like "This could be a bug in firedancer" if it's an expected case. Just log what we are doing, e.g. "Dropping incoming shred <SHRED_INFO_HERE_INCLUDING_SLOT_MERKLE_ETC> as it is already in the shred store". Chat with Philip but I'd probably log this at WARNING

src/disco/shred/fd_shred_tile.c
```diff
@@ -1039,6 +1039,17 @@ after_frag( fd_shred_ctx_t *    ctx,
         fec = fd_store_insert( ctx->store, ctx->round_robin_id, (fd_hash_t *)fd_type_pun( &ctx->out_merkle_roots[fset_k] ) );
       } FD_STORE_SHARED_LOCK_END;
 
+      if( FD_UNLIKELY( !fec ) ) {
+        /* fec can be null for several reasons, but the most likely case
+        that Firedancer can run into during regular operation is when it
+        is our leader slot and someone is sending us back our own FEC
+        set shreds. We could end up trying to insert our own FEC set
```
mmcgee-jump: Style nit but kbowers style comments are always double spaced, and these following lines should be indented a few more spaces to line up with word "fec"

mmcgee-jump (dismissed):
.

src/disco/shred/fd_shred_tile.c
```diff
@@ -1039,6 +1039,17 @@ after_frag( fd_shred_ctx_t *    ctx,
         fec = fd_store_insert( ctx->store, ctx->round_robin_id, (fd_hash_t *)fd_type_pun( &ctx->out_merkle_roots[fset_k] ) );
       } FD_STORE_SHARED_LOCK_END;
 
+      if( FD_UNLIKELY( !fec ) ) {
+        /* fec can be null for several reasons, but the most likely case
```
lidatong, line 1043: are there actually other reasons? i'd just explicitly state store_insert returns null if it's already been inserted

src/discof/forest/fd_forest.c
```diff
@@ -668,6 +668,18 @@ fd_forest_fec_clear( fd_forest_t * forest, ulong slot, uint fec_set_idx, uint ma
...
+  if( FD_UNLIKELY( fd_forest_blk_idxs_test( ele->cmpl, max_shred_idx ) ) ) {
+    /* It's possible the the fec_resolver evicted something that we
+       actually have the completed version of (ex. it was our own leader
+       block, but for some reason someone is also sending us our own
+       block through turbine.  fec_resolver is still going to buffer
+       those fec sets, because our own shreds don't go through
+       fec_resolver.  So then we are in a state where we have received
+       fec_complete messages for all the FECs in our slot, but
+       fec_resolver possibly has some incomplete ctxs for some of those
+       FEC sets, that will eventually get evicted). */
```
```suggestion
    /* It's possible the fec_resolver evicted something that we
       already completed.  One way this can happen is if we produce a FEC set as leader, but for some reason a validator sends us shreds from that same
       FEC set. fec_resolver is still going to buffer
       those fec sets, because our own shreds don't go through
       fec_resolver.  So then we are in a state where we have received
       fec_complete messages for all the FECs in our slot, but
       fec_resolver possibly has some incomplete ctxs for some of those
       FEC sets, that will eventually get evicted). */
```

src/disco/shred/fd_shred_tile.c
emwang-jump: technically if the store is full or corrupt, but its possible we should ERR/CRIT in those cases??

PR #6605, opened 2025-10-03
leader: patch for data cost calculation

ibhatt-jumptrading:


src/discof/bank/fd_bank_tile.c
```diff
@@ -194,7 +194,7 @@ handle_microblock( fd_bank_ctx_t *     ctx,
     ctx->metrics.txn_result[ fd_bank_err_from_runtime_err( err ) ]++;
 
     uint actual_execution_cus = (uint)(txn_ctx->compute_budget_details.compute_unit_limit - txn_ctx->compute_budget_details.compute_meter);
-    uint actual_acct_data_cus = (uint)(txn_ctx->loaded_accounts_data_size);
+    uint actual_acct_data_cus = (uint)(FD_COMPUTE_BUDGET_HEAP_COST * ((txn_ctx->loaded_accounts_data_size + (FD_COMPUTE_BUDGET_ACCOUNT_DATA_COST_PAGE_SIZE-1UL)) / FD_COMPUTE_BUDGET_ACCOUNT_DATA_COST_PAGE_SIZE ));
```
mmcgee-jump: Should you not just fix the `loaded_accounts_data_size` value?

ibhatt-jumptrading: the loaded account data size isn't equal to the CUs for it. We could add a flag in the txn ctx for this value, but imo it would muck up the txn_ctx even more

mmcgee-jump: As a principle I think the runtime functions should compute any necessary value and provide back to the caller. Caller should not need to know details of how the runtime works and do a CU calculation, but this is fine for now

ibhatt-jumptrading: i'll just do it properly

mmcgee-jump (approved):

PR #6607, opened 2025-10-03
repair: remove keyguard client and async all signing tasks

emwang-jump:


src/discof/repair/fd_repair_tile.c
```diff
@@ -187,25 +185,60 @@ typedef struct fd_fec_sig fd_fec_sig_t;
...
-  ulong       next;         /* used internally by fd_map_chain */
+  ulong       key;        /* map key, unique nonce */
   ulong       buflen;
   union {
     uchar           buf[sizeof(fd_repair_msg_t)];
     fd_repair_msg_t msg;
   };
+  fd_pong_data_t  pong_data;
 };
 typedef struct fd_repair_pending_sign fd_repair_pending_sign_t;
```
lidatong: nit but would namespace like

`sign_ping`
`sign_pong`
`sign_req`
`sign_map`
`sign_queue`

(don't need `fd_repair` prefix either since it's `.c`)

PR #6615, opened 2025-10-06
rdisp: nits

intrigus-lgtm:
Implement preconditions of `fd_rdisp_footprint`/`fd_rdisp_new` as documented.
Fix typos.
Fix reference to `fd_rdisp_txn`, `LAST_REF_WAS_WRITE`, `MAX_ACCT_PER_TXN`.
Use `add_txn` instead of (`fd_rdisp_add_txn`) for consistency (all other references are without `fd_rdisp`) with the rest of the code.

### DIFF
--- a/src/discof/replay/fd_rdisp.c
+++ b/src/discof/replay/fd_rdisp.c
@@ -80,7 +80,7 @@
    outgoing edges.  The type is actually a union of bitfield, but C
    bitfields are gross, so we just do it manually with macros.  If the
    high bit is set, that means the transaction storing this edge_t value
-   is it the last in this specific DAG, and the lower 31 bits of the
+   is the last in this specific DAG, and the lower 31 bits of the
    value are an index in the map_pool for the account pubkey for this
    DAG.  See the comments about the hidden edge outgoing from node 7 in
    the DAG at the top of this file for an example.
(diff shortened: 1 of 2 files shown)


src/discof/replay/fd_rdisp.c
```diff
@@ -483,7 +483,8 @@ ulong fd_rdisp_align( void ) { return 128UL; }
 ulong
 fd_rdisp_footprint( ulong depth,
                     ulong block_depth ) {
-  if( FD_UNLIKELY( (depth>FD_RDISP_MAX_DEPTH) | (block_depth>FD_RDISP_MAX_BLOCK_DEPTH) ) ) return 0UL;
+  if( FD_UNLIKELY( (depth>FD_RDISP_MAX_DEPTH)             |  (depth<2) |
+                   (block_depth>FD_RDISP_MAX_BLOCK_DEPTH) | (block_depth<4) ) ) return 0UL;
```
ptaffet-jump: The 2 and 4 should have `UL` suffixes, and the alignment should be better

intrigus-lgtm: Added `UL` suffixes.
Hope the alignment is better now?
I had a look at the contribution guide and the existing codebase, but still not sure whether I got the "correct" alignment.

PR #6628, opened 2025-10-06
[WIP / DO NOT MERGE] flamenco/runtime: Add missing log for conformance

CharlesMoyesFD:
A plethora of outstanding mismatches in sol_vm_syscall_cpi_c_diff were caused by a missing `"Program is not deployed"` log for the `!fd_bpf_upgradeable_loader_state_is_program()` special case. Note that Agave (naturally) excludes built-in programs from this log.

Agave reference (`programs/bpf_loader/src/lib.rs:428`):
```
match &executor.program {
    ProgramCacheEntryType::FailedVerification(_)
    | ProgramCacheEntryType::Closed
    | ProgramCacheEntryType::DelayVisibility => {
        ic_logger_msg!(log_collector, "Program is not deployed");  // <-- THIS IS THE EQUIVALENT LOG
        Err(Box::new(InstructionError::UnsupportedProgramId) as Box<dyn std::error::Error>)
    }
    ProgramCacheEntryType::Loaded(executable) => execute(executable, invoke_context),
    _ => Err(Box::new(InstructionError::UnsupportedProgramId) as Box<dyn std::error::Error>),
}
```

CI will fail until https://github.com/firedancer-io/firedancer/pull/6575 is merged. That is expected. Until then, this PR will remain in a holding pattern with "WIP" status.


src/flamenco/runtime/program/fd_bpf_loader_program.c
```diff
@@ -2478,6 +2479,11 @@ fd_bpf_loader_program_execute( fd_exec_instr_ctx_t * ctx ) {
          Program account and program data account discriminants get checked when loading in program accounts
          into the program cache. If the discriminants are incorrect, the program is marked as closed. */
       if( FD_UNLIKELY( !fd_bpf_upgradeable_loader_state_is_program( program_account_state ) ) ) {
+        uchar migrated_yet;
+        if( !fd_is_non_migrating_builtin_program(program_id) &&
+            !fd_is_migrating_builtin_program(ctx->txn_ctx, program_id, &migrated_yet) ) {
+          fd_log_collector_msg_literal(ctx, "Program is not deployed");
```
ripatel-fd:
```suggestion
        if( !fd_is_non_migrating_builtin_program( program_id ) &&
            !fd_is_migrating_builtin_program( ctx->txn_ctx, program_id, &migrated_yet ) ) {
          fd_log_collector_msg_literal( ctx, "Program is not deployed" );
```

src/flamenco/runtime/program/fd_bpf_loader_program.c
```diff
@@ -2478,6 +2479,11 @@ fd_bpf_loader_program_execute( fd_exec_instr_ctx_t * ctx ) {
          Program account and program data account discriminants get checked when loading in program accounts
          into the program cache. If the discriminants are incorrect, the program is marked as closed. */
       if( FD_UNLIKELY( !fd_bpf_upgradeable_loader_state_is_program( program_account_state ) ) ) {
+        uchar migrated_yet;
+        if( !fd_is_non_migrating_builtin_program( program_id ) &&
+            !fd_is_migrating_builtin_program( ctx->txn_ctx, program_id, &migrated_yet ) ) {
```
mjain-jump: any reason we need these migrating builtin checks?

topointon-jump (approved):

src/flamenco/runtime/program/fd_bpf_loader_program.c
cmoyes-jump: Builtin programs should never log "Program is not deployed". My understanding is that the migration status only affects whether they run as native builtin programs or as BPF programs, but they're still always available.

cmoyes-jump: I realize now that we would not have reached this part of the code if we were migrating, removing

Updated the comment

mjain-jump (approved):

PR #6642, opened 2025-10-07
backtest: skip checksum

kbhargava-jump:


ripatel-fd:
Not a fan of this checksum stuff, why not just compare file size? Then we don't need to make that optional because it doesn't impact performance

kbhargava-jump:
> Not a fan of this checksum stuff, why not just compare file size? Then we don't need to make that optional because it doesn't impact performance

im removing it soon anyways when i remove the script

PR #6643, opened 2025-10-07
Fix firedancer-dev backtest requiring keys (pt. 2)

ripatel-fd:


src/discof/replay/fd_replay_tile.c
```diff
@@ -1161,6 +1162,16 @@ publish_reset( fd_replay_tile_t *  ctx,
   ctx->replay_out->chunk = fd_dcache_compact_next( ctx->replay_out->chunk, sizeof(fd_poh_reset_t), ctx->replay_out->chunk0, ctx->replay_out->wmark );
 }
 
+static ulong
+next_leader_slot( fd_replay_tile_t const * ctx,
```
mmcgee-jump, line 1166: What exactly is this "fixing"? You're just removing functionality from backtest that we actually need to simulate real production performance.

mmcgee-jump (changes requested):
.

ripatel-fd:
> What exactly is this "fixing"? 

@mmcgee-jump I can't run backtest without random unrelated file system accesses to ~/.firedancer/fd1/identity_key.json. This file usually doesn't exist, so I just get a non-descriptive exit code 1 error. 

Backtest does asynchronous single ledger replay without fork choice or block production
The replay tile managing a key in that context makes no sense. 

We should make a new subcommand if we want a topology that exercises block production and fork choice on historical data.

ripatel-fd:
Oh I see, didn't realize backtest is already doing block production. Rewriting this PR

ripatel-fd:
Well, this is still valid. firedancer-dev backtest is never going to get a leader slot unless we also have functionality to forcibly inject ourselves into a leader schedule (and break consensus rules)

PR #6646, opened 2025-10-07
snapshots: remove temp files

cali-jumptrading:
We should not leave temporary files in the user's snapshot directory so we can cleanly run backtest and auto-update ledgers.

### DIFF
--- a/src/discof/restore/fd_snaprd_tile.c
+++ b/src/discof/restore/fd_snaprd_tile.c
@@ -598,6 +598,20 @@ rename_snapshots( fd_snaprd_tile_t * ctx ) {
   }
 }
 
+static void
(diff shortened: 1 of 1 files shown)


mmcgee-jump:
Currently this breaks the sandbox, but even if fixed it weakens it and it's not clear what for? to `cleanly run backtest and auto-update ledgers.` ? Those are not relevant to production so I would not compromise the sandbox for them

cali-jumptrading:
> Currently this breaks the sandbox, but even if fixed it weakens it and it's not clear what for? to `cleanly run backtest and auto-update ledgers.` ? Those are not relevant to production so I would not compromise the sandbox for them

Ah feel free to revert then.

PR #6653, opened 2025-10-07
progcache: support concurrent usage

ripatel-fd:


src/flamenco/progcache/fd_prog_load.c
```diff
@@ -0,0 +1,183 @@
...
+                  ulong *                   out_sz,
+                  fd_funk_txn_xid_t *       out_xid ) {
+  fd_pubkey_t prog_addr = FD_LOAD( fd_pubkey_t, _prog_addr );
+
+  fd_funk_txn_xid_t _out_xid;
+  if( !out_xid ) out_xid = &_out_xid;
+  fd_account_meta_t const * meta = fd_funk_get_acc_meta_readonly(
+      accdb, xid, &prog_addr, NULL, NULL, out_xid );
+  if( FD_UNLIKELY( !meta ) ) return NULL;
+  fd_txn_account_t _rec[1];
```
mjain-jump, line 116: slighter preference towards using `FD_TXN_ACCOUNT_DECL`

src/flamenco/progcache/fd_prog_load.c
```diff
@@ -0,0 +1,183 @@
...
+  fd_txn_account_t * rec = fd_txn_account_join( fd_txn_account_new( _rec, &prog_addr, (void *)meta, 0 ), accdb->wksp );
+  if( FD_UNLIKELY( !rec ) ) FD_LOG_CRIT(( "fd_txn_account_new failed" ));
```
mjain-jump, line 118: i think `fd_txn_account_init_from_funk_readonly` does exactly this already

mjain-jump: ah nvm, i see the wksp is different

mjain-jump: actually, you should be able to use it

src/flamenco/progcache/fd_prog_load.c
```diff
@@ -0,0 +1,183 @@
...
+    elf = fd_get_executable_program_content_for_upgradeable_loader( accdb, xid, rec, out_sz, &sub_xid );
+  } else if( !memcmp( owner, fd_solana_bpf_loader_v4_program_id.key, sizeof(fd_pubkey_t) ) ) {
+    elf = fd_get_executable_program_content_for_v4_loader( rec, out_sz );
+  } else if( !memcmp( owner, fd_solana_bpf_loader_program_id.key, sizeof(fd_pubkey_t) ) ||
+             !memcmp( owner, fd_solana_bpf_loader_deprecated_program_id.key, sizeof(fd_pubkey_t) ) ) {
+    elf = fd_get_executable_program_content_for_v1_v2_loaders( rec, out_sz );
+  }
+
+  if( FD_LIKELY( elf ) ) {
+    /* Return the XID with the last modification slow (unclear if they can differ) */
```
mjain-jump:
```suggestion
    /* Return the XID with the last modification slot (unclear if they can differ) */
```

src/flamenco/progcache/fd_prog_load.c
```diff
@@ -0,0 +1,183 @@
...
+fd_get_executable_program_content_for_v1_v2_loaders( fd_txn_account_t const * program_acc,
+                                                     ulong *                  program_data_len ) {
+  *program_data_len = fd_txn_account_get_data_len( program_acc );
+  return fd_txn_account_get_data( program_acc );
+}
+
+uchar const *
+fd_prog_load_elf( fd_funk_t const *         accdb,
+                  fd_funk_txn_xid_t const * xid,
+                  void const *              _prog_addr,
```
mjain-jump, line 106: why not an `fd_pubkey_t const *`?

src/flamenco/progcache/fd_prog_load.c
```diff
@@ -0,0 +1,183 @@
...
+  } else if( !memcmp( owner, fd_solana_bpf_loader_v4_program_id.key, sizeof(fd_pubkey_t) ) ) {
+    elf = fd_get_executable_program_content_for_v4_loader( rec, out_sz );
+  } else if( !memcmp( owner, fd_solana_bpf_loader_program_id.key, sizeof(fd_pubkey_t) ) ||
+             !memcmp( owner, fd_solana_bpf_loader_deprecated_program_id.key, sizeof(fd_pubkey_t) ) ) {
+    elf = fd_get_executable_program_content_for_v1_v2_loaders( rec, out_sz );
+  }
+
+  if( FD_LIKELY( elf ) ) {
+    /* Return the XID with the last modification slow (unclear if they can differ) */
+    *out_xid = sub_xid.ul[0]>out_xid->ul[0] ? sub_xid : *out_xid;
```
mjain-jump: the last modification slot for the programdata will always be >= the last modification slot for the program, so you can probably omit the usage of `sub_xid` entirely

src/flamenco/progcache/fd_progcache_admin.c
```diff
@@ -0,0 +1,405 @@
...
+/* Algorithm to estimate size of cache metadata structures (rec_pool
+   object pool and rec_map hashchain table).
+
+   FIXME Carefully balance this */
+
+static ulong
+fd_progcache_est_rec_max1( ulong wksp_footprint,
+                           ulong mean_cache_entry_size ) {
+  return wksp_footprint / mean_cache_entry_size;
+}
```
mjain-jump, line 12: OOC whats the purpose of keeping both `fd_progcache_est_rec_max1` and `fd_progcache_est_rec_max`?

src/flamenco/progcache/fd_progcache_rec.c
```diff
@@ -0,0 +1,107 @@
...
+fd_progcache_rec_new( void *                     mem,
+                      fd_sbpf_elf_info_t const * elf_info,
+                      ulong                      load_slot,
+                      fd_features_t const *      features,
+                      void const *               progdata,
+                      ulong                      progdata_sz ) {
+
+  /* Format object */
+
+  int   has_calldests = !fd_sbpf_enable_stricter_elf_headers_enabled( elf_info->sbpf_version );
```
mjain-jump, line 17: calldests can be null if the text section is empty, might want to guard against that too

mjain-jump: (even though this case is caught by the vm validate step)

src/flamenco/progcache/fd_progcache_rec.c
```diff
@@ -0,0 +1,107 @@
...
+  ulong pc_max        = fd_ulong_max( 1UL, elf_info->text_cnt );
```
mjain-jump: doing the above should allow you to remove the `fd_ulong_max` special casing for all callers

src/flamenco/progcache/fd_progcache_rec.c
```diff
@@ -0,0 +1,107 @@
...
+
+  FD_SCRATCH_ALLOC_INIT( l, mem );
+  fd_progcache_rec_t * rec           = FD_SCRATCH_ALLOC_APPEND( l, fd_progcache_rec_align(),  sizeof(fd_progcache_rec_t) );
+  void *               calldests_mem = NULL;
+  if( has_calldests )
```
mjain-jump: put this in `{ }` for readability

src/flamenco/progcache/fd_progcache_rec.c
mjain-jump: you can just port over [these](https://github.com/firedancer-io/firedancer/pull/6647) changes

src/flamenco/progcache/fd_progcache_rec.c
```diff
@@ -0,0 +1,107 @@
...
+    /*               */calldests_mem = FD_SCRATCH_ALLOC_APPEND( l, fd_sbpf_calldests_align(), fd_sbpf_calldests_footprint( pc_max ) );
+  void *               rodata_mem    = FD_SCRATCH_ALLOC_APPEND( l, 8UL,                       elf_info->bin_sz );
+  FD_SCRATCH_ALLOC_FINI( l, fd_progcache_rec_align() );
+  memset( rec, 0, sizeof(fd_progcache_rec_t) );
+  rec->calldests_off = has_calldests ? (uint)( (ulong)calldests_mem - (ulong)mem ) : 0U;
+  rec->rodata_off    = (uint)( (ulong)rodata_mem    - (ulong)mem );
```
mjain-jump: spacing looks a bit weird here

src/flamenco/progcache/fd_prog_load.c
ripatel-fd: This macro shouldn't exist at all

src/flamenco/progcache/fd_progcache_rec.c
```diff
@@ -0,0 +1,107 @@
...
+
+  fd_sbpf_syscalls_t _syscalls[ FD_SBPF_SYSCALLS_SLOT_CNT ];
+  fd_sbpf_syscalls_t * syscalls = fd_sbpf_syscalls_join( fd_sbpf_syscalls_new( _syscalls ) );
+  int syscalls_err = fd_vm_syscall_register_slot( syscalls, load_slot, features, /* is_deploy */ 0 );
+  if( FD_UNLIKELY( syscalls_err!=FD_VM_SUCCESS ) ) FD_LOG_CRIT(( "fd_vm_syscall_register_slot failed" ));
+
+  /* Run ELF loader */
+
+  fd_sbpf_loader_config_t config = {0};
+  if( FD_UNLIKELY( 0!=fd_sbpf_program_load( prog, progdata, progdata_sz, syscalls, &config ) ) ) {\
```
mjain-jump:
```suggestion
  if( FD_UNLIKELY( 0!=fd_sbpf_program_load( prog, progdata, progdata_sz, syscalls, &config ) ) ) {
```

src/flamenco/progcache/fd_prog_load.c
ripatel-fd: This method doesn't give me the XID the record was found in. I don't want to change this function either because that means updating 40 call sites.

src/flamenco/progcache/fd_progcache_rec.c
```diff
@@ -0,0 +1,107 @@
...
+  /* Loader requires syscall table */
+
+  fd_sbpf_syscalls_t _syscalls[ FD_SBPF_SYSCALLS_SLOT_CNT ];
+  fd_sbpf_syscalls_t * syscalls = fd_sbpf_syscalls_join( fd_sbpf_syscalls_new( _syscalls ) );
+  int syscalls_err = fd_vm_syscall_register_slot( syscalls, load_slot, features, /* is_deploy */ 0 );
+  if( FD_UNLIKELY( syscalls_err!=FD_VM_SUCCESS ) ) FD_LOG_CRIT(( "fd_vm_syscall_register_slot failed" ));
+
+  /* Run ELF loader */
+
+  fd_sbpf_loader_config_t config = {0};
```
mjain-jump: are you sure its fine to not set the min and max sbpf versions here?

src/flamenco/progcache/fd_prog_load.c
ripatel-fd: Could do, but the internals use fd_funk_rec_key_t ... 🤔 Would prefer to keep it opaque for now

src/flamenco/progcache/fd_prog_load.c
ripatel-fd: @mjain-jump I don't think this is true. Can't you withdraw and redeploy an existing program without changing the programdata account at all?

src/flamenco/progcache/fd_progcache_rec.c
ripatel-fd: I see, I think I'm missing the fix that you recently PRed

src/flamenco/progcache/fd_progcache_user.h
```diff
@@ -0,0 +1,142 @@
...
+/* fd_progcache_pull does a cache fill (on cache miss) or returns an
+   existing cache entry (on cache hit).  Returns a pointer to the cache
+   entry. */
+
+fd_progcache_rec_t const *
+fd_progcache_pull( fd_progcache_t *           cache,
+                   fd_funk_t *                accdb,
+                   fd_funk_txn_xid_t const *  xid,
+                   void const *               prog_addr,
+                   fd_prog_load_env_t const * opt );
```
mjain-jump: nit: the `opt` argument name differs from the definition in `fd_progcache_user.c`

src/flamenco/progcache/fd_progcache_rec.h
```diff
@@ -0,0 +1,119 @@
...
+  uint rodata_sz;
+
+  uint calldests_off;  /* offset to sbpf_calldests map */
+  uint rodata_off;     /* offset to rodata segment */
+
+  /* SBPF version, SIMD-0161 */
+  uchar sbpf_version;
+
+  uint executable : 1;
+  uint invalidate : 1;  /* override this result at the next slot */
```
mjain-jump: i think these fields need to be better documented (when they're used, updated, etc.) its not super clear atm without reading the code

src/flamenco/progcache/fd_prog_load.c
mjain-jump: if the program is redeployed with the same `programdata_address` then the `slot` value in the `programdata` account metadata is updated to the current slot, requiring modification of the programdata account.

src/flamenco/progcache/fd_progcache_rec.c
ripatel-fd: good point

src/flamenco/progcache/fd_progcache_user.c
```diff
@@ -0,0 +1,684 @@
...
+fd_progcache_load_fork( fd_progcache_t *          cache,
+                        fd_funk_txn_xid_t const * xid,
+                        ulong                     epoch_slot0 ) {
+  /* Skip if already on the correct fork */
+  if( FD_LIKELY( (!!cache->fork_depth) & (!!fd_funk_txn_xid_eq( &cache->fork[ 0 ], xid ) ) ) ) return;
+  cache->metrics->fork_switch_cnt++;
+  fd_progcache_load_fork_slow( cache, xid, epoch_slot0 ); /* switch fork */
+}
+
+/* fd_progcache_query searches for a program cache entry on the current
```
mjain-jump, line 161: stray comment

src/flamenco/progcache/fd_progcache_user.h
```diff
@@ -0,0 +1,142 @@
...
+void *
+fd_progcache_leave( fd_progcache_t * cache,
+                    void **          opt_shfunk );
+
+/* Record-level operations ********************************************/
+
+/* fd_progcache_peek queries the program cache for an existing cache
+   entry.  Does not fill the cache.  Returns a pointer to the entry on
+   cache hit (invalidated by the next non-const API call).  Returns NULL
+   on cache miss. */
```
mjain-jump, line 145: not sure if this is documented somewhere else already, but should also include that this may also return NULL if the program's last verified slot was in a prior epoch

src/flamenco/progcache/fd_progcache_user.c
```diff
@@ -0,0 +1,684 @@
...
+                    fd_ulong_bswap( FD_LOAD( ulong, (uchar const *)prog_addr+16 ) ),
+                    fd_ulong_bswap( FD_LOAD( ulong, (uchar const *)prog_addr+24 ) ),
+                    found_rec->last_slot_verified,
+                    xid->ul[0] ));
+    }
+    goto miss;
+  }
+
+  /* Entry is too old */
+  if( found_rec->last_slot_verified < env->epoch_slot0 ) goto miss;
```
mjain-jump: peek already handles this, so either this or the check in peek should be removed

src/flamenco/progcache/fd_progcache_user.c
```diff
@@ -0,0 +1,684 @@
...
+
+  /* Create a tombstone */
+
+  void * rec_mem = fd_funk_val_truncate( funk_rec, funk->alloc, funk->wksp, fd_progcache_rec_align(), fd_progcache_rec_footprint( NULL ), NULL );
+  if( FD_UNLIKELY( !rec_mem ) ) {
+    FD_LOG_ERR(( "Program cache is out of memory: fd_alloc_malloc failed (requested align=%lu sz=%lu)",
+                  fd_progcache_rec_align(), fd_progcache_rec_footprint( NULL ) ));
+  }
+  fd_progcache_rec_t * rec = fd_progcache_rec_new_nx( rec_mem, slot );
+  rec->invalidate = 1;
```
mjain-jump, line 670: whats the purpose of having both `invalidate` and `executable`? can we use one or the other?

mjain-jump: if i understand correctly, its pretty much the equivalent of the `last_modified_slot` and `failed_verification` flags from the previous implementation?

src/flamenco/runtime/program/fd_bpf_loader_program.c
```diff
@@ -2577,29 +2550,14 @@ fd_bpf_loader_program_execute( fd_exec_instr_ctx_t * ctx ) {
...
-                                                  program_id,
-                                                  &cache_entry )!=0 ) ) {
+    fd_prog_load_env_t load_env[1]; fd_prog_load_env_from_bank( load_env, ctx->txn_ctx->bank );
+    fd_progcache_rec_t const * cache_entry =
+        fd_progcache_pull( ctx->txn_ctx->progcache,
+                           ctx->txn_ctx->funk,
+                           ctx->txn_ctx->xid,
+                           program_id,
+                           load_env );
+    if( FD_UNLIKELY( !cache_entry ) ) {
```
mjain-jump, line 2561: can you write a small comment here explaining why the cache fetch can fail

src/flamenco/progcache/fd_progcache_user.c
ripatel-fd: @mjain-jump To be honest, I didn't really understand what "last_modified_slot" does, since it was zero most of the time.

`invalidate` means: "This program was modified at the current fork graph node, you need to reload it in a future slot"

src/flamenco/progcache/fd_progcache_user.c
ripatel-fd: no it's supposed to be there

src/flamenco/progcache/fd_progcache_user.h
ripatel-fd: > if the program's last verified slot was in a prior epoch

that's also a cache miss

src/flamenco/progcache/fd_progcache_user.c
ripatel-fd: good find

src/flamenco/runtime/program/fd_bpf_loader_program.c
ripatel-fd: I added that comment to the documentation of the pull function

src/flamenco/progcache/fd_progcache_user.c
mjain-jump: there's a list called `programs_to_reverify` inside the txn ctx that gets amended every time `fd_deploy_program` is called (for upgrades / deployments). in txn finalize we iterate over this list and set the `last_slot_modified` field to the current slot for each of the cache entries. so i think yes it's pretty much equivalent to your `invalidate` flag. what's the purpose of `executable` in your PR?

src/flamenco/progcache/fd_progcache_user.c
mjain-jump: i see, but it looks like you used the wrong function name in your documentation (should be `fd_progcache_fork_has_xid`?)

src/flamenco/progcache/fd_prog_load.c
mjain-jump: for the program cache though it makes it clear what the expected arguments are

src/flamenco/progcache/fd_progcache_user.c
ripatel-fd: > what's the purpose of executable in your PR?

This is documented

ripatel-fd: > Entries are either non-executable (e.g. programs that failed
   verification) or executable.  Non-executable entry objects consist
   only of this header struct.  Executable entry objects are variable-
   sized and contain additional structures past this header (rodata/ROM
   segment, control flow metadata, ...).
> [...]
> uint executable : 1;  /* is this an executable entry? */

src/flamenco/progcache/fd_progcache_user.c
ripatel-fd: This comment refers to the query routine overall, but some parts of query (like fork_has_xid) is split out into separate functions.

src/flamenco/progcache/fd_progcache_admin.c
ripatel-fd: rec_max1 will do the actual estimate (need to tune this in the future), rec_max is a wrapper that adds some min bounds

src/flamenco/progcache/fd_prog_load.c
ripatel-fd: Ok good to know, thanks. I'll simplify this.

src/flamenco/progcache/fd_progcache_rec.c
ripatel-fd: Fixed

src/flamenco/progcache/fd_progcache_user.c
```diff
@@ -0,0 +1,696 @@
...
+    /* Skip over records from an older epoch (FIXME could bail early
+       here if the chain is ordered) */
+    ulong found_slot = rec->pair.xid->ul[0];
+    if( FD_UNLIKELY( found_slot<epoch_slot0 ) ) continue;
+
+    /* Skip over records that are older than what we already have */
+    if( FD_UNLIKELY( (long)found_slot<best_slot ) ) continue;
+
+    /* Confirm that record is part of the current fork
+       FIXME this has bad performance / pointer-chasing */
```
ripatel-fd, line 217: can remove this FIXME, i actually addressed this problem

src/disco/metrics/metrics.xml
```diff
@@ -1062,4 +1062,14 @@ metric introduced.
     <gauge name="StartSlot" summary="The slot at which the backtest started" />
 </tile>
 
+<tile name="exec">
+    <counter name="ProgcacheMisses" summary="Number of program cache misses" />
+    <counter name="ProgcacheHits" summary="Number of program cache hits" />
+    <counter name="ProgcacheFills" summary="Number of program cache insertions" />
```
mmcgee-jump, line 1068: split to enum?

ProgCacheInsertResult
  - Success
  - SuccessTombstone
  - Duplicate

ripatel-fd: Will do this and other suggestions discussed in chat in the next PR

PR #6660, opened 2025-10-08
funk: fix deep asan use after poison

two-heart:
Now block level test vectors run with deep asan without crashes


two-heart:
Failing benchmark since branches outside the FD repo cannot post comments on issues.

PR #6668, opened 2025-10-09
snapshot: adding unlinkat to seccomp

ibhatt-jumptrading:


src/app/firedancer/config/default.toml
```diff
@@ -1489,7 +1489,7 @@ user = ""
     #
     # In development, you can disable the sandbox for testing and
     # debugging with the `--no-sandbox` argument to `firedancer-dev`.
-    sandbox = false
+    sandbox = true
```
ripatel-fd, line 1492: did you mean to change this default?

ibhatt-jumptrading: yes

PR #6674, opened 2025-10-09
elf: fix uninitialized read

mjain-jump:
Closes https://github.com/firedancer-io/auditor-internal/issues/329

### DIFF
--- a/src/ballet/sbpf/fd_sbpf_loader.c
+++ b/src/ballet/sbpf/fd_sbpf_loader.c
@@ -220,19 +220,47 @@ struct fd_sbpf_loader {
 };
 typedef struct fd_sbpf_loader fd_sbpf_loader_t;
 
-/* Queries a single string from a section which is marked as SHT_STRTAB.
-   Returns an ElfParserError on failure. On success, returns 0 and
-   writes the queried string into out_str, which should be a
-   pre-allocated buffer of at least maximum_length bytes. The resulting
-   string is guaranteed to be null-terminated on success.
+/* fd_sbpf_slice_cstr_eq is a helper method for checking equality
+   between a slice of memory to a null-terminated C-string.  Unlike
+   strcmp, this function does not include the null-terminator in the
+   comparison.  Returns 1 if the first slice_len bytes of the slice and
+   cstr are equal, and 0 otherwise. */
+static inline int
+fd_sbpf_slice_cstr_eq( uchar const * slice,
+                       ulong         slice_len,
+                       char const *  cstr ) {
+  return !!(slice_len==strlen( cstr ) && fd_memeq( slice, cstr, slice_len ));
+}
+
+/* fd_sbpf_slice_cstr_start_with is a helper method for checking that a
+   null-terminated C-string is a prefix of a slice of memory.  Returns 1
+   if the first strlen(cstr) bytes of cstr is a prefix of slice, and 0
+   otherwise. */
+static inline int
+fd_sbpf_slice_cstr_start_with( uchar const * slice,
+                               ulong         slice_len,
+                               char const *  cstr ) {
+  ulong cstr_len = strlen( cstr );
+  return !!(slice_len>=cstr_len && fd_memeq( slice, cstr, cstr_len ));
+}
+
+/* fd_sbpf_lenient_get_string_in_section queries a single string from a
+   section which is marked as SHT_STRTAB.  Returns an ElfParserError on
+   failure, and leaves *out_slice and *out_slice_len in an undefined
+   state.  On success, returns 0 and sets *out_slice to a pointer into
+   elf_bytes corresponding to the beginning of the string within the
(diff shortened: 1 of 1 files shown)


ripatel-fd:
Well, looking closer into it, seems like the auditor-internal report is a false positive. 
Sure, the memeq call touches uninitialized bytes, but this never causes UB or wrong results. 

Please properly specify the behavior of `fd_sbpf_lenient_get_string_in_section`

Currently, this is underspecified:

```
On success, returns 0 and
   writes the queried string into out_str, which should be a
   pre-allocated buffer of at least maximum_length bytes. The resulting
   string is guaranteed to be null-terminated on success.
```

Could be rewritten as:

```
Writes a cstr into out_str on success. Leaves bytes beyond the null terminator uninitialized.
```

But if this is the case, why do we use `fd_memeq` down below instead of just using `strncmp`?

two-heart:
@ripatel-fd I don't follow this reasoning, `name` could contain anything, say it already contains b".symtab\0", now if the bytes starting `(uchar const *)elf_bytes+string_range_start` with length `string_range_end-string_range_start` is just the bytes `.sym` it would incorrectly match.

two-heart:
The original code also seems to not match agave, since it does not compare the 0 byte at the end.

ripatel-fd:
> The original code also seems to not match agave, since it does not compare the 0 byte at the end.

Good point

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -220,19 +220,44 @@ struct fd_sbpf_loader {
 };
 typedef struct fd_sbpf_loader fd_sbpf_loader_t;
 
+/* Helper method for checking equality between a slice of memory to a
+   null-terminated C-string. Unlike strcmp, this function does not
+   include the null-terminator in the comparison. Returns 1 if the first
+   slice_len bytes of the slice and cstr are equal, and 0 otherwise. */
+int
```
ripatel-fd:
```suggestion
static inline int
```

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -220,19 +220,44 @@ struct fd_sbpf_loader {
...
+fd_sbpf_slice_cstr_eq( uchar const * slice,
+                       ulong         slice_len,
+                       char const *  cstr ) {
+  return !!(slice_len==strlen( cstr ) && fd_memeq( slice, cstr, slice_len ));
+}
+
+/* Helper method for checking that a null-terminated C-string is a
+   prefix of a slice of memory. Returns 1 if the first strlen(cstr)
+   bytes of cstr is a prefix of slice, and 0 otherwise. */
+int
```
ripatel-fd:
```suggestion
static inline int
```

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -220,19 +220,44 @@ struct fd_sbpf_loader {
 };
 typedef struct fd_sbpf_loader fd_sbpf_loader_t;
 
+/* Helper method for checking equality between a slice of memory to a
+   null-terminated C-string. Unlike strcmp, this function does not
+   include the null-terminator in the comparison. Returns 1 if the first
+   slice_len bytes of the slice and cstr are equal, and 0 otherwise. */
```
ripatel-fd: sorry for being super pedantic, but per convention we limit doc comments to 75 cars with, two spaces after each sentence, and start docs with `<function_name> ...`

(feel free to ignore this)

src/ballet/sbpf/fd_sbpf_loader.c
ripatel-fd: I think it wouldn't hurt to use a overflow-safe addition operation here and also check that offset_in_file is in bounds. To defend against invalid input arguments. The function documentation doesn't specify any restrictions on the input arguments, so technically there's nothing preventing someone from passing in an out-of-bounds 'offset in section' parameter.

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -1644,7 +1674,7 @@ fd_sbpf_program_relocate( fd_sbpf_program_t *             prog,
       /* Update the calldests
          https://github.com/anza-xyz/sbpf/blob/v0.12.2/src/elf.rs#L1027-L1032 */
       uint pc_hash;
-      int err = fd_sbpf_register_function_hashed_legacy( loader, prog, "", (ulong)target_pc, &pc_hash );
+      int err = fd_sbpf_register_function_hashed_legacy( loader, prog, (uchar const *)"", 0UL, (ulong)target_pc, &pc_hash );
```
ripatel-fd:
```suggestion
      int err = fd_sbpf_register_function_hashed_legacy( loader, prog, NULL, 0UL, (ulong)target_pc, &pc_hash );
```

PR #6679, opened 2025-10-09
fuzz, replay: wire in block dumping

mjain-jump:


src/discof/replay/fd_replay_tile.c
```diff
@@ -1446,6 +1463,11 @@ dispatch_task( fd_replay_tile_t *  ctx,
       fd_funk_txn_xid_t xid = { .ul = { task->txn_exec->slot, task->txn_exec->slot } };
       fd_runtime_update_program_cache( bank, ctx->funk, &xid, txn_p, ctx->runtime_spad );
 
+      /* Add the transaction to the block dumper if necessary */
```
ibhatt-jumptrading: not sure if this works with forks. We might accidentally dump txns from a competing fork that we are executing

mjain-jump: should work fine because its for backtest only. i'll add a comment

kbhargava-jump: do we want this to work for when we have forks? @nipatel-jump has created `simulate`, which is basically backtest with forks allowed since we capture all the fecs, so if we see a mismatch live we can replay it again. would be pretty useful to capture blocks in simulate if possible

mjain-jump: there's some other fields to add before that happens (such as `parent_lt_hash` in order to match the effects, and then also harness behavior needs to be optimized for larger inputs too. could address those in follow-up PRs. The main purpose of this PR for now is to help collect seed corpuses for fuzzing.

src/flamenco/runtime/tests/fd_dump_pb.h
```diff
@@ -53,15 +53,86 @@
...
 #include "../info/fd_instr_info.h"
-#include "../info/fd_runtime_block_info.h"
 #include "../../vm/fd_vm.h"
 #include "generated/block.pb.h"
 #include "generated/elf.pb.h"
+#include "../../../disco/fd_txn_p.h"
+
+/* The amount of memory allocated towards dumping blocks from ledgers */
+#define FD_BLOCK_DUMP_CTX_SPAD_MEM_MAX (5UL<<30)
+#define FD_BLOCK_DUMP_CTX_MAX_TXN_CNT  (10000UL)
```
cmoyes-jump, line 63: These limits are okay?

src/flamenco/runtime/tests/fd_dump_pb.c
```diff
@@ -878,20 +1015,12 @@ fd_dump_instr_to_protobuf( fd_exec_txn_ctx_t * txn_ctx,
...
-      position = fd_cstr_append_cstr(position, txn_ctx->capture_ctx->dump_proto_output_dir);
-      position = fd_cstr_append_cstr(position, "/instr-");
-      position = fd_cstr_append_cstr(position, encoded_signature);
-      position = fd_cstr_append_cstr(position, "-");
-      position = fd_cstr_append_ushort_as_text(position, '0', 0, instruction_idx, 3); // Assume max 3 digits
-      position = fd_cstr_append_cstr(position, ".instrctx");
-      fd_cstr_fini(position);
-
+      char output_filepath[ PATH_MAX ];
+      sprintf( output_filepath, "%s/instr-%s-%hu.instrctx", txn_ctx->capture_ctx->dump_proto_output_dir, encoded_signature, instruction_idx );
```
cmoyes-jump: Change these to snprintf ?

src/flamenco/runtime/tests/fd_dump_pb.c
```diff
@@ -325,256 +328,440 @@ dump_blockhash_queue( fd_blockhashes_t const * queue,
...
+create_block_context_protobuf_from_block( fd_block_dump_ctx_t * dump_ctx,
+                                          fd_banks_t *          banks,
+                                          fd_bank_t *           bank,
+                                          fd_funk_t *           funk ) {
+  /* We should use the bank fields and funk txn from the parent slot in
+     order to capture the block context from before the current block
+     was executed, since dumping is happening in the block finalize
+     step. */
+  fd_bank_t *                    parent_bank    = fd_banks_get_parent( banks, bank );
+  fd_funk_txn_xid_t              parent_xid     = { .ul = { fd_bank_slot_get( parent_bank ), fd_bank_slot_get( parent_bank ) } };
```
cmoyes-jump: nit: a bit weird this is called twice

src/flamenco/runtime/tests/fd_dump_pb.h
mjain-jump: yep they are very generous, and can be bumped if needed

src/flamenco/runtime/tests/fd_dump_pb.c
mjain-jump: good idea, fixed

src/flamenco/runtime/tests/fd_dump_pb.c
mjain-jump: fixed

PR #6680, opened 2025-10-09
snapshots: add ancestors

cali-jumptrading:


src/discof/restore/utils/fd_ssmsg.h
```diff
@@ -368,6 +368,11 @@ struct fd_snapshot_manifest {
   /* The fork_id in the status cache for the root slot. */
   ushort txncache_fork_id;
 
+  /* A list of ancestor slots. The bound comes from
```
mmcgee-jump, line 371: I don't think this is the right bound? It's the max bit width of the rolling bit field but the excess could be arbitrary

cali-jumptrading: Do you have a particular bound in mind? We could try increasing the bound by another power of 2, but I find that to be a bit excessive... And it seems the excess is unbounded but shouldn't be populated unless there's a bug according to the comment here: https://github.com/anza-xyz/agave/blob/master/accounts-db/src/rolling_bit_field.rs#L21

mmcgee-jump: I think it's unbounded from Agave in general. I'd just explain why that is and how it could happen, and set a reasonable bound ourselves (4096 or 8192 probably fine).

cali-jumptrading: Done

PR #6687, opened 2025-10-09
runtime: stakes and epoch boundary cleanup

ibhatt-jumptrading:


src/flamenco/rewards/fd_epoch_rewards.c
```diff
@@ -50,10 +77,12 @@ fd_epoch_rewards_new( void * shmem, ulong stake_account_max ) {
...
 
-  memset( epoch_rewards, 0, sizeof(fd_epoch_rewards_t) );
 
-  epoch_rewards->magic              = FD_EPOCH_REWARDS_MAGIC;
-  epoch_rewards->stake_account_max_ = stake_account_max;
+  FD_COMPILER_MFENCE();
+  epoch_rewards->magic = FD_EPOCH_REWARDS_MAGIC;
+  FD_COMPILER_MFENCE();
+
+  epoch_rewards->stake_account_max = stake_account_max;
```
yufeng-jump: Probably doesn't make a difference to correctness, but I think the general pattern is that fence and magic setting are the last thing that happens, and other fields get set before magic.

yufeng-jump (approved):

kbhargava-jump (approved):

PR #6692, opened 2025-10-10
firedancer-dev: read gossvf config parameters

davidkleymann:
Currently, firedancer-dev hardcodes `shred_version` and `allow_private_address` both to 0. This prevents use on local deployments without a source of shreds.

This PR sets both parameters to the values from the config.


mmcgee-jump:
This is just for the development `gossip` command, not the production validator

davidkleymann:
@mmcgee-jump I cleaned up the commits, is it ready to merge?

mmcgee-jump (approved):

PR #6703, opened 2025-10-10
snapshots: split loading into snapld tile and general belt sanding

amass-jump:


src/discof/restore/utils/fd_sshttp.c
```diff
@@ -129,11 +129,6 @@ fd_sshttp_init( fd_sshttp_t * http,
   http->sockfd = socket( AF_INET, SOCK_STREAM|SOCK_NONBLOCK, 0 );
   if( FD_UNLIKELY( -1==http->sockfd ) ) FD_LOG_ERR(( "socket() failed (%d-%s)", errno, fd_io_strerror( errno ) ));
 
-  int optval = 1;
```
mmcgee-jump, line 132: Why this went away?

amass-jump: It caused a failure in the sandbox (due to setsockopt syscall), and rather than fixing that I decided to simplify & remove it because I'm pretty sure TCP_NODELAY can have no possible effect on RX speed. That was true after I tested again after removing in any event.

PR #6705, opened 2025-10-11
sched: clarify task scheduling policy and fix minor bug

yufeng-jump:
Should be ever so slightly faster due to a full round of dispatching at the start of a block, and scheduling should look more front loaded overall.

### DIFF
--- a/src/discof/replay/fd_replay_tile.c
+++ b/src/discof/replay/fd_replay_tile.c
@@ -123,8 +123,6 @@ fd_block_id_ele_get_idx( fd_block_id_ele_t * ele_arr, fd_block_id_ele_t * ele ) 
   return (ulong)(ele - ele_arr);
 }
 
-FD_STATIC_ASSERT( FD_PACK_MAX_BANK_TILES<=64UL, exec_bitset );
-
 struct fd_replay_tile {
   fd_wksp_t * wksp;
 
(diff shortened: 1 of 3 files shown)


src/discof/replay/fd_replay_tile.c
```diff
@@ -1769,6 +1754,7 @@ after_credit( fd_replay_tile_t *  ctx,
   }
 
   *charge_busy = replay( ctx, stem );
+  *opt_poll_in = !*charge_busy;
```
yufeng-jump, line 1757: This ensures a full round of dispatching at the start of a block.

src/discof/replay/fd_sched.c
```diff
@@ -653,12 +675,22 @@ fd_sched_task_next_ready( fd_sched_t * sched, fd_sched_task_t * out, ulong exec_
...
+
+      /* Dispatch more sigverify tasks only if at least one exec tile is
+         executing transactions or completely idle.  Allow at most one
+         sigverify task in-flight per tile, and only dispatch to
+         completely idle tiles. */
+      ulong exec_fully_ready_bitset = exec_ready_bitset0;
+      ulong sigverify_ready_bitset = sched->sigverify_ready_bitset[ 0 ] & exec_fully_ready_bitset;
       ulong sigverify_queued_cnt = block->txn_parsed_cnt-block->txn_sigverify_in_flight_cnt-block->txn_sigverify_done_cnt;
-      if( FD_LIKELY( sigverify_queued_cnt>0UL && exec_tile_cnt>1UL ) ) { /* Leave one exec tile idle for critical path execution. */
+      if( FD_LIKELY( sigverify_queued_cnt>0UL && fd_ulong_popcnt( sigverify_ready_bitset )>fd_int_if( block->txn_exec_in_flight_cnt>0U, 0, 1 ) ) ) {
```
yufeng-jump, line 686: This is the bug that caused scheduling to be less front loaded than it could be.  Up to one exec tile was not getting sigverify tasks when it could be.

PR #6714, opened 2025-10-13
docs: fixed broken paths 

bobtajson:
Fixed broken paths to files in docs


ripatel-fd:
These links are not broken. The book is rendered to a website here and the links make sense under the context: https://docs.firedancer.io/guide/initializing.html
The book is not meant to be viewed as GitHub Markdown.

PR #6715, opened 2025-10-13
bmtree: short-circuit to avoid UB

intrigus-lgtm:
Use `||` so `proof_depth` is guaranteed to not be out of range in the shift.
With just `|` this would have been possible as it doesn't short-circuit


intrigus-lgtm:
> I looked at fd_ulong_shift_left too, but it generates an extra instruction on both GCC and Clang. This patch has no performance difference on Clang, but it creates a branch on GCC. Suppose it doesn't really matter for performance, but I don't like the idea of making code worse to suppress linter false positives.

Could you elaborate a bit on why you consider this a linter false positive?
As far as my understanding of the C standard goes, a compiler could do the shift before the check.
Because it did the shift, the check is now redundant, because it would have been UB to do it.
Therefore the check could be removed as far as I understand.

Now in practice I couldn't get it to actually generate wrong code, but I don't see why a compiler shouldn't do that in the future.

Anything I'm missing?

ripatel-fd:
> because it would have been UB to do it

The UB is a good point. I thought that only the result of the expression is undefined, but it's true UB

PR #6731, opened 2025-10-13
Fix: fini not considering the queue mode

sense112:
## Description:

When upgrading from version 709 to 711 with queue mode set to `simple` , the following error occurs during the configure command:

```
fd1:[group]:main src/app/shared/commands/configure/configure.c(89)[configure_stage]: ethtool-channels ... undoing ... device `bond0` has partial ethtool-channels network configuration
fd1:[group]:main src/app/shared/commands/configure/fd_ethtool_ioctl.c(139)[fd_ethtool_ioctl_rxfh_set_default]: RUN: `ethtool --set-rxfh-indir enp1s0f0 default`
fd1:[group]:main src/app/shared/commands/configure/fd_ethtool_ioctl.c(92)[fd_ethtool_ioctl_channels_set_num]: RUN: `ethtool --set-channels enp1s0f0 combined 48`
fd1:[group]:main src/app/shared/commands/configure/fd_ethtool_ioctl.c(139)[fd_ethtool_ioctl_rxfh_set_default]: RUN: `ethtool --set-rxfh-indir enp1s0f0 default`
fd1:[group]:main src/app/shared/commands/configure/fd_ethtool_ioctl.c(139)[fd_ethtool_ioctl_rxfh_set_default]: RUN: `ethtool --set-rxfh-indir enp1s0f1 default`
fd1:[group]:main src/app/shared/commands/configure/fd_ethtool_ioctl.c(92)[fd_ethtool_ioctl_channels_set_num]: RUN: `ethtool --set-channels enp1s0f1 combined 48`
fd1:[group]:main src/app/shared/commands/configure/fd_ethtool_ioctl.c(139)[fd_ethtool_ioctl_rxfh_set_default]: RUN: `ethtool --set-rxfh-indir enp1s0f1 default`
fd1:[group]:main src/app/shared/commands/configure/configure.c(97)[configure_stage]: ethtool-channels ... clean was unable to get back to an unconfigured state ... device `bond0` has partial ethtool-channels network configuration
```

## Root Cause:

The `fdctl configure init all` process currently executes the following sequence:

1. Runs the `Check` command to verify `ethtool` configuration (considering the configured queue mode).
2. If the check fails, it runs the `fini` process, which currently **ignores** the configured queue mode and forces max combined.
3. The `Check` command re-runs which then expects `simple`  mode (1 combined) but sees the `fini`-enforced (max combined)  configuration, causing it to fail with error message `clean was unable to get back to an unconfigured state`.
4. This failure prevents reaching the init step, so the node never gets properly configured.

## Solution:

This PR updates the `fini` process to respect the configured queue mode. This ensures the second check passes, allowing the init process to execute and properly configure the node.

### DIFF
--- a/src/app/shared/commands/configure/ethtool-channels.c
+++ b/src/app/shared/commands/configure/ethtool-channels.c
@@ -321,7 +321,7 @@ check( fd_config_t const * config ) {
 }
 
 static int
-fini_device( char const * device ) {
+fini_device( char const * device, uint num_channels ) {
   int error = 0;
 
   fd_ethtool_ioctl_t ioc;
@@ -347,7 +347,7 @@ fini_device( char const * device ) {
   /* This should happen first, otherwise changing the number of channels may fail */
   error |= (0!=fd_ethtool_ioctl_rxfh_set_default( &ioc ));
 
-  error |= (0!=fd_ethtool_ioctl_channels_set_num( &ioc, 0 /* max */ ));
+  error |= (0!=fd_ethtool_ioctl_channels_set_num( &ioc, num_channels ));
 
   /* Some drivers (i40e) do not always evenly redistribute the RXFH table
      when increasing the channel count, so we run this again just in case. */
@@ -382,15 +382,18 @@ static int
 fini( fd_config_t const * config,
       int                 pre_init FD_PARAM_UNUSED ) {
   int done = 0;
+  int dedicated_mode = (0==strcmp( config->net.xdp.rss_queue_mode, "dedicated" ));
+  uint const num_channels = !dedicated_mode ? config->layout.net_tile_count : 0;
+
   if( FD_UNLIKELY( device_is_bonded( config->net.interface ) ) ) {
     char line[ 4096 ];
     char const * bond_devices[ 16 ];
     device_read_slaves( config->net.interface, line, bond_devices );
     for( ulong i=0UL; bond_devices[ i ]!=NULL; i++ ) {
-      done |= fini_device( bond_devices[ i ] );
+      done |= fini_device( bond_devices[ i ], num_channels );
     }
   } else {
-    done = fini_device( config->net.interface );
+    done = fini_device( config->net.interface, num_channels );
   }
   return done;
 }


amass-jump:
Hi,

The intention of the `fini` step is to completely undo/remove any configuration done by `init`. So the logic flow here for init should roughly be:

- Run check(). If `check_device_is_configured()` is true for all the bond devices, then check() succeeds and init is done (already configured). In your example, it did not succeed.
- check() will now either return a partially-configured result or a not-configured one, based on the return of `check_device_is_modified()`. In your example, the devices were detected as modified
- Because the state is modified, fini() will run to reset us to a default/unconfigured state so that we can later run the init() logic again. We run fini() and then run check() again, and this time we expect that what was partially configured before is now fully not configured
- In your example, that second check() returned partially configured again (after fini()), which is an error.

So the real issue here is why fini() failed with your setup. Happy to help look into that. We need to figure out why the second round of check_device_is_modified() calls returned true.

The given changes are not correct as fini() should not depend on any firedancer configuration, as the intention is to undo anything firedancer has done.

HGuillemet:
I have similar error on a machine with X540 and no bond.
Feel free to reach me on discord if you'd like to experiment with my configuration.

sense112:
Hi @amass-jump,

Thank you for the reply. I agree and understand the intention of the fini. Looking through the code again, the second check seems to be running the exact same code path execution as the first check, what causes the second check to check for not configured vs the first check to check if it is configured? 

both the first and second check run `result = stage->check( config );` - I don't see a flag in here to check for configured vs checking for reset back to default.

sense112:
Ah nevermind, I think I understand the code path a bit better now, will test again on the node and revert back afterwards

HGuillemet:
@amass-jump In my case at least (X540), the problem is that after `ethtool --set-channels combined 63` and `ethtool  --set-rxfh-indir default`, the RXFD tables looks like this:
```
RX flow hash indirection table for enp193s0f0 with 16 RX ring(s):
    0:      0     1     2     3     4     5     6     7
    8:      8     9    10    11    12    13    14    15
   16:      0     1     2     3     4     5     6     7
   24:      8     9    10    11    12    13    14    15
   32:      0     1     2     3     4     5     6     7
   40:      8     9    10    11    12    13    14    15
   48:      0     1     2     3     4     5     6     7
   56:      8     9    10    11    12    13    14    15
   64:      0     1     2     3     4     5     6     7
   72:      8     9    10    11    12    13    14    15
   80:      0     1     2     3     4     5     6     7
   88:      8     9    10    11    12    13    14    15
   96:      0     1     2     3     4     5     6     7
  104:      8     9    10    11    12    13    14    15
  112:      0     1     2     3     4     5     6     7
  120:      8     9    10    11    12    13    14    15
  ```
  while `check_device_is_modified` expects to see the table dispatching into 63 ring buffers.
  Hence the `device has partial ethtool-channels network configuration` error.
  
  Isn't there a confusion between number of channels and number of rings ?

**Comment by amass-jump** [comment]
@HGuillemet,

> Isn't there a confusion between number of channels and number of rings ?

Yes, looks like it. Could you please give this a try and let me know if that fixes the issues you are seeing? https://github.com/firedancer-io/firedancer/pull/6793

**Comment by HGuillemet** [comment]
@amass-jump
Yes it does.

"dedicated" doesn't work with message "failed to isolate queue zero", but that's maybe expected with X540 ?

**Comment by amass-jump** [comment]
@HGuillemet awesome, thanks for verifying. I will take a look at getting `dedicated` working with X540, but for now we wanted to fix issues with the default `simple`.

@sense112 Can you try the same?

**Comment by sense112** [comment]
@amass-jump  So I am on the same NIC as HGuillemet`X540-AT2`.

My table looks very similar `sudo ethtool -x enp1s0f0`:

```
RX flow hash indirection table for enp1s0f0 with 48 RX ring(s):
    0:      0     1     2     3     4     5     6     7
    8:      8     9    10    11    12    13    14    15
   16:      0     1     2     3     4     5     6     7
   24:      8     9    10    11    12    13    14    15
   32:      0     1     2     3     4     5     6     7
   40:      8     9    10    11    12    13    14    15
   48:      0     1     2     3     4     5     6     7
   56:      8     9    10    11    12    13    14    15
   64:      0     1     2     3     4     5     6     7
   72:      8     9    10    11    12    13    14    15
   80:      0     1     2     3     4     5     6     7
   88:      8     9    10    11    12    13    14    15
   96:      0     1     2     3     4     5     6     7
  104:      8     9    10    11    12    13    14    15
  112:      0     1     2     3     4     5     6     7
  120:      8     9    10    11    12    13    14    15
RSS hash key:
03:06:33:32:05:7b:d1:41:f9:23:03:58:88:c0:30:a1:e0:f4:89:e6:54:89:46:20:4a:42:29:a2:22:88:d3:90:56:b2:c2:0f:b8:c7:2b:c5
RSS hash function:
    toeplitz: on
    xor: off
    crc32: off
```

Your branch is working on my side, I also pulled some additional logs from that code snipped, it expects 16 but reset to 0

```
Index: 0 tableValue: 0, q: 0
Index: 1 tableValue: 1, q: 1
Index: 2 tableValue: 2, q: 2
Index: 3 tableValue: 3, q: 3
Index: 4 tableValue: 4, q: 4
Index: 5 tableValue: 5, q: 5
Index: 6 tableValue: 6, q: 6
Index: 7 tableValue: 7, q: 7
Index: 8 tableValue: 8, q: 8
Index: 9 tableValue: 9, q: 9
Index: 10 tableValue: 10, q: 10
Index: 11 tableValue: 11, q: 11
Index: 12 tableValue: 12, q: 12
Index: 13 tableValue: 13, q: 13
Index: 14 tableValue: 14, q: 14
Index: 15 tableValue: 15, q: 15
Index: 16 tableValue: 0, q: 16 <- Fails at this point
```

As for his dedicated test (just for the sake of completeness) I also got the error `error configuring network device (enp1s0f0), failed to isolate queue zero. Try `net.xdp.rss_queue_mode="simple"``

amass-jump:
Thanks to both of you for the assistance in diagnosing this issue. 

It turns out this is because `ixgbe` limits the RXFH indirection table to a maximum of 16 (or 64) queues, regardless of the number of "channels" the device has:

https://github.com/torvalds/linux/blob/master/drivers/net/ethernet/intel/ixgbe/ixgbe_ethtool.c#L2800

I will submit a patch for this (both to fix the check_device_is_modified() logic, but also such that rss_queue_mode=dedicated works with this driver).

Closing this for now

amass-jump:
https://github.com/firedancer-io/firedancer/pull/6806

sense112:
@amass-jump, amazing, thank you for the support and for the fix!

HGuillemet:
> It turns out this is because `ixgbe` limits the RXFH indirection table to a maximum of 16 (or 64) queues, regardless of the number of "channels" the device has:
> 
> https://github.com/torvalds/linux/blob/master/drivers/net/ethernet/intel/ixgbe/ixgbe_ethtool.c#L2800

Any idea why the driver limits to 16 ?
Maybe this should be reported ?

amass-jump:
> Any idea why the driver limits to 16 ?
Maybe this should be reported ?

This commit has some more info: https://github.com/torvalds/linux/commit/e85d3d55875f7a1079edfbc4e4e98d6f8aea9ac7

PR #6735, opened 2025-10-14
replay: runtime support for block equivocation

ibhatt-jumptrading:
Adding replay/runtime support for equivocating on a block. However, this does not add support for generic eviction.

### DIFF
--- a/src/discof/replay/fd_replay_tile.c
+++ b/src/discof/replay/fd_replay_tile.c
@@ -667,8 +667,8 @@ replay_block_start( fd_replay_tile_t *  ctx,
 
   /* Create a new funk txn for the block. */
 
-  fd_funk_txn_xid_t xid        = { .ul = { slot, slot } };
-  fd_funk_txn_xid_t parent_xid = { .ul = { parent_slot, parent_slot } };
+  fd_funk_txn_xid_t xid        = { .ul = { slot, bank_idx } };
+  fd_funk_txn_xid_t parent_xid = { .ul = { parent_slot, parent_bank_idx } };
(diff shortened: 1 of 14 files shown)


src/discof/replay/fd_replay_tile.c
```diff
@@ -1778,10 +1775,47 @@ after_credit( fd_replay_tile_t *  ctx,
      and pass it to the scheduler. */
 
   fd_reasm_fec_t * fec;
-  if( FD_LIKELY( fd_sched_can_ingest( ctx->sched ) && !fd_banks_is_full( ctx->banks ) && (fec = fd_reasm_out( ctx->reasm )) ) ) {
-    /* If sched is full or there are no free banks, we cannot ingest any
-       more FEC sets into the scheduler. */
-    process_fec_set( ctx, fec );
+  if( FD_LIKELY( fd_sched_can_ingest( ctx->sched, 1UL ) && !fd_banks_is_full( ctx->banks ) && (fec = fd_reasm_peek( ctx->reasm )) ) ) {
```
lidatong, line 1782: dont think peek is quite the right way to do this. what if there's another non equivocating FEC on the longest fork that immediately follows that you can replay?

this would block until the scheduler has enough FEC sets, so a leader who just repeatedly spam equivocates the last FEC set will effectively DOS us

think you need to queue up the equivocations and only process them when they're idle

lidatong (approved):

src/discof/replay/fd_replay_tile.c
ibhatt-jumptrading: discussed offline, will be addressed with a larger change in reasm

PR #6746, opened 2025-10-14
Fix error when enabling already-enabled ntuple

HGuillemet:
On ConnectX-4, and probably other NICS, ntuple feature is activated by default and cannot be changed.
This causes `fdctl init` to fail to configure the net tile with ` rss_queue_mode = "dedicated"`.

This PR proposes 2 fixes.
* A minor one that changes the error message when the ioctl returns a positive value. Current error message printed a misleading message using an `errno` that were not set by `ioctl`.
* A test to only try to enable ntuple if it's not already enabled. Maybe a more generic fix would be make such test in `fd_ethtool_ioctl_feature_set` instead of specifically for ntuple.

### DIFF
--- a/src/app/shared/commands/configure/ethtool-channels.c
+++ b/src/app/shared/commands/configure/ethtool-channels.c
@@ -117,9 +117,17 @@ init_device( char const *        device,
       else         return 1;
     }
 
-    if( FD_UNLIKELY( 0!=fd_ethtool_ioctl_feature_set( &ioc, FD_ETHTOOL_FEATURE_NTUPLE, 1 ) ) ) {
-      if( strict ) FD_LOG_ERR(( "error configuring network device (%s), failed to enable ntuple feature. Try `net.xdp.rss_queue_mode=\"simple\"`", device ));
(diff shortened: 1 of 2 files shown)


amass-jump:
Hi @HGuillemet,

> A test to only try to enable ntuple if it's not already enabled. Maybe a more generic fix would be make such test in fd_ethtool_ioctl_feature_set instead of specifically for ntuple.

This should already be the case for all calls to fd_ethtool_ioctl_feature_set, see https://github.com/firedancer-io/firedancer/blob/main/src/app/shared/commands/configure/fd_ethtool_ioctl.c#L284

Are you able to provide logging and/or debug where this call is failing?

HGuillemet:
@amass-jump I wrote this patch against v0.711. And indeed you already fix it some days ago on main branch.  So I guess you can close this PR.

However, being able to test `rss_queue_mode = "dedicated"`  with 0.711 and this patch, I tried it on mainnet and voting performance are way worse than with `simple`. But this would be best discussed elsewhere.

PR #6753, opened 2025-10-14
gossip: refactor peer sampling and fix invalid active set state

### LINKED ISSUE #7345: gossip: refactor sampler management
Right now samplers are not aware of contact info CRDS movements. For example, a contact info may be dropped from the table and still remain in the active push set (when it should be dropped) and peer samplers. This results in the `Sample peer not found in CRDS. Likely dropped` log. 

An event-driven "Contact info CRDS update" callback in fd_gossip can be used to manage the different samplers.

ravyu-jump:
- Convert wpeer sampler into a standalone library: used by CRDS peer sampler and Active Set bucket samplers
- Add callback `fd_crds_ci_change_fn` that tracks individual contact info events (insert/drop, stake change, identity change)
- Refactor active set: Active set is fully contained, maintaining samplers, push state, and state machines (with APIs to update internal state). 
- Changes collectively fix invalid active set states where:
  - a dropped peer might still be in a bucket
  - if node's identity change is triggered, the new identity might still be in a sample set or bucket


mmcgee-jump (changes requested):
handle identity change

PR #6759, opened 2025-10-15
ready: properly check status of tiles that can shutdown

anwayde:
The genesis tile shuts down after creating the genesis, so a `test_fddev`-like smoke test for the full client needs this.


src/app/shared/commands/ready.c
```diff
@@ -19,6 +19,9 @@ ready_cmd_fn( args_t *   args FD_PARAM_UNUSED,
        anyway. */
     if( FD_UNLIKELY( tile->is_agave ) ) continue;
 
+    /* Don't wait for tiles that are allowed to shutdown. */
```
mmcgee-jump: Not quite the right fix. You should wait for it, but wait for it to either become status==1UL or status==2UL

PR #6771, opened 2025-10-15
gossip: add missing origin rejects to purge table

### LINKED ISSUE #7445: Gossip: Separate purge table for CRDS values originating from nodes with no contact info
Due to the issue described in https://github.com/firedancer-io/firedancer/issues/7393, we end up re-requesting these values again in pull requests since we generate filters that do not contain these values. An easy work around is to maintain a separate no-contact-info table of crds hashes, very similar to the purge table the CRDS db maintains.

ravyu-jump:
Reduces pull response drops due to no origin contact info to near 0


ravyu-jump:
Converting to draft, metrics show minimal improvements in pull response duplicate rates.

PR #6779, opened 2025-10-16
startup: improve gossip and snapshot startup speed

mmcgee-jump:
- Switched HTTP downloader to blocking sockets for now, for a ~50% perf win. Couldn't quite figure out the issue with nonblocking, it's some coalescing thing where each recvfrom returns too little data and the syscall overhead is higher
- Added some info on gossip tile backpressure to watch command
- Fixed gossip tile backpressure caused by too infrequent credit updates from consumers
- Re-enabled fast gossip pull request loop for extra fast gossip table boot

### DIFF
--- a/src/discof/restore/utils/fd_sshttp.c
+++ b/src/discof/restore/utils/fd_sshttp.c
@@ -125,9 +125,18 @@ fd_sshttp_init( fd_sshttp_t * http,
     "Host: " FD_IP4_ADDR_FMT "\r\n\r\n",
     (int)path_len, path, FD_IP4_ADDR_FMT_ARGS( addr.addr ) ) );
 
+  /* TODO: Figure out recv coalescing properly and switch stream back
+     to non-blocking. */
   http->addr = addr;
-  http->sockfd = socket( AF_INET, SOCK_STREAM|SOCK_NONBLOCK, 0 );
+  http->sockfd = socket( AF_INET, SOCK_STREAM, 0 );
   if( FD_UNLIKELY( -1==http->sockfd ) ) FD_LOG_ERR(( "socket() failed (%d-%s)", errno, fd_io_strerror( errno ) ));
+
+  struct timeval timeout = {
+    .tv_sec = 0UL,
+    .tv_usec = (ulong)10e3,
+  };
+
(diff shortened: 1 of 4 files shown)


src/discof/restore/utils/fd_sshttp.c
```diff
@@ -125,10 +125,19 @@ fd_sshttp_init( fd_sshttp_t * http,
...
-  http->sockfd = socket( AF_INET, SOCK_STREAM|SOCK_NONBLOCK, 0 );
+  http->sockfd = socket( AF_INET, SOCK_STREAM, 0 );
   if( FD_UNLIKELY( -1==http->sockfd ) ) FD_LOG_ERR(( "socket() failed (%d-%s)", errno, fd_io_strerror( errno ) ));
 
+  struct timeval timeout = {
+    .tv_sec = 0UL,
+    .tv_usec = (ulong)10e3,
+  };
+
+  if( FD_UNLIKELY( -1==setsockopt( http->sockfd, SOL_SOCKET, SO_RCVTIMEO, &timeout, sizeof(timeout) ) ) ) FD_LOG_ERR(("setsockopt SO_RCVTIMEO failed (%d-%s)", errno, fd_io_strerror(errno)));
```
amass-jump, line 139: I believe this will break the current snapld seccomp policy

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -125,10 +125,19 @@ fd_sshttp_init( fd_sshttp_t * http,
...
+  /* TODO: Figure out recv coalescing properly and switch stream back
+     to non-blocking. */
   http->addr = addr;
-  http->sockfd = socket( AF_INET, SOCK_STREAM|SOCK_NONBLOCK, 0 );
+  http->sockfd = socket( AF_INET, SOCK_STREAM, 0 );
   if( FD_UNLIKELY( -1==http->sockfd ) ) FD_LOG_ERR(( "socket() failed (%d-%s)", errno, fd_io_strerror( errno ) ));
 
+  struct timeval timeout = {
+    .tv_sec = 0UL,
+    .tv_usec = (ulong)10e3,
```
amass-jump, line 136: just confirming you intend to block for 10ms? How'd you select that OOC?

amass-jump (approved):

src/discof/restore/utils/fd_sshttp.c
mmcgee-jump: It's enough to keep the tile heartbeat reasonably current, and react to control messages, assuming the download stalled, without degrading performance

mmcgee-jump: (Hopefully temporary anyway, until we can revert to NONBLOCK with the right coalescing)

amass-jump (approved):

PR #6781, opened 2025-10-16
fix: invalid SVC type assumption

losman0s:
### Problem

There is currently an invalid equivalence between "already scheduled" and "current type is dynamic" in the SVC timer scheduling function. This results in attempting to read the expiry of an instant timer, which segfaults.

### Solution

Tighten the check to ensure the current timer is dynamic.
For clarity, also changing `in_sched` -> `old_svc_type == FD_QUIC_SVC_DYNAMIC` in the subsequent removal check.


akhinvasara-jumptrading:
Incorporated into https://github.com/firedancer-io/firedancer/pull/6784 - thank you very much!

PR #6786, opened 2025-10-16
Drop backtest stderr logs to NOTICE

ripatel-fd:
Reduce stderr log spam (INFO logs still available via /tmp)


src/flamenco/runtime/tests/run_ledger_backtest.sh
```diff
@@ -202,7 +202,6 @@ echo "
     max_live_slots = 32
     max_fork_width = 4
 [log]
-    level_stderr = \"INFO\"
```
mmcgee-jump, line 205: Can you make it so `if env["CI"]` it keeps INFO ?

mmcgee-jump (changes requested):
.

PR #6800, opened 2025-10-16
quic: improve keep-alive impact on client svc schedule

akhinvasara-jumptrading:
1. After servicing a client conn to send the keep-alive ping, we continue to default schedule for last_activity+idle_timeout/2. But that's in the past! So we end up in an INSTANT scheduling loop until the peer responds to reset last_activity.... bad. This PR fixes that. 

2. Suppose at time _t_, the user calls fd_quic_conn_let_die() to _prolong_ the lifetime, providing a new keep_alive_duration longer than the currently implicit one. The connection is still alive, but _t_ is _after_ when we'd have sent the keep-alive (last_activity+idle_timeout/2). The next service for this conn is scheduled for last_activity+idle_timeout (after change 1 in this PR), so we'd just timeout the conn isntead of ever sending the keep-alive. This PR changes let_die() to check if we're in this particular situation, and if so, schedules the conn immediately in an attempt to save the connection.

3. minor: This PR also expands the conn_close frame debug log.

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -365,6 +365,8 @@ fd_quic_init( fd_quic_t * quic ) {
 
   switch( config->role ) {
   case FD_QUIC_ROLE_SERVER:
+    if( FD_UNLIKELY( config->keep_alive ) ) { FD_LOG_WARNING(( "server keep-alive not supported" )); return NULL; }
+    break;
   case FD_QUIC_ROLE_CLIENT:
     break;
   default:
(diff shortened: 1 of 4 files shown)


src/waltz/quic/fd_quic.c
```diff
@@ -2909,12 +2909,20 @@ fd_quic_svc_poll( fd_quic_t *      quic,
     fd_quic_cb_conn_final( quic, conn ); /* inform user before freeing */
     fd_quic_conn_free( quic, conn );
     break;
-  default:
-    /* prep idle timeout or keep alive at idle timeout/2 */
-    fd_quic_svc_prep_schedule( conn, conn->last_activity + (conn->idle_timeout_ns>>(quic->config.keep_alive)) );
+  default: {
```
akhinvasara-jumptrading, line 2914: @ripatel-fd This is the only important part

ripatel-fd (dismissed):

PR #6803, opened 2025-10-17
ci: set hugepages for codeql workflow

two-heart:


ripatel-fd:
@two-heart I don't think this will do anything, since the hugepages stage will never decrease reservations, only increase (to avoid memory fragmentation). What are you trying to do here? It looks like you want to free up more memory for CodeQL. In that case, I suggest we set up a dedicated CodeQL runner without any huge pages, that won't get any other CI jobs.

ripatel-fd (changes requested):
.

ripatel-fd:
Will tackle this with @kbhargava-jump

PR #6805, opened 2025-10-17
quic: premature 1-RTT packets

losman0s:
### Problem

`fd_quic` currently allows 1-RTT "flag" packets (`PING`, `MAX_DATA`, etc) to be sent even when handshake is ongoing.
Those packets are then discarded server-side (at least in `quinn-rs`) due to decryption failure.
From my quick testing, the current scheduling of e.g. PINGs vs handshake completion does not lead to many (any?) such occurrences in normal operation, but I did observe some when artificially delaying the handshake replies.

Opening this PR for consideration from eyes more familiar with `fd_quic`. 

### Solution

Require the connection to be in ACTIVE state before allowing `fd_quic_enc_level_appdata_id` due to pending flag packets.

This PR also gates PINGs consistently between the [first](https://github.com/firedancer-io/firedancer/blob/1d552bf8a1744ee32e5dc612f3412754391217df/src/waltz/quic/fd_quic.c#L2887) and [second](https://github.com/firedancer-io/firedancer/blob/1d552bf8a1744ee32e5dc612f3412754391217df/src/waltz/quic/fd_quic.c#L3866-L3867) scheduling sites.

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -888,7 +888,11 @@ fd_quic_tx_enc_level( fd_quic_conn_t * conn, int acks ) {
     return fd_quic_enc_level_appdata_id;
   }
 
-  if( conn->flags && conn->upd_pkt_number >= app_pkt_number ) {
+  /* only allow 1-RTT "flag" frames when we have the keys, to prevent e.g. early 1-RTT PINGs */
+  uint flags_pending = conn->flags & ~(FD_QUIC_CONN_FLAGS_CLOSE_SENT | FD_QUIC_CONN_FLAGS_PING_SENT);
+  if( ( flags_pending != 0U )
+      & ( conn->upd_pkt_number >= app_pkt_number )
+      & fd_uint_extract_bit( conn->keys_avail, fd_quic_enc_level_appdata_id ) ) {
     return fd_quic_enc_level_appdata_id;
   }
 
@@ -2891,7 +2895,7 @@ fd_quic_svc_poll( fd_quic_t *      quic,
       }
     } else if( quic->config.keep_alive & !!(conn->let_die_time_ns > now) ) {
       /* send PING */
-      if( !( conn->flags & FD_QUIC_CONN_FLAGS_PING ) ) {
+      if( !( conn->flags & ( FD_QUIC_CONN_FLAGS_PING | FD_QUIC_CONN_FLAGS_PING_SENT ) ) ) {
         conn->flags         |= FD_QUIC_CONN_FLAGS_PING;
         conn->upd_pkt_number = FD_QUIC_PKT_NUM_PENDING;     /* update to be sent in next packet */
       }
@@ -3876,8 +3880,7 @@ fd_quic_conn_service( fd_quic_t * quic, fd_quic_conn_t * conn, long now ) {
   /* Send new rtt measurement probe? */
   if( FD_UNLIKELY( now > conn->last_ack + (long)conn->rtt_period_ns ) ) {
     /* send PING */
-    if( !( conn->flags & ( FD_QUIC_CONN_FLAGS_PING | FD_QUIC_CONN_FLAGS_PING_SENT ) )
-        && conn->state == FD_QUIC_CONN_STATE_ACTIVE ) {
+    if( !( conn->flags & ( FD_QUIC_CONN_FLAGS_PING | FD_QUIC_CONN_FLAGS_PING_SENT ) ) ) {
       conn->flags         |= FD_QUIC_CONN_FLAGS_PING;
       conn->upd_pkt_number = FD_QUIC_PKT_NUM_PENDING;     /* update to be sent in next packet */
     }


src/waltz/quic/fd_quic.c
```diff
@@ -876,7 +876,10 @@ fd_quic_tx_enc_level( fd_quic_conn_t * conn, int acks ) {
     return fd_quic_enc_level_appdata_id;
   }
 
-  if( conn->flags && conn->upd_pkt_number >= app_pkt_number ) {
+  /* only allow 1-RTT "flag" frames when connection is ACTIVE, to prevent e.g. early 1-RTT PINGs */
+  if( conn->flags
```
akhinvasara-jumptrading: While we're at it, let's ignore the CLOSE_SENT and PING_SENT flags. Also would be nice to get rid of the branches in evaluation (use & instead of &&).

src/waltz/quic/fd_quic.c
```diff
@@ -876,7 +876,10 @@ fd_quic_tx_enc_level( fd_quic_conn_t * conn, int acks ) {
...
+      && conn->upd_pkt_number >= app_pkt_number
+      && conn->state == FD_QUIC_CONN_STATE_ACTIVE ) {
```
akhinvasara-jumptrading: Instead of state ACTIVE, let's check that we have 1-rtt keys available. 
fd_uint_extract_bit( conn->keys_avail, fd_quic_enc_level_appdata_id )

src/waltz/quic/fd_quic.c
```diff
@@ -2884,7 +2887,8 @@ fd_quic_svc_poll( fd_quic_t *      quic,
       }
     } else if( quic->config.keep_alive & !!(conn->let_die_time_ns > now) ) {
       /* send PING */
-      if( !( conn->flags & FD_QUIC_CONN_FLAGS_PING ) ) {
+      if( !( conn->flags & ( FD_QUIC_CONN_FLAGS_PING | FD_QUIC_CONN_FLAGS_PING_SENT ) )
+          && conn->state == FD_QUIC_CONN_STATE_ACTIVE ) {
```
akhinvasara-jumptrading: We should def check PING_SENT here, good catch thank you!

Also, let's simplify by removing the STATE_ACTIVE check here and at the other call site as wel, because you've fixed tx_enc_level to only return appdata enc_level once the keys are available. The gen_frames logic only tries sending these flag-based frames if enc_level is appdata.

akhinvasara-jumptrading (changes requested):
Great catches, thank you! 🔥  Left a couple suggestions to tighten it up and grab some adjacent wins.

src/waltz/quic/fd_quic.c
losman0s: Done

src/waltz/quic/fd_quic.c
losman0s: I actually used that check initially, but changed to ACTIVE state check since, iiuc:
- client-side 1-RTT keys are available as soon as `Handshake` is received in response to `Initial` (at which point client is in `HANDSHAKE_COMPLETE`)
- **server still rejects short headers until its connection is in `Established` state**, which only happens once it receives `Handshake` back from the client.
- the earliest safe signal that we can send 1-RTT is if we have 1-RTT keys locally & `peer_enc_level` == `appdata`, which anyway only happens when we receive HANDSHAKE_DONE / transition to ACTIVE (unless there is some packet reordering, but idk how often that happens/significant it is)

Lmk if you still want me to amend.

src/waltz/quic/fd_quic.c
losman0s: Done. I thought this was avoiding some wasteful code paths downstream, but I see that they're all already avoided due to `enc_level != appdata`.
And the re-arming assignments are probably insignificant (I assume it's not an issue to have the PINGs re-armed potentially multiple times before one is actually sent).

src/waltz/quic/fd_quic.c
akhinvasara-jumptrading: tldr: yes pls, amending would be great :) thank you!

I think amending would still be good, because waiting for the handshake done turns the 1-rtt into 2-rtt. 
It's def buggy to send 1-rtt before we have the keys for it - but the uncommon case of packet reordering isn't worth always waiting the extra round trip. 
It's also OK for the server to fail to decrypt, it should just drop the packet, not kill the connection. And we will just retx the packet anyway.

losman0s:
I pushed a minimal revision of the `test_quic_bw` test, which started to fail with this commit.

It looks like the test couldn't handle the new flow:
1. client `PING` is not gated by `ACTIVE`, so `PING` gets sent as soon as 1-RTT keys are received from server
2. server sends `HANDSHAKE_DONE` as soon as it receives `PING` (vs previously when it received `Hanshake` back from client)
3. client receives `HANDSHAKE_DONE` while still in `HANDSHAKE` (vs previously `HANDSHAKE_COMPLETE`), which it ignores
4. client only receives and processes `HANDSHAKE_DONE` at retry from server

The test was only cranking 20 iterations as fast as possible, even though it uses wall clock, so it was never reaching the retry.

One thing about it is that it shows that by sending 1-RTT `PING` as soon as we have the keys, we might be introducing a systematic 500ms delay to HS completion, which could be worse than a RTT from just gating `appdata` enc_level on `ACTIVE` state?

Lmk what you want to do.

PR #6820, opened 2025-10-17
snapshots: rename snaprd to snapct

amass-jump:


amass-jump:
This PR has no functional changes, just a simple refactor. ct is short for "control", as this tile no longer "reads" anything, it just controls the pipeline

mmcgee-jump (approved):

PR #6831, opened 2025-10-21
fix: Prevent integer underflow in Ghost consensus vote processing

vah13:
Integer Underflow (CWE-190) in Ghost Consensus

VULNERABILITY DESCRIPTION:
The Ghost consensus algorithm had a integer underflow vulnerability in fd_ghost_replay_vote() when processing stake-weighted vote updates. The code detected overflow conditions but only logged them and continued execution with corrupted values, allowing consensus weight manipulation.

VULNERABLE CODE LOCATIONS:
1. Line 415-420: prev->replay_stake subtraction (FIXED: early return)
2. Line 423-426: ancestor->weight subtraction (FIXED: clamp to zero)

SECURITY IMPACT:
- Consensus integrity compromise
- Fork choice manipulation
- Validator network split potential
- Byzantine fault tolerance violation

TESTING:
Fuzzing confirmed the vulnerability with payload triggering wraparound:
  prev_replay_stake: 9693583160302274054 - 9693583160302274182
  = 18446744073709551488 (wrapped value)


ripatel-fd:
> but only logged them and continued execution with corrupted values

Thank you for the contribution. FD_LOG_CRIT aborts the process and core dumps, it does not continue execution.

PR #6832, opened 2025-10-21
snapshots: restructure config options

amass-jump:


src/disco/topo/fd_topo.h
```diff
@@ -502,26 +502,34 @@ struct fd_topo_tile {
       int slices_fd;
     } shredcap;
 
-    struct {
+    struct fd_topo_tile_snapct {
```
amass-jump, line 503: Note: This change is so that I can copy this struct from the topo* in unprivliged_init to the tile's context. I don't see this pattern being done already, but it avoids having to define this struct a third time in the code (fd_config + fd_topo + snapct_ctx)

src/util/fd_util_base.h
```diff
@@ -391,6 +391,10 @@ __extension__ typedef unsigned __int128 uint128;
     (__typeof__(&(x)))_fd_aopm;                                                                \
   }))
 
+/* FD_ARRAY_CNT returns the number of elements of the given array. */
+
+#define FD_ARRAY_CNT(x) (sizeof(x)/sizeof((x)[0]))
+
```
amass-jump: Thoughts?

src/app/firedancer/config/default.toml
```diff
@@ -400,34 +338,97 @@ user = ""
...
+        #  /snapshot.tar.bz2
+        #  /incremental-snapshot.tar.bz2
+        #
+        # If these peers have snapshots at the given paths, they will be
+        # considered as additional sources for downloading snapshots
+        # when enabled alongside gossip and entrypoint sources.
+        # The url to fetch snapshots from.  The paths /snapshot.tar.bz2
+        # and /incremental-snapshot.tar.bz2 will be queried from this
+        # URL.
+        servers = []
```
amass-jump, line 408: Certainly open to ideas on what to name this... `http` is super confusing because gossip uses http too. One could argue that gossip peers are also "servers", though I was trying to indicate that this is a dedicated snapshot-serving thing

src/discof/restore/fd_snapct_tile.c
```diff
@@ -21,26 +21,34 @@
 
 #define NAME "snapct"
 
+/* FIXME: Implement gossip allow_list and block_list logic. */
+/* FIXME: Implement full_effective_age_cancel_threshold */
+/* FIXME: Implement min_speed_mib and other download health logic */
+/* FIXME: Implement max_retry_abort and retry logic in general */
```
amass-jump, line 27: I added a handful of FIXME's (here and other places throughout) that we should plan to tackle in followup PRs. Wanted to try to keep this one as small as possible

src/app/firedancer/config/testnet.toml
```diff
@@ -10,7 +10,3 @@
     max_account_records = 200_000_000
 [layout]
     gossvf_tile_count = 4
-[snapshots]
-    [[snapshots.sources.http]]
-        enabled = true
-        url = "http://solana-testnet-rpc.jumpisolated.com:8899"
```
amass-jump, line 16: Note: I don't think this should be in the default testnet config, but we can leave it in there for developer convenience for the time being?

src/app/firedancer/config/default.toml
```diff
@@ -264,6 +264,7 @@ user = ""
     # specified in `[net.interface]`.
     host = ""
 
+#TODO-AM: Tweak all the wording
```
amass-jump: I still need to clean up the comments here, but otherwise it should be good. Wanted to get the ball rolling on review while I worked on that

src/app/firedancer/config/default.toml
```diff
@@ -520,10 +518,6 @@ user = ""
         # is not recommended to change this setting.
         mean_cache_entry_size = 131072
 
-# This section configures the "groove" persistent account database.
-# [groove]
-# ...
-
```
amass-jump, line 517: ?

src/app/firedancer/config/default.toml
```diff
@@ -396,38 +333,89 @@ user = ""
...
         #
-        # If these peers have snapshots at the given paths, they will be
+        # If these servers have snapshots at the given paths, they will be
         # considered as additional sources for downloading snapshots
-        # when enabled alongside gossip sources.
-        [[snapshots.sources.http]]
-            # Whether to enable fetching snapshots from a list of
-            # HTTP peers.
-            enabled = false
+        # when enabled alongside gossip and entrypoint sources.
```
mmcgee-jump: This language is a little confusing, and to me implies they are only considered if gossip / entrypoint sources are also enabled

src/app/firedancer/config/default.toml
```diff
@@ -396,38 +333,89 @@ user = ""
...
+        # snapshot.  A local snapshot that has an older age than one
+        # we can download might actually result in a faster catchup
+        # time because of the faster loading speed.  The full age
+        # should be set higher because, as full snapshots are much
+        # larger than incremental snapshots, the time penalty for
+        # downloading a full snapshot is much higher.
+        max_local_full_effective_age = 1000
+        max_local_incremental_age = 200
+
+        # If any HTTP server URLs are listed, the following paths will be
```
mmcgee-jump, line 391: This can be an IP address or a domain right?

mmcgee-jump: might be worth clarifying

mmcgee-jump: (Probably some examples would help here too)

src/app/firedancer/config/default.toml
mmcgee-jump: addresses ? endpoints?

src/app/firedancer/config/default.toml
```diff
@@ -396,38 +333,89 @@ user = ""
...
+            # Allow downloading of snapshots from arbitrary gossip peers
+            # that are hosting a snapshot server.  This may allow faster
+            # snapshot download if a peer in a local data-center is
+            # serving a good snapshot, but it is extremely dangerous as
+            # the peer could serve you a specially crafted malicious
+            # snapshot with no way to tell.
+            allow_any = true
+
+            # Allow downloading of snapshots from the gossip peers
+            # identified by the given base58-encoded public keys.
```
mmcgee-jump: Maybe clarify that they must also be advertising snapshots in gossip

src/app/firedancer/config/default.toml
```diff
@@ -396,38 +333,89 @@ user = ""
...
+    [snapshots.download]
+        # The minimum acceptable speed for snapshot download from a peer, in
+        # megabits per second.  If the initial download speed falls below
+        # this threshold, the validator will abandon the download, find a
+        # new peer and start downloading the snapshot from them.
+        #
+        # The initial download speed is measured over every 10 mib for the
+        # first 400 mib of downloaded content.  The validator continues
+        # to measure download speed over every 10 mib until the download is
+        # complete, but only the initial download speed is used to decide
```
mmcgee-jump: A little unclear to me what the relevance of "continuing to measure" is? Why do we do that?

src/app/firedancer/config/default.toml
```diff
@@ -396,38 +333,89 @@ user = ""
...
+        # The maximum number of times to abort and retry when encountering a
+        # slow snapshot download.  If the maximum attempts is reached while
+        # downloading a snapshot the validator will exit with an error.
+        #
+        # The number of attempts is reset when starting a download for a
+        # new full or incremental snapshot.
+        #
+        # If set to zero, the validator will continually retry until
+        # completing snapshot download.
+        max_retry_abort = 5
```
mmcgee-jump: I think this language might of come from Agave but imo would be clearer just calling this "max_retries"

src/app/firedancer/config/default.toml
```diff
@@ -396,38 +333,89 @@ user = ""
...
+        # TODO: refine the initial download speed measurement and threshold.
+        #
+        # If set to zero, no minimum threshold will be applied.
+        min_speed_mib = 0
+
+        # The maximum number of times to abort and retry when encountering a
+        # slow snapshot download.  If the maximum attempts is reached while
+        # downloading a snapshot the validator will exit with an error.
+        #
+        # The number of attempts is reset when starting a download for a
```
mmcgee-jump: So we could try 25 different incrementals (5 each for 5 different fulls) before aborting?

src/app/firedancer/config/default.toml
mmcgee-jump: Yah fine to delete this kind of stuff

src/app/firedancer/config/testnet.toml
mmcgee-jump: Yeah let's leave it until the defaults are working well. I think we can also add an environment variable which adds to the servers[] list, and add this URL to that env var on all of our dev boxes

src/app/shared/fd_config.c
```diff
@@ -461,6 +461,13 @@ fd_config_validatef( fd_configf_t const * config ) {
   if( FD_UNLIKELY( config->layout.sign_tile_count < 2 ) ) {
     FD_LOG_ERR(( "layout.sign_tile_count must be >= 2" ));
   }
+
+  if( FD_UNLIKELY( config->snapshots.sources.gossip.allow_any && config->snapshots.sources.gossip.allow_list_cnt>0UL ) ) {
```
mmcgee-jump, line 465: Might be worth checking the allow/disallow lists don't overlap here for sanity also

src/util/fd_util_base.h
mmcgee-jump: Not generally a fan of this, kinda like Linus rant about `make_u32_from_two_u16` here: https://lore.kernel.org/lkml/CAHk-=wjLCqUUWd8DzG+xsOn-yVL0Q=O35U9D6j6=2DUWX52ghQ@mail.gmail.com/

mmcgee-jump (approved):
looks good! just some detail nits for the toml

src/discof/restore/fd_snapct_tile.c
```diff
@@ -64,29 +73,32 @@ typedef struct fd_snapct_gossip_ci_entry fd_snapct_gossip_ci_entry_t;
...
+    int dir_fd;
+    int full_snapshot_fd;
+    int incremental_snapshot_fd;
   } local_out;
 
-  uchar in_kind[ MAX_IN_LINKS ];
+  char http_full_snapshot_name[ PATH_MAX ];
+  char http_incr_snapshot_name[ PATH_MAX ];
+
+  void const * gossip_in_mem;
```
cali-jumptrading: I prefer not using a void const * here... it is a bit confusing that this is the gossip in link. Out of curiousity, why did you remove the typed anonymous struct definitions for the in links?

src/app/shared/fd_config.h
```diff
@@ -123,33 +122,33 @@ struct fd_configf {
...
-
     struct {
+      uint max_local_full_effective_age;
+      uint max_local_incremental_age;
 
       struct {
-        int enabled;
+        int   allow_any;
+        ulong allow_list_cnt;
+        char  allow_list[ 16 ][ FD_BASE58_ENCODED_32_SZ ];
```
cali-jumptrading: Also out of curiosity, is 16 an arbitrary number? I'm wondering if users will ever need more than 16 in their allow list. Could we have a defined constant for this and in other places in the snapshot config?

src/app/firedancer/config/default.toml
```diff
@@ -396,38 +333,89 @@ user = ""
...
+        # catchup time.  To that end, they are a tradeoff between the
+        # catchup time from a snapshot slot and download time of the
+        # snapshot.  A local snapshot that has an older age than one
+        # we can download might actually result in a faster catchup
+        # time because of the faster loading speed.  The full age
+        # should be set higher because, as full snapshots are much
+        # larger than incremental snapshots, the time penalty for
+        # downloading a full snapshot is much higher.
+        max_local_full_effective_age = 1000
+        max_local_incremental_age = 200
```
cali-jumptrading: maybe also explain the `max_local_incremental_age` parameter and how it differs from the `max_local_full_effective_age` parameter

src/discof/restore/fd_snapct_tile.c
```diff
@@ -748,13 +744,13 @@ after_credit( fd_snapct_tile_t *  ctx,
         ctx->flush_ack = 0;
         ctx->state = FD_SNAPCT_STATE_FLUSHING_FULL_HTTP_RESET;
         FD_LOG_WARNING(( "error downloading snapshot from http://" FD_IP4_ADDR_FMT ":%hu/snapshot.tar.bz2",
-                         FD_IP4_ADDR_FMT_ARGS( ctx->addr.addr ), ctx->addr.port ));
+                         FD_IP4_ADDR_FMT_ARGS( ctx->addr.addr ), fd_ushort_bswap( ctx->addr.port ) ));
```
cali-jumptrading, line 749: Out of curiosity, why is the port stored in network order? I recall there was an effort to keep all ports stored in host order in firedancer...

amass-jump: The definition of `fd_ip4_port_t` says the port is supposed to be in network order, and in general I would think it strange to store a network order IP addr next to a host order port

https://github.com/firedancer-io/firedancer/blob/main/src/util/net/fd_net_headers.h#L62

But I can make it all host order if we prefer that?

src/app/shared/fd_config.h
amass-jump: yeah, 16 is arbitrary. If an operator tried to use more in a config, the validator would not boot. How many more do you think we'd need?

We could add a constant but it would only get used in the one place here, I'm not sure how useful that is

src/discof/restore/fd_snapct_tile.c
cali-jumptrading: I think we had wanted to fix the port endianness confusion but never finished it. https://github.com/firedancer-io/firedancer/pull/6062

I don't care one way or the other, but @ripatel-fd can chime in here too.

src/app/firedancer/config/default.toml
amass-jump: I had a small note about that:

```
        #  The full age
        # should be set higher because, as full snapshots are much
        # larger than incremental snapshots, the time penalty for
        # downloading a full snapshot is much higher.
```

In general I didn't want to get too carried away with adding commentary on the config, as I think we should do a final cleanup pass at the end to rewrite config comments and whatnot, once all the logic/behavior/etc. has been finalized

src/discof/restore/fd_snapct_tile.c
amass-jump: I removed those as all of the members of those structs were/are unused for input links, only the base memory address is needed. What should the type of the pointer be? `void const *` seemed reasonable given I think this is just a pointer to memory but not of any particular type

src/app/firedancer/config/default.toml
amass-jump: I didn't touch the max_retry_abort or min_download_speed params at all, other than to move them around a bit. Neither of them are actually implmented yet, and I think maybe we should just remove them from the config and add them back in once we have decided on how we'd like that logic to work

src/discof/restore/fd_snapct_tile.c
cali-jumptrading: seems like the in link memory type was `fd_wksp_t *`. Also we had some assertions to check for overruns like https://github.com/firedancer-io/firedancer/blob/main/src/discof/restore/fd_snapct_tile.c#L804C3-L804C74 that used some of the other fields like wmark and chunk0.

Granted, we don't have strong abstractions for what an "in link" is. But in general, our codebase seems to follow this pattern... https://github.com/firedancer-io/firedancer/blob/main/src/discof/replay/fd_replay_tile.c#L92

Not a big deal, but I just don't prefer using `void const *` when possible.

src/app/shared/fd_config.h
cali-jumptrading: yeah... I'm not sure how much more we would need, but I could see a user having more than 16 known validators. In this case, a user may want to build up a larger number of trusted known validators, given that we don't implement the "trusted snapshot hashes" semantic that agave does. Since by default, we will enable `allow_any`, but encourage users to use the `allow_list` for safety reasons.

Maybe 32? or 64?

cali-jumptrading: Also... yeah ideally it'd be nice to have constants for these values but I can also see why you'd be against them. We don't have a shared constants header for config options so constants have to be defined multiple times in multiple places which defeats the point of having constants...

Could do something like this and define a local constant: https://github.com/firedancer-io/firedancer/blob/main/src/disco/topo/fd_topo.h#L196

amass-jump: Updated to 32. Also, I lied earlier, I forgot that we had the allow/block lists defined twice, once in fd_config.h and again in fd_topo.h. So I added a constant and used it in both places. Nice catch

contrib/offline-replay/offline_replay.toml
```diff
@@ -31,10 +31,11 @@
...
-    download = false
     incremental_snapshots = false
-    minimum_download_speed_mib = 0
-    maximum_local_snapshot_age = 0
-    maximum_download_retry_abort = 0
+    [snapshots.sources]
+        servers = []
+        [snapshots.sources.gossip]
+            allow_any = false
+            allow_list = []
```
amass-jump, line 39: @mmcgee-jump @cali-jumptrading Does anyone feel strongly about keeping a (redundant) `download=false` config option that is equivalent to `servers=[], gossip_any=false, gossip_list=[]` ? To me this seems to mostly be a testing / dev thing so not super critical to have (which is why I made changes as above), but happy to revert

cali-jumptrading: No, it doesn't seem necessary after these changes.

cali-jumptrading (approved):

PR #6838, opened 2025-10-21
restore: keep rec_pool lock while loading snap

ripatel-fd:
Remove some atomic CAS operations from snapshot loader

### DIFF
--- a/src/discof/restore/fd_snapin_tile.c
+++ b/src/discof/restore/fd_snapin_tile.c
@@ -72,6 +72,7 @@ struct fd_snapin_tile {
   uchar *         acc_data;
 
   fd_funk_txn_xid_t xid[1]; /* txn XID */
+  fd_funk_txn_t *   funk_txn;
 
   fd_stem_context_t *      stem;
   fd_ssparse_t *           ssparse;
@@ -123,6 +124,7 @@ static inline int
 should_shutdown( fd_snapin_tile_t * ctx ) {
   if( FD_UNLIKELY( ctx->state==FD_SNAPSHOT_STATE_SHUTDOWN ) ) {
     FD_LOG_NOTICE(( "loaded %.1fM accounts from snapshot in %.3f seconds", (double)ctx->metrics.accounts_inserted/1e6, (double)(fd_log_wallclock()-ctx->boot_timestamp)/1e9 ));
+    fd_funk_rec_pool_unlock( ctx->funk->rec_pool );
   }
   return ctx->state==FD_SNAPSHOT_STATE_SHUTDOWN;
 }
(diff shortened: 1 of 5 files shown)


src/discof/restore/fd_snapin_tile.c
```diff
@@ -611,6 +636,7 @@ handle_control_frag( fd_snapin_tile_t *  ctx,
         fd_funk_txn_remove_published( ctx->funk );
       } else {
         fd_funk_txn_cancel( ctx->funk, ctx->xid );
```
ripatel-fd, line 639: This will deadlock, need to temporarily release record lock before doing txn level operations.

src/discof/restore/fd_snapin_tile.c
```diff
@@ -123,6 +124,7 @@ static inline int
 should_shutdown( fd_snapin_tile_t * ctx ) {
   if( FD_UNLIKELY( ctx->state==FD_SNAPSHOT_STATE_SHUTDOWN ) ) {
     FD_LOG_NOTICE(( "loaded %.1fM accounts from snapshot in %.3f seconds", (double)ctx->metrics.accounts_inserted/1e6, (double)(fd_log_wallclock()-ctx->boot_timestamp)/1e9 ));
+    fd_funk_rec_pool_unlock( ctx->funk->rec_pool );
```
mmcgee-jump, line 127: do this in `FD_SNAPSHOT_MSG_CTRL_SHUTDOWN` not here

src/discof/restore/fd_snapin_tile.c
```diff
@@ -607,12 +632,15 @@ handle_control_frag( fd_snapin_tile_t *  ctx,
...
 
+      fd_funk_rec_pool_unlock( ctx->funk->rec_pool );
       if( ctx->full ) {
         fd_funk_txn_remove_published( ctx->funk );
       } else {
         fd_funk_txn_cancel( ctx->funk, ctx->xid );
+        ctx->funk_txn = NULL;
         fd_funk_txn_xid_copy( ctx->xid, fd_funk_last_publish( ctx->funk ) );
       }
+      fd_funk_rec_pool_lock( ctx->funk->rec_pool, 1 );
```
mmcgee-jump, line 643: Not here, I would lock on INIT messages, and unlock on FAIL

src/discof/restore/fd_snapin_tile.c
```diff
@@ -650,16 +681,20 @@ handle_control_frag( fd_snapin_tile_t *  ctx,
...
       }
       FD_TEST( !fd_funk_last_publish_is_frozen( ctx->funk ) );
 
       /* Make 'Last published' XID equal the restored slot number */
       fd_funk_txn_xid_t target_xid = { .ul = { ctx->bank_slot, 0UL } };
       fd_funk_txn_prepare( ctx->funk, ctx->xid, &target_xid );
       fd_funk_txn_publish_into_parent( ctx->funk, &target_xid );
+      ctx->funk_txn = NULL;
       fd_funk_txn_xid_copy( ctx->xid, &target_xid );
+      fd_funk_rec_pool_lock( ctx->funk->rec_pool, 1 );
```
amass-jump, line 697: Is there any need to re-lock here? No additional data/funk stuff should come after snapct has published a DONE control message. Further to what @mmcgee-jump has mentioned already, I think we can lock() on an INIT control message, and unlock() on NEXT/DONE/FAIL control messages. I don't think any additional locks or unlocks should be necessary, unless I'm missing something

Some context on this is that snapct tightly controls the state of the entire pipeline with these control messages and waiting for them to flush through the pipeline before proceeding to new states. There is more info in `src/discof/restore/utils/fd_ssctrl.h`

ripatel-fd:
@amass-jump , can you take over this PR?

PR #6839, opened 2025-10-21
Enable gossip-sourced snapshots in devnet.toml

ripatel-fd:


ripatel-fd:
Oh ok let's just merge it then when it's the default. Not urgent

PR #6842, opened 2025-10-22
gossip: tag bad CRDS values instead of dropping them in gossvf

ravyu-jump:


src/flamenco/gossip/fd_gossip.c
```diff
@@ -551,7 +554,7 @@ process_push_crds( fd_gossip_t *                       gossip,
                    uchar const *                       payload,
                    long                                now,
                    fd_stem_context_t *                 stem ) {
-  /* overrides_fast here, either count duplicates or purge if older (how!?) */
+  if( FD_UNLIKELY( value->gossvf_outcome!=GOSSVF_OUTCOME( SUCCESS_PUSH ) ) ) return -1;
```
mmcgee-jump, line 557: Do this check in the parent function

PR #6847, opened 2025-10-22
deps.sh: Arch Linux support

ripatel-fd:
So Eric can use Firedancer too

### DIFF
--- a/deps.sh
+++ b/deps.sh
@@ -284,11 +284,47 @@ check_macos_pkgs () {
   PACKAGE_INSTALL_CMD=( brew install ${MISSING_FORMULAE[*]} )
 }
 
+check_arch_pkgs () {
+  local REQUIRED_PKGS=(
+    base-devel        # C/C++ compiler, make, etc.
+    curl              # download rustup
+    zstd              # build system
+    cmake             # Agave (protobuf-src)
+    clang             # Agave (bindgen)
+    perl              # Agave (OpenSSL)
+    protobuf          # Agave, solfuzz
+    systemd-libs      # Agave
+  )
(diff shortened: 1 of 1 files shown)


deps.sh
```diff
@@ -284,11 +284,49 @@ check_macos_pkgs () {
   PACKAGE_INSTALL_CMD=( brew install ${MISSING_FORMULAE[*]} )
 }
 
+check_arch_pkgs () {
+  local REQUIRED_PKGS=(
+    base-devel        # C/C++ compiler, make, etc.
+    curl              # download rustup
+    pkgconf           # build system
```
esemeniuc: this is part of base-devel, can see https://archlinux.org/packages/core/any/base-devel/

deps.sh
```diff
@@ -284,11 +284,49 @@ check_macos_pkgs () {
...
+    zstd              # build system
+    gzip              # build system
```
esemeniuc: ditto

deps.sh
```diff
@@ -284,11 +284,49 @@ check_macos_pkgs () {
...
+    cmake             # Agave (protobuf-src)
+    clang             # Agave (bindgen)
+    perl              # Agave (OpenSSL)
+    protobuf          # Agave, solfuzz
+    systemd-libs      # Agave
+  )
+  if [[ $DEVMODE == 1 ]]; then
+    REQUIRED_PKGS+=( autoconf automake bison flex gettext gmp lcov )
```
esemeniuc:
```suggestion
    REQUIRED_PKGS+=( gmp lcov )
```
base-devel includes first few

PR #6849, opened 2025-10-22
Remove valloc

ripatel-fd:


ripatel-fd:
@kbhargava-jump can you check whether fd_ledger in offline replay still works>

kbhargava-jump:
> @kbhargava-jump can you check whether fd_ledger in offline replay still works>

yup this PR works fine for fd_ledger minify

PR #6856, opened 2025-10-22
genesis: always load genesis, verify hashes

mmcgee-jump:


src/discof/genesis/fd_genesi_tile.seccomppolicy
```diff
@@ -20,17 +21,57 @@ fsync: (eq (arg 0) logfile_fd)
...
+             (eq (arg 4) 0)
+             (eq (arg 5) 0))
+
+# client: need to read from connections established by the client
+#
+# arg 0 is the file descriptor that we recieve messages from.  We will
+# allow receiving to any of the descriptors opened in the client.  Only
+# restrict from receiving to the logfile and STDERR
+recvfrom: (and (not (or (eq (arg 0) logfile_fd)
+                        (eq (arg 0) 2)))
```
amass-jump, line 68: OOC, why not also checking e.g.g out_dir_fd & co for this & sendto?

src/discof/genesis/fd_genesi_tile.c
```diff
@@ -1,38 +1,90 @@
...
+            int    items,
+            int    size ) {
+  fd_alloc_t * alloc = (fd_alloc_t *)opaque;
+  ulong num = (ulong)items * (ulong)size;
+  void * result = fd_alloc_malloc( alloc, 16UL, num + 8UL );
+  if( FD_UNLIKELY( !result ) ) {
+    return NULL;
+  }
+  *(ulong *)result = num;
+  return (uchar *)result + 8UL;
```
amass-jump: OOC, why do you store the `num` here?

src/discof/genesis/fd_genesi_tile.c
```diff
@@ -187,11 +405,15 @@ privileged_init( fd_topo_t *      topo,
...
-  ctx->fd = open( tile->genesi.genesis_path, O_RDONLY|O_CLOEXEC );
-  if( FD_UNLIKELY( -1==ctx->fd ) ) {
+  ctx->local_genesis = 1;
+  ctx->in_fd = open( tile->genesi.genesis_path, O_RDONLY|O_CLOEXEC );
+  if( FD_UNLIKELY( -1==ctx->in_fd ) ) {
     if( FD_LIKELY( errno==ENOENT  ) ) {
+      FD_LOG_INFO(( "no local genesis.bin file found at `%s`", tile->genesi.genesis_path ));
+
       if( FD_UNLIKELY( !tile->genesi.entrypoints_cnt ) ) {
         FD_LOG_ERR(( "This node is bootstrapping the cluster as it has no gossip entrypoints provided, but "
```
amass-jump, line 411: What do you think about having a very explicit config option for bootstrapping a new cluster, rather than using the entrypoints size==0 check?

src/discof/genesis/fd_genesi_tile.c
```diff
@@ -287,8 +583,10 @@ populate_allowed_fds( fd_topo_t const *      topo,
 
 fd_topo_run_tile_t fd_tile_genesi = {
   .name                     = "genesi",
+  .rlimit_file_cnt_fn       = rlimit_file_cnt,
```
amass-jump, line 579: Why don't we need the allow_connect / allow_renameat flags?

mmcgee-jump: Haha, we do of course, but only causes a problem on ubuntu at the moment

src/discof/genesis/fd_genesi_tile.c
mmcgee-jump: Yeah it would probably make more sense, this overload is a bit strange (you could want to "rebootstrap" a cluster which does not perform genesis but has no entrypoint in concept). It kinda comes from Agave originally

src/discof/genesis/fd_genesi_tile.c
mmcgee-jump: Oops, had chatgpt write the malloc/free and it messed this up, I think its there to support a theoretical realloc, which would need to remember the size

src/app/firedancer/topology.c
```diff
@@ -862,6 +863,15 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
     tile->genesi.entrypoints_cnt        = config->gossip.entrypoints_cnt;
     fd_memcpy( tile->genesi.entrypoints, config->gossip.resolved_entrypoints, tile->genesi.entrypoints_cnt * sizeof(fd_ip4_port_t) );
 
+    tile->genesi.has_expected_genesis_hash = !!strcmp( config->consensus.expected_genesis_hash, "" );
+
+    if( FD_UNLIKELY( strcmp( config->consensus.expected_genesis_hash, "") && !fd_base58_decode_32( config->consensus.expected_genesis_hash, tile->genesi.expected_genesis_hash ) ) ) {
```
ripatel-fd:
```suggestion
    if( FD_UNLIKELY( strcmp( config->consensus.expected_genesis_hash, "" ) && !fd_base58_decode_32( config->consensus.expected_genesis_hash, tile->genesi.expected_genesis_hash ) ) ) {
```

src/discof/genesis/fd_genesi_tile.c
```diff
@@ -1,38 +1,90 @@
...
 #include <sys/stat.h>
 #include <unistd.h>
+#include <netinet/in.h>
+#include <linux/fs.h>
 
 #include "generated/fd_genesi_tile_seccomp.h"
-struct fd_genesi_tile {
-  int fd;
 
+#if __has_include("<bzlib.h>")
```
ripatel-fd: i think some old compilers might not support this, typically i see this pattern
```
#if defined(__has_include)
#if __has_include(...)
```
probably not worth doing though

src/discof/genesis/fd_genesi_tile.c
```diff
@@ -1,38 +1,90 @@
...
+#include <bzlib.h>
+#else
+#include "../../../opt/include/bzlib.h"
```
ripatel-fd: Don't do this, the prefix is opt-msan sometimes, just fail build

src/discof/genesis/fd_genesi_tile.c
```diff
@@ -1,38 +1,90 @@
...
+bz2_malloc( void * opaque,
+            int    items,
+            int    size ) {
+  fd_alloc_t * alloc = (fd_alloc_t *)opaque;
+  ulong num = (ulong)items * (ulong)size;
+  void * result = fd_alloc_malloc( alloc, 16UL, num + 8UL );
+  if( FD_UNLIKELY( !result ) ) {
+    return NULL;
+  }
+  *(ulong *)result = num;
```
ripatel-fd: can we ditch this, since this is only useful for realloc?

src/discof/genesis/fd_genesi_tile.c
```diff
@@ -95,6 +156,56 @@ initialize_accdb( fd_genesi_tile_t * ctx ) {
...
+  uchar mainnet_hash[ 32 ];
+  FD_TEST( fd_base58_decode_32( "5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d", mainnet_hash ) );
+
+  uchar testnet_hash[ 32 ];
+  FD_TEST( fd_base58_decode_32( "4uhcVJyU9pJkvQyS88uRDiswHXSCkY3zQawwpjk2NsNY", testnet_hash ) );
+
+  uchar devnet_hash[ 32 ];
+  FD_TEST( fd_base58_decode_32( "EtWTRABZaYq6iMfeYKouRu166VU2xqa1wcaWoxPkrZBG", devnet_hash ) );
+
+  switch ( genesis->cluster_type ) {
```
ripatel-fd:
```suggestion
  switch( genesis->cluster_type ) {
```

src/discof/genesis/fd_genesis_client.c
ripatel-fd, line 1: can you add a fuzzer for this guy, or add a github issue to add one later?

ripatel-fd (approved):

src/discof/genesis/fd_genesis_client.c
mmcgee-jump: Filed https://github.com/firedancer-io/firedancer/issues/6859

ripatel-fd (approved):

PR #6823, opened 2025-10-18
resolf: fix database accesses

ripatel-fd:
- **accdb: add admin API**
- **accdb: add peek API**
- **resolf: fix accdb races**
- **funk: remove unused txn APIs**

PR #6862, opened 2025-10-22
accdb: add admin API

ripatel-fd:
Add accdb API for fork operations


ripatel-fd:
Merging https://github.com/firedancer-io/firedancer/pull/6823 instead

PR #6865, opened 2025-10-22
Fix OOB shifts in fd_vm_interp_core (#3872)

### LINKED ISSUE #3872: OOB shifts in fd_vm_interp_core
Should probably use `fd_{uint,ulong}_shift_{left,right}` here. Should also check via `make -j asm` that the resulting bytecode doesn't get less efficient. 

```
src/flamenco/vm/fd_vm_interp_core.c:1049:46: runtime error: shift exponent 4294967295 is too large for 32-bit type 'int'
src/flamenco/vm/fd_vm_interp_core.c:1063:41: runtime error: shift exponent 18446744073709551615 is too large for 64-bit type 'long'
src/flamenco/vm/fd_vm_interp_core.c:1049:46: runtime error: shift exponent 4096 is too large for 32-bit type 'int'
```

Reproduce using UBSan on Clang

Ayushjhax:
## Summary
Fixes out-of-bounds (OOB) shift operations in the VM interpreter that were causing UBSan errors.

Closes #3872

## Problem
The VM interpreter had 4 arithmetic right shift (ARSH) operations using raw C shift operators without bounds checking:

- `0xc4` (ARSH_IMM): `(int)reg_dst >> imm` 
- `0xc7` (ARSH64_IMM): `(long)reg_dst >> imm`
- `0xcc` (ARSH_REG): `(int)reg_dst >> (uint)reg_src`
- `0xcf` (ARSH64_REG): `(long)reg_dst >> reg_src`

When shift amounts ≥ bit width (32 for int, 64 for long), this caused undefined behavior and UBSan errors:

## Solution
Replaced raw shift operators with safe helper functions from `fd_bits.h`:

- `fd_int_shift_right()` for 32-bit shifts
- `fd_long_shift_right()` for 64-bit shifts

These functions cap shift amounts to prevent UB while preserving arithmetic right shift semantics.

## Changes
- `src/flamenco/vm/fd_vm_interp_core.c`: Fixed 4 ARSH operations (lines 972, 985, 989, 1003)

## Testing
- No compilation errors
- UBSan errors eliminated  
- Existing ARSH tests pass
- Performance impact minimal (uses cmov instructions)

### DIFF
--- a/src/flamenco/vm/fd_vm_interp_core.c
+++ b/src/flamenco/vm/fd_vm_interp_core.c
@@ -976,7 +976,7 @@ interp_exec:
   /* 0xc0 - 0xcf ******************************************************/
 
   FD_VM_INTERP_INSTR_BEGIN(0xc4) /* FD_SBPF_OP_ARSH_IMM */
-    reg[ dst ] = (ulong)(uint)( (int)reg_dst >> imm ); /* FIXME: WIDE SHIFTS, STRICT SIGN EXTENSION */
+    reg[ dst ] = (ulong)(uint)fd_int_shift_right( (int)reg_dst, (int)imm );
   FD_VM_INTERP_INSTR_END;
 
   FD_VM_INTERP_BRANCH_BEGIN(0xc5) /* FD_SBPF_OP_JSLT_IMM */ /* FIXME: CHECK IMM SIGN EXTENSION */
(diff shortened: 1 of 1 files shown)


ripatel-fd (changes requested):
Thanks for the contribution. This is not an issue for the immediate shifts because those are checked by the bytecode verifier. For the register shifts, this is a known technical UB issue but doesn't cause any issues on any known GCC or Clang versions on x86. 

We'll investigate this a bit further before accepting this fix because performance numbers of the VM interpreter are very sensitive to changes.

Ayushjhax:
Thanks for the clarification that makes sense.

ripatel-fd:
We fixed this upstream so I'm going to close this. Thank you again for the contribution

PR #6870, opened 2025-10-23
snapshots: simplify gossip peer tracking logic

amass-jump:


src/discof/restore/fd_snapct_tile.c
```diff
@@ -856,65 +838,58 @@ gossip_frag( fd_snapct_tile_t *  ctx,
...
+      fd_ip4_port_t new_addr    = msg->contact_info.contact_info->sockets[ FD_CONTACT_INFO_SOCKET_RPC ];
+      new_addr.port             = fd_ushort_bswap( new_addr.port );
+      fd_ip4_port_t gossip_addr = msg->contact_info.contact_info->sockets[ FD_CONTACT_INFO_SOCKET_GOSSIP ];
+      gossip_addr.port          = fd_ushort_bswap( gossip_addr.port );
+      int addr_is_entrypoint    = is_entrypoint( ctx, gossip_addr );
+
+      if( FD_LIKELY( (ctx->config.entrypoints_enabled && addr_is_entrypoint) || ctx->config.gossip_peers_enabled ) ) {
+        fd_snapct_gossip_ci_entry_t * entry = ctx->gossip.ci_table + msg->contact_info.idx;
+        if( FD_UNLIKELY( !fd_pubkey_eq( &entry->pubkey, (fd_pubkey_t const *)msg->origin_pubkey ) ) ) {
+          FD_TEST( fd_pubkey_check_zero( &entry->pubkey ) );
```
amass-jump, line 854: The main idea behind these changes is that the gossip frags tell us the exact index to use for this contact info / peer, so we don't need a pool, we just need an array. We still have the map overlaid on top of the array so we can do fast lookups by pubkey (e.g. for snapshot hashes or to iterate the table)

src/discof/restore/fd_snapct_tile.c
```diff
@@ -856,65 +838,58 @@ gossip_frag( fd_snapct_tile_t *  ctx,
...
+          FD_TEST( ULONG_MAX==gossip_ci_map_idx_query_const( ctx->gossip.ci_map, (fd_pubkey_t const *)msg->origin_pubkey, ULONG_MAX, ctx->gossip.ci_table ) );
```
amass-jump, line 855: Sanity check that there are no other CIs in the map with this pubkey. Not strictly required but a nice safety check

src/discof/restore/fd_snapct_tile.c
```diff
@@ -1267,11 +1242,12 @@ unprivileged_init( fd_topo_t *      topo,
   if( FD_UNLIKELY( !tile->snapct.maximum_download_retry_abort ) ) ctx->config.maximum_download_retry_abort = UINT_MAX;
   else                                                            ctx->config.maximum_download_retry_abort = tile->snapct.maximum_download_retry_abort;
 
-  ctx->selector = fd_sspeer_selector_join( fd_sspeer_selector_new( _selector, FD_SSPING_MAX_PEERS, ctx->config.incremental_snapshot_fetch, 1UL ) );
+  ctx->selector = fd_sspeer_selector_join( fd_sspeer_selector_new( _selector, TOTAL_PEERS_MAX, ctx->config.incremental_snapshot_fetch, 1UL ) );
+
+  fd_memset( _ci_table, 0, sizeof(fd_snapct_gossip_ci_entry_t) * GOSSIP_PEERS_MAX );
```
mmcgee-jump, line 1247: Is this memset needed? I like to avoid these if possible since across the codebase they add up to some measurable boot time overhead

amass-jump: We do need to initialize the table elements, at least the pubkey field to zero. I could replace this with a for loop over the elements to set pubkey to zero if that'd be better?

cali-jumptrading (approved):

PR #6876, opened 2025-10-23
bank: clarify spad usage

mmcgee-jump:


ripatel-fd (commented):
I find this construction worse ...

ibhatt-jumptrading:
spad is no longer used

PR #6878, opened 2025-10-23
xdp: work around Clang compiler bug

ripatel-fd:
Works around a rare BPF_PROG_LOAD failure with ASan due to a Clang
compiler bug


ripatel-fd:
To reproduce

```
$ clang --version
clang version 19.1.7 (RESF 19.1.7-1.el9)
Target: x86_64-redhat-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Configuration file: /etc/clang/x86_64-redhat-linux-gnu-clang.cfg

$ make -j CC=clang EXTRAS=asan
```

PR #6886, opened 2025-10-23
Fix accidentally negated bool check for block builder in bundle tile

esemeniuc:
We should fetch when we don't have the builder info **OR** its expired. Reduces number of network requests to the block engine


ripatel-fd (approved):
Thank you

PR #6888, opened 2025-10-23
ssping: fix uninitialized read

ripatel-fd:


ripatel-fd:
@amass-jump This fixes an uninitialized read on `peer->deadline_nanos` found by MSan. Could you review this fix? I'm not sure about the ssping logic, so I'm not sure if it logically makes sense, but at least it fixes the crash.

src/discof/restore/utils/fd_ssping.c
```diff
@@ -358,6 +359,7 @@ send_pings( fd_ssping_t *     ssping,
       .msg_iov = iovs + msg_cnt,
       .msg_iovlen = 1,
     };
+    msgs[ msg_cnt ].msg_len = 0;
```
amass-jump, line 362: Did MSAN complain about this? This is supposed to be an "output-only" field (that we don't need), so technically I don't think it's incorrect to leave uninitialized?

src/discof/restore/utils/fd_ssping.c
```diff
@@ -226,6 +226,7 @@ fd_ssping_add( fd_ssping_t * ssping,
     if( FD_UNLIKELY( !peer_pool_free( ssping->pool ) ) ) return;
     peer = peer_pool_ele_acquire( ssping->pool );
     FD_TEST( peer );
+    memset( peer, 0, sizeof(fd_ssping_peer_t) );
```
amass-jump, line 229: Theoretically speaking we do not need to initialize `deadline_nanos` for peers in the `unpinged` state, but it's fine to do so anyway to silence MSAN warnings. However I think it'd be better to directly init fields to 0 where applicable rather than memset the whole thing

src/discof/restore/utils/fd_ssping.c
ripatel-fd: No it did not complain about this

PR #6905, opened 2025-10-24
restore: prefetch map chain

ripatel-fd:
Optimize snapshot loading by prefetching the index in 16-wide
batches.  Also fix a bug in funk_base that selected a worse hash
function than xxh128.

### DIFF
--- a/src/discof/restore/fd_snapin_tile.c
+++ b/src/discof/restore/fd_snapin_tile.c
@@ -487,6 +487,110 @@ process_account_data( fd_snapin_tile_t *            ctx,
   ctx->acc_data += result->account_data.data_sz;
 }
 
+/* streamlined_insert inserts an unfragmented account.
+   Only used while loading a full snapshot, not an incremental. */
+
+static void
+streamlined_insert( fd_snapin_tile_t * ctx,
+                    fd_funk_rec_t *    rec,
+                    uchar const *      frame,
+                    ulong              slot ) {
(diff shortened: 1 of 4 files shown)


src/discof/restore/fd_snapin_tile.c
```diff
@@ -487,6 +487,111 @@ process_account_data( fd_snapin_tile_t *            ctx,
...
+  fd_funk_rec_map_shmem_private_chain_t * chain_tbl = fd_funk_rec_map_shmem_private_chain( rec_map->map, 0UL );
+
+  /* Derive map chains */
+  uint chain_idx[ FD_SSPARSE_ACC_BATCH_MAX ];
+  ulong chain_mask = rec_map->map->chain_cnt-1UL;
+  for( ulong i=0UL; i<FD_SSPARSE_ACC_BATCH_MAX; i++ ) {
+    uchar const * frame  = result->account_batch.batch[ i ];
+    uchar const * pubkey = frame+0x10UL;
+    fd_funk_rec_key_t key = FD_LOAD( fd_funk_rec_key_t, pubkey );
+    ulong         memo   = fd_funk_rec_key_hash( &key, rec_map->map->seed );
```
ripatel-fd: this should use hash1

amass-jump:
What kind of increase to accounts/sec did you see from this?

ripatel-fd:
> What kind of increase to accounts/sec did you see from this?

@amass-jump About 1.6x throughput vs previous commit for a mainnet snapshot

PR #6906, opened 2025-10-24
restore: ILP hashmap walk

ripatel-fd:
Uses instruction-level parallelism to parallel-query a hashmap.
Slightly improves over just prefetching the chain heads during
later stages of loading a snapshot.

Also improves stat rendering in 'firedancer-dev snapshot-load'.


src/app/firedancer/topology.c
```diff
@@ -356,10 +356,10 @@ fd_topo_initialize( config_t * config ) {
 
   if( FD_LIKELY( snapshots_enabled ) ) {
   /* TODO: Revisit the depths of all the snapshot links */
```
amass-jump, line 358: Can you please mirror these changes into backtest.c & snapshot_load.c?

amass-jump (dismissed):

amass-jump (dismissed):

PR #6911, opened 2025-10-27
vinyl: I/O layer

ripatel-fd:


src/vinyl/bstream/fd_vinyl_bstream.h
```diff
@@ -0,0 +1,469 @@
...
+
+FD_FN_CONST static inline ulong
+fd_vinyl_bstream_pair_sz( ulong val_esz ) {
+  return fd_ulong_align_up( sizeof(fd_vinyl_bstream_phdr_t) + val_esz + FD_VINYL_BSTREAM_FTR_SZ, FD_VINYL_BSTREAM_BLOCK_SZ );
+}
+
+/* fd_vinyl_bstream_hash returns a 64-bit hash of the 64-bit seed and
+   the sz byte buffer buf.  Assumes buf is points to well aligned range
+   of blocks stable for the duration of the call and sz is a multiple of
+   FD_VINYL_IO_BLOCK_SZ.  sz 0 returns seed.  Retains no interest in
```
two-heart: `FD_VINYL_IO_BLOCK_SZ` is not defined anywhere

src/vinyl/bstream/fd_vinyl_bstream.c
```diff
@@ -0,0 +1,260 @@
...
+
+  if( FD_UNLIKELY( !block ) ) return "NULL block";
+  if( FD_UNLIKELY( !dst   ) ) return "NULL dst";
+
+  ulong move_info_sz = block->move.info_sz;
+
+  int bad_ctl        = (block->move.ctl != fd_vinyl_bstream_ctl( FD_VINYL_BSTREAM_CTL_TYPE_MOVE,
+                                                                 FD_VINYL_BSTREAM_CTL_STYLE_RAW,
+                                                                 FD_VINYL_BSTREAM_BLOCK_SZ ) );
+  int bad_match      = fd_vinyl_seq_eq( block->move.seq, seq );
```
two-heart: inverted?

src/vinyl/meta/fd_vinyl_meta.c
```diff
@@ -0,0 +1,162 @@
...
+    ulong start_idx = ele0[ ele_idx ].memo & (ele_max-1UL);
+
+    if( !( ((hole_idx<start_idx) & (start_idx<=ele_idx)                       ) |
+           ((hole_idx>ele_idx) & ((hole_idx<start_idx) | (start_idx<=ele_idx))) ) ) {
+
+      ulong line_idx = ele0[ ele_idx ].line_idx;
+      if( FD_LIKELY( line_idx< line_cnt  ) ) {
+        (void)line;
+//      FD_CRIT( line[ line_idx ].ele_idx==ele_idx, "corruption detected" );
+//      line[ line_idx ].ele_idx = hole_idx;
```
two-heart, line 146: looks unfinished

src/vinyl/bstream/fd_vinyl_bstream.h
kbowers-jump: Yep ... should be FD_VINYL_BSTREAM_BLOCK_SZ.

src/vinyl/bstream/fd_vinyl_bstream.c
kbowers-jump: Yep ... as per the top level commit comments, move support isn't tested yet (which is why it didn't go in as a PR right off the bat ... Richie and Javier wanted to start kicking the tires as the lower layers they need should be well tested).

src/vinyl/meta/fd_vinyl_meta.c
kbowers-jump: Hmmm ... line 144 should be deleted and the next two lines should be uncommented in a later commit.  What's going on here is that I broke the giant mass of code into a bunch of commits that could be reviewed and tested individually.

The vinyl "meta" commit ended up before the vinyl "line" commit.  To keep the vinyl "meta" commit working without the "line" commit, the minimal amount of stuff needed in the meta data structure to support fast cache line eviction sequence management is commented out in the "meta" commit.  The "line" commit should finish this.

Since Richie and Javier want to start using the lower layers of vinyl now, I'd either include the "line" commit in this PR or keep this as is with the understanding that the cache line operation support will be filled in with a later PR.

Regardless, good catch ... you are doing a thorough job.

ripatel-fd: Will fix in a future PR

PR #6912, opened 2025-10-27
restore: initial vinyl integration

ripatel-fd:


src/app/firedancer/config/default.toml
```diff
@@ -491,6 +493,13 @@ user = ""
     # setting.
     max_database_transactions = 2048
 
+# FIXME rename 'vinyl' to human-readable name like 'accounts.root_db'
+[vinyl]
+    enabled = false
+    path = "/home/{user}/.firedancer/{name}.accounts.vinyl"
```
mmcgee-jump, line 499: Path goes in `[paths]` section not here, and follow convention of other paths

src/app/firedancer/config/default.toml
```diff
@@ -491,6 +493,13 @@ user = ""
     # setting.
     max_database_transactions = 2048
 
+# FIXME rename 'vinyl' to human-readable name like 'accounts.root_db'
```
mmcgee-jump, line 496: Maybe just do this here? I would just merge both of them into [accdb] or whatever

ripatel-fd: I'm not going to do refactors that affect funk in this PR because this is going to stay a deep integration branch.
We can do it after it's merged.

PR #6913, opened 2025-10-27
restore: use O_DIRECT to create vinyl DB

ripatel-fd:


src/discof/restore/utils/fd_vinyl_io_wd.c
```diff
@@ -0,0 +1,382 @@
...
+
+#include <errno.h>
+#include <unistd.h>
+
+static inline void
+wd_write( int          fd,
+          ulong        off,
+          void const * buf,
+          ulong        sz ) {
+  ssize_t ssz = pwrite( fd, buf, sz, (off_t)off );
```
ripatel-fd, line 11: write with O_DIRECT will block until the DMA transaction is complete. This is silly because we could double-buffer and pipeline another write while the device is busy. io_uring is one option, but for simplicity, we should probably use a worker thread here (is practically always blocked on disk I/O, therefore never scheduled to CPU / very low overhead)

ripatel-fd: @mmcgee-jump please advise how we can spawn another thread in the same tile to schedule background I/O ops. Our only two reasonable choices are: using worker threads, or using io_uring. The former is simpler and more secure. 

A worker thread would be scheduled to the same CPU as the main snapin tile and will synchronize via low-overhead futexes (probably via POSIX API).

PR #6915, opened 2025-10-27
Remove C++ support

ripatel-fd:


ripatel-fd:
Do not merge, this affects Shelby

PR #6931, opened 2025-10-27
vm, serialization: store offsets relative to the start of each region

topointon-jump:
If there are gaps between the regions, the serialization offsets could be calculated incorrectly.


src/flamenco/runtime/program/fd_bpf_loader_serialization.c
```diff
@@ -164,7 +161,7 @@ write_account( fd_borrowed_account_t *   account,
   uchar const * data = account ? fd_borrowed_account_get_data( account )     : NULL;
   ulong         dlen = account ? fd_borrowed_account_get_data_len( account ) : 0UL;
 
-  acc_region_metas[instr_acc_idx].original_data_len = dlen;
+    acc_region_metas[instr_acc_idx].original_data_len = dlen;
```
mjain-jump: extra indent

mjain-jump (dismissed):

PR #6946, opened 2025-10-28
sched: store staged_head_bank_idx before potentially overwriting it

two-heart:


src/discof/replay/fd_sched.c
```diff
@@ -1506,7 +1514,7 @@ subtree_abandon( fd_sched_t * sched, fd_sched_block_t * block ) {
       }
     }
 
-    if( FD_UNLIKELY( in_order && block->staged && sched->active_bank_idx==sched->staged_head_bank_idx[ block->staging_lane ] ) ) {
```
yufeng-jump, line 1509: Instead of caching the predicate, could we move this if block to before the if( abandon ) block?

PR #6947, opened 2025-10-28
dockerfiles for gcc-8, gcc-9 and gcc-10

anwayde:
The next step would be to update `build.sh` to use these and perform builds for gcc-8, gcc-9 and gcc-10 in containers, which would be do-able even on RHEL/Rocky 9 hosts.


ripatel-fd:
What's the purpose of these?

PR #6949, opened 2025-10-28
fd_config: fix bundle tile keepalive config check

esemeniuc:
Error will never be printed as it impossible to match both predicates


ripatel-fd:
Thank you

PR #6968, opened 2025-10-28
flamenco, runtime: remove unused block error macros

topointon-jump:


topointon-jump:
we will actually use these when we add block-level verification

PR #6978, opened 2025-10-29
elf: solfuzz v2 elf harness

mjain-jump:


ripatel-fd (dismissed):
If we're vendoring flatbuffers code or check in generated code, please add the involved libraries and their license excerpts to NOTICE

src/flamenco/runtime/tests/fetch_and_generate.sh
```diff
@@ -2,7 +2,8 @@
 set -euo pipefail
 
 # Allow overriding proto version; default pinned
-PROTO_VERSION="${PROTO_VERSION:-v1.0.5}"
+PROTO_VERSION="${PROTO_VERSION:-v3.0.0}"
+FLATCC="${FLATCC:-../../../../opt/bin/flatcc}"
```
cmoyes-jump: seems like we should define a PROJ_ROOT or something similar and base everything off that

deps.sh
```diff
@@ -133,15 +133,16 @@ fetch () {
...
-  checkout_repo lz4       https://github.com/lz4/lz4                "v1.10.0"
-  checkout_repo s2n       https://github.com/awslabs/s2n-bignum     "" "4d2e22a"
-  checkout_repo openssl   https://github.com/openssl/openssl        "openssl-3.6.0"
-  checkout_repo secp256k1 https://github.com/bitcoin-core/secp256k1 "v0.7.0"
+  checkout_repo zstd      https://github.com/facebook/zstd            "v1.5.7"
+  checkout_repo lz4       https://github.com/lz4/lz4                  "v1.10.0"
+  checkout_repo s2n       https://github.com/awslabs/s2n-bignum       "" "4d2e22a"
+  checkout_repo openssl   https://github.com/openssl/openssl          "openssl-3.6.0"
+  checkout_repo secp256k1 https://github.com/bitcoin-core/secp256k1   "v0.7.0"
+  checkout_repo flatcc    https://github.com/firedancer-io/flatcc.git "master"
```
ripatel-fd: Why do we maintain a fork? Ideally we'd not do this so it's easier to track upstream bug fixes

mjain-jump: The latest version (0.6.1) causes deps.sh to fail. this is fixed in flatcc master, but I don’t want to pin deps.sh to master in case of breaking changes so I made a fork to keep it stable

src/flamenco/runtime/tests/fetch_and_generate.sh
mjain-jump: fixed

cmoyes-jump (approved):

PR #6980, opened 2025-10-29
fuzz: move proto-specific functions away from solfuzz-generic headers

mjain-jump:


src/flamenco/runtime/tests/fd_solfuzz_codec.h
```diff
@@ -0,0 +1,65 @@
...
+/* fd_solfuzz_codec.h provides Protobuf encoding / decoding helper
+   functions for all harnesses. */
+
+#include "../../../ballet/nanopb/pb_encode.h"
+#include "../../../ballet/nanopb/pb_decode.h"
+
+FD_PROTOTYPES_BEGIN
+
+static FD_FN_UNUSED void *
+sol_compat_decode_lenient( void *               decoded,
```
ripatel-fd, line 13: The sol_compat namespace should only be used in the public ABI. 
use fd_solfuzz for everything else.

src/flamenco/runtime/tests/fd_solfuzz_codec.h
```diff
@@ -0,0 +1,65 @@
...
+    return NULL;
+  }
+  ulong size;
+  if( FD_UNLIKELY( !pb_get_encoded_size( &size, decode_type, decoded ) ) ) {
+    pb_release( decode_type, decoded );
+    return NULL;
+  }
+  if( FD_UNLIKELY( size != in_sz ) ) {
+    pb_release( decode_type, decoded );
+    return NULL;
```
ripatel-fd, line 44: why do we need this?

src/flamenco/runtime/tests/fd_harness_common.h
ripatel-fd, line 1: please keep this in fd_solfuzz_private.h, no point in moving it

ripatel-fd (changes requested):

PR #6984, opened 2025-10-29
quic: fd_quic_connect failure frees tls_hs

akhinvasara-jumptrading:
This PR fixes a tls_hs leak inside fd_quic_connect if fd_quic_tls_hs_new set an alert on the tls_hs.


nlgripto:
closes https://github.com/firedancer-io/auditor-internal/issues/332

PR #7000, opened 2025-10-30
gui: fix transactions chart for firedancer

jherrera-jump:
transactions chart is minimally useable, doesn't support transaction state timestamps yet


mmcgee-jump (dismissed):
nice

PR #7006, opened 2025-10-31
progcache: fix invalidation GC

ripatel-fd:
fd_progcache must maintain an invariant that no cache invalidation
records exist at the root.  Adds missing garbage collection for
this case.  Fixes a bug that results in invalidations incorrectly
being reordered with newly rooted records.

### DIFF
--- a/src/flamenco/progcache/fd_progcache_user.c
+++ b/src/flamenco/progcache/fd_progcache_user.c
@@ -649,6 +649,11 @@ fd_progcache_invalidate( fd_progcache_t *          cache,
 
   if( FD_UNLIKELY( !cache || !funk->shmem ) ) FD_LOG_CRIT(( "NULL progcache" ));
 
+  /* Select a fork node to create invalidate record in
+     Do not create invalidation records at the funk root */
+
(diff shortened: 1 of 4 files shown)


src/flamenco/progcache/fd_progcache_user.c
```diff
@@ -203,6 +203,14 @@ fd_progcache_search_chain( fd_progcache_t const *    cache,
   for( ulong i=0UL; i<cnt; i++, ele_idx=rec_tbl[ ele_idx ].map_next ) {
     fd_funk_rec_t * rec = &rec_tbl[ ele_idx ];
 
+    //FD_LOG_NOTICE(( "found ele_idx=%u xid %lu:%lu rec %016lx%016lx%016lx%016lx",
```
ibhatt-jumptrading: delete this

ripatel-fd: I kept this in deliberately for future debugging

mjain-jump: remove / fd log debug

ripatel-fd: see above comment

PR #7012, opened 2025-10-31
snapshots: more accurate ssping latencies and fix crash

amass-jump:


src/discof/restore/utils/fd_ssping.c
```diff
@@ -313,9 +312,10 @@ recv_pings( fd_ssping_t * ssping,
     fd_ssping_peer_t * peer = peer_map_ele_query( ssping->map, &key, NULL, ssping->pool );
     if( FD_UNLIKELY( peer==NULL || ( peer->state!=PEER_STATE_PINGED && peer->state!=PEER_STATE_REFRESHING ) ) ) continue;
 
+    long now = fd_log_wallclock();
+
     deadline_list_ele_remove( peer->state==PEER_STATE_PINGED ? ssping->pinged : ssping->refreshing, peer, ssping->pool );
-    FD_TEST( peer->deadline_nanos-PEER_DEADLINE_NANOS_PING<now );
-    peer->latency_nanos  = (ulong)(now - (peer->deadline_nanos - PEER_DEADLINE_NANOS_PING));
+    peer->latency_nanos  = (ulong)fd_long_max( now - (peer->deadline_nanos - PEER_DEADLINE_NANOS_PING), 1L );
```
cali-jumptrading, line 318: confused what a latency of 1 means.

cali-jumptrading: In what cases could we get a latency of 1?

amass-jump: This is mainly to prevent cases where the latency could be computed as negative (plus I figured zero might be more problematic than 1 and there is no effective difference)

I think this FD_TEST was hit for one of two reasons
- the system clock moved backwards (not all that uncommon)
- a peer was pinged and the response was received in the same call to fd_ssping_advance(). This could happen if we are busy doing a lot of work in that one function call

In general moving the timestamp to this loop makes the ping latencies more accurate in cases where the `now` timestamp is stale

cali-jumptrading (approved):

src/discof/restore/utils/fd_ssping.c
mmcgee-jump: It goes backwards as it's using CLOCK_REALTIME, which is system adjustable. Maybe the right fix here might be to switch these measurements to use CLOCK_MONOTONIC or something like that? Not sure if MONOTONIC can go backwards either

PR #7017, opened 2025-10-31
restore: vinyl snapshot loader

ripatel-fd:
This patch integrates the snapshot loader with Firedancer's new disk-backed account database.
Adds a line-rate capable O_DIRECT write path with cache bypass. 
Has minimal impact to the existing funk-based database logic, except for minor cleanups along the way.

- **restore: separate out funk DB logic**
- **restore: initial vinyl integration**
- **restore: async O_DIRECT via worker thread**


src/tango/dcache/test_dcache.c
```diff
@@ -1,28 +1,31 @@
 #include "../fd_tango.h"
+#include "fd_dcache_private.h"
+
+FD_STATIC_ASSERT( sizeof(fd_dcache_private_hdr_t)==FD_DCACHE_ALIGN, unit_test );
```
jvarela-jump, line 6:
```suggestion
FD_STATIC_ASSERT( alignof(fd_dcache_private_hdr_t)==FD_DCACHE_ALIGN, unit_test );

```

src/app/firedancer-dev/main.c
```diff
@@ -42,9 +45,13 @@ fd_topo_obj_callbacks_t * CALLBACKS[] = {
   &fd_obj_cb_banks,
   &fd_obj_cb_funk,
   &fd_obj_cb_bank_hash_cmp,
+  &fd_obj_cb_vinyl_meta,
+  &fd_obj_cb_vinyl_meta_ele,
   NULL,
 };
 
+extern configure_stage_t fd_cfg_stage_vinyl;
```
amass-jump, line 53: I think the others are in main.h?

jvarela-jump (changes requested):

src/tango/dcache/fd_dcache_private.h
```diff
@@ -11,7 +11,7 @@
 /* fd_dcache_private_hdr specifies the detailed layout of the shared
    memory region. */
 
-struct __attribute__((aligned(FD_DCACHE_ALIGN))) fd_dcache_private_hdr {
+struct __attribute__((aligned(128UL))) fd_dcache_private_hdr {
```
jvarela-jump:
```suggestion
struct __attribute__((aligned(FD_DCACHE_ALIGN))) fd_dcache_private_hdr {
```

src/disco/metrics/metrics.xml
```diff
@@ -1050,6 +1050,10 @@ metric introduced.
     <gauge name="AccountsInserted" summary="Number of accounts inserted during snapshot loading. Might decrease if snapshot load is aborted and restarted" />
 </tile>
 
+<tile name="snapwr">
+    <gauge name="VinylBytesWritten" summary="Number of bytes written so far to the vinyl snapshot file. Might decrease if snapshot creation is aborted and restarted" />
```
amass-jump, line 1055: Does this not have a `State` like all the other snapshot pipeline tiles?

lidatong (approved):

src/discof/restore/fd_snapwr_tile.c
```diff
@@ -0,0 +1,273 @@
...
+static void
+metrics_write( fd_snapwr_t * ctx ) {
+  FD_MGAUGE_SET( SNAPWR, VINYL_BYTES_WRITTEN, ctx->metrics.last_off );
+}
+
+/* handle_control_frag handles an administrative frag from the snapin
+   tile. */
+
+static void
+handle_control_frag( fd_snapwr_t * ctx,
```
amass-jump, line 184: It seems suboptimal that this snapshot pipeline tile does not participate in the same "lockstep" control flow process that all the others do. Namely, if snapwr is locked up on a long write() syscall, the rest of the pipeline may reset and start loading a totally new snapshot while snapwr is still writing the old one. Does this not matter?

ripatel-fd: > It seems suboptimal that this snapshot pipeline tile does not participate in the same "lockstep" control flow process that all the others do.

I agree. Unfortunately, in order to achieve good O_DIRECT bandwidth without using an async kernel API, blocking for a few milliseconds is necessary. Unless we use signals, but Firedancer's sandbox is not configured correctly for signal handlers

> Namely, if snapwr is locked up on a long write() syscall, the rest of the pipeline may reset and start loading a totally new snapshot while snapwr is still writing the old one.

@amass-jump The snapin tile blocks until the snapwr tile acknowledges the shutdown signal. So, I don't think there's any problematic race condition here.

src/app/firedancer-dev/main.c
ripatel-fd: @amass-jump The other configure stages are in src/app/shared. But vinyl is full-Firedancer specific, so vinyl should not be in shared. All the other full-Firedancer-specific tiles and obj callback declarations are in this file, so I thought it'd be appropriate to have vinyl here too for consistency.

src/disco/metrics/metrics.xml
ripatel-fd: Good point, I'll add this.

ripatel-fd: fixed.

src/app/firedancer-dev/commands/snapshot_load.c
```diff
@@ -193,7 +205,11 @@ snapshot_load_cmd_fn( args_t *   args,
   puts( "- stall: Waiting on upstream tile"         );
   puts( "- acc:   Number of accounts"               );
   puts( "" );
-  puts( "--------------------------------------------[ct],[ld],[dc],[in]--------[ct],[ld],[dc],[in]--------------" );
+  fputs( "--------------------------------------------", stdout );
+  if( snapwr_tile ) fputs( "--------------", stdout );
+  fputs( "[ct],[ld],[dc],[in]--------[ct],[ld],[dc],[in]", stdout );
+  if( snapwr_tile ) fputs( ",[wr]" , stdout );
```
amass-jump, line 211: WDYT about similar changes in `watch`?

amass-jump: Are the list of dummy tiles in fd_tile_unit_test.c important? (I was just looking at files I changed when adding the snapld tile)

src/discof/restore/fd_snapwr_tile.c
```diff
@@ -0,0 +1,276 @@
...
+                     ulong         meta_ctl,
+                     ulong         meta_sig ) {
+  switch( meta_ctl ) {
```
amass-jump, line 187: does it make sense to FD_TEST that we are in IDLE state when receiving any of these?

src/discof/restore/fd_snapin_tile.c
```diff
@@ -872,25 +682,30 @@ populate_allowed_fds( fd_topo_t      const * topo FD_PARAM_UNUSED,
...
-populate_allowed_seccomp( fd_topo_t const *      topo FD_PARAM_UNUSED,
-                          fd_topo_tile_t const * tile FD_PARAM_UNUSED,
+populate_allowed_seccomp( fd_topo_t const *      topo,
+                          fd_topo_tile_t const * tile,
                           ulong                  out_cnt,
                           struct sock_filter *   out ) {
-
-  populate_sock_filter_policy_fd_snapin_tile( out_cnt, out, (uint)fd_log_private_logfile_fd() );
-  return sock_filter_policy_fd_snapin_tile_instr_cnt;
+  (void)topo;
```
amass-jump, line 689: OOC, are we not supposed to use FD_PARAM_UNUSED?

src/discof/restore/fd_snapwr_tile.c
amass-jump: I see, I didn't catch at first that any control frag processing in snapwr will block snapin (thus preventing the control frag from being forwarded to snapct). So this should all be fine 👍

src/discof/restore/fd_snapin_tile.c
ripatel-fd: Kevin prefers `(void)`.

src/discof/restore/fd_snapwr_tile.c
ripatel-fd: I don't think so, snapin->snapwr is simplified. snapin just sends a SHUTDOWN while snapin is still in PROCESSING state, because snapwr just blindly writes to offsets specified by snapin. It isn't aware what kind of snapshot is being written and whether anything was reset.

src/app/firedancer-dev/commands/snapshot_load.c
ripatel-fd: Not that I'm aware of since fd_tile_unit_test.c is probably not supposed to test the snapshot loader

PR #7024, opened 2025-11-02
txncache: fix short copy

yufeng-jump:
A short copy would leak txnpages.  Moreover, when more than one txnpage is copied at once, copying uints into ushorts would cause txnpage 0 to be considered free when it it not, as the upper ushort of a uint would be all zeroes in the case of a valid txnpage index.

### DIFF
--- a/src/flamenco/runtime/fd_txncache.c
+++ b/src/flamenco/runtime/fd_txncache.c
@@ -8,7 +8,7 @@ struct blockcache {
   uint * heads;          /* The hash table for the blockhash.  Each entry is a pointer to the head of a linked list of
                             transactions that reference this blockhash.  As we add transactions to the bucket, the head
                             pointer is updated to the new item, and the new item is pointed to the previous head. */
-  uint * pages;          /* A list of the txnpages containing the transactions for this blockcache. */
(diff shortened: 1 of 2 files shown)


src/flamenco/runtime/fd_txncache.c
```diff
@@ -304,7 +304,10 @@ fd_txncache_finalize_fork( fd_txncache_t *       tc,
 static inline void
 remove_blockcache( fd_txncache_t * tc,
                    blockcache_t *  blockcache ) {
-  memcpy( tc->txnpages_free+tc->shmem->txnpages_free_cnt, blockcache->pages, blockcache->shmem->pages_cnt*sizeof(ushort) );
+  for( ushort i=0; i<blockcache->shmem->pages_cnt; i++ ) {
+    FD_TEST( blockcache->pages[ i ]<USHORT_MAX );
+    tc->txnpages_free[ tc->shmem->txnpages_free_cnt+i ] = (ushort)blockcache->pages[ i ];
```
mmcgee-jump: Wouldn't it be cleaner to make `_blockcache_pages` ushort and keep the memcpy?

PR #7025, opened 2025-11-03
snapshots: implement gossip allow_list / block_list

amass-jump:


src/discof/restore/fd_snapct_tile.c
```diff
@@ -420,6 +427,39 @@ init_load( fd_snapct_tile_t *  ctx,
...
 }
 
+static void
+log_download( fd_snapct_tile_t * ctx,
+              int                full,
+              fd_ip4_port_t      addr,
+              ulong              slot ) {
+  for( gossip_ci_map_iter_t iter = gossip_ci_map_iter_init( ctx->gossip.ci_map, ctx->gossip.ci_table );
+      !gossip_ci_map_iter_done( iter, ctx->gossip.ci_map, ctx->gossip.ci_table );
+      iter = gossip_ci_map_iter_next( iter, ctx->gossip.ci_map, ctx->gossip.ci_table ) ) {
```
cali-jumptrading, line 437: Could we avoid this iteration by keeping the gossip peer address associated with the rpc address as metadata in snapct? I had to keep metadata for each peer when implementing https so it might be something we end up doing later.

src/discof/restore/fd_snapct_tile.c
```diff
@@ -807,50 +849,89 @@ gossip_frag( fd_snapct_tile_t *  ctx,
...
-        entry->pubkey = *(fd_pubkey_t const *)msg->origin_pubkey;
-        entry->rpc_addr.l = 0UL;
-        entry->wallclock_nanos = msg->wallclock_nanos;
-        gossip_ci_map_idx_insert( ctx->gossip.ci_map, msg->contact_info.idx, ctx->gossip.ci_table );
+        entry->pubkey      = *pubkey;
+        entry->rpc_addr.l  = 0UL;
+        entry->added_nanos = fd_log_wallclock();
+        if( ctx->config.sources.gossip.allow_any ) {
+          entry->allowed = 1;
+          for( ulong i=0UL; i<ctx->config.sources.gossip.block_list_cnt; i++ ) {
```
cali-jumptrading, line 866: perhaps use a hash set here to avoid iteration? so we can test if a pubkey is in the block_list/allow_list. Probably not a huge performance impact tho

src/discof/restore/fd_snapct_tile.c
```diff
@@ -807,50 +849,89 @@ gossip_frag( fd_snapct_tile_t *  ctx,
...
         entry->rpc_addr = new_addr;
         if( FD_LIKELY( !!cur_addr.l ) ) {
           int removed = fd_ssping_remove( ctx->ssping, cur_addr );
           if( FD_LIKELY( removed ) ) fd_sspeer_selector_remove( ctx->selector, cur_addr );
         }
         if( FD_LIKELY( !!new_addr.l ) ) fd_ssping_add( ctx->ssping, new_addr );
+        if( !ctx->config.sources.gossip.allow_any ) {
+          FD_BASE58_ENCODE_32_BYTES( pubkey->uc, pubkey_b58 );
+          if( FD_LIKELY( !!new_addr.l ) ) {
+            FD_LOG_NOTICE(( "allowed gossip peer added with public key `%s` and RPC address `" FD_IP4_ADDR_FMT ":%hu`",
```
cali-jumptrading, line 898: is this for debugging? perhaps bump this down to FD_LOG_DEBUG/FD_LOG_INFO before merging if these logs can spam.

src/discof/restore/fd_snapct_tile.c
```diff
@@ -807,50 +849,89 @@ gossip_frag( fd_snapct_tile_t *  ctx,
...
+                            pubkey_b58, FD_IP4_ADDR_FMT_ARGS( new_addr.addr ), fd_ushort_bswap( new_addr.port ) ));
+          } else {
+            FD_LOG_WARNING(( "allowed gossip peer with public key `%s` does not advertise an RPC address", pubkey_b58 ));
```
cali-jumptrading, line 901: same here

src/discof/restore/fd_snapct_tile.c
```diff
@@ -807,50 +849,89 @@ gossip_frag( fd_snapct_tile_t *  ctx,
...
+      if( rem_idx==ULONG_MAX ) break;
+      FD_TEST( entry->allowed && rem_idx==msg->contact_info.idx );
       fd_ip4_port_t addr = entry->rpc_addr;
       if( FD_LIKELY( !!addr.l ) ) {
         int removed = fd_ssping_remove( ctx->ssping, addr );
         if( FD_LIKELY( removed ) ) fd_sspeer_selector_remove( ctx->selector, addr );
       }
+      if( !ctx->config.sources.gossip.allow_any ) {
+        FD_BASE58_ENCODE_32_BYTES( pubkey->uc, pubkey_b58 );
+        FD_LOG_WARNING(( "allowed gossip peer removed with public key `%s` and RPC address `" FD_IP4_ADDR_FMT ":%hu`",
```
cali-jumptrading, line 924: also same here

src/discof/restore/fd_snapct_tile.c
amass-jump: We could, but I'm not a fan of adding redundant state when a cheap (and in this case, O(1) time) computation would suffice. Maintaining that extra state introduces more opportunities for bugs

src/discof/restore/fd_snapct_tile.c
amass-jump: These are going to be short arrays, generally linear search performs best for cases like this despite what the CS textbooks would have us believe

src/discof/restore/fd_snapct_tile.c
amass-jump: No, I intended for this to be end-user logging. It is not logged in allow_any mode, as that would spam, so we only get as many logs as the user has allowed peers configured.

src/discof/restore/fd_snapct_tile.c
amass-jump: This one is even more important for the operators. Basically it is saying that a allowed peer they have configured is not able to be used, so they should probably review the config

src/discof/restore/fd_snapct_tile.c
cali-jumptrading: how so? You're going to be searching up to 32 pubkeys, with each pubkey doing a 32 byte memcmp.

src/discof/restore/fd_snapct_tile.c
cali-jumptrading: how is this an O(1) time computation? If the user has enabled `allow_any`, won't this `log_download` function have to iterate through all gossip peers, which could be up to 2000?

src/discof/restore/fd_snapct_tile.c
cali-jumptrading: Are you sure these logs don''t spam? As I recall, gossip ci messages are sent on an interval.

src/discof/restore/fd_snapct_tile.c
amass-jump: Sorry, what I meant by `time` was we only run this twice per run of the validator (once for full and once for incr, maybe a few extra times if a download fails).

Yes, it would iterate through the full list. But we are already doing that often anyway (for the gossip saturation)

src/discof/restore/fd_snapct_tile.c
amass-jump: These are only logged once per peer, on the first gossip message we receive for that peer (we could get another log if that peer changes their RPC address, which is not a normal thing to happen)

src/discof/restore/fd_snapct_tile.c
cali-jumptrading: gossip saturation only occurs when we are collecting peers though... it's not long running in the background. Anyways, adding the metadata here probably will have a negligible perf impact anyways, so it's fine to keep as an iteration for now

src/discof/restore/fd_snapct_tile.c
mmcgee-jump: I would confirm this, peers republish every 5s with a new timestamp, but I'm not sure if gossip rebroadcasts that as an "update" or ignores it.

src/discof/restore/fd_snapct_tile.c
amass-jump: I think if we ran this allow_list / block_list logic more than once per peer, it might make sense to introduce a hash set or whatnot. But given it's at most once per peer I don't think the tradeoff to add the extra state is there

src/discof/restore/fd_snapct_tile.c
amass-jump: Internally speaking, there is one gossip message for new/update and one for remove. We handle the "upsert"-style logic for the former and only do this logging when we see a new peer (not on all the additional messages for that peer)

amass-jump: More specifically, this logic only runs when the RPC address of a new or existing peer changes

```
 885       /* Maybe update the RPC address of a new or existing allowed gossip peer */
 886       fd_ip4_port_t cur_addr = entry->rpc_addr;
 887       fd_ip4_port_t new_addr = msg->contact_info.contact_info->sockets[ FD_CONTACT_INFO_SOCKET_RPC ];
 888       if( FD_UNLIKELY( new_addr.l!=cur_addr.l ) ) {
``

src/discof/restore/fd_snapct_tile.c
cali-jumptrading: hmm... if you did add the extra state, that could save you a significant number of memcmps though (1 + 2 + 3 ... + 32) to be exact. And we should try to process gossip frags as quickly as possible to not backpressure gossip...

wait no, you'd actually save up to 2000 memcmps, one for each peer.

amass-jump: We benchmarked this and using the basic linear search was faster than using a hash map

PR #7027, opened 2025-11-03
build: arm64 brew support, install coreutils on macos

es3n1n-osec:
Hi, Homebrew on arm64 macos uses a different folder for Cellars. I also think that the Cellar location can be manually changed as well, so the current check would fail for these builds too. A better way to check whether a formula is installed is to directly ask Homebrew. 

Plus, `nproc` is not a thing by default on macos. It should either be installed with `coreutils`, or the usage of it should be swapped to a `sysctl -n hw.logicalcpu`. Given that it probably will be easier to maintain everything with keeping in mind that you can freely use `nproc`, the current change is just installing `coreutils`.

### DIFF
--- a/deps.sh
+++ b/deps.sh
@@ -266,13 +266,13 @@ check_alpine_pkgs () {
 }
 
 check_macos_pkgs () {
-  local REQUIRED_FORMULAE=( perl autoconf gettext automake flex bison protobuf )
+  local REQUIRED_FORMULAE=( perl autoconf gettext automake flex bison protobuf coreutils )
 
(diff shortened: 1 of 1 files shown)


ripatel-fd:
Thanks for contribution. I'm fine with merging this as-is. It'd be nicer to remove the coreutils dependency and add an nprocs wrapper to deps.sh. Since, afaik, we're not using the nproc shell command anywhere else in the build system.

ripatel-fd:
On the other hand, keeping coreutils makes sense because we depend on a bunch of GNU-specific extensions for many util commands.

PR #7029, opened 2025-11-03
SIMD-0334: Fix alt_bn128_pairing syscall length check

LStan:
anza-xyz/solana-sdk#324
anza-xyz/agave#8747


cmoyes-jump (approved):

LStan:
Implemented in #7217

PR #7037, opened 2025-11-03
gui: publish protocol limits

jherrera-jump:


src/disco/pack/fd_pack.c
```diff
@@ -2613,6 +2632,23 @@ fd_pack_rebate_cus( fd_pack_t              * pack,
     in_wcost_table->total_cost -= rebate->writer_rebates[i].rebate_cus;
     /* Important: Even if this is 0, don't delete it from the table so
        that the insert order doesn't get messed up. */
+
+    /* Update the write account heap. */
+    int in_heap = 0;
```
ptaffet-jump: Don't you also need to do something similar when the transaction is scheduled? Though I suppose that is then a little harder because you have the possibility of a pubkey going on to the list and then off the list when the rebate is processed. If you're going to only handle the simple case, then you need to specify the exact semantics in the doc comment.

src/disco/pack/fd_pack.c
```diff
@@ -2613,6 +2632,23 @@ fd_pack_rebate_cus( fd_pack_t              * pack,
...
+    for( ulong i = 0UL; i < FD_PACK_TOP_WRITERS_CNT; i++ ) {
```
ptaffet-jump: I think it's true that if the `in_wcost_table->total_cost` is less than the last element in the heap you don't need to even consider inserting it, right? I'd add that check

PR #7038, opened 2025-11-03
txncache: add txn generation

mmcgee-jump:


src/flamenco/runtime/fd_txncache.c
```diff
@@ -205,6 +205,7 @@ fd_txncache_insert_txn( fd_txncache_t *         tc,
     ulong txnhash_offset = blockcache->shmem->txnhash_offset;
     memcpy( txnpage->txns[ txn_idx ]->txnhash, txnhash+txnhash_offset, 20UL );
     txnpage->txns[ txn_idx ]->fork_id = fork_id;
+    txnpage->txns[ txn_idx ]->generation = blockcache->shmem->generation;
```
yufeng-jump: I believe we need the generation of the block where the transaction landed, not the referenced blockhash.  The referenced blockhash is still well alive and has the same generation through the whole sequence of events, whereas the minority fork that landed the transaction is the one that got recycled.

src/flamenco/runtime/fd_txncache.c
```diff
@@ -441,7 +457,7 @@ fd_txncache_query( fd_txncache_t *       tc,
   for( uint head=blockcache->heads[ head_hash ]; head!=UINT_MAX; head=tc->txnpages[ head/FD_TXNCACHE_TXNS_PER_PAGE ].txns[ head%FD_TXNCACHE_TXNS_PER_PAGE ]->blockcache_next ) {
     fd_txncache_single_txn_t * txn = tc->txnpages[ head/FD_TXNCACHE_TXNS_PER_PAGE ].txns[ head%FD_TXNCACHE_TXNS_PER_PAGE ];
 
-    int descends = txn->fork_id.val==fork_id.val || descends_set_test( fork->descends, txn->fork_id.val );
+    int descends = (txn->fork_id.val==fork_id.val || descends_set_test( fork->descends, txn->fork_id.val )) && blockcache->shmem->generation==txn->generation;
```
yufeng-jump: Same here, you wanna check against the generation of the pack/replay fork, not the generation of the referenced blockhash?

PR #7040, opened 2025-11-03
reasm: fix for uaf in out deque

ibhatt-jumptrading:


src/discof/reasm/fd_reasm.c
```diff
@@ -429,6 +439,20 @@ fd_reasm_publish( fd_reasm_t * reasm, fd_hash_t const * merkle_root ) {
...
+       relatively expensive check since it iterates over the entire out
+       queue, however it does not happen in the common case. */
+    if( FD_UNLIKELY( head->in_out_queue ) ) {
+      for( out_iter_t iter = out_iter_init( reasm->out ); !out_iter_done( reasm->out, iter ); iter = out_iter_next( reasm->out, iter ) ) {
+        ulong * ele = out_iter_ele( reasm->out, iter );
+        if( *ele==pool_idx( reasm->pool, head ) ) {
+          *ele = INVALID_OUT_IDX;
+          break;
+        }
+      }
```
lidatong: hm so this is O(n^2) which is not great and also i dont think you need to introduce another idea of invalid out idx

what i was thinking is you instead have a `.free` flag that indicates whether every element is free or not

if free = 1 it's in the pool, otherwise not

on acquire and release you set / unset `.free`

and then in publish, at the very end, you `while pop deque ... push if free = 0 otherwise dont` (tracking when youve wrapped around)

PR #7042, opened 2025-11-04
banks: fix pool race

topointon-jump:


ibhatt-jumptrading:
this doesn't cause a race in the client since this code is only used in fuzzing/testing

topointon-jump:
Closing in favour of https://github.com/firedancer-io/firedancer/pull/7046

PR #7046, opened 2025-11-04
banks: acquire locks before checking cow dirty bit

topointon-jump:


ibhatt-jumptrading:
we only need to guard the dirty bit with a bank specific lock while we are executing. Like yufeng said, that handling could perhaps be bugged

topointon-jump:
If one thread is advancing the root, but another thread is also calling locking_modify on the new root, then there is a race condition where the old root pool element could be lost. The removal of the pool element + updating of the dirty bit in advance_root needs to be atomic w.r.t any calls to locking_modify on the new root.

Updated the PR so that any change to the dirty bit is protected by the lock on the new roots element, which will ensure atomicity w.r.t locking_modify.

PR #7051, opened 2025-11-04
gui: terminate strings copied from block engine tile

two-heart:
otherwise we can have inter-tile memory corruptions


src/disco/gui/fd_gui.c
```diff
@@ -156,6 +156,10 @@ fd_gui_new( void *                shmem,
 
 
   gui->block_engine.has_block_engine = 0;
+  /* Guarantee that copied strings are always NULL-terminated */
```
mmcgee-jump: Does the bundle tile make this guarantee? Can we just FD_TEST the termination?

PR #7056, opened 2025-11-04
codeql: add query to check for wrong pool ele_acquire null check

intrigus-lgtm:
pool ele_acquire can never return NULL (even when the pool is empty) so checking for that is likely to be a misunderstanding or superflous.


two-heart (dismissed):
lgtm

PR #7057, opened 2025-11-04
chore: fix wrong fd_pool acquire usage

intrigus-lgtm:
calling acquire on a pool with no space left is UB and cannot be assumed to return NULL
Fixes all the results that https://github.com/firedancer-io/firedancer/pull/7056 finds.

### DIFF
--- a/src/flamenco/runtime/fd_bank.c
+++ b/src/flamenco/runtime/fd_bank.c
@@ -53,9 +53,6 @@ fd_bank_footprint( void ) {
       FD_LOG_CRIT(( "Failed to acquire " #name " pool element: pool is full" ));                                   \
     }                                                                                                              \
     fd_bank_##name##_t * child_##name = fd_bank_##name##_pool_ele_acquire( name##_pool );                          \
-    if( FD_UNLIKELY( !child_##name ) ) {                                                                           \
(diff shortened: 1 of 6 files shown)


ripatel-fd (dismissed):
Can we just introduce an acquire_safe API here instead of removing error checks?

intrigus-lgtm:
> Can we just introduce an acquire_safe API here instead of removing error checks?

I don't fully understand what you are proposing to do.
This PR is basically just a continuation of https://github.com/firedancer-io/firedancer/pull/6739
All those error checks are redundant (as far as I understand the code) and will therefore NEVER trigger.

src/flamenco/runtime/fd_bank.c
```diff
@@ -53,9 +53,6 @@ fd_bank_footprint( void ) {
     }                                                                                                              \
     fd_rwlock_write( fd_bank_get_##name##_pool_lock( bank ) );                                                     \
```
two-heart: Unrelated to your correct fix, this lock should be acquired before the pool acquire, for the pool acquire to be correct

intrigus-lgtm: Good catch, but this had already been fixed in https://github.com/firedancer-io/firedancer/pull/7032

PR #7061, opened 2025-11-04
firedancer: bank correctly report error codes

ibhatt-jumptrading:


src/discof/bank/fd_bank_tile.c
```diff
@@ -185,11 +189,6 @@ handle_microblock( fd_bank_ctx_t *     ctx,
     FD_TEST( txn_ctx->flags & FD_TXN_P_FLAGS_EXECUTE_SUCCESS );
     txn->flags |= FD_TXN_P_FLAGS_EXECUTE_SUCCESS | FD_TXN_P_FLAGS_SANITIZE_SUCCESS;
 
-    /* Stash the result in the flags value so that pack can inspect it. */
-    /* TODO: Need to translate the err to a hacky Frankendancer style err
```
mmcgee-jump, line 189: What happened to the todo? Doesn't this still need to be done?

ibhatt-jumptrading: firedancer error codes already map to agave error codes

PR #7064, opened 2025-11-04
snapshots: lthash v3

cali-jumptrading:


src/disco/stem/fd_stem.c
```diff
@@ -356,6 +356,8 @@ STEM_(run1)( ulong                        in_cnt,
     cr_max = fd_ulong_min( cr_max, out_depth[ cons_out[ cons_idx ] ] );
   }
 
+  if( FD_UNLIKELY( burst>cr_max ) ) FD_LOG_ERR(( "one or more out links have insufficient depth for STEM_BURST %lu", burst ));
```
cali-jumptrading: This is to ensure we don't allow stem to start up if `STEM_BURST` is greater than the maximum available credits across all out links. If this scenario occurs, stem will never be able to process input frags because it will always think it is backpressured.

src/app/shared/fd_action.h
```diff
@@ -5,7 +5,7 @@
 
 union fdctl_args {
   struct {
-    char  tile_name[ 7UL ];
+    char  tile_name[ 8UL ];
```
cali-jumptrading: Needed to increase the tile name size so we can store names like "snaplta".

ripatel-fd: We'd have to shift the print offsets in all the tools, like monitor, mem, etc. Can we just rename these to `lthash+` and `lthash-`?

ripatel-fd: Or `snaph+` and `snaph-`

cali-jumptrading: ohh that might work.

cali-jumptrading: could do `snapha` and `snaphs`? nah that's kinda confusing... `snaphs` could easily refer to snaphash

cali-jumptrading: changed the names to `snapla` and `snapls`... let me know if you don't like these names

src/disco/metrics/generated/fd_metrics_snaplta.c
```diff
@@ -0,0 +1,8 @@
+/* THIS FILE IS GENERATED BY gen_metrics.py. DO NOT HAND EDIT. */
```
amass-jump: Don't think you want these files in here... we should probably fix the generator script to delete files that shouldn't be there

src/disco/stem/fd_stem.c
amass-jump: Nice! One nit is adding logging for `cr_max` too?

src/app/shared/fd_config_parse.c
```diff
@@ -83,6 +83,7 @@ fd_config_extract_podf( uchar *        pod,
   CFG_POP      ( uint,   layout.exec_tile_count                              );
   CFG_POP      ( uint,   layout.sign_tile_count                              );
   CFG_POP      ( uint,   layout.gossvf_tile_count                            );
+  CFG_POP      ( uint,   layout.snapla_tile_count                           );
```
amass-jump:
```suggestion
  CFG_POP      ( uint,   layout.snapla_tile_count                            );
```

src/flamenco/runtime/fd_hashes.h
```diff
@@ -56,6 +56,29 @@ fd_hashes_account_lthash( fd_pubkey_t const       * pubkey,
...
+   to the account data.  data_len is the length of the account data.
+   lthash_out points to where the computed lthash value will be written
+   (2048 bytes).
+
+   On return, lthash_out contains the computed lthash.  This function
+   assumes all pointers are valid and properly aligned.  The account
+   data pointer must be readable for account->dlen bytes. */
+void
+fd_hashes_account_lthash_simple( uchar const         pubkey[ static FD_HASH_FOOTPRINT ],
+                                 uchar const         owner[ static FD_PUBKEY_FOOTPRINT ],
```
amass-jump: I am curious why pubkey has HASH footpoint while owner has PUBKEY footprint

src/flamenco/runtime/fd_hashes.h
```diff
@@ -56,6 +56,29 @@ fd_hashes_account_lthash( fd_pubkey_t const       * pubkey,
...
+   pubkey points to the account's public key (32 bytes).  owner points
+   to the account's owner (32 bytes).  lamports is the account's
+   lamports.  executable is the account's executable flag.  data points
+   to the account data.  data_len is the length of the account data.
+   lthash_out points to where the computed lthash value will be written
+   (2048 bytes).
+
+   On return, lthash_out contains the computed lthash.  This function
+   assumes all pointers are valid and properly aligned.  The account
+   data pointer must be readable for account->dlen bytes. */
```
amass-jump: account->dlen?

src/disco/metrics/metrics.xml
```diff
@@ -1059,6 +1059,18 @@ metric introduced.
     <gauge name="VinylBytesWritten" summary="Number of bytes written so far to the vinyl snapshot file. Might decrease if snapshot creation is aborted and restarted" />
 </tile>
 
+<tile name="snapla">
+    <gauge name="State" summary="State of the tile. 0=hashing, 1=done, 2=shutdown" />
```
amass-jump: Is the summary correct? it seems like snapla and snapls still use the same state enum as all the other pipeline tiles

src/app/shared/commands/watch/watch.c
```diff
@@ -268,22 +268,27 @@ write_snapshots( config_t const * config,
...
-  double snapld_idle_pct = 100.0*(double)diff_tile( config, "snapld", prev_tile, cur_tile, MIDX( COUNTER, TILE, REGIME_DURATION_NANOS_CAUGHT_UP_POSTFRAG ) )/(double)snapld_total_ticks;
-  double snapdc_idle_pct = 100.0*(double)diff_tile( config, "snapdc", prev_tile, cur_tile, MIDX( COUNTER, TILE, REGIME_DURATION_NANOS_CAUGHT_UP_POSTFRAG ) )/(double)snapdc_total_ticks;
-  double snapin_idle_pct = 100.0*(double)diff_tile( config, "snapin", prev_tile, cur_tile, MIDX( COUNTER, TILE, REGIME_DURATION_NANOS_CAUGHT_UP_POSTFRAG ) )/(double)snapin_total_ticks;
-
-  PRINT( "⚡ \033[1m\033[93mSNAPSHOTS...\033[0m\033[22m \033[1mSTATE\033[22m %s \033[1mPCT\033[22m %.1f %% \033[1mRX\033[22m %3.f MB/s \033[1mACC\033[22m %3.1f M/s \033[1mBACKP\033[22m %3.0f%%,%3.0f%%,%3.0f%%,%3.0f%% \033[1mBUSY\033[22m %3.0f%%,%3.0f%%,%3.0f%%,%3.0f%%\033[K\n",
+  ulong snapls_tile_idx   = fd_topo_find_tile( &config->topo, "snapls", 0UL );
+  ulong snapls_total_ticks = snapls_tile_idx!=ULONG_MAX ? total_regime( &cur_tile[ fd_topo_find_tile( &config->topo, "snapls", 0UL )*FD_METRICS_TOTAL_SZ ] )-total_regime( &prev_tile[ fd_topo_find_tile( &config->topo, "snapls", 0UL )*FD_METRICS_TOTAL_SZ ] )  : 0UL;
```
amass-jump:
```suggestion
  ulong snapls_total_ticks = snapls_tile_idx!=ULONG_MAX ? total_regime( &cur_tile[ snapls_tile_idx*FD_METRICS_TOTAL_SZ ] )-total_regime( &prev_tile[ snapls_tile_idx*FD_METRICS_TOTAL_SZ ] )  : 0UL;
```

?

src/app/shared/commands/watch/watch.c
```diff
@@ -268,22 +268,27 @@ write_snapshots( config_t const * config,
...
+  snapct_total_ticks  = fd_ulong_max( snapct_total_ticks, 1UL );
+  snapld_total_ticks  = fd_ulong_max( snapld_total_ticks, 1UL );
+  snapdc_total_ticks  = fd_ulong_max( snapdc_total_ticks, 1UL );
+  snapin_total_ticks  = fd_ulong_max( snapin_total_ticks, 1UL );
+  snapls_total_ticks = fd_ulong_max( snapls_total_ticks, 1UL );
+
+  double snapct_backp_pct  = 100.0*(double)diff_tile( config, "snapct", prev_tile, cur_tile, MIDX( COUNTER, TILE, REGIME_DURATION_NANOS_BACKPRESSURE_PREFRAG ) )/(double)snapct_total_ticks;
```
amass-jump: Can you please fix the indentation in this file? I think the diff is much more complex than it needs to be due to these changes?

src/app/shared/commands/watch/watch.c
```diff
@@ -292,10 +297,12 @@ write_snapshots( config_t const * config,
     snapld_backp_pct,
     snapdc_backp_pct,
     snapin_backp_pct,
+    snapls_backp_pct,
```
amass-jump: I don't think it's great that the watch output is completely ignoring the `snapla` tiles, especially given that this information would be useful for the operator to tune the tile count (if they are backpressuring) 

Perhaps we can combine them somehow? Taking the max for backpressure would be reasonable for example, similarly with min for idle.

What do you think?

src/discof/restore/fd_snapct_tile.c
```diff
@@ -1210,13 +1210,13 @@ unprivileged_init( fd_topo_t *      topo,
       ctx->snapld_in_mem = topo->workspaces[ topo->objs[ in_link->dcache_obj_id ].wksp_id ].wksp;
       FD_TEST( !has_snapld_dc );
       has_snapld_dc = 1;
-    } else if( 0==strcmp( in_link->name, "snapin_ct" ) ) {
+    } else if( 0==strcmp( in_link->name, "snapin_ct" ) || 0==strcmp( in_link->name, "snapls_ct" ) ) {
       ctx->in_kind[ i ] = IN_KIND_SNAPIN;
```
amass-jump:
```suggestion
      ctx->in_kind[ i ] = IN_KIND_ACK;
```
?

src/app/firedancer/topology.c
```diff
@@ -389,16 +403,23 @@ fd_topo_initialize( config_t * config ) {
...
     /**/               fd_topob_link( topo, "snapdc_in",    "snapdc_in",    16384UL,                                  USHORT_MAX,                    1UL );
-    /**/               fd_topob_link( topo, "snapin_ct",    "snapin_ct",    128UL,                                    0UL,                           1UL );
-
-    /**/               fd_topob_link( topo, "snapin_manif", "snapin_manif", 2UL,                                      sizeof(fd_snapshot_manifest_t),1UL );
+    if( FD_UNLIKELY( snapshot_lthash_disabled ) ) {
+                       fd_topob_link( topo, "snapin_ct",    "snapin_ct",    128UL,                                    0UL,                           1UL );
+    } else {
+                       fd_topob_link( topo, "snapls_ct",    "snapls_ct",    128UL,                                    0UL,                           1UL );
+    }
+    /**/               fd_topob_link( topo, "snapin_manif", "snapin_manif", 4UL,                                      sizeof(fd_snapshot_manifest_t),1UL );
```
amass-jump, line 429: Why 2 -> 4?

src/disco/metrics/metrics.xml
cali-jumptrading: oh maybe not, it was from  an older branch. will update!

src/flamenco/runtime/fd_hashes.h
cali-jumptrading: I'll change that to data_len.

src/flamenco/runtime/fd_hashes.h
cali-jumptrading: uhhh no idea... but I'll just use FD_HASH_FOOTPRINT for both of them!

src/disco/metrics/generated/fd_metrics_snaplta.c
cali-jumptrading: These files are generated but they do need to be checked in.

cali-jumptrading: oh wait I see, it's because of a rename, I'll remove these files.

src/app/firedancer/topology.c
```diff
@@ -389,16 +403,23 @@ fd_topo_initialize( config_t * config ) {
...
     /**/               fd_topob_link( topo, "snapct_repr",  "snapct_repr",  128UL,                                    0UL,                           1UL )->permit_no_consumers = 1; /* TODO: wire in repair later */
     if( FD_LIKELY( config->tiles.gui.enabled ) ) {
       /**/             fd_topob_link( topo, "snapct_gui",   "snapct_gui",   128UL,                                    sizeof(fd_snapct_update_t),    1UL );
     }
     if( vinyl_enabled ) {
       /**/             fd_topob_link( topo, "snapin_wr",    "snapin_wr",    4UL,                                      16UL<<20,                      1UL );
     }
+    if( FD_LIKELY( !snapshot_lthash_disabled ) ) {
+    FOR(lta_tile_cnt) fd_topob_link( topo,  "snapla_ls",    "snapla_ls",    128UL,                                    sizeof(fd_lthash_value_t),     1UL );
+    /**/              fd_topob_link( topo,  "snapin_ls",    "snapin_ls",    256UL,                                    sizeof(fd_snapshot_full_account_t), 1UL );
```
amass-jump, line 443: This is a ~2.5GB dcache, right? Is there no way to write out the data compactly such that we aren't wasting 9.999MB for small accounts (which are the vast majority?)

src/app/firedancer/topology.c
cali-jumptrading: This is for the `STEM_BURST` value increasing in `fd_snapin_tile` to 4. Recall that STEM_BURST limits the number of frags you can publish and if your out link has a depth less than your STEM_BURST, you won't be able to publish any frags at all.

src/app/shared/commands/watch/watch.c
cali-jumptrading: I think this can be a great follow up PR.

src/app/firedancer/topology.c
cali-jumptrading: Do you have any suggestions that don't break the current dcache model? Otherwise let's leave this for a follow up PR.

src/app/shared/commands/watch/watch.c
cali-jumptrading: Done

src/disco/stem/fd_stem.c
cali-jumptrading: Done

src/app/shared/fd_config_parse.c
cali-jumptrading: Done

src/disco/metrics/generated/fd_metrics_snaplta.c
cali-jumptrading: Done

src/flamenco/runtime/fd_hashes.h
cali-jumptrading: Done

src/flamenco/runtime/fd_hashes.h
cali-jumptrading: Done

src/app/shared/commands/watch/watch.c
cali-jumptrading: Done

src/discof/restore/fd_snapct_tile.c
cali-jumptrading: Done

.github/workflows/benchmark.yml
```diff
@@ -63,7 +63,7 @@ jobs:
         DUMP_DIR=../dump ./src/flamenco/runtime/tests/run_ledger_backtest.sh \
           -l mainnet-368528500-perf \
           -y 5 -m 2000000 -e 368529500 -c 3.0.0 \
-          --log baseline.log --skip-checksum
+          --log baseline.log --skip-checksum -nlt
```
cali-jumptrading: This ledger also contains the wrong lthash. I don't want to block this PR from merging, so I will fix the ledgers after this PR is merged and remove the -nlt flag.

src/app/firedancer/main.c
```diff
@@ -101,7 +103,6 @@ fd_topo_run_tile_t * TILES[] = {
   &fd_tile_dedup,
   &fd_tile_resolv,
   &fd_tile_pack,
-  &fd_tile_bank,
```
amass-jump, line 105: As discussed this seems to be a rebase conflict?

src/discof/restore/fd_snapla_tile.c
```diff
@@ -0,0 +1,437 @@
...
+typedef struct fd_snapla_tile fd_snapla_tile_t;
+
+static inline int
+should_shutdown( fd_snapla_tile_t * ctx ) {
+  return ctx->state==FD_SNAPSHOT_STATE_SHUTDOWN;
+}
+
+static ulong
+scratch_align( void ) {
+  return 128UL;
```
amass-jump: ?

src/discof/restore/fd_snapla_tile.c
```diff
@@ -0,0 +1,437 @@
...
+          ctx->account_hdr.executable = result->account_header.executable;
+          ctx->account_hdr.lamports   = result->account_header.lamports;
+        }
+        ctx->accounts_seen++;
+        break;
+      case FD_SSPARSE_ADVANCE_ACCOUNT_DATA:
+        if( FD_LIKELY( ctx->hash_account ) ) {
+          fd_memcpy( ctx->data + ctx->acc_data_sz, result->account_data.data, result->account_data.data_sz );
+          ctx->acc_data_sz += result->account_data.data_sz;
+        }
```
amass-jump, line 203: Is it not possible to hash the account header/data directly in the dcache without memcpy'ing everything out first?

src/discof/restore/fd_snapls_tile.c
```diff
@@ -0,0 +1,485 @@
...
+      ctx->in.mtu                    = in_link->mtu;
+      ctx->in.pos                    = 0UL;
+      ctx->in_kind[ i ]              = IN_KIND_SNAPIN;
+    } else if( FD_LIKELY( 0==strcmp( in_link->name, "snapla_ls" ) ) ) {
+      ctx->adder_in[ adder_idx ].wksp    = in_wksp->wksp;
+      ctx->adder_in[ adder_idx ].chunk0  = fd_dcache_compact_chunk0( ctx->adder_in[ adder_idx ].wksp, in_link->dcache );
+      ctx->adder_in[ adder_idx ].wmark   = fd_dcache_compact_wmark ( ctx->adder_in[ adder_idx ].wksp, in_link->dcache, in_link->mtu );
+      ctx->adder_in[ adder_idx ].mtu     = in_link->mtu;
+      ctx->in_kind[ i ]                  = IN_KIND_SNAPLA;
+      if( FD_LIKELY( adder_idx==0UL ) ) ctx->adder_in_offset = i;
```
amass-jump, line 399: This assumes all the adder input links are contiguous? Not sure were enforce that anywhere though?

src/discof/restore/fd_snapls_tile.c
```diff
@@ -0,0 +1,485 @@
...
+
+  ctx->metrics.full.accounts_hashed        = 0UL;
+  ctx->metrics.incremental.accounts_hashed = 0UL;
+
+  ctx->state                        = FD_SNAPSHOT_STATE_IDLE;
+  ctx->full                         = 1;
+  ctx->hash_account                 = 0;
+
+  ctx->num_hash_tiles               = fd_topo_tile_name_cnt( topo, "snapla" );
+  ctx->hash_tile_idx                = tile->kind_id;
```
amass-jump: ?

src/discof/restore/fd_snapls_tile.c
```diff
@@ -0,0 +1,485 @@
...
+
+static int
+recv_acks( fd_snapls_tile_t * ctx,
+           ulong               in_idx,
+           ulong               sig ) {
+  if( FD_UNLIKELY( ctx->num_acks==0UL ) ) ctx->ack_sig = sig;
+
+  /* for extremely small snapshots (ie. backtest), there may be race
+     conditions where we may receive control messages out of order
+     before all acks for a single control message are received. */
```
amass-jump: Is this because of the ERR sig? That one is expected to be "asynchronous", i.e. can be raised by any tile in the pipeline at any time. It's safe to forward it downstream at any time as well. The other control signals are generated at a single point (snapct) and only after the pipeline is "flushed" so I'm not sure I understand this OOO condition

src/app/firedancer/topology.c
```diff
@@ -1038,6 +1080,8 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
     tile->snapin.txncache_obj_id = fd_pod_query_ulong( config->topo.props, "txncache", ULONG_MAX );
 
     tile->snapin.use_vinyl = !!config->firedancer.vinyl.enabled;
+    tile->snapin.lthash_disabled = !!config->development.snapshots.disable_lthash_verification;
+    FD_LOG_WARNING(("lthash disabled = %d", tile->snapin.lthash_disabled));
```
amass-jump: should we remove this or only log if disabled?

src/discof/restore/fd_snapin_tile_private.h
```diff
@@ -274,6 +306,69 @@ fd_snapin_read_account( fd_snapin_tile_t *  ctx,
...
+                                  uchar const *      pubkey,
+                                  int                has_data,
+                                  int *              early_exit ) {
+  if( FD_UNLIKELY( ctx->lthash_disabled ) ) return;
+
+  if( FD_LIKELY( has_data ) ) {
+    fd_snapshot_full_account_t * existing_account = fd_chunk_to_laddr( ctx->hash_out.wksp, ctx->hash_out.chunk );
+    fd_snapshot_account_hdr_init( &existing_account->hdr, pubkey, owner, lamports, executable, data_len );
+    fd_memcpy( existing_account->data, data, data_len );
+    fd_stem_publish( ctx->stem, ctx->out_ct_idx, FD_SNAPSHOT_HASH_MSG_SUB, ctx->hash_out.chunk, sizeof(fd_snapshot_full_account_t), 0UL, 0UL, 0UL );
```
amass-jump: can't this just be sizeof(hdr) + data_len? Do we need to publish/memcpy the full 10MB every time?

src/discof/restore/fd_snapin_tile_funk.c
```diff
@@ -3,14 +3,34 @@
 #include "fd_snapin_tile_private.h"
 #include "../../flamenco/accdb/fd_accdb_sync.h"
 
-void
+int
 fd_snapin_process_account_header_funk( fd_snapin_tile_t *            ctx,
```
amass-jump, line 7: I think @ripatel-fd should take a look at the changes in this file before we merge

src/discof/restore/fd_snapla_tile.c
cali-jumptrading: This is the slow path outside of batched accounts. It will not be possible to hash the account header/data directly in the dcache unless we have an API that gaurantees those dcache entries will not be overwritten by the producer.

src/app/firedancer/topology.c
cali-jumptrading: oh yeah we should remove it

src/discof/restore/fd_snapls_tile.c
cali-jumptrading: ah this was for debugging, will remove.

src/app/firedancer/main.c
cali-jumptrading: forgot about this one! will fix

src/discof/restore/fd_snapls_tile.c
cali-jumptrading: I think all the adder_in links are instantiated with a `FOR` macro in topology. so they would all be continuous.

src/discof/restore/fd_snapin_tile_private.h
cali-jumptrading: I didn't want to break the dcache model... but I can try just publishing the sizeof(hdr) + data_len.

src/discof/restore/fd_snapls_tile.c
cali-jumptrading: Looking at the snapct code, it seems possible that we could call `init_load`, which sends an `INIT_FULL` message to the snapshot pipeline tiles, and then send an `NEXT/DONE` message due to this: https://github.com/firedancer-io/firedancer/blob/main/src/discof/restore/fd_snapct_tile.c#L1064

Would you rather have `snapct` wait for the `INIT_FULL/INIT_INCR` message to be properly flushed and acked before moving on to reading the snapshot? I think this could be a cleaner solution. I had assumed all control messages needed to be acked before moving on and didn't realize you had a special case for `INIT_FULL/INIT_INCR`

cali-jumptrading: Ok I just added ack logic for snapct for this message type.

src/discof/restore/fd_snapin_tile_private.h
cali-jumptrading: done

src/discof/restore/fd_snapin_tile_funk.c
ripatel-fd: Looks good to me

amass-jump (dismissed):

src/discof/restore/fd_snapin_tile.c
```diff
@@ -827,8 +867,8 @@ unprivileged_init( fd_topo_t *      topo,
 
 /* Control fragments can result in one extra publish to forward the
    message down the pipeline, in addition to the result / malformed
-   message / etc. */
-#define STEM_BURST 2UL
+   message. Can send one duplicate account message as well. */
+#define STEM_BURST 3UL
```
cali-jumptrading, line 871: decreasing this from 4 to 3 seems to have fixed some of the perf issues I was seeing with lthash verification.

src/app/shared/commands/watch/watch.c
```diff
@@ -327,34 +327,57 @@ write_snapshots( config_t const * config,
...
+      snapld_backp_pct,
+      snapdc_backp_pct,
+      snapin_backp_pct,
+      snapls_backp_pct,
+      100.0-snapct_idle_pct-snapct_backp_pct,
+      100.0-snapld_idle_pct-snapld_backp_pct,
+      100.0-snapdc_idle_pct-snapdc_backp_pct,
+      100.0-snapin_idle_pct-snapin_backp_pct,
+      100.0-snapls_idle_pct );
+  } else {
```
cali-jumptrading, line 366: Added a if statement to guard whether we print the tile metrics for the snapls tile

PR #2437, opened 2024-07-16
ethtool: disable generic-receive-offload, which is entirely incompatible with AF_XDP and QUIC

nbridge-jump:
generic-receive-offload causes multiple UDP packet payloads to be merged, and the IP and UDP headers to be rewritten.
QUIC short packets cannot handle this weird behavior.

PR #7071, opened 2025-11-05
ethtool: disable RX-UDP-GRO-FWD instead of GRO

amass-jump:
In #2437 GRO was disabled to solve an issue with AF_XDP and GRO, namely the fact that GRO was creating "superpackets" out of aggregated small UDP packets.  AF_XDP was then delivering these to the socket which was breaking QUIC.

As it turns out, we only need to disable rx-udp-gro-forwarding to fix this, not the entire generic-receive-offload feature. This PR swaps the disable logic for the more specific feature.

Wonderfully, this means we can re-enable the overarching GRO feature, which is also done in this PR. This can have a massive impact on throughput for sockets-based TCP flows, such as the HTTP snapshot download.  An increase from ~400MBps to ~900MBps was observed. :rocket:

### DIFF
--- a/book/guide/initializing.md
+++ b/book/guide/initializing.md
@@ -9,8 +9,7 @@ so Firedancer can run correctly. It does the following:
 * **hyperthreads** Checks hyperthreaded pair for critical CPU cores.
 * **ethtool-channels** Configures the number of channels on the network
 device.
-* **ethtool-offloads** Disable generic-receive-offload (GRO) and GRE
-segmentation offload on the network device.
+* **ethtool-offloads** Modify offload feature flags on the network device.
 * **ethtool-loopback** Disable tx-udp-segmentation on the loopback
 device.
 
@@ -188,7 +187,7 @@ privileges, and cannot be performed with capabilities.
 
 ## ethtool-offloads
 XDP is incompatible with a feature of network devices called
-`generic-receive-offload`. This feature must be disabled for Firedancer
+`rx-udp-gro-forwarding`. This feature must be disabled for Firedancer
 to work. GRE segmentation offload is also disabled.
 
 The command run by the stage is similar to running
--- a/src/app/shared/commands/configure/ethtool-offloads.c
+++ b/src/app/shared/commands/configure/ethtool-offloads.c
@@ -1,12 +1,20 @@
-/* This stage disables various ethtool features on the main and
-   loopback interfaces that are known to be incompatible with AF_XDP.
+/* This stage checks and modifies various ethtool features on the main
+   and loopback interfaces.
 
-     - "Generic Receive Offload": If left enabled, GRO will mangle UDP
-     packets in a way that causes AF_XDP packets to get corrupted.
+     - "Generic Receive Offload": If enabled, may greatly increase
+     throughput for sockets-based TCP flows, such as HTTP snapshot
+     downloads.  This stage will log a warning if GRO is disabled, but
+     does not modify the flag.
+
+     - "RX UDP GRO Forwarding": If left enabled, may aggregate multiple
+     UDP packets into a single large superpacket.  This would normally
+     be split later for socket recv() calls, but AF_XDP delivers the
+     full superpacket which confuses the application layer.  Disabled
+     by this stage.
 
      - "GRE Segmentation Offload": This feature has been known to cause
      corruption of packets sent via normal sockets while XDP is in use
-     on the same system. */
+     on the same system.  Disabled by this stage. */
 
 #include "configure.h"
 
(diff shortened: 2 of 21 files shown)


src/app/shared/commands/configure/ethtool-offloads.c
```diff
@@ -25,16 +31,13 @@ enabled( fd_config_t const * config ) {
      the virtual device as part of netns setup, not here */
   if( config->development.netns.enabled ) return 0;
 
-  /* only enable if network stack is XDP */
-  if( 0!=strcmp( config->net.provider, "xdp" ) ) return 0;
```
amass-jump, line 29: Note this change. We enable GRO for all providers but only disable the other two features in XDP mode.

src/app/shared/commands/configure/ethtool-offloads.c
```diff
@@ -1,8 +1,14 @@
-/* This stage disables various ethtool features on the main and
-   loopback interfaces that are known to be incompatible with AF_XDP.
+/* This stage modifies various ethtool features on the main and
+   loopback interfaces.
 
-     - "Generic Receive Offload": If left enabled, GRO will mangle UDP
-     packets in a way that causes AF_XDP packets to get corrupted.
+     - "Generic Receive Offload": Enabled to increase throughput for
```
mmcgee-jump: Do we want to specifically enable this, rather than just not touching it?

mmcgee-jump (dismissed):

src/app/shared/commands/configure/ethtool-offloads.c
ripatel-fd: It greatly increases snapshot download speeds

ripatel-fd: Well, good point by @mmcgee-jump though. If it's disabled by default, there's probably a good reason for it.

ripatel-fd (dismissed):

src/app/shared/commands/configure/ethtool-offloads.c
amass-jump: I think it makes sense to enable this explicitly, given it has a large throughput impact on sockets traffic and no (known) downside to XDP traffic. It is generally on by default anyway. I added logic in check() to ignore/allow it to be disable if we get an NOTSUPP error

book/guide/initializing.md
```diff
@@ -188,7 +187,7 @@ privileges, and cannot be performed with capabilities.
 
 ## ethtool-offloads
 XDP is incompatible with a feature of network devices called
-`generic-receive-offload`. This feature must be disabled for Firedancer
+`rx-udp-gro-forwarding`. This feature must be disabled for Firedancer
 to work. GRE segmentation offload is also disabled.
```
nbridge-jump, line 191: Might be worth adjusting the language to indicate this only affects XDP in skb mode. Sockets mode should work fine.  And we are sure(?) drv mode works fine

src/app/shared/commands/configure/ethtool-offloads.c
mmcgee-jump: If it's enabled by default, the only reason it would be disabled is if an operator explicitly disabled it for some reason. It feels a bit weird to then change it back, although I don't feel super strongly. We do not go out of our way to reconfigure a other settings that are already the defaults, though.

nbridge-jump: I can see arguments either way. But if we choose to turn it on, we should document that fact

book/guide/initializing.md
amass-jump: I was hoping to avoid making too many statements/assumptions about the behavior in various modes. We could _probably_ only disable these features in XDP `drv` provider mode. But it would not be too surprising that in certain strange edge-case setups / situations `drv` could still be affected?

To keep things simpler I just left it the way it was as disabling in all provider modes. If we find a need to enable this feature (when using `drv`) in the future perhaps we could revisit then

src/app/shared/commands/configure/ethtool-offloads.c
amass-jump: How about I just change it to a warning if disabled, like we do for hyperthreads?

mmcgee-jump: SGTM

ripatel-fd: I think we should just not touch it and leave a note in the Firedancer book instructing operators to manually enable GRO for better TCP throughput. The reason is that it's impossible to know whether enabling GRO is safe (countless possible driver bugs), whereas for the other system changes we know that they'll work.

mmcgee-jump: Strong disagree. We should not let operators run a suboptimal config because there might be speculated bugs in a driver.

nbridge-jump: They should have the option of running with it disabled

mmcgee-jump: Yes, they would just disable it themselves to do that

amass-jump:
Before merge I'm going to add some logic to prevent spamming the same warning several times

nbridge-jump (approved):

PR #7097, opened 2025-11-06
rpc: implement a few more monitoring methods

mmcgee-jump:


src/discof/replay/fd_replay_tile.h
```diff
@@ -26,6 +26,20 @@ struct fd_replay_slot_completed {
 
   ulong transaction_count;
 
+  struct {
```
ibhatt-jumptrading, line 29: we should probably move these to another file where we define the runtime types and break the fd_types dependency. Don't need to do in this pr, but maybe worth leaving a todo

src/discof/replay/fd_replay_tile.c
```diff
@@ -1387,6 +1402,8 @@ publish_reset( fd_replay_tile_t *  ctx,
   }
   reset->next_leader_slot = ctx->next_leader_slot;
 
+  if( FD_LIKELY( ctx->rpc_enabled ) ) bank->refcnt++;
```
ibhatt-jumptrading, line 1405: where do we release the rpc refcnt

ibhatt-jumptrading: doesn't seem like we use it in rpc tile

mmcgee-jump: We return a frag to replay to release it

PR #7100, opened 2025-11-06
restore: narrow account batching to 8-wide

ripatel-fd:
Improves snapshot loading throughput by nearly 2x when disabling
bstream integrity hashes, on Zen 4.  It is unclear why 16-wide
batching was that slow.  Presumably due to more frequent batching
failures, bigger code footprint, and more frequent data cache
misses.

### DIFF
--- a/src/discof/restore/utils/fd_ssparse.h
+++ b/src/discof/restore/utils/fd_ssparse.h
@@ -59,7 +59,7 @@ typedef struct acc_vec acc_vec_t;
 
 /* FD_SSPARSE_ACC_BATCH_MAX controls the max number of accounts in a
    batch. */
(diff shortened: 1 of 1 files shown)


amass-jump (approved):

ripatel-fd:
```
(bstream pair hashing and block trailing memset disabled for this test)

Clang 19,    8-wide: snapin CPU  69% (bottlenecked by snapdc) 
GCC   11.5,  8-wide: snapin CPU  69% (bottlenecked by snapdc) 
Clang 19,   16-wide: snapin CPU 100%, acc=7.6M/s
GCC   11.5, 16-wide: snapin CPU 100%, acc=7.6M/s
```

PR #7104, opened 2025-11-06
flamenco, types: enforce upper bound deque sizes

ravyu-jump:


ravyu-jump:
Closed because Agave doesn't enforce these + such OOBs are unreachable for the relevant types (vote_state)

PR #7105, opened 2025-11-06
snapshots: snapld uses nonblocking socket again after GRO enabled

amass-jump:
Also a few fixes / cleaning in fd_sshttp

### DIFF
--- a/src/discof/restore/fd_snapld_tile.c
+++ b/src/discof/restore/fd_snapld_tile.c
@@ -169,7 +169,10 @@ after_credit( fd_snapld_tile_t *  ctx,
               fd_stem_context_t * stem,
               int *               opt_poll_in FD_PARAM_UNUSED,
               int *               charge_busy ) {
-  if( ctx->state!=FD_SNAPSHOT_STATE_PROCESSING ) return;
+  if( ctx->state!=FD_SNAPSHOT_STATE_PROCESSING ) {
+    fd_log_sleep( (long)1e6 );
+    return;
+  }
 
   uchar * out = fd_chunk_to_laddr( ctx->out_dc.mem, ctx->out_dc.chunk );
 
@@ -204,12 +207,9 @@ after_credit( fd_snapld_tile_t *  ctx,
           fd_ssctrl_meta_t * meta = (fd_ssctrl_meta_t *)out;
           ulong next_chunk = fd_dcache_compact_next( ctx->out_dc.chunk, sizeof(fd_ssctrl_meta_t), ctx->out_dc.chunk0, ctx->out_dc.wmark );
           memmove( fd_chunk_to_laddr( ctx->out_dc.mem, next_chunk ), out, data_len );
-          char const * full_name;
-          char const * incr_name;
-          fd_sshttp_snapshot_names( ctx->sshttp, &full_name, &incr_name );
           meta->total_sz = fd_sshttp_content_len( ctx->sshttp );
           FD_TEST( meta->total_sz!=ULONG_MAX );
-          fd_memcpy( meta->name, ctx->load_full ? full_name : incr_name, PATH_MAX );
+          fd_memcpy( meta->name, fd_sshttp_snapshot_name( ctx->sshttp ), PATH_MAX );
           fd_stem_publish( stem, 0UL, FD_SNAPSHOT_MSG_META, ctx->out_dc.chunk, sizeof(fd_ssctrl_meta_t), 0UL, 0UL, 0UL );
           ctx->out_dc.chunk = next_chunk;
         }
--- a/src/discof/restore/utils/fd_sshttp.c
+++ b/src/discof/restore/utils/fd_sshttp.c
@@ -8,21 +8,25 @@
 
 #include <unistd.h>
 #include <errno.h>
+#include <poll.h>
 #include <stdlib.h>
 
 #include <sys/socket.h>
-#include <netinet/tcp.h>
 #include <netinet/in.h>
+
+#define FD_SSHTTP_MAGIC (0xF17EDA2CE5811900) /* FIREDANCE HTTP V0 */
 
 #define FD_SSHTTP_STATE_INIT  (0) /* start */
 #define FD_SSHTTP_STATE_REQ   (1) /* sending request */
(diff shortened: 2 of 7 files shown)


src/discof/restore/utils/fd_sshttp.c
```diff
@@ -125,29 +126,23 @@ fd_sshttp_init( fd_sshttp_t * http,
...
     .sin_family = AF_INET,
     .sin_port   = addr.port,
     .sin_addr   = { .s_addr = addr.addr }
   };
 
-  if( FD_UNLIKELY( -1==connect( http->sockfd, fd_type_pun_const( &addr_in ), sizeof(addr_in) ) ) ) {
+  if( FD_LIKELY( -1==connect( http->sockfd, fd_type_pun_const( &addr_in ), sizeof(addr_in) ) ) ) {
     if( FD_UNLIKELY( errno!=EINPROGRESS ) ) {
       if( FD_UNLIKELY( -1==close( http->sockfd ) ) ) FD_LOG_ERR(( "close() failed (%d-%s)", errno, fd_io_strerror( errno ) ));
-      return;
```
amass-jump, line 150: This `return` caused issues because we stay in the INIT state, which means every future call to advance() would return EAGAIN and the caller would never be able to recover / reset. Instead we "fall through" to the send request state where we will be able to return ERROR to the caller for a reset. In other words, init() can not fail.

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -267,7 +262,7 @@ follow_redirect( fd_sshttp_t *        http,
 
   FD_LOG_NOTICE(( "following redirect to http://" FD_IP4_ADDR_FMT ":%hu%.*s",
                   FD_IP4_ADDR_FMT_ARGS( http->addr.addr ), fd_ushort_bswap( http->addr.port ),
-                  (int)headers[ 0 ].value_len, headers[ 0 ].value ));
+                  (int)location_len, location ));
```
amass-jump, line 268: @cali-jumptrading am I missing something here, or is this change correct?

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -125,29 +126,23 @@ fd_sshttp_init( fd_sshttp_t * http,
...
-  if( FD_UNLIKELY( -1==setsockopt( http->sockfd, SOL_SOCKET, SO_RCVTIMEO, &timeout, sizeof(timeout) ) ) ) FD_LOG_ERR(("setsockopt SO_RCVTIMEO failed (%d-%s)", errno, fd_io_strerror(errno)));
-
   struct sockaddr_in addr_in = {
     .sin_family = AF_INET,
     .sin_port   = addr.port,
     .sin_addr   = { .s_addr = addr.addr }
   };
 
-  if( FD_UNLIKELY( -1==connect( http->sockfd, fd_type_pun_const( &addr_in ), sizeof(addr_in) ) ) ) {
+  if( FD_LIKELY( -1==connect( http->sockfd, fd_type_pun_const( &addr_in ), sizeof(addr_in) ) ) ) {
```
cali-jumptrading, line 146: why the change to likely?

amass-jump: For nonblocking sockets, the expected outcome is for connect to return -1 and EINPROGRESS. In fact I think that is the only outcome (it's possible localhost may connect immediately but I would be surprised)

src/discof/restore/utils/fd_sshttp.c
cali-jumptrading: did you run it on testnet? should be the correct change if everything's working

cali-jumptrading (approved):

src/discof/restore/utils/fd_sshttp.c
amass-jump: Yes, it works fine now. But tbh I'm actually confused how it worked before

cali-jumptrading: who knows?

amass-jump:
@ripatel-fd suggested we should keep this as a blocking recv() to save on power. I am going to look into whether we can do that while still getting an accurate busy%, similar perf, etc.

mmcgee-jump:
Power usage does not matter for this tile, it shuts down once the snapshot is loaded

amass-jump:
> Power usage does not matter for this tile, it shuts down once the snapshot is loaded

Beyond simply saving power, a blocking poll seems to benefit throughput somewhat when we are being network-limited

src/discof/restore/fd_snapld_tile.c
```diff
@@ -169,7 +169,10 @@ after_credit( fd_snapld_tile_t *  ctx,
               fd_stem_context_t * stem,
               int *               opt_poll_in FD_PARAM_UNUSED,
               int *               charge_busy ) {
-  if( ctx->state!=FD_SNAPSHOT_STATE_PROCESSING ) return;
+  if( ctx->state!=FD_SNAPSHOT_STATE_PROCESSING ) {
+    fd_log_sleep( (long)1e6 );
```
ripatel-fd, line 173: You might want to call STEM_(in_update) and metrics_write before going to sleep

src/discof/restore/utils/fd_sshttp.c
```diff
@@ -364,19 +362,31 @@ static int
...
+  if( FD_UNLIKELY( -1==read && errno==EAGAIN ) ) {
+    if( FD_UNLIKELY( ++http->empty_recvs>8UL ) ) {
+      /* If we have gone several iterations without having any data to
+         read, sleep the thread for up to one millisecond, or until
+         the socket is readable again, whichever comes first. */
+      struct pollfd pfd = {
+        .fd = http->sockfd,
+        .events = POLLIN,
+      };
+      poll( &pfd, 1 /*fds*/, 1 /*ms*/ );
```
ripatel-fd: poll uses a different syscall on x86 (`poll`) vs arm (`ppoll`), can you use `fd_syscall_poll` instead?

ripatel-fd (dismissed):

ripatel-fd (approved):

PR #7106, opened 2025-11-06
fuzz: increase runner spad memory

ravyu-jump:


src/flamenco/runtime/tests/fd_solfuzz.c
```diff
@@ -80,7 +80,7 @@ fd_solfuzz_runner_new( fd_wksp_t *                         wksp,
   /* Allocate objects */
   ulong const txn_max  = 16UL;
   ulong const rec_max  = 1024UL;
-  ulong const spad_max = 1000000000UL;
+  ulong const spad_max = 1500000000UL; /* 1.5GB to accommodate 128 accouts 10MB each */
```
mjain-jump:
```suggestion
  ulong const spad_max = 1500000000UL; /* 1.5GB to accommodate 128 accounts 10MB each */
```

ibhatt-jumptrading: really should be sized out to sizeof(fd_exec_stack_t) + some other data

mjain-jump (approved):

src/flamenco/runtime/tests/fd_solfuzz.c
mjain-jump: that would probably blow up our memory requirement. i dont think we need to size out for worst-case for fuzzing

mjain-jump (approved):

PR #7115, opened 2025-11-06
Update MACHINE and OBJ_DIR to icelake for solfuzz CI

nlgripto:
lets us hit avx512 instructions


ripatel-fd:
this is going to lose us AVX2 coverage though, right?

nlgripto:
@ripatel-fd we'll get the cross-arch setup in a followup

PR #7128, opened 2025-11-07
pack/resolv: fix transaction expiration logic

### LINKED ISSUE #7052: resolv: wrong expiration logic
There is a confusion in `resolv` `after_frag` between slot and block height. 
Example: A transaction received during slot S referencing the blockhash of slot S-153 should not expire if 4 slots were skipped in-between (which is common), because the block height difference is 149 only.
The result is that many transactions are wrongly ignored by a leader if there were skips during last 151 slots.

[The offending code](https://github.com/firedancer-io/firedancer/blob/e97dd839818a30be7a22a3c1c28bcfcaa567e192/src/discoh/resolv/fd_resolv_tile.c#L394-L403).

A similar check is done in [pack](https://github.com/firedancer-io/firedancer/blob/e97dd839818a30be7a22a3c1c28bcfcaa567e192/src/disco/pack/fd_pack_tile.c#L37-L41), but this time including some arbitrary margin to cope with block height/slot difference.

* A simple fix would be to replace 151 in `resolv` with this `TRANSACTION_LIFETIME_SLOTS` constant, but it's not satisfactory. 

* I would personally remove these imperfect expiration checks in both `resolv` and `pack` and rely only on checks done in `bank`. 
If these checks are security barriers, then they don't hold long. An attacker could clutter pack with transactions that won't land using other ways. During normal operation, transactions not being forwarded anymore, the number of expired txs is not high.

* Another option is to have `bank` send the true block height along with the block hash and use it for expiration checks.

If you are interested, I can post a PR implementing the option you think is best.

HGuillemet:
Currently, Firedancer discards transactions before sending them to bank tile by considering the slot difference between the reference block and the current slot. This is deviating from the protocol and causes transactions to be wrongly ignored if at least 1 slot was skipped during the last 151 slots. The difference in block heights should be used instead. 

Fixes #7052

* Blockhash map in `resolv` now maps block hashes to block heights instead of to slots.
* `resolv` forwards to `pack` the txs that have a reference block height earlier or equal to 151 before last root block height, instead of the last completed slot.
* `pack` now receives from `resolv` the last root block height as message signature instead of the last observed slot. Only the root block has a finalized block height that we can base on to discard expired txs. When a transaction is received with a new root block height, it discards all buffered txs referencing a slot with a block height earlier than 151 before.
* When we become leader, `pack` discards all txs  based on leader slot block height.
* Add block height in Firedancer `replay`->`resolv` "root advanced" message.
* Add block height to "became leader" message
* Add block height to Frankendancer `bank`->`resolv` "rooted bank" message.
* Replace slot by block height in Frankendancer `bank`->`resolv` "blockhash completed" message.

[The changes to agave](https://github.com/HGuillemet/firedancer-agave/commit/905ac673a8940e14f11ca8cc45b6739d8c41fbe0)
Shall I open a PR against some branch of firedancer agave repo ?


src/disco/fd_txn_p.h
```diff
@@ -15,7 +15,7 @@ struct __attribute__((aligned(64))) fd_txn_p {
      uint rebated_cus; /* requested_exec_plus_acct_data_cus-actual used CUs. Pack reads this for CU rebating. */
      uint actual_consumed_cus; /* non_execution_cus+real execution CUs+real account data cus. PoH reads this for block CU counting. */
    } bank_cu; /* Populated by bank. */
-   ulong blockhash_slot; /* Slot provided by resolv tile when txn arrives at the pack tile. Used when txn is in extra storage in pack. */
+   ulong block_height; /* Block height provided by resolv tile when txn arrives at the pack tile. Used when txn is in extra storage in pack. */
```
mmcgee-jump: This is the `blockhash_block_height`, not `block_height`

src/disco/pack/fd_pack_tile.c
```diff
@@ -861,17 +854,17 @@ during_frag( fd_pack_ctx_t * ctx,
     ulong addr_table_sz = 32UL*txn->addr_table_adtl_cnt;
     FD_TEST( addr_table_sz<=32UL*FD_TXN_ACCT_ADDR_MAX );
 
-    if( FD_UNLIKELY( (ctx->leader_slot==ULONG_MAX) & (sig>ctx->highest_observed_slot) ) ) {
-      /* Using the resolv tile's knowledge of the current slot is a bit
+    if( FD_UNLIKELY( (ctx->leader_slot==ULONG_MAX) & (sig>ctx->root_block_height) ) ) {
+      /* Using the resolv tile's knowledge of the current last root block height is a bit
```
mmcgee-jump: Rewrap to 72 columns

src/disco/pack/fd_pack_tile.c
```diff
@@ -861,17 +854,17 @@ during_frag( fd_pack_ctx_t * ctx,
...
          of a hack, since we don't get any info if there are no
          transactions and we're not leader.  We're actually in exactly
          the case where that's okay though.  The point of calling
          expire_before long before we become leader is so that we don't
          drop new but low-fee-paying transactions when pack is clogged
          with expired but high-fee-paying transactions.  That can only
          happen if we are getting transactions. */
-      ctx->highest_observed_slot = sig;
-      ulong exp_cnt = fd_pack_expire_before( ctx->pack, fd_ulong_max( ctx->highest_observed_slot, TRANSACTION_LIFETIME_SLOTS )-TRANSACTION_LIFETIME_SLOTS );
+      ctx->root_block_height = sig;
```
mmcgee-jump, line 1008: This isn't the root block height though? It's the block height of the referenced blockhash, I don't think this logic works

mmcgee-jump (changes requested):
Thank you for the contribution!

src/disco/fd_txn_p.h
HGuillemet: Right, I missed that because this field seems only used when `FD_PACK_USE_EXTRA_STORAGE` is on.

src/disco/pack/fd_pack_tile.c
HGuillemet: Ok. Comments are 72-wide but not the code ?

src/disco/pack/fd_pack_tile.c
HGuillemet: The idea is to pass the root block height as the sig, and use it to discard already buffered txs, and the reference (= blockhash) block height of the sent tx in the `reference_block_height` of the `fd_txn_m`.
Why do you think it won't work ?

src/disco/pack/fd_pack_tile.c
```diff
@@ -1076,18 +1069,18 @@ after_frag( fd_pack_ctx_t *     ctx,
...
         fd_histf_sample( ctx->insert_duration, (ulong)insert_duration );
         ctx->current_bundle->bundle = NULL;
       }
     } else {
-      ulong blockhash_slot = sig;
+      ulong block_height = sig;
       ulong deleted;
       long insert_duration = -fd_tickcount();
-      int result = fd_pack_insert_txn_fini( ctx->pack, ctx->cur_spot, blockhash_slot, &deleted );
+      int result = fd_pack_insert_txn_fini( ctx->pack, ctx->cur_spot, block_height, &deleted );
```
HGuillemet: There is a problem here though. `fd_pack_insert_txn_fini` should be called with the reference block hash, not the root block hash passed using `sig`.
Couldn't we remove this third argument and use `ctx->cur_spot->txnp->reference_block_height` instead, `ctx->cur_sport` being the second argument ?

HGuillemet: Pushed a new commit where the reference block height is read by `fd_pack_insert_fini` from `reference_block_height`, when the tx come from extra storage or directly from `resolv`.

HGuillemet:
Merged with current main. Fixed for full FD. Tested on testnet with full FD.

PR #7130, opened 2025-11-07
replay: improve reliability of replay to tower vote states sending

topointon-jump:
Eliminate several failure modes when sending vote accounts from replay->tower.

### DIFF
--- a/src/discof/replay/fd_replay_tile.c
+++ b/src/discof/replay/fd_replay_tile.c
@@ -549,10 +549,10 @@ publish_stake_weights( fd_replay_tile_t *   ctx,
    - stake:          The stake amount associated with this vote account
    - vote_tower_out: Output structure to populate with vote state information
 
-   Failure modes:
-   - Vote account data is too large (returns -1)
-   - Vote account is not found in Funk (returns -1) */
-static int
+   This function should never fail, as the vote_states cache is guarenteed to
+   only contain valid vote accounts.
+*/
+static void
 fd_replay_out_vote_tower_from_funk(
     fd_accdb_user_t *         accdb,
     fd_funk_txn_xid_t const * xid,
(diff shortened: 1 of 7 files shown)


src/discof/replay/fd_replay_tile.c
```diff
@@ -617,12 +617,12 @@ buffer_vote_towers( fd_replay_tile_t *        ctx,
...
     if( FD_UNLIKELY( ctx->vote_tower_out_len >= (FD_REPLAY_TOWER_VOTE_ACC_MAX-1UL) ) ) FD_LOG_ERR(( "vote_tower_out_len too large" ));
-    if( FD_UNLIKELY( fd_replay_out_vote_tower_from_funk( ctx->accdb,
-                                                         xid,
-                                                         vote_account_pubkey,
-                                                         vote_state->stake,
-                                                         &ctx->vote_tower_out[ctx->vote_tower_out_len++] ) ) ) {
+    fd_replay_tower_t * vote_tower_out = &ctx->vote_tower_out[ ctx->vote_tower_out_len ];
+    int err = fd_replay_out_vote_tower_from_funk( ctx->accdb, xid, vote_account_pubkey, vote_state->stake, vote_tower_out );
+    if( FD_UNLIKELY( err ) ) {
       FD_LOG_DEBUG(( "failed to get vote state for vote account %s", FD_BASE58_ENC_32_ALLOCA( vote_account_pubkey->uc ) ));
```
mmcgee-jump: But why would this happen?

topointon-jump: Vote accounts could be closed after they are added to the `vote_states` cache - example https://solscan.io/account/Dxy5rr77bH8j34QupCLj6hMGb5UahYhhYfC8RE3jse4p

topointon-jump: We don't want to duplicate the tower parsing logic here, so I added some basic sanity-checks (lamports=0, discriminant being non-zero), and Tower itself should skip them if they are corrupted any further.

src/discof/replay/fd_replay_tile.c
```diff
@@ -585,6 +585,17 @@ fd_replay_out_vote_tower_from_funk(
       return -1;
     }
 
+    /* If the vote account has no lamports, no data, or the first 2 bytes are zeros,
+       then it is not a valid vote account and should be skipped */
+    if( FD_UNLIKELY( fd_accdb_ref_lamports( peek->acc ) == 0 ) ) {
+      FD_LOG_WARNING(( "vote account %s has no lamports", FD_BASE58_ENC_32_ALLOCA( pubkey->uc ) ));
+      return -1;
+    }
+    if( FD_UNLIKELY( data_sz == 0UL || memcmp( vote_tower_out->acc, "\0\0", 2 ) == 0 ) ) {
```
ibhatt-jumptrading: can you explain why we need to check if the first two bytes are 0

topointon-jump: It's to check that the vote account has a valid discriminant. Added a comment. Tower should do full checking of the vote account state, otherwise we have to do this parsing twice, but this is a sanity-check that the account state has been initialized.

ibhatt-jumptrading:
i was thinking about this a little bit, we could just cross reference with the current vote states because that stays up to date so if an account is not in vote_states but is in vote_states_prev, we get rid of it. So I propose we just add stakes_t_1 to fd_vote_states.c/h. This makes it so we can reliably just reference the current vote states for the source of truth and don't need to rely on funk failure modes or introspecting on the actual account data. This should be a quick fix: you just need to copy how the vote states has the stake values from t-2

topointon-jump:
updated this so that `fd_replay_out_vote_tower_from_funk` can never fail (cc @ibhatt-jumptrading @mmcgee-jump)

PR #7142, opened 2025-11-08
bundle_tile: clear errorno to get deterministic value when iterating through directories

esemeniuc:
Errorno was not cleared before usage


mmcgee-jump (approved):

mmcgee-jump:
Thank you! Nice find

PR #7153, opened 2025-11-10
replay: reasm fec out bug fix

nipatel-jump:


src/discof/replay/fd_replay_tile.c
```diff
@@ -1731,6 +1731,18 @@ replay( fd_replay_tile_t *  ctx,
   return charge_busy;
 }
 
+static int
+verify_process_fec( fd_replay_tile_t * ctx ) {
```
ibhatt-jumptrading: call this can_process_fec

src/discof/replay/fd_replay_tile.c
```diff
@@ -1731,6 +1731,18 @@ replay( fd_replay_tile_t *  ctx,
...
+  fd_reasm_fec_t * fec;
+  if( FD_UNLIKELY( !fd_sched_can_ingest( ctx->sched, 1UL ) ) )               return 0;
+  if( FD_UNLIKELY( (fec = fd_reasm_peek( ctx->reasm ))==NULL ) )             return 0;
+  if( FD_UNLIKELY( fec->fec_set_idx==0 ) && fd_banks_is_full( ctx->banks ) ) return 0;
```
ibhatt-jumptrading: I think you just did the FD_UNLIKELY around the first cond

ibhatt-jumptrading: I would also make this condition is banks is not full so the standard case can exit early without a reasm lookup.

should be smth like
if( banks_not_full && fec->idx!=0 ) pop_from_reasm()

nipatel-jump: I think its a little more complicated than this, but I made the change so that we only query in the one case we need to.

src/discof/replay/fd_replay_tile.c
```diff
@@ -1731,6 +1731,29 @@ replay( fd_replay_tile_t *  ctx,
...
 
+static int
+can_process_fec( fd_replay_tile_t * ctx ) {
+  fd_reasm_fec_t * fec;
+  if( FD_UNLIKELY( !fd_sched_can_ingest( ctx->sched, 1UL ) ) ) return 0;
+  if( FD_UNLIKELY( (fec = fd_reasm_peek( ctx->reasm ))==NULL ) ) return 0;
+
+  /* If fec_set_idx is 0, we need a new bank for a new slot.  Banks must
+     not be full in this case. */
+  if( FD_UNLIKELY( fec->fec_set_idx==0 && fd_banks_is_full( ctx->banks ) ) ) return 0;
```
ibhatt-jumptrading: sorry for the nits but why don't you just do
if( banks_is_full && fec_set_idx==0 ) return 0;
return 1

src/discof/replay/fd_replay_tile.c
```diff
@@ -1731,6 +1731,22 @@ replay( fd_replay_tile_t *  ctx,
...
```
ibhatt-jumptrading: super nit but flip these two so we short circuit out better

PR #7165, opened 2025-11-11
pack: fix top_writers performance bug

jherrera-jump:


src/disco/pack/fd_pack.c
```diff
@@ -2705,18 +2687,23 @@ fd_pack_end_block( fd_pack_t * pack ) {
...
+          fd_memcpy( &pack->top_writers[ FD_PACK_TOP_WRITERS_CNT-1UL ], writer, sizeof(fd_pack_addr_use_t) );
+          fd_pack_writer_cost_sort_insert( pack->top_writers, FD_PACK_TOP_WRITERS_CNT );
+      }
+
       /* Clearing the cost field here is unnecessary (since it gets
          cleared on insert), but makes debugging a bit easier. */
-      pack->written_list[ pack->written_list_cnt - 1UL - i ]->total_cost = 0UL;
-      acct_uses_remove( pack->writer_costs, pack->written_list[ pack->written_list_cnt - 1UL - i ] );
+      writer->total_cost = 0UL;
+      acct_uses_remove( pack->writer_costs, writer );
```
ptaffet-jump, line 2700: This is fine, but counter-intuitively, what you want is a min-heap data structure to capture the top K.
Something like:

```
iterate over elements
  if heap is not full:
    insert current element into heap
  else:
    if current element is larger than the min in the heap, pop the min and push the current element
```

How different will the performance be in practice? Pretty minimal with a heap this small.

ptaffet-jump: Then I guess you'd need to sort one time at the end to make the order what the caller is expecting

src/disco/pack/fd_pack.h
```diff
@@ -299,9 +280,18 @@ void fd_pack_set_block_limits( fd_pack_t * pack, fd_pack_limits_t const * limits
...
+
+   The opt_limits_usage->top_writers field is ignored. */
 void fd_pack_get_block_limits( fd_pack_t * pack, fd_pack_limits_usage_t * opt_limits_usage, fd_pack_limits_t * opt_limits );
 
+/* fd_pack_get_top_writers copies the top FD_PACK_TOP_WRITERS_CNT by
+   writer cost writers into top_writers.
+
+   This should be called at the end of the relevant leader slot right
+   after fd_pack_end_block, otherwise the copied data will be stale. */
+void fd_pack_get_top_writers( fd_pack_t * pack, fd_pack_addr_use_t * top_writers );
```
ptaffet-jump, line 293: `pack` should be `const`, right? And you can use `[static FD_PACK_TOP_WRITERS_CNT]` (see other places in the codebase for this obtuse syntax) to make it more clear. The comment should make it clear if they are in order or not

PR #7171, opened 2025-11-11
replay: fix next_leader_slot check

topointon-jump:
`fd_multi_epoch_leaders_get_next_slot` returns ULONG_MAX, not 0, if it can't find the next leader slot


mmcgee-jump (changes requested):
I see 3 callers in the file but only 2 updated?

mmcgee-jump (approved):

PR #7175, opened 2025-11-12
accdb: add vinyl bench-query-rand

ripatel-fd:
- Adds a async/pipelined random key-value read benchmark
- Fixes init of vinyl_data allocator


jvarela-jump (approved):

jvarela-jump:
Required https://github.com/firedancer-io/firedancer/pull/7194, which has just been merged.

PR #7177, opened 2025-11-12
util: use __builtin_uaddl_overflow for SCRATCH_ALLOC

ripatel-fd:


src/util/bits/fd_bits.h
```diff
@@ -999,11 +999,11 @@ fd_ulong_svw_dec_tail( uchar const * b,
     ulong _align = (align);                                                                         \
     ulong _sz    = (sz);                                                                            \
     ulong _scratch_alloc = fd_ulong_align_up( _##layout, (align) );                                 \
```
amass-jump: Unrelated to your changes but I wonder if this should be `_align` and not `(align)` such that the latter is only evaluated once

ripatel-fd: Fixing

amass-jump (approved):

PR #7178, opened 2025-11-12
vinyl: fix client_idx selection

ripatel-fd:


ptaffet-jump (approved):
Forwarding @kbowers-jump 's approval

PR #7180, opened 2025-11-12
gui: shred vis replay timestamps

jherrera-jump:


src/disco/gui/fd_gui_tile.c
```diff
@@ -282,6 +284,18 @@ after_frag( fd_gui_ctx_t *      ctx,
       fd_gui_plugin_message( ctx->gui, sig, ctx->buf, fd_clock_now( ctx->clock ) );
       break;
     }
+    case IN_KIND_EXEC_REPLAY: {
+      FD_TEST( ctx->is_full_client );
+      if( FD_LIKELY( sig>>32==FD_EXEC_TT_TXN_EXEC ) ) {
+        fd_exec_task_done_msg_t * msg = (fd_exec_task_done_msg_t *)ctx->buf;
+
+        long tsorig_ns = ctx->ref_wallclock + (long)((double)(fd_frag_meta_ts_decomp( tsorig, fd_tickcount() ) - ctx->ref_tickcount) / fd_tempo_tick_per_ns( NULL ));
+        long tspub_ns = ctx->ref_wallclock + (long)((double)(fd_frag_meta_ts_decomp( tspub, fd_tickcount() ) - ctx->ref_tickcount) / fd_tempo_tick_per_ns( NULL ));
```
mmcgee-jump: Just call `fd_tickcount` once it's expensive, and I'd cache too `fd_tempo_tick_per_ns` in the tile context too, to avoid the weird FD_ONCE lock stuff

src/disco/gui/fd_gui_tile.c
```diff
@@ -105,6 +107,7 @@ typedef struct {
      here to handle those cases until fd_clock is more widely adopted. */
   long ref_wallclock;
   long ref_tickcount;
+  const double tick_per_ns;
```
mmcgee-jump, line 110: Hmm what does `const` do here?

jherrera-jump: nothing really, just indicates to readers that the value doesn't change

PR #7182, opened 2025-11-12
topo: allow artificial raising of wksp size

ripatel-fd:


ripatel-fd:
https://github.com/firedancer-io/firedancer/commit/9086c0b3785d1bed4e0245f53396fb47a14d8534

PR #7189, opened 2025-11-12
Add basic Thread Safety Analyser support for fd_rwlock

two-heart:
Thread Safety Analyser (TSA) is a clang feature to statically (no runtime overhead) check lock semantics. With this we are able to, at compile time, detect things like a missing _end_locking_modify calls, unlocking without locking, and more. This commit adds the compiler flag and annotations to the point that the checks fd_rwlock compile. For more focused reviews, I will add more precise checking per component in follow up PRs. Including things like:
- GUARDED_BY declare what field needs what lock on access
- ACQUIRED_AFTER declare a unified order for lock acquisition to prevent deadlocks

This deprecates and is more accurate than the removed CodeQL queries.


src/flamenco/fd_rwlock.h
```diff
@@ -4,16 +4,15 @@
...
+typedef struct FD_CAPABILITY("fd_rwlock") fd_rwlock {
   ushort value; /* Bits 0..16 are
 
                     0: Unlocked
                     1..=0xFFFE: Locked by N readers
                     0xFFFF: Write locked */
-};
-
-typedef struct fd_rwlock fd_rwlock_t;
+} fd_rwlock_t;
```
ripatel-fd: can you revert back to a separate typedef?

two-heart: yes, done

src/flamenco/runtime/context/fd_capture_ctx.c
intrigus-lgtm, line 1: Remove empty file?

PR #7190, opened 2025-11-12
Documentation updates

esemeniuc:
Document memory configuration

### DIFF
--- a/doc/testing.md
+++ b/doc/testing.md
@@ -26,7 +26,10 @@ make -j
 make run-unit-test
 ```
 
-For large page and NUMA configuration, refer to `./test.sh --help`.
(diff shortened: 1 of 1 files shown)


deps.sh
```diff
@@ -199,6 +199,7 @@ check_debian_pkgs () {
     pkgconf            # build system
     zstd               # build system
     gzip               # build system
+    libbz2-dev         # blocksort
```
ripatel-fd: Why did you need to include this dependency? Firedancer should compile fine without it

esemeniuc: it fails to link

esemeniuc: nvm had some stuff from v0.7 that make it error

esemeniuc: updated now

PR #7195, opened 2025-11-12
gui: fix query_rankings on full client

jherrera-jump:


src/disco/gui/fd_gui.c
```diff
@@ -1536,6 +1537,8 @@ static fd_gui_slot_t *
 fd_gui_clear_slot( fd_gui_t *      gui,
                    ulong           _slot,
                    ulong           _parent_slot ) {
+  gui->summary.startup_slot = fd_ulong_if( gui->summary.startup_slot==ULONG_MAX, _slot, gui->summary.startup_slot );
```
mmcgee-jump: This feels a bit implicit to me .. hard to prove what it ends up being. Isn't it just the genesis or snapshot slot anyway?

mmcgee-jump (approved):

PR #7207, opened 2025-11-13
runtime: remove cluster version toml option

ibhatt-jumptrading:


.github/workflows/benchmark.yml
```diff
@@ -91,7 +91,7 @@ jobs:
         sudo $OBJDIR/bin/firedancer-dev configure init keys
         DUMP_DIR=../dump ./src/flamenco/runtime/tests/run_ledger_backtest.sh \
           -l mainnet-368528500-perf \
-          -y 5 -m 2000000 -e 368529500 -c 3.0.0 \
```
kbhargava-jump, line 94: can you remove the `-c 3.0.0` on line 65 too. its okay if baseline fails on this PR

PR #7211, opened 2025-11-13
gui: vote slot message

jherrera-jump:


src/disco/gui/fd_gui.c
```diff
@@ -2540,6 +2547,12 @@ fd_gui_handle_tower_update( fd_gui_t *                   gui,
                             long                         now ) {
   (void)now;
 
+  if( FD_LIKELY( tower->vote_slot!=ULONG_MAX && gui->summary.slot_voted!=tower->vote_slot ) ) {
```
mmcgee-jump: What's this used for?

jherrera-jump: overview page. Changed to fetch slot from bank vote states.

book/api/websocket.md
```diff
@@ -1568,6 +1576,7 @@ initially replay one but the cluster votes on the other one.
 | transaction_fee              | `string\|null` | Total amount of transaction fees that this slot collects in lamports after any burning |
 | priority_fee                 | `string\|null` | Total amount of priority fees that this slot collects in lamports after any burning |
 | tips                         | `string\|null` | Total amount of tips that this slot collects in lamports, across all block builders, after any commission to the block builder is subtracted |
+| vote_slot                    | `number\|null` | The most recent slot for which this valiadtor had voted for as of the time that this slot was replayed.  Will typically be one less than `slot` |
```
mmcgee-jump: Language is a bit vague, and I wouldn't be able to use this without reading the code as a 3rd party dev:

> The highest slot that this validator has landed a vote on chain for, as-of the end of this slot. Typically this would be one less than `slot` implying a vote latency or one, but could be much less. Will be `null` if this validator has not landed any vote at all.

The latter part still needs even more clarity. Is vote landing based on this validators vote_account? When the validator is first booted, do we show `null`, or would we shot vote latencies for votes cast from a prior validator run (e.g. -2500, if we crashed a few minutes ago), etc etc

mmcgee-jump (changes requested):
.

book/api/websocket.md
jherrera-jump: sg. Updated the comment assuming that bank vote_states is always in sync with the accounts DB at the end of a slot.

src/disco/gui/fd_gui_peers.c
```diff
@@ -676,6 +677,12 @@ fd_gui_peers_handle_vote_update( fd_gui_peers_ctx_t *  peers,
   for( ulong i=0UL; i<vote_cnt; i++ ) {
     if( FD_UNLIKELY( votes_sorted[ i ].stake==ULONG_MAX ) ) continue;
 
+    if( FD_UNLIKELY( !memcmp( &votes_sorted[ i ].node_account, identity->uc, sizeof(fd_pubkey_t) ) && peers->slot_voted!=votes_sorted[ i ].last_vote_slot ) ) {
```
mmcgee-jump, line 696: I don't think this is quite right. Not seeing a vote from our identity on a slot does not mean it stays at what it was, we may have switched to a fork which has a lower vote_slot now. I think you need to actually pull the vote slot out of each slot properly.

Although this is OK for now and we can fix later, with an appropriate comment

jherrera-jump: `votes_sorted ` is just a copy of the `vote_states` bank field that has been sorted by stake descending and deduplicated. Deduplicated here means if multiple vote accounts point to the same identity key, we go with the one with the most stake. This logic will need to change once `SIMD-0180: Use Vote Account Address To Key Leader Schedule` hits mainnet.

As long as the vote account exists, it will be in `vote_states`, which get initialized at snapshot load and gets updated by the runtime. So, on any given fork, `last_voted_slot` should reflect the last landed vote for ALL the vote accounts from the perspective of that fork's bank, even if that slot didn't have a landed vote for some of the vote accounts.

I'll leave this as a comment in the code.

PR #7218, opened 2025-11-13
ethtool: support multiple XDP tiles with rss_queue_mode=dedicated

amass-jump:
Also:
- Dedup ntuple ports and ignore zero
- Fini device before simple fallback after failed dedicated init
- Fix some minor init failures on various devices

### DIFF
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -1005,9 +1005,10 @@ dynamic_port_range = "8900-9000"
         #   traffic.
         #
         # "dedicated"
-        #   Reserves a dedicated hardware queue for each net tile.  For
-        #   now, only supports net_tile_count = 1.  Uses ethtool 'ntuple'
-        #   to route interesting packets based on UDP dst port.  This mode
+        #   Reserves a dedicated hardware queue for each net tile.  Uses
+        #   ethtool 'ntuple' rules to route incoming packets based on
+        #   UDP dst port.  Packets are load balanced across the dedicated
+        #   queues based on lowest bits of source address.  This mode
         #   may not work with some network device setups.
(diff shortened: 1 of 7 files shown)


src/app/fdctl/config/default.toml
```diff
@@ -1005,9 +1005,10 @@ dynamic_port_range = "8900-9000"
         #   traffic.
         #
         # "dedicated"
-        #   Reserves a dedicated hardware queue for each net tile.  For
-        #   now, only supports net_tile_count = 1.  Uses ethtool 'ntuple'
-        #   to route interesting packets based on UDP dst port.  This mode
+        #   Reserves a dedicated hardware queue for each net tile.  Uses
+        #   ethtool 'ntuple' rules to route interesting packets based on
```
ripatel-fd:
```suggestion
        #   ethtool 'ntuple' rules to route incoming packets based on
```

src/app/firedancer/config/default.toml
```diff
@@ -1004,9 +997,10 @@ user = ""
         #   traffic.
         #
         # "dedicated"
-        #   Reserves a dedicated hardware queue for each net tile.  For
-        #   now, only supports net_tile_count = 1.  Uses ethtool 'ntuple'
-        #   to route interesting packets based on UDP dst port.  This mode
+        #   Reserves a dedicated hardware queue for each net tile.  Uses
+        #   ethtool 'ntuple' rules to route interesting packets based on
```
ripatel-fd:
```suggestion
        #   ethtool 'ntuple' rules to route incoming packets based on
```

ripatel-fd (approved):

ripatel-fd (approved):

PR #7227, opened 2025-11-13
misc: fix compilation without dev deps

mjain-jump:


src/flamenco/runtime/tests/fd_solfuzz_private.h
```diff
@@ -10,13 +10,15 @@
 #include "../../../ballet/nanopb/pb_decode.h"
 #include "generated/context.pb.h"
 
+#ifdef FD_HAS_FLATCC
```
ripatel-fd, line 13:
```suggestion
#if FD_HAS_FLATCC
```

These should all be `#if` not `#ifdef`. Because `FD_HAS_FLATCC=0` means there's no flatcc

src/flamenco/runtime/tests/fd_solfuzz.h
ripatel-fd, line 1: Can we just disable solfuzz entirely if flatbuffers is not found?

PR #7228, opened 2025-11-14
net: fix test_fib4

akhinvasara-jumptrading:
test_fib4 had diverged from the example defined in the comments. For example, there was an invalid subnet definition and unintended collisions between the local and main table. This PR brings the test implementation back in line with the included example.

### DIFF
--- a/src/waltz/ip/test_fib4.c
+++ b/src/waltz/ip/test_fib4.c
@@ -15,17 +15,105 @@ fib2_mem[ 1<<18 ];
 
 static void
 test_fib_print( fd_fib4_t const * fib,
-                char const *      actual ) {
+                char const *      actual,
+                ulong             reorder_after ) {
   static char dump_buf[ 8192 ];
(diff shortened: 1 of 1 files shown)


mjain-jump (dismissed):

src/waltz/ip/test_fib4.c
```diff
@@ -15,17 +15,105 @@ fib2_mem[ 1<<18 ];
...
+          break;
+        }
+      }
+
+      if( FD_UNLIKELY( !found ) ) {
+        fwrite( dump_buf, 1, sz, stderr );
+        fflush( stderr );
+        FD_LOG_ERR(( "FAIL: line %lu from dump not found in expected (unordered section)", i ));
+      }
+    }
```
ripatel-fd, line 116: why do we need all this?

ripatel-fd:
> unintended collisions between the local and main table

There's no such thing as collisions in different routing tables. The local table just has a higher priority

PR #7241, opened 2025-11-14
net: use fd_map_dynamic.c for fib4 hmap

akhinvasara-jumptrading:
This PR switches the hash map in fib4 from slot_para to map_dynamic to avoid locking overhead. For this use case, occasional inconsistent/torn reads are acceptable (just misroute or drop a couple packets, rarely). They are mitigated by atomic operations on the core next_hop struct. This improves xdp tx perf by 10-20%, depending on the xdp mode / ZC.

### DIFF
--- a/src/app/shared/fd_obj_callbacks.c
+++ b/src/app/shared/fd_obj_callbacks.c
@@ -216,7 +216,7 @@ fib4_align( fd_topo_t const *     topo FD_FN_UNUSED,
 static void
 fib4_new( fd_topo_t const *     topo,
            fd_topo_obj_t const * obj ) {
-  FD_TEST( fd_fib4_new( fd_topo_obj_laddr( topo, obj->id ), VAL("route_max"), VAL("route_peer_max"), VAL("route_peer_seed") ) );
+  FD_TEST( fd_fib4_new( fd_topo_obj_laddr( topo, obj->id ), VAL("route_max"), VAL("route_peer_max") ) );
 }
 
 fd_topo_obj_callbacks_t fd_obj_cb_fib4 = {
--- a/src/app/shared/fd_obj_callbacks.c
+++ b/src/app/shared/fd_obj_callbacks.c
@@ -216,7 +216,7 @@ fib4_align( fd_topo_t const *     topo FD_FN_UNUSED,
 static void
 fib4_new( fd_topo_t const *     topo,
            fd_topo_obj_t const * obj ) {
-  FD_TEST( fd_fib4_new( fd_topo_obj_laddr( topo, obj->id ), VAL("route_max"), VAL("route_peer_max") ) );
+  FD_TEST( fd_fib4_new( fd_topo_obj_laddr( topo, obj->id ), VAL("route_max"), VAL("route_peer_max"), VAL("route_peer_seed") ) );
 }
 
 fd_topo_obj_callbacks_t fd_obj_cb_fib4 = {
--- a/src/disco/net/xdp/fd_xdp_tile.c
+++ b/src/disco/net/xdp/fd_xdp_tile.c
@@ -229,8 +229,8 @@ typedef struct {
   fd_net_flusher_t tx_flusher[2]; /* one per XSK */
 
   /* Route and neighbor tables */
-  fd_fib4_t const * fib_local;
-  fd_fib4_t const * fib_main;
+  fd_fib4_t *       fib_local;
+  fd_fib4_t *       fib_main;
   fd_neigh4_hmap_t  neigh4[1];
   fd_netlink_neigh4_solicit_link_t neigh4_solicit[1];
 
--- a/src/waltz/ip/fd_fib4.c
+++ b/src/waltz/ip/fd_fib4.c
@@ -23,7 +23,8 @@ fd_fib4_footprint( ulong route_max,
                    ulong route_peer_max ) {
   if( route_max==0 || route_max>UINT_MAX ||
       route_peer_max==0 || route_peer_max>UINT_MAX ) return 0UL;
-  ulong hmap_footprint = _fd_fib4_hmap_footprint( route_peer_max );
+  int   lg_slot_cnt    = fd_fib4_hmap_est_lg_slot_cnt( route_peer_max ); FD_TEST( lg_slot_cnt>0 );
+  ulong hmap_footprint = fd_fib4_hmap_footprint( lg_slot_cnt );
   if( FD_UNLIKELY( !hmap_footprint ) ) return 0UL;
 
   return FD_LAYOUT_FINI( FD_LAYOUT_APPEND( FD_LAYOUT_APPEND( FD_LAYOUT_APPEND( FD_LAYOUT_APPEND( FD_LAYOUT_INIT,
@@ -37,7 +38,8 @@ fd_fib4_footprint( ulong route_max,
 void *
 fd_fib4_new( void * mem,
              ulong  route_max,
-             ulong  route_peer_max ) {
+             ulong  route_peer_max,
+             ulong  route_peer_seed ) {
 
   if( FD_UNLIKELY( !mem ) ) {
     FD_LOG_WARNING(( "NULL mem" ));
@@ -57,7 +59,8 @@ fd_fib4_new( void * mem,
   }
 
   FD_SCRATCH_ALLOC_INIT( l, mem );
-  ulong  hmap_footprint = _fd_fib4_hmap_footprint( route_peer_max );
+  int   fib_lg_slot_cnt = fd_fib4_hmap_est_lg_slot_cnt( route_peer_max ); FD_TEST( fib_lg_slot_cnt>0 );
+  ulong hmap_footprint  = fd_fib4_hmap_footprint( fib_lg_slot_cnt );
   if( FD_UNLIKELY( !hmap_footprint ) ) {
     FD_LOG_WARNING(( "invalid fd_fib4_hmap_footprint" ));
     return NULL;
@@ -75,7 +78,7 @@ fd_fib4_new( void * mem,
 
   /* hmap_footprint!=0 --> lg_slot_cnt > 0  */
   fd_fib4_hmap_entry_t * map = fd_fib4_hmap_join(
-                                 fd_fib4_hmap_new( fib4_hmap_mem, _fd_fib4_hmap_lg_slot_cnt( route_peer_max ) )
+                                 fd_fib4_hmap_new( fib4_hmap_mem, fib_lg_slot_cnt, route_peer_seed )
                                );
   FD_TEST( map );
 
@@ -134,18 +137,18 @@ fd_fib4_cnt( fd_fib4_t const * fib ) {
 /* fd_fib4_hmap_insert adds a new entry (key=ip4_dst, value=hop) to the fib4
    hmap. Assume the netmask for the ip4_dst entry is 32, and ip4_dst is not 0.
    Fails if the map capacity has hit configured hmap_max. Returns 1 on success,
-   or 0 on failure.
+   or 0 on failure. Assumes we are the only writer.
 */
 
 static inline int
-_fd_fib4_hmap_insert( fd_fib4_t *    fib,
-                     uint            ip4_dst,
-                     fd_fib4_hop_t * hop ) {
+fd_fib4_hmap_insert_entry( fd_fib4_t *     fib,
+                           uint            ip4_dst,
+                           fd_fib4_hop_t * hop ) {
 
   if( FD_UNLIKELY( fib->hmap_cnt>=fib->hmap_max ) ) return 0;
   FD_TEST( hop );
 
-  fd_fib4_hmap_entry_t * map = fd_fib4_hmap( fib );
+  fd_fib4_hmap_entry_t * map   = fd_fib4_hmap( fib );
   fd_fib4_hmap_entry_t * entry = fd_fib4_hmap_insert( map, ip4_dst );
   if( FD_UNLIKELY( !entry ) ) {
     /* update existing entry */
@@ -153,7 +156,13 @@ _fd_fib4_hmap_insert( fd_fib4_t *    fib,
   }
   FD_TEST( entry );
 
-  fd_fib4_hop_st_atomic( &entry->next_hop, hop );
+  fd_fib4_hmap_entry_t to_enter = {
+    .dst_addr = ip4_dst,
+    .hash     = fd_uint_hash( ip4_dst ),
+    .next_hop = *hop
+  };
+  fd_fib4_hmap_entry_st( entry, &to_enter );
+
   fib->hmap_cnt++;
   return 1;
 }
@@ -167,7 +176,7 @@ fd_fib4_insert( fd_fib4_t *     fib,
 
   FD_TEST( hop );
   if( ip4_dst!=0 && prefix==32 ) {
-    if( _fd_fib4_hmap_insert( fib, ip4_dst, hop ) ) return 1;
+    if( fd_fib4_hmap_insert_entry( fib, ip4_dst, hop ) ) return 1;
     FD_LOG_WARNING(( "Failed to insert /32 route " FD_IP4_ADDR_FMT " into fib4 hashmap", FD_IP4_ADDR_FMT_ARGS(ip4_dst) ));
     return 0;
   }
@@ -225,17 +234,16 @@ fd_fib4_insert( fd_fib4_t *     fib,
 }
 
 fd_fib4_hop_t
-fd_fib4_lookup( fd_fib4_t const * fib,
-                uint              ip4_dst,
-                ulong             flags ) {
+fd_fib4_lookup( fd_fib4_t  * fib,
+                uint         ip4_dst,
+                ulong        flags ) {
   if( FD_UNLIKELY( flags ) ) {
     return fd_fib4_hop_tbl_const( fib )[0]; /* dead route */
   }
 
   if( fib->hmap_cnt>0 ) {
-    fd_fib4_hmap_entry_t * map   = fd_fib4_hmap( (fd_fib4_t *)fib );
-
-    fd_fib4_hop_t next_hop = fd_fib4_hmap_query_atomic( map, ip4_dst );
+    fd_fib4_hmap_entry_t * map      = fd_fib4_hmap ( fib );
+    fd_fib4_hop_t          next_hop = fd_fib4_hmap_query_hop( map, ip4_dst );
     if( next_hop.rtype!=FD_FIB4_RTYPE_UNSPEC ) {
       return next_hop;
     }
--- a/src/waltz/ip/fd_fib4.c
+++ b/src/waltz/ip/fd_fib4.c
@@ -23,26 +23,21 @@ fd_fib4_footprint( ulong route_max,
                    ulong route_peer_max ) {
   if( route_max==0 || route_max>UINT_MAX ||
       route_peer_max==0 || route_peer_max>UINT_MAX ) return 0UL;
-  ulong elem_max       = fd_fib4_hmap_get_ele_max( route_peer_max   );
-  ulong probe_max      = fd_fib4_hmap_get_probe_max( elem_max );
-  ulong lock_cnt       = fd_fib4_hmap_get_lock_cnt( elem_max );
-  ulong hmap_footprint = fd_fib4_hmap_footprint( elem_max, lock_cnt, probe_max );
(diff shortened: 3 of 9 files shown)


mmcgee-jump:
Can we just switch this stuff to use message passing? Not sure why we need a custom concurrent shared memory structure for this

ripatel-fd:
> Can we just switch this stuff to use message passing? Not sure why we need a custom concurrent shared memory structure for this

Maintaining replicas of routing tables across different net tiles results in a significantly more complex control plane. The "custom concurrent" part is a single atomic read. Replication of netlink messages is ~hundreds of lines of code and inherently bug-prone due to possibility of dropped updates, and will further decrease performance due to having to poll additional input links.

src/waltz/ip/fd_fib4_private.h
```diff
@@ -45,31 +48,115 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
...
-static inline ulong   fd_fib4_hmap_get_ele_max   ( ulong max_cnt  ) { return fd_ulong_pow2_up( max_cnt + ( max_cnt>>1 ) ); }
-/* Get the hashmap's probe limit (75% of total capacity). Higher than requested size to avoid probe failure */
-static inline ulong   fd_fib4_hmap_get_probe_max ( ulong elem_max ) { return elem_max - ( elem_max>>2 );                      }
-/* Get the hashmap's lock count. Each lock roughly protects 1/16 elements in the hashmap */
-static inline ulong   fd_fib4_hmap_get_lock_cnt  ( ulong elem_max ) { return ( elem_max<=32 ) ? 1 : ( elem_max>>4 );          }
+  return lg_slot_cnt;
+}
+
+static inline ulong
+_fd_fib4_hmap_footprint( ulong route_peer_max) {
```
ripatel-fd: can we remove this redundant footprint functino

src/waltz/ip/fd_fib4_private.h
```diff
@@ -45,31 +48,115 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
...
+  return (fd_fib4_hmap_entry_t const *)( (ulong)fib + fib->hmap_join_offset );
+}
+
+static inline fd_fib4_hmap_entry_t *
+fd_fib4_hmap( fd_fib4_t * fib ) {
+  return (fd_fib4_hmap_entry_t*)fd_fib4_hmap_const( fib );
+}
+
+static inline int
+_fd_fib4_hmap_lg_slot_cnt( ulong route_peer_max ) {
```
ripatel-fd: fd_fib4_hmap_est_lg_slot_cnt to follow convention

src/waltz/ip/fd_fib4_private.h
```diff
@@ -45,31 +48,115 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
...
-#include "../../util/tmpl/fd_map_slot_para.c"
+static inline fd_fib4_hmap_entry_t const *
+fd_fib4_hmap_const( fd_fib4_t const * fib ) {
+  return (fd_fib4_hmap_entry_t const *)( (ulong)fib + fib->hmap_join_offset );
+}
+
+static inline fd_fib4_hmap_entry_t *
+fd_fib4_hmap( fd_fib4_t * fib ) {
+  return (fd_fib4_hmap_entry_t*)fd_fib4_hmap_const( fib );
+}
```
ripatel-fd, line 157: It's not nice to cast const pointers to non-const. Can you just copy paste the offset calculation

src/waltz/ip/fd_fib4_private.h
```diff
@@ -45,31 +48,115 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
...
-struct __attribute__((aligned(16))) fd_fib4_hmap_entry {
-  uint dst_addr; /* Little endian. All 32-bits defined */
   fd_fib4_hop_t next_hop;
-};
+  fd_fib4_hop_ld_atomic( &next_hop, &entry->next_hop );
 
-typedef struct fd_fib4_hmap_entry fd_fib4_hmap_entry_t;
+  /* confirm it's still the same IP we expected */
+  int torn_read = FD_VOLATILE_CONST( entry->dst_addr )!=dst_addr;
+  if( FD_UNLIKELY( torn_read ) ) return null;
```
ripatel-fd: I wouldn't worry about this overrun check. There's no guarantee that the entry gets updated dst_addr first, then hop info, so it's not complete anyways

src/waltz/ip/fd_fib4_private.h
```diff
@@ -45,31 +48,115 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
...
+  #endif
+}
+
+/* Atomically write hop, then write hash and key. Assumes no other writers */
+static void
+fd_fib4_hmap_entry_move( fd_fib4_hmap_entry_t * dst,
+                         fd_fib4_hmap_entry_t const * src) {
+  fd_fib4_hop_st_atomic( &dst->next_hop, &src->next_hop );
+  dst->hash     = src->hash;
+  dst->dst_addr = src->dst_addr; /* write key last, so query hits valid data */
```
ripatel-fd: The compiler is free to reorder this if there are no compiler fences

src/app/shared/fd_obj_callbacks.c
```diff
@@ -216,7 +216,7 @@ fib4_align( fd_topo_t const *     topo FD_FN_UNUSED,
 static void
 fib4_new( fd_topo_t const *     topo,
            fd_topo_obj_t const * obj ) {
-  FD_TEST( fd_fib4_new( fd_topo_obj_laddr( topo, obj->id ), VAL("route_max"), VAL("route_peer_max"), VAL("route_peer_seed") ) );
+  FD_TEST( fd_fib4_new( fd_topo_obj_laddr( topo, obj->id ), VAL("route_max"), VAL("route_peer_max") ) );
```
ripatel-fd: We'll definitely want to keep a hashmap seed for hashdos resistance

src/waltz/ip/fd_fib4.c
```diff
@@ -182,7 +159,7 @@ fd_fib4_insert( fd_fib4_t *     fib,
                 fd_fib4_hop_t * hop ) {
 
   if( ip4_dst!=0 && prefix==32 ) {
-    if( fd_fib4_hmap_insert( fib, ip4_dst, hop )==FD_MAP_SUCCESS ) return 1;
+    if( _fd_fib4_hmap_insert( fib, ip4_dst, hop ) ) return 1;
```
ripatel-fd: `_fd` is not a valid namespace prefix. All Firedancer symbols should be prefixed with `fd_`

src/waltz/ip/fd_fib4_private.h
```diff
@@ -45,31 +48,115 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
 
 
 /* Hashmap private APIs */
+struct __attribute__((aligned(16))) fd_fib4_hmap_entry {
+  uint dst_addr; /* Little endian. All 32-bits defined */
+  fd_fib4_hop_t next_hop;
+  uint hash;
```
ripatel-fd: can we remove the hash here?

ripatel-fd: unless storing the hash improves performance

ripatel-fd: Also the struct layout seems wrong. fib4_hop_t needs to be 16 byte aligned to support atomic reads. So the way this is currently aligned results in a 48 byte struct when it could be 32

src/waltz/ip/fd_fib4_private.h
```diff
@@ -45,31 +48,115 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
...
+static inline void
+fd_fib4_hop_st_atomic( fd_fib4_hop_t       * dst,
+                       fd_fib4_hop_t const * src ) {
+  # if FD_HAS_X86
+   FD_VOLATILE( *(__m128i *)( dst ) ) = *(__m128i const *)( src );
+  # elif FD_HAS_INT128
+    FD_VOLATILE( *(uint128 *)( dst ) ) = *(uint128 const *)( src );
+  # else
+    FD_VOLATILE( *dst ) = *src;
+  #endif
```
ripatel-fd: weird indentation

src/waltz/ip/fd_fib4_private.h
```diff
@@ -45,31 +48,115 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
...
+static inline void
+fd_fib4_hop_ld_atomic( fd_fib4_hop_t       * dst,
+                       fd_fib4_hop_t const * src ) {
+  #if FD_HAS_X86
+    *(__m128i *)( dst ) = FD_VOLATILE_CONST( *(__m128i const *)( src ) );
+  #elif FD_HAS_INT128
+    *(uint128 *)( dst ) = FD_VOLATILE_CONST( *(uint128 const *)( src ) );
+  # else
+    *dst = *src;
+  #endif
```
ripatel-fd: weird indentation

ripatel-fd (changes requested):
Looks good, though I we should use a seeded hash function for HashDoS resistance. The `fib4_hmap_entry` struct layout needs optimization. Rest is just nits

src/waltz/ip/fd_fib4_private.h
akhinvasara-jumptrading: Sorry, a bit confused. MEMOIZE requires the hash field? Or you're saying to not MEMOIZE?

akhinvasara-jumptrading: now that layout is fixed, it will be 32 bytes regardless of the hash field exists. So we might as well MEMOIZE (keep hash)?

src/disco/net/xdp/fd_xdp_tile.c
```diff
@@ -229,8 +229,8 @@ typedef struct {
   fd_net_flusher_t tx_flusher[2]; /* one per XSK */
 
   /* Route and neighbor tables */
-  fd_fib4_t const * fib_local;
-  fd_fib4_t const * fib_main;
+  fd_fib4_t *       fib_local;
+  fd_fib4_t *       fib_main;
```
ripatel-fd, line 233: these should stay read-only, since they are managed by the netlink tile

src/waltz/ip/fd_fib4.c
```diff
@@ -62,32 +59,33 @@ fd_fib4_new( void * mem,
...
-  ulong hmap_footprint     = fd_fib4_hmap_footprint( hmap_elem_max, hmap_lock_cnt, hmap_probe_max );
-  FD_TEST( hmap_footprint );
-  void * fib4_hmap_mem     = FD_SCRATCH_ALLOC_APPEND( l, fd_fib4_hmap_align(), hmap_footprint );
-  void * fib4_hmap_ele_mem = FD_SCRATCH_ALLOC_APPEND( l, alignof(fd_fib4_hmap_entry_t), hmap_elem_max*sizeof(fd_fib4_hmap_entry_t) );
-  FD_TEST( fib4_hmap_mem );
-  FD_TEST( fib4_hmap_ele_mem );
+  int   fib_lg_slot_cnt = fd_fib4_hmap_est_lg_slot_cnt( route_peer_max ); FD_TEST( fib_lg_slot_cnt>0 );
+  ulong hmap_footprint  = fd_fib4_hmap_footprint( fib_lg_slot_cnt );
+  if( FD_UNLIKELY( !hmap_footprint ) ) {
+    FD_LOG_WARNING(( "invalid fd_fib4_hmap_footprint" ));
```
ripatel-fd, line 65:
```suggestion
    FD_LOG_WARNING(( "invalid fib_lg_slot_cnt" ));
```

src/waltz/ip/fd_fib4.c
```diff
@@ -149,36 +136,35 @@ fd_fib4_cnt( fd_fib4_t const * fib ) {
...
-  fd_fib4_hmap_entry_t * ele = fd_fib4_hmap_query_ele( query );
-  ele->dst_addr              = ip4_dst;
-  FD_TEST( hop );
-  ele->next_hop = *hop;
   fib->hmap_cnt++;
-
-  fd_fib4_hmap_publish( query );
-
-  return FD_MAP_SUCCESS;
+  return 1;
```
ripatel-fd, line 167: Can we keep FD_MAP_SUCCESS and FD_MAP_ERR_FULL here? Saves us from having to update call sites

src/waltz/ip/fd_fib4_private.h
```diff
@@ -15,18 +20,16 @@ struct __attribute__((aligned(16))) fd_fib4_key {
...
 struct __attribute__((aligned(FD_FIB4_ALIGN))) fd_fib4 {
-  ulong hmap_offset;
-  ulong hmap_elem_offset;
-  ulong hmap_cnt;
-  ulong hmap_max;
-  ulong generation;
-  ulong cnt;
-  ulong max;
-  ulong hop_off;
+  ulong  hmap_join_offset; /* compute join by adding offset to fib4 base */
```
ripatel-fd, line 23:
```suggestion
  ulong  hmap_join_offset; /* offset to first hashmap slot */
```

src/waltz/ip/fd_fib4_private.h
```diff
@@ -46,31 +49,123 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
...
+}
 
-/* Get the hashmap's total capacity (50% extra capacity beyond the requested size to optimize performance) */
-static inline ulong   fd_fib4_hmap_get_ele_max   ( ulong max_cnt  ) { return fd_ulong_pow2_up( max_cnt + ( max_cnt>>1 ) ); }
-/* Get the hashmap's probe limit (75% of total capacity). Higher than requested size to avoid probe failure */
-static inline ulong   fd_fib4_hmap_get_probe_max ( ulong elem_max ) { return elem_max - ( elem_max>>2 );                      }
-/* Get the hashmap's lock count. Each lock roughly protects 1/16 elements in the hashmap */
-static inline ulong   fd_fib4_hmap_get_lock_cnt  ( ulong elem_max ) { return ( elem_max<=32 ) ? 1 : ( elem_max>>4 );          }
+static inline int
+fd_fib4_hmap_est_lg_slot_cnt( ulong route_peer_max ) {
```
ripatel-fd, line 160: Document that -1 is for invalid lg

src/waltz/ip/fd_fib4_private.h
```diff
@@ -46,31 +49,123 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
...
+  if( FD_UNLIKELY( !route_peer_max )) return -1;
+
+  #define FD_FIB4_DEFAULT_SPARSITY (2.5)
```
ripatel-fd, line 163: just make this a `float const`?

src/waltz/ip/fd_fib4_private.h
```diff
@@ -46,31 +49,123 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
...
+  ulong slot_cnt_bound = (ulong)( FD_FIB4_DEFAULT_SPARSITY * (double)route_peer_max );
```
ripatel-fd, line 164: Use of double requires FD_HAS_DOUBLE guard, would just use float here

src/waltz/ip/fd_fib4_private.h
```diff
@@ -46,31 +49,123 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
...
+  FD_VOLATILE( dst->xmm[1] ) = src->xmm[1];
+# elif FD_HAS_INT128
+  FD_VOLATILE( dst->uf[0] )  = src->uf[0];
+  FD_VOLATILE( dst->uf[1] )  = src->uf[1];
+# elif FD_HAS_AVX
+  FD_VOLATILE( dst->avx[0] ) = src->avx[0];
+# else
+  dst->dst_addr = src->dst_addr;
+  dst->hash     = src->hash;
+  dst->next_hop = src->next_hop;
```
ripatel-fd, line 88: these should still be volatile stores

src/waltz/ip/fd_fib4_private.h
```diff
@@ -46,31 +49,123 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
...
+  dst->xmm[1] = FD_VOLATILE_CONST( src->xmm[1] );
+#elif FD_HAS_INT128
+  dst->uf[0] = FD_VOLATILE_CONST( src->uf[0] );
+  dst->uf[1] = FD_VOLATILE_CONST( src->uf[1] );
+# elif FD_HAS_AVX
+  dst->avx[0] = FD_VOLATILE_CONST( src->avx[0] );
+# else
+  dst->dst_addr = src->dst_addr;
+  dst->hash     = src->hash;
+  dst->next_hop = src->next_hop;
```
ripatel-fd, line 109: These should still be volatile loads

src/waltz/ip/fd_fib4_private.h
```diff
@@ -46,31 +49,123 @@ FD_FN_CONST static inline fd_fib4_hop_t *       fd_fib4_hop_tbl      ( fd_fib4_t
...
+fd_fib4_hmap_entry_ld( fd_fib4_hmap_entry_t       * dst,
+                       fd_fib4_hmap_entry_t const * src ) {
+#if FD_HAS_X86
+  dst->xmm[0] = FD_VOLATILE_CONST( src->xmm[0] );
+  dst->xmm[1] = FD_VOLATILE_CONST( src->xmm[1] );
+#elif FD_HAS_INT128
+  dst->uf[0] = FD_VOLATILE_CONST( src->uf[0] );
+  dst->uf[1] = FD_VOLATILE_CONST( src->uf[1] );
+# elif FD_HAS_AVX
+  dst->avx[0] = FD_VOLATILE_CONST( src->avx[0] );
```
ripatel-fd, line 105: FD_HAS_AVX is dead code because FD_HAS_X86 is always true

src/waltz/ip/fd_fib4_private.h
ripatel-fd: sgtm

src/disco/net/xdp/fd_xdp_tile.c
akhinvasara-jumptrading: would def have preferred that, but unfortunately map_dynamic query is not const

src/waltz/ip/fd_fib4.c
akhinvasara-jumptrading: sure - would just require either 1) directly including fd_map.h or 2) including map.h in map_dynamic but not actually using it there
do you have a preference?

src/waltz/ip/fd_fib4_private.h
akhinvasara-jumptrading: hmm.... agree that FD_HAS_AVX should be the first check
but are you saying that the FD_HAS_INT128 is also dead code? I assume that's not true, from the example of fd_funk_txn_xid_st_atomic

ripatel-fd: No, there are architectures that have native u128 load support but are not x86.

src/disco/net/xdp/fd_xdp_tile.c
ripatel-fd: This will definitely crash then. The hashmap is memory mapped read-only to XDP tiles (at least, it should be), so writes by XDP tiles will segfault. If you use a non-const map query, then there would be multiple unsynchronized writers. This is not the concurrency model we were designing for. The solution is to use (or introduce) a map_dynamic const query operation.

src/waltz/ip/fd_fib4.c
ripatel-fd: Include map.h

ripatel-fd (changes requested):
hashmap query should be const

PR #7247, opened 2025-11-14
txn, sched: allow 128-instruction transactions to pass through to the…

yufeng-jump:
… runtime

### DIFF
--- a/src/ballet/txn/fd_txn_parse.c
+++ b/src/ballet/txn/fd_txn_parse.c
@@ -8,7 +8,8 @@ fd_txn_parse_core( uchar const             * payload,
                    ulong                     payload_sz,
                    void                    * out_buf,
                    fd_txn_parse_counters_t * counters_opt,
-                   ulong *                   payload_sz_opt ) {
+                   ulong *                   payload_sz_opt,
+                   ulong                     instr_max ) {
   ulong i = 0UL;
   /* This code does non-trivial parsing of untrusted user input, which
      is a potentially dangerous thing.  The main invariants we need to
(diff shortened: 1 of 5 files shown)


src/discof/replay/fd_sched.c
```diff
@@ -1318,7 +1318,8 @@ fd_sched_parse_txn( fd_sched_t * sched, fd_sched_block_t * block, fd_sched_alut_
                                     fd_ulong_min( FD_TXN_MTU, block->fec_buf_sz-block->fec_buf_soff ),
                                     txn,
                                     NULL,
-                                    &pay_sz );
+                                    &pay_sz,
+                                    128UL );
```
ibhatt-jumptrading: can you make this is a #define

src/ballet/txn/fd_txn_parse.c
```diff
@@ -244,7 +248,15 @@ fd_txn_parse_core( uchar const             * payload,
 
   if( FD_LIKELY( counters_opt   ) ) counters_opt->success_cnt++;
   if( FD_LIKELY( payload_sz_opt ) ) *payload_sz_opt = i;
-  return fd_txn_footprint( instr_cnt, addr_table_cnt );
+
+  ulong footprint = fd_txn_footprint( instr_cnt, addr_table_cnt );
+  /* FIXME remove this check when 64 instructions is enforced cluster wide. */
```
topointon-jump:
```suggestion
  /* FIXME remove this check when static_instruction_limit is activated on all networks */
```
Just so we can grep for this feature.

ptaffet-jump (approved):
:'( but thanks

PR #7254, opened 2025-11-14
gui: handle oversized config accounts

jherrera-jump:
fix crash on devnet

### DIFF
--- a/src/disco/gui/fd_gui_config_parse.h
+++ b/src/disco/gui/fd_gui_config_parse.h
@@ -12,6 +12,7 @@
 #define FD_GUI_CONFIG_PARSE_VALIDATOR_INFO_ICON_URI_SZ (  80UL)
 #define FD_GUI_CONFIG_PARSE_VALIDATOR_INFO_KEYBASE_USERNAME_SZ (80UL)
 #define FD_GUI_CONFIG_PARSE_VALIDATOR_INFO_MAX_SZ      ( 576UL) /* does not include size of ConfigKeys */
+#define FD_GUI_CONFIG_PARSE_MAX_VALID_ACCT_SZ          (FD_GUI_CONFIG_PARSE_CONFIG_KEYS_MAX_SZ+FD_GUI_CONFIG_PARSE_VALIDATOR_INFO_MAX_SZ)
 
 /* The size of a ConfigKeys of length 2, which is the expected length of ValidatorInfo */
 #define FD_GUI_CONFIG_PARSE_CONFIG_KEYS_MAX_SZ         (1UL + (32UL + 1UL)*2UL)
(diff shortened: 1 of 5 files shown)


src/discof/restore/fd_snapin_tile.c
```diff
@@ -517,7 +518,7 @@ handle_data_frag( fd_snapin_tile_t *  ctx,
         early_exit = fd_snapin_process_account_data( ctx, result );
 
         /* We exepect ConfigKeys Vec to be length 2 */
-        if( FD_UNLIKELY( ctx->gui_out.idx!=ULONG_MAX && !memcmp( result->account_data.owner, fd_solana_config_program_id.key, sizeof(fd_hash_t) ) && result->account_data.data_sz && *(uchar *)result->account_data.data==2UL ) ) {
+        if( FD_UNLIKELY( ctx->gui_out.idx!=ULONG_MAX && !memcmp( result->account_data.owner, fd_solana_config_program_id.key, sizeof(fd_hash_t) ) && result->account_data.data_sz && *(uchar *)result->account_data.data==2UL && result->account_data.data_sz<=FD_GUI_CONFIG_PARSE_MAX_VALID_ACCT_SZ ) ) {
```
mmcgee-jump, line 527: Is this possible that it's more? Isn't that enforced by the config program itself?

jherrera-jump: On devnet theres an account with like 6k bytes in every snapshot that crashes the validator. Maybe an older version of the config program had a bug on that cluster?

src/disco/gui/fd_gui_config_parse.h
```diff
@@ -12,6 +12,7 @@
 #define FD_GUI_CONFIG_PARSE_VALIDATOR_INFO_ICON_URI_SZ (  80UL)
 #define FD_GUI_CONFIG_PARSE_VALIDATOR_INFO_KEYBASE_USERNAME_SZ (80UL)
 #define FD_GUI_CONFIG_PARSE_VALIDATOR_INFO_MAX_SZ      ( 576UL) /* does not include size of ConfigKeys */
+#define FD_GUI_CONFIG_PARSE_MAX_VALID_ACCT_SZ          (FD_GUI_CONFIG_PARSE_CONFIG_KEYS_MAX_SZ+FD_GUI_CONFIG_PARSE_VALIDATOR_INFO_MAX_SZ)
```
mmcgee-jump, line 15: `fd_gui_peers_handle_config_account` has an error branch that doesn't free to cJSON object

src/discof/restore/fd_snapin_tile.c
mmcgee-jump: I would add a comment about how this happens / why it's ok to enforce our own limit here

mmcgee-jump (approved):

PR #7258, opened 2025-11-14
snapshots: allow loading uncompressed archives from disk

amass-jump:


src/disco/stem/fd_stem.c
```diff
@@ -787,6 +787,13 @@ STEM_(run)( fd_topo_t *      topo,
                fd_alloca( FD_STEM_SCRATCH_ALIGN, STEM_(scratch_footprint)( polled_in_cnt, tile->out_cnt, reliable_cons_cnt ) ),
                ctx );
 
+#ifdef STEM_CALLBACK_METRICS_WRITE
+  /* Write final metrics state before shutting down */
+  FD_COMPILER_MFENCE();
+  STEM_CALLBACK_METRICS_WRITE( ctx );
+  FD_COMPILER_MFENCE();
+#endif
+
```
amass-jump, line 796: This change makes it such that all stem tiles will perform a final metrics write on shutdown

src/discof/restore/fd_snapdc_tile.c
```diff
@@ -14,10 +14,12 @@
 
 /* The snapdc tile is a state machine that decompresses the full and
    optionally incremental snapshot byte stream that it receives from the
-   snapld tile. */
+   snapld tile.  In the event that the snapshot is already uncompressed,
+   this tile simply copies the stream to the next tile in the pipeline. */
```
cali-jumptrading, line 18: Do you think it's worth not spinning up this tile at all if loading from an uncompressed snapshot? It would make topology code way more complicated and messy... but would save us this extra memcpy.

src/disco/stem/fd_stem.c
```diff
@@ -787,6 +787,13 @@ STEM_(run)( fd_topo_t *      topo,
                fd_alloca( FD_STEM_SCRATCH_ALIGN, STEM_(scratch_footprint)( polled_in_cnt, tile->out_cnt, reliable_cons_cnt ) ),
                ctx );
 
+#ifdef STEM_CALLBACK_METRICS_WRITE
+  /* Write final metrics state before shutting down */
+  FD_COMPILER_MFENCE();
+  STEM_CALLBACK_METRICS_WRITE( ctx );
+  FD_COMPILER_MFENCE();
+#endif
```
cali-jumptrading, line 795: Perhaps keep this to a separate PR?

src/discof/restore/utils/fuzz_ssarchive_parser.c
```diff
@@ -26,12 +26,13 @@ LLVMFuzzerTestOneInput( uchar const * data,
...
   ulong incremental_slot;
   uchar decoded_hash[ FD_HASH_FOOTPRINT ];
+  int   is_zstd;
 
   char* name_cstr = (char *)malloc( size + 1UL );
   assert( name_cstr );
   fd_cstr_fini( fd_cstr_append_text( fd_cstr_init( name_cstr ), (char const *)data, size ) );
 
-  fd_ssarchive_parse_filename( name_cstr, &full_slot, &incremental_slot, decoded_hash );
+  fd_ssarchive_parse_filename( name_cstr, &full_slot, &incremental_slot, decoded_hash, &is_zstd );
```
cali-jumptrading, line 35: You might have already done this, but make sure the fuzzer doesn't crash with the new additions!

src/discof/restore/fd_snapdc_tile.c
amass-jump: I considered this, but I think it is not a good idea to try that. Ultimately we are concerned with throughput, not latency, and memcpy should always be higher throughput than any other tile in the pipeline. Also, we want to be able to handle both compressed and uncompressed snapshots in the same pipeline at runtime, it's not a configuration option

src/discof/restore/utils/fuzz_ssarchive_parser.c
amass-jump: Yep, the fuzzer still passes 👍

PR #7262, opened 2025-11-15
gui: add end slot reason, blockhash

jherrera-jump:


src/disco/gui/fd_gui_printf.c
```diff
@@ -1552,6 +1552,25 @@ fd_gui_printf_slot_transactions_request( fd_gui_t * gui,
...
+            }
+            case FD_PACK_END_SLOT_REASON_MICROBLOCK: {
+              jsonp_string( gui->http, "end_slot_reason", "microblock_limit" );
+              break;
+            }
+            case FD_PACK_END_SLOT_REASON_LEADER_SWITCH: {
+              jsonp_string( gui->http, "end_slot_reason", "leader_switch" );
+              break;
+            }
+            default: __builtin_unreachable();
```
mmcgee-jump: Would just do FD_LOG_ERR rather than unreachable, the compiler does wacky things with this and makes it hard to debug if someone adds a variant or anything.

mmcgee-jump (changes requested):
.

PR #7263, opened 2025-11-17
banks: acquire write lock for tree modifications

two-heart:


ibhatt-jumptrading:
advance_root_prepare does not modify tree state therefore doesn't need a write lock here. 

edit: it actually modifies state, but happens single threaded with replay so this is okay

two-heart:
Makes sense, in this case we would not need any locks at all. We might want to be more explicit in documenting what functions are called in multithreaded vs single threaded contexts.

PR #7265, opened 2025-11-17
banks: remove unused and broken validate function

two-heart:
there were two bugs: (1) never expanding the macro, because iter was part of the definition (2) using bank_ instead of banks_


ibhatt-jumptrading:
validate function isn't used anywhere, it probably just needs to be removed/reworked

two-heart:
@ibhatt-jumptrading, okay, changed this PR to just remove the function

PR #7266, opened 2025-11-17
banks: fix wrong log message

two-heart:


src/flamenco/runtime/fd_bank.c
```diff
@@ -1068,14 +1068,15 @@ fd_banks_new_bank( fd_banks_t * banks,
   if( FD_UNLIKELY( !parent_bank ) ) {
     FD_LOG_CRIT(( "Invariant violation: parent bank for bank index %lu does not exist", parent_bank_idx ));
   }
-  if( FD_UNLIKELY( !(parent_bank->flags&FD_BANK_FLAGS_INIT) ) ) {
```
ibhatt-jumptrading, line 1071: thanks for the change, why did you change where this was logged

two-heart: Good question, without a good answer, moved it back

PR #7268, opened 2025-11-17
net: only allow IP Ethertype for TX

ripatel-fd:
Crash Firedancer if an app tile attempts to send a packet with an
Ethertype other than IP.  Protects the kernel and network from
buggy code.

### DIFF
--- a/src/disco/net/xdp/fd_xdp_tile.c
+++ b/src/disco/net/xdp/fd_xdp_tile.c
@@ -830,16 +830,22 @@ after_frag( fd_net_ctx_t *      ctx,
   }
 
   /* Construct (inner) ip header */
-  uint   ihl         = FD_IP4_GET_LEN( *(fd_ip4_hdr_t *)iphdr );
-  uint   ver         = FD_IP4_GET_VERSION( *(fd_ip4_hdr_t *)iphdr );
-  uint   ip4_saddr   = FD_LOAD( uint, iphdr+12 );
-  ushort ethertype   = FD_LOAD( ushort, frame+12 );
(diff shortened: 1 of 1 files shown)


src/disco/net/xdp/fd_xdp_tile.c
```diff
@@ -830,16 +830,22 @@ after_frag( fd_net_ctx_t *      ctx,
...
-  uint   ver         = FD_IP4_GET_VERSION( *(fd_ip4_hdr_t *)iphdr );
-  uint   ip4_saddr   = FD_LOAD( uint, iphdr+12 );
-  ushort ethertype   = FD_LOAD( ushort, frame+12 );
-  if( ethertype==fd_ushort_bswap( FD_ETH_HDR_TYPE_IP ) && ver!=0x4 ) {
+  uint   ihl       = FD_IP4_GET_LEN( *(fd_ip4_hdr_t *)iphdr );
+  uint   ver       = FD_IP4_GET_VERSION( *(fd_ip4_hdr_t *)iphdr );
+  uint   ip4_saddr = FD_LOAD( uint, iphdr+12 );
+  ushort ethertype = FD_LOAD( ushort, frame+12 );
+
+  if( FD_UNLIKELY( ethertype!=fd_ushort_bswap( FD_ETH_HDR_TYPE_IP ) ) ) {
```
amass-jump, line 838: We do not support VLAN-tagged interfaces, right?

ripatel-fd: Not at the moment

PR #7273, opened 2025-11-17
net: native XDP bond support 

ripatel-fd:
Adds support for driver/zero-copy support for bonded network interfaces.  This is done by introducing `[net.xdp.native_bond]`, which causes XDP tiles to attach to slave devices instead of the virtual bond device.


src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -205,39 +168,55 @@ init( fd_config_t const * config ) {
...
     FD_TEST( !only_dedicated );
     FD_LOG_WARNING(( "error configuring network device (%s), rss_queue_mode \"auto\" attempted"
                      " \"dedicated\" configuration but falling back to \"simple\".", config->net.interface ));
     /* Wipe partial dedicated configuration before simple init */
-    if( is_bonded ) { for( ulong i=0UL; bond_devices[ i ]!=NULL; i++ ) fini_device( bond_devices[ i ] ); }
-    else            { fini_device( config->net.interface ); }
+    if( is_bonded ) {
+      fd_bonding_slave_iter_t iter_[1];
+      fd_bonding_slave_iter_t * iter = fd_bonding_slave_iter_init( iter_, config->net.interface );
+      for( ; !failed && !fd_bonding_slave_iter_done( iter );
```
amass-jump: don't need `!failed` here

src/app/fdctl/config/default.toml
```diff
@@ -1016,6 +1016,17 @@ dynamic_port_range = "8900-9000"
         #   back to "simple" mode if necessary.
         rss_queue_mode = "simple"
 
+        # Enable native network bonding support.
```
amass-jump, line 1019: Do you think we should add a note about the fact that we do not support bond devices being added / removed from the bond while running?

src/app/fdctl/config/default.toml
```diff
@@ -1016,6 +1016,17 @@ dynamic_port_range = "8900-9000"
...
+        #
+        # If native_bond is set to 'true' and [net.interface] is a 'bond'
+        # type network device, install XDP config at bond slave devices
+        # instead of the bond device itself.
+        #
+        # Requires net_tile_count to be divisble by the number of slave
+        # devices.  This causes outgoing traffic to be load balanced with
+        # an arbitrary/opaque hash policy.  Supports up to 16 slave devices.
+        native_bond = false
```
amass-jump: Not true by default?

src/disco/net/fd_net_tile_topo.c
```diff
@@ -221,3 +269,98 @@ fd_topos_net_tile_finish( fd_topo_t * topo,
...
+
+  /* Add loopback unless found */
+
+  uint lo_idx = if_nametoindex( "lo" ); FD_TEST( lo_idx );
+  ADD_IF_IDX( lo_idx );
+
+  /* Done with config discovery */
+
+  *fds_cnt = if_cnt;
+  int next_fd = 123462;
```
amass-jump, line 317: where does 123462 come from...?

src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -163,13 +126,13 @@ init_device( char const *        device,
     /* Remove a queue from the rxfh table for each net tile. */
     uint rxfh_queue_cnt;
     FD_TEST( 0==fd_ethtool_ioctl_rxfh_get_queue_cnt( &ioc, &rxfh_queue_cnt ) );
-    if( FD_UNLIKELY( net_tile_cnt>=rxfh_queue_cnt ) ) {
+    if( FD_UNLIKELY( queue_cnt>=rxfh_queue_cnt ) ) {
       if( strict ) FD_LOG_ERR(( "error configuring network device (%s), too many net tiles %u for queue count %u.  "
                                 "Try `net.xdp.rss_queue_mode=\"simple\"` or reduce net tile count",
                                 device, net_tile_cnt, rxfh_queue_cnt ));
```
amass-jump, line 132: nit: should we change the error log too?

amass-jump (approved):

src/disco/net/fd_net_tile_topo.c
ripatel-fd: @amass-jump This was always the hardcoded base file descriptor number for XDP sockets... Introduced by @mmcgee-jump

src/app/shared/commands/configure/ethtool-channels.c
ripatel-fd: Happy to, but I couldn't figure out a clean message

src/app/fdctl/config/default.toml
ripatel-fd: Not for now ... Until the feature is better tested

src/app/fdctl/config/default.toml
ripatel-fd: OK will do.

book/snippets/bonding.ansi
```diff
@@ -0,0 +1,10 @@
+$ sudo fdctl configure init bonding
+[32mNOTICE [0m hugetlbfs ... bonding ... unconfigured ... /sys/class/net/bond0/bonding/miimon is 200ms, want at least 5000ms
```
mmcgee-jump: why `hugetlbfs` ?

ripatel-fd: fixing

amass-jump (approved):

book/snippets/bonding.ansi
```diff
@@ -0,0 +1,10 @@
+$ sudo fdctl configure init bonding
+[32mNOTICE [0m bonding ... unconfigured ... /sys/class/net/bond0/bonding/miimon is 200ms, want at least 5000ms
+[32mNOTICE [0m bonding ... configuring
+[32mNOTICE [0m RUN: `echo 5000 | sudo tee /sys/class/net/bond0/bonding/miimon`
+[32mNOTICE [0m RUN: `echo 5000 | sudo tee /sys/class/net/bond0/bonding/downdelay`
+[32mNOTICE [0m RUN: `echo 5000 | sudo tee /sys/class/net/bond0/bonding/peer_notif_delay`
+[32mNOTICE [0m bonding ... done
+
+$ /sys/class/net/bond0/bonding/miimon
```
mmcgee-jump:
```suggestion
$ cat /sys/class/net/bond0/bonding/miimon
```

src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -115,10 +73,15 @@ static int
 init_device( char const *        device,
              fd_config_t const * config,
              int                 dedicated_mode,
-             int                 strict ) {
+             int                 strict,
+             uint                device_cnt ) {
   FD_TEST( dedicated_mode || strict );
 
   uint const net_tile_cnt = config->layout.net_tile_count;
+  if( FD_UNLIKELY( net_tile_cnt%device_cnt!=0 ) ) {
```
mmcgee-jump, line 81: Doesn't this only apply if it's in bonded mode?

src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -273,16 +252,21 @@ check_device_is_modified( char const * device ) {
 static int
 check_device_is_configured( char const *        device,
                             fd_config_t const * config,
-                            int                 dedicated_mode ) {
+                            int                 dedicated_mode,
+                            uint                device_cnt ) {
   uint const net_tile_cnt = config->layout.net_tile_count;
+  if( FD_UNLIKELY( net_tile_cnt%device_cnt!=0 ) ) {
```
mmcgee-jump, line 258: As above

src/app/shared/commands/run/run.c
```diff
@@ -719,6 +727,11 @@ fdctl_check_configure( config_t const * config ) {
                  "Firedancer.", check.message ));
 
   if( FD_LIKELY( !config->development.netns.enabled && 0==strcmp( config->net.provider, "xdp" ) ) ) {
+    check = fd_cfg_stage_bonding.check( config, FD_CONFIGURE_CHECK_TYPE_RUN );
```
mmcgee-jump: I'd probably do if `fd_cfg_stage_bonding.enabled( config ) ...` here since the conditions of when to check this are a little involved

mmcgee-jump: And you currently will call check even in some disabled cases, which could cause to erronously return unconfigured?

src/app/firedancer/config/default.toml
```diff
@@ -1011,6 +1011,17 @@ user = ""
...
+        # Enable native network bonding support.
+        #
+        # If native_bond is set to 'true' and [net.interface] is a 'bond'
+        # type network device, install XDP config at bond slave devices
+        # instead of the bond device itself.
+        #
+        # Requires net_tile_count to be divisble by the number of slave
+        # devices.  This causes outgoing traffic to be load balanced with
+        # an arbitrary/opaque hash policy.  Supports up to 16 slave devices.
+        native_bond = false
```
mmcgee-jump: I don't see why we would default this to off? See my suggested language in 1.5 channel

mmcgee-jump (changes requested):
.

src/app/shared/commands/configure/ethtool-channels.c
ripatel-fd: @mmcgee-jump This branch gets never taken if not in bonded mode, because then `device_cnt` will be 1.

src/app/shared/commands/configure/ethtool-channels.c
ripatel-fd: @mmcgee-jump This branch gets never taken if not in bonded mode, because then device_cnt will be 1.

src/app/firedancer/config/default.toml
ripatel-fd: Ok, I'll enable it by default

src/app/shared/commands/run/run.c
ripatel-fd: addressed

PR #7283, opened 2025-11-17
replay: move bank frozen, do not require frozen in tower

emwang-jump:


ibhatt-jumptrading:
why is it fine for the bank to not be frozen

emwang-jump:
> why is it fine for the bank to not be frozen

we only publish slot complete to tower for a bank we will freeze

PR #7306, opened 2025-11-18
net: improve xdp fseq cacheing

akhinvasara-jumptrading:
Caches sequence numbers for the XDP rings. The sequence numbers owned by the kernel are updated only when needed. The sequence numbers owned by us are updated intermittently. Improves net_tx_full_fail_cnt/net_tx_submit_cnt and net_xdp_tx_ring_empty_descs/net_tx_submit_cnt by ~10x, and increases tx pps by ~9% .

### DIFF
--- a/src/disco/net/xdp/test_xdp_tile1.c
+++ b/src/disco/net/xdp/test_xdp_tile1.c
@@ -717,15 +717,17 @@ xdp_reset( fd_tile_test_ctx_t * test_ctx,
   fd_memset( xsk->ring_fr.frame_ring,  0, xsk->ring_fr.depth * sizeof(ulong)           );
   fd_memset( xsk->ring_cr.frame_ring,  0, xsk->ring_cr.depth * sizeof(ulong)           );
 
+  xdp_fr_ring_prod = xsk->ring_fr.depth/2;
+  xdp_fr_ring_cons = 0;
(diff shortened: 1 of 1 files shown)


src/disco/net/xdp/fd_xdp_tile.c
```diff
@@ -1146,25 +1176,33 @@ before_credit( fd_net_ctx_t *      ctx,
...
   }
 
-  uint comp_cons = FD_VOLATILE_CONST( *rr_xsk->ring_cr.cons );
-  uint comp_prod = FD_VOLATILE_CONST( *rr_xsk->ring_cr.prod );
+  uint comp_cons = rr_xsk->ring_cr.cached_cons;
+  uint comp_prod = rr_xsk->ring_cr.cached_prod; /* might be stale */
+  if( FD_UNLIKELY( comp_cons==comp_prod ) ) {
+    comp_prod = rr_xsk->ring_cr.cached_prod = FD_VOLATILE_CONST( *rr_xsk->ring_cr.prod );
+  }
+  FD_TEST( comp_prod >= comp_cons );
```
ripatel-fd: Does not handle uint32 wraparound

ripatel-fd (changes requested):
uint32 wraparound bug

PR #7312, opened 2025-11-18
config: use default XDP mode

ripatel-fd:
Allow the kernel to select driver XDP when available instead of
always using SKB mode.


src/waltz/xdp/fd_xdp1.c
```diff
@@ -253,7 +253,7 @@ fd_xdp_install( uint           if_idx,
   if(      !strcmp( xdp_mode, "skb"     ) ) uxdp_mode = XDP_FLAGS_SKB_MODE;
   else if( !strcmp( xdp_mode, "drv"     ) ) uxdp_mode = XDP_FLAGS_DRV_MODE;
   else if( !strcmp( xdp_mode, "hw"      ) ) uxdp_mode = XDP_FLAGS_HW_MODE;
```
mmcgee-jump, line 255: Whats "hw" mode and why undocumented?

ripatel-fd: @mmcgee-jump It's a kernel feature for NICs that have eBPF in hardware. Which are basically vaporware

PR #7313, opened 2025-11-18
harness: reject deprecated vote states

ibhatt-jumptrading:


mjain-jump (approved):

topointon-jump:
Should we also change solfuzz not to generate these?

PR #7324, opened 2025-11-19
shmem: work around ASLR failure

ripatel-fd:
Works around occasional run-test-vectors failures

### DIFF
--- a/src/util/shmem/fd_shmem_user.c
+++ b/src/util/shmem/fd_shmem_user.c
@@ -121,7 +121,9 @@ fd_shmem_private_map_rand( ulong size,
     }
   }
 
-  FD_LOG_ERR(( "unable to find random address for memory map after 1000 attempts" ));
+  FD_LOG_WARNING(( "Failed 1000 attempts to find a virtual memory range suitable for an mmap() allocation of 0x%lx, retrying with kernel default ASLR ...", size ));
+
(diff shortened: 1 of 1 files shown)


src/util/shmem/fd_shmem_user.c
```diff
@@ -121,7 +121,9 @@ fd_shmem_private_map_rand( ulong size,
     }
   }
 
-  FD_LOG_ERR(( "unable to find random address for memory map after 1000 attempts" ));
+  FD_LOG_WARNING(( "Failed 1000 attempts to find a virtual memory range suitable for an mmap() allocation of 0x%lx, retrying with kernel default ASLR ...", size ));
```
mmcgee-jump, line 124: Blocking so I can review this in detail. I'm not sure "kernel default ASLR" does anything here with mmap, otherwise there was no point writing this whole thing in the first place.

mmcgee-jump (changes requested):
.

ripatel-fd:
> I'm not sure "kernel default ASLR" does anything here with mmap, otherwise there was no point writing this whole thing in the first place.

@mmcgee-jump On x86, kernel ASLR uses (at most) 32-bit offsets for randomization. It's much more conservative to prevent other brk-like growing regions from crashing into an mmap()ed region, like the stack. On Firedancer, memory regions don't grow post startup, so we can use a much wider ASLR range.

Relevant sysctls are randomize_va_space, mmap_rnd_bits

PR #7327, opened 2025-11-19
net: fix IGMP IPv4 verihl order

two-heart:


ripatel-fd:
cc @jvarela-jump @0x0ece this might affect SNP

PR #7338, opened 2025-11-19
sbpf: fix calldest map for SHT_NOBITS .text

ripatel-fd:
Fixes a mismatch in the ELF loader for SHT_NOBITS .text sections.
If SHT_NOBITS is set and sh_size is not zero, then the instruction
count is zero, but the call destination map is scaled by sh_size.

Therefore, the size of the calldest map and text_cnt need to be
decoupled.

Note that this would not cause a bank hash mismatch, since zero
instruction ELF programs are guaranteed to fail verification (due
to entrypoint being out of bounds).

Found by solfuzz repo 9bede7116179ceb81bacac9a55e27042fc8c98f8423edae4dbac95838c952c62

### DIFF
--- a/src/ballet/sbpf/fd_sbpf_loader.c
+++ b/src/ballet/sbpf/fd_sbpf_loader.c
@@ -1,7 +1,6 @@
 #include "fd_sbpf_loader.h"
 #include "fd_sbpf_instr.h"
 #include "fd_sbpf_opcodes.h"
-#include "../../util/fd_util.h"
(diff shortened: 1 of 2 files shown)


ripatel-fd:
> Do you mind also adding the fixture to test-vectors and updating `test-vectors-commit-sha.txt`? 🙏

@topointon-jump This was from a flatbuffers input, not even a fixture. And flatbuffers support for test_sol_compat is missing. So I can't unfortunately.

PR #7341, opened 2025-11-19
sbpf: fix mismatch in parse_dynamic_relocations

ripatel-fd:
Fixes incorrect integer overflow handling and encoding of an
Option<> type.

### DIFF
--- a/src/ballet/sbpf/fd_sbpf_loader.c
+++ b/src/ballet/sbpf/fd_sbpf_loader.c
@@ -1176,7 +1176,8 @@ fd_sbpf_lenient_elf_parse( fd_sbpf_elf_info_t * info,
       dynamic_table[ dyn.d_tag ] = dyn.d_un.d_val;
     }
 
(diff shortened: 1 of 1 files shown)


src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -1193,33 +1194,37 @@ fd_sbpf_lenient_elf_parse( fd_sbpf_elf_info_t * info,
       }
 
       /* https://github.com/anza-xyz/sbpf/blob/v0.12.2/src/elf_parser/mod.rs#L430-L444 */
-      ulong offset = ULONG_MAX;
-      for( ulong i=0; i<ehdr.e_phnum; i++ ) {
-        /* Again... */
+      _Bool offset_found = 0;
```
mjain-jump: why not `uchar`?

ripatel-fd: Because _Bool has only two possible values, 0 and 1

mjain-jump: i thought using bool was discouraged

mjain-jump (approved):

PR #7342, opened 2025-11-19
fuzz: add flatbuffer fixture execution

mjain-jump:


mjain-jump:
Requires commit SHA update after https://github.com/firedancer-io/test-vectors/pull/166

ripatel-fd (dismissed):

PR #7361, opened 2025-11-20
runtime: clean up block harness init

ripatel-fd:
- Move all lock guards into their own scopes to defend against
  lock leaks
- Reject input instead of crashing if sysvars are missing

### DIFF
--- a/src/flamenco/runtime/tests/fd_block_harness.c
+++ b/src/flamenco/runtime/tests/fd_block_harness.c
@@ -324,51 +324,66 @@ fd_solfuzz_pb_block_ctx_create( fd_solfuzz_runner_t *                runner,
 
   /* Initialize the current running epoch stake and vote accounts */
 
-  fd_vote_states_t * vote_states = fd_bank_vote_states_locking_modify( bank );
-  vote_states = fd_vote_states_join( fd_vote_states_new( vote_states, FD_RUNTIME_MAX_VOTE_ACCOUNTS, 999UL ) );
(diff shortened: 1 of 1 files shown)


src/flamenco/runtime/tests/fd_block_harness.c
```diff
@@ -324,88 +324,106 @@ fd_solfuzz_pb_block_ctx_create( fd_solfuzz_runner_t *                runner,
...
+  }
 
   /* Finish init epoch bank sysvars */
   fd_epoch_schedule_t epoch_schedule_[1];
   fd_epoch_schedule_t * epoch_schedule = fd_sysvar_epoch_schedule_read( funk, xid, epoch_schedule_ );
-  FD_TEST( epoch_schedule );
+  if( FD_UNLIKELY( !epoch_schedule ) ) {
+    FD_LOG_WARNING(( "Block test vector missing epoch_schedule sysvar" ));
+    return NULL;
+  }
```
mjain-jump, line 379: i actually prefer crashing here because otherwise small mutator changes that alter this behavior go unnoticed

mjain-jump (changes requested):
scoping looks good, but dont agree with the silent input rejection

ripatel-fd:
@mjain-jump there is no silent input rejection. `test_sol_compat` logs a warning and the shared interface reports up an error. It is up to the shared interface user to crash though if it gracefully reports an invalid input. It doesn't make sense to make the entire interface crash.

PR #7364, opened 2025-11-20
gui: show producer shreds

jherrera-jump:


jherrera-jump:
can be merged after https://github.com/firedancer-io/firedancer/pull/7385

PR #7392, opened 2025-11-21
gui: fix cJSON_Parse overrun

jherrera-jump:
cJSON_ParseWithLengthOpts seems to access memory one past the end of the parseable region, even when `require_null_terminated=0`

found by fuzzer, credit to @nlgripto for find + fix


mmcgee-jump:
Huh, weird. Can you file an issue or something for any other users of same function?

PR #7398, opened 2025-11-22
accdb: add vinyl-funk overlay prototype

ripatel-fd:
Adds a second accdb backend using a read-only vinyl base layer
and a read-write funk overlay.  This is close to the target design,
but excludes the "migrating records from vinyl to funk" step.

This solution is probably not production-shippable, but it allows
us to get some real-world exposure to the vinyl top layer.


src/flamenco/accdb/fd_accdb_ref.h
```diff
@@ -24,6 +24,7 @@
 struct fd_accdb_ref {
   ulong rec_laddr;
   ulong meta_laddr;
+  uchar address[32];  /* only for vinyl requests */
```
jvarela-jump, line 27: Probably worth using a macro here for the address size, and wherever the address is initialized or memcpy-ed?

jvarela-jump (approved):

src/flamenco/accdb/fd_accdb_ref.h
ripatel-fd: Good point, I'll change this to use `fd_pubkey_t` in the next PR.

PR #7404, opened 2025-11-24
restore: fix Zstandard frame handling

ripatel-fd:
The snapdc tile is not conformant to the Zstandard spec.
Namely, Zstandard compressors are supposed to accept arbitrary
concatenation of frames and ignore skippable frames.

Fixes a bug where the snapshot loader aborted after the first
Zstandard frame.

### DIFF
--- a/src/discof/restore/fd_snapdc_tile.c
+++ b/src/discof/restore/fd_snapdc_tile.c
@@ -18,8 +18,9 @@
    this tile simply copies the stream to the next tile in the pipeline. */
 
 struct fd_snapdc_tile {
-  int full;
-  int is_zstd;
+  uint full    : 1;
+  uint is_zstd : 1;
+  uint dirty   : 1;  /* in the middle of a frame? */
   int state;
 
   ZSTD_DCtx * zstd;
@@ -95,6 +96,7 @@ handle_control_frag( fd_snapdc_tile_t *  ctx,
   /* All control messages cause us to want to reset the decompression stream */
   ulong error = ZSTD_DCtx_reset( ctx->zstd, ZSTD_reset_session_only );
   if( FD_UNLIKELY( ZSTD_isError( error ) ) ) FD_LOG_ERR(( "ZSTD_DCtx_reset failed (%lu-%s)", error, ZSTD_getErrorName( error ) ));
+  ctx->dirty = 0;
 
   switch( sig ) {
     case FD_SNAPSHOT_MSG_CTRL_INIT_FULL: {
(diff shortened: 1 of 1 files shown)


src/discof/restore/fd_snapdc_tile.c
```diff
@@ -184,6 +186,8 @@ handle_data_frag( fd_snapdc_tile_t *  ctx,
   uchar * out = fd_chunk_to_laddr( ctx->out.wksp, ctx->out.chunk );
 
   if( FD_UNLIKELY( !ctx->is_zstd ) ) {
+    uchar const * in = data+ctx->in.frag_pos;
```
amass-jump: merge conflict?

src/discof/restore/fd_snapdc_tile.c
```diff
@@ -18,8 +18,9 @@
    this tile simply copies the stream to the next tile in the pipeline. */
 
 struct fd_snapdc_tile {
-  int full;
-  int is_zstd;
+  uint full    : 1;
+  uint is_zstd : 1;
+  uint dirty   : 1;  /* in the middle of a frame? */
```
amass-jump, line 23: not sure this is ever being initialized, can we set it to zero in the FD_SNAPSHOT_MSG_CTRL_INIT_ handling?

src/discof/restore/fd_snapdc_tile.c
```diff
@@ -235,22 +241,7 @@ handle_data_frag( fd_snapdc_tile_t *  ctx,
...
-         some trailing padding or garbage at the end of the snapshot.
-
-         This is not valid under the snapshot format and indicates a
-         problem so we abandon the snapshot. */
-      ctx->state = FD_SNAPSHOT_STATE_ERROR;
-      fd_stem_publish( stem, 0UL, FD_SNAPSHOT_MSG_CTRL_ERROR, 0UL, 0UL, 0UL, 0UL, 0UL );
-      return 0;
-    }
-
-    ctx->state = FD_SNAPSHOT_STATE_FINISHING;
```
amass-jump, line 252: It is now impossible to enter the FINISHING state, right? We should update other logic / assertions / etc. to reflect that

ripatel-fd: @amass-jump No, FINISHING state is entered the same way as the uncompressed case. Via a control frame

src/discof/restore/fd_snapdc_tile.c
ripatel-fd: No but good catch

src/discof/restore/fd_snapdc_tile.c
ripatel-fd: Good catch

src/discof/restore/fd_snapdc_tile.c
ripatel-fd: I was wrong, FINISHING state is unused

src/discof/restore/fd_snapdc_tile.c
```diff
@@ -182,9 +174,9 @@ handle_data_frag( fd_snapdc_tile_t *  ctx,
   uchar const * data = fd_chunk_to_laddr_const( ctx->in.wksp, chunk );
   uchar const * in  = data+ctx->in.frag_pos;
   uchar * out = fd_chunk_to_laddr( ctx->out.wksp, ctx->out.chunk );
+  FD_TEST( ctx->in.frag_pos<sz );
```
amass-jump: In the zstd case, frag_pos can == sz because all input data has been consumed but there is still output data in the internal zstd buffer that needs to be flushed out. While in non-zstd case frag_pos must be < sz

PR #7419, opened 2025-11-24
net: use fd_map_slot for fib4 hmap

akhinvasara-jumptrading:
This PR switches the hash map in fib4 from map_slot_para to map_slot to avoid locking overhead. For this use case, occasional inconsistent/torn reads are acceptable (just misroute or drop a couple packets, rarely). They are mitigated by atomic operations on the entry. This improves xdp tx perf by 10-20%, depending on the xdp mode / ZC.

This PR replaces https://github.com/firedancer-io/firedancer/pull/7241

### DIFF
--- a/src/waltz/ip/fd_fib4.c
+++ b/src/waltz/ip/fd_fib4.c
@@ -136,12 +136,14 @@ fd_fib4_clear( fd_fib4_t * fib4_join ) {
 
   if( fib4->hmap_cnt==0 ) return;
 
+
   fib4->hmap_cnt = 0;
+  fd_fib4_hmap_t * hmap_join = fd_type_pun( fib4_join->hmap_join );
   void *           hmap_mem  = fd_fib4_hmap_mem( fib4 );
   ulong            elem_max  = fd_fib4_hmap_get_ele_max( fib4->hmap_max );
+  FD_TEST( fd_fib4_hmap_leave( hmap_join ) );
   FD_TEST( fd_fib4_hmap_new( hmap_mem, elem_max, 1 ) );
 
-  fd_fib4_hmap_t * hmap_join = fd_type_pun( fib4_join->hmap_join );
   ulong            probe_max = fd_fib4_hmap_get_probe_max( elem_max );
   ulong            seed      = fib4->seed;
   FD_TEST( fd_fib4_hmap_join( hmap_join, hmap_mem, elem_max, probe_max, seed ) );
--- a/src/waltz/ip/fd_fib4.h
+++ b/src/waltz/ip/fd_fib4.h
@@ -41,8 +41,11 @@
 struct fd_fib4_priv;
 typedef struct fd_fib4_priv fd_fib4_priv_t;
 struct fd_fib4 {
-   fd_fib4_priv_t * priv;          /* local ptr to shared fib4_priv_t */
-   uchar            hmap_join[64]; /* local join to the hmap - punned internally */
+   /* local ptr to shared fib4_priv_t */
+   fd_fib4_priv_t * priv;
+
+   /* local join to the hmap - punned internally */
+   uchar            hmap_join[64] __attribute__((aligned(8)));
  };
 typedef struct fd_fib4 fd_fib4_t;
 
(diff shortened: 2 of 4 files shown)


src/waltz/ip/fd_fib4.c
```diff
@@ -113,88 +130,86 @@ fd_fib4_delete( void * mem ) {
...
-  ulong probe_max = fd_fib4_hmap_get_probe_max( elem_max );
-  ulong lock_cnt  = fd_fib4_hmap_get_lock_cnt( elem_max );
-  ulong seed      = fd_fib4_hmap_seed( hmap );
-  ulong ignored[ fd_fib4_hmap_lock_max( ) ];
-  FD_TEST( fd_fib4_hmap_lock_range( hmap, 0, lock_cnt, FD_MAP_FLAG_BLOCKING, ignored )==FD_MAP_SUCCESS );
-  FD_TEST( fd_fib4_hmap_new( fd_fib4_hmap_mem( fib4 ), elem_max, lock_cnt, probe_max, seed ) );
-  fd_memset( fd_fib4_hmap_ele_mem( fib4 ), 0, elem_max*sizeof(fd_fib4_hmap_entry_t) );
+  void *           hmap_mem  = fd_fib4_hmap_mem( fib4 );
+  ulong            elem_max  = fd_fib4_hmap_get_ele_max( fib4->hmap_max );
+  FD_TEST( fd_fib4_hmap_new( hmap_mem, elem_max, 1 ) );
```
ripatel-fd, line 145: might want to leave the caller before calling new

src/waltz/ip/fd_fib4.c
```diff
@@ -98,12 +89,38 @@ fd_fib4_new( void * mem,
...
+    return NULL;
+  }
+
+  if( FD_UNLIKELY( !fd_ulong_is_aligned( (ulong)shmem, fd_fib4_align() ) ) ) {
+    FD_LOG_WARNING(( "unaligned shmem" ));
+    return NULL;
+  }
+
+  fd_fib4_priv_t * priv       =  fd_type_pun( shmem );
+  fd_fib4_hmap_t * hmap_join  =  fd_type_pun( join->hmap_join );
```
ripatel-fd, line 109: can you add an FD_STATIC_ASSERT here that `sizeof(fd_fib4_hmap_t)<=size of the hmap_join buffer`

src/waltz/ip/fd_fib4.h
```diff
@@ -34,10 +34,16 @@
...
-struct fd_fib4;
+/* fd_fib4_t is a local handle to a fib4 object. Use fd_fib4_{join,leave}
+   to join the fd_fib4_t to the shmem fib4. A fd_fib4_t can be stack declared,
+   e.g. fd_fib4_t fib4[1]; */
+
+struct fd_fib4_priv;
+typedef struct fd_fib4_priv fd_fib4_priv_t;
+struct fd_fib4 {
+   fd_fib4_priv_t * priv;          /* local ptr to shared fib4_priv_t */
+   uchar            hmap_join[64]; /* local join to the hmap - punned internally */
```
ripatel-fd: missing aligned attribute (compiler could technically underalign the join object)

src/waltz/ip/fd_fib4_private.h
```diff
@@ -14,63 +18,141 @@ struct __attribute__((aligned(16))) fd_fib4_key {
...
+   If the destination address is not found, returns a route with rtype
+   set to UNSPEC. The result is not guaranteed to be valid - the caller
+   is responsible for validating the result. */
+
+static inline fd_fib4_hop_t
+fd_fib4_hmap_query_hop( fd_fib4_hmap_t const * map,
+                        uint                   dst_addr ) {
+  static fd_fib4_hop_t null = {0};
+  fd_fib4_hmap_entry_t const * entry = fd_fib4_hmap_query( map, &dst_addr );
+  if( !entry ) return null;
```
ripatel-fd: remvoe the static null here and just return `(fd_fib4_hop_t){0}`

ripatel-fd (changes requested):
looks good

src/waltz/ip/fd_fib4.c
akhinvasara-jumptrading: there's one in fd_fib4_private.h already, right after the map_slot include

ripatel-fd (approved):

PR #7430, opened 2025-11-25
flamenco, vm: implement increase_cpi_account_info_limit

topointon-jump:
Implementing `increase_cpi_account_info_limit` (https://github.com/solana-foundation/solana-improvement-documents/pull/339), which improves UX/reduces work for programs which make CPI calls that were _themselves_ invoked with more than 64 accounts - they no longer have to de-dup the list of account infos to make a CPI call.

This SIMD also starts charging for serialized account infos and instruction account metas in CPIs.

- [x] Implementation
- [x] Unit tests
- [x] Ledger tests
- [x] Fuzzing


mjain-jump (approved):
looks good, is it fuzz tested?

topointon-jump:
VM CPI fuzzer has been running for a while and not showing any bugs

PR #7389, opened 2025-11-21
flamenco, vm: implement provide_instruction_data_offset_in_vm_r2

topointon-jump:
Implement https://github.com/solana-foundation/solana-improvement-documents/pull/321

- [x] Implementation
- [x] Ledger tests

PR #7459, opened 2025-11-27
flamenco, vm: add bpf loader serialization test fixtures

topointon-jump:
Adds a set of fixtures and a unit test fixture runner for vm input serialization, generated by running the input through the Agave vm input serializers. Compares the output buffer, the vm regions and the region metadata. Includes coverage of `stricter_abi_and_runtime_constraints` and `account_data_direct_mapping`.

Also tests the `provide_instruction_data_offset_in_vm_r2` feature gate.

Adjusted some of the serialization code to make it more amenable for unit testing.


topointon-jump:
This pr depends on, and adds tests for, https://github.com/firedancer-io/firedancer/pull/7389. Needs to be rebased when 7389 is merged.

src/flamenco/progcache/fd_progcache_rec.c
```diff
@@ -89,7 +89,8 @@ fd_progcache_rec_new( void *                          mem,
                    0,
                    FD_FEATURE_ACTIVE( load_slot, features, account_data_direct_mapping ),
                    FD_FEATURE_ACTIVE( load_slot, features, stricter_abi_and_runtime_constraints ),
-                   0 );
+                   0,
```
cmoyes-jump, line 92: nit: could be useful to have in-line comments with what these values are, much like in fd_bpf_loader_program.c

src/flamenco/runtime/program/fd_bpf_loader_program.c
```diff
@@ -402,19 +402,17 @@ fd_bpf_execute( fd_exec_instr_ctx_t *      instr_ctx,
...
+                                                  &instruction_data_offset, &input_sz, input );
   if( FD_UNLIKELY( err ) ) {
     return err;
   }
 
-  if( FD_UNLIKELY( input==NULL ) ) {
-    return FD_EXECUTOR_INSTR_ERR_MISSING_ACC;
-  }
-
   fd_sha256_t _sha[1];
```
topointon-jump, line 417: `fd_exec_instr_ctx_get_index_of_instr_account_in_transaction` throws `FD_EXECUTOR_INSTR_ERR_MISSING_ACC` if the account is missing, so propagating the error is the correct thing to do here and doesn't change the behaviour.

src/flamenco/runtime/program/test_bpf_loader_serialization_fixtures/aligned_data_len_0.json
```diff
@@ -0,0 +1,58 @@
+{
```
mjain-jump: ik this is still WIP but would it be better to have these as test vector fixtures instead ofo having a bunch of JSONs committed to the repo?

topointon-jump: flattened these into a single json, to avoid having a bunch of files. later I agree would be good to have a fuzz harness specifically for the bpf serializer. for now though I just want to get the unit tests in.

src/flamenco/progcache/fd_progcache_rec.c
topointon-jump: I added these but it was quite messy, especially with the IDE's own hints, so I reverted it

cmoyes-jump (approved):

PR #7470, opened 2025-11-28
flamenco, test: ensure no workspace memory leaks during solfuzz execution

topointon-jump:
Add a check at the end of every test execution to ensure that we do not leak memory. Note that we need to compact the fd_alloc instances before doing this, otherwise unused superblocks may be leaked.


cmoyes-jump (approved):

ripatel-fd:
Great catch. Sorry about this.

PR #7471, opened 2025-11-28
metrics_record: add new dev command for streaming metrics into CSV

amass-jump:


ripatel-fd:
Hmm ... We already have `firedancer-dev metrics`. Why not have a `--csv` flag there? 
Not opposed to having a separate command either, but I think grouping very similar functionality under the same high-level key is better user UX.

src/app/shared_dev/commands/metrics_record.c
```diff
@@ -0,0 +1,246 @@
...
+
+  fd_io_buffered_ostream_flush( out );
+  fd_io_buffered_ostream_fini( out );
+
+  fd_topo_leave_workspaces( &config->topo );
+}
+
+action_t fd_action_metrics_record = {
+  .name          = "metrics-record",
+  .description   = "Continuously print a select subset of metrics to STDOUT in CSV format",
```
ripatel-fd, line 242: Graphite format is also very popular

amass-jump:
> Hmm ... We already have firedancer-dev metrics. Why not have a --csv flag there?

This tool differs from the existing `metrics` command in several ways:
- continuous output (until SIGINT) vs one-time dump
- CSV output vs Prometheus text format
- Metrics filtering / "selector" CLI vs all metrics at once

To me this felt like too many things to try to cram into the existing tool and hide behind a cohesive CLI. But I can work on combining it with the existing command if you feel strongly

src/app/shared_dev/commands/metrics_record.c
amass-jump:
```
metric_path value timestamp\n
```

I don't think this is appropriate for this tool, as the idea is to output continuous tabular / data-frame style data that is all sampled at the same interval. So there's no need to include `metric_path` for each row

ripatel-fd (dismissed):

ripatel-fd:
@amass-jump I don't have a strong opinion on this, sgtm

PR #7479, opened 2025-12-01
backtest: replay from pcap

ripatel-fd:
Define a minimal pcapng-based file format to replace RocksDB as a
shred source for `firedancer-dev backtest`.

Add a fd_blockstore2shredcap tool to perform the conversion.

Allow `firedancer-dev` backtest to replay from such pcap files.

### DIFF
--- a/contrib/codeql/nightly/TileUnionMismatch.ql
+++ b/contrib/codeql/nightly/TileUnionMismatch.ql
@@ -35,6 +35,4 @@ where
   not tileName.matches("%_" + t.toString() + "_%") and
   /* net "inheritance" case */
   not t.toString() = "net"
-  /* backtest is called archiver */
-  and not (t.toString() = "archiver" and tileName = "fd_backtest_tile.c")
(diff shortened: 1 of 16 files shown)


src/discof/backtest/shredcap.md
```diff
@@ -0,0 +1,93 @@
...
+
+The format for shredcap v0.1 bank hash v0 metadata is as follows:
+
+- Metadata Type: `0x1` (bank hash v0)
+- Slot number (uint64)
+- Bank hash (32 bytes): Bank hash seen after replaying a slot
+- Data shred count (uint64): The number of data shreds with block data
+  ingested to produce this bank hash
+
+#### Endpoint v0
```
mmcgee-jump, line 71: This is a bit of a weird construction. Is there not a field on the packet itself you can just tag as being a shred?

ripatel-fd:
@mmcgee-jump Great idea re removing endpoint. I'm going to make another version of this format "v0.2" that removes `Endpoint v0` and adds EPB options instead.

PR #7481, opened 2025-12-01
repair: fix null query in peers

emwang-jump:
map_dynamic_remove in the wrong place causing all kinds of problems


src/discof/repair/fd_repair_tile.c
```diff
@@ -608,6 +608,7 @@ after_sign( ctx_t             * ctx,
         this request to the inflights table, pretend we've sent it and
         let the inflight timeout request it down the line.
   */
+  if( FD_UNLIKELY( memcmp( &pending->msg.shred.to, &null_pubkey, 32UL ) == 0 ) ) __asm__("int $3"); /* should REALLY not be possible i think*/
```
ibhatt-jumptrading: log crit also interrupts

ripatel-fd: Don't check this in, it's x86 specific

ripatel-fd (dismissed):
int $3

src/discof/repair/fd_repair_tile.c
emwang-jump: shouldve left wip in the title, not merging this for now to see why we ever had a null pubkey in the first place

ripatel-fd: sorry

PR #7484, opened 2025-12-01
Remove RocksDB from full Firedancer

ripatel-fd:


ripatel-fd:
Need to investigate whether this is a practical approach for large RocksDB instances

PR #7493, opened 2025-12-01
net: validate udp net_len

akhinvasara-jumptrading:
This PR primarily adds upd header net_len validation to the xdp tile - packets with clearly incorrect net_len will be dropped by the network stack. The socket tile handcrafts the headers, so already provides this guarantee. 

A future patch will explicitly define the guarantees made by the network stack and factor out validation logic. 

Minor accompaniments:

- To prevent future code drift around validation logic, this PR also switches repair to use the common fd_ip4_udp_hdr_strip, instead of performing its own validation.
- Send tile stops assuming that quic will always produce fixed size ip headers, using the FD_IP4_GET_LEN macro and adding minor validation.

### DIFF
--- a/src/discof/send/fd_send_tile.c
+++ b/src/discof/send/fd_send_tile.c
@@ -163,9 +163,11 @@ quic_tx_aio_send( void *                    _ctx,
     if( FD_UNLIKELY( batch[ i ].buf_sz<FD_NETMUX_SIG_MIN_HDR_SZ ) ) continue;
     uchar * buf = batch[ i ].buf;
     fd_ip4_hdr_t * ip4_hdr    = fd_type_pun( buf );
-    fd_udp_hdr_t * udp_hdr    = fd_type_pun( buf + sizeof(fd_ip4_hdr_t) );
-    uchar        * payload    = buf + sizeof(fd_ip4_hdr_t) + sizeof(fd_udp_hdr_t);
-    ulong          payload_sz = batch[ i ].buf_sz - sizeof(fd_ip4_hdr_t) - sizeof(fd_udp_hdr_t);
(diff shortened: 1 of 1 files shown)


src/discof/repair/fd_repair_tile.c
```diff
@@ -710,16 +710,12 @@ after_fec( ctx_t      * ctx,
...
-  uchar *              data = (uchar              *)( (ulong)udp + sizeof(fd_udp_hdr_t) );
-  if( FD_UNLIKELY( (ulong)udp+sizeof(fd_udp_hdr_t) > (ulong)eth+sz ) ) return;
-  ulong udp_sz = fd_ushort_bswap( udp->net_len );
-  if( FD_UNLIKELY( udp_sz<sizeof(fd_udp_hdr_t) ) ) return;
-  ulong data_sz = udp_sz-sizeof(fd_udp_hdr_t);
-  if( FD_UNLIKELY( (ulong)data+data_sz > (ulong)eth+sz ) ) return;
-
+  fd_eth_hdr_t * eth; fd_ip4_hdr_t * ip4; fd_udp_hdr_t * udp;
+  uchar * data; ulong data_sz;
+  if( FD_UNLIKELY( !fd_ip4_udp_hdr_strip( ctx->buffer, sz, &data, &data_sz, &eth, &ip4, &udp ) ) ) {
```
mmcgee-jump: This can FD_TEST now, since it's a copied buffer

akhinvasara-jumptrading: cool cool, was originally intending to leave the behavior unchanged, but ill switch it over to FD_TEST rn

PR #7497, opened 2025-12-01
fuzz, flamenco: fix stack smashing in vote program

mjain-jump:


src/flamenco/runtime/program/fd_vote_program.c
```diff
@@ -970,7 +970,7 @@ check_and_filter_proposed_vote_state( fd_vote_state_t *           vote_state,
   /* Index into the slot_hashes, starting at the oldest known slot hash */
   // https://github.com/anza-xyz/agave/blob/v2.0.1/programs/vote/src/vote_state/mod.rs#L264
   ulong slot_hashes_index = deq_fd_slot_hash_t_cnt( slot_hashes );
-  ulong proposed_lockouts_indexes_to_filter[ MAX_LOCKOUT_HISTORY ];
+  ulong proposed_lockouts_indexes_to_filter[ lockouts_len ];
```
ripatel-fd: We should crash if we encounter a vote state with lockouts_len>MAX_LOCKOUT_HISTORY from a trust source (accounts DB) or reject the input if it's from an untrusted place. We should not do an unbounded stack allocation

mjain-jump: lockouts_len is provided in the instr data, which is bounded by the MTU size. So theoretically in production it can be > MAX_LOCKOUT_HISTORY

mjain-jump: i made a constant bound for it

PR #7520, opened 2025-12-02
backtest: don't use sudo to run backtest

ibhatt-jumptrading:


kbhargava-jump:
backtest runs will still run as root, log will show:
```
src/app/shared_dev/boot/fd_dev_boot.c(150)[fd_dev_main]: insufficient permissions to execute command `backtest`, rerunning as root
```

PR #7529, opened 2025-12-02
vm: fix epoch rewards padding

intrigus-lgtm:
The epoch rewards struct has 16 byte alignment so
it has 15 bytes of padding at the end.
(81 bytes that are actually used, but 96 bytes in total)
(This issue was also rediscovered by @Nicola-Osec after I forgot to PR this)


ripatel-fd (approved):

ripatel-fd:
@mjain-jump The syscall fuzzer should have trivially caught this

PR #7535, opened 2025-12-03
Intrigus/fix/ss http resolve archive misc

intrigus-lgtm:


src/discof/restore/utils/fd_ssresolve.c
```diff
@@ -272,6 +272,11 @@ fd_ssresolve_parse_redirect( fd_ssresolve_t *        ssresolve,
     }
   }
 
+  if( FD_UNLIKELY( !location ) ) {
```
ripatel-fd:
```suggestion
  if( FD_UNLIKELY( !location_len ) ) {
```

intrigus-lgtm: Is there a specific reason you prefer this?
When location is not NULL, `location_len >= 1` should be guaranteed by line 264.
But of course happy to change this.

intrigus-lgtm:
> Thank you very much for the fixes, these look great. Would you mind adding unit tests reproducing the bugs you've fixed though? If you don't have time to do this, I'm happy to take it over. Just blocking for now to keep myself honest because if I merge this, I'll just forget to add tests later.

A few of those functions are in `static` functions.
How should I test those?
I could either make them non-static or test them indirectly.

src/discof/restore/utils/fd_ssresolve.c
ripatel-fd: We typically always do "slice empty?" type checks on the length field. It's a more complete solution because often slices with an invalid (but not NULL) base pointer and a zero size length are considered valid

intrigus-lgtm: Makes sense!
Should I align https://github.com/firedancer-io/firedancer/blob/8997d39c5f3e0d0b8cf1ac845804ab260fbac4ff/src/discof/restore/utils/fd_sshttp.c#L451-L455 to this pattern as well?

ripatel-fd: yes please

intrigus-lgtm: Done.

PR #7548, opened 2025-12-04
gui: fix est slot duration

jherrera-jump:


src/disco/gui/fd_gui.c
```diff
@@ -2787,6 +2787,28 @@ fd_gui_handle_replay_update( fd_gui_t *                gui,
   slot_complete_event->timestamp = slot_completed->completed_time;
   slot_complete_event->shred_idx = USHORT_MAX;
   slot_complete_event->slot      = slot->slot;
+
+  ulong duration_sum = 0UL;
+  ulong slot_cnt = 0UL;
+
+  /* If we've just caught up we should truncate our slot history to avoid including catch-up slots */
```
mmcgee-jump: Maybe a cleaner way to do this basing it off turbine slot

PR #7553, opened 2025-12-04
flamenco, runtime, vm: clean up bounds on the number of instruction accounts

topointon-jump:
Corrects the value of FD_INSTR_ACCT_MAX to a larger bound derived from the transaction MTU. Without this, it is possible to cause a conformance divergence.

Credit to @nick0ve for reporting this in https://github.com/firedancer-io/auditor-internal/issues/348.  

We were also using FD_INSTR_ACCT_MAX in several places where we shouldn't have been. This PR corrects that, so that we only use this larger bound where it's actually necessary.

Thankfully, https://github.com/solana-foundation/solana-improvement-documents/pull/406 fixes this by limiting the number of instruction accounts to 255 at the transaction sanitization level. Once that SIMD is implemented and activated we can use the same 255 limit everywhere.

Increases the size of `fd_runtime_t` by 470KB.


src/flamenco/runtime/info/fd_instr_info.c
```diff
@@ -33,18 +33,14 @@ fd_instr_info_init_from_txn_instr( fd_instr_info_t *      instr,
...
   instr->program_id = txn_instr->program_id;
-
-  /* See note in fd_instr_info.h.  TLDR: capping this value at 256
-     should have literally 0 effect on program execution, down to the
-     error codes.  This is purely for the sake of not increasing the
-     overall memory footprint of the transaction context.  If this
-     change causes issues, we may need to increase the array sizes in
-     the instr info. */
-  instr->acct_cnt = fd_ushort_min( txn_instr->acct_cnt, FD_INSTR_ACCT_MAX );
+  instr->acct_cnt = txn_instr->acct_cnt;
```
mjain-jump: nit: alignment of = sign

src/flamenco/runtime/program/fd_bpf_loader_serialization.c
```diff
@@ -266,8 +266,8 @@ fd_bpf_loader_input_serialize_aligned( fd_exec_instr_ctx_t *     ctx,
                                        ulong *                   instr_data_offset ) {
   fd_pubkey_t * txn_accs = ctx->txn_out->accounts.account_keys;
 
-  uchar  acc_idx_seen[ FD_INSTR_ACCT_MAX ] = {0};
-  ushort dup_acc_idx[ FD_INSTR_ACCT_MAX ]  = {0};
+  uchar  acc_idx_seen[ FD_TXN_ACCT_ADDR_MAX ] = {0};
+  ushort dup_acc_idx[ FD_TXN_ACCT_ADDR_MAX ]  = {0};
```
mjain-jump, line 272: you should probably add a comment mentioning something about transaction sanitization / loading ensuring that the instr account indices are always in range here

src/flamenco/runtime/program/fd_native_cpi.c
```diff
@@ -28,9 +28,9 @@ fd_native_cpi_native_invoke( fd_exec_instr_ctx_t *             ctx,
     instr_info->program_id = (uchar)program_id;
   }
 
-  fd_pubkey_t instr_acct_keys[ FD_INSTR_ACCT_MAX ];
-  uchar       acc_idx_seen[ FD_INSTR_ACCT_MAX ];
-  memset( acc_idx_seen, 0, FD_INSTR_ACCT_MAX );
+  fd_pubkey_t instr_acct_keys[ FD_VM_CPI_MAX_INSTRUCTION_ACCOUNTS ];
+  uchar       acc_idx_seen[ FD_TXN_ACCT_ADDR_MAX ];
```
mjain-jump:
```suggestion
  uchar       acc_idx_seen[ FD_TXN_ACCT_ADDR_MAX ] = { 0 };
```

src/flamenco/runtime/fd_runtime_const.h
```diff
@@ -147,8 +147,63 @@ static const fd_cluster_version_t FD_RUNTIME_CLUSTER_VERSION = {
...
+   Therefore, the maximum number of instruction accounts is:
+     (MTU - fixed overhead) / (size of instruction account)
+   = (MTU
+       - signature count (1 byte, value=1)
+       - signature (64 bytes)
+       - signature count in header (1 byte)
+       - readonly signed count (1 byte)
+       - readonly unsigned count (1 byte)
+       - account count (1 byte, compact-u16 value=2)
+       - 2 account addresses (32 + 32 bytes)
```
mjain-jump: technically it could be 1 account if the program id is the fee payer address (which is obviously invalid and will fail out very early), but maybe clarify this

src/flamenco/vm/syscall/fd_vm_syscall_cpi.c
```diff
@@ -58,17 +58,25 @@ int
...
+                           fd_pubkey_t const        instr_acct_keys[ FD_VM_CPI_MAX_INSTRUCTION_ACCOUNTS ],
+                           fd_instruction_account_t instruction_accounts[ FD_VM_CPI_MAX_INSTRUCTION_ACCOUNTS ],
                            ulong *                  instruction_accounts_cnt,
                            fd_pubkey_t const *      signers,
                            ulong                    signers_cnt ) {
 
   /* De-duplicate the instruction accounts, using the same logic as Solana */
   ulong deduplicated_instruction_accounts_cnt = 0;
-  fd_instruction_account_t deduplicated_instruction_accounts[256] = {0};
+  fd_instruction_account_t deduplicated_instruction_accounts[ FD_VM_CPI_MAX_INSTRUCTION_ACCOUNTS ] = {0};
```
mjain-jump, line 69: nit: align `=` signs and variables

mjain-jump (dismissed):

src/flamenco/runtime/fd_runtime_const.h
```diff
@@ -105,7 +105,7 @@ static const fd_cluster_version_t FD_RUNTIME_CLUSTER_VERSION = {
 /* The bpf loader's serialization footprint is bounded in the worst case
    by 64 unique writable accounts which are each 10MiB in size (bounded
```
ibhatt-jumptrading, line 106: comments shouldn't exceed 80 chars

src/flamenco/runtime/info/fd_instr_info.c
```diff
@@ -33,18 +33,14 @@ fd_instr_info_init_from_txn_instr( fd_instr_info_t *      instr,
...
-
-  /* See note in fd_instr_info.h.  TLDR: capping this value at 256
-     should have literally 0 effect on program execution, down to the
-     error codes.  This is purely for the sake of not increasing the
-     overall memory footprint of the transaction context.  If this
-     change causes issues, we may need to increase the array sizes in
-     the instr info. */
-  instr->acct_cnt = fd_ushort_min( txn_instr->acct_cnt, FD_INSTR_ACCT_MAX );
+  instr->acct_cnt   = txn_instr->acct_cnt;
+  if( FD_UNLIKELY( instr->acct_cnt > FD_INSTR_ACCT_MAX ) ) {
```
ibhatt-jumptrading, line 38: i don't think this should be a crit

ibhatt-jumptrading: nvm

ripatel-fd (changes requested):
The fix looks good, but can we add a manual test passing a worst-case transaction into the runtime? This makes it easier to verify the behavior in the future without having to rely on external fuzzing infra.

topointon-jump:
@ripatel-fd done! will migrate it over to the new unit test framework when we have it, for now I copy/pasted the other test setup code

PR #7557, opened 2025-12-04
net: factor out header validation

akhinvasara-jumptrading:
This PR centralizes network header validation. The XDP tile guarantees that packets sent downstream will have passed validation.

For ease of review/reversability of desired, the first commit performs the refactor. The second commit changes the reported xdp metrics to take advantage of the new error codes.

The RX perf impact of this change is slightly positive (tested using Pktgen-DPDK from rd-merenguel1 (mlx5) to rd-chachal1 on (mlx5): 
before:  8.617 Mpps
After: 8.713 Mpps


ripatel-fd:
`8.713e+06 Mpps` that's 8 trillion wow :)

src/util/net/fd_net_headers.h
ripatel-fd, line 1: Can we keep these in src/disco/net? These are rather specific to Firedancer and the util/net bits aim to be more generic (shared across other projects)

src/util/net/fd_udp.h
```diff
@@ -82,6 +82,29 @@ fd_ip4_udp_check( uint                 ip4_saddr,
...
 }
 
+/* fd_udp_hdr_validate validates a UDP header.
+
+   hdr points to the UDP header.
+   udp_sz is the number of bytes from beginning of UDP header to end of packet.
+
+   Returns FD_NET_SUCCESS if valid, or FD_NET_ERR_INVAL_UDP_HDR if the header is invalid:
+   - net_len and/or udp_sz is impossibly small
+   - net_len is larger than udp_sz */
```
ripatel-fd, line 92: This doesn't define what makes a UDP header "valid" and it also doesn't look at the UDP checksum. I'd call this `fd_udp_hdr_bounds_check` or similar, and move it out of util/net to disco/net

src/disco/metrics/metrics.xml
```diff
@@ -54,7 +54,8 @@ metric introduced.
 <tile name="net">
     <counter name="RxPktCnt" summary="Packet receive count." />
     <counter name="RxBytesTotal" summary="Total number of bytes received (including Ethernet header)." />
-    <counter name="RxUnderszCnt" summary="Number of incoming packets dropped due to being too small." />
+    <counter name="RxInvalidIp4Cnt" summary="Number of incoming packets dropped due to invalid IP4 header." />
```
ripatel-fd, line 57: IPv4*

akhinvasara-jumptrading:
Closed in favor of https://github.com/firedancer-io/firedancer/pull/7647 given git write access changes.

PR #7565, opened 2025-12-05
genesis: replacing fd types genesis type with handrolled one

ibhatt-jumptrading:


ripatel-fd (changes requested):
Since we accept genesis blobs from remote peers this needs some more security measures. Could you add a `fuzz_genesis_parse.c` with a generic libfuzzer harness? You can check the gossip message parser fuzzer for a scaffold and the in-tree docs for libFuzzer build instructions.

src/discof/genesis/fd_genesi_tile.c
```diff
@@ -283,19 +272,10 @@ after_credit( fd_genesi_tile_t *  ctx,
     }
 
     FD_TEST( !ctx->bootstrap );
-
-    fd_bincode_decode_ctx_t decode_ctx = {
-      .data    = decompressed+512UL,
-      .dataend = decompressed+512UL+fd_tar_meta_get_size( meta ),
-    };
+    ulong size = 512UL+fd_tar_meta_get_size( meta );
```
ripatel-fd:
```suggestion
    ulong size = fd_tar_meta_get_size( meta );
```

This looks wrong. The previous code added 512 to both the start and end pointers to shift the access away from the tar block. This just adds 512 bytes out of bounds to the parser

src/flamenco/runtime/fd_genesis_parse.h
```diff
@@ -0,0 +1,87 @@
...
+#define FD_GENESIS_BUILTIN_MAX_COUNT (16UL)
+#define FD_GENESIS_MAX_MESSAGE_SIZE  (1024UL*1024UL*10UL) /* 10MiB */
+
+#define FD_GENESIS_TYPE_TESTNET     (0)
+#define FD_GENESIS_TYPE_MAINNET     (1)
+#define FD_GENESIS_TYPE_DEVNET      (2)
+#define FD_GENESIS_TYPE_DEVELOPMENT (3)
+
+struct fd_genesis_account {
+  uchar             pubkey[32];
```
ripatel-fd, line 20: I suggest using `fd_pubkey_t` here.

src/flamenco/runtime/fd_genesis_parse.h
```diff
@@ -0,0 +1,87 @@
...
+  fd_account_meta_t meta;
+  uchar             data[];
```
ripatel-fd, line 22: Not opposed to VLAs but I'm surprised it compiles under strict checks

src/flamenco/runtime/fd_genesis_parse.h
```diff
@@ -0,0 +1,87 @@
...
+    uchar warmup;
+    ulong first_normal_epoch;
+    ulong first_normal_slot;
+  } epoch_schedule;
+
+  ulong accounts_len;
+  uint  accounts_off[ FD_GENESIS_ACCOUNT_MAX_COUNT ];
+  ulong builtin_len;
+  uint  builtin_off[ FD_GENESIS_BUILTIN_MAX_COUNT ];
+
```
ripatel-fd, line 76: maybe add a comment `/* variable size data follows */`

src/flamenco/runtime/fd_genesis_parse.h
```diff
@@ -0,0 +1,87 @@
...
+};
+typedef struct fd_genesis fd_genesis_t;
+
+FD_PROTOTYPES_BEGIN
+
+fd_genesis_t *
+fd_genesis_parse( uchar const * mem_in,
+                  ulong         sz_in,
+                  ulong *       sz_out,
+                  uchar *       genesis_out );
```
ripatel-fd: Couple polishing things:
- This wraps a constructor (of the fd_genesis_t class). Constructors typically take the input memory region as the first argument and return the newly created object. So, `genesis_out` should be the first argument.
- `uchar * genesis_out` implies to the reader this has 1 byte alignment. I would rename this to `void * genesis_mem` and document the alignment requirement. 
- Instead of a `sz_out` pointer, it might be cleaner to store the size in the genesis object itself. 
- How about renaming the input to `uchar const * bin` and `ulong bin_sz`?
- Add a short comment that this is a bincode parser and add a ref link to the corresponding Agave type.
- What happens if parsing fails? Add documentation
- On first sight, I thought that the caller initializes `*sz_out` with the buffer size of `genesis_out`, and then on return `*sz_out` contains a different value. This is a pretty typical pattern for "placement new" type constructors and serializers. But then I noticed that `genesis_out` is assumed to have some constant memory footprint. Please document

src/flamenco/runtime/fd_genesis_parse.c
```diff
@@ -0,0 +1,156 @@
...
+  CHECK_LEFT( 8U ); genesis->builtin_len  = FD_LOAD( ulong, CURSOR ); INC( 8U );
+  if( FD_UNLIKELY( genesis->builtin_len>FD_GENESIS_BUILTIN_MAX_COUNT ) ) {
+    FD_LOG_WARNING(( "genesis builtin length %lu exceeds supported max count %lu", genesis->builtin_len, FD_GENESIS_BUILTIN_MAX_COUNT ));
+    return NULL;
+  }
+  for( ulong i=0UL; i<genesis->builtin_len; i++ ) {
+    fd_genesis_account_t * account = FD_SCRATCH_ALLOC_APPEND( l, alignof(fd_genesis_account_t), sizeof(fd_genesis_account_t) );
+    genesis->builtin_off[ i ] = (uint)( (ulong)account - (ulong)genesis );
+    if( FD_UNLIKELY( genesis->builtin_off[ i ] + sizeof(fd_genesis_account_t) > FD_GENESIS_MAX_MESSAGE_SIZE ) ) {
+      FD_LOG_WARNING(( "genesis builtin offset %lu exceeds supported max size %lu", genesis->builtin_off[ i ] + sizeof(fd_genesis_account_t), FD_GENESIS_MAX_MESSAGE_SIZE ));
```
ripatel-fd, line 78: Maybe a more operator friendly warning would be `cannot decode genesis blob: genesis parameters exceed hardcoded buffer size`. I guess it doesn't matter because it never realistically gets hit

src/flamenco/runtime/fd_genesis_parse.h
```diff
@@ -0,0 +1,87 @@
...
+#define FD_GENESIS_TYPE_DEVELOPMENT (3)
+
+struct fd_genesis_account {
+  uchar             pubkey[32];
+  fd_account_meta_t meta;
+  uchar             data[];
+};
+typedef struct fd_genesis_account fd_genesis_account_t;
+
+struct fd_genesis {
```
ripatel-fd, line 26: This is a rather aggressive change request, but why not make `fd_genesis` a descriptor, like some of the gossip parsers and `txn_t` are? 

Instead of containing a copy of accounts, you could make the genesis descriptor object point into the original serialized blob. That way `fd_genesis` doesn't have to be variable sized (it just contains some arrays of account/builtin pointers).

ibhatt-jumptrading: we pass fd_genesis across tiles so i figured it would be cleaner to just separate the type out. it's also more consistent with how we layout the snapshot manifest message

src/flamenco/runtime/fd_genesis_parse.h
ibhatt-jumptrading: this matches the pattern for snapshot manifest better imo

src/flamenco/runtime/fd_genesis_parse.h
ibhatt-jumptrading: done

src/discof/genesis/fd_genesi_tile.c
ibhatt-jumptrading: yes it is wrong, fixed

src/flamenco/runtime/fd_genesis_parse.h
ripatel-fd: sounds good to me

src/flamenco/runtime/fd_genesis_parse.h
ripatel-fd: OK sounds good

src/discof/replay/fd_replay_tile.c
```diff
@@ -1263,7 +1271,7 @@ boot_genesis( fd_replay_tile_t *  ctx,
   // TODO: Do not pass the fd_types type between tiles, it have offsets
   // that are unsafe and can't be validated as being in-bounds.  Need to
   // pass an actual owned genesis type.
```
ripatel-fd: You can remove this comment

ripatel-fd (dismissed):
Looks great

alpeng-jump (approved):

PR #7570, opened 2025-12-09
snapshots: fix pipeline state machine edge cases

amass-jump:


src/discof/restore/fd_snapdc_tile.c
```diff
@@ -118,27 +118,26 @@ handle_control_frag( fd_snapdc_tile_t *  ctx,
...
     case FD_SNAPSHOT_MSG_CTRL_NEXT:
     case FD_SNAPSHOT_MSG_CTRL_DONE:
-      FD_TEST( ctx->state==FD_SNAPSHOT_STATE_PROCESSING ||
-               ctx->state==FD_SNAPSHOT_STATE_ERROR );
-      if( FD_UNLIKELY( ctx->is_zstd && ctx->dirty ) ) {
-        FD_LOG_WARNING(( "encountered end-of-file in the middle of a compressed frame" ));
-        ctx->state = FD_SNAPSHOT_STATE_ERROR;
-        fd_stem_publish( stem, 0UL, FD_SNAPSHOT_MSG_CTRL_ERROR, 0UL, 0UL, 0UL, 0UL, 0UL );
-        return;
-      }
```
amass-jump, line 140: This check could never trigger before (we always reset dirty to 0 above), but also returning here deadlocks the pipeline

src/discof/restore/fd_snapdc_tile.c
```diff
@@ -118,27 +118,26 @@ handle_control_frag( fd_snapdc_tile_t *  ctx,
...
       ctx->is_zstd = !!msg->zstd;
+      ctx->dirty = 0;
       ctx->in.frag_pos = 0UL;
       ctx->metrics.incremental.compressed_bytes_read      = 0UL;
       ctx->metrics.incremental.decompressed_bytes_written = 0UL;
       break;
     }
     case FD_SNAPSHOT_MSG_CTRL_FAIL:
-      FD_TEST( ctx->state==FD_SNAPSHOT_STATE_PROCESSING ||
-               ctx->state==FD_SNAPSHOT_STATE_ERROR );
```
amass-jump, line 128: It's also possible for tiles to receive the FAIL control message in the IDLE state. This is the cause of several of the FD_TEST assertions we've seen.

This happens when snapct sends out a DONE, and an early tile immediately handles it and goes to IDLE. But a later tile may fail and generate ERROR, which causes snapct to send out a FAIL control message which is looped back through the pipeline

src/discof/restore/fd_snapin_tile.c
```diff
@@ -161,6 +161,7 @@ verify_slot_deltas_with_bank_slot( fd_snapin_tile_t * ctx,
 static void
 transition_malformed( fd_snapin_tile_t *  ctx,
                       fd_stem_context_t * stem ) {
+  if( FD_UNLIKELY( ctx->state==FD_SNAPSHOT_STATE_ERROR ) ) return;
   ctx->state = FD_SNAPSHOT_STATE_ERROR;
   fd_stem_publish( stem, ctx->out_ct_idx, FD_SNAPSHOT_MSG_CTRL_ERROR, 0UL, 0UL, 0UL, 0UL, 0UL );
 }
```
amass-jump, line 167: Not a bug, but no need to generate extra ERROR messages when we are already in that state

src/discof/restore/fd_snapls_tile.c
```diff
@@ -313,34 +312,6 @@ returnable_frag( fd_snapls_tile_t *  ctx,
...
 }
 
-static void
-after_credit( fd_snapls_tile_t *  ctx,
-              fd_stem_context_t *  stem,
-              int *                opt_poll_in FD_PARAM_UNUSED,
-              int *                charge_busy FD_PARAM_UNUSED ) {
-  if( FD_UNLIKELY( ctx->hash_accum.received_lthashes==ctx->num_hash_tiles && ctx->hash_accum.awaiting_ack ) ) {
-    fd_lthash_sub( &ctx->hash_accum.calculated_lthash, &ctx->running_lthash );
-    if( FD_UNLIKELY( memcmp( &ctx->hash_accum.expected_lthash, &ctx->hash_accum.calculated_lthash, sizeof(fd_lthash_value_t) ) ) ) {
```
amass-jump, line 323: This doesn't need to be asynchronous, because receiving all the NEXT/DONE "acks" from the snapla's implies that we must have already seen the HASH_RESULT messages from each of them. So as soon as we get the last ack we can check the lthash for correctness.

src/discof/restore/utils/fd_ssctrl.h
```diff
@@ -139,4 +148,37 @@ typedef struct fd_snapshot_full_account fd_snapshot_full_account_t;
 
 #define FD_SNAPSHOT_MAX_SNAPLA_TILES (8UL)
 
+/* fd_ssctrl_test_maybe_error is only used in testing and randomly generates
+   error control messages from tiles in the snapshots pipeline in appropriate
+   states. */
+
+#define FD_SNAPSHOT_TEST_ERROR_NANOS (0L)
```
cali-jumptrading: Think this test logic should be done in a proper unit test or maybe as a sub-command of snapshot-load.

src/discof/restore/fd_snapdc_tile.c
```diff
@@ -161,6 +160,8 @@ handle_data_frag( fd_snapdc_tile_t *  ctx,
                   fd_stem_context_t * stem,
                   ulong               chunk,
                   ulong               sz ) {
+  if( FD_UNLIKELY( fd_ssctrl_test_maybe_error( &ctx->state, stem, 0UL ) ) ) return 0;
```
cali-jumptrading: This looks confusing without additional context. Apparently nothing happens in this function when the define FD_SNAPSHOT_TEST_ERROR_NANOS is not set.

If you truly want something like this I think it would be better to guard this call with a defined macro

src/discof/restore/fd_snapin_tile.c
```diff
@@ -617,7 +619,7 @@ handle_control_frag( fd_snapin_tile_t *  ctx,
                ctx->state==FD_SNAPSHOT_STATE_ERROR );
       if( FD_UNLIKELY( ctx->state!=FD_SNAPSHOT_STATE_FINISHING ) ) {
         transition_malformed( ctx, stem );
-        return;
+        break;
```
cali-jumptrading, line 622: why the break here instead of return? Do we still need to forward the control message if we are generating an error?

amass-jump: Yes, every control message needs to be forwarded down the pipeline immediately, or deferred for later (such as with snapls). But we can't outright drop a control message or the pipeline will be locked on waiting for that message to be flushed forever (each control message is only generated a single time and we do not generate new control messages until it's been flushed).

src/discof/restore/utils/fd_ssctrl.h
amass-jump: I agree, I am not super happy with how I added the testing logic here. I will revert it for now and just merge the fixes

PR #7571, opened 2025-12-10
runtime: port finalize_account to accdb API

ripatel-fd:


src/flamenco/runtime/fd_slot_acc.h
ripatel-fd, line 1: did not mean to commit this

mmcgee-jump (approved):

PR #7576, opened 2025-12-10
fix: oob access on fd_log_private_shared_lock

cavemanloverboy:
image


cavemanloverboy:
if it's indeed intentional can we make the static array and the local lock variable size 2?

ripatel-fd:
> if it's indeed intentional can we make the static array and the local lock variable size 2?

@cavemanloverboy Yes, the current code is just UB ... Thanks for flagging

cavemanloverboy:
> > if it's indeed intentional can we make the static array and the local lock variable size 2?
> 
> @cavemanloverboy Yes, the current code is just UB ... Thanks for flagging

do you want the current PR or do you want me to expand the array size?

ripatel-fd:
We are going to fix this, thank you for reporting.

mmcgee-jump:
The code as written is correct, although it's confusing. The pointer is replaced on boot with a 4k mmaped page. Index 1 and 0 serve different purposes, so this change is not correct.

PR #7577, opened 2025-12-11
fix(hfokrk): lamport pct ulong overflow

lidatong:


ripatel-fd (approved):
Could we add this to test_hfork?

PR #7583, opened 2025-12-15
replay: adding better metrics

ibhatt-jumptrading:


src/discof/replay/fd_replay_tile.c
```diff
@@ -1600,8 +1619,20 @@ replay( fd_replay_tile_t *  ctx,
...
-  if( FD_UNLIKELY( !fd_sched_can_ingest( ctx->sched, 1UL ) ) ) return 0;
-  if( FD_UNLIKELY( (fec = fd_reasm_peek( ctx->reasm ))==NULL ) ) return 0;
+  if( FD_UNLIKELY( !fd_sched_can_ingest( ctx->sched, 1UL ) ) ) {
+    ctx->metrics.sched_full++;
+    return 0;
+  }
+
+  if( FD_UNLIKELY( (fec = fd_reasm_peek( ctx->reasm ))==NULL ) ) {
+    ctx->metrics.reasm_empty++;
+    ctx->metrics.reasm_latest_slot    = ULONG_MAX;
```
mmcgee-jump: Initialize these to zero.

ibhatt-jumptrading: these get zero'd out in unrpriv init already

src/disco/metrics/metrics.xml
```diff
@@ -794,9 +794,18 @@ metric introduced.
...
+  <gauge name="ReasmFree" summary="The number of free FEC sets in the reassembly queue" />
+  <gauge name="ReasmLatestSlot" summary="Slot of the latest FEC set in the reassembly queue that can be replayed" />
+  <gauge name="ReasmLatestFecIdx" summary="FEC set index of the latest FEC set in the reassembly queue that can be replayed" />
+
   <counter name="SlotsTotal" summary="Count of slots replayed successfully" />
   <counter name="TransactionsTotal" summary="Count of transactions processed overall on the current fork" />
 
+  <counter name="SchedFull" summary="Times where sched is full and a FEC set can't be processed" />
+  <counter name="ReasmEmpty" summary="Times where reasm is empty and a FEC set can't be processed" />
+  <counter name="LeaderBidWait" summary="Times where replay is blocked by the the PoH tile not sending an end of leader message" />
```
yufeng-jump: the the

src/discof/replay/fd_replay_tile.c
```diff
@@ -1600,8 +1619,20 @@ replay( fd_replay_tile_t *  ctx,
...
+    ctx->metrics.reasm_latest_fec_idx = ULONG_MAX;
```
yufeng-jump: We might want these to monotonically increase, rather than resetting to ULONG_MAX.  The way metrics make their way into the database, we might miss some high watermark values if we reset?  I don't think we lose any information if we keep the reasm_latest_* values untouched in this branch, since reasm_empty will go up and that'll indicate an empty reasm.

yufeng-jump (approved):

src/discof/replay/fd_replay_tile.c
ibhatt-jumptrading: yeah agreed

PR #7585, opened 2025-12-15
flamenco, runtime: implement static_instruction_limit

topointon-jump:
Implement https://github.com/solana-foundation/solana-improvement-documents/pull/160

- [x] Implementation
- [x] Unit tests
- [x] Ledger

Also drive-by fix to make `fd_executor_verify_transaction` take a const bank.

### DIFF
--- a/src/flamenco/runtime/tests/run_backtest_all.sh
+++ b/src/flamenco/runtime/tests/run_backtest_all.sh
@@ -95,3 +95,4 @@ src/flamenco/runtime/tests/run_ledger_backtest.sh -l mainnet-376969880-r2 -y 1 -
 src/flamenco/runtime/tests/run_ledger_backtest.sh -l mainnet-376969880-simd-339 -y 1 -m 2000000 -e 376969900 -o H6iVbVaDZgDphcPbcZwc5LoznMPWQfnJ1AM7L1xzqvt5
 src/flamenco/runtime/tests/run_ledger_backtest.sh -l breakpoint-385786458 -y 1 -m 2000000 -e 385786458
 src/flamenco/runtime/tests/run_ledger_backtest.sh -l localnet-deprecate-rent-exemption-threshold -y 1 -m 1000 -e 260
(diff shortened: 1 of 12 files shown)


src/flamenco/runtime/tests/run_backtest_all.sh
```diff
@@ -94,3 +94,4 @@ src/flamenco/runtime/tests/run_ledger_backtest.sh -l devnet-422969842 -y 1 -m 20
 src/flamenco/runtime/tests/run_ledger_backtest.sh -l mainnet-376969880-r2 -y 1 -m 2000000 -e 376969900 -o 5xXZc66h4UdB6Yq7FzdBxBiRAFMMScMLwHxk2QZDaNZL
 src/flamenco/runtime/tests/run_ledger_backtest.sh -l mainnet-376969880-simd-339 -y 1 -m 2000000 -e 376969900 -o H6iVbVaDZgDphcPbcZwc5LoznMPWQfnJ1AM7L1xzqvt5
 src/flamenco/runtime/tests/run_ledger_backtest.sh -l breakpoint-385786458 -y 1 -m 2000000 -e 385786458
+src/flamenco/runtime/tests/run_ledger_backtest.sh -l localnet-static-instruction-limit -y 1 -m 1000 -e 191
```
topointon-jump, line 98: This ledger passes through an epoch boundary in which `static_instruction_limit` is activated, and then tries to execute a transaction with 64 instructions which succeeds. Note that the script to create this ledger also sent transactions with 65 instructions, but because these were filtered out by Agave's sanitisation code (which is what this feature changes) they don't appear in the ledger.

ibhatt-jumptrading:
this doesn't fully cover the case where a transaction is worst case sized for > 64 instructions because it will violate limits in the transaction parser (fd_txn_parse). The correct thing to do here imo is to change txn parse to always fail if there are >64 top level instructions.

topointon-jump:
^Discussed offline, will pr in a cleanup pr in after the feature flag is activated everywhere (~3 weeks)

PR #7587, opened 2025-12-15
fd_types cleanup

ripatel-fd:


src/flamenco/rewards/fd_rewards.h
```diff
@@ -7,6 +7,25 @@
 #include "../stakes/fd_stake_delegations.h"
 #include "../stakes/fd_vote_states.h"
 
+struct fd_prev_epoch_inflation_rewards {
+  ulong validator_rewards;
```
ibhatt-jumptrading: can you just align these

mmcgee-jump (dismissed):

topointon-jump (approved):

src/flamenco/rewards/fd_rewards.h
ripatel-fd: Yep, thanks 👍

PR #7597, opened 2025-12-17
simd-0332: reduce chacha rounds for turbine from 20 to 8

0x0ece:
- Added a few missing functions for chacha8
- Made wsample configurable with chacha8 vs chacha20 (sampler->rng_algo)
- Made sdest use chacha8 vs chacha20 based on staked wsample. Note: leader schedule (that also uses wsample) doesn’t need chacha8
- Added frank feature gate and cleaned up existing ones
- Put all together in shred tile (note: works for frank, doesn’t work yet for full client)
- Added a conformance test with Rust code to generate the C input - found a bug 🎉 

Agave: https://github.com/anza-xyz/agave/pull/7353

### DIFF
--- a/src/ballet/wsample/fd_wsample.c
+++ b/src/ballet/wsample/fd_wsample.c
@@ -389,18 +389,12 @@ fd_wsample_delete( void * shmem  ) {
   return shmem;
 }
 
-
-
-fd_chacha_rng_t * fd_wsample_get_rng( fd_wsample_t * sampler ) { return sampler->rng; }
-
-
-/* TODO: Should this function exist at all? */
 void
-fd_wsample_seed_rng( fd_chacha_rng_t * rng,
-                     uchar seed[static 32] ) {
-  fd_chacha20_rng_init( rng, seed );
+fd_wsample_seed_rng( fd_wsample_t * sampler,
+                     uchar          seed[ 32 ],
+                     int            use_chacha8 ) {
+  fd_chacha_rng_init( sampler->rng, seed, use_chacha8 ? FD_CHACHA_RNG_ALGO_CHACHA8 : FD_CHACHA_RNG_ALGO_CHACHA20 );
 }
-
 
 fd_wsample_t *
 fd_wsample_restore_all( fd_wsample_t * sampler ) {
@@ -597,7 +591,7 @@ fd_wsample_remove_idx( fd_wsample_t * sampler,
    but operations with mask registers are frustratingly slow.  Instead,
    we first define x'=x+1, so then v0<=x is equivalent to v0-x'<0, or
    whether v0-x' has the high bit set.  Because of
-   fd_chacha20_rng_ulong_roll's contract, we know x<ULONG_MAX, so forming
+   fd_chacha_rng_ulong_roll's contract, we know x<ULONG_MAX, so forming
    x' is safe. We then use a _mm512_permutexvar_epi8 to essentially
    broadcast the high byte from each of the ulongs (really we just need
    the high bit) to the whole vector.  A popcnt gives us our base
@@ -734,7 +728,7 @@ fd_wsample_sample_and_remove_many( fd_wsample_t * sampler,
   for( ulong i=0UL; i<cnt; i++ ) {
     if( FD_UNLIKELY( !sampler->unremoved_weight ) ) { idxs[ i ] = FD_WSAMPLE_EMPTY;         continue; }
     if( FD_UNLIKELY(  sampler->poisoned_mode    ) ) { idxs[ i ] = FD_WSAMPLE_INDETERMINATE; continue; }
-    ulong unif = fd_chacha20_rng_ulong_roll( sampler->rng, sampler->unremoved_weight+sampler->poisoned_weight );
+    ulong unif = fd_chacha_rng_ulong_roll( sampler->rng, sampler->unremoved_weight+sampler->poisoned_weight );
     if( FD_UNLIKELY( unif>=sampler->unremoved_weight ) ) {
       idxs[ i ] = FD_WSAMPLE_INDETERMINATE;
       sampler->poisoned_mode = 1;
@@ -771,7 +765,7 @@ ulong
 fd_wsample_sample( fd_wsample_t * sampler ) {
   if( FD_UNLIKELY( !sampler->unremoved_weight ) ) return FD_WSAMPLE_EMPTY;
   if( FD_UNLIKELY(  sampler->poisoned_mode    ) ) return FD_WSAMPLE_INDETERMINATE;
-  ulong unif = fd_chacha20_rng_ulong_roll( sampler->rng, sampler->unremoved_weight+sampler->poisoned_weight );
+  ulong unif = fd_chacha_rng_ulong_roll( sampler->rng, sampler->unremoved_weight+sampler->poisoned_weight );
   if( FD_UNLIKELY( unif>=sampler->unremoved_weight ) ) return FD_WSAMPLE_INDETERMINATE;
   return (ulong)fd_wsample_map_sample( sampler, unif );
 }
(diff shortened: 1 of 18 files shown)


ripatel-fd (dismissed):

src/disco/shred/fd_stake_ci.c
```diff
@@ -4,7 +4,7 @@
 
 #define SORT_NAME sort_pubkey
 #define SORT_KEY_T fd_shred_dest_weighted_t
-#define SORT_BEFORE(a,b) (memcmp( (a).pubkey.uc, (b).pubkey.uc, 32UL )<0)
+#define SORT_BEFORE(a,b) (memcmp( (a).pubkey.uc, (b).pubkey.uc, 32UL )>0)
```
0x0ece, line 7: I believe this is a bug...

src/disco/shred/test_shred_dest_conformance.c
```diff
@@ -0,0 +1,694 @@
...
+    8
+  );
+
+  FD_LOG_NOTICE(( "pass" ));
+  fd_halt();
+  return 0;
+}
+
+/* ============================================================================
+   RUST REFERENCE IMPLEMENTATION
```
ripatel-fd, line 506: Should probably move to contrib and make it an actual .rs file so we can compile and run it

0x0ece: It's all private stuff in rust 😭

src/disco/shred/fd_shred_dest.h
```diff
@@ -161,7 +161,8 @@ fd_shred_dest_idx_t *
 fd_shred_dest_compute_first( fd_shred_dest_t          * sdest,
                              fd_shred_t const * const * input_shreds,
                              ulong                      shred_cnt,
-                             fd_shred_dest_idx_t      * out );
+                             fd_shred_dest_idx_t      * out,
+                             int                        use_chacha8 );
```
ptaffet-jump, line 167: Should be mentioned in the function comment

src/ballet/wsample/fd_wsample.c
```diff
@@ -127,7 +127,7 @@ struct __attribute__((aligned(64UL))) fd_wsample_private {
   uint               height;
   char               restore_enabled;
   char               poisoned_mode;
-  /* Two bytes of padding here */
+  /* One byte of padding here */
```
ptaffet-jump: This is still two bytes now, right?

src/ballet/wsample/fd_wsample.h
```diff
@@ -135,10 +138,10 @@ void * fd_wsample_new_fini( void * shmem, ulong poisoned_weight );
 /* fd_wsample_get_rng returns the value provided for rng in new. */
 fd_chacha_rng_t * fd_wsample_get_rng( fd_wsample_t * sampler );
 
-/* fd_wsample_seed_rng seeds the ChaCha20 rng with the provided seed in
+/* fd_wsample_seed_rng seeds the ChaCha rng with the provided seed in
    preparation for sampling.  This function is compatible with Solana's
    ChaChaRng::from_seed. */
-void fd_wsample_seed_rng( fd_chacha_rng_t * rng, uchar seed[static 32] );
+void fd_wsample_seed_rng( fd_wsample_t * sampler, uchar seed[ 32 ], int use_chacha8 );
```
ptaffet-jump: I know this is my mistake, but could you fix the formatting? Also, isn't `static 32` better? I know there's an even better syntax that was more gross, but I thought `static 32` was better than just `32`

src/disco/shred/fd_shred_dest.h
```diff
@@ -153,15 +153,18 @@ static inline ulong fd_shred_dest_cnt_all     ( fd_shred_dest_t * sdest ) { retu
    input_shreds[i] for i in [0, shred_cnt).  shred_cnt must be in [0,
    67].  The destination index for input_shreds[i] is stored at out[i].
    input_shreds==NULL is fine if shred_cnt==0, in which case this
-   function is a no-op.  Returns out on success and NULL on failure.
+   function is a no-op.  If use_chacha8 is true, the tree is computed
```
ptaffet-jump: true -> `non-zero`

src/disco/shred/fd_shred_dest.h
```diff
@@ -205,7 +208,8 @@ fd_shred_dest_compute_children( fd_shred_dest_t          * sdest,
                                 ulong                      out_stride,
                                 ulong                      fanout,
                                 ulong                      dest_cnt,
-                                ulong                    * opt_max_dest_cnt );
+                                ulong                    * opt_max_dest_cnt,
+                                int                        use_chacha8 );
```
ptaffet-jump, line 214: Sorry, but same here

src/disco/shred/fd_shred_tile.c
```diff
@@ -260,6 +262,26 @@ typedef struct {
...
+   https://github.com/anza-xyz/agave/blob/v3.1.4/core/src/shred_fetch_stage.rs#L456 */
+static inline int
+fd_shred_check_feature_activation( ulong shred_slot, ulong feature_slot, fd_shred_ctx_t * ctx ) {
+  fd_epoch_leaders_t const * lsched = fd_stake_ci_get_lsched_for_slot( ctx->stake_ci, shred_slot );
+  if( lsched==NULL ) {
+    /* when the shred slot is out of a reasonable range it won't be propagated,
+       so the feature activation doesn't really matter */
+    return 0;
+  }
+  fd_epoch_schedule_t default_schedule[1] = {{ lsched->slot_cnt, lsched->slot_cnt, 0, 0, 0 }};
```
ptaffet-jump, line 292: Can @topointon-jump or someone weigh in on if this is correct, even for a test cluster with warmup slots? Or if this is good enough?

src/ballet/wsample/fd_wsample.h
0x0ece: To my understanding there's no difference static vs not, so we ended up not adding static anywhere. This means >= 32. Then yes, there's the weird syntax to say "exactly 32", but we're never using it.

src/disco/shred/fd_shred_tile.c
```diff
@@ -917,7 +939,8 @@ after_frag( fd_shred_ctx_t *    ctx,
             the shred, but still send it to the blockstore. */
           fd_shred_dest_t * sdest = fd_stake_ci_get_sdest_for_slot( ctx->stake_ci, shred->slot );
           if( FD_UNLIKELY( !sdest ) ) break;
-          fd_shred_dest_idx_t * dests = fd_shred_dest_compute_children( sdest, &shred, 1UL, ctx->scratchpad_dests, 1UL, fanout, fanout, max_dest_cnt );
+          int use_chacha8 = fd_shred_check_feature_activation( shred->slot, ctx->features_activation->switch_to_chacha8_turbine, ctx );
```
ptaffet-jump: This is probably fine, but would it be possible to convert the activation epoch to a slot when it is updated so that all the divisions move off the critical path?

ptaffet-jump: I.e. `int use_chacha8 = shred->slot >= ctx->switch_to_chacha8_slot;` or something like that

src/disco/shred/fd_shred_tile.c
topointon-jump: This won't work properly for clusters with warmup slots, debatable if this is a big issue or not (I personally don't think it is)

ptaffet-jump: Okay, I'd probably put this in the "good enough" category then

src/disco/shred/fd_shred_tile.c
0x0ece: good point. This is true for all turbine features, so I did it for all feature slots. The aux function now converts the feature_slot, and the check for activation remains the classic `shred_slot >= feature_slot`.

src/disco/shred/fd_shred_tile.c
0x0ece: I captured this in the comment of the fn, so we remember

PR #7606, opened 2025-12-18
http: do not search outside the header value

intrigus-lgtm:
Header values in picohttp are not null-terminated so `strstr` would happily search outside the header value and exceed the length of the header value.
This could lead to _DoS in artifical circumstances_ **which don't apply to fd in practice**. There is always a null byte in memory _somewhere_ after the header value even if it is not actually part of the header value. This is because `fd_http_server_ws_frame`s are allocated after the request buffer that contains the headers and the ws_frames contain four bytes of padding that is zero, because the whole memory we're operating on, has been allocated with `mmap` anonymously which zeroes the memory QED.

Drive-by fix: the `Upgrade` header field's token has to contain the value "websocket", but should be treated as an ASCII case-insensitive value: https://datatracker.ietf.org/doc/html/rfc6455#section-4.2.1

### DIFF
--- a/src/waltz/http/fd_http_server.c
+++ b/src/waltz/http/fd_http_server.c
@@ -575,13 +575,14 @@ read_conn_http( fd_http_server_t * http,
 
   conn->upgrade_websocket = 0;
   int compress_websocket = 0;
-  if( FD_UNLIKELY( upgrade_key && !strncmp( upgrade_key, "websocket", 9UL ) ) ) {
+  if( FD_UNLIKELY( upgrade_key && !strncasecmp( upgrade_key, "websocket", 9UL ) ) ) {
     conn->request_bytes_len = (ulong)result;
     conn->upgrade_websocket = 1;
 
 #if FD_HAS_ZSTD
     for( ulong i=0UL; i<num_headers; i++ ) {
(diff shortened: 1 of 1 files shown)


src/waltz/http/fd_http_server.c
```diff
@@ -581,7 +581,9 @@ read_conn_http( fd_http_server_t * http,
 
 #if FD_HAS_ZSTD
     for( ulong i=0UL; i<num_headers; i++ ) {
-      if( FD_LIKELY( headers[ i ].name_len==22UL && !strncasecmp( headers[ i ].name, "Sec-WebSocket-Protocol", 22UL ) && strstr( headers[ i ].value, "compress-zstd" ) ) ) {
+      // Use `memmem` as the header is not guaranteed to be null terminated
+      // `memmem` also matches "randomprefix_compress-zstd_randomsuffix" but that's fine, we don't want to fully parse the header value
```
mmcgee-jump: Wouldn't it be better to just check value_len == 13 and use strncasecmp here too?

intrigus-lgtm: I was thinking that there might be cases where there are multiple protocols specified in the header, but the existing code and documentation quite explicitly say that this should just be a single value.
Changed to use value_len == 13 and strncmp, because the header field value is usually case-sensitive.

PR #7610, opened 2025-12-18
accdb: implement account reclaims

ripatel-fd:
- **accdb: remove unused guard structs**
- **accdb: implement record deletions / reclaims**


src/flamenco/accdb/fd_accdb_admin.c
```diff
@@ -214,6 +215,35 @@ fd_accdb_cancel( fd_accdb_admin_t *        accdb,
...
+  fd_funk_xid_key_pair_t pair = rec->pair;
+  fd_funk_rec_query_t query[1];
+  int rm_err = fd_funk_rec_map_remove( funk->rec_map, &pair, NULL, query, FD_MAP_FLAG_BLOCKING );
+  if( FD_UNLIKELY( rm_err!=FD_MAP_SUCCESS ) ) FD_LOG_CRIT(( "fd_funk_rec_map_remove failed (%i-%s)", rm_err, fd_map_strerror( rm_err ) ));
+
+  /* Phase 2: Invalidate record */
+
+  fd_funk_rec_t * old_rec = query->ele;
+  memset( &old_rec->pair, 0, sizeof(fd_funk_xid_key_pair_t) );
+  FD_COMPILER_MFENCE();
```
topointon-jump, line 238: Just so I understand, `fd_accdb_txn_cancel_one` has the `FD_COMPILER_MFENCE` both before and after the memset - do we need to do that here too? If not, why not?

ripatel-fd: Good point, I'll add another compiler fence. We should always finish removing the account from the record map before we change the map key.

PR #7612, opened 2025-12-18
restore: snapwm and multiple snapwr (vinyl)

jvarela-jump:
Two architectural optimizations to the snapshot load pipeline under vinyl:

1. `snapin` tile is split into two:

- `snapin` handles parsing and bstream pair generation.
- `snapwm` handles vinyl meta_map and bstream allocation.

    With this change: `snapin`->`snapwm`->`snapwh`->`snapwr`.

2. The topology now supports multiple `snapwr` tiles, and `snapwh` handles multiple fseq accordingly.
- the number of `snapwr` tiles is configurable from the toml.

The measured speedup is around ~1.6x.


src/discof/restore/fd_snapwm_tile.c
```diff
@@ -0,0 +1,382 @@
...
+  fd_snapwm_vinyl_privileged_init( ctx, topo, tile );
+}
+
+static inline fd_snapwm_out_link_t
+out1( fd_topo_t const *      topo,
+      fd_topo_tile_t const * tile,
+      char const *           name ) {
+  ulong idx = fd_topo_find_tile_out_link( topo, tile, name, 0UL );
+
+  if( FD_UNLIKELY( idx==ULONG_MAX ) ) return (fd_snapwm_out_link_t){ .idx = ULONG_MAX, .mem = NULL, .chunk0 = 0, .wmark = 0, .chunk = 0, .mtu = 0 };
```
ripatel-fd:
```suggestion
  if( FD_UNLIKELY( idx==ULONG_MAX ) ) return (fd_snapwm_out_link_t){0};
```

src/discof/restore/fd_snapwm_tile.c
```diff
@@ -0,0 +1,382 @@
...
+
+  ulong mtu = topo->links[ tile->out_link_id[ idx ] ].mtu;
+  if( FD_UNLIKELY( mtu==0UL ) ) return (fd_snapwm_out_link_t){ .idx = idx, .mem = NULL, .chunk0 = ULONG_MAX, .wmark = ULONG_MAX, .chunk = ULONG_MAX, .mtu = mtu };
```
ripatel-fd:
```suggestion
  if( FD_UNLIKELY( mtu==0UL ) ) return (fd_snapwm_out_link_t){0};
```

src/discof/restore/fd_snapin_tile_vinyl.c
```diff
@@ -1,115 +1,14 @@
...
-  - Phase 2: once read is done
-    - Replay all elements written to bstream (io_mm)
-    - Populate the vinyl_meta index while replaying
-
-\**********************************************************************/
-
 void
 fd_snapin_vinyl_privileged_init( fd_snapin_tile_t * ctx,
                                  fd_topo_t *        topo,
                                  fd_topo_tile_t *   tile ) {
```
ripatel-fd: This function can be removed

src/discof/restore/fd_snapin_tile_vinyl.c
```diff
@@ -688,71 +201,10 @@ fd_snapin_read_account_vinyl( fd_snapin_tile_t *  ctx,
...
-    FD_LOG_WARNING(( "failed to read account %s: account data size (%lu bytes) exceeds buffer size (%lu bytes)",
-                     acct_addr_b58, (ulong)meta->dlen, data_max ));
-  }
-  memcpy( data, mmio+seq_data, meta->dlen );
+  (void)ctx;
+  (void)acct_addr;
+  (void)meta;
+  (void)data;
+  (void)data_max;
+  FD_LOG_ERR(( "fd_snapin_read_account_vinyl is not supported here" ));
```
ripatel-fd: Is snapin_read_account even used anymore? Maybe this can just be removed entirely

src/discof/restore/fd_snapwm_tile.c
```diff
@@ -0,0 +1,382 @@
...
+/* verify_slot_deltas_with_slot_history verifies the 'SlotHistory'
+   sysvar account after loading a snapshot.  The full database
+   architecture is only instantiated after snapshot loading, so this
+   function uses a primitive/cache-free mechanism to query the parts of
+   the account database that are available.
+
+   Returns 0 if verification passed, -1 if not. */
+
+static int
+verify_slot_deltas_with_slot_history( fd_snapwm_tile_t * ctx ) {
```
ripatel-fd, line 69: Is there a way we could avoid duplication of this function between snapin and snapwm?

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -171,50 +174,73 @@ backtest_topo( config_t * config ) {
...
     if( vinyl_enabled ) {
-      fd_topo_link_t * snapin_wh = fd_topob_link( topo, "snapin_wh", "snapin_wr", 4UL, 16UL<<20, 1UL );
-      fd_topob_link( topo, "snapwh_wr", "snapin_wr", 4UL, 0UL, 1UL );
-      fd_pod_insertf_ulong( topo->props, 8UL, "obj.%lu.app_sz", snapin_wh->dcache_obj_id );
+      if( FD_LIKELY( snapshot_lthash_disabled ) ) {
+        fd_topob_link( topo, "snapwm_ct", "snapwm_ct",   128UL,   0UL,                            1UL );
+      } else {
+        /* TODO pending */
+      }
+      fd_topob_link( topo, "snapin_txn", "snapin_txn",    4UL,   (ulong)((3764697600UL+64UL)/4),  1UL ); /* mtu=(sizeof(fd_sstxncache_entry_t)*(FD_SNAPIN_TXNCACHE_MAX_ENTRIES+1UL))/depth */
```
cali-jumptrading: Any way this number can be represented using a `sizeof(some type)`?

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -171,50 +174,73 @@ backtest_topo( config_t * config ) {
...
+      fd_topob_link( topo, "snapin_wm", "snapin_wm",     16UL,   128UL<<20,                       1UL ); /* FD_SSPARSE_ACC_BATCH_MAX * 16<<20 */
```
cali-jumptrading: Same here, would prefer the generic version using `FD_SSPARSE_ACC_BATCH * 16`.

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -171,50 +174,73 @@ backtest_topo( config_t * config ) {
...
+      fd_topo_link_t * snapwm_wh =
+      fd_topob_link( topo, "snapwm_wh", "snapwm_wr",     16UL,    16UL<<20,                       1UL );
```
cali-jumptrading: would be nice to have a comment to explain the hardcoded 16<<20UL ( I understand it is 16 MB chunks )

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -171,50 +174,73 @@ backtest_topo( config_t * config ) {
...
+      fd_topob_link( topo, "snapwh_wr", "snapwm_wr",     16UL,    0UL,                            1UL );
```
cali-jumptrading: This is a link for passing control messages right? Maybe we should comment this too... I got a bit confused at first.

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -223,29 +249,35 @@ backtest_topo( config_t * config ) {
...
-      FOR(lta_tile_cnt) fd_topob_tile_out( topo, "snapla", i,                "snapla_ls",   i                                          );
-      /**/              fd_topob_tile_in ( topo, "snapls", 0UL, "metric_in", "snapin_ls",   0UL, FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
-      FOR(lta_tile_cnt) fd_topob_tile_in ( topo, "snapls", 0UL, "metric_in", "snapla_ls",   i,    FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
-      /**/              fd_topob_tile_out( topo, "snapls", 0UL,              "snapls_ct",   0UL                                        );
-    }
-
     if( vinyl_enabled ) {
-      fd_topob_tile_out( topo, "snapin", 0UL,              "snapin_wh", 0UL );
-      fd_topob_tile_in ( topo, "snapwh", 0UL, "metric_in", "snapin_wh", 0UL, FD_TOPOB_RELIABLE, FD_TOPOB_POLLED );
+      if( FD_LIKELY( !snapshot_lthash_disabled ) ) {
```
cali-jumptrading: the two negatives (!snaphot_lthash_disabled) are a bit confusing here. Maybe stick to the original `if( snapshot_lthash_disabled )`?

src/app/firedancer-dev/commands/backtest.c
```diff
@@ -223,29 +249,35 @@ backtest_topo( config_t * config ) {
...
+      fd_topob_tile_out( topo, "snapin", 0UL,              "snapin_txn",0UL );
+      fd_topob_tile_in ( topo, "snapwm", 0UL, "metric_in", "snapin_txn",0UL, FD_TOPOB_RELIABLE, FD_TOPOB_POLLED );
+      fd_topob_tile_out( topo, "snapwm", 0UL,              "snapwm_wh", 0UL );
+      fd_topob_tile_in ( topo, "snapwh", 0UL, "metric_in", "snapwm_wh", 0UL, FD_TOPOB_RELIABLE, FD_TOPOB_POLLED );
       fd_topob_tile_out( topo, "snapwh", 0UL,              "snapwh_wr", 0UL );
-      fd_topob_tile_in ( topo, "snapwr", 0UL, "metric_in", "snapwh_wr", 0UL, FD_TOPOB_RELIABLE, FD_TOPOB_POLLED );
-      fd_topob_tile_uses( topo, snapwr_tile, &topo->objs[ topo->links[ fd_topo_find_link( topo, "snapin_wh", 0UL ) ].dcache_obj_id ], FD_SHMEM_JOIN_MODE_READ_ONLY );
+      FOR(snapwr_tile_cnt) fd_topob_tile_in ( topo, "snapwr", i, "metric_in", "snapwh_wr", 0UL, FD_TOPOB_RELIABLE, FD_TOPOB_POLLED );
+      FOR(snapwr_tile_cnt) fd_topob_tile_uses( topo, &topo->tiles[ fd_topo_find_tile( topo, "snapwr", i ) ], &topo->objs[ topo->links[ fd_topo_find_link( topo, "snapwm_wh", 0UL ) ].dcache_obj_id ], FD_SHMEM_JOIN_MODE_READ_ONLY );
```
cali-jumptrading, line 291: maybe add comment explaning that wr shares the dcache with wh?

src/discof/restore/fd_snapin_tile_vinyl.c
jvarela-jump: Done.

src/discof/restore/fd_snapin_tile_vinyl.c
jvarela-jump: It has been removed on the vinyl side of the tile.
`verify_slot_deltas_with_slot_history` still invokes `snapin_read_account` under funk.

src/app/firedancer-dev/commands/backtest.c
jvarela-jump: Done.

src/app/firedancer-dev/commands/backtest.c
jvarela-jump: Done.

src/app/firedancer-dev/commands/backtest.c
jvarela-jump: Done.

src/app/firedancer-dev/commands/backtest.c
jvarela-jump: Yes, good point. An explanation has been added.

src/app/firedancer-dev/commands/backtest.c
jvarela-jump: Yes, good point. An explanation has been added.

src/app/firedancer/topology.c
```diff
@@ -1237,26 +1271,29 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
...
+
+    tile->snapwm.lthash_disabled = !!config->development.snapshots.disable_lthash_verification;
+    strcpy( tile->snapwm.vinyl_path, config->paths.accounts );
+    tile->snapwm.vinyl_meta_map_obj_id  = fd_pod_query_ulong( config->topo.props, "vinyl.meta_map",  ULONG_MAX );
+    tile->snapwm.vinyl_meta_pool_obj_id = fd_pod_query_ulong( config->topo.props, "vinyl.meta_pool", ULONG_MAX );
+
+    ulong wm_wr_link_id = fd_topo_find_link( &config->topo, "snapwm_wh", 0UL );
+    FD_TEST( wm_wr_link_id!=ULONG_MAX );
+    fd_topo_link_t * wm_wr_link = &config->topo.links[ wm_wr_link_id ];
+    tile->snapwm.snapwr_depth = wm_wr_link->depth;
```
cali-jumptrading: Is this needed when you already have the dcache obj id shared?

src/discof/restore/fd_snapwm_tile.c
jvarela-jump: That is a good point, but just one observation: the version in snapin would be removed once funk is deprecated from the snapshot load pipeline.

src/discof/restore/fd_snapin_tile_private.h
```diff
@@ -306,7 +281,7 @@ fd_snapin_read_account( fd_snapin_tile_t *  ctx,
                         uchar *             data,
                         ulong               data_max ) {
   if( ctx->use_vinyl ) {
-    fd_snapin_read_account_vinyl( ctx, acct_addr, meta, data, data_max );
+    FD_LOG_ERR(( "read account is not supported under vinyl" ));
```
cali-jumptrading, line 287: Did this change because the account reads happen in the wm tile? Maybe we can add a comment that this will go away when snapshot load into funk goes away.

src/app/firedancer-dev/commands/backtest.c
jvarela-jump: Ok, restructuring backtest, full topology and snapshot_load using this format wherever applicable. Once funk is deprecated from the snapshot load pipeline, the code will be simplified.
```
  if( vinyl_enabled ) {
    ...
  }
  if( snapshot_lthash_disabled ) {
    if( vinyl_enabled ) {
      ...
    } else {
      ...
    }
  } else {
    if( vinyl_enabled ) {
      ...
    } else {
      ...
    }
  }
```

src/discof/restore/fd_snapwm_tile_private.h
```diff
@@ -0,0 +1,185 @@
...
+  long boot_timestamp;
+
+  fd_stem_context_t *      stem;
+
+  fd_sstxncache_entry_t * txncache_entries;
+  ulong const *           txncache_entries_len_ptr;
+
+  struct {
+    /* Account counters (full + incremental) */
+    ulong accounts_loaded;
```
cali-jumptrading, line 42: would these metrics for the incremental snapshot only be useful?

src/discof/restore/fd_snapin_tile.c
```diff
@@ -827,13 +795,25 @@ unprivileged_init( fd_topo_t *      topo,
...
+       uses the dcache of the snapin_txn link as the scratch memory.
+       The app field of the dcache is used to communicate the
+       txncache_entries_len value. */
+    fd_snapin_out_link_t snapin_txn = out1( topo, tile, "snapin_txn" );
+    FD_TEST( !!snapin_txn.mem );
+    fd_topo_link_t const * out_link_txn = &topo->links[ tile->out_link_id[ snapin_txn.idx ] ];
+    ulong depth = out_link_txn->depth;
+    FD_TEST( ( depth*snapin_txn.mtu )==( sizeof(fd_sstxncache_entry_t)*FD_SNAPIN_TXNCACHE_MAX_ENTRIES ) );
+    ctx->txncache_entries                 = fd_chunk_to_laddr( snapin_txn.mem, snapin_txn.chunk0 );
+    ctx->txncache_entries_len_vinyl_ptr   = (ulong*)fd_dcache_app_laddr( out_link_txn->dcache );
```
cali-jumptrading, line 812: why do we need the extra `txncache_entries_len_vinyl_ptr`?

src/discof/restore/fd_snapwm_tile.c
```diff
@@ -0,0 +1,407 @@
...
+  l = FD_LAYOUT_APPEND( l, fd_vinyl_io_mm_align(),    fd_vinyl_io_mm_footprint( FD_SNAPWM_IO_SPAD_MAX     ) );
+  return FD_LAYOUT_FINI( l, scratch_align() );
+}
+
+static void
+metrics_write( fd_snapwm_tile_t * ctx ) {
+  FD_MGAUGE_SET( SNAPWM, ACCOUNTS_LOADED,        ctx->metrics.accounts_loaded );
+  FD_MGAUGE_SET( SNAPWM, ACCOUNTS_REPLACED,      ctx->metrics.accounts_replaced );
+  FD_MGAUGE_SET( SNAPWM, ACCOUNTS_IGNORED,       ctx->metrics.accounts_ignored );
+  FD_MGAUGE_SET( SNAPWM, STATE, (ulong)ctx->state );
```
cali-jumptrading: nit: alignment

src/discof/restore/fd_snapwm_tile.c
```diff
@@ -0,0 +1,407 @@
...
+                 ulong               sig,
+                 ulong               chunk,
+                 ulong               sz,
+                 ulong               ctl    FD_PARAM_UNUSED,
+                 ulong               tsorig FD_PARAM_UNUSED,
+                 ulong               tspub  FD_PARAM_UNUSED,
+                 fd_stem_context_t * stem ) {
+  FD_TEST( ctx->state!=FD_SNAPSHOT_STATE_SHUTDOWN );
+
+  ctx->stem = stem;
```
cali-jumptrading: this ctx->stem pointer doesn't seem to be needed.

src/discof/restore/fd_snapwm_tile.c
```diff
@@ -0,0 +1,407 @@
...
+  FD_TEST( !!ctx->in.wksp          );
+  FD_TEST( !!ctx->txncache_entries );
+
+  fd_snapwm_vinyl_unprivileged_init( ctx, topo, tile, _io_mm, _io_wd );
+}
+
+/* Control fragments can result in one extra publish to forward the
+   message down the pipeline, in addition to the result / malformed
+   message. Can send one duplicate account message as well. */
+#define STEM_BURST 3UL
```
cali-jumptrading: I think this should be 2... looks like wm doesn't send any other messages besides control messages downstream. Though it looks like this may increase when you add the lthash vinyl tiles

src/discof/restore/fd_snapwm_tile_private.h
```diff
@@ -0,0 +1,185 @@
...
+/* fd_snapwm_vinyl_wd_init transitions the vinyl backend from generic
+   vinyl accessor (io_mm) to fast dumb direct account insertion (io_wd).
+   This must be called before calling fd_snapwm_process_account_*.
+   Starts the snapwr tile (waits for the snapwr tile to ack). */
+
+void
+fd_snapwm_vinyl_wd_init( fd_snapwm_tile_t * ctx );
+
+/* fd_snapwm_vinyl_wd_fini transitions the vinyl backend from fast dumb
+   direct account insertion (io_wd) back to generic mode (io_mm).
```
cali-jumptrading, line 149: I don't understand why we need to transition from generic mode to direct mode and back... why can't we just start and end in direct mode?

src/discof/restore/fd_snapin_tile.c
cali-jumptrading: wait nvm I figured it out

src/discof/restore/fd_snapwm_tile_private.h
jvarela-jump: Yes, good point. The incremental count could be inferred from the difference total-full.
The metrics were ported from the current version of snapin into snapwm, to keep it consistent with the current metrics.

src/discof/restore/fd_snapwm_tile_private.h
jvarela-jump: Good question: `snapwm` follows the original vinyl implementation from `snapin`. But in short: inside `handle_control_frag`  there is some order needed between `fd_snapwm_vinyl_txn_begin` (which uses `io_mm`) and `fd_snapwm_vinyl_wd_init` (which uses `io_wd`). Because the latter is considered to be a "fast dumb" version, the ctx always goes back to `io_mm` whenever possible. The original version makes the transitions more clear (and verbose). Having said that, nothing would prevent us from starting from `io_wd`, but it would require changes to the state machine. We could revise this at a later time though, once we start streamlining the snapshot load pipeline, if that sounds ok?

src/discof/restore/fd_snapwm_tile.c
jvarela-jump: Done.

src/discof/restore/fd_snapin_tile_private.h
jvarela-jump: Ok. A comment has been added.

src/discof/restore/fd_snapwm_tile.c
jvarela-jump: Removed.

src/discof/restore/fd_snapwm_tile.c
jvarela-jump: Yes, that is correct, good point. Setting STEM_BURST to 2, until the lthash vinyl PR is ready.

src/discof/restore/fd_snapwm_tile_private.h
cali-jumptrading: yes thank you for the explanation!

src/discof/restore/fd_snapwm_tile_private.h
cali-jumptrading: ok makes sense!

cali-jumptrading: Though... if we did have incremental only metrics we could see how accounts/duplicate accounts progress in realtime as we load the incremental. Also... if we reset the incremental snapshot, do we rollback the total counts?

cali-jumptrading: Ah I see we do rollback correctly! thanks

PR #7615, opened 2025-12-19
flamenco, features: update key for disable_sbpf_v0_execution

topointon-jump:


topointon-jump:
This causes a hash collision in our features.. if/when this is ever activated this will be re-keyed so it's not worth changing now

PR #7617, opened 2025-12-19
Simplify txn timing accounting

ripatel-fd:
- Replace uchar percentage compression of transaction timings with
  float nanosecond duration
- Add txn and VM metrics to bank tile
- Simplify executor time measurements


src/discof/bank/fd_bank_tile.c
```diff
@@ -472,19 +502,10 @@ handle_bundle( fd_bank_ctx_t *     ctx,
...
-    long tx_load_end_ticks    = fd_long_if( ctx->txn_out[ i ].details.exec_start_timestamp!=LONG_MAX,   ctx->txn_out[ i ].details.exec_start_timestamp,   tx_start_ticks         );
-    long tx_end_ticks         = fd_long_if( ctx->txn_out[ i ].details.commit_start_timestamp!=LONG_MAX, ctx->txn_out[ i ].details.commit_start_timestamp, tx_load_end_ticks      );
-
-    trailer->txn_preload_end_pct = (uchar)(((double)(tx_preload_end_ticks - microblock_start_ticks) * (double)UCHAR_MAX) / (double)microblock_duration_ticks);
-    trailer->txn_start_pct       = (uchar)(((double)(tx_start_ticks       - microblock_start_ticks) * (double)UCHAR_MAX) / (double)microblock_duration_ticks);
-    trailer->txn_load_end_pct    = (uchar)(((double)(tx_load_end_ticks    - microblock_start_ticks) * (double)UCHAR_MAX) / (double)microblock_duration_ticks);
-    trailer->txn_end_pct         = (uchar)(((double)(tx_end_ticks         - microblock_start_ticks) * (double)UCHAR_MAX) / (double)microblock_duration_ticks);
+    trailer->txn_preload_end_nanos = (float)( txn_start_ticks[ i ] - microblock_start_ticks ) * ctx->ns_per_tick;
```
jherrera-jump, line 505: is it possible to make `txn_preload_end_nanos` and `txn_start_nanos` distinct? Otherwise in the frontend the `validate` state will always show as having zero duration. this state should include pre-execution validation (e.g. status cache).

src/flamenco/runtime/fd_runtime.h
```diff
@@ -215,6 +211,12 @@ struct fd_txn_out {
     fd_pubkey_t                 programs_to_reverify[ MAX_TX_ACCOUNT_LOCKS ];
   } details;
 
+  /* fd_tickcount() observations at various txn processing steps */
+  struct {
+    long exec_start_ticks;
+    long exec_end_ticks;
```
jherrera-jump, line 217: Why the change from commit_start to exec_end? For microblocks in the leader pipeline with more than one txn (e.g. bundles), `exec_end` is not equivalent to `commit_start` since all the commits happen after all executions have completed.

jherrera-jump (changes requested):
While we're here, could we also rename the variable to something more intelligible actually refering to the states displayed on the frontend. Maybe

txn_preload_end -> txn_validate_start
txn_start -> txn_load_start
txn_load_end -> txn_exec_start
txn_end -> txn_commit_start

src/discof/bank/fd_bank_tile.c
ripatel-fd: Could we just remove this state for full FD? I think this comes Frankendancer, but Firedancer's pipeline conceptually has one stage fewer. The status cache checks are already part of the "prepare" parts in full Firedancer, which are already counted towards another metric.

jherrera-jump: hmm fair point, I think just leave it as 0 and I can remove it later

PR #7620, opened 2025-12-22
populate the slot field in became_leader_t

cavemanloverboy:
image


cavemanloverboy:
technically unnecessary because the slot is communciated via the `sig`, but cmon

PR #7626, opened 2025-12-22
solcap v3

ripatel-fd:
This PR is a rewrite of solcap structured runtime event logging infrastructure

- Events are now encoded in Protobuf, including account data
- Event types are specified/documented in https://github.com/firedancer-io/solana-core.git
- Protobuf encoding is done via a new schema-less high-performance encoder
- The solcap custom file formats in v1 (completely custom) and v2 (pcapng extension) are ditched for standard pcapng (UDP/IP encapsulation)
- Adds a 'bank sequence number' to uniquely identify banks for the lifetime of a Firedancer run
- Config option cleanup & documentation


contrib/test/run_solcap_tests.sh
ripatel-fd, line 1: Going to bring this back, accidental deletion

PR #7639, opened 2025-12-30
txn: change parser to parse txns with >64 instrs

ibhatt-jumptrading:
currently transactions can be successfully parsed and included in a block if they have more than 64 instructions. However, they will always fail to execute successfully due to runtime protocol limits. Until this point, firedancer/frankendancer would reject these transactions as being invalid

1. the firedancer transaction scheduler uses fd_txn_parse_core to parse transactions from blocks produced by other validators and needs to support parsing >64 instr transactions. If it doesn't we will incorrectly mark a block as invalid leading to a hard fork
2. this will enable us to pack transactions into our own blocks in both firedancer and frankendancer with >64 instrs. Even though these transactions will fail, they can still be included in a block and a fee will be collected


ripatel-fd:
@ibhatt-jumptrading Why do we need this? (Can you add this to the commit description for future reference)

src/ballet/txn/fd_txn_parse.c
```diff
@@ -145,7 +141,7 @@ fd_txn_parse_core( uchar const             * payload,
     /* Need to assign addr_table_lookup_cnt,
        addr_table_adtl_writable_cnt, addr_table_adtl_cnt,
        _padding_reserved_1 later */
-    parsed->instr_cnt                     = instr_cnt;
+    parsed->instr_cnt                     = fd_ushort_min( instr_cnt, FD_TXN_INSTR_MAX );
```
yufeng-jump, line 144: Does the runtime fail a >64 instr transaction based on this field or another field that isn’t clipped?

src/flamenco/runtime/tests/run_backtest_ci.sh
```diff
@@ -29,3 +29,4 @@ src/flamenco/runtime/tests/run_ledger_backtest.sh -l localnet-stake-v3.0.0 -y 1
 src/flamenco/runtime/tests/run_ledger_backtest.sh -l mainnet-378539412 -y 5 -m 2000000 -e 378539445
 src/flamenco/runtime/tests/run_ledger_backtest.sh -l devnet-422969842 -y 1 -m 2000000 -e 422969848
 src/flamenco/runtime/tests/run_ledger_backtest.sh -l breakpoint-385786458 -y 1 -m 2000000 -e 385786458
+src/flamenco/runtime/tests/run_ledger_backtest.sh -l devnet-427963044 -y 1 -m 2000000 -e 427963085
```
yufeng-jump: nit: usually new line at end of file

src/ballet/txn/fd_txn_parse.c
ibhatt-jumptrading: we set the has_unparsed_instrs flag and fail based on if it's set or not

src/ballet/txn/fd_txn_parse.c
```diff
@@ -145,7 +141,8 @@ fd_txn_parse_core( uchar const             * payload,
...
+    parsed->has_unparsed_instrs           = (instr_cnt>FD_TXN_INSTR_MAX) ? 1 : 0;
```
mjain-jump, line 145:
```suggestion
    parsed->has_unparsed_instrs           = !!(instr_cnt>FD_TXN_INSTR_MAX);
```

src/flamenco/runtime/fd_executor.c
```diff
@@ -1714,6 +1714,17 @@ fd_execute_txn( fd_runtime_t *      runtime,
...
+  /* If the transaction has more than FD_TXN_INSTR_MAX instructions,
+     we know that the transaction has to fail.  Firedancer's txn parsing
+     logic ignores any instructions past the 64th one.  If the txn has
+     yet to fail at this point and we know that there were unparsed
+     transactions, we want to fail the transaction with an instruction
+     error. */
+  if( FD_UNLIKELY( txn->has_unparsed_instrs ) ) {
+    txn_out->err.exec_err = FD_EXECUTOR_INSTR_ERR_MAX_INSN_TRACE_LENS_EXCEEDED;
+    return FD_RUNTIME_TXN_ERR_INSTRUCTION_ERROR;
+  }
```
mjain-jump, line 1726: isnt this check redundant because `fd_execute_instr` would always throw an error during instr stack pushing after 64 instructions?

ibhatt-jumptrading: we don't parse more than 64 instructions so it's not redundant

yufeng-jump (approved):

src/ballet/txn/fd_txn.h
```diff
@@ -85,18 +85,19 @@
    to 33. FIXME: We should petition to limit this to approx 8. */
 #define FD_TXN_ADDR_TABLE_LOOKUP_MAX (127UL)
 
-/* FD_TXN_INSTR_MAX: The (inclusive) maximum number of instructions a transaction
-   can have.  As of Solana 1.15.0, this is limited to 64. */
+/* FD_TXN_INSTR_MAX: The (inclusive) maximum number of instructions a
+   transaction can have.  As of Solana 1.15.0, this is limited to 64.
+   A transaction can have more than 64 instructions, but if more than
+   that many instructions are executed, a transaction will fail. */
```
ptaffet-jump, line 91: This corrects my wrong comment, but it isn't worded very well as is, because it contradicts itself. Maybe it should be "the maximum number of instructions a successful transaction can have" or "the maximum number of instructions a transaction can execute."

src/ballet/txn/fd_txn.h
```diff
@@ -285,13 +286,24 @@ struct fd_txn {
...
   /* instr_cnt: The number of instructions in this transaction.
-     instr_cnt in [0, FD_TXN_INSTR_MAX]. */
+     instr_cnt in [0, FD_TXN_INSTR_MAX].  There can be more than
+     FD_TXN_INSTR_MAX instructions in a transaction, but they will
+     not be parsed or counted here. */
   ushort      instr_cnt;
 
+  /* has_unparsed_instrs: 1 if the transaction has more than
+     FD_TXN_INSTR_MAX instructions, 0 otherwise.  */
+  uchar       has_unparsed_instrs;
```
ptaffet-jump, line 296: Why not just use the space that's currently called `_padding_reserved_1` for this?

src/ballet/txn/fd_txn_parse.c
ptaffet-jump: Yeah, I've gone back and forth on whether this should be clipped or not. It's hard to say. I think this is okay though.

src/ballet/txn/fd_txn.h
```diff
@@ -285,13 +286,24 @@ struct fd_txn {
...
+                                     insert anyways */
+
   /* instr: The array of instructions in this transaction. It's a "flexible array
      member" since C does not allow the pretty typical 0-len array at the end
      of the struct trick.
-     Indexed [0, instr_cnt). */
+     Indexed [0, min(instr_cnt, FD_TXN_INSTR_MAX)).  All transactions
+     past the 64th instruction are not parsed because they will
+     automatically fail to execute in the runtime due to protocol
+     limits. */
```
ptaffet-jump, line 306: Since `instr_cnt` is clamped at `FD_TXN_INSTR_MAX`, I think it is still indexed `[0, instr_cnt)`. I'd move the rest of the note to the doc comment for `instr_cnt`, and start it with `WARNING: `.

src/ballet/txn/fd_txn_parse.c
```diff
@@ -169,7 +166,11 @@ fd_txn_parse_core( uchar const             * payload,
        program ID can't come from a table. */
     CHECK( (0UL < (ulong)program_id) & ((ulong)program_id < (ulong)acct_addr_cnt) );
 
-    if( parsed ){
+    /* We only care to parse the first FD_TXN_INSTR_MAX (64)
+       instructions because will never execute more than that many
+       instructions in a single transaction since it will lead to an
+       execution failure.  We ignore the rest of the instructions. */
```
ptaffet-jump, line 172:
```suggestion
    /* We only care to parse the first FD_TXN_INSTR_MAX (64)
       instructions. We will never execute more than that many
       instructions in a single transaction (since it will lead to an
       execution failure) and the worst-case size of the fd_txn_t is excessively large if we don't add this optimization.  We ignore the rest of the instructions. */
```

ptaffet-jump (changes requested):
Overall, this is a good fix, but we realized IRL that this is still not correct because instructions past 64 can be compute budget instructions that we need to handle properly in order to fail the transaction properly. The best fix is probably just to get the static limit feature gate activated ASAP

topointon-jump:
> The best fix is probably just to get the static limit feature gate activated ASAP

`static_instruction_limit` is 3rd on the testnet queue

ibhatt-jumptrading:
easier to just wait for static_instruction_limit to get activated

PR #7647, opened 2026-01-02
net: factor out header validation

akhinvasara-jumptrading:
This PR centralizes network header validation. The XDP tile guarantees that packets sent downstream will have passed validation.

For ease of review/reversability of desired, the first commit performs the refactor. The second commit changes the reported xdp metrics to take advantage of the new error codes.

The RX perf impact of this change is slightly positive (tested using Pktgen-DPDK from rd-merenguel1 (mlx5) to rd-chachal1 on (mlx5):
before: 8.617 Mpps
After: 8.713 Mpps


src/util/net/fd_net_headers.h
```diff
@@ -1,8 +1,10 @@
 #ifndef HEADER_fd_src_util_net_fd_net_headers_h
 #define HEADER_fd_src_util_net_fd_net_headers_h
 
-#include "fd_udp.h"
 #include "fd_eth.h"
+#include "fd_ip4.h"
+#include "fd_udp.h"
+#include "fd_gre.h"
```
ripatel-fd, line 7: Can you revert these header additions?

src/disco/net/fd_net_checks.h
```diff
@@ -0,0 +1,179 @@
+#ifndef HEADER_fd_src_disco_net_fd_net_checks_h
+#define HEADER_fd_src_disco_net_fd_net_checks_h
+
+/* fd_net_checks.h is the authoritative definition of 'valid' RX packets from the
+   firedancer networking stack. Any FD networking implementation should drop received
+   fragments that fail these checks. These functions 'validate' only against these checks
+   for firedancer, rather than validating conformance to the internet protocols themselves. */
```
ripatel-fd: Can you precisely specify the filtering to be done by the net tile on the RX path? 
Either in a Markdown document or in this header.

src/disco/net/fd_net_checks.h
```diff
@@ -0,0 +1,179 @@
...
+
+#include "../../util/net/fd_net_headers.h"
+#include "fd_net_common.h"
+
+/* FD_NET_SUCCESS, FD_NET_ERR_* error codes are defined in fd_net_common.h */
+
+/* fd_ip4_hdr_validate performs a series of checks on an IPv4 header. The firedancer
+   networking stack guarantees that it will only forward RX packets with an IP header
```
ripatel-fd: "forward" is ambiguous. It could mean IP forwarding to a remote host. 
Would clarify that this means forwarding to a downstream app tile.

src/disco/net/fd_net_checks.h
```diff
@@ -0,0 +1,179 @@
...
+
+   hdr points to the UDP header.
+   udp_sz is the number of bytes from beginning of UDP header to end of packet.
+
+   Returns FD_NET_SUCCESS if valid, or FD_NET_ERR_INVAL_UDP_HDR if the header is invalid:
+   - net_len and/or udp_sz is impossibly small
+   - net_len is larger than udp_sz */
+
+FD_FN_PURE static inline int
+fd_udp_hdr_validate( fd_udp_hdr_t const * hdr,
```
ripatel-fd:
```suggestion
fd_udp_hdr_check_len( fd_udp_hdr_t const * hdr,
```

src/disco/net/fd_net_checks.h
```diff
@@ -0,0 +1,179 @@
...
+  ulong iplen = FD_IP4_GET_LEN( *hdr );
+  if( FD_UNLIKELY( (iplen<20UL) | (iplen>ip4_sz) ) ) return FD_NET_ERR_INVAL_IP4_HDR;
+
+  uchar const proto = hdr->protocol;
+  if( FD_UNLIKELY( !fd_ulong_extract_bit( proto_allow_mask, proto ) ) ) return FD_NET_ERR_DISALLOW_IP_PROTO;
+
+  if( proto==FD_IP4_HDR_PROTOCOL_GRE ) {
+    ulong overhead = iplen + sizeof(fd_gre_hdr_t);
+    if( FD_UNLIKELY( overhead>ip4_sz ) ) return FD_NET_ERR_INVAL_GRE_HDR;
+  }
```
ripatel-fd: I would remove the special casing on proto here and move it to `fd_gre_hdr_check_len`

src/disco/net/fd_net_checks.h
```diff
@@ -0,0 +1,179 @@
...
+   that has passed these checks.
+
+   hdr points to the IP header.
+   ip4_sz is the number of bytes from beginning of ip hdr to end of packet.
+   proto_allow_mask is a bitmask of allowed IP protocols. Must be one of FD_IP4_HDR_PROTO_MASK_*.
+
+   Returns FD_NET_SUCCESS (defined in fd_net_common.h) if valid, or FD_NET_ERR_* on failure. */
+
+FD_FN_PURE static inline int
+fd_ip4_hdr_validate( fd_ip4_hdr_t const * hdr,
```
ripatel-fd:
```suggestion
fd_ip4_hdr_check_len( fd_ip4_hdr_t const * hdr,
```

src/disco/net/xdp/fd_xdp_tile.c
```diff
@@ -896,6 +899,30 @@ after_frag( fd_net_ctx_t *      ctx,
...
+      FD_DTRACE_PROBE( net_tile_err_rx_noip );
+      ctx->metrics.rx_ip_inv_cnt++;
+      break;
+    case FD_NET_ERR_INVAL_IP4_HDR:
+    case FD_NET_ERR_DISALLOW_IP_PROTO:
+      ctx->metrics.rx_ip_inv_cnt++;
+      break;
+    case FD_NET_ERR_INVAL_UDP_HDR:
+      ctx->metrics.rx_udp_inv_cnt++;
+      break;
```
ripatel-fd: IMO it would be cleaner to turn the rx drop metrics into an enum. 
For example `net_rx_check{rx_check="udp_invalid"}`.
If you make the NET_ERR return codes positive you can directly index into the enum, avoiding the switch statement. You can then also just use FD_MCNT_ENUM_COPY to copy the whole array into the metrics region.

src/disco/net/fd_net_checks.h
```diff
@@ -0,0 +1,179 @@
...
+   Returns 1 on success and 0 on failure */
+
+static inline int
+fd_ip4_udp_hdr_strip( uchar const *         data,
+                      ulong                 data_sz,
+                      uchar ** const        opt_payload,
+                      ulong *               opt_payload_sz,
+                      fd_eth_hdr_t ** const opt_eth,
+                      fd_ip4_hdr_t ** const opt_ip4,
+                      fd_udp_hdr_t ** const opt_udp ) {
```
ripatel-fd, line 176: Could you add a fuzzer for this function?

The fuzzer should pass random inputs to `data` / `data_sz`, and then validate that:
- `opt_payload` and `opt_payload_sz` are in bounds of `data` and `data_sz`
- `opt_eth` is in bounds
- `opt_ip4` is in bounds
- `opt_udp` is in bounds

ripatel-fd (changes requested):

src/util/net/fd_net_headers.h
akhinvasara-jumptrading: I added them so that users of any net headers can just include fd_net_headers, as the name implies. But if you would still prefer all users to individually include the particular headers they need, I'll revert - lmk?

src/disco/net/fd_net_checks.h
akhinvasara-jumptrading: It additionally checks version and that the protocol is allowed - does this check_len name still make sense? I'll stage the change for now, but don't love it.

src/disco/net/xdp/fd_xdp_tile.c
akhinvasara-jumptrading: Def makes sense - I'd considered that, but left it as is to minimize operator-facing changes. I'll switch it over, thanks. Just going to squash both commits together.

src/disco/net/fd_net_checks.h
akhinvasara-jumptrading: A full fuzz target in the style of fuzz_quic_wire? Or a unit test that that does a bunch of random sampling?

PR #7658, opened 2026-01-03
net: helpers for fd_xdp_ring_t

akhinvasara-jumptrading:
This PR factors out the cached fseq logic into fd_xdp_ring_{empty,full} for use in the xdp tile.  It adds extensive unit tests for those helpers, and for net_tx_ready (which was slightly refactored for cleaner testing).

This PR also fixes the logic inversion bug reported in https://github.com/firedancer-io/auditor-internal/issues/360.

### DIFF
--- a/src/disco/net/xdp/fd_xdp_tile.c
+++ b/src/disco/net/xdp/fd_xdp_tile.c
@@ -397,27 +397,19 @@ net_check_gre_interface_exists( fd_net_ctx_t * ctx ) {
 }
 
 
-/* net_tx_ready returns 1 if the current XSK is ready to submit a TX send
-   job.  If the XSK is blocked for sends, returns 0.  Reasons for block
-   include:
-   - No XSK TX buffer is available
-   - XSK TX ring is full */
+/* net_tx_ready returns 1 if we can submit a job to this TX ring, and 0 otherwise.
+   Reasons for block include:
+   - No TX buffer is available (free ring empty)
+   - TX ring is full
+
+   tx_ring: pointer to the XDP TX ring
+   free_ring: pointer to the free TX ring */
 
 static int
-net_tx_ready( fd_net_ctx_t * ctx,
-              uint           xsk_idx ) {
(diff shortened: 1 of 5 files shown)


src/disco/net/xdp/fd_xdp_tile.c
```diff
@@ -1135,7 +1104,11 @@ net_rx_event( fd_net_ctx_t * ctx,
     FD_LOG_CRIT(( "mcache corruption detected: chunk=%u chunk0=%u wmark=%u",
                   freed_chunk, ctx->umem_chunk0, ctx->umem_wmark ));
   }
-  ulong freed_off = (freed_chunk - ctx->umem_chunk0)<<FD_CHUNK_LG_SZ;
+
+  uint  fill_prod  = fill_ring->cached_prod;
+  uint  fill_mask  = (fill_ring->depth)-1U;
+  ulong frame_mask = FD_NET_MTU - 1UL;
```
ripatel-fd, line 1112: Can you add a static assertion checking that FD_NET_MTU is a power of 2?

ripatel-fd (approved):

src/waltz/xdp/fd_xsk.h
```diff
@@ -153,6 +153,40 @@ struct __attribute__((aligned(64UL))) fd_xdp_ring {
...
+  }
+  return ring->cached_prod == ring->cached_cons;
+}
+
+/* fd_xdp_ring_full returns 1 if the ring is full, 0 otherwise.
+   Assumes caller is the producer to this ring (fill, tx). */
+static inline int
+fd_xdp_ring_full( fd_xdp_ring_t * ring ) {
+  /* If potentially stale cached_cons says we have more space,
+      it's definitely not full. Else, refresh cached fseq. */
```
ripatel-fd: "fseq" is not a thing in XDP. This is a module in src/tango/fseq that is unused in the xdp library

ripatel-fd (dismissed):
I think the fd_xsk.h API is a bit much, I would have just done the comparisons in line. But looks extensively tested, so I'm happy with it

PR #7659, opened 2026-01-05
inflight: match layout/append alignment

two-heart:
not a problem in practice but still a mismatched value 8 vs 128


ripatel-fd:
This code is fine as is

PR #7665, opened 2026-01-05
fuzz: break instr harness dependency on funk

ibhatt-jumptrading:


src/flamenco/runtime/tests/fd_instr_harness.c
```diff
@@ -217,21 +201,16 @@ fd_solfuzz_pb_instr_ctx_create( fd_solfuzz_runner_t *                runner,
...
-    fd_txn_account_t * acc = &accts[i];
-    if ( !fd_executor_pubkey_is_bpf_loader( fd_txn_account_get_owner( acc ) ) ) {
+    fd_account_meta_t * meta = txn_out->accounts.metas[i];
+    if( !fd_executor_pubkey_is_bpf_loader( fd_type_pun( meta->owner ) ) ) {
       continue;
     }
 
-    fd_account_meta_t const * meta = fd_txn_account_get_meta( acc );
-    if( meta == NULL ) {
+    if( meta==NULL ) {
```
mjain-jump: after your changes i think this is always false so everything in here is probably dead code

ripatel-fd (approved):
LGTM

PR #7667, opened 2026-01-05
tower: pass vote acct info to gui

jherrera-jump:


src/choreo/tower/fd_tower.h
```diff
@@ -448,6 +448,13 @@ struct fd_tower_vote {
 };
 typedef struct fd_tower_vote fd_tower_vote_t;
 
+struct fd_tower_vote_lat {
+  ulong slot; /* vote slot */
+  ulong conf; /* confirmation count */
+  ulong latency; /* optional. ULONG_MAX by default */
+};
+typedef struct fd_tower_vote_lat fd_tower_vote_lat_t;
```
lidatong: i think you won't actually need this type if you instead make the anonymous struct in `votes_v3` a named struct (can just lift it out of `fd_voter` so `fd_voter_vote_v3` or something)

src/discof/tower/fd_tower_tile.h
```diff
@@ -67,6 +68,14 @@ struct fd_tower_slot_done {
...
   ulong vote_txn_sz;
   uchar vote_txn[ FD_TPU_MTU ];
+
+  /* The latest balance in lamports of our vote account, or ULONG_MAX if
+     our account is not found. */
+  ulong vote_acct_bal;
+
+  /* Our current on-chain tower with vote latencies optionally included. */
+  ulong tower_cnt;
+  fd_tower_vote_lat_t tower[ FD_TOWER_VOTE_MAX ];
```
lidatong: `fd_voter_vote_v3`

lidatong (commented):
looks reasonable, just the comment about naming the struct in `fd_voter` vs. adding a new type to api

lidatong (approved):

PR #7669, opened 2026-01-06
xdp: lock xsk descriptor rings and exclude from core dumps

tristan-carter:
Resolves TODO and brings the allocation of memory for the xsk descriptor rings up to standard with fd_shmem.

Benchmarked with fddev on AMD EPYC 9254P (24 cores, set to min 2.9GHz) with 384GB RAM (warmed for 20 mins).

Mean TPS over 300 seconds:
Before: 119,924
After: 119,737

Avoided including fd_numa_mlock() from fd_shmem_private.h (and instead reimplemented its functionality) to avoid including a private header from a separate fd component. Chose not to refactor the given shmem function into a public header to avoid heavily tweaking shmem.

### DIFF
--- a/src/disco/topo/fd_topo.c
+++ b/src/disco/topo/fd_topo.c
@@ -235,13 +235,12 @@ fd_topo_tile_extra_normal_pages( fd_topo_tile_t const * tile ) {
       /* tx ring */
       xsk_rings_sz_bytes += tile->xdp.xdp_tx_queue_size * xdp_desc_sz_bytes;
 
-      /* completion ring */ 
+      /* completion ring */
       xsk_rings_sz_bytes += tile->xdp.xdp_tx_queue_size * xdp_address_sz_bytes;
       /* free ring */
       xsk_rings_sz_bytes += tile->xdp.free_ring_depth   * xdp_address_sz_bytes;
 
-      /* Rounds up number of pages */
-      key_pages += ( ( xsk_rings_sz_bytes + FD_SHMEM_NORMAL_PAGE_SZ - 1UL ) / FD_SHMEM_NORMAL_PAGE_SZ );
+      key_pages += fd_ulong_align_up( xsk_rings_sz_bytes, FD_SHMEM_NORMAL_PAGE_SZ ) / FD_SHMEM_NORMAL_PAGE_SZ;
 
       /* All 4 rings must store a ring header. This is 320 bytes
          per ring as of linux v6.18.3, however could change in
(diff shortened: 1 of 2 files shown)


ripatel-fd (commented):
Thanks for this contribution. This is a good change, I'm surprised that the kernel doesn't mlock XDP rings already: https://elixir.bootlin.com/linux/v6.18.3/source/net/xdp/xsk_queue.c#L49

Before merging this, I'll need to double check the mlock rlimits set by the sandbox, but I think it should be fine.

ripatel-fd:
@mmcgee-jump, can you sanity check whether we need to explicitly track XDP rings in the memlock reservations?

mmcgee-jump:
We do yeah, to make sure the mlock reservations are correct. You can either put these in a workspace or just manually add the bytes to the fd_topo.c memlock_bytes thing

tristanx86:
> We do yeah, to make sure the mlock reservations are correct. You can either put these in a workspace or just manually add the bytes to the fd_topo.c memlock_bytes thing

Thanks for pointing this out, to avoid including linux headers into fd_topo.c I have manually done the calculation based on ABI constants and added lots of extra room where constants are not in ABI.

I have assumed the size of the descriptors to always be 16 bytes based on https://elixir.bootlin.com/linux/v6.18.3/source/include/uapi/linux/if_xdp.h#L165 being in the ABI.

I have assumed the size of each ring header to be one 4KB page since the size of these headers could change in the future, although are currently 320 bytes each as of linux v6.18.3 https://elixir.bootlin.com/linux/v6.18.3/source/net/xdp/xsk_queue.h#L16, allowing for a safe margin.

src/disco/topo/fd_topo.c
```diff
@@ -221,6 +221,35 @@ fd_topo_tile_extra_normal_pages( fd_topo_tile_t const * tile ) {
...
+      /* tx ring */
+      xsk_rings_sz_bytes += tile->xdp.xdp_tx_queue_size * xdp_desc_sz_bytes;
+
+      /* completion ring */ 
+      xsk_rings_sz_bytes += tile->xdp.xdp_tx_queue_size * xdp_address_sz_bytes;
+      /* free ring */
+      xsk_rings_sz_bytes += tile->xdp.free_ring_depth   * xdp_address_sz_bytes;
+
+      /* Rounds up number of pages */
+      key_pages += ( ( xsk_rings_sz_bytes + FD_SHMEM_NORMAL_PAGE_SZ - 1UL ) / FD_SHMEM_NORMAL_PAGE_SZ );
```
ripatel-fd: You can use `fd_ulong_align_up` here

tristanx86: Thanks for this suggestion, I have switched to using this now and removed some trailing whitespace

ripatel-fd (approved):

PR #7689, opened 2026-01-07
gossip: update shred_version filtering

amass-jump:
See https://github.com/anza-xyz/agave/pull/5659


src/discof/gossip/fd_gossvf_tile.c
```diff
@@ -437,7 +437,7 @@ static inline int
 is_entrypoint( fd_gossvf_tile_ctx_t * ctx,
                fd_ip4_port_t          addr ) {
   for( ulong i=0UL; i<ctx->entrypoints_cnt; i++ ) {
-    if( FD_UNLIKELY( addr.addr==ctx->entrypoints[ i ].addr && addr.port==ctx->entrypoints[ i ].port ) ) return 1;
+    if( FD_UNLIKELY( addr.l==ctx->entrypoints[ i ].l ) ) return 1;
```
mmcgee-jump: IIRC we have avoided this since the padding bytes may not match, not sure if they are expected to always be zeroed

PR #7693, opened 2026-01-07
ed25519: more avx512 inconsistencies

0x0ece:
This essentially restores the AVX512 code before https://github.com/firedancer-io/firedancer/pull/1311 because of subtle details that were not properly addressed.
(There's a minor detail on the constant `FD_R43X6_QUAD_1112d` (previously `FD_R43X6_QUAD_1122d`, but that should make no difference.)


src/ballet/ed25519/test_ed25519.c
```diff
@@ -5,6 +5,9 @@
 #include "test_ed25519_wycheproof.c"
 #include "test_ed25519_cctv.c"
 
+#define FD_ED25519_UNIT_TESTS
```
jvarela-jump, line 8: Maybe add ```#undef FD_ED25519_UNIT_TESTS``` at the end, just for completeness?

src/ballet/ed25519/fd_ed25519_user.c
```diff
@@ -277,7 +277,7 @@ int fd_ed25519_verify_batch_single_msg( uchar const   msg[], /* msg_sz */
 
     /* Check public key and point r */
     if( FD_UNLIKELY( res ) ) {
-      return res == 1 ? FD_ED25519_ERR_PUBKEY : FD_ED25519_ERR_SIG;
+      return res == -1 ? FD_ED25519_ERR_PUBKEY : FD_ED25519_ERR_SIG;
```
jvarela-jump, line 280: I think `fd_ed25519_verify` is missing the same change?

PR #7694, opened 2026-01-07
gossip: remove stake filter in ping tracker

amass-jump:
See https://github.com/anza-xyz/agave/pull/7708


src/flamenco/gossip/fd_ping_tracker.c
```diff
@@ -305,27 +303,6 @@ fd_ping_tracker_track( fd_ping_tracker_t * ping_tracker,
     peer_map_ele_insert( ping_tracker->peers, peer, ping_tracker->pool );
     lru_list_ele_push_tail( ping_tracker->lru, peer, ping_tracker->pool );
   } else {
-    if( FD_LIKELY( peer_stake>=1000000000UL || is_entrypoint( ping_tracker, peer_address ) ) ) {
```
mmcgee-jump, line 308: You need to keep the is_entrypoint branch no?

mmcgee-jump (changes requested):
.

src/flamenco/gossip/fd_ping_tracker.c
amass-jump: I thought it would be better to remove it. It seemed like a strange and possibly dangerous case to have a node that starts off not being an entrypoint, then changes it's gossip address and starts being trusted as a entrypoint later.

Besides as the name suggests, the entrypoint special case for pinging is only really needed at startup so that we can seed the process of filling the table with ContactInfos. After startup this would no longer matter.

Thoughts?

mmcgee-jump: Well the purpose of the ping is just to prevent a DDoS, but you can't DDoS an IP address that had to be manually specified (as an entrypoint), so there's no reason to ping it.

It does appear to simplify the logic to remove but I can't easily verify the invariants in my head anymore. Previously, there are clear conditions (e.g. validity change always generates a change callback, which isn't true anymore) so it's a bit harder to reason about.

amass-jump: That's reasonable, I added it back in. I have another change (not PR'd yet) that cleans all this up a bit but added it back for now

PR #7695, opened 2026-01-07
fuzz, txn: fix how rollback accounts are handled

mjain-jump:


src/flamenco/runtime/tests/fd_txn_harness.c
```diff
@@ -473,27 +473,37 @@ fd_solfuzz_pb_txn_run( fd_solfuzz_runner_t * runner,
     }
 
     /* If the transaction is a fees-only transaction, we have to create rollback accounts to iterate over and save. */
+    fd_pubkey_t *         pubkeys_to_save  = txn_out->accounts.keys;
     fd_account_meta_t * * accounts_to_save = txn_out->accounts.metas;
     ulong                 accounts_cnt     = txn_out->accounts.cnt;
     if( txn_out->err.is_fees_only ) {
-      accounts_to_save = fd_spad_alloc( runner->spad, alignof(fd_txn_account_t), sizeof(fd_txn_account_t) * 2 );
+      pubkeys_to_save  = fd_spad_alloc( runner->spad, alignof(fd_pubkey_t), sizeof(fd_pubkey_t) * 2 );
```
ibhatt-jumptrading: don't we want to save the rollbacks when the transaction isn't just fees only and fails w an instruction error?

src/flamenco/runtime/tests/fd_txn_harness.c
```diff
@@ -473,27 +473,37 @@ fd_solfuzz_pb_txn_run( fd_solfuzz_runner_t * runner,
...
+      accounts_to_save = fd_spad_alloc( runner->spad, alignof(fd_account_meta_t *), sizeof(fd_account_meta_t *) * 2 );
```
ibhatt-jumptrading: you can also stack allocate here

src/flamenco/runtime/tests/fd_txn_harness.c
mjain-jump: they already are - fees-only transactions only happen when `fd_executor_load_transaction_accounts` fails. if it makes it into `fd_execute_txn` then this codepath isnt hit regardless of instr error or not

src/flamenco/runtime/tests/fd_txn_harness.c
mjain-jump: fixed

src/flamenco/runtime/tests/fd_txn_harness.c
```diff
@@ -472,28 +472,41 @@ fd_solfuzz_pb_txn_run( fd_solfuzz_runner_t * runner,
...
 
     /* Capture borrowed accounts */
     for( ulong j=0UL; j<accounts_cnt; j++ ) {
-      fd_account_meta_t * meta = accounts_to_save[j];
-      fd_pubkey_t * pubkey     = &txn_out->accounts.keys[j];
+      fd_pubkey_t *       pubkey = &pubkeys_to_save[j];
+      fd_account_meta_t * meta   = accounts_to_save[j];
 
-      if( !( fd_runtime_account_is_writable_idx( txn_in, txn_out, runner->bank, (ushort)j ) || j==FD_FEE_PAYER_TXN_IDX ) ) continue;
+      if( !( txn_out->err.is_fees_only ||                                                                /* Capture all accounts in fees-only transactions */
```
ibhatt-jumptrading: doesn't the fee payer and the nonce have to writable here

mjain-jump: the indices passed into `fd_runtime_account_is_writable_idx` dont correspond to the rollback account's indices in the transaction, so the writable check would be incorrect. i refactored the code to be a bit more readable

PR #7736, opened 2026-01-09
accdb: anonymous references

ripatel-fd:
So far, the accdb_ref APIs could only reference funk- and vinyl-managed accounts. 

Plenty of accounts in Firedancer are custom-allocated.
This patch adds a `fd_accdb_init_{ro,rw}_nodb()` that allows users to create accdb-compatible account references to externally-managed accounts.

Additionally, a `accdb_v0` API is added which is a minimal accdb implementation for testing, with `O(n)` query complexity, and without fork awareness.

### DIFF
--- a/src/flamenco/accdb/fd_accdb_impl_v0.c
+++ b/src/flamenco/accdb/fd_accdb_impl_v0.c
@@ -170,10 +170,15 @@ fd_accdb_user_v0_open_ro( fd_accdb_user_t *         accdb_,
       FD_LOG_CRIT(( "accdb_user_v0_open_ro failed: account %s is currently in use", address_b58 ));
     }
     rec->reader_cnt++;
-    found = ro;
-    FD_STORE( fd_pubkey_t, found->address, rec->key );
-    found->meta = &rec->meta;
+
+    *ro = (fd_accdb_ro_t) {0};
+    FD_STORE( fd_pubkey_t, ro->ref->address, rec->key );
+    ro->ref->accdb_type = FD_ACCDB_TYPE_V0;
+    ro->ref->user_data  = 0UL;
+    ro->meta            = &rec->meta;
+
     accdb->base.ro_active++;
+    found = ro;
   }
 
   fd_rwlock_unwrite( &v0->lock );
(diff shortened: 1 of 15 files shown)


src/flamenco/accdb/fd_accdb_impl_v0.c
```diff
@@ -0,0 +1,327 @@
...
+
+FD_FN_CONST ulong
+fd_accdb_v0_align( void ) {
+  return alignof(fd_accdb_v0_t);
+}
+
+ulong
+fd_accdb_v0_footprint( ulong rec_cnt ) {
+  ulong rec_sz;
+  if( FD_UNLIKELY( !__builtin_umull_overflow( rec_cnt, sizeof(fd_accdb_v0_rec_t), &rec_sz ) ) ) return 0UL;
```
topointon-jump: Is the negation correct here? `__builtin_umull_overflow` returns true if there is an overflow

src/flamenco/accdb/fd_accdb_impl_v0.c
```diff
@@ -0,0 +1,327 @@
...
+                          fd_accdb_rw_t *           rw,
+                          fd_funk_txn_xid_t const * xid,
+                          void const *              address,
+                          ulong                     data_max,
+                          int                       flags ) {
+  fd_accdb_user_v0_t * accdb = (fd_accdb_user_v0_t *)accdb_;
+  fd_accdb_v0_t *      v0    = accdb->v0;
+  (void)xid;
+
+  fd_accdb_rw_t * found = NULL;
```
topointon-jump: Looks like we never assign to this?

ibhatt-jumptrading (dismissed):

src/flamenco/accdb/fd_accdb_impl_v0.c
```diff
@@ -0,0 +1,327 @@
...
+    FD_LOG_CRIT(( "accdb_user_v0_close_rw failed: account %s not found", address_b58 ));
+  }
+  fd_accdb_v0_rec_t * rec = &v0->rec[ idx ];
+
+  if( FD_UNLIKELY( !rec->writer_cnt ||
+                   !user->base.rw_active ||
+                   user->base.ro_active ) ) {
+    FD_LOG_CRIT(( "accdb_user_v0_close_rw failed: invalid ref count detected" ));
+  }
+  rec->writer_cnt--;
```
topointon-jump, line 283: Do we ever increment this? Sorry if I missed it

src/flamenco/accdb/fd_accdb_impl_v0.c
ripatel-fd: Good catch!

src/flamenco/accdb/fd_accdb_impl_v0.c
ripatel-fd: No - good catch

topointon-jump (approved):

PR #7771, opened 2026-01-11
sysvar: fix FD_UNLIKELY scope

two-heart:
for ==0 this is not a correctness problem, but it the opposite of the intended branch hint


ripatel-fd (approved):

ripatel-fd:
Thanks 👑

PR #7773, opened 2026-01-12
runtime: migrate bundle config read to accdb

riptl:
Also fixes a potential OOB read if the bundle config account is
malicious.

### DIFF
--- a/src/discof/replay/fd_replay_tile.c
+++ b/src/discof/replay/fd_replay_tile.c
@@ -1177,34 +1177,34 @@ maybe_become_leader( fd_replay_tile_t *  ctx,
   fd_bank_t *       bank = prepare_leader_bank( ctx, ctx->next_leader_slot, now_nanos, &ctx->reset_block_id, stem );
   fd_funk_txn_xid_t xid  = { .ul = { ctx->next_leader_slot, ctx->leader_bank->data->idx } };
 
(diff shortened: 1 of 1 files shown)


src/discof/replay/fd_replay_tile.c
```diff
@@ -1185,26 +1185,26 @@ maybe_become_leader( fd_replay_tile_t *  ctx,
...
        if it is the first time in an epoch. */
-    fd_txn_account_t tip_receiver_acc[1];
-    err = fd_txn_account_init_from_funk_readonly( tip_receiver_acc,
-                                                  (fd_hash_t *)tip_receiver->b,
-                                                  funk,
-                                                  &xid );
-    if( FD_LIKELY( !err ) ) {
+    fd_accdb_ro_t tip_receiver_acc[1];
+    if( FD_LIKELY( fd_accdb_open_ro( ctx->accdb, tip_receiver_acc, &xid, tip_receiver ) ) ) {
       memcpy( tip_receiver_owner, tip_receiver_acc->meta->owner, sizeof(fd_acct_addr_t) );
```
topointon-jump: nit but should we use `fd_accdb_ref_owner` here instead of `tip_receiver_acc->meta->owner`?

ripatel-fd: Yes, good catch

PR #7800, opened 2026-01-12
runtime: fix initial vote state cache fill

riptl:


ibhatt-jumptrading:
should probably be done for tower too

ripatel-fd:
> should probably be done for tower too

@ibhatt-jumptrading The tower fix requires a bit more work, which I'll do in a separate PR. 
This unblocks async replay, but tower is only required for live replay.

src/discof/replay/fd_replay_tile.c
```diff
@@ -1054,18 +1054,14 @@ init_after_snapshot( fd_replay_tile_t * ctx ) {
...
-          break;
-        }
-        if( FD_LIKELY( fd_accdb_peek_test( peek ) ) ) break;
-        FD_SPIN_PAUSE();
+      fd_accdb_ro_t ro[1];
+      if( FD_UNLIKELY( !fd_accdb_open_ro( ctx->accdb, ro, &xid, &vote_state->vote_account ) ) ) {
+        ctx->runtime_stack.vote_accounts.stale_accs[ stale_vote_acc_cnt++ ] = vote_state->vote_account;
+        FD_BASE58_ENCODE_32_BYTES( vote_state->vote_account.uc, acc_cstr );
+        FD_LOG_DEBUG(( "vote account %s from manifest is stale", acc_cstr ));
+        break;
```
ripatel-fd: bug ...

PR #7843, opened 2026-01-13
fix(notar,tower): map key null for slot number keys

lidatong:


src/disco/shred/fd_shred_tile.c
```diff
@@ -980,6 +980,10 @@ after_frag( fd_shred_ctx_t *    ctx,
         FD_STORE(uint, fd_chunk_to_laddr( ctx->shred_out_mem, ctx->shred_out_chunk ) + sz, nonce );
         sz += 4UL;
 
+        if( FD_LIKELY( !is_code ) ) {
+          FD_LOG_HEXDUMP_NOTICE(( "sending shred", shred, FD_SHRED_MIN_SZ ));
+        }
```
emwang-jump: delete

PR #7858, opened 2026-01-14
Make vinyl limit config more robust

riptl:
Fixes a few backtests on vinyl which fail due to too low limits.

- If the user sets max_cache_entries lower than the sum of quotas
  on all links, log a warning and increase the cache size
- run_ledger_backtest.sh: run with at least 1 million record map
  capacity

### DIFF
--- a/src/app/firedancer/topology.c
+++ b/src/app/firedancer/topology.c
@@ -1563,6 +1563,23 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
         FD_VINYL_IO_TYPE_UR : FD_VINYL_IO_TYPE_BD;
     tile->vinyl.uring_depth = config->firedancer.vinyl.io_uring.queue_depth;
 
+    /* Minimum bound for cache entry count */
+    ulong required_cache_entries = 0UL;
+    for( ulong i=0UL; i<(tile->uses_obj_cnt); i++ ) {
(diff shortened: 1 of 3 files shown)


ibhatt-jumptrading (dismissed):

src/app/firedancer/topology.c
```diff
@@ -1563,6 +1563,23 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
...
+      if( strcmp( rq_obj->name, "vinyl_rq" ) ) continue;
+
+      ulong quota_max;
+      FD_TEST( (quota_max = fd_pod_queryf_ulong( config->topo.props, ULONG_MAX, "obj.%lu.quota_max", rq_obj_id ))!=ULONG_MAX );
+      required_cache_entries += quota_max;
+    }
+    if( FD_UNLIKELY( required_cache_entries > tile->vinyl.vinyl_line_max ) ) {
+      tile->vinyl.vinyl_line_max = required_cache_entries;
+      FD_LOG_WARNING(( "insufficient [vinyl.max_cache_entries] (%lu) for config, increasing cache entry limit to %lu and continuing",
+                       tile->vinyl.vinyl_line_max, required_cache_entries ));
```
ripatel-fd, line 1579: this is buggy

PR #7873, opened 2026-01-15
restore: snapshot load lthash (vinyl)

jvarela-jump:
Adding lthash verification to the snapshot-load pipeline under vinyl.
Two new tiles are introduced in `src/discof/restore`:

- `snaplh` (the equivalent of `snapla` for funk): it computes lthash add (on all accounts loaded) and lthash sub (duplicates).
- `snaplv` (the equivalent of `snapls` for funk): it handles incoming requests from `snapwm`, manages and forwards requests to `snaplh` tiles, and forwards results to `snapct`.

Other main differences compared to the architecture under funk:
- `snaplh` is not directly in the state-machine path (although it does follow it).
- `snaplh` does not parse the snapshot; instead, it reads every bstream pair (account) that is being sent to vinyl.
- `snaplv` and `snaplh` are located farther inside the pipeline.


src/app/firedancer/config/default.toml
```diff
@@ -790,12 +790,20 @@ user = ""
...
     # error.
     snapla_tile_count = 4
 
+    # How many snapshot lthash add/sub tiles to run.  Snapshot lthash
+    # add/sub tiles compute the lthash of all accounts in the loaded
+    # snapshot in parallel under vinyl. Too few and snapshot loading
+    # might be delayed.  Once the lthash of the accounts in a snapshot
+    # is calculated, if it does not match the validator will abort with
+    # an error.  Preferably a power of two.
+    snaplh_tile_count = 8
```
ripatel-fd: @jvarela-jump can you unify this and just call this `snapshot_hash_tile_count` ?

These config options should be understandable to the user, and the user probably doesn't understand what the difference between "snapla" and "snaplh" is, even after they read the documentation

ripatel-fd (approved):

src/app/firedancer/config/default.toml
jvarela-jump: Good point. Done.

PR #7875, opened 2026-01-15
repair: verify fec chaining, duplicate confirmation repair, equivocation

emwang-jump:
Closes [#7830](https://github.com/firedancer-io/firedancer/issues/7830)

### DIFF
--- a/src/app/firedancer-dev/commands/repair.c
+++ b/src/app/firedancer-dev/commands/repair.c
@@ -257,7 +257,6 @@ repair_topo( config_t * config ) {
 
   FOR(net_tile_cnt)    fd_topob_tile_in(  topo, "repair",  0UL,          "metric_in", "net_repair",    i,            FD_TOPOB_UNRELIABLE, FD_TOPOB_POLLED   ); /* No reliable consumers of networking fragments, may be dropped or overrun */
   /**/                 fd_topob_tile_in(  topo, "repair",  0UL,          "metric_in", "gossip_out",    0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED   );
-  /**/                 fd_topob_tile_in(  topo, "repair",  0UL,          "metric_in", "replay_epoch",  0UL,          FD_TOPOB_UNRELIABLE, FD_TOPOB_POLLED   );
                        fd_topob_tile_in(  topo, "repair",  0UL,          "metric_in", "snapin_manif",  0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED   );
   FOR(shred_tile_cnt)  fd_topob_tile_in(  topo, "repair",  0UL,          "metric_in", "shred_out",     i,            FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED   );
   FOR(shred_tile_cnt)  fd_topob_tile_out( topo, "repair", 0UL,                        "repair_shred",  i                                                    );
@@ -910,9 +909,8 @@ repair_cmd_fn_requests( args_t *   args,
       fd_forest_ref_t * req = fd_forest_reqslist_iter_ele( iter, dlist, pool );
       fd_forest_blk_t * blk = fd_forest_pool_ele( fd_forest_pool( forest ), req->idx );
 
-      printf("%-15lu %-12u %-12u %-12u %-20ld %-12u %-10u\n",
+      printf("%-15lu %-12u %-12u %-20ld %-12u %-10u\n",
               blk->slot,
-              blk->consumed_idx,
               blk->buffered_idx,
               blk->complete_idx,
               blk->first_shred_ts,
--- a/src/discof/forest/fd_forest.c
+++ b/src/discof/forest/fd_forest.c
@@ -6,6 +6,9 @@ static void ver_inc( ulong ** ver ) {
 
 #define VER_INC ulong * ver __attribute__((cleanup(ver_inc))) = fd_forest_ver( forest ); ver_inc( &ver )
 
+static fd_hash_t empty_mr   = { .ul = { 0, 0, 0, 0 } };
+static fd_hash_t invalid_mr = { .ul = { ULONG_MAX, ULONG_MAX, ULONG_MAX, ULONG_MAX } };
+
 void *
 fd_forest_new( void * shmem, ulong ele_max, ulong seed ) {
   FD_TEST( fd_ulong_is_pow2( ele_max ) );
@@ -488,7 +491,6 @@ advance_consumed_frontier( fd_forest_t * forest, ulong slot, ulong parent_slot )
       consumed_remove( forest, fd_forest_pool_idx( pool, head ) );
       while( FD_LIKELY( child ) ) { /* add children to consumed frontier */
         consumed_insert( forest, fd_forest_pool_idx( pool, child ) );
-
         fd_forest_deque_push_tail( queue, fd_forest_pool_idx( pool, child ) );
         child = fd_forest_pool_ele( pool, child->sibling );
       }
@@ -534,13 +536,13 @@ acquire( fd_forest_t * forest, ulong slot, ulong parent_slot ) {
   blk->confirmed   = 0;
   blk->consumed    = 0;
 
-  blk->consumed_idx = UINT_MAX;
   blk->buffered_idx = UINT_MAX;
   blk->complete_idx = UINT_MAX;
 
   fd_forest_blk_idxs_null( blk->idxs );
   blk->lowest_verified_fec = UINT_MAX;
   memset( blk->merkle_roots, 0, sizeof( blk->merkle_roots ) ); /* expensive*/
+  blk->confirmed_bid = empty_mr;
 
   blk->est_buffered_tick_recv = 0;
 
@@ -567,7 +569,18 @@ fd_forest_blk_insert( fd_forest_t * forest, ulong slot, ulong parent_slot ) {
   FD_TEST( slot > fd_forest_root_slot( forest ) ); /* caller error - inval */
 # endif
   fd_forest_blk_t * ele = query( forest, slot );
-  if( FD_LIKELY( ele ) ) { return ele; }
+  if( FD_LIKELY( ele ) ) {
+    // potentially may need to update the parent_slot, if this
+    // this was a sentinel block that was created for a confirmed msg
+    if( FD_UNLIKELY( parent_slot != ele->parent_slot ) ) {
+      ele->parent_slot = parent_slot;
+      subtrees_orphaned_remove( forest, slot ); // if this is a sentinel block, then it must be in subtrees
+    } else {
+      return ele;
+    }
+  } else {
+    ele = acquire( forest, slot, parent_slot );
+  }
 
   fd_forest_ancestry_t * ancestry = fd_forest_ancestry( forest );
   fd_forest_frontier_t * frontier = fd_forest_frontier( forest );
@@ -583,8 +596,6 @@ fd_forest_blk_insert( fd_forest_t * forest, ulong slot, ulong parent_slot ) {
   ulong                  null     = fd_forest_pool_idx_null( pool );
 
   fd_forest_blk_t * parent = NULL;
-
-  ele = acquire( forest, slot, parent_slot );
 
   if(        FD_LIKELY  ( parent = fd_forest_ancestry_ele_query ( ancestry, &parent_slot, NULL, pool ) ) ) { /* parent is in ancestry, ele makes new frontier */
     fd_forest_frontier_ele_insert( frontier, ele, pool );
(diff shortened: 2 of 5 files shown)


src/app/firedancer-dev/commands/repair.c
```diff
@@ -164,6 +166,7 @@ repair_topo( config_t * config ) {
   /**/                 fd_topob_link( topo, "snapin_manif", "snapin_manif", 2UL,                                      sizeof(fd_snapshot_manifest_t), 1UL );
 
   /**/                 fd_topob_link( topo, "genesi_out",  "genesi_out",    2UL,                                        128,     1UL );
```
lidatong:
```suggestion
  /**/                 fd_topob_link( topo, "genesi_out",   "genesi_out",   2UL,                                        128,     1UL );
```

src/app/firedancer/topology.c
```diff
@@ -866,10 +866,10 @@ fd_topo_initialize( config_t * config ) {
...
   }
 
-  /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "genesi_out",    0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
-  /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "gossip_out",    0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
-  /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "tower_out",     0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
-  /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "replay_epoch",  0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
+  /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "genesi_out",   0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
+  /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "gossip_out",   0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
+  /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "tower_out",    0UL,          FD_TOPOB_UNRELIABLE, FD_TOPOB_POLLED );
```
lidatong: i remember we discussed this but it needs to be unreliable because it can't keep up?

src/disco/gui/fd_gui_tile.c
```diff
@@ -466,14 +466,14 @@ after_frag( fd_gui_ctx_t *      ctx,
     case IN_KIND_SHRED_OUT: {
       FD_TEST( ctx->is_full_client );
       long tsorig_nanos = ctx->ref_wallclock + (long)((double)(fd_frag_meta_ts_decomp( tsorig, fd_tickcount() ) - ctx->ref_tickcount) / ctx->tick_per_ns);
-      if( FD_LIKELY( sz!=0 && sz!=FD_SHRED_DATA_HEADER_SZ + FD_SHRED_MERKLE_ROOT_SZ * 2 + sizeof(int) ) ) {
-        ulong slot = fd_disco_shred_out_shred_sig_slot( sig );
-        int is_turbine = fd_disco_shred_out_shred_sig_is_turbine( sig );
-        ulong shred_idx  = fd_disco_shred_out_shred_sig_shred_idx( sig );
+      if( FD_LIKELY( sz!=0 && fd_disco_shred_out_msg_type( sig )==FD_SHRED_OUT_MSG_TYPE_SHRED ) ) {
```
lidatong, line 465: can you file an issue to follow-up PR to remove the dependency on parsing sigs?

src/discof/forest/fd_forest.c
```diff
@@ -667,13 +676,58 @@ fd_forest_blk_insert( fd_forest_t * forest, ulong slot, ulong parent_slot ) {
...
+                             fd_hash_t   * cmr ) {
   VER_INC;
   FD_TEST( shred_idx <= FD_SHRED_IDX_MAX );
   fd_forest_blk_t * ele = query( forest, slot );
 # if FD_FOREST_USE_HANDHOLDING
   if( FD_UNLIKELY( !ele ) ) FD_LOG_ERR(( "fd_forest: fd_forest_data_shred_insert: ele %lu is not in the forest. data_shred_insert should be preceded by blk_insert", slot ));
 # endif
+
+  /* Pre-filtering on merkle root.
+     If we have knowledge of the canonical merkle root, we can reject
```
```suggestion
     If we have knowledge of the confirmed merkle root, we can reject
```

src/discof/forest/fd_forest.c
```diff
@@ -667,13 +676,58 @@ fd_forest_blk_insert( fd_forest_t * forest, ulong slot, ulong parent_slot ) {
...
+     shreds that don't match it.  Else, we'll accept any and all shreds,
+     and invalidating the merkle root if we see more than 1 version of
+     the FEC. */
+  uint fec_idx = fec_set_idx / 32UL;
```
lidatong, line 723: double-check we never let through `fec_set_idx >= 1024` with new 32 fec resolver to avoid UB overflow below

src/discof/forest/fd_forest.c
```diff
@@ -706,21 +760,54 @@ fd_forest_blk_parent_update( fd_forest_t * forest, ulong slot, ulong parent_slot
...
+           we know we must've received all the data for version B!
+       (2) the second case is that maybe both version B and version A
+           get completed, one after the other.  In this case we've first
+           overwritten from { 0 } to B.  But when version A arrives,
+           what do we do ???? if A is the correct version but we ignore it,
+           when we chain verify down the line we'll evict B and try to repair
+           for A, but fec_resolver AINT gonna let it through! vice versa
+           if B is the correct version, but we choose to overwrite the fec when A arrive.
+           no way around it (unless) we ask shred to re-deliver ouch. or
+           we refactor so that we do fec-based forest ouch.
```
lidatong: file an issue for 2, to implement eviction of completed fec sets in fec resolver

src/discof/forest/fd_forest.c
```diff
@@ -706,21 +760,54 @@ fd_forest_blk_parent_update( fd_forest_t * forest, ulong slot, ulong parent_slot
...
+    FD_LOG_WARNING(( "fd_forest_fec_insert: fec_resolver a version of fec_set_idx %u we dont have recorded. current_mr %s, received_mr %s", fec_set_idx, mr_b58, mr_recv_b58 ));
+    /* there are two cases:
+
+       (1) the most common is that we've received several shreds from
+           version A, and all the shreds from version B.  In forest we
+           have recorded hash = { 0 } for this fec set because we've
+           received a mix of merkle roots, so we nulled the FEC set.
+           Then fec_resolver completes version B, and delivers.  We can
+           safely overwrite our recorded merkle root with B because
+           we know we must've received all the data for version B!
```
```suggestion
       (1) the first and common case is we receive a mix of shreds from equivocating FEC siblings A & B.  In forest we
           have recorded hash = { 0 } for this fec set because we've
           received a mix of merkle roots, so we nulled the FEC set.
           Let's say fec_resolver then completes version B, and delivers (a separate FEC completion msg from the original shred msgs we were tracking).  We can
           safely overwrite our recorded merkle root with B because
           we know we must've received all the data for version B!
```

src/discof/forest/fd_forest.c
```diff
@@ -706,21 +760,54 @@ fd_forest_blk_parent_update( fd_forest_t * forest, ulong slot, ulong parent_slot
...
+       (2) the second case is that maybe both version B and version A
+           get completed, one after the other.  In this case we've first
```
```suggestion
       (2) the second case is that we get two FEC completion msgs: one for both version A and B. They
           get completed, one after the other.  In this case we've first
```

src/discof/forest/fd_forest.h
```diff
@@ -16,6 +16,100 @@
    Forest constructs the ancestry tree backwards, and then repairs the
    tree forwards (using BFS). */
 
+/* Merkle root tracking.
+   For each FEC set in the slot, we track the first merkle root we have
+   received for that FEC set. It's stored in the merkle_roots array.
+   Then for any shred index that arrives after, the merkle root is
+   compared to the merkle root we have stored.
```
lidatong:
```suggestion
   For each FEC set in the slot, we record the merkle root of the first shred we
   receive in `.merkle_roots[ fec_set_idx / 32 ]`.
   Then for any shred in the same FEC inserted later, the merkle root of the new shred is
   compared to the merkle root we have stored.
```

src/discof/forest/fd_forest.h
```diff
@@ -16,6 +16,100 @@
...
+                            root to a null hash for later detection.
+
+  Note we don't verify the chain on each FEC arrival, because we can't
+  tell whether the CMR of the following FEC is incorrect or if the
+  current MR we have is incorrect.  We can only verify the chain when
+  we get a confirmation of a block_id.
+
+  Eventually one of two things happen:
+  1. We are able to complete version of the FEC with the merkle root we
+     have stored.  This is the regular case, and means we only saw one
```
lidatong:
```suggestion
     have stored.  This is the common case, and means we only saw one
```

src/discof/forest/fd_forest.h
```diff
@@ -16,6 +16,100 @@
...
+       need to repair (and we won't be making any new requests for that
+       FEC set).
+
+       We can't ignore shreds of FEC_B though, because it's difficult to
+       differentiate between a slot where we haven't finished repairing,
+       or a slot where we can't repair because the version we have is a
+       bad version and fake.  So merkle chaining verification can only
+       be performed on slots that have all the shreds received.
+
+  3. We receive some shreds from FEC_A, but get a completion for FEC_B.
```
lidatong:
```suggestion
  3. We receive some shreds from both FEC_A and FEC_B, but get a FEC completion for FEC_B.
```

src/discof/forest/fd_forest.h
```diff
@@ -16,6 +16,100 @@
...
+        overwritten from the null_hash to the FEC_B merkle root. TODO /
+        dbl check this case.
+
+  So unfortunately...because of case 2, we can no longer rely on the
+  condition that the slot will complete with all the FEC completions.
+  But we can still rely on that at lease some version of all the shreds
+  in the slot will arrive eventually.
+
+  As soon as we have a confirmed block id, we can verify the slot by
+  verifying the chain of merkle roots. backwards.  As the CMRs match,
```
lidatong:
```suggestion
  verifying the chain of merkle roots backwards.  As the CMRs correctly chain,
```

src/discof/forest/fd_forest.h
```diff
@@ -16,6 +16,100 @@
...
+  the verified bit on each FEC set is set to 1. If it doesn't match, we
```
lidatong:
```suggestion
  the verified bit on each FEC set is set to 1. If they don't chain, we
```

src/discof/forest/fd_forest.h
```diff
@@ -61,6 +160,17 @@ struct __attribute__((aligned(128UL))) fd_forest_blk {
   fd_forest_blk_idxs_t idxs[fd_forest_blk_idxs_word_cnt]; /* data shred idxs */
   fd_forest_blk_idxs_t cmpl[fd_forest_blk_idxs_word_cnt]; /* last shred idx of every FEC set that has been completed by shred_tile */
 
+  struct {
+    fd_hash_t mr;
+    fd_hash_t cmr;
+  } merkle_roots[ FD_FEC_BLK_MAX + 1 ]; /* +1 -> .cmr will be populated when the block id is confirmed */
+  fd_forest_merkle_t    merkle_recvd[fd_forest_merkle_word_cnt];
```
lidatong: do we need `merkle_recvd` if we have two sentinels for example merkle root 0 and {FFFFF} ?

src/discof/forest/fd_forest.h
```diff
@@ -61,6 +160,17 @@ struct __attribute__((aligned(128UL))) fd_forest_blk {
...
+  fd_forest_merkle_t merkle_verified[fd_forest_merkle_word_cnt];
```
lidatong: instead of merkle_verified can it track a `verified` cursor similar to `consumed`?

src/discof/forest/fd_forest.h
```diff
@@ -636,29 +756,50 @@ fd_forest_code_shred_insert( fd_forest_t * forest, ulong slot, uint shred_idx );
...
-fd_forest_fec_insert( fd_forest_t * forest, ulong slot, ulong parent_slot, uint last_shred_idx, uint fec_set_idx, int slot_complete, int ref_tick );
+fd_forest_fec_insert( fd_forest_t * forest,
+                      ulong slot,
+                      ulong parent_slot,
+                      uint last_shred_idx,
+                      uint fec_set_idx,
+                      int slot_complete,
+                      int ref_tick,
+                      fd_hash_t * mr,
+                      fd_hash_t * cmr );
```
lidatong: spacing

src/app/firedancer/topology.c
emwang-jump: It could probably keep up, but there's kind of no need to because later confirmations will retroactively confirm earlier ones we miss

emwang-jump: I'll leave a comment

lidatong: would be inclined to go the other way and keep it reliable unless it has trouble keeping up actually -- one less consideration when debugging any potential issues

src/discof/forest/fd_forest.h
emwang-jump: Yeah good point, I'll do that in a follow up PR

src/discof/forest/fd_forest.h
emwang-jump: So true :"> will also do in a follow up PR

src/discof/forest/fd_forest.c
emwang-jump: Yeup it's impossible, `fec_resolver` rejects any shred with index > than 32 * 1024 (in the default config). It would also fail the FD_TEST above rather than cause overflow

src/app/firedancer/topology.c
emwang-jump: makes sense, perf-ed and ran the monitoring tools on it and the confirmation verifying seems to be negligible, so making this reliable

PR #7876, opened 2026-01-15
snapshots: add max_retry_abort

cali-jumptrading:


src/discof/restore/fd_snapin_tile.c
```diff
@@ -599,20 +599,24 @@ handle_control_frag( fd_snapin_tile_t *  ctx,
         ctx->metrics.accounts_loaded   = ctx->metrics.full_accounts_loaded;
         ctx->metrics.accounts_replaced = ctx->metrics.full_accounts_replaced;
         ctx->metrics.accounts_ignored  = ctx->metrics.full_accounts_ignored;
+
+        fd_funk_txn_xid_t incremental_xid = { .ul={ LONG_MAX, LONG_MAX } };
```
cali-jumptrading, line 603: Fixes the fd_accdb failure when retrying an incremental snapshot. On retrying an incremental snapshot, we only get a fail control message and then an init control message, so the logic to make a new incremental txn should be in response to init instead of the next control message.

src/app/firedancer/config/default.toml
```diff
@@ -346,6 +346,16 @@ user = ""
...
 
+    # Controls how many times the validator will retry loading a full or
+    # incremental snapshot before aborting.  Reasons for retrying a
+    # snapshot load include snapshot corruption, slow download speed,
+    # server availability, and snapshot age.  A peer that is serving us
+    # a snapshot may go offline, or network congestion may cause slow
+    # download speeds.  While downloading a snapshot, a newer snapshot
+    # may be available and we may cancel and retry the snapshot download
+    # if the full_effective_age_cancel_threshold is met.
+    max_retry_abort = 5
```
ripatel-fd, line 357: i would have just called this `max_retry`, but I think the name is fine

cali-jumptrading: yeah makes sense, feel free to change it later.

PR #7878, opened 2026-01-15
development: more sophisticated core dumps

mmcgee-jump:


src/util/shmem/fd_shmem_user.c
```diff
@@ -241,9 +242,11 @@ fd_shmem_join( char const *               name,
     FD_LOG_WARNING(( "fd_numa_mlock(\"%s\",%lu KiB) failed (%i-%s); attempting to continue",
                     path, sz>>10, errno, fd_io_strerror( errno ) ));
 
-  if( FD_UNLIKELY( madvise( shmem, sz, MADV_DONTDUMP ) ) )
```
ibhatt-jumptrading, line 244: do we need to explictly do an madvise DODUMP here? I remember having to do that but could be wrong...

mmcgee-jump: No

src/app/firedancer/topology.c
```diff
@@ -434,13 +432,13 @@ fd_topo_initialize( config_t * config ) {
   fd_topob_wksp( topo, "poh_shred"    );
   fd_topob_wksp( topo, "poh_replay"   );
 
-  fd_topob_wksp( topo, "funk"         );
+  fd_topob_wksp( topo, "funk"         )->core_dump_level = FD_TOPO_CORE_DUMP_LEVEL_FULL;
   fd_topob_wksp( topo, "progcache"    );
   fd_topob_wksp( topo, "fec_sets"     );
   fd_topob_wksp( topo, "txncache"     );
```
ibhatt-jumptrading, line 438: can we add dump_level_full for tower replay and banks

ibhatt-jumptrading, line 438: does this get correctly overriden if the user specifies a lower coredump level for funk/store

mmcgee-jump: I looked at a dump and they are not big at all, so I left them in regular

mmcgee-jump: I don't understand the question? This is the minimum core dump level it needs to be included

ibhatt-jumptrading: misread this

PR #7887, opened 2026-01-16
shred: temporary feature flagged fix 32

emwang-jump:
waiting for feature flags to get reactivated on testnet to confirm


ptaffet-jump:
Looking pretty good so far

PR #7892, opened 2026-01-17
accdb: add overlay API

riptl:
Adds an API to do short-lived local database transactions.
These are designed for 'revert protect' / 'bundle' type use cases
during block production.


ripatel-fd:
@ibhatt-jumptrading This is the API that I want to use for the bank_tile's bundle execution path. I don't have the energy to finish the implementation right now.

PR #7896, opened 2026-01-18
[config]: Update testnet address for jito block engine

esemeniuc:
Address list: https://jito-foundation.gitbook.io/mev/mev-payment-and-distribution/on-chain-addresses#testnet


src/app/fdctl/config/testnet-jito.toml
```diff
@@ -1,5 +1,3 @@
-scratch_directory = "/data/fd1"
```
esemeniuc, line 1: nowhere else uses this path

PR #7931, opened 2026-01-21
rpc: implement getMultipleAccounts

riptl:


src/discof/rpc/fd_rpc_tile.c
```diff
@@ -977,7 +975,170 @@ getMinimumBalanceForRentExemption( fd_rpc_tile_t * ctx,
...
+static fd_http_server_response_t
+getMultipleAccounts( fd_rpc_tile_t * ctx,
+                     ulong           request_id,
+                     cJSON const *   params ) {
+  int param_cnt = cJSON_GetArraySize( params );
+  if( param_cnt!=2 ) {
+    /* In theory, the second argument (options) is not required.  But if
+       it is omitted, it implies Base58-encoded account data, which we
+       deliberately don't support. */
+    return (fd_http_server_response_t){ .status = 400 };
```
ripatel-fd, line 984: not true, this one has base64 default

src/discof/rpc/fd_rpc_tile.c
```diff
@@ -977,7 +975,164 @@ getMinimumBalanceForRentExemption( fd_rpc_tile_t * ctx,
...
+  char const * encoding = cJSON_GetStringValue( cJSON_GetObjectItemCaseSensitive( config, "encoding" ) );
+  if( !encoding ) encoding = "base64";
+  _Bool use_zstd = 0; (void)use_zstd;
+  if( 0==strcmp( encoding, "base64+zstd" ) && FD_HAS_ZSTD ) {
+    use_zstd = 1;
+  } else if( 0==strcmp( encoding, "base64" ) ||
+             0==strcmp( encoding, "jsonParsed" ) ) {
+    encoding = "base64";
+  } else {
+    fd_http_server_printf( ctx->http, "{\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32602,\"message\":\"Invalid param: this server only supports 'base64' account data encoding\"},\"id\":%lu}\n", request_id );
```
jherrera-jump, line 1002: Response codes, error messages, etc. need to be conformant with Agave's RPC implementation.

I think Agave for most of these error cases switches to default params if it can't parse them.

PR #7939, opened 2026-01-21
rekey alt_bn128_little_endian feature

0x0ece:
Context: https://github.com/anza-xyz/agave/pull/10101

### DIFF
--- a/src/flamenco/features/feature_map.json
+++ b/src/flamenco/features/feature_map.json
@@ -253,6 +253,6 @@
   {"name":"deprecate_rent_exemption_threshold","pubkey":"rent6iVy6PDoViPBeJ6k5EJQrkj62h7DPyLbWGHwjrC"},
   {"name":"static_instruction_limit","pubkey":"64ixypL1HPu8WtJhNSMb9mSgfFaJvsANuRkTbHyuLfnx"},
   {"name":"vote_state_v4","pubkey":"Gx4XFcrVMt4HUvPzTpTSVkdDVgcDSjKhDN1RqRS6KDuZ"},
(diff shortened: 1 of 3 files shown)


src/flamenco/features/feature_map.json
```diff
@@ -253,6 +253,6 @@
   {"name":"deprecate_rent_exemption_threshold","pubkey":"rent6iVy6PDoViPBeJ6k5EJQrkj62h7DPyLbWGHwjrC"},
   {"name":"static_instruction_limit","pubkey":"64ixypL1HPu8WtJhNSMb9mSgfFaJvsANuRkTbHyuLfnx"},
   {"name":"vote_state_v4","pubkey":"Gx4XFcrVMt4HUvPzTpTSVkdDVgcDSjKhDN1RqRS6KDuZ"},
-  {"name":"alt_bn128_little_endian","pubkey":"bnS3pWfLrxHRJvMyLm6EaYQkP7A2Fe9DxoKv4aGA8YM"},
+  {"name":"alt_bn128_little_endian","pubkey":"bn2oPgpkzQPT3tohMaAsMVGjhDmmDa4jCaVPqCFmtxM"},
```
topointon-jump, line 256: Just for context - in case there are any fixtures with the old pubkey in test vectors, we can also use the "old" json field to support fixtures created with the old key.

0x0ece: Yes, this is not the case. Sam rekeyed for extra precaution but this code just landed and we don't have test vectors yet (we need solfuzz-agave based on 4.0)

PR #7945, opened 2026-01-21
rpc: implement getTokenAccountBalance

riptl:
For now, only supports SPL Token, not SPL Token 2022


jherrera-jump:
Closing for now as we're focusing on minimal rpc methods for monitoring

PR #7955, opened 2026-01-21
snapshots: add min download speed

cali-jumptrading:


src/app/firedancer/config/default.toml
```diff
@@ -356,6 +356,13 @@ user = ""
     # if the full_effective_age_cancel_threshold is met.
     max_retry_abort = 5
 
+    # The minimum download speed the validator will tolerate in
+    # megabytes per second (MB/s).  If the download speed of a snapshot
+    # falls below the min_download_speed threshold, the validator will
+    # cancel the download and retry downloading the snapshot from a
+    # different peer.
+    min_download_speed = 1000
```
cali-jumptrading: I will change this to 100, needed this for testing

src/discof/restore/fd_snapld_tile.c
```diff
@@ -33,6 +34,11 @@ typedef struct fd_snapld_tile {
   int   load_file;
   int   sent_meta;
 
+  ulong  bytes_in_epoch;
+  double download_speed;
+  long   start_epoch;
+  long   end_epoch;
```
jvarela-jump: Probably rename as `start_batch` and `end_batch`?
( Just to avoid confusion with an `epoch` )

src/discof/restore/fd_snapld_tile.c
```diff
@@ -269,6 +289,22 @@ after_credit( fd_snapld_tile_t *  ctx,
         if( FD_LIKELY( data_len!=0UL ) ) {
           fd_stem_publish( stem, 0UL, FD_SNAPSHOT_MSG_DATA, ctx->out_dc.chunk, data_len, 0UL, 0UL, 0UL );
           ctx->out_dc.chunk = fd_dcache_compact_next( ctx->out_dc.chunk, data_len, ctx->out_dc.chunk0, ctx->out_dc.wmark );
+          ctx->bytes_in_epoch += data_len;
+
+          /* measure download speed every 100 MiB */
+          if(ctx->bytes_in_epoch>=100<<20UL) {
+            ctx->end_epoch = fd_log_wallclock();
+            ctx->download_speed = (double)(ctx->bytes_in_epoch)*1e3 / (double)(ctx->end_epoch - ctx->start_epoch);
```
jvarela-jump: Maybe just add a comment, stating that `end` and `start` are measured in `ns`, download_speed needs to be computed in `MB/s`, therefore the `*1e3` coefficient.

Side comment: there will be a rounding error here between MiB and MB, since the branch is taken in MiB, but the calculation is MB/s.

cali-jumptrading: resolved

src/discof/restore/fd_snapld_tile.c
cali-jumptrading: done

PR #7957, opened 2026-01-21
snapshots: propagate retries to metrics

cali-jumptrading:


src/discof/restore/fd_snapct_tile.c
```diff
@@ -765,6 +765,10 @@ after_credit( fd_snapct_tile_t *  ctx,
       }
 
       ctx->full_retries++;
+      ctx->metrics.full.num_retries++;
+
+      /* reset the incremental retries when we retry a full snapshot */
```
mmcgee-jump: Not sure if we want to do this? I feel the metric is more useful as total incremental attempts? If you want one for the current full, maybe add that also with a name like "CurrentIncrementalNumRetries" ?

src/discof/restore/fd_snapct_tile.c
```diff
@@ -791,6 +795,7 @@ after_credit( fd_snapct_tile_t *  ctx,
       }
 
       ctx->incr_retries++;
+      ctx->metrics.incremental.num_retries++;
```
mmcgee-jump, line 789: Bit weird to have the duplicated counters. Can you just write the metric directly out of `ctx->incr_retries` ?

src/discof/restore/fd_snapct_tile.c
cali-jumptrading: Just left it as total incremental attempts

src/discof/restore/fd_snapct_tile.c
cali-jumptrading: yeah fixed

Issue #7278, opened 2025-11-17
Auto affinity improvements


mmcgee-jump:
- [x] Exclude core 0 from `auto` affinity by default https://github.com/firedancer-io/firedancer/pull/7964
- [x] Add an `excluded_cores` list to configuration file, which will not get scheduled on https://github.com/firedancer-io/firedancer/pull/7964
- [x] Investigate why `floating` tiles get a weird affinity assignment by default, they should be unpinned
   - upon investigation, this looks ok
- [x] If the CPU has more cores than tiles when excluding HT pairs, we do not need to schedule on HT pairs and should ignore them by default https://github.com/firedancer-io/firedancer/pull/8308
- [ ] If Firedancer is started with an affinity already (e.g. from systemd), we should constrain the available cores for "auto" to this set

PR #7964, opened 2026-01-22
affinity: add ability to exclude cores

0x0ece:
Addresses 1 and 5 in: https://github.com/firedancer-io/firedancer/issues/7278
- added `excluded_cores` config value, default to "0+"
- added ability to parse "N+" in affinity (currently undocumented except for `excluded_cores`)

### DIFF
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -696,6 +696,26 @@ dynamic_port_range = "8900-9000"
     # determined automatically as well.
     agave_affinity = "auto"
 
+    # Logical CPU cores to blocklist from being used by Firedancer, when
+    # using auto affinity.  If not using auto affinity, this option is
+    # ignored.
+    #
+    # Blocklisting cores is useful if there are specific cores that are
+    # used by other processes on the system, and you do not want
+    # Firedancer running there and competing for the core.
+    #
+    # Cores are specified as a comma separated list of single cores like
+    # "0", ranges like "0-10", or ranges with stride like "0-10/2".
+    # Cores can also be specified with an 'h' suffix to indicate both
+    # the core and its hyperthread sibling should be blocklisted.  For
+    # example, "2h" blocklists core 2 and its hyperthread sibling (if
+    # any, if there is no hyperthread sibling the program will proceed).
+    #
+    # By default, both core 0 and its hyperthread sibling are
+    # blocklisted to prevent interference with OS kernel threads that
+    # often run on these cores.
+    blocklist_cores = "0h"
+
     # The number of threads to spawn per-fork for the unified scheduler.
     # The replay stage, which is a part of the Agave subprocess, uses
     # these threads for transaction execution.  The threads stay within
--- a/src/app/firedancer/topology.c
+++ b/src/app/firedancer/topology.c
@@ -1080,8 +1080,25 @@ fd_topo_initialize( config_t * config ) {
                        "Not all cores in the affinity will be used by Firedancer. You may wish to increase the number of tiles in the system by increasing "
                        "individual tile counts in the [layout] section of the configuration file.",
                        topo->tile_cnt, affinity_tile_cnt ));
-  }
+  } else {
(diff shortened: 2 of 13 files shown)


src/app/fdctl/config/default.toml
```diff
@@ -696,6 +696,18 @@ dynamic_port_range = "8900-9000"
     # determined automatically as well.
     agave_affinity = "auto"
 
+    # Logical CPU cores to exclude when computing the affinity.
```
mmcgee-jump:
```
    # Logical CPU cores to blacklist from being used by Firedancer, when
    # using auto affinity.  If not using auto affinity, this option is
    # ignored.
    #
    # Blacklisting cores is useful if there are specific cores that are
    # used by other processes on the system, and you do not want
    # Firedancer running there and competing for the core.
    #
    # Cores are specified as a comma separated list of single cores like
    # "0", ranges like "0-10", or ranges with stride like "0-10/2".
    # Cores can also be specified with an 'h' suffix to indicate both
    # the core and its hyperthread sibling should be blacklisted.  For
    # example, "2h" blacklists core 2 and its hyperthread sibling (if
    # any, if there is no hyperthread sibling the program will proceed).
    #
    # By default, both core 0 and its hyperthread sibling are
    # blacklisted to prevent interference with OS kernel threads that
    # often run on these cores.
    blacklist_cores = "0,0h"
```

Suggest language like this. Also maybe `h` is clearer than `+` ?

src/util/tile/fd_tile_private.h
```diff
@@ -52,6 +52,13 @@ int
 fd_cpuset_setaffinity( ulong               tid,
                        fd_cpuset_t const * mask );
 
+/* fd_topob_sibling_idx returns the sibling CPU (hyperthreaded pair) of
+   the provided CPU, if there is one, otherwise return ULONG_MAX.
+   On error, logs an error and exits the process. */
+
+ulong
+fd_topob_sibling_idx( ulong cpu_idx );
```
mmcgee-jump:
```suggestion
fd_tile_private_sibling_idx( ulong cpu_idx );
```

src/app/firedancer/topology.c
```diff
@@ -1081,8 +1081,22 @@ fd_topo_initialize( config_t * config ) {
                        "Not all cores in the affinity will be used by Firedancer. You may wish to increase the number of tiles in the system by increasing "
                        "individual tile counts in the [layout] section of the configuration file.",
                        topo->tile_cnt, affinity_tile_cnt ));
-  }
+  } else {
+    ushort excluded_cores[ FD_TILE_MAX ];
+    topo->excluded_cores_cnt = fd_tile_private_cpus_parse( config->layout.excluded_cores, excluded_cores );
+    if( FD_UNLIKELY( topo->agave_affinity_cnt>FD_TILE_MAX ) ) {
```
```suggestion
    if( FD_UNLIKELY( topo->excluded_cores_cnt>FD_TILE_MAX ) ) {
```

src/app/firedancer/topology.c
```diff
@@ -1081,8 +1081,22 @@ fd_topo_initialize( config_t * config ) {
...
+      FD_LOG_ERR(( "The CPU string in the configuration file under [layout.excluded_cores] specifies more CPUs than Firedancer can use. "
+                    "You should reduce the number of CPUs in the excluded cores string." ));
+    }
 
+    for( ulong i=0UL; i<topo->excluded_cores_cnt; i++ ) {
+      if( FD_UNLIKELY( excluded_cores[ i ]==USHORT_MAX ) ) {
+        FD_LOG_ERR(( "The CPU string in the configuration file under [layout.excluded_cores] contains invalid values: `f`. "
```
mmcgee-jump: What is `f` here?

0x0ece: It should never happen in practice, but if we use the same parsing fn a user could input a string with the `f` letter for floating. The result would contain a USHORT_MAX. The error msg is intended to say "hey, you can't use `f` here". I'll add a comment.

mmcgee-jump (approved):

PR #7965, opened 2026-01-22
pack: significantly reduce mem footprint, delet small_microblocks

cavemanloverboy:
# Summary
In the current implementation of pack, it is possible to dispatch multiple transactions for execution at once by turning off `SMALL_MICROBLOCKS` or executing bundles.  However, in pracice, this is only effective for bundles for obvious reasons (atomicity, revert protection). Otherwise, when scheduling individual transactions, once we've decided that we are scheduling a transaction the best thing to do is to simply dispatch it versus holding it until we find another (one or several) to dispatch. Holding it once this decision is made is just adding latency. Additionally, with the current setup in firedancer, larger microblocks means that account locks are held for longer (until the whole microblock is done executing).

In this PR, we commit to the small microblocks design. By doing so, we can significantly decrease the memory footprint of both the firedancer and frankendancer application. This is because the maximum message sizes for the `pack_bank`, `bank_pack` and `bank_poh` messages can be decreased significantly from their current values. The benefit of this is tremendous. For 32 banks, the decrease in the memory footprint is on the order of 68 GB.

The reduction in memory footprint is the main motivation for the change. The simplification of the logic and the IBRL, although also significant, is secondary. Presently: we have for 32 banks:

1. `pack_bank` link with `USHORT_MAX` mtu, `USHORT_MAX` number of messages. This is ≈4GiB.
2. `bank_pack` link with `USHORT_MAX` mtu, `16384UL` number of messages, per link. This is ≈32GiB.
3. `bank_poh`  link with `USHORT_MAX` mtu, `16384UL` number of messages, per link. THis is ≈32GiB.

After this change, we can expect (for the same 32 bank setup):

1. `pack_bank` link with `FD_PACK_BANK_MTU` (10,928 bytes) mtu, `256UL` number of messages. This is ≈0.0026 GiB (1 shared link).
2. `bank_pack` link with `FD_PACK_REBATE_MAX_SZ` (12,880 bytes) mtu, `32UL` number of messages, per link. This is ≈0.0004 GiB per link, ≈0.012 GiB total for 32 banks. (FYI the prev implementation used max num accounts per txn, but we only do rebates for writable accounts.)
3. `bank_poh` link with `FD_BANK_POH_MTU` (10,936 bytes) mtu, `32UL` number of messages, per link. This is ≈0.0003 GiB per link, ≈0.010 GiB total for 32 banks.

**Total memory savings: ~68 GiB** (from 68 GiB down to 0.025 GiB for these three link types).

# Changes Made

### 1. Updated `MAX_TXN_PER_MICROBLOCK` to 5
With `SMALL_MICROBLOCKS=1` permanently enabled, microblocks are limited to bundle size (max 5 transactions). Updated the constant from the theoretical maximum of 30 (based on fitting into a `USHORT_MAX` message) to the effective maximum of 5 that reflects actual runtime behavior.

### 2. Optimized `pack_bank` link sizing
Replaced the overly conservative `USHORT_MAX` MTU with `FD_PACK_BANK_MTU`, calculated to fit the worst-case microblock: 5 transactions plus the bank trailer. Since this is a shared link (only one exists), kept a reasonably high depth of 256 messages. Since we do not schedule transactions on a bank until the bank idle bit is set again, that means every bank should really only have one pending message at a time so even this is overkill.

### 3. Optimized `bank_pack` link sizing
Previously used `USHORT_MAX` MTU to accommodate an overly pessimistic rebate size. Updated to `FD_PACK_REBATE_MAX_SZ`, now calculated based on the actual worst case: 5 transactions with 64 writable accounts each. (FYI the prev implementation used max num accounts per txn, but we only do rebates for writable cus...) Since these are per-bank links (32 total), reduced depth to 32 messages. The `skip_cnt` logic in pack guarantees that we do not set bank idle bits faster than we process rebates (one `poll_cursor` check for every full fd_stem round robin), so we should never even get to 2 pending rebates.

Updated the HEADROOM calculation in rebate sum logic to match the new 5×64 maximum instead of the old 31×128 assumption.

### 4. Optimized `bank_poh` link sizing
Similar optimization to `pack_bank`, but uses `FD_BANK_POH_MTU` to account for the different trailer type (`fd_microblock_trailer_t` vs `fd_microblock_bank_trailer_t`). These are also per-bank links, so reduced depth to 32 messages. This one might actually get backpressured if poh is slow, but `32UL` is reasonably high imo.

### 5. Removed all `SMALL_MICROBLOCKS` conditional code
Since `SMALL_MICROBLOCKS` is always 1, removed all `#if !SMALL_MICROBLOCKS` sections that supported the legacy large microblock mode:
- `wait_duration` delay logic that batched transactions before dispatch
- `wait_duration_ticks` array in pack tile context
- `FD_IMPORT` of `pack_delay.bin` delay configuration data


cavemanloverboy:
if you guys actually wanna move forward with this, I think we might actually need to update the bank tiles to publish in `after_credit` so we don't overrun. don't wanna spend more time on this without confirmation y'all will accept it tho

cavemanloverboy:
also, this reduces the mem footprint of the discoh bank tiles themselves, since those preallocate for the txn abi sidecars in franken.

ripatel-fd:
Appreciate this contribution and sorry that we couldn't upstream it. This is mostly fixed in full Firedancer and we are doing further memory optimizations there. Frankendancer is EOL and won't receive further optimizations.

PR #7972, opened 2026-01-22
rpc: fix getHealth

jherrera-jump:
- rpc now retains bank ownership until the bank is older than the largest finalized slot (as determined by tower)
- getHealth now works

### DIFF
--- a/src/discof/replay/fd_replay_tile.c
+++ b/src/discof/replay/fd_replay_tile.c
@@ -1297,9 +1297,6 @@ publish_reset( fd_replay_tile_t *  ctx,
   }
   reset->next_leader_slot = ctx->next_leader_slot;
 
-  if( FD_LIKELY( ctx->rpc_enabled ) ) bank->data->refcnt++;
-  FD_LOG_DEBUG(( "bank (idx=%lu, slot=%lu) refcnt incremented to %lu", bank->data->idx, reset->completed_slot, bank->data->refcnt ));
-
   fd_stem_publish( stem, ctx->replay_out->idx, REPLAY_SIG_RESET, ctx->replay_out->chunk, sizeof(fd_poh_reset_t), 0UL, 0UL, fd_frag_meta_ts_comp( fd_tickcount() ) );
   ctx->replay_out->chunk = fd_dcache_compact_next( ctx->replay_out->chunk, sizeof(fd_poh_reset_t), ctx->replay_out->chunk0, ctx->replay_out->wmark );
 }
--- a/src/discof/rpc/fd_rpc_tile.c
+++ b/src/discof/rpc/fd_rpc_tile.c
@@ -1,5 +1,6 @@
 #include "../replay/fd_replay_tile.h"
 #include "../genesis/fd_genesi_tile.h"
+#include "../tower/fd_tower_tile.h"
 #include "../../disco/topo/fd_topo.h"
 #include "../../disco/keyguard/fd_keyload.h"
 #include "../../disco/keyguard/fd_keyswitch.h"
@@ -27,7 +28,8 @@
 #define FD_HTTP_SERVER_RPC_MAX_REQUEST_LEN 8192UL
 
 #define IN_KIND_REPLAY (0)
-#define IN_KIND_GENESI (0)
+#define IN_KIND_GENESI (1)
+#define IN_KIND_TOWER  (2)
 
 #define FD_RPC_COMMITMENT_PROCESSED (0)
 #define FD_RPC_COMMITMENT_CONFIRMED (1)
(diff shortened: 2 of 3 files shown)


src/discof/replay/fd_replay_tile.c
```diff
@@ -1301,9 +1301,6 @@ publish_reset( fd_replay_tile_t *  ctx,
   }
   reset->next_leader_slot = ctx->next_leader_slot;
 
-  if( FD_LIKELY( ctx->rpc_enabled ) ) bank->data->refcnt++;
```
mmcgee-jump, line 1300: Why removed?

src/discof/rpc/fd_rpc_tile.c
```diff
@@ -291,15 +300,7 @@ returnable_frag( fd_rpc_tile_t *     ctx,
...
-        break;
-      }
-      case REPLAY_SIG_RESET: {
-        fd_poh_reset_t const * reset = fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk );
-
-        ulong prior_processed_idx = ctx->processed_idx;
-        ctx->processed_idx = reset->bank_idx;
-
-        if( FD_LIKELY( prior_processed_idx!=ULONG_MAX ) ) fd_stem_publish( stem, ctx->replay_out->idx, prior_processed_idx, 0UL, 0UL, 0UL, 0UL, 0UL );
+        ctx->processed_idx = slot_completed->bank_idx;
```
mmcgee-jump, line 316: So the "processed" slot is always the last completed one, not the reset slot?

src/discof/rpc/fd_rpc_tile.c
```diff
@@ -794,12 +842,13 @@ getHealth( fd_rpc_tile_t * ctx,
...
-  else {
-    ulong slots_behind = fd_ulong_sat_sub( ctx->cluster_confirmed_slot, ctx->banks[ ctx->confirmed_idx ].slot );
-    if( FD_LIKELY( slots_behind<=128UL ) ) fd_http_server_printf( ctx->http, "{\"jsonrpc\":\"2.0\",\"result\":\"ok\",\"id\":%lu}\n", request_id );
-    else                                   fd_http_server_printf( ctx->http, "{\"jsonrpc\":\"2.0\",\"error\":{\"code\":%d,\"message\":\"Node is unhealthy\",\"data\":{\"slotsBehind\":%lu}},\"id\":%lu}\n", FD_RPC_ERROR_NODE_UNHEALTHY, slots_behind, request_id );
+  int health_status = _getHealth( ctx );
+
+  switch( health_status ) {
+    case FD_RPC_HEALTH_STATUS_UNKNOWN: fd_http_server_printf( ctx->http, "{\"jsonrpc\":\"2.0\",\"error\":{\"code\":%d,\"message\":\"Node is unhealthy\",\"data\":{\"slotsBehind\":null}},\"id\":%lu}\n", FD_RPC_ERROR_NODE_UNHEALTHY, request_id ); break;
+    case FD_RPC_HEALTH_STATUS_BEHIND:  fd_http_server_printf( ctx->http, "{\"jsonrpc\":\"2.0\",\"result\":\"ok\",\"id\":%lu}\n", request_id ); break;
```
mmcgee-jump: These are flipped?

src/discof/rpc/fd_rpc_tile.c
```diff
@@ -1261,6 +1310,23 @@ static fd_http_server_response_t
 rpc_http_request( fd_http_server_request_t const * request ) {
   fd_rpc_tile_t * ctx = (fd_rpc_tile_t *)request->ctx;
 
+  if( FD_UNLIKELY( request->method==FD_HTTP_SERVER_METHOD_GET && !strcmp( request->path, "/health" ) ) ) {
```
mmcgee-jump, line 1326: Why up here instead of with the regular functions below?

mmcgee-jump: Oh this is /health, not RPC getHealth? But is the return code 200 always then?

src/discof/rpc/fd_rpc_tile.c
jherrera-jump: In Agave, its the bank we've just voted for (handle_votable_bank). Right now tower is not really built out to replicate this exactly, so I went with latest replayed slot, which is slightly more eager than Agave but I can't see how it would make a difference from an end-user perspective, since the `processed` commitment level already doesn't provide finalization guarantees.

I think reset_slot might be a bit too eager since it might be a future leader slot for a bank that hasn't been frozen yet.

src/discof/rpc/fd_rpc_tile.c
jherrera-jump: oops

src/discof/rpc/fd_rpc_tile.c
jherrera-jump: yup, https://solana.com/docs/rpc/http#health-check

src/discof/replay/fd_replay_tile.c
jherrera-jump: switched away from reset_slot, see response below

src/discof/rpc/fd_rpc_tile.c
jherrera-jump: added a comment

src/discof/rpc/fd_rpc_tile.c
mmcgee-jump: Haha why is this stuff so braindamaged ...

PR #7979, opened 2026-01-22
flamenco, elf: ensure dynamic string table exists

topointon-jump:


cmoyes-jump (dismissed):

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -636,6 +636,12 @@ fd_sbpf_r_bpf_64_32( fd_sbpf_loader_t *              loader,
                      ulong                           r_offset,
                      fd_sbpf_loader_config_t const * config ) {
 
+  /* Ensure the dynamic string table exists.
+     https://github.com/anza-xyz/sbpf/blob/v0.12.2/src/elf_parser/mod.rs#L532 */
```
mjain-jump: better to use this link
https://github.com/anza-xyz/sbpf/blob/v0.12.2/src/elf.rs#L1256-L1259

mjain-jump: actually, not sure if this is correct? below we check this:
```
info->shndx_dynsymtab<0
```
because agave checks if the dynamic symbol table exists. im not sure what the associated agave check for your fix is?

topointon-jump: It's actually the second block that contains this check, specifically the call to `dynamic_symbol_name` here:
```rust
                    let name = elf
                        .dynamic_symbol_name(symbol.st_name as Elf64Word)
                        .map_err(|_| ElfError::UnknownSymbol(symbol.st_name as usize))?;
```

which contains

```rust
self.dynamic_symbol_names_section_header.ok_or(ElfParserError::NoDynamicStringTable)?,
```

Have updated the permalink and comment to make this more clear

PR #7983, opened 2026-01-23
firedancer: add wait_for_supermajority_with_bank_hash

jherrera-jump:
- Adds cluster restart flags. GUI compatibility not yet added.
- Fix incomplete parsing of hard forks in snapshot manifest
- fix missing ipecho->replay link
- drive by fix compiling with FD_DISABLE_OPTIMIZATION


src/app/firedancer/config/default.toml
```diff
@@ -527,6 +527,21 @@ telemetry = true
     # deadlocked waiting on their votes to land.
     wait_for_vote_to_start_leader = true
 
+    # If set to zero or some number smaller than the snapshot slot,
```
mmcgee-jump: Do we need to keep this option the same as agave? It seems like it should really be `wait_for_supermajority_at_boot` and it's always just the snapshot slot?

mmcgee-jump: It can actually be rolled into the following one too?

`wait_for_supermajority_with_bank_hash` ?

src/app/firedancer/topology.c
```diff
@@ -580,13 +582,15 @@ fd_topo_initialize( config_t * config ) {
   FOR(gossvf_tile_cnt) fd_topob_link( topo, "gossvf_gossip", "gossvf_gossip", config->net.ingress_buffer_size,          sizeof(fd_gossip_view_t)+FD_NET_MTU, 1UL );
   /**/                 fd_topob_link( topo, "gossip_gossvf", "gossip_gossvf", 65536UL*4UL,                              sizeof(fd_gossip_ping_update_t), 1UL ); /* TODO: Unclear where this depth comes from ... fix */
   /**/                 fd_topob_link( topo, "gossip_out",    "gossip_out",    65536UL*4UL,                              sizeof(fd_gossip_update_message_t), 1UL ); /* TODO: Unclear where this depth comes from ... fix */
+  /**/                 fd_topob_link( topo, "gossip_wfs",    "gossip_wfs",    2UL,                                      0UL, 1UL );
```
mmcgee-jump: No new link for this. I'd stick it on gossip_out

src/discof/restore/utils/fd_ssmsg.h
```diff
@@ -377,8 +377,17 @@ struct fd_snapshot_manifest {
...
+     progression.  This contains the list of slots which have
+     historically undergone a hard fork.  The typical case for these is
+     a feature is deactivated and the cluster is restarted.
+
+     Sometimes, a given slot needs to be hard forked multiple times.
+     hard_forks_cnts contains the number of times a particular slot was
+     hard forked. */
   ulong hard_forks_len;
   ulong hard_forks[ 64UL ];
+  ulong hard_forks_cnts[ 64UL ];
```
mmcgee-jump, line 401: Not opposed to this, but philosophically it seems a bit redundant since the hard_forks array can just have the item repeated 2, 3, etc times?

src/discof/gossip/fd_gossip_tile.c
```diff
@@ -115,6 +143,14 @@ during_housekeeping( fd_gossip_tile_ctx_t * ctx ) {
 
     fd_keyswitch_state( ctx->keyswitch, FD_KEYSWITCH_STATE_COMPLETED );
   }
+
+  if( FD_UNLIKELY( ctx->wfs_state==FD_GOSSIP_WFS_STATE_START && ctx->last_wallclock>ctx->wfs_next_check_5secs ) ) {
+    if( FD_UNLIKELY( maybe_done_waiting_for_supermajority( ctx, ctx->last_wallclock ) ) ) {
```
mmcgee-jump: Any reason why is this level triggered and not edge?

src/discof/gossip/fd_gossip_tile.c
```diff
@@ -243,21 +287,54 @@ handle_local_duplicate_shred( fd_gossip_tile_ctx_t *            ctx,
...
+
+  fd_bank_t bank[1];
+  if( FD_UNLIKELY( !fd_banks_bank_query( bank, ctx->banks, bank_idx ) ) ) {
+    FD_LOG_CRIT(( "invariant violation: replay bank is NULL at bank index %lu", bank_idx ));
+  }
+
+  fd_vote_states_t const * vote_states = fd_bank_vote_states_locking_query( bank );
+  fd_vote_states_iter_t iter_[1];
+
+  for( fd_vote_states_iter_t * iter = fd_vote_states_iter_init( iter_, vote_states );
```
mmcgee-jump: Don't you just need epoch stakes here? I'd pull this from the epoch message out of replay, rather than using shared state like this.

src/discof/replay/fd_replay_tile.c
```diff
@@ -2447,6 +2449,9 @@ out1( fd_topo_t const *      topo,
 
   if( FD_UNLIKELY( idx==ULONG_MAX ) ) return (fd_replay_out_link_t){ .idx = ULONG_MAX, .mem = NULL, .chunk0 = 0, .wmark = 0, .chunk = 0 };
 
+  ulong mtu = topo->links[ tile->out_link_id[ idx ] ].mtu;
```
mmcgee-jump: Is this needed? No new out link?

src/discof/replay/fd_replay_tile.c
```diff
@@ -1115,7 +1130,7 @@ static inline int
 maybe_become_leader( fd_replay_tile_t *  ctx,
                      fd_stem_context_t * stem ) {
   FD_TEST( ctx->is_booted );
-  if( FD_LIKELY( ctx->next_leader_slot==ULONG_MAX || ctx->is_leader || (!ctx->has_identity_vote_rooted && ctx->wait_for_vote_to_start_leader) || ctx->replay_out->idx==ULONG_MAX ) ) return 0;
+  if( FD_LIKELY( ctx->next_leader_slot==ULONG_MAX || ctx->is_leader || (!ctx->has_identity_vote_rooted && ctx->wait_for_vote_to_start_leader && !ctx->wfs_enabled) || ctx->replay_out->idx==ULONG_MAX || !ctx->wfs_complete ) ) return 0;
```
mmcgee-jump: Don't understand this change? Should just be?


```suggestion
  if( FD_LIKELY( ctx->next_leader_slot==ULONG_MAX || ctx->is_leader || (!ctx->has_identity_vote_rooted && ctx->wait_for_vote_to_start_leader ) || ctx->replay_out->idx==ULONG_MAX || !ctx->wfs_complete ) ) return 0;
```

src/discof/replay/fd_replay_tile.c
```diff
@@ -160,6 +163,16 @@ struct fd_replay_tile {
   int has_identity_vote_rooted;
   int wait_for_vote_to_start_leader;
 
+  /* These flags are initialized from config parameters. Typically used
+     during a cluster restart/hard-fork, they incur additional
+     validation checks are run at boot to ensure the correct snapshot is
```
mmcgee-jump: Comment language a bit weird, would rephrase

src/discof/gossip/fd_gossip_tile.c
```diff
@@ -270,6 +347,13 @@ returnable_frag( fd_gossip_tile_ctx_t * ctx,
     case IN_KIND_EPOCH:         handle_epoch( ctx, fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk ) ); break;
     case IN_KIND_GOSSVF:        handle_packet( ctx, sig, fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk ), sz, stem ); break;
     case IN_KIND_TOWER:         handle_local_duplicate_shred( ctx, sig, fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk ), stem ); break;
+    case IN_KIND_REPLAY_OUT:    {
+      fd_replay_slot_completed_t const * src = fd_chunk_to_laddr( ctx->in[ in_idx ].mem, chunk );
+
+      if( FD_LIKELY( sig==REPLAY_SIG_SLOT_COMPLETED ) ) handle_wait_for_supermajority( ctx, src->bank_idx );
```
mmcgee-jump: Actually I don't understand this plumbing at all. Why is gossip listening to replay? To get epoch stakes? You can pull those from the snapshot manifest link (or genesis link, if booting cluster), which is cleaner, no replay->gossip, more of a DAG dataflow, and no bank refcnt etc.

src/discof/restore/utils/fd_ssmsg.h
jherrera-jump: kept it this way since this is how it's stored in the manifest and also the eventual consumer (the shred version calculation) uses the hashed value of the count, so we'd convert it to a single array then convert back back which seems like extra steps with little benefit.

src/discof/gossip/fd_gossip_tile.c
jherrera-jump: based on a wrong assumption, will fix

src/discof/gossip/fd_gossip_tile.c
jherrera-jump: ah good point. I think Agave is based on latest active stake but that makes no sense. Should simplify all the plumbing too

src/discof/replay/fd_replay_tile.c
jherrera-jump: Even after we've completed waiting for supermajority, want to forcibly disable `wait_for_vote_to_start_leader` during cluster restarts since it will take a some slots for votes to start rooting, so the cluster could still deadlock if all the nodes have that param enabled.

I will switch this to a validation check in the config. i.e. if wait_for_supermajority_with_bank_hash is provided, then the operator must manually set `wait_for_vote_to_start_leader` to false.

src/discof/gossip/fd_gossip_tile.c
jherrera-jump: Confirmed that agave is in fact based on latest active stake. Still changed to fetch from the manifest.

src/discof/restore/utils/fd_ssmsg.h
mmcgee-jump: Seems reasonable

src/discof/gossip/fd_gossip_tile.c
mmcgee-jump: still seems like this is doing a periodic check rather than just doing live computation as stakes change etc?

src/app/firedancer/config/default.toml
```diff
@@ -1743,6 +1749,14 @@ telemetry = true
         allow_private_address = false
 
     [development.genesis]
+        # During normal operation, the genesi tile from performs a
+        # validation check on the genesis hash which ensures it belongs
+        # to one of the three primary clusters (testnet, mainnet,
+        # devnet). During development, this option can be disabled,
+        # which allows joining a local cluster in, for example, mainnet
+        # mode with a different genesis hash.
+        validate_genesis_hash = true
```
mmcgee-jump, line 1738: Is this from agave CLI, or new to us?

jherrera-jump: New to us and very much optional. Added it when testing this feature.

mmcgee-jump: Ok move this to development section? Can't tell why an operator would want this option. Also comments in this file are double spaced.

mmcgee-jump: Oh it already is :) haha

src/disco/topo/fd_topo.h
```diff
@@ -236,6 +236,8 @@ struct fd_topo_tile {
       ulong  max_purged;
       ulong  max_failed;
 
+      fd_pubkey_t wait_for_supermajority_with_bank_hash;
```
mmcgee-jump: Not a pubkey

src/discof/gossip/fd_gossip_tile.c
```diff
@@ -265,6 +297,26 @@ returnable_frag( fd_gossip_tile_ctx_t * ctx,
     case IN_KIND_EPOCH:         handle_epoch( ctx, fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk ) ); break;
     case IN_KIND_GOSSVF:        handle_packet( ctx, sig, fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk ), sz, stem ); break;
     case IN_KIND_TOWER:         handle_local_duplicate_shred( ctx, sig, fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk ), stem ); break;
+    case IN_KIND_SNAPIN_MANIF:    {
```
mmcgee-jump: You can get many of these if the snapshot fails / reloads multiple times etc. Or you get full and then incremental. I think you need to zero, e.g. wfs_stakes_cnt before processing a new one?

src/discof/gossip/fd_gossip_tile.c
mmcgee-jump: Still not done?

jherrera-jump: so there was a reason this is level triggered. Agave considered contact infos with wallclock older than 15s as offline, so we'd need to poll the contact info map anyways to update expired contact info entries.

One thing I did get wrong though is Agave uses 1 second sleep duration in their poll loop so will change to match that.

edit: fixed to be edge triggered

src/discof/gossip/fd_gossip_tile.c
```diff
@@ -173,7 +210,11 @@ after_credit( fd_gossip_tile_ctx_t * ctx,
               int *                  charge_busy FD_PARAM_UNUSED ) {
   ctx->stem = stem;
 
-  if( FD_UNLIKELY( !ctx->my_contact_info->shred_version ) ) return;
+  if( FD_UNLIKELY( ctx->wfs_state==FD_GOSSIP_WFS_STATE_PUBLISH ) ) {
+    fd_stem_publish( ctx->stem, ctx->gossip_out->idx, FD_GOSSIP_UPDATE_TAG_WFS_DONE, ctx->gossip_out->chunk, 0UL, 0UL, 0UL, 0UL );
```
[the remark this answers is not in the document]
jherrera-jump: chunk is unused since sz is 0, but needs to be included since consumers validatate provided chunk is in range

src/app/shared/fd_config.c
```diff
@@ -423,6 +423,10 @@ fd_config_fill( fd_config_t * config,
     }
   }
 
+  if( FD_UNLIKELY( config->is_firedancer && strcmp( config->firedancer.consensus.wait_for_supermajority_with_bank_hash, "" ) && (!config->consensus.expected_shred_version || config->consensus.wait_for_vote_to_start_leader) ) ) {
+    FD_LOG_ERR(( "Config option [consensus.wait_for_supermajority_with_bank_hash] requires consensus.expected_shred_version!=0 and consensus.wait_for_vote_to_start_leader==false." ));
```
mmcgee-jump, line 427: Disagrees with the TOML comment that it will be force disabled?

src/discof/gossip/fd_gossip_tile.c
```diff
@@ -102,6 +106,41 @@ gossip_ping_tracker_change_fn( void *        _ctx,
...
+     updates, including refreshes, so any updates we missed at boot will
+     show up shortly after. */
+  if( FD_LIKELY( !ctx->my_contact_info->shred_version || ctx->wfs_state!=FD_GOSSIP_WFS_STATE_START ) ) return;
+
+  /* gossvf should filter out messages with mismatching shred version */
+  FD_TEST( ci->shred_version==ctx->my_contact_info->shred_version );
+
+  /* ignore unstaked peers */
+  ulong stake_idx = fd_stake_weight_key_sort_split( ctx->wfs_stakes, ctx->wfs_stakes_cnt, (fd_stake_weight_t){ .key = *identity } );
+  if( FD_UNLIKELY( stake_idx>=ctx->wfs_stakes_cnt || memcmp( identity->uc, ctx->wfs_stakes[ stake_idx ].key.uc, sizeof(fd_pubkey_t) ) ) ) return;
```
mmcgee-jump, line 126: Huh, why is our own stake excluded?

src/discof/gossip/fd_gossip_tile.c
```diff
@@ -173,7 +212,11 @@ after_credit( fd_gossip_tile_ctx_t * ctx,
               int *                  charge_busy FD_PARAM_UNUSED ) {
   ctx->stem = stem;
 
-  if( FD_UNLIKELY( !ctx->my_contact_info->shred_version ) ) return;
+  if( FD_UNLIKELY( ctx->wfs_state==FD_GOSSIP_WFS_STATE_PUBLISH ) ) {
+    fd_stem_publish( ctx->stem, ctx->gossip_out->idx, FD_GOSSIP_UPDATE_TAG_WFS_DONE, ctx->gossip_out->chunk, 0UL, 0UL, 0UL, 0UL );
+    ctx->wfs_state = FD_GOSSIP_WFS_STATE_DONE;
```
mmcgee-jump, line 217: *opt_poll_in = 0;

src/discof/gossip/fd_gossip_tile.c
```diff
@@ -173,7 +212,11 @@ after_credit( fd_gossip_tile_ctx_t * ctx,
               int *                  charge_busy FD_PARAM_UNUSED ) {
   ctx->stem = stem;
 
-  if( FD_UNLIKELY( !ctx->my_contact_info->shred_version ) ) return;
```
mmcgee-jump, line 176: Where did this go?

src/discof/gossip/fd_gossip_tile.c
```diff
@@ -325,6 +398,17 @@ unprivileged_init( fd_topo_t *      topo,
...
+  ulong banks_obj_id = fd_pod_query_ulong( topo->props, "banks", ULONG_MAX );
+  FD_TEST( banks_obj_id!=ULONG_MAX );
+  ulong banks_locks_obj_id = fd_pod_query_ulong( topo->props, "banks_locks", ULONG_MAX );
+  FD_TEST( banks_locks_obj_id!=ULONG_MAX );
+
+  ctx->wfs_state  = fd_int_if( memcmp( tile->gossip.wait_for_supermajority_with_bank_hash.uc, ((fd_pubkey_t){ 0 }).uc, sizeof(fd_pubkey_t) ), FD_GOSSIP_WFS_STATE_INIT, FD_GOSSIP_WFS_STATE_DONE );
+  ctx->wfs_bank_idx = ULONG_MAX;
+  memset( ctx->wfs_active, 0, sizeof(ctx->wfs_active) );
+
+  FD_TEST( fd_banks_join( ctx->banks, fd_topo_obj_laddr( topo, banks_obj_id ), fd_topo_obj_laddr( topo, banks_locks_obj_id ) ) );
```
mmcgee-jump: Is banks used now? Thought you can remove all this?

src/discof/replay/fd_replay_tile.c
```diff
@@ -2478,12 +2473,19 @@ returnable_frag( fd_replay_tile_t *  ctx,
       process_vote_txn_sent( ctx, fd_chunk_to_laddr( ctx->in[ in_idx ].mem, chunk ) );
       break;
     }
+    case IN_KIND_GOSSIP_OUT: {
+        FD_TEST( sig==FD_GOSSIP_UPDATE_TAG_WFS_DONE );
```
mmcgee-jump: nit 4 spaces?

src/discof/replay/fd_replay_tile.c
```diff
@@ -1481,6 +1497,18 @@ on_snapshot_message( fd_replay_tile_t *  ctx,
...
+
+    if( FD_UNLIKELY( ctx->wfs_enabled && memcmp( ctx->expected_bank_hash.uc, fd_bank_bank_hash_get( bank ).uc, sizeof(fd_hash_t) ) ) ) {
+      FD_BASE58_ENCODE_32_BYTES( ctx->expected_bank_hash.uc, expected_bank_hash_cstr );
+      FD_BASE58_ENCODE_32_BYTES( fd_bank_bank_hash_get( bank ).uc, actual_bank_hash_cstr );
+      FD_LOG_ERR(( "[consensus.wait_for_supermajority_with_bank_hash] expected_bank_hash=%s does not match snapshot slot"
+                   "=%lu bank_hash=%s. If you are loading a snapshot from the network, check that the slot matches the "
+                   "cluster restart slot. ", expected_bank_hash_cstr, snapshot_slot, actual_bank_hash_cstr ));
+    }
+    if( FD_UNLIKELY( ctx->wfs_enabled ) ) {
+      FD_LOG_NOTICE(( "Waiting for supermajority at snapshot slot %lu", snapshot_slot ));
```
mmcgee-jump: For most routine log line like this prefer lowercase

```suggestion
      FD_LOG_NOTICE(( "waiting for supermajority at snapshot slot %lu", snapshot_slot ));
```

src/discof/replay/fd_replay_tile.c
```diff
@@ -2478,12 +2473,19 @@ returnable_frag( fd_replay_tile_t *  ctx,
...
+        FD_LOG_NOTICE(( "Done waiting for supermajority. More than 80 percent of cluster stake has joined." ));
+        break;
+    }
     case IN_KIND_RPC:
     case IN_KIND_GUI: {
       fd_bank_t bank[1];
       FD_TEST( fd_banks_bank_query( bank, ctx->banks, sig ) );
       bank->data->refcnt--;
       FD_LOG_DEBUG(( "bank (idx=%lu, slot=%lu) refcnt decremented to %lu for %s", bank->data->idx, fd_bank_slot_get( bank ), bank->data->refcnt, ctx->in_kind[in_idx]==IN_KIND_RPC ? "rpc" : "gui" ));
+
```
mmcgee-jump: nit whitespace

src/discof/gossip/fd_gossip_tile.c
jherrera-jump: `identity` is the identity pubkey associated with the activity update.

fd_stake_weight_key_sort_split does a binary search but if the queried pubkey is not in the array it still returns an index, so we need to do an additional check to see if the query matches the result. Will clarify the comment a bit.

src/discof/gossip/fd_gossip_tile.c
```diff
@@ -325,6 +403,14 @@ unprivileged_init( fd_topo_t *      topo,
 
   FD_TEST( fd_rng_join( fd_rng_new( ctx->rng, ctx->rng_seed, ctx->rng_idx ) ) );
 
+  ulong banks_obj_id = fd_pod_query_ulong( topo->props, "banks", ULONG_MAX );
```
mmcgee-jump: Don't need this?

PR #7988, opened 2026-01-23
flamenco: fix logs in system program

anarcheuz:
Agave uses the Debug trait for eg. `Create Account: account Address { address: SysvarRecentB1ockHashes11111111111111111111, base: None } already in use`. 

Some logs are also updated to match 3.1.0.

### DIFF
--- a/src/flamenco/runtime/program/fd_system_program.c
+++ b/src/flamenco/runtime/program/fd_system_program.c
@@ -13,6 +13,15 @@
 
 #define FD_SYSTEM_PROGRAM_INSTR_FOOTPRINT (FD_TXN_MTU + sizeof(fd_system_program_instruction_t))
 
+#define FD_FMT_ADDRESS(account_b58, base, out_fmt)                                               \
+  char out_fmt[ 128UL ];                                                                         \
+  FD_BASE58_ENCODE_32_BYTES( base->key, base_b58 );                                              \
+  if( base ) {                                                                                   \
+    snprintf( out_fmt, 128UL, "Address { address: %s, base: Some(%s) }", account_b58, base_b58 ); \
+  } else {                                                                                       \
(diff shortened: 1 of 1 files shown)


src/flamenco/runtime/program/fd_system_program.c
```diff
@@ -160,10 +160,16 @@ fd_system_program_allocate( fd_exec_instr_ctx_t *   ctx,
     /* Max msg_sz: 35 - 2 + 125 = 158 */
     FD_BASE58_ENCODE_32_BYTES( account->pubkey->key, account_b58 );
     FD_BASE58_ENCODE_32_BYTES( base->key,            base_b58    );
-    fd_log_collector_printf_inefficient_max_512( ctx,
-      "Allocate: 'to' (account %s, base %s) must sign",
-      account_b58,
-      base ? base_b58 : "None" );
+    if( base ) {
```
mjain-jump: dont think its relevant to your pr but isnt this unsafe if youre accessing base->key before this check?

anarcheuz: This is fine because FD_BASE58_ENCODE_32_BYTES checks if the pointer is NULL. `base->key` is the same pointer as `base`

src/flamenco/runtime/program/fd_system_program.c
```diff
@@ -160,10 +160,16 @@ fd_system_program_allocate( fd_exec_instr_ctx_t *   ctx,
...
+      fd_log_collector_printf_inefficient_max_512( ctx,
+        "Allocate: 'to' account Address { address: %s, base: Some(%s) } must sign",
```
mjain-jump: can you make the output format for the address a macro or something? just so we dont have this branch

anarcheuz: I've changed to a Macro. Please recheck if it's acceptable.

PR #8005, opened 2026-01-26
conformance: propagate error codes in test harness

cmoyes-jump:
Previously, we erred on the side of skipping tests instead of propagating useful error codes and checking behavioral consistency with Agave.

fd_instr_harness.c:
- Modified `fd_solfuzz_pb_instr_ctx_create()` to return proper `FD_EXECUTOR_INSTR_ERR_*` error codes instead of 0/1
- Updated callers to check for non-zero (error) instead of `!result`

fd_vm_harness.c:
- `fd_solfuzz_pb_syscall_run()`: Allocate effects early so context creation errors can return effects with proper error codes instead of skipping
- `fd_solfuzz_pb_vm_interp_run()`: Same pattern


src/flamenco/runtime/tests/fd_instr_harness.c
```diff
@@ -299,7 +299,7 @@ fd_solfuzz_pb_instr_ctx_create( fd_solfuzz_runner_t *                runner,
     uint index = test_ctx->instr_accounts[j].index;
     if( index >= test_ctx->accounts_count ) {
       FD_LOG_NOTICE( ( "instruction account index out of range (%u > %u)", index, test_ctx->instr_accounts_count ) );
-      return 0;
+      return FD_EXECUTOR_INSTR_ERR_NOT_ENOUGH_ACC_KEYS;
```
mjain-jump: use `FD_EXECUTOR_INSTR_ERR_MISSING_ACC` here, not enough account keys is deprecated in agave

mjain-jump: actually nvm i see you have that error used below

src/flamenco/runtime/tests/fd_instr_harness.c
```diff
@@ -52,7 +52,7 @@ fd_solfuzz_pb_instr_ctx_create( fd_solfuzz_runner_t *                runner,
   fd_features_t * features = fd_bank_features_modify( runner->bank );
   fd_exec_test_feature_set_t const * feature_set = &test_ctx->epoch_context.features;
   if( !fd_solfuzz_pb_restore_features( features, feature_set ) ) {
-    return 0;
+    return FD_EXECUTOR_INSTR_ERR_GENERIC_ERR;
```
mjain-jump: maybe FD_LOG_ERR with `"invariant violation: unsupported feature ID"`? solfuzz should never generate features that are unsupported by either target

src/flamenco/runtime/tests/fd_instr_harness.c
```diff
@@ -129,7 +129,7 @@ fd_solfuzz_pb_instr_ctx_create( fd_solfuzz_runner_t *                runner,
 
   if( FD_UNLIKELY( test_ctx->accounts_count > MAX_TX_ACCOUNT_LOCKS ) ) {
     FD_LOG_NOTICE(( "too many accounts" ));
-    return 0;
+    return FD_EXECUTOR_INSTR_ERR_MAX_ACCS_EXCEEDED;
```
mjain-jump: could fd_log_err with `"invariant violation: max txn accounts exceeded (<test_ctx->accounts_count> > <MAX_TX_ACCOUNT_LOCKS>)"`

src/flamenco/runtime/tests/fd_instr_harness.c
```diff
@@ -252,7 +252,7 @@ fd_solfuzz_pb_instr_ctx_create( fd_solfuzz_runner_t *                runner,
 
   if( FD_UNLIKELY( test_ctx->instr_accounts_count > MAX_TX_ACCOUNT_LOCKS ) ) {
     FD_LOG_NOTICE(( "too many instruction accounts" ));
-    return 0;
+    return FD_EXECUTOR_INSTR_ERR_MAX_ACCS_EXCEEDED;
```
mjain-jump: same as above, this should be easy to fix up as well

src/flamenco/runtime/tests/fd_instr_harness.c
```diff
@@ -271,7 +271,7 @@ fd_solfuzz_pb_instr_ctx_create( fd_solfuzz_runner_t *                runner,
 
   fd_epoch_schedule_t epoch_schedule_[1];
   fd_epoch_schedule_t * epoch_schedule = fd_sysvar_cache_epoch_schedule_read( ctx->sysvar_cache, epoch_schedule_ );
-  if( FD_UNLIKELY( !epoch_schedule ) ) { return 0; }
+  if( FD_UNLIKELY( !epoch_schedule ) ) { return FD_EXECUTOR_INSTR_ERR_GENERIC_ERR; }
```
mjain-jump: could do `FD_TEST( epoch_schedule );` like we do for the other sysvars

src/flamenco/runtime/tests/fd_instr_harness.c
mjain-jump: this should be fixed up already in solfuzz, so you can do the same as above here

src/flamenco/runtime/tests/fd_instr_harness.c
```diff
@@ -323,10 +323,12 @@ fd_solfuzz_pb_instr_ctx_create( fd_solfuzz_runner_t *                runner,
...
   }
 
-  /* Early returning only happens in instruction execution. */
+  /* For non-syscalls (instruction execution), program_id must be in the input accounts.
+     For syscalls, we skip this check because the program_id was already added to
+     txn_out->accounts at lines 169-181 if it wasn't in the input. */
   if( !is_syscall && !found_program_id ) {
     FD_LOG_NOTICE(( " Unable to find program_id in accounts" ));
-    return 0;
+    return FD_EXECUTOR_INSTR_ERR_MISSING_ACC;
```
mjain-jump: also already fixed up in solfuzz, can do the same as above

cmoyes-jump:
> i think the correct approach for this PR is to make the instr context creation infallible both here and in solfuzz-agave. ideally these harness rejections dont pass by silently, and instead FD_LOG_ERR so we can make the appropriate fixup in solfuzz.

Makes sense. Updated.

PR #8025, opened 2026-01-27
runtime: harden feature activation logic

ripatel-fd:
- Only consider accounts owned by the feature program for
  activation (fixes a stray FD_LOG_WARNING for account
  9LZdXeKGeBV6hRLdxS1rHbHoEUsKqesCC2ZAPTPKJAbK, which was
  initialized a year ago, but is still owned by the system
  program)
- Do not crash with FD_LOG_CRIT if the feature account is too
  small to activate, log a warning instead
- Simplify serialization logic

### DIFF
--- a/src/flamenco/runtime/fd_runtime.c
+++ b/src/flamenco/runtime/fd_runtime.c
@@ -464,35 +464,34 @@ fd_feature_activate( fd_bank_t *               bank,
     return;
   }
 
-  FD_BASE58_ENCODE_32_BYTES( addr->uc, addr_b58 );
-  fd_feature_t feature[1];
-  int decode_err = 0;
-  if( FD_UNLIKELY( !fd_bincode_decode_static( feature, feature, fd_accdb_ref_data_const( ro ), fd_accdb_ref_data_sz( ro ), &decode_err ) ) ) {
+  if( FD_UNLIKELY( !fd_pubkey_eq( fd_accdb_ref_owner( ro ), &fd_solana_feature_program_id ) ) ) {
+    /* Feature account not yet initialized */
     fd_accdb_close_ro( accdb, ro );
-    FD_LOG_WARNING(( "Failed to decode feature account %s (%d)", addr_b58, decode_err ));
     return;
   }
-  fd_accdb_close_ro( accdb, ro );
 
-  if( feature->has_activated_at ) {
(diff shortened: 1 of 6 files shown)


src/flamenco/runtime/fd_runtime.c
```diff
@@ -464,7 +464,21 @@ fd_feature_activate( fd_bank_t *               bank,
...
+  if( FD_UNLIKELY( !fd_pubkey_eq( fd_accdb_ref_owner( ro ), &fd_solana_feature_program_id ) ) ) {
+    /* Feature account not yet initialized */
+    fd_accdb_close_ro( accdb, ro );
+    return;
+  }
+
   FD_BASE58_ENCODE_32_BYTES( addr->uc, addr_b58 );
+  ulong data_space = fd_accdb_ref_data_sz( ro );
+  if( FD_UNLIKELY( data_space<9UL ) ) {
+    /* Feature account too small to be valid */
```
ibhatt-jumptrading: can you define this as a const somewhere

src/flamenco/runtime/fd_runtime.c
```diff
@@ -464,35 +464,34 @@ fd_feature_activate( fd_bank_t *               bank,
...
-    fd_features_set( features, id, feature->activated_at);
-  } else {
-    FD_LOG_DEBUG(( "Feature %s not activated at %lu, activating", addr_b58, feature->activated_at ));
+  FD_BASE58_ENCODE_32_BYTES( addr->uc, addr_b58 );
+
+  fd_feature_t feature;
+  if( FD_UNLIKELY( !fd_feature_decode( &feature, fd_accdb_ref_data_const( ro ), fd_accdb_ref_data_sz( ro ) ) ) ) {
+    FD_LOG_WARNING(( "cannot activate feature %s, corrupt account data", addr_b58 ));
+    FD_LOG_HEXDUMP_NOTICE(( "corrupt feature account", fd_accdb_ref_data_const( ro ), fd_accdb_ref_data_sz( ro ) ));
+    fd_accdb_close_ro( accdb, ro );
```
ibhatt-jumptrading, line 479: if feature account is malformed should we crash here

ripatel-fd: IMO no because Agave doesn't crash either here

PR #8053, opened 2026-01-28
rebate writable alt cus unconditionally 

cavemanloverboy:


src/disco/pack/fd_pack.c
```diff
@@ -1964,20 +1965,23 @@ fd_pack_schedule_impl( fd_pack_t          * pack,
...
+      out_txnp->pack_cu.requested_exec_plus_acct_data_cus = cur->txn->pack_cu.requested_exec_plus_acct_data_cus;
+      out_txnp->pack_cu.non_execution_cus       = cur->txn->pack_cu.non_execution_cus;
+      out_txnp->scheduler_arrival_time_nanos    = cur->txn->scheduler_arrival_time_nanos;
+      out_txnp->source_tpu                      = cur->txn->source_tpu;
+      out_txnp->source_ipv4                     = cur->txn->source_ipv4;
+      out_txnp->flags                           = cur->txn->flags;
     }
+    /* Copy the ALT accounts from the source fd_txn_e_t */
+    ulong alt_acct_cnt = (ulong)txn->addr_table_adtl_cnt;
+    fd_memcpy( out->alt_accts, cur->txn_e->alt_accts, alt_acct_cnt * sizeof(fd_acct_addr_t) );
```
ptaffet-jump, line 1996: Can you make this non-temporal if supported?

src/disco/pack/fd_pack_rebate_sum.c
```diff
@@ -91,17 +91,18 @@ fd_pack_rebate_sum_add_txn( fd_pack_rebate_sum_t         * s,
...
-        if( FD_UNLIKELY( !in_table ) ) {
-          in_table = rmap_insert( s->map, accts[j] );
-          in_table->rebate_cus = 0UL;
-          s->inserted[ s->writer_cnt++ ] = in_table;
-        }
-        in_table->rebate_cus += rebated_cus;
+    /* ALT accounts are pre-resolved by resolv_tile and passed via
+       fd_txn_e_t, so we always rebate even if bank sanitization
+       failed (e.g. due to LUT deactivation). */
+    accts = adtl_writable[i];
```
ptaffet-jump, line 98: Can you make this conditional on accts==NULL and update the documentation for the function?

src/discof/execle/fd_execle_tile.c
```diff
@@ -130,7 +131,19 @@ during_frag( fd_execle_tile_t * ctx,
   if( FD_UNLIKELY( chunk<ctx->pack_in_chunk0 || chunk>ctx->pack_in_wmark || sz>USHORT_MAX ) )
     FD_LOG_ERR(( "chunk %lu %lu corrupt, not in range [%lu,%lu]", chunk, sz, ctx->pack_in_chunk0, ctx->pack_in_wmark ));
 
-  fd_memcpy( dst, src, sz-sizeof(fd_microblock_execle_trailer_t) );
+  /* Pack sends fd_txn_e_t (with ALT accounts), but PoH expects fd_txn_p_t.
+     We copy only the fd_txn_p_t portion to the PoH output buffer, and keep
+     a pointer to the original fd_txn_e_t for accessing ALT accounts in rebates.
+     Credits aren't returned until we call housekeeping later so this ptr should
```
ptaffet-jump: This is true, but we code all the tiles as if it weren't true. Could you copy the ALTs to somewhere in the tile context in during_frag?

src/discof/execle/fd_execle_tile.c
```diff
@@ -130,7 +131,19 @@ during_frag( fd_execle_tile_t * ctx,
...
+     remain valid */
+  ulong txn_cnt = (sz-sizeof(fd_microblock_execle_trailer_t))/sizeof(fd_txn_e_t);
+  fd_txn_e_t * src_txn_e = (fd_txn_e_t *)src;
```
ptaffet-jump: Make this `const`

ptaffet-jump (commented):
Thanks for the contribution! A few minor things. The biggest thing is the memory lifetime. If you notice, we code all the tiles to be safe if it gets overrun during `during_frag`. Even though that shouldn't happen with a reliable consumer, it could happen e.g. if the producer tile were compromised, and we may move away from reliable links in some places in the future (though probably not this place). In this case, I think the security impact would be minimal because you're not reading any offsets, but it would be better to keep it consistent.

src/disco/pack/fd_pack_rebate_sum.h
```diff
@@ -76,12 +76,12 @@ void * fd_pack_rebate_sum_new( void * mem );
...
-   [0, TXN(txn[i])->addr_table_adtl_writable_cnt ).  If txn[i] does not
-   load any accounts writably from address lookup tables or if the
-   SANITIZE_SUCCESS flag is not set, adtl_writable[i] is ignored and can
-   be NULL.  txn_cnt must be in [0, MAX_TXN_PER_MICROBLOCK], where
-   txn_cnt==0 is a no-op.  txn and adtl_writable can be NULL if
-   txn_cnt==0.
+   [0, TXN(txn[i])->addr_table_adtl_writable_cnt ).  If adtl_writable is
+   NULL or adtl_writable[i] is NULL, ALT account rebates are skipped
+   (this can happen if no transactions load accounts writably from
+   address lookup tables or if the resolved addresses are not available).
```
ptaffet-jump: Additionally, if the transaction txn[i] loads writable
   accounts from one or more address lookup tables, addtl_writable[i]
must either be NULL or point to the first writable account address that it loaded.
When not NULL, adtl_writable is indexed addtl_writable[i][j] for j in
   [0, TXN(txn[i])->addr_table_adtl_writable_cnt ). If adtl_writable[i] is NULL, ALT account rebates are skipped
   (this can happen if no transactions load accounts writably from
   address lookup tables or if the resolved addresses are not available). txn_cnt must be in [0, MAX_TXN_PER_MICROBLOCK], where txn_cnt==0 is a
   no-op.  txn and adtl_writable can be NULL if txn_cnt==0.

src/disco/pack/fd_pack_rebate_sum.c
```diff
@@ -91,16 +91,22 @@ fd_pack_rebate_sum_add_txn( fd_pack_rebate_sum_t         * s,
       }
       in_table->rebate_cus += rebated_cus;
     }
-    if( FD_LIKELY( txn->flags & FD_TXN_P_FLAGS_SANITIZE_SUCCESS ) ) {
+    /* ALT accounts are pre-resolved by resolv_tile and passed via
+       fd_txn_e_t, so we always rebate even if bank sanitization
+       failed (e.g. due to LUT deactivation).  If adtl_writable is NULL
+       or adtl_writable[i] is NULL, we do not rebate ALT accounts. */
+    if( FD_LIKELY( adtl_writable ) ) {
```
ptaffet-jump: Sorry, my last comment wasn't clear. I just meant adtl_writable[i] could be NULL. I don't see a use for `adtl_writable` being NULL, and it just complicates the explanation in the comment.

src/discof/execle/fd_execle_tile.c
```diff
@@ -130,7 +131,17 @@ during_frag( fd_execle_tile_t * ctx,
...
-  fd_memcpy( dst, src, sz-sizeof(fd_microblock_execle_trailer_t) );
+  /* Pack sends fd_txn_e_t (with ALT accounts), but PoH expects fd_txn_p_t.
+     We copy the fd_txn_p_t portion to the PoH output buffer, and copy the
+     ALT accounts to the tile context for rebates. */
+  ulong txn_cnt = (sz-sizeof(fd_microblock_execle_trailer_t))/sizeof(fd_txn_e_t);
+  fd_txn_e_t const * src_txn_e = (fd_txn_e_t const *)src;
+  fd_txn_p_t       * dst_txn_p = (fd_txn_p_t       *)dst;
+  for( ulong i=0UL; i<txn_cnt; i++ ) {
+    fd_memcpy( dst_txn_p + i, src_txn_e[i].txnp, sizeof(fd_txn_p_t) );
+    fd_memcpy( ctx->_alt_accts[i], src_txn_e[i].alt_accts, sizeof(ctx->_alt_accts[i]) );
```
ptaffet-jump: Hmm I know we copy the whole txn_p, but I think we should only copy the addresses that are used.  Make sure that you check the value though so that a compromised pack tile can't cause it to read out of bounds.

src/discoh/bank/fd_bank_tile.c
```diff
@@ -120,7 +121,19 @@ during_frag( fd_bank_ctx_t * ctx,
...
+     a pointer to the original fd_txn_e_t for accessing ALT accounts in rebates.
+     Credits aren't returned until we call housekeeping later so this ptr should
+     remain valid */
+  ulong txn_cnt = (sz-sizeof(fd_microblock_execle_trailer_t))/sizeof(fd_txn_e_t);
+  fd_txn_e_t * src_txn_e = (fd_txn_e_t *)src;
+  fd_txn_p_t * dst_txn_p = (fd_txn_p_t *)dst;
+  for( ulong i=0UL; i<txn_cnt; i++ ) {
+    fd_memcpy( dst_txn_p + i, src_txn_e[i].txnp, sizeof(fd_txn_p_t) );
+  }
+  ctx->_pack_txn_e = src_txn_e;
```
ptaffet-jump: Sorry, the changes you made to `discof/execle` to make it robust against overrun need to be done here too please

cavemanloverboy:
> Sorry to keep nit-picking. I think this is pretty close!

no worries, gimme 5 min

src/disco/pack/fd_pack_rebate_sum.h
```diff
@@ -67,21 +67,23 @@ void * fd_pack_rebate_sum_new( void * mem );
...
-   no-op.  txn and adtl_writable can be NULL if txn_cnt==0.
+   fields set.
+   
+   Additionally, if the transaction txn[i] loads writable accounts
+   from one or more address lookup tables, addtl_writable[i] must
+   either be NULL or point to the first writable account address that
+   it loaded.  When not NULL, adtl_writable is indexed
+   addtl_writable[i][j] for j in [0, TXN(txn[i])->addr_table_adtl_writable_cnt ).
+   If adtl_writable[i] is NULL, ALT account rebates are skipped (this can
+   happen if no transactions load accounts writably from address
```
ptaffet-jump: (this can happen is the transaction doesn't load any writable accounts from address lookup tables)

ptaffet-jump (dismissed):

ptaffet-jump (approved):

PR #8062, opened 2026-01-28
Use getsockopt() to get the error value of connect

hrvoje-osec:
Adds `fd_bundle_client_get_connect_result` which will follow the man page and do a `getsockopt` call to read the `SO_ERROR` option at level `SOL_SOCKET` to get the error value.

Additionally checks the error value once `poll` sets the `POLLOUT` flag as this flag is always set when there is some kind of result from the initial `connect()` call, meaning both on success and failure.


yufeng-jump:
This PR introduces a seccomp violation to the bundle tile due to the use of getsockopt()

PR #8065, opened 2026-01-28
Fix CPU topo in `fd_shmem`

### LINKED ISSUE #4428: Auto affinity fails for offline CPUs`
```
$ echo 0 | sudo tee /sys/devices/system/cpu/cpu10/online
$ build/native/gcc/bin/fddev
...
ERR     03-08 20:48:42.163545 815486 f0   pidns src/app/fdctl/run/run.c(200): Unable to set the thread affinity for tile sign:0 on cpu 10. It is likely that the affinity you have specified for this tile in [layout.affinity] of your configuration file contains a CPU (10) which does not exist on this machine.
```

HGuillemet:
`fd_shmem_admin` assumes no CPU were made offline. This may result in wrong numa assignment or errors when trying, eg, to set a tile affinity for cpu 63 on a 64-cpu machine with some of them offline.

This is particularly problematic since a warning is generated at startup that suggests to put pack and poh siblings offline.

This PR proposes a fix by using the information reported by `fd_cpu_topo.c`.

Should fix #4428, at least my comment. Not sure about @ripatel-fd original issue.


HGuillemet:
Replaced by #10430

PR #8068, opened 2026-01-28
flamenco, runtime, vm: fix accounts_resize_delta to use signed arithmetic

topointon-jump:
https://github.com/firedancer-io/auditor-internal/issues/371


two-heart (approved):

ibhatt-jumptrading:
feels weird that we just cast down ulong to long in some places now

topointon-jump:
The casts are disgusting but unfortunately necessary to support negative resize deltas.

Agave has the same casts: 
https://github.com/anza-xyz/agave/blob/v2.1.14/sdk/src/transaction_context.rs#L1103C9-L1103C82
```rust
let length_delta = (new_length as i64).saturating_sub(old_length as i64);
```

Also added a unit test for this case specifically.

PR #8073, opened 2026-01-28
flamenco, vote: fix signer checks for transactions with large number of instruction accounts

topointon-jump:
We need to de-dup the signers array otherwise we run into issues with large numbers of instruction accounts

### DIFF
--- a/src/flamenco/runtime/context/fd_exec_instr_ctx.c
+++ b/src/flamenco/runtime/context/fd_exec_instr_ctx.c
@@ -127,9 +127,8 @@ fd_exec_instr_ctx_try_borrow_last_program_account( fd_exec_instr_ctx_t const * c
 
 int
 fd_exec_instr_ctx_get_signers( fd_exec_instr_ctx_t const * ctx,
-                               fd_pubkey_t const *         signers[static FD_TXN_SIG_MAX] ) {
-  fd_memset( signers, 0, sizeof(fd_pubkey_t const *) * FD_TXN_SIG_MAX );
-
+                               fd_pubkey_t const *         signers[static FD_TXN_SIG_MAX],
+                               ulong *                     signers_cnt ) {
   ulong j = 0UL;
   for( ushort i=0; i<ctx->instr->acct_cnt; i++ ) {
     if( fd_instr_acc_is_signer_idx( ctx->instr, i, NULL ) ) {
@@ -143,18 +142,20 @@ fd_exec_instr_ctx_get_signers( fd_exec_instr_ctx_t const * ctx,
       }
 
       /* Skip if duplicate signer */
-      if( fd_signers_contains( signers, pubkey ) ) {
+      if( FD_UNLIKELY( fd_signers_contains( signers, j, pubkey ) ) ) {
         continue;
       }
 
       /* This should never be possible */
       if( FD_UNLIKELY( j>=FD_TXN_SIG_MAX ) ) {
-        FD_LOG_CRIT(( "invariant violation: too many signers" ));
+        FD_LOG_CRIT(( "invariant violation: too many signers (cnt=%lu, FD_TXN_SIG_MAX=%lu)", j, (ulong)FD_TXN_SIG_MAX ));
       }
 
       signers[j++] = pubkey;
     }
   }
+
+  *signers_cnt = j;
   return FD_EXECUTOR_INSTR_SUCCESS;
 }
 
--- a/src/flamenco/runtime/program/fd_vote_program.c
+++ b/src/flamenco/runtime/program/fd_vote_program.c
@@ -735,6 +735,7 @@ authorize( fd_exec_instr_ctx_t *         ctx,
            fd_pubkey_t const *           authorized,
            fd_vote_authorize_t           vote_authorize,
            fd_pubkey_t const *           signers[static FD_TXN_SIG_MAX],
+           ulong                         signers_cnt,
            fd_sol_sysvar_clock_t const * clock ) {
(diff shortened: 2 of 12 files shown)


src/flamenco/runtime/context/fd_exec_instr_ctx.c
```diff
@@ -129,16 +129,30 @@ int
...
+      if( fd_signers_contains( signers, pubkey ) ) {
+        continue;
+      }
+
+      /* This should never be possible */
+      if( FD_UNLIKELY( j>=FD_TXN_SIG_MAX ) ) {
+        FD_LOG_CRIT(( "invariant violation: too many signers" ));
+      }
+
+      signers[j++] = pubkey;
```
mjain-jump, line 154: dont we need to save an output signer count somewhere?

topointon-jump: at the moment the api relies on the unset pubkeys being zerod out, hence the memset. we should change the api but lets do it in a different pr

ibhatt-jumptrading: this pr is small enough where i think it would be fine to change the api

topointon-jump: changed, agree it's better this way

src/flamenco/runtime/context/fd_exec_instr_ctx.c
```diff
@@ -127,18 +127,35 @@ fd_exec_instr_ctx_try_borrow_last_program_account( fd_exec_instr_ctx_t const * c
 
 int
 fd_exec_instr_ctx_get_signers( fd_exec_instr_ctx_t const * ctx,
-                               fd_pubkey_t const *         signers[static FD_TXN_SIG_MAX] ) {
+                               fd_pubkey_t const *         signers[static FD_TXN_SIG_MAX],
```
ibhatt-jumptrading, line 130: why not just bump up the FD_TXN_SIG_MAX bound here instead of doing duplicate checking

src/flamenco/runtime/context/fd_exec_instr_ctx.c
```diff
@@ -127,18 +127,35 @@ fd_exec_instr_ctx_try_borrow_last_program_account( fd_exec_instr_ctx_t const * c
...
         return err;
       }
+
+      /* Skip if duplicate signer */
+      if( fd_signers_contains( signers, j, pubkey ) ) {
+        continue;
+      }
+
+      /* This should never be possible */
+      if( FD_UNLIKELY( j>=FD_TXN_SIG_MAX ) ) {
```
ibhatt-jumptrading, line 150: is this true? i thought you could have an instruction that references the same signer 128 times

topointon-jump: you can (which is what this pr addresses), but `j` is the number of _unique_ signer pubkeys, which can never be more than `FD_TXN_SIG_MAX`

src/flamenco/runtime/context/fd_exec_instr_ctx.c
topointon-jump: that was my original change. either approach works.

after chatting with @mjain-jump we decided to go with the de-duplication for a couple reasons:
- less change surface area and slightly less stack pressure
- more closely matches agave (they use a hashset)
- we'd need to introduce a different constant for the transaction parser limit, which needs to stay at `FD_TXN_SIG_MAX`

I am ok with either approach

src/flamenco/runtime/context/fd_exec_instr_ctx.c
```diff
@@ -127,18 +127,35 @@ fd_exec_instr_ctx_try_borrow_last_program_account( fd_exec_instr_ctx_t const * c
...
       int err = fd_runtime_get_key_of_account_at_index( ctx->txn_out,
                                                         idx_in_txn,
-                                                        &signers[j++] );
+                                                        &pubkey );
       if( FD_UNLIKELY( err ) ) {
         return err;
       }
+
+      /* Skip if duplicate signer */
+      if( fd_signers_contains( signers, j, pubkey ) ) {
```
mjain-jump: `FD_UNLIKELY`

src/flamenco/runtime/program/fd_vote_program.c
```diff
@@ -1456,7 +1470,8 @@ process_authorize_with_seed_instruction( /* invoke_context */
   if( FD_UNLIKELY( !clock ) ) return FD_EXECUTOR_INSTR_ERR_UNSUPPORTED_SYSVAR;
 
   fd_pubkey_t * expected_authority_keys[FD_TXN_SIG_MAX] = { 0 };
-  fd_pubkey_t   single_signer                        = { 0 };
+  fd_pubkey_t   single_signer                           = { 0 };
+  ulong         expected_authority_keys_cnt             = 0UL;
```
mjain-jump: nit: swap these two around, the `expected_authority_keys_cnt` should be under the `expected_authority_keys` for clarity

src/flamenco/runtime/program/fd_vote_program.c
```diff
@@ -1485,6 +1501,7 @@ process_authorize_with_seed_instruction( /* invoke_context */
       new_authority,
       authorization_type,
       (fd_pubkey_t const **)expected_authority_keys,
+      expected_authority_keys_cnt,
```
mjain-jump, line 1504: is this not always just 1? no need for an intermediate variable

topointon-jump: no, if there are no signers than expected_authority_keys_cnt is 0 (the if statement above this)

PR #8079, opened 2026-01-29
add fd_genesis_blob_size method

mubarizkyc:


ripatel-fd:
Hi @mubarizkyc, thank you for the contribution. I'm curious, what's your use case for this method.
Your change looks useful, but this pattern is a bit error-prone. It's easy for the "blob_size" and serializer to drift out of sync.

mubarizkyc:
saw https://github.com/firedancer-io/firedancer/blob/main/src/flamenco/genesis/fd_genesis_create.h#L63 so added the method ,if this looks fragile ,we can use a dry-run based approch with  encoder ,if its on prefrence

mubarizkyc:
#8097

PR #8086, opened 2026-01-29
tower: only count completely successful votes from replay

yufeng-jump:


src/app/firedancer/topology.c
```diff
@@ -887,12 +887,15 @@ fd_topo_initialize( config_t * config ) {
...
+  FOR(execrp_tile_cnt) fd_topob_tile_in (   topo, "replay",  0UL,          "metric_in", "execrp_replay", i,            FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
+
   FOR(execrp_tile_cnt) fd_topob_tile_in (   topo, "execrp",  i,            "metric_in", "replay_execrp", 0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
+  FOR(execrp_tile_cnt) fd_topob_tile_out(   topo, "execrp",  i,                         "execrp_replay", i                                                  );
+  FOR(execrp_tile_cnt) fd_topob_tile_out(   topo, "execrp",  i,                         "execrp_sig",    i                                                  );
 
   /**/                 fd_topob_tile_in (   topo, "tower",   0UL,          "metric_in", "dedup_resolv",  0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
-  /**/                 fd_topob_tile_in (   topo, "tower",   0UL,          "metric_in", "replay_execrp", 0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
+  FOR(execrp_tile_cnt) fd_topob_tile_in (   topo, "tower",   0UL,          "metric_in", "execrp_replay", i,            FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
```
yufeng-jump: This is the only real link change.  Other changes are just drive-by grouping.

src/discof/tower/fd_tower_tile.c
```diff
@@ -949,9 +949,10 @@ returnable_frag( ctx_t *             ctx,
     return 0;
   }
   case IN_KIND_EXECRP: {
-    if( FD_LIKELY( (sig>>32)==FD_EXECRP_TT_TXN_EXEC ) ) {
-      fd_execrp_txn_exec_msg_t * msg = fd_chunk_to_laddr( ctx->in[in_idx].mem, chunk );
-      count_vote_txn( ctx, TXN(msg->txn), msg->txn->payload );
+    if( FD_LIKELY( (sig>>32)==FD_EXECRP_TT_TXN_SIGVERIFY ) ) {
```
mmcgee-jump: Does this necessarily happen after execution?

mmcgee-jump (dismissed):
.

src/discof/tower/fd_tower_tile.c
yufeng-jump: No.  But I don't think it cares about execution status here?

lidatong: agave only sources replay votes that are successfully executed. if two nodes, one agave and one firedancer, are entirely disconnected from gossip, then it's possible firedancer confirms something agave does not if we don't verify successful execution

otoh, this could happen anyway between different nodes regardless of agave / firedancer, because of gossip jitter

ibhatt-jumptrading (approved):

PR #8087, opened 2026-01-29
flamenco: remove dead check

anarcheuz:
non existing accounts should have the value 0 in the newer accountsdb implementation.

### DIFF
--- a/src/flamenco/runtime/fd_executor.c
+++ b/src/flamenco/runtime/fd_executor.c
@@ -1670,9 +1670,7 @@ fd_executor_txn_check( fd_runtime_t * runtime,
         }
       }
 
-      if( starting_lamports != ULONG_MAX ) {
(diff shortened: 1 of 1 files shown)


src/flamenco/runtime/fd_executor.c
```diff
@@ -1670,9 +1670,7 @@ fd_executor_txn_check( fd_runtime_t * runtime,
         }
       }
 
-      if( starting_lamports != ULONG_MAX ) {
```
mjain-jump, line 1673: good start, i think the checks for starting_dlen being ulong max can also be cleaned up while youre at it

mjain-jump (approved):
looks good, can do a follow-up to clean the starting_dlen stuff

mjain-jump:
should be fully addressed by https://github.com/firedancer-io/firedancer/pull/8107

ripatel-fd (approved):

PR #8088, opened 2026-01-29
runtime: fix account meta null check

two-heart:
we should took the reference, so this is not the NULL check we want


src/flamenco/runtime/info/fd_instr_info.c
```diff
@@ -67,7 +67,7 @@ fd_instr_info_sum_account_lamports( fd_instr_info_t const * instr,
     ushort idx_in_txn = instr->accounts[i].index_in_transaction;
     fd_accdb_rw_t const * ref = &txn_out->accounts.account[ idx_in_txn ];
 
-    if( !ref || instr->is_duplicate[i] ) {
+    if( !ref->meta || instr->is_duplicate[i] ) {
```
ripatel-fd: This does not look correct either way. All references in [0,instr->acct_cnt) should have a valid meta. For non-existent accounts, it points to zeros but it's still a valid pointer.

two-heart: At this point the acc should always exist, so removing this check

PR #8116, opened 2026-02-02
accdb: write path IBRL

ripatel-fd:
Greatly improves io_ur write throughput and latency. Previously,
io_ur's write path had the same performance profile as synchronous
blocking pwrite(). This was due to too frequent write ops (small
jobs), and too frequent blocking waiting for writes to complete.

This commit adds a write-back cache for append-only writes backed
by a ring buffer. This allows the backend to asynchronously queue
writes to disk and continue serving other requests without waiting
for completions.

Technical description of changes:

- read and write paths refactored into separate C files for
  maintainability
- wb_ring is the ring buffer allocation policy for the write-back
  cache
- wq_ring is a ring reorder buffer for write completions (in case
  writes complete out-of-order)
- io_commit now yields before completing writes
- io_read now reads from the write-back cache for uncompleted
  writes
- reads are now allowed while write jobs are still inflight
- a coding scheme for io_uring user_data was added to support the
  above

### DIFF
--- a/src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
+++ b/src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
@@ -1,5 +1,7 @@
 #include "fd_vinyl_io_ur_private.h"
 
+#pragma GCC diagnostic ignored "-Wunused-function"
+
 /* This backend uses scratch pad memory as a write-back cache.  Every
    byte in the scratch pad has one of the following states:
 
@@ -38,19 +40,26 @@ fd_vinyl_io_wq_completion( fd_vinyl_io_ur_t * ur ) {
   FD_CRIT( ur->cqe_pending      >0, "stray completion"       );
   FD_CRIT( ur->cqe_write_pending>0, "stray write completion" );
 
+  /* interpret CQE */
   struct io_uring_cqe * cqe = fd_io_uring_cq_head( ring->cq );
   if( FD_UNLIKELY( !cqe ) ) FD_LOG_CRIT(( "no write completion found" ));
   if( ur_udata_req_type( cqe->user_data )!=UR_REQ_WRITE ) {
     FD_LOG_CRIT(( "unexpected CQE type while flushing write queue" ));
   }
-
   int cqe_res = cqe->res;
   if( cqe_res<0 ) {
     FD_LOG_ERR(( "io_uring write failed (%i-%s)", -cqe_res, fd_io_strerror( -cqe_res ) ));
   }
-  /* FIXME detect short writes */
 
-  ulong wq_idx = ur_udata_idx( cqe->user_data );
+  /* advance write cursor */
+  ulong       wq_idx  = ur_udata_idx( cqe->user_data );
+  wq_desc_t * wq_desc = wq_ring_desc( &ur->wq, wq_idx );
+  ulong       req_sz  = wq_desc->sz;
+  if( FD_UNLIKELY( (uint)cqe_res != req_sz ) ) {
+    FD_LOG_ERR(( "database write failed (short write): requested to write %lu bytes, but only wrote %i bytes (at bstream seq 0x%016lx)",
+                 req_sz, cqe_res,
+                 wq_desc->seq - req_sz ));
+  }
   ur->seq_clean = wq_ring_complete( &ur->wq, wq_idx );
 
   fd_io_uring_cq_advance( ring->cq, 1U );
@@ -93,9 +102,10 @@ wq_enqueue( fd_vinyl_io_ur_t * ur,
 
   fd_io_uring_t * ring     = ur->ring;
   ulong           cq_depth = ring->cq->depth;
+  wq_ring_t *     wq       = &ur->wq;
   for(;;) {
     wq_clean( ur );
-    if( wq_ring_free_cnt( &ur->wq )>=2 && ur->cqe_pending+2 <= cq_depth ) break;
+    if( wq_ring_free_cnt( wq )>=2 && ur->cqe_pending+2 <= cq_depth ) break;
     wq_wait( ur );
   }
 
@@ -107,7 +117,8 @@ wq_enqueue( fd_vinyl_io_ur_t * ur,
   ulong dev_off  = seq % dev_sz;
 
   ulong wsz = fd_ulong_min( sz, dev_sz - dev_off );
-  ulong wq0 = wq_ring_enqueue( &ur->wq, seq+wsz );
+  ulong wq0 = wq_ring_enqueue( wq, seq+wsz );
+  wq_ring_desc( wq, wq0 )->sz = (uint)wsz; /* remember request size for CQE */
   struct io_uring_sqe * sqe = fd_io_uring_get_sqe( ring->sq );
   if( FD_UNLIKELY( !sqe ) ) FD_LOG_CRIT(( "unbalanced SQ" ));
   *sqe = (struct io_uring_sqe) {
@@ -126,7 +137,8 @@ wq_enqueue( fd_vinyl_io_ur_t * ur,
 
   sz -= wsz;
   if( sz ) {
-    ulong wq1 = wq_ring_enqueue( &ur->wq, seq+wsz+sz );
+    ulong wq1 = wq_ring_enqueue( wq, seq+wsz+sz );
+    wq_ring_desc( wq, wq1 )->sz = (uint)sz;
     sqe = fd_io_uring_get_sqe( ring->sq );
     if( FD_UNLIKELY( !sqe ) ) FD_LOG_CRIT(( "unbalanced SQ" ));
     *sqe = (struct io_uring_sqe) {
--- a/src/vinyl/io/ur/wq_ring.h
+++ b/src/vinyl/io/ur/wq_ring.h
@@ -8,7 +8,8 @@
 
 struct wq_desc {
   ulong seq;
-  ulong done : 1;
+  uint  sz;
+  uint  done : 1;
 };
 
 typedef struct wq_desc wq_desc_t;
@@ -51,6 +52,13 @@ wq_ring_is_full( wq_ring_t const * ring ) {
   return (ring->wq1 - ring->wq0) >= ring->max;
 }
 
+FD_FN_PURE static inline wq_desc_t *
+wq_ring_desc( wq_ring_t * ring,
+              ulong       wq ) {
+  ulong mask = ring->max-1UL;
+  return &ring->ring[ wq&mask ];
+}
+
 static inline ulong
 wq_ring_enqueue( wq_ring_t * ring,
                  ulong       seq ) {
@@ -59,6 +67,7 @@ wq_ring_enqueue( wq_ring_t * ring,
   ulong mask = ring->max-1UL;
   ring->ring[ wq&mask ] = (wq_desc_t){
     .seq  = seq,
+    .sz   = 0U,
     .done = 0
   };
   ring->wq1 = wq+1UL;


src/vinyl/io/ur/fd_vinyl_io_ur_private.h
```diff
@@ -0,0 +1,242 @@
...
+/* fd_vinyl_io_ur_rd_t extends fd_vinyl_io_rd_t.  Describes an inflight
+   read request.  Each object gets created with a fd_vinyl_io_read()
+   call, has at least the lifetime of a io_uring SQE/CQE transaction,
+   and gets destroyed with fd_vinyl_io_poll().
+
+   Each fd_vinyl_io_rd_t describes a contiguous read in bstream seq
+   space.  When mapped to the device, this typically results in a single
+   contiguous read. */
+
+struct fd_vinyl_io_ur_rd;
```
cali-jumptrading, line 49: why the forward declared struct before the struct definition?

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
```diff
@@ -0,0 +1,242 @@
...
+
+static inline void
+bd_read( int    fd,
+         ulong  off,
+         void * buf,
+         ulong  sz ) {
+  ssize_t ssz = pread( fd, buf, sz, (off_t)off );
+  if( FD_LIKELY( ssz==(ssize_t)sz ) ) return;
+  if( ssz<(ssize_t)0 ) FD_LOG_CRIT(( "pread(fd %i,off %lu,sz %lu) failed (%i-%s)", fd, off, sz, errno, fd_io_strerror( errno ) ));
+  /**/                 FD_LOG_CRIT(( "pread(fd %i,off %lu,sz %lu) failed (unexpected sz %li)", fd, off, sz, (long)ssz ));
```
cali-jumptrading, line 132: This FD_LOG_CRIT is positioned a bit confusingly to me... I feel like it should not be aligned with the previous FD_LOG_CRIT because that one is in an if expression.

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
```diff
@@ -0,0 +1,242 @@
...
+   3 bits are always zero for 8 byte aligned pointers). */
+
+#define UR_REQ_READ      0
+#define UR_REQ_READ_PART 1
+#define UR_REQ_WRITE     2
+
+static inline ulong
+ur_udata_pack_idx( ulong req_type, /* UR_REQ_* */
+                   ulong idx ) {
+  return (idx<<3) | (req_type & 0x7UL);
```
cali-jumptrading: maybe make the 7UL a defined constant?

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
```diff
@@ -0,0 +1,369 @@
+/* fd_vinyl_io_ur_rd.c provides io_ur read methods. */
+
+#include "fd_vinyl_io_ur_private.h"
+
+/* io_ur reads are either served by the write-back cache or by disk,
+   see the explanation in fd_vinyl_io_ur_wb.c.
+
+   wb_read serves the tail part of a read from the write-back cache.
+   [seq0,seq0+sz) is the requested bstream range (assumed to be in
+   [seq_past,seq_present)).  The tail part is written to dst.  Returns
```
cali-jumptrading: I feel like "tail part" of a read is a bit confusing. Maybe we can reword this to say
```
wb_read serves the remaining portion of a read that was not fully served by reading from disk...
```

Or maybe some extra context like explaining how reads hit the disk first, and then the write back cache if the write has not made it to disk yet.

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
```diff
@@ -0,0 +1,369 @@
...
+         uchar *            dst0,
+         ulong              seq0,
+         ulong              sz ) {
+  if( FD_UNLIKELY( !sz ) ) return 0UL;
+
+  ulong seq_cache = io->seq_cache;
+  ulong seq1      = seq0 + sz;
+
+  /* expect most reads to hit disk */
+  FD_CRIT( seq1 <= io->base->seq_present, "out-of-bounds read request" );
```
cali-jumptrading: if seq1 <= io->base->seq_present, it means the sequence range is on disk right?

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
```diff
@@ -0,0 +1,369 @@
...
+  if( FD_LIKELY( fd_vinyl_seq_le( seq1, seq_cache ) ) ) return sz;
```
cali-jumptrading, line 41: wait so I'm getting confused now... does `seq_cache` point to the start of the sequence range in the write back cache? if `seq1` is less than `seq_cache`, than doesn't that mean seq1 exists on disk? I'm confused why you return sz when `seq1` < `seq_cache`.

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
```diff
@@ -0,0 +1,369 @@
...
+
+  /* request seq range served by cache */
+  ulong req0 = fd_ulong_max( seq0, seq_cache );
+  ulong req1 = seq1;
+  FD_CRIT( fd_vinyl_seq_ge( req0, wb_ring_seq0( &io->wb ) ), "write back cache read failed" );
```
cali-jumptrading: maybe a more descriptive error message, like `req start is out of bounds of wb ring`

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
```diff
@@ -0,0 +1,369 @@
...
+  FD_CRIT( fd_vinyl_seq_le( req1, wb_ring_seq1( &io->wb ) ), "write back cache read failed" );
+
+  /* output range */
+  uchar * dst = dst0 + (req0-seq0);
+  ulong   rsz = req1 - req0;
+  FD_CRIT( dst>=dst0 && dst<dst0+sz && rsz<=sz, "invariant violation" );
+
+  /* find source range, copy into output */
+  uchar const * spad     = fd_vinyl_io_ur_spad( io );
```
cali-jumptrading: the write back ring lives in a spad-like memory? I may be totally not understanding this, but is `fd_vinyl_io_ur_spad` getting the wb_ring memory? If it's a getter, I would prefer naming it so that it is clear

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
```diff
@@ -0,0 +1,369 @@
...
+  ulong const   src_off0 = wb_ring_seq_to_off( &io->wb, req0 );
+  FD_CRIT( src_off0+rsz<=io->wb.max, "invariant violation" );
+  if( FD_LIKELY( fd_vinyl_seq_ge( req0, io->wb.seq1 ) ) ) {
+    /* entire read is served by part1 of wb cache */
+    fd_memcpy( dst, spad+src_off0, rsz );
+  } else {
+    /* slow path */
```
cali-jumptrading: Also I don't understand the slow path vs fast path 😅

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
ripatel-fd: This struct is self-referential, so combined with the typedef it's required.

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
ripatel-fd: Good point, will fix

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
ripatel-fd: Well it's just `(1<<3)-1` (3-bit mask), so I think the constant should be the 3. 
Point taken though, I will introduce `REQ_TYPE_BIT_CNT`

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
ripatel-fd: Good suggestion. It's actually more aggressive than that- It checks the write-back cache first, and serves from there even if the content is fully available on disk. (Because a memcpy is faster than reading from disk). I will clarify the comment.

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
ripatel-fd: No, `seq_present` means "all the data up to here has been committed by the user". But "commit" in vinyl terms just means that it's readable (regardless if it's in dirty cache / not on disk).

I will add some more code comments.

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
ripatel-fd: From docs:

```
         [clean] [write] [wait] [future] [used] [free]
         ^       ^       ^      ^        ^      ^
      seq_cache  |   seq_write  |    seq_future |
             seq_clean     seq_present       wb.seq1

   The region [seq_past,seq_cache) is outside of the write cache.

   The region [seq_cache,seq_present), i.e. spanning 'clean', 'write',
   and 'wait', is part of the write cache.  Read requests that fall in
   this range are served from the write cache.

   The logical address space maps to the physical address space as a
   ring.  The ring mapping is not modulo, see wb_ring for details. 
```

> if seq1 is less than seq_cache, than doesn't that mean seq1 exists on disk

Correct, if the upper bound of the request is below the start of the cached region, the cache does not contain the requested range at all. The return value of this function is "read this many bytes from disk". E.g. if you call `wb_read(sz=100)` and it returns 0, it means that all 100 bytes were read from cache.

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
ripatel-fd: Yeah sorry I will give it a better name.

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
ripatel-fd: This algorithm should live in wb_ring.h. Sorry for the mess.
`wb_ring.h` is a compactly stored ring of messages that tracks two contiguous partitions (part0 and part1).
In logical space (bstream sequence numbers), part0 and part1 are contiguous (part1 comes after part0).
In physical space (the actual ring buffer byte array), part1 is at the left side, and immediately afterwards is part0.
The two ranges meet at the wraparound point.

In typical case, the request is made against part1 (the left part). In the "slow path", the request could span both part0 and part1, or only be in part0.

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
ripatel-fd: I made them less descriptive (`invariant violation`) because this code is unreachable assuming even if the io_ur user misbehaves, as long as io_ur state is valid

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
ripatel-fd: @cali-jumptrading I moved this algorithm to wb_ring.h. Is the call site clearer now with the `wb_ring_translate` method?

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
cali-jumptrading: ohh got it thanks

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
cali-jumptrading: ohh I see, it returns the bytes you still need to read, not the number of bytes actually read. makes sense

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
cali-jumptrading: oh yeah it looks way better! And I get why you have two partitions now, you want the part immediately after the allocated space in the buffer and the part that is immediately before the start of allocated space in the buffer.

cali-jumptrading: It was a bit unclear from the docs what the partitions were... but up to you if you want to explain them, since its an implementation detail of the ring buffer.

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
```diff
@@ -0,0 +1,369 @@
...
+  FD_CRIT( fd_vinyl_seq_ge( req0, wb_ring_seq0( &io->wb ) ), "invariant violation" );
+  FD_CRIT( fd_vinyl_seq_le( req1, wb_ring_seq1( &io->wb ) ), "invariant violation" );
+
+  /* output range */
+  uchar * dst = dst0 + (req0-seq0);
+  ulong   rsz = req1 - req0;
+  FD_CRIT( dst>=dst0 && dst<dst0+sz && rsz<=sz, "invariant violation" );
+
+  /* find source range, copy into output */
+  uchar const *  spad = fd_vinyl_io_ur_wb_buf( io );
```
cali-jumptrading, line 55: ok so this is a nitpick, but I feel like spad is a confusing name. perhaps `wb_buf`? Anyways not a big deal.

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
```diff
@@ -0,0 +1,369 @@
...
+     start. */
+
+  int   dev_fd   = ur->dev_fd;
+  ulong dev_base = ur->dev_base;
+  ulong dev_sz   = ur->dev_sz;
+
+  ulong dev_off = seq0 % dev_sz;
+
+  ulong rsz = fd_ulong_min( sz, dev_sz - dev_off );
+  bd_read( dev_fd, dev_base + dev_off, dst, rsz );
```
cali-jumptrading, line 115: I'm a little confused how dev_base is guaranteed to contain the bytes we want to read... probably just don't have the full bstream context here. I would prefer an api here that gets called to do the read and potential read wraparound... but up to you.

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
```diff
@@ -0,0 +1,249 @@
...
+/* fd_vinyl_io_ur_t extends fd_vinyl_io_t. */
+
+struct fd_vinyl_io_ur {
+  fd_vinyl_io_t            base[1];
+  int                      dev_fd;       /* File descriptor of block device */
+  ulong                    dev_sync;     /* Offset to block that holds bstream sync (BLOCK_SZ multiple) */
+  ulong                    dev_base;     /* Offset to first block (BLOCK_SZ multiple) */
+  ulong                    dev_sz;       /* Block store byte size (BLOCK_SZ multiple) */
+  fd_vinyl_io_ur_rd_t *    rd_head;      /* Pointer to queue head */
+  fd_vinyl_io_ur_rd_t **   rd_tail_next; /* Pointer to queue &tail->next or &rd_head if empty. */
```
cali-jumptrading: There must be a reason why you decided to use a pointer to pointer for the tail... but I'm kinda confused why.

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
```diff
@@ -0,0 +1,369 @@
...
+      if( !cq_cnt ) {
+        if( FD_UNLIKELY( blocking ) ) FD_LOG_CRIT(( "io_uring_submit_and_wait() returned but no CQEs ready" ));
+        return FD_VINYL_ERR_AGAIN;
+      }
+    }
+
+    struct io_uring_cqe * cqe = fd_io_uring_cq_head( ring->cq );
+    if( FD_UNLIKELY( !cqe ) ) FD_LOG_CRIT(( "fd_io_uring_cq_head() returned NULL despite io_uring_cq_ready()>=1" ));
+    if( ur_udata_req_type( cqe->user_data )==UR_REQ_WRITE ) {
+      fd_vinyl_io_wq_completion( ur );
```
cali-jumptrading, line 447: is it ok to throw away wq completions?

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
```diff
@@ -0,0 +1,529 @@
...
+    if( wq_ring_free_cnt( &ur->wq )>=2 && ur->cqe_pending+2 <= cq_depth ) break;
+    wq_wait( ur );
+  }
+
+  /* Map seq into the bstream store.  If we hit the store and with more
+     to go, wrap around and finish the write at the store start. */
+
+  ulong dev_base = ur->dev_base;
+  ulong dev_sz   = ur->dev_sz;
+  ulong dev_off  = seq % dev_sz;
```
cali-jumptrading, line 117: you do this pretty often, wondering if it makes sense to make a helper function for this.

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
```diff
@@ -0,0 +1,529 @@
...
+      .off       = dev_base,
+      .addr      = (ulong)( src + wsz ),
+      .len       = (uint)sz,
+      .flags     = IOSQE_FIXED_FILE,
+      .user_data = ur_udata_pack_idx( UR_REQ_WRITE, wq1 ),
+    };
+    ur->sqe_prep_cnt++;
+    ur->sqe_write_tot_sz += sz;
+    ur->cqe_pending++;
+    ur->cqe_write_pending++;
```
cali-jumptrading, line 156: same here, seems like enqueueing sqes for reads and writes could be done with a helper function to reduce the code duplication. up to you though

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
```diff
@@ -0,0 +1,529 @@
...
+  fd_vinyl_io_ur_t * ur = (fd_vinyl_io_ur_t *)io; /* Note: io must be non-NULL to have even been called */
+
+  int flag_blocking = !!( flags & FD_VINYL_IO_FLAG_BLOCKING );
+
+  int bad_align = !fd_ulong_is_aligned( sz, FD_VINYL_BSTREAM_BLOCK_SZ );
+  int bad_sz    = sz > ur->wb.max;
+
+  if( FD_UNLIKELY( bad_align | bad_sz ) ) FD_LOG_CRIT(( bad_align ? "misaligned sz" : "sz too large" ));
+
+  /* An alloc op discards a previous alloc */
```
cali-jumptrading, line 226: why does an alloc op discard a previous one? Wouldn't that throw away data?

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
```diff
@@ -0,0 +1,249 @@
...
+#include <unistd.h> /* pread, pwrite */
+
+/* WQ_DEPTH sets the max number of write queue jobs that io_ur can track
+   at once.
+
+   WQ_BLOCK_SZ sets the write block size.
+
+   WQ_DEPTH*WQ_BLOCK_SZ is thus the write window size (bandwidth-delay
+   product).  For example, with a 32 MiB window size, and 1ms write
+   latency, the max write rate is 32 GiB/s. */
```
jvarela-jump:
```suggestion
   latency, the max write rate is ~ 32 GiB/s. */
```

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
```diff
@@ -0,0 +1,529 @@
...
+  ulong seq_clean_new = wb_ring_alloc_seq0( &ur->wb, sz );
+  if( FD_UNLIKELY( fd_vinyl_seq_gt( seq_clean_new, ur->seq_clean ) ) ) {
+    /* This alloc cannot proceed until some writes happen */
+    if( !flag_blocking ) return NULL;
+    wb_flush_seq( ur, seq_clean_new );
+  }
+
+  /* At this point, we have enough clean space to alloc sz bytes. */
+
+  ulong seq = ur->base->seq_future;
```
cali-jumptrading, line 243: The docs above say the write cache spans from seq_clean to seq_present... so why are we comparing `seq_future` to `wb_ring_seq1`?

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
```diff
@@ -0,0 +1,529 @@
...
+  if( FD_UNLIKELY( fd_vinyl_seq_ne( seq, wb_ring_seq1( &ur->wb ) ) ) ) {
+    FD_LOG_CRIT(( "seq_future (%lu) and write-back buffer (%lu) out of sync", seq, wb_ring_seq1( &ur->wb ) ));
+  }
+  wb_ring_alloc( &ur->wb, sz );
+  if( FD_UNLIKELY( fd_vinyl_seq_ne( seq+sz, wb_ring_seq1( &ur->wb ) ) ) ) {
+    FD_LOG_CRIT(( "seq_future (%lu) and write-back buffer (%lu) out of sync", seq, wb_ring_seq1( &ur->wb ) ));
+  }
+  ur->base->spad_used = sz;
+  if( fd_vinyl_seq_gt( ur->wb.seq0, ur->seq_cache ) ) ur->seq_cache = ur->wb.seq0;
```
cali-jumptrading, line 252: how does `ur->wb.seq0` get out of sync with `ur->seq_cache`? I thought `seq_cache` is the start of the write back ring cache.

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
```diff
@@ -0,0 +1,529 @@
...
+/* fd_vinyl_io_ur_commit 'commits' previous appends, making them visible
+   to subsequent read calls. */
+
+int
+fd_vinyl_io_ur_commit( fd_vinyl_io_t * io,
+                       int             flags ) {
+  fd_vinyl_io_ur_t * ur = (fd_vinyl_io_ur_t *)io; /* Note: io must be non-NULL to have even been called */
+  (void)flags;
+
+  wb_ring_trim( &ur->wb, ur->base->seq_future );
```
cali-jumptrading, line 333: why do you do a trim first?

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
```diff
@@ -0,0 +1,529 @@
...
+  ur->base->seq_present = ur->base->seq_future;
+  ur->base->spad_used   = 0UL;
+
+  /* Keep io_uring pipeline fed */
+
+  if( ur->base->seq_present - ur->seq_write >= WQ_BLOCK_SZ ) {
```
cali-jumptrading, line 339: would you ever have to do more than one `wq_enqueue_seq`? what if there is more than one `WQ_BLOCK_SZ` of writes that need to happen?

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
```diff
@@ -0,0 +1,529 @@
...
+  ur->seq_write = seq1;
+
+  fd_io_uring_t * ring = ur->ring;
+  int submit_cnt = fd_io_uring_submit( ring->sq, ring->ioring_fd, 0, 0U );
+  if( FD_UNLIKELY( submit_cnt<0 ) ) FD_LOG_ERR(( "io_uring_submit failed (%i-%s)", -submit_cnt, fd_io_strerror( -submit_cnt ) ));
+  ur->sqe_sent_cnt += (ulong)submit_cnt;
+}
+
+/* wb_flush_seq does a blocking flush of the write cache until the
+   given sequence number. */
```
cali-jumptrading, line 184: in general I would appreciate more context in the comments for these functions. But I get that some of these functions are implementation-specific and not exposed to users so the comments might unnecessary. Also just don't have the context for this so getting confused easily haha

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
```diff
@@ -0,0 +1,529 @@
...
+    if( fd_vinyl_seq_gt( seq_until, ur->seq_write ) ) {
+      wq_enqueue_seq( ur, seq_until );
+    }
+  }
+
+  return FD_VINYL_SUCCESS;
+}
+
+/* fd_vinyl_io_ur_copy does a blocking read of an old block, then copies
+   it out to the tail.  This is not ideal.  Instead, we should enqueue
```
cali-jumptrading, line 350: I don't understand this. What does a read of an `old block` mean and what is the `tail` we are copying to?

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
```diff
@@ -0,0 +1,249 @@
...
+                      ulong           sz,
+                      int             flags );
+
+ulong
+fd_vinyl_io_ur_append( fd_vinyl_io_t * io,
+                       void const *    _src,
+                       ulong           sz );
+
+ulong
+fd_vinyl_io_ur_copy( fd_vinyl_io_t * io,
```
cali-jumptrading, line 182: would be nice to have more documentation for these apis.

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
```diff
@@ -0,0 +1,529 @@
...
+                             "misaligned sz" ));
+
+  /* Map the dst to the bstream (updating seq_future) and map the src
+     and dst regions onto the device.  Then copy as much as we can at a
+     time, handling device wrap around and copy buffering space. */
+
+  ulong seq = seq_future;
+
+  for(;;) {
+    ulong csz = fd_ulong_min( sz, spad_max );
```
cali-jumptrading, line 386: why can we only read up to spad_max at a time?

cali-jumptrading (commented):
have more questions, appreciate your answers so far!

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
ripatel-fd: This is copy pasted from Kevin's io_bd backend. But I think you found a bug, it seems like there's missing bounds checking for the reads. I need to check that the read is within range `[seq_past,seq_present)`.

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
ripatel-fd: This is copy-pasted from Kevin's io_bd backend. I think this is some trick/optimization to deal with sentinels / end of the list.

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
ripatel-fd: It's not thrown away, it's delivered via that call

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
ripatel-fd: Yes, good point ... Should probably abstract the bstream translation

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
ripatel-fd: i'd prefer to keep those counters here to make it easy for users to grep where they are incremented/decremented. But we can probably remove a few of these

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
cali-jumptrading: ohh got it I thought it was the same as the rq_completion thing that returns a pointer to an entry. makes sense!

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
ripatel-fd: @cali-jumptrading This is documented in `fd_vinyl_io.h`. There is only one outstanding alloc at a time, and you can finish the alloc by calling `append`. So, if you want to write 4 records, you'd do
- alloc
- append
- alloc 
- append
- alloc
- append
- alloc
- append
- commit

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
cali-jumptrading: ok so I realized the documentation is in the base header for vinyl and that this is an implementation header. I will read the base header's documentation now.

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
ripatel-fd: `seq0` gets extended backwards in some weird trim edge cases, to avoid gaps in the middle of the ring buffer. I definitely need to document this.

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
ripatel-fd:
```
     commit: finish all outstanding appends, moving all blocks in the
     bstream's present to the bstream's past.  This will empty the io's
     append scratch pad.  The underlying implementation is free to
     process outstanding appends in any order (and free to interleave
     them arbitrarily with outstanding reads).
```

Because of the documentation "this will empty the io's append scratch pad", to conform with the interface.
Allocs happen in this scratch pad (conceptually, there's a complex ring buffer instead of a scratch pad in io_ur)

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
ripatel-fd: @cali-jumptrading `wq_enqueue_seq` can do arbitrarily big appends. `WQ_BLOCK_SZ` is just the lower bound. Should be more like `WQ_BLOCK_SZ_TARGET`

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
ripatel-fd: It takes an arbitrary range in the bstream and copy pastes it to the end of the bstream.
Will clarify docs.

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
ripatel-fd: `spad_max` is the max supported copy size. 
Looking at this algorithm again, it's completely wrong because `read_imm` and `append` are both going to write to the same `wb_ring` which results in weird overlapping memcpy.
Need to rethink this.

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
ripatel-fd: Good point 👍  Will improve documentation.

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
ripatel-fd: `[seq_present,seq_future)` contains uncommitted data. These are appends that have not yet been committed, which means they are not visible to reads. Therefore, we conceptually don't consider them part of the write-back cache (the data hasn't been written yet).

On data commit, instead of copying the data into the write-back cache, we just extend the write-back cache to the uncommitted data region, and the uncommitted data will magically fall in place.

`seq_future` here acts as the "place where the next allocation should be put".

However, `wb_ring` does not only describe the write-back cache. It contains both the cache and the uncommitted data. And data flows from uncommitted to cache.

ripatel-fd: At least that's the notional concept I've tried to write up. We might as well change the notions to consider uncommitted writes as part of the write-back cache. Might be less cognitive overhead.

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
```diff
@@ -0,0 +1,249 @@
...
+void
+fd_vinyl_io_wq_completion( fd_vinyl_io_ur_t * io );
+
+/* io_uring userdata encoding ******************************************
+
+   io_uring userdata are arbitrary 64-bit words that are provided in SQE
+   and echoed back in corresponding CQE.  We use the userdata to encode
+   which request completed upon CQE receipt.  We need to minimally pack
+   the request type (read or write) and the request identifier.  For the
+   write path, this is an index, for the read path, this is a pointer to
```
jvarela-jump:
```suggestion
   write path, this is an index; for the read path, this is a pointer to
```

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
```diff
@@ -0,0 +1,249 @@
...
+static inline ulong
+ur_udata_pack_idx( ulong req_type, /* UR_REQ_* */
+                   ulong idx ) {
+  return (idx<<UR_REQ_TYPE_WIDTH) | (req_type & UR_REQ_TYPE_MASK);
+}
+
+static inline ulong
+ur_udata_pack_ptr( ulong  req_type,
+                   void * ptr ) {
+  return ( ((ulong)ptr) & ~0x7UL ) | (req_type & 0x7UL);
```
jvarela-jump:
```suggestion
  return ( ((ulong)ptr) & ~UR_REQ_TYPE_MASK ) | (req_type & UR_REQ_TYPE_MASK);
```

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
```diff
@@ -0,0 +1,249 @@
...
+}
+
+static inline ulong
+ur_udata_req_type( ulong udata ) {
+  return udata & 0x7UL;
```
jvarela-jump:
```suggestion
  return udata & UR_REQ_TYPE_MASK;
```

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
```diff
@@ -0,0 +1,249 @@
...
+}
+
+static inline ulong
+ur_udata_idx( ulong udata ) {
+  return udata >> 3;
```
jvarela-jump:
```suggestion
  return udata >> UR_REQ_TYPE_WIDTH;
```

src/vinyl/io/ur/fd_vinyl_io_ur_private.h
```diff
@@ -0,0 +1,249 @@
...
+}
+
+static inline void *
+ur_udata_ptr( ulong udata ) {
+  return (void *)( udata & ~0x7UL );
```
jvarela-jump:
```suggestion
  return (void *)( udata & ~UR_REQ_TYPE_MASK );
```

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
cali-jumptrading: makes sense! I think this model is simple and sleek, just needs to be documented.

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
cali-jumptrading: ok got it! I'll review the header `fd_vinyl_io.h`.

src/vinyl/io/ur/fd_vinyl_io_ur_wb.c
cali-jumptrading: got it makes sense. If my understanding is correct, `seq_cache` is the `fd_vinyl_io_t`'s view of where the write back cache starts. But due to trimming, the `wb_ring`'s internal seq0 might get extended backwards, so `ur->seq_cache` is not always the same as `wb.seq0`.

It seems cumbersome to have to maintain both seqs (`seq_cache` and `seq0`) by hand. Would an api like `advance_seq_cache` make sense?

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
```diff
@@ -0,0 +1,369 @@
...
+                  bad_dst ? "misaligned or NULL dst" :
+                  bad_sz  ? "misaligned sz"          :
+                            "not in past" ));
+
+  /* At this point, we have a valid read request.  Serve tail from
+     write-back cache. */
+
+  sz = wb_read( ur, dst, seq0, sz );
+  if( !sz ) return;
+  seq1 = seq0 + sz;
```
jvarela-jump:
```suggestion
```

src/vinyl/io/ur/fd_vinyl_io_ur_rd.c
```diff
@@ -0,0 +1,369 @@
...
+                  bad_seq ? "misaligned seq"         :
+                  bad_dst ? "misaligned or NULL dst" :
+                  bad_sz  ? "misaligned sz"          :
+                            "not in past" ));
+
+  /* At this point, we have a valid read request.  Serve tail from
+     write-back cache. */
+
+  sz = wb_read( ur, dst, seq0, sz );
+  if( !sz ) return;
```
jvarela-jump:
```suggestion
  ulong pend_sz = wb_read( ur, dst, seq0, sz );
  if( !pend_sz ) return;
  ulong read_sz = sz - pend_sz;
```
This is part 1 of 2.
Part 2 of 2 is the one that seems to require `read_sz`.
Could you please double check?

cali-jumptrading (approved):

PR #8117, opened 2026-02-02
reasm: remove xid from xid map when publishing slot

topointon-jump:
Fixes a bug where we were never removing entries from the xid map which would eventually cause a crash as the map would fill up.

### DIFF
--- a/src/discof/reasm/fd_reasm.c
+++ b/src/discof/reasm/fd_reasm.c
@@ -347,6 +347,9 @@ fd_reasm_insert( fd_reasm_t *      reasm,
     fec->popped         = 1;
     bid_t * bid         = bid_insert( reasm->bid, slot );
     bid->idx            = pool_idx( pool, fec );
+    xid_t * xid         = xid_insert( reasm->xid, ( slot << 32 ) | fec_set_idx );
+    if( FD_UNLIKELY( !xid ) ) FD_LOG_CRIT(( "xid map full, slot=%lu fec_set_idx=%u", slot, fec_set_idx ));
+    xid->idx            = pool_idx( pool, fec );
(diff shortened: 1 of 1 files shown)


src/discof/reasm/fd_reasm.c
```diff
@@ -521,9 +522,13 @@ fd_reasm_publish( fd_reasm_t * reasm, fd_hash_t const * merkle_root ) {
     bid_t * bid = bid_query( reasm->bid, head->slot, NULL );
     if( FD_UNLIKELY( bid ) ) bid_remove( reasm->bid, bid  ); /* only first FEC */
 
+    /* remove from the xid map */
+    xid_t * xid = xid_query( reasm->xid, ( head->slot << 32 ) | head->fec_set_idx, NULL );
+    if( FD_LIKELY( xid ) ) xid_remove( reasm->xid, xid );
```
lidatong: FD_LOG_CRIT here too if it's not in there

src/discof/reasm/fd_reasm.c
```diff
@@ -347,6 +347,9 @@ fd_reasm_insert( fd_reasm_t *      reasm,
     fec->popped         = 1;
     bid_t * bid         = bid_insert( reasm->bid, slot );
     bid->idx            = pool_idx( pool, fec );
+    xid_t * xid         = xid_insert( reasm->xid, ( slot << 32 ) | fec_set_idx );
+    if( FD_UNLIKELY( !xid ) ) FD_LOG_CRIT(( "xid map full, slot=%lu fec_set_idx=%u", slot, fec_set_idx ));
```
lidatong, line 351: you don't need this here, it's empty

PR #8120, opened 2026-02-02
flamenco, elf: ensure dynamic string table exists

topointon-jump:
.


mjain-jump:
dont forget to update the test vectors commit sha after

PR #8129, opened 2026-02-02
fuzz, capture: split up protobuf dumping and capture ctx

mjain-jump:


src/flamenco/runtime/tests/fd_dump_pb.h
```diff
@@ -66,6 +66,31 @@ FD_PROTOTYPES_BEGIN
...
+
+  /* Block Capture */
+  int                      dump_block_to_pb;
+
+  /* Syscall Capture */
+  int                      dump_syscall_to_pb;
+  char const *             dump_syscall_name_filter;
+
+  /* ELF Capture */
+  int                      dump_elf_to_pb;
```
ripatel-fd: use a bit field here

src/flamenco/runtime/fd_runtime_helpers.h
```diff
@@ -9,6 +9,7 @@
 #include "info/fd_instr_info.h"
 #include "../features/fd_features.h"
 #include "../capture/fd_capture_ctx.h"
+#include "tests/fd_dump_pb.h"
```
ibhatt-jumptrading: does this need to be here

src/flamenco/runtime/tests/fd_dump_pb.c
```diff
@@ -1148,10 +1148,14 @@ FD_SPAD_FRAME_BEGIN( dump_ctx->spad ) {
 void
 fd_dump_vm_syscall_to_protobuf( fd_vm_t const * vm,
                                 char const *    fn_name ) {
+  char const * syscall_name_filter = vm->instr_ctx->runtime->log.dump_proto_ctx->dump_syscall_name_filter;
```
ibhatt-jumptrading, line 1151: a cleaner way to do this to avoid a strcmp for every syscall is to map the syscall function pointers to the syscall names so all you have to do is a pointer comparison

src/flamenco/runtime/fd_runtime_helpers.h
mjain-jump: could put it in fd_runtime.h but figured this was cleaner since fd_runtime.h's only import is fd_runtime_helpers.h

ibhatt-jumptrading: forward declare it so you don't need to do the include in the .h file

mjain-jump: fixed

src/flamenco/runtime/tests/fd_dump_pb.c
mjain-jump: theres no current map that does this, and this would also need to be maintained separately every time a new syscall is added, so i think the current implementation should be fine for now

PR #8130, opened 2026-02-02
repair: forest eviction policy

### LINKED ISSUE #7838: repair: forest eviction policy

emwang-jump:
Closes #7838


src/discof/forest/fd_forest.c
```diff
@@ -992,7 +1007,7 @@ fd_forest_publish( fd_forest_t * forest, ulong new_root_slot ) {
         could be publishing backwards to a slot that we don't have. */
 
   if( FD_UNLIKELY( !new_root_ele ) ) {
-    new_root_ele = fd_forest_blk_insert( forest, new_root_slot, 0 );
+    new_root_ele = fd_forest_blk_insert( forest, new_root_slot, new_root_slot );
```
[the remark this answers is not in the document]
emwang-jump: this is pretty good dang

src/disco/shred/fd_fec_resolver.h
```diff
@@ -188,6 +188,7 @@ fd_fec_resolver_advance_slot_old( fd_fec_resolver_t * resolver,
 #define FD_FEC_RESOLVER_SHRED_IGNORED   (-1)
 #define FD_FEC_RESOLVER_SHRED_OKAY      ( 0)
 #define FD_FEC_RESOLVER_SHRED_COMPLETES ( 1)
+#define FD_FEC_RESOLVER_SHRED_DUPLICATE ( 2)
```
ptaffet-jump: This needs to go in the metrics, and the values below RETVAL_CNT, etc. need to be updated. IMO DUPLICATE should be a negative return value though

src/disco/shred/fd_fec_resolver.c
```diff
@@ -666,7 +666,10 @@ fd_fec_resolver_add_shred( fd_fec_resolver_t         * resolver,
     /* First, check to make sure this is not a duplicate */
     int shred_dup = !!(fd_uint_if( is_data_shred, ctx->set->data_shred_rcvd, ctx->set->parity_shred_rcvd ) & (1U << in_type_idx));
 
-    if( FD_UNLIKELY( shred_dup ) ) return FD_FEC_RESOLVER_SHRED_IGNORED;
+    if( FD_UNLIKELY( shred_dup ) ) {
+      if( FD_LIKELY( out_merkle_root && is_repair ) ) memcpy( out_merkle_root, ctx->root.hash, sizeof(fd_bmtree_node_t) );
+      return is_repair ? FD_FEC_RESOLVER_SHRED_DUPLICATE : FD_FEC_RESOLVER_SHRED_IGNORED;
```
ptaffet-jump: You need to add to the documentation when exactly DUPLICATE will be returned. I don't really understand the difference.  I think this code path should not depend on repair though.

I think ideally this change would be something like: populate the merkle root also when returning SHRED_IGNORED, but that is extra computation on several paths, so I'm not sure it is worth it.

So then I guess the docs should be something like: if the shred is part of an in progress FEC set but has already been received, returns DUPLICATE and populates out_merkle_root if it is non-NULL.

emwang-jump: Yeah looked into populating the merkle root when returning SHRED_IGNORED but didn't seem to be worth it.  Updated the docs to include DUPLICATE!

src/disco/store/fd_store.c
```diff
@@ -189,8 +189,8 @@ fd_store_insert( fd_store_t * store,
 
     fd_store_fec_t * fec = fd_store_pool_ele( &pool, idx );
     if( FD_UNLIKELY( fec ) ) {
-      FD_BASE58_ENCODE_32_BYTES( merkle_root->uc, _merkle_root );
-      FD_LOG_WARNING(( "duplicate insert %s", _merkle_root ));
+      //FD_BASE58_ENCODE_32_BYTES( merkle_root->uc, _merkle_root );
```
ptaffet-jump: Do you think it's worth it to increment a metric counter here?

ptaffet-jump:
The changes to fd_fec_resolver and the shred tile look good to me. I'll review the rest later.

src/discof/forest/fd_forest.c
```diff
@@ -514,15 +511,384 @@ query( fd_forest_t * forest, ulong slot ) {
...
+     2. Connected, unconfirmed leaves
+     3. Orphaned,  confirmed   leaves
+
+  We follow a general heuristic of evicting by newest slot in the
+  category first, with an exception.  If the newest slot in the category
+  is the parent of the slot we are adding, we take the second newest
+  slot in the category.  This is to avoid needlessly creating an orphan
+  that would immediately get evicted again by its parent getting
+  requested again.  If we have confirmations we can also avoid adding
+  new slots that we are certain won't get confirmed.
```
```suggestion
  We follow a general heuristic of evicting the leaf (youngest descendant) in each
  category first, with an exception.  If the leaf happens to be
  the parent of the new slot we are attempting to insert, we pick a different fork to evict from.  This is to avoid getting stuck in a cycle, where repair is repeatedly evicting ancestors with descedants, and therefore never repairing the fork.  If we have confirmations we can also avoid adding
  new slots that we are certain won't get confirmed.
```

src/discof/forest/fd_forest.c
```diff
@@ -514,15 +511,384 @@ query( fd_forest_t * forest, ulong slot ) {
...
+  to replay 6 to see it), we can follow a general policy of evicting the
+  newest unconfirmed leaf. Newest implies furthest from the root. So we
+  would evict 1995' first, and then add 6. */
+
+
+#define SLOT_IGNORE 0
+#define SLOT_INSERT 1
+
+static int
+fd_forest_evict( fd_forest_t * forest, ulong new_slot, ulong parent_slot, ulong * evicted ) {
```
lidatong: no `fd_forest` prefix if it's private to the module

src/discof/repair/fd_repair_tile.c
```diff
@@ -645,9 +645,8 @@ after_shred( ctx_t      * ctx,
 
     int slot_complete = !!(shred->data.flags & FD_SHRED_DATA_FLAG_SLOT_COMPLETE);
     int ref_tick      = shred->data.flags & FD_SHRED_DATA_REF_TICK_MASK;
-    fd_forest_blk_insert( ctx->forest, shred->slot, shred->slot - shred->data.parent_off );
-    //if( FD_UNLIKELY( ctx->profiler.enabled && shred->slot == ctx->profiler.end_slot ) ) fd_forest_blk_parent_update( ctx->forest, shred->slot, shred->slot - shred->data.parent_off );
-    fd_forest_data_shred_insert( ctx->forest, shred->slot, shred->slot - shred->data.parent_off, shred->idx, shred->fec_set_idx, slot_complete, ref_tick, src, mr, cmr );
+    fd_forest_blk_t * blk = fd_forest_blk_insert( ctx->forest, shred->slot, shred->slot - shred->data.parent_off );
```
lidatong: add metric and log here instead of inside library

src/discof/forest/fd_forest.c
```diff
@@ -514,15 +511,384 @@ query( fd_forest_t * forest, ulong slot ) {
...
+        1. One fork in orphans, and root is alone (common situation in
+           catchup). The new slot's parent is the tip of the orphan
+           fork.  Ignore the slot in this case.
+        2. One long fork, and the new slot's parent is the tip of the
+           fork. Force a root in this case. */
+    if( fd_forest_orphaned_ele_query( orphaned, &parent_slot, NULL, pool ) ) return SLOT_IGNORE;
+    *evicted       = fd_forest_pool_ele( pool, forest->root )->slot;
+    ulong new_root = fd_forest_pool_ele( pool, forest->root )->child;
+    if( FD_UNLIKELY( !fd_forest_pool_ele( pool, new_root )->confirmed ) ) {
+      FD_LOG_WARNING(( "Forest tried to force root on slot %lu but it is not confirmed yet. Ignoring.", fd_forest_pool_ele( pool, new_root )->slot ));
```
lidatong: move this into tile logging and add metric

src/discof/forest/fd_forest.c
```diff
@@ -558,16 +924,17 @@ acquire( fd_forest_t * forest, ulong slot, ulong parent_slot ) {
...
 fd_forest_blk_t *
 fd_forest_blk_insert( fd_forest_t * forest, ulong slot, ulong parent_slot ) {
 # if FD_FOREST_USE_HANDHOLDING
   FD_TEST( slot > fd_forest_root_slot( forest ) ); /* caller error - inval */
 # endif
+  /* we don't want to add a slot to the forest chains older than root,
+   to avoid filling forest up with junk. Can't rely on publishing to
+   prune these useless subtrees if we are having trouble rooting.
+   TODO: do the same with reasm/store/shred? */
+  if( FD_UNLIKELY( parent_slot < fd_forest_root_slot( forest ) ) ) return NULL;
```
lidatong: do this check at the tile level

src/discof/forest/fd_forest.c
lidatong: there's only two cases, so cleaner to just return the evicted slot (ULONG_MAX if slot_ignore)

src/discof/forest/fd_forest.c
```diff
@@ -514,15 +511,384 @@ query( fd_forest_t * forest, ulong slot ) {
...
+  if( FD_UNLIKELY( unconfrmd_orphan )) {
+    *evicted = unconfrmd_orphan->slot;
+    clear_leaf( forest, unconfrmd_orphan->slot );
+    return SLOT_INSERT;
+  }
+  if( FD_UNLIKELY( unconfrmd_leaf )) {
+    *evicted = unconfrmd_leaf->slot;
+    clear_leaf( forest, unconfrmd_leaf->slot );
+    return SLOT_INSERT;
+  }
```
lidatong, line 813: with above change to return the slot # you this can be just one liner 
```suggestion
  if( FD_UNLIKELY( unconfrmd_orphan )) return clear_leaf( forest, unconfrmd_orphan->slot );
```

src/discof/forest/fd_forest.c
```diff
@@ -514,15 +511,384 @@ query( fd_forest_t * forest, ulong slot ) {
...
+  if( FD_UNLIKELY( confirmed_orphan )) {
+    fd_forest_blk_t * parent = query( forest, parent_slot );
+    /* Always accept a new orphan, as it could bring us closer to confirmation */
```
```suggestion
    /* Always accept a new orphan subtree root, as it could bring us closer to confirmation */
```

lidatong: is this actually the right policy? what if it's an orphan on another fork? i assume this is handling the case it's an ancestor of the confirmed orphan right?

src/discof/forest/fd_forest.c
```diff
@@ -514,15 +511,384 @@ query( fd_forest_t * forest, ulong slot ) {
...
+       tree, and the parent of the non confirmed fork is is our parent.
+       in this case we should just ignore this insert. TODO: optionally
+       we could evict the non confirmed fork if its a separate fork.
+         (2) we could have one orphan fork where parent_slot is at the
+       tip, and everything in main tree is confirmed. in this case we
+       should also ignore this insert. */
+    return SLOT_IGNORE;
+  }
+  /* unreachable */
+  return SLOT_IGNORE;
```
lidatong: remove this and last `if` and instead FD_TEST( confirmed_leaf )

src/discof/forest/fd_forest.c
emwang-jump: Yeah this is debatable, would probably get changed to explicitly check it's the ancestor of the confirmed orphan

src/discof/forest/fd_forest.c
emwang-jump: its more verbose (will need to return a struct or pass multiple pointers) to differentiate between force rooting vs. evicting and also trying to force root vs. failed eviction, is it okay to leave this here? I've moved everything else into tile logging

PR #8131, opened 2026-02-02
rpc: serve genesis.tar.bz2

jherrera-jump:


src/discof/rpc/fd_rpc_tile.c
```diff
@@ -202,6 +233,21 @@ struct fd_rpc_tile {
   int has_genesis_hash;
   uchar genesis_hash[ 32 ];
 
+#define FD_RPC_TAR_SZ (FD_GENESIS_MAX_MESSAGE_SIZE + 4UL*512UL)
+  uchar genesis_tar[ FD_RPC_TAR_SZ ];
+  ulong genesis_tar_sz;
+
+  /* From bzip2 docs:
+       To guarantee that the compressed data will fit in its buffer,
```
mmcgee-jump, line 260: What the heck lmao

mmcgee-jump: Is this floor though, maybe it should be ceil(x/100) ?

mmcgee-jump: Could be worth fuzzing this whole rx path

mmcgee-jump (approved):

src/discof/rpc/fd_rpc_tile.c
jherrera-jump: yeah, working on fuzz test for rx path

mmcgee-jump (approved):

PR #8138, opened 2026-02-03
Use mask for stake history offset update

mubarizkyc:
assuming ```fd_stake_history_size``` will always be a power of two


mmcgee-jump:
Not sure why there's 6 commits here, please submit clean PRs if you want them considered for merging

PR #8139, opened 2026-02-03
accdb: fix overlapping bitmask value

two-heart:


ripatel-fd (approved):

ripatel-fd:
Thank you

intrigus-lgtm (commented):
Only had a passing look at this, but it seems that `FD_ACCDB_FLAG_DONTZERO` is implemented internally as CREATE|TRUNCATE?
At least when one looks at any fd_accdb_open_rw implementation, there is no separate flag handling for DONTZERO:
https://github.com/firedancer-io/firedancer/blob/4e0729ce3c6b89baf7aa8f260ed15aaff86ba14d/src/flamenco/accdb/fd_accdb_impl_v1.c#L143-L219
and that's also why the tests are failing here as DONTZERO with this PR is not included as a valid flag:
```c
if( FD_UNLIKELY( flags & ~(FD_ACCDB_FLAG_CREATE|FD_ACCDB_FLAG_TRUNCATE) ) ) {
    FD_LOG_CRIT(( "invalid flags for open_rw: %#02x", (uint)flags ));
  }
```

PR #8140, opened 2026-02-03
update_sysvar_epoch_schedule_write

mubarizkyc:
updated  ```fd_update_sysvar_epoch_schedule_write``` to make it a bit consistent with ```fd_sysvar_{*}_write```


ripatel-fd:
@mubarizkyc Could you remove the trailing whitespace from the patch?

mubarizkyc:
@ripatel-fd check now

PR #8151, opened 2026-02-04
deps: openssl + msan

two-heart:
to make backtest with msan work

### DIFF
--- a/deps.sh
+++ b/deps.sh
@@ -502,66 +502,74 @@ install_openssl () {
   cd "$PREFIX/git/openssl"
 
   echo "[+] Configuring OpenSSL"
-  ./config \
-    -fPIC \
-    --prefix="$PREFIX" \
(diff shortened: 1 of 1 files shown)


deps.sh
```diff
@@ -502,66 +502,77 @@ install_openssl () {
...
+    no-srtp
+    no-ts
     no-whirlpool
+  )
+
+  if [[ $MSAN == 1 ]]; then
+    CONFIG_OPTS+=( enable-msan no-asm )
+    export CFLAGS="$EXTRA_CFLAGS"
+    export CXXFLAGS="$EXTRA_CXXFLAGS"
+    export LDFLAGS="$EXTRA_LDFLAGS"
```
ripatel-fd: Could you remove these exports? I'm worried these will bleed into other build functions since they are global variables.

PR #8157, opened 2026-02-04
Add insert_maybe_exist interface to fd_map_dynamic

palmtenor:
In many occasions we could be in the situation where the key we are trying to insert is already in the map. We often need to do a bit different processing (different response, update timestamp, etc.) for that case, hence currently we are doing a lot of query-check-insert, which would do map lookup twice.

This commits adds a `insert_maybe_exist` interface the either return the existing map entry or insert a new one, and communication about which case it is to the caller.

### DIFF
--- a/src/util/tmpl/fd_map_dynamic.c
+++ b/src/util/tmpl/fd_map_dynamic.c
@@ -110,6 +110,15 @@
 
     mymap_t * mymap_insert( mymap_t * map, ulong key );
 
+    // Same as mymap_insert, except if key is already in the map, returns
+    // the existing map entry and set value pointed by exist to 1.
+    // Otherwise, the inserted map entry will be returned and set value
+    // pointed by exist to 0.
+    // Still returns NULL on other failures and value pointed by exist
+    // will be undefined.
+
+    mymap_t * mymap_insert_maybe_exist( mymap_t * map, ulong key, int * exist );
+
     // Remove entry from map, fast O(1).  Assumes map is a current join
     // and that entry points to a full entry currently in the map.
     // Removal performance very slightly more optimal if sizeof(mymap_t)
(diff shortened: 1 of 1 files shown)


kbowers-jump:
Does fd_map_slot.c upsert already cover this case?  ```     // mymap_upsert:
     //
     //   ... join is a current local join and key points to the key of
     //   ... interest (for maps that use key sentinels, *key should not
     //   ... be a sentinel key), retains no interest in join or key.
     //
     //   ele = mymap_upsert( join, key );
     //
     //   if( FD_UNLIKELY( !ele ) ) {
     //
     //     ... element *key is not in map and the map is too full to
     //     ... insert it
     //
     //   } else if( FD_UNLIKELY( mymap_ele_is_free( ele ) ) ) {
     //
     //     ... element *key was not in map and should be inserted at
     //     ... ele.  See insert for details how to complete inserting
     //     ... ele.
     //
     //     ... Note that if mymap uses key sentinels to indicate when
     //     ... an element is free, this code path will never be taken.
     //     ... That is, the caller will not be able to tell via
     //     ... mymap_ele_is_free if an upsert is inserting or updating
     //     ... as the upsert initialized ele's key field (implicitly
     //     ... marking the element as not free such).
     //
     //   } else {
     //
     //     ... element *key is stored in the map at ele.  See update
     //     ... for details how to complete updating ele.
     //
     //     ... Note that if mymap uses key sentinels to indicate when
     //     ... an element is free, we might actually be inserting ele
     //     ... here (see note above).
     //
     //   }

     myele_t * mymap_upsert( mymap_t * join, ulong const * key );```

palmtenor:
I think mostly yes. But `mymap_upsert` is not available in the non-slot (`fd_map_dynamic`) map?
`fd_map_dynamic` also doesn't have the explicit entry-free-indicator (like `mymap_ele_is_free`) to make the upsert result more clear... (if it's an insert or update)

palmtenor:
Actually I see your point now, we didn't have `fd_map_slot` in our FD commit...

Issue #8144, opened 2026-02-03
SIMD-0444: Relax program data account check in migration


topointon-jump:
https://github.com/anza-xyz/agave/pull/9891

https://github.com/solana-foundation/solana-improvement-documents/pull/444

PR #8184, opened 2026-02-05
flamenco, runtime: implement relax_programdata_account_check_migration

topointon-jump:
https://github.com/firedancer-io/firedancer/issues/8144

- [x] Implementation

When Agave 4.0 is cut:
- [ ] Ledger
- [ ] Fuzzing


mjain-jump (approved):
approving assuming ledger and unit tests pass

PR #8188, opened 2026-02-06
backtest: add replay histogram

ripatel-fd:


src/disco/metrics/metrics.xml
```diff
@@ -1169,6 +1169,10 @@ metric introduced.
 <tile name="backt" telemetry="false">
     <gauge name="FinalSlot" summary="The slot after which the backtest will complete" />
     <gauge name="StartSlot" summary="The slot at which the backtest started" />
+
+    <histogram name="SlotReplayDurationSeconds" min="0.01" max="0.5" converter="seconds">
```
mmcgee-jump, line 1173: Why is this a backtest metric and not `replay` tile metric? Seems useful in production

ripatel-fd: @mmcgee-jump Could add in the future. Replay incrementally replays as block data comes in, and handles multiple forks, so it's more difficult to track the timing for those cases.

mmcgee-jump: It's literally trivial? Add a start_time or replayed_time field to the bank?

ripatel-fd: @mmcgee-jump don't have time to work on this at the moment, feel free to do the change

PR #8192, opened 2026-02-06
xdp: fix uninitialized variable in test

intrigus-lgtm:
The during_frag_expected_sz variable was not initialized leading to mismatches in fd_memeq later on. This bug is surprisingly difficult to trigger as neither msan, ubsan nor asan were able to catch the problem. (At least with enabled optimizations)


nlgripto:
crazy find

intrigus-lgtm:
Sorry for the spam, but MSAN even with optimizations disabled didn't catch that we were comparing an uninitialized padding byte as the struct wasn't packed.
And this issue only reproduced in CI, which made debugging this very annoying.

mmcgee-jump (approved):

Issue #6848, opened 2025-10-22
Invalid tile_heardbeat in metrics


ripatel-fd:
These are not unix nanos

```
# HELP tile_heartbeat The last UNIX timestamp in nanoseconds that the tile heartbeated.
# TYPE tile_heartbeat gauge
tile_heartbeat{kind="netlnk",kind_id="0"} 54892910537315
tile_heartbeat{kind="net",kind_id="0"} 54892967467391
tile_heartbeat{kind="metric",kind_id="0"} 54892943773352
tile_heartbeat{kind="snapct",kind_id="0"} 52738181652845
tile_heartbeat{kind="snapld",kind_id="0"} 52738181957809
tile_heartbeat{kind="snapdc",kind_id="0"} 52738181961784
tile_heartbeat{kind="snapin",kind_id="0"} 52739266361844
```

PR #8201, opened 2026-02-07
fix: use UNIX nanoseconds for tile_heartbeat gauge

N3ro47:
Fixes [Issue #6848](https://github.com/firedancer-io/firedancer/issues/6848)

Changed `tile_heartbeat` to report UNIX nanoseconds using `fd_log_wallclock()` instead of raw ticks.

I'm still learning the tribal knowledge of the project, but my assessment is that `fd_log_wallclock` is performant enough for the housekeeping path. It maps to `clock_gettime(CLOCK_REALTIME)`, which typically utilizes the vDSO path.


N3ro47:
Done. Also removed obsolete ns_per_tic and tic/toc logic. The monitor loop now uses fd_log_wallclock() directly

PR #8207, opened 2026-02-09
shred, firedancer, frankendancer: correctly compute feature activation slot for turbine features

topointon-jump:
Changes in Frank:
- feature activation slots now return turbine slots (old code was epochs)
- passing features from poh->shred only when they're different (reduced a lot the number of messages, before it was pretty much every slot if not more)

Changes in Firedancer:
- pass epoch schedule
- compute turbine slot from feature slot

### DIFF
--- a/src/disco/shred/fd_shred_tile.c
+++ b/src/disco/shred/fd_shred_tile.c
@@ -254,6 +254,7 @@ typedef struct {
     };
   } pending_batch;
 
+  fd_epoch_schedule_t            epoch_schedule[1];
   fd_shred_features_activation_t features_activation[1];
   /* too large to be left in the stack */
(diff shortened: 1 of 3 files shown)


src/discoh/pohh/fd_pohh_tile.c
```diff
@@ -1456,11 +1461,14 @@ static inline void
 publish_features_activation(  fd_pohh_tile_t *    ctx,
                               fd_stem_context_t * stem ) {
   uchar * dst = (uchar *)fd_chunk_to_laddr( ctx->shred_out->mem, ctx->shred_out->chunk );
-  fd_shred_features_activation_t * act_data = (fd_shred_features_activation_t *)dst;
+  /* send epoch_schedule */
+  fd_memcpy( dst, ctx->epoch_schedule, sizeof(fd_epoch_schedule_t) );
+  /* send fefeatures_activation */
```
jvarela-jump:
```suggestion
  /* send features_activation */
```

PR #8208, opened 2026-02-09
backtest: move ledger testing to binary

kbhargava-jump:


ripatel-fd:
@kbhargava-jump can we revive this now that we have shredcap?

PR #8218, opened 2026-02-10
resolv: use exact block height check

cavemanloverboy:
resolv just checks if a transaction's blockhash slot is within 151 slots from the current reference slot, which needlessly filters some transactions in the case there were some recently skipped slots. the exact block height check is possible no additional overhead in full firedancer, and with a slight modification to discoh's message.

note: this REQUIRES the changes to the agave submodule to work

https://github.com/cavemanloverboy/agave-fd/commit/5149fb5b5744c3913634cc7f2dd0ba56d034b017


HGuillemet:
Duplicate of #7128

PR #8219, opened 2026-02-10
use only strict cost ,for ed25519 sig verification

mubarizkyc:
refrencing [from](https://github.com/anza-xyz/agave/pull/10158)


mubarizkyc:
@0x0ece probably  this also requires merging refrenced agave commit

PR #8225, opened 2026-02-10
txsend: light belt sanding

mmcgee-jump:


src/discof/txsend/fd_txsend_tile.c
```diff
@@ -192,416 +166,324 @@ quic_tx_aio_send( void *                    _ctx,
...
+   doesn't expect this behavior and will underflow.  (That's also not
+   ideal, in case some plugin wanted to listen reliably on quic->verify
+   they could not, if it got underflowed).  Here though, we want to
+   avoid dropping outgoing votes to verify, since they might be needed
+   for liveness of a small cluster.
+
+   We thus take the trade of servicing QUIC in after_credit, which means
+   it could theoretically get backpressured by verify, however this
+   isn't realistic in practice, as verify polls round robin and there's
+   only one vote per slot. */
```
ripatel-fd, line 205: Unfortunate shortcoming of the stem design. Shouldn't have independent actors blocking each other and track flow control credits per out link. But should be good enough for the non-contended case.

PR #8258, opened 2026-02-11
runtime: simplify bank object initialization

ripatel-fd:
Simplifies addressing of objects that are shared across multiple
banks.  For convenience, each bank retains position-independent
pointers to objects managed by fd_banks.  Since the difference
between (bank[i]-fd_banks) and (bank[i]<->cost_tracker) is constant
for each i, each bank only needs to store a pointer to fd_banks.

Also, we remove the assumption that certain bank fields are
retained while the bank is in the free pool.  (All fields are
initialized on bank object creation now.)  This allows us to install
stronger UAF detection using MSan.

### DIFF
--- a/src/flamenco/runtime/fd_bank.c
+++ b/src/flamenco/runtime/fd_bank.c
@@ -272,7 +272,6 @@ void *
 fd_banks_new( void * shmem,
               ulong  max_total_banks,
               ulong  max_fork_width,
-              int    larger_max_cost_per_block,
               ulong  seed ) {
   if( FD_UNLIKELY( !shmem ) ) {
     FD_LOG_WARNING(( "NULL shmem" ));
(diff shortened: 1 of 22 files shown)


src/flamenco/runtime/fd_bank.c
```diff
@@ -554,11 +537,17 @@ fd_banks_init_bank( fd_bank_t *  bank_l,
...
+  bank->vote_states_dirty    = 1;
+  fd_bank_vote_states_t * vote_states = fd_bank_vote_states_pool_ele( vote_states_pool, bank->vote_states_pool_idx );
+  if( FD_UNLIKELY( !fd_vote_states_new( vote_states->data, FD_RUNTIME_MAX_VOTE_ACCOUNTS, banks->data->hash_seed ) ) ) {
+    FD_LOG_WARNING(( "Failed to create vote states" ));
+    return NULL;
+  }
   fd_rwlock_new( &bank_l->locks->vote_states_lock[ bank->idx ] );
 
-  bank->cost_tracker_pool_idx = fd_bank_cost_tracker_pool_idx_null( fd_bank_get_cost_tracker_pool( bank ) );
+  bank->cost_tracker_pool_idx  = fd_bank_cost_tracker_pool_idx_null( NULL );
```
ibhatt-jumptrading, line 561: nit extra space here

PR #8261, opened 2026-02-12
replay: make scheduler depth configurable

ripatel-fd:
Make replay dispatcher depth configurable to save memory.
fd_rdisp_new takes a considerable amount of time with MSan (30s)
with the previous default.

Decreases live client depth to 1M and backtest depth to 8K.

### DIFF
--- a/src/app/firedancer-dev/commands/backtest.c
+++ b/src/app/firedancer-dev/commands/backtest.c
@@ -520,6 +520,7 @@ extern int * fd_log_private_shared_lock;
 
 static void
 backtest_cmd_topo( config_t * config ) {
+  config->firedancer.development.replay.scheduler_depth = 8192UL;
   backtest_topo( config );
 }
 
--- a/src/discof/replay/fd_sched.h
+++ b/src/discof/replay/fd_sched.h
@@ -50,6 +50,9 @@
    more ingest, more ready, more done ...
    ... */
 
+#define FD_SCHED_MIN_DEPTH 478
+#define FD_SCHED_MAX_DEPTH FD_RDISP_MAX_DEPTH
+
 struct fd_sched;
 typedef struct fd_sched fd_sched_t;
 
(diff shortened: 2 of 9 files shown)


src/app/firedancer-dev/commands/backtest.c
```diff
@@ -520,6 +520,7 @@ extern int * fd_log_private_shared_lock;
 
 static void
 backtest_cmd_topo( config_t * config ) {
+  config->firedancer.development.replay.scheduler_depth = 8192UL;
```
yufeng-jump, line 523: Change default in run_ledger_backtest.sh and offline_replay.toml

src/discof/replay/fd_sched.h
```diff
@@ -50,6 +50,9 @@
    more ingest, more ready, more done ...
    ... */
 
+#define FD_SCHED_MIN_DEPTH (32UL)
```
yufeng-jump: Can you do 512 and in sched.c assert that this is larger than FD_SCHED_MAX_TXN_PER_FEC

src/discof/replay/fd_sched.h
```diff
@@ -50,6 +50,9 @@
...
+#define FD_SCHED_MAX_DEPTH ((1UL<<20)-1UL)
```
yufeng-jump: Why not just make it 1<<23

yufeng-jump (dismissed):
lgtm overall just a few nits

src/discof/replay/fd_sched.h
ripatel-fd: Yes that's a bug, good catch

src/app/firedancer-dev/commands/backtest.c
ripatel-fd: Would actually prefer to leave this here, because backtest should have sane defaults. Ideally, we'd have a separate default config snippet for each subcommand

PR #8266, opened 2026-02-12
restore: reduce snapla tile footprint

ripatel-fd:
Removes unused manifest parser


ripatel-fd:
Looks like this was required, unfortunate

PR #8273, opened 2026-02-12
tsa: base support

two-heart:
Introduces basic Thread Safety Analyser support. In contrast to #7189, this does only adds support + a script to see the TSA warnings. As discussed offline with @ripatel-fd, follow up PRs will add annotations support to individual components. Later with the follow-up PRs merged we can decide if we want to run TSA on every PR, or if we want to use the sarif v2 output from the checker script and feed it into GitHub code scanning in nightly.

### DIFF
--- a/src/flamenco/fd_rwlock.h
+++ b/src/flamenco/fd_rwlock.h
@@ -4,8 +4,9 @@
 /* A very simple read-write spin lock. */
 
 #include "../util/fd_util_base.h"
+#include "../util/sanitize/fd_tsa.h"
 
-struct fd_rwlock {
+struct FD_CAPABILITY("fd_rwlock") fd_rwlock {
   ushort value; /* Bits 0..16 are
 
                     0: Unlocked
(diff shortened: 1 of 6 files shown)


src/flamenco/fd_rwlock.h
```diff
@@ -15,14 +16,15 @@ struct fd_rwlock {
...
+
 static inline fd_rwlock_t *
 fd_rwlock_new( fd_rwlock_t * lock ) {
   lock->value = 0;
   return 0;
 }
 
 static inline void
-fd_rwlock_write( fd_rwlock_t * lock ) {
+fd_rwlock_write( fd_rwlock_t * lock ) FD_ACQUIRE( lock ) FD_NO_THREAD_SAFETY_ANALYSIS  {
```
ripatel-fd, line 27: why is this disabled?

two-heart: To suppress TSA checks inside the lock primitive implementation itself, while still keeping lock state tracking for callers. This is the common way to annotate the lock itself, see e.g. https://github.com/torvalds/linux/blob/master/tools/perf/util/mutex.c#L58

PR #8283, opened 2026-02-12
snapshots: verify epoch stakes

cali-jumptrading:


src/discof/restore/fd_snapin_tile.c
```diff
@@ -153,6 +153,35 @@ verify_slot_deltas_with_slot_history( fd_snapin_tile_t * ctx ) {
   return 0;
 }
 
+/* verification of epoch stakes from manifest
+   TODO: is leader schedule epoch always 1 slot ahead of bank epoch?
```
ripatel-fd, line 157: No it's not for all the clusters, it gets determined somewhere in the middle of the epoch (leader schedule offset)

cali-jumptrading: I think we assume the leader schedule epoch is always 1 ahead of the bank epoch in the ssmanifest parser... I'll make the change and open a new PR then.

ripatel-fd: I see, yeah we can fix it in a follow up PR

PR #8287, opened 2026-02-12
replay: reasm eviction policy

emwang-jump:
Closes https://github.com/firedancer-io/firedancer/issues/7837
needs:
- (1) fec resolver v2 PR
- (2) repair fec chaining PR

### DIFF
--- a/src/app/firedancer/topology.c
+++ b/src/app/firedancer/topology.c
@@ -8,6 +8,7 @@
 #include "../../discof/tower/fd_tower_tile.h"
 #include "../../discof/resolv/fd_resolv_tile.h"
 #include "../../discof/repair/fd_repair.h"
+#include "../../discof/reasm/fd_reasm.h"
 #include "../../discof/replay/fd_replay_tile.h"
 #include "../../disco/net/fd_net_tile.h"
 #include "../../discof/restore/fd_snapct_tile.h"
@@ -449,7 +450,7 @@ fd_topo_initialize( config_t * config ) {
   fd_topob_wksp( topo, "gossip_out"    );
 
   fd_topob_wksp( topo, "shred_out"     );
-  fd_topob_wksp( topo, "repair_shred"  );
+  fd_topob_wksp( topo, "repair_out" );
   fd_topob_wksp( topo, "replay_epoch"  );
   fd_topob_wksp( topo, "replay_execrp" );
   fd_topob_wksp( topo, "replay_out"    );
@@ -481,6 +482,7 @@ fd_topo_initialize( config_t * config ) {
   fd_topob_wksp( topo, "banks"         );
   fd_topob_wksp( topo, "banks_locks"   );
   fd_topob_wksp( topo, "store"         )->core_dump_level = FD_TOPO_CORE_DUMP_LEVEL_FULL;
+  fd_topob_wksp( topo, "rnonce"        );
 
   fd_topob_wksp( topo, "gossip_sign"   );
   fd_topob_wksp( topo, "sign_gossip"   );
@@ -648,7 +650,7 @@ fd_topo_initialize( config_t * config ) {
   /**/                 fd_topob_link( topo, "sign_txsend",   "sign_txsend",   128UL,                                    sizeof(fd_ed25519_sig_t)*2UL,  1UL ); /* TODO: Depth probably doesn't need to be 128 */
 
   FOR(shred_tile_cnt)  fd_topob_link( topo, "shred_out",     "shred_out",     shred_depth,                              FD_SHRED_OUT_MTU,              3UL ); /* TODO: Pretty sure burst of 3 is incorrect here */
-  FOR(shred_tile_cnt)  fd_topob_link( topo, "repair_shred",  "repair_shred",  shred_depth,                              sizeof(fd_ed25519_sig_t),      1UL );
+  /**/                 fd_topob_link( topo, "repair_out",    "repair_out",    shred_depth,                              FD_SHRED_OUT_MTU,              1UL );
   /**/                 fd_topob_link( topo, "tower_out",     "tower_out",     16384UL,                                  sizeof(fd_tower_msg_t),        2UL ); /* conf + slot_done. see explanation in fd_tower_tile.h for link_depth */
   /**/                 fd_topob_link( topo, "txsend_out",    "txsend_out",    128UL,                                    FD_TPU_RAW_MTU,                1UL );
 
@@ -917,14 +919,15 @@ fd_topo_initialize( config_t * config ) {
 
   /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "genesi_out",    0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "gossip_out",    0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
+  /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "replay_out",    0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   FOR(shred_tile_cnt)  fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "shred_out",     i,            FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   if( snapshots_enabled ) {
                        fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "snapin_manif",  0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   }
   /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "tower_out",     0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
-  FOR(shred_tile_cnt)  fd_topob_tile_out(   topo, "repair",  0UL,                       "repair_shred",  i                                                  );
+  /**/                 fd_topob_tile_out(   topo, "repair",  0UL,                       "repair_out",    0                                                  );
 
-  FOR(shred_tile_cnt)  fd_topob_tile_in(    topo, "replay",  0UL,          "metric_in", "shred_out",     i,            FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
+  /**/                 fd_topob_tile_in (   topo, "replay",  0UL,          "metric_in", "repair_out",    0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   /**/                 fd_topob_tile_in (   topo, "replay",  0UL,          "metric_in", "genesi_out",    0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   /**/                 fd_topob_tile_out(   topo, "replay",  0UL,                       "replay_out",    0UL                                                );
   /**/                 fd_topob_tile_out(   topo, "replay",  0UL,                       "replay_epoch",  0UL                                                );
@@ -996,7 +999,6 @@ fd_topo_initialize( config_t * config ) {
   FOR(shred_tile_cnt)    fd_topob_tile_in ( topo, "shred",   i,            "metric_in", "replay_epoch",  0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   FOR(shred_tile_cnt)    fd_topob_tile_in ( topo, "shred",   i,            "metric_in", "gossip_out",    0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   FOR(shred_tile_cnt)    fd_topob_tile_out( topo, "shred",   i,                         "shred_out",     i                                                  );
-  FOR(shred_tile_cnt)    fd_topob_tile_in ( topo, "shred",   i,            "metric_in", "repair_shred",  i,            FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   FOR(shred_tile_cnt)    fd_topob_tile_in ( topo, "shred",   i,            "metric_in", "ipecho_out",    0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   FOR(shred_tile_cnt)    fd_topob_tile_in ( topo, "shred",   i,            "metric_in", "tower_out",     0UL,          FD_TOPOB_UNRELIABLE, FD_TOPOB_POLLED );
   FOR(shred_tile_cnt)    fd_topob_tile_out( topo, "shred",   i,                         "shred_net",     i                                                  );
@@ -1207,7 +1209,7 @@ fd_topo_initialize( config_t * config ) {
      It's not super security sensitive, but for good hygiene, we make it
      an object. */
   if( 1 /* just restrict the scope for these variables in this big function */ ) {
-    fd_topo_obj_t * rnonce_ss_obj = fd_topob_obj( topo, "rnonce_ss", "repair_shred" );
+    fd_topo_obj_t * rnonce_ss_obj = fd_topob_obj( topo, "rnonce_ss", "rnonce" );
     fd_topo_tile_t * repair_tile = &topo->tiles[ fd_topo_find_tile( topo, "repair", 0UL ) ];
     fd_topob_tile_uses( topo, repair_tile, rnonce_ss_obj, FD_SHMEM_JOIN_MODE_READ_WRITE );
     for( ulong i=0UL; i<shred_tile_cnt; i++ ) {
@@ -1348,16 +1350,32 @@ fd_topo_initialize( config_t * config ) {
   FD_TEST( fd_pod_insertf_ulong( topo->props, fec_sets_obj->id, "fec_sets" ) );
 
    /* The Store fec_max parameter is the max number of FEC sets that can
-      be retained before the Tower consensus algorithm indicates a new
-      "root", which allows pruning FECs.
+      be retained by store.
 
-      The default value is from multiplying max_live_slots by the
-      maximum number of FEC sets in a block (currently 32768 but can be
-      reduced to 1024 once FECs are restricted to always be size 32,
-      given the current consensus limit of 32768 shreds per block). */
+      The base value is from multiplying max_live_slots by the maximum
+      number of FEC sets in a block (which is 1024, given the current
+      consensus limit of 32768 shreds per block).  This is notably, the
+      total capacity of reasm.
 
-  if( FD_UNLIKELY( !fd_ulong_is_pow2( config->firedancer.runtime.max_live_slots ) ) ) FD_LOG_ERR(( "max_live_slots must be a power of 2 for store" ));
-  fd_topo_obj_t * store_obj = setup_topo_store( topo, "store", config->firedancer.runtime.max_live_slots * 1024UL /* FIXME temporary hack to run on 512 gb boxes */, (uint)shred_tile_cnt );
+      Store needs to hold _at least_ depth(shred_out) +
+      depth(repair_out) + 1 more FEC sets than reasm.  Shred inserts
+      FECs into store (before publishing to shred_out) and Replay
+      inserts FECs into reasm (after consuming from repair_out), so
+      store can be up to depth(shred_out) + depth(repair_out) ahead of
+      reasm.  We + 1 because replay tile can be mid-process of ingesting
+      a FEC set.
+
+      store is exactly this total depth ahead of reasm, then this means
+      the links must be full and shred is backpressured via repair_out
+      -> shred_out.  Thus, no more FEC sets will be inserted to store
+      until reasm inserts the next FEC, at which point it must be at
+      capacity (given store is _at least_ total depth larger) and will
+      evict a FEC which is together evicted from store. */
+
+  fd_topo_link_t * repair_out_link = &topo->links[ fd_topo_find_link( topo, "repair_out", 0UL ) ];
+  ulong store_fec_max = fd_ulong_pow2_up( config->firedancer.runtime.max_live_slots * FD_FEC_BLK_MAX + (shred_depth * shred_tile_cnt) + repair_out_link->depth + 1 );
+
+  fd_topo_obj_t * store_obj = setup_topo_store( topo, "store", store_fec_max, (uint)shred_tile_cnt );
   FOR(shred_tile_cnt) fd_topob_tile_uses( topo, &topo->tiles[ fd_topo_find_tile( topo, "shred", i ) ], store_obj, FD_SHMEM_JOIN_MODE_READ_WRITE );
   fd_topob_tile_uses( topo, &topo->tiles[ fd_topo_find_tile( topo, "replay", 0UL ) ], store_obj, FD_SHMEM_JOIN_MODE_READ_WRITE );
   FD_TEST( fd_pod_insertf_ulong( topo->props, store_obj->id, "store" ) );
--- a/src/discof/forest/fd_forest.c
+++ b/src/discof/forest/fd_forest.c
@@ -212,7 +212,7 @@ fd_forest_init( fd_forest_t * forest, ulong root_slot ) {
   root_ele->sibling          = null;
   root_ele->buffered_idx     = 0;
   root_ele->complete_idx     = 0;
-  root_ele->confirmed        = 1;
+  root_ele->chain_confirmed  = 1;
 
   root_ele->merkle_roots[0].mr = (fd_hash_t){ .key = { 0 } };
 
@@ -607,7 +607,7 @@ latest_confirmed_slot( fd_forest_t * forest, ulong root_idx ) {
 
   while( FD_LIKELY( !fd_forest_deque_empty( queue ) ) ) {
     fd_forest_blk_t * blk = fd_forest_pool_ele( pool, fd_forest_deque_pop_head( queue ) );
-    if( FD_LIKELY( blk->confirmed || memcmp( &blk->confirmed_bid, &empty_mr, sizeof( fd_hash_t ) ) != 0 ) ) {
+    if( FD_LIKELY( blk->chain_confirmed || memcmp( &blk->confirmed_bid, &empty_mr, sizeof( fd_hash_t ) ) != 0 ) ) {
       latest_confirmed = blk;
     }
     fd_forest_blk_t * child = fd_forest_pool_ele( pool, blk->child );
@@ -635,7 +635,7 @@ gca( fd_forest_t * forest, fd_forest_blk_t * blk1, fd_forest_blk_t * blk2 ) {
 #define UPDATE_BEST_CANDIDATE( best_confrmd, best_unconfrmd, ele, filter )                         \
   if( FD_UNLIKELY( filter ) ) continue;                                                            \
   do {                                                                                             \
-    if( FD_UNLIKELY( ele->confirmed ) ) {                                                          \
+    if( FD_UNLIKELY( ele->chain_confirmed ) ) {                                                    \
       if( FD_LIKELY( !best_confrmd ) ) best_confrmd = ele;                                         \
       else                             best_confrmd = fd_ptr_if( best_confrmd->slot < ele->slot, ele, best_confrmd ); \
     } else {                                                                                                          \
@@ -798,7 +798,7 @@ evict( fd_forest_t * forest, ulong new_slot, ulong parent_slot ) {
            fork. Force a root in this case. */
     if( fd_forest_orphaned_ele_query( orphaned, &parent_slot, NULL, pool ) ) return ULONG_MAX;
     ulong new_root = fd_forest_pool_ele( pool, forest->root )->child;
-    if( FD_UNLIKELY( !fd_forest_pool_ele( pool, new_root )->confirmed ) ) return ULONG_MAX;
+    if( FD_UNLIKELY( !fd_forest_pool_ele( pool, new_root )->chain_confirmed ) ) return ULONG_MAX;
 
     FD_LOG_WARNING(( "Forest force rooting on slot %lu", fd_forest_pool_ele( pool, new_root )->slot ));
     ulong evicted_slot = fd_forest_pool_ele( pool, forest->root )->slot;
@@ -880,14 +880,14 @@ acquire( fd_forest_t * forest, ulong slot, ulong parent_slot, ulong * evicted ) 
   fd_forest_blk_t * blk  = fd_forest_pool_ele_acquire( pool );
   ulong             null = fd_forest_pool_idx_null( pool );
 
-  blk->slot        = slot;
-  blk->parent_slot = parent_slot;
-  blk->next        = null;
-  blk->parent      = null;
-  blk->child       = null;
-  blk->sibling     = null;
-  blk->confirmed   = 0;
-  blk->consumed    = 0;
+  blk->slot            = slot;
+  blk->parent_slot     = parent_slot;
+  blk->next            = null;
+  blk->parent          = null;
+  blk->child           = null;
+  blk->sibling         = null;
+  blk->chain_confirmed = 0;
+  blk->consumed        = 0;
 
   blk->buffered_idx = UINT_MAX;
   blk->complete_idx = UINT_MAX;
@@ -1185,7 +1185,7 @@ fd_forest_fec_insert( fd_forest_t * forest, ulong slot, ulong parent_slot, uint 
     /* check for a child that is confirmed */
     fd_forest_blk_t * child = fd_forest_pool_ele( fd_forest_pool( forest ), ele->child );
     while( FD_UNLIKELY( child ) ) {
-      if( FD_UNLIKELY( child->confirmed ) ) {
+      if( FD_UNLIKELY( child->chain_confirmed ) ) {
         ele->confirmed_bid       = child->merkle_roots[0].cmr;
         ele->lowest_verified_fec = fec_idx + 1; /* populate the block id with the confirmed child's CMR */
         break;
@@ -1234,7 +1234,7 @@ fd_forest_fec_chain_verify( fd_forest_t * forest, fd_forest_blk_t * ele, fd_hash
   ele->lowest_verified_fec = fec_idx+1;
   ele->confirmed_bid       = *bid; /* confirmed */
 
-  while( FD_UNLIKELY( !ele->confirmed ) ) {
+  while( FD_UNLIKELY( !ele->chain_confirmed ) ) {
     if( FD_UNLIKELY( !fd_hash_eq( expected_mr, &ele->merkle_roots[fec_idx].mr ) ) ) return ele;
 
     /* This FEC merkle is correct, and the chained merkle is correct. */
@@ -1244,7 +1244,7 @@ fd_forest_fec_chain_verify( fd_forest_t * forest, fd_forest_blk_t * ele, fd_hash
     if( FD_UNLIKELY( fec_idx==0 ) ) {
       /* hop to the parent slot, but first we've made it through this
          slot successfully verifying the chain! mark it confirmed! */
-      ele->confirmed = 1;
+      ele->chain_confirmed = 1;
       ele = fd_forest_pool_ele( fd_forest_pool( forest ), ele->parent );
       if( FD_UNLIKELY( !ele || ele->complete_idx == UINT_MAX || ele->buffered_idx != ele->complete_idx ) ) {
         /* can't verify the chain further */
@@ -1272,9 +1272,18 @@ fd_forest_fec_clear( fd_forest_t * forest, ulong slot, uint fec_set_idx, uint ma
   for( uint i=fec_set_idx; i<=fec_set_idx+max_shred_idx; i++ ) {
     fd_forest_blk_idxs_remove( ele->idxs, i );
   }
-  if( FD_UNLIKELY( ele->complete_idx == fec_set_idx+max_shred_idx ) ) ele->complete_idx = UINT_MAX;
-  if( FD_UNLIKELY( fec_set_idx == 0 ) )                               ele->buffered_idx = UINT_MAX;
-  else                                                                ele->buffered_idx = fd_uint_if( ele->buffered_idx != UINT_MAX, fd_uint_min( ele->buffered_idx, fec_set_idx - 1 ), UINT_MAX );
+
+  /* There is a chance that the repair iterator is on this exact slot.
+     This means that this slot is in the requests list, and also at the
+     head of it. If we fec_clear on a range that is less than the
+     iterator's next_shred_idx, then the iterator will pop the slot as
+     "done" (next_shred_idx > complete_idx) without ever rerequesting
+     this fec. We must mark the slot incomplete so that the iterator can
+     re-request everything. */
+  ele->complete_idx = UINT_MAX;
+
+  if( FD_UNLIKELY( fec_set_idx == 0 ) ) ele->buffered_idx = UINT_MAX;
+  else                                  ele->buffered_idx = fd_uint_if( ele->buffered_idx != UINT_MAX, fd_uint_min( ele->buffered_idx, fec_set_idx - 1 ), UINT_MAX );
 
   uint fec_idx = fec_set_idx / 32UL;
   memset( &ele->merkle_roots[fec_idx].mr, 0, sizeof(fd_hash_t) );
@@ -1296,13 +1305,26 @@ fd_forest_fec_clear( fd_forest_t * forest, ulong slot, uint fec_set_idx, uint ma
     }
     if( FD_UNLIKELY( !has_requests_anc ) ) {
       requests_insert( forest, fd_forest_requests( forest ), fd_forest_reqslist( forest ), fd_forest_pool_idx( pool, ele ) );
+
+      /* remove any children than are in the requests list */
+      ulong * queue = fd_forest_deque( forest );
+      fd_forest_deque_push_tail( queue, fd_forest_pool_idx( pool, ele ) );
+      while( FD_LIKELY( fd_forest_deque_cnt( queue ) ) ) {
+        fd_forest_blk_t * child = fd_forest_pool_ele( pool, fd_forest_deque_pop_head( queue ) );
+        if( FD_LIKELY( child != ele ) ) requests_remove( forest, fd_forest_requests( forest ), fd_forest_reqslist( forest ), &forest->iter, fd_forest_pool_idx( pool, child ) );
+        child = fd_forest_pool_ele( pool, child->child );
+        while( FD_LIKELY( child ) ) {
+          fd_forest_deque_push_tail( queue, fd_forest_pool_idx( pool, child ) );
+          child = fd_forest_pool_ele( pool, child->sibling );
+        }
+      }
     }
     /* TODO we could update consumed, but it's not that necessary since
        clearing a fec of a completed slot shouldn't really affect the
        notion of when we completed the slot.  consumed is also updated
        mainly for metrics.  For now we leave it alone. */
   }
-  FD_LOG_INFO(( "fd_forest: fd_forest_fec_clear: cleared fec_set_idx %u to %u, slot %lu", fec_set_idx, fec_set_idx+max_shred_idx, slot ));
+  FD_LOG_INFO(( "fd_forest: fd_forest_fec_clear: cleared slot %lu fec set %u to %u", slot, fec_set_idx, fec_set_idx+max_shred_idx ));
 }
 
 fd_forest_blk_t const *
@@ -1380,9 +1402,9 @@ fd_forest_publish( fd_forest_t * forest, ulong new_root_slot ) {
     fd_forest_pool_ele_release( pool, head );
   }
 
-  new_root_ele->parent    = null; /* unlink new root from parent */
-  new_root_ele->confirmed = 1;
-  forest->root            = fd_forest_pool_idx( pool, new_root_ele );
+  new_root_ele->parent          = null; /* unlink new root from parent */
+  new_root_ele->chain_confirmed = 1;
+  forest->root                  = fd_forest_pool_idx( pool, new_root_ele );
 
   /* 3. New root is in orphaned. This is the case where maybe the
         expected snapshot slot has jumped far ahead.  Invariants tell
@@ -1825,6 +1847,7 @@ fd_forest_print( fd_forest_t const * forest ) {
   fd_forest_frontier_print( forest );
   fd_forest_orphaned_print( forest );
   printf("\n");
+
   fflush(stdout);
 }
 
--- a/src/discof/forest/fd_forest.h
+++ b/src/discof/forest/fd_forest.h
@@ -126,6 +126,7 @@
    to turn on additional runtime checks and logging. */
 
 #include "../../disco/fd_disco_base.h"
+#include "../../disco/shred/fd_fec_set.h"
 
 #ifndef FD_FOREST_USE_HANDHOLDING
 #define FD_FOREST_USE_HANDHOLDING 1
@@ -139,8 +140,6 @@
 #define SET_NAME fd_forest_blk_idxs
 #define SET_MAX  FD_SHRED_BLK_MAX
 #include "../../util/tmpl/fd_set.c"
-
-#define FD_FEC_BLK_MAX (FD_SHRED_BLK_MAX / 32UL) /* 1024 */
 
 /* fd_forest_blk_t implements a left-child, right-sibling n-ary
    tree. Each ele maintains the `pool` index of its left-most child
@@ -169,14 +168,18 @@ struct __attribute__((aligned(128UL))) fd_forest_blk {
     fd_hash_t mr;
     fd_hash_t cmr;
   } merkle_roots[ FD_FEC_BLK_MAX ]; /* */
+
   fd_hash_t confirmed_bid;  /* confirmed block id - can't be wrapped in the above struct because we can create sentinel blocks
-                               on confirmation, and don't know the index of the last fec set until we repair the slot
-                               hash_null if not confirmed. */
+                               on confirmation, and don't know the index of the last fec set until we repair the slot.
+                               hash_null if unknown.  Otherwise populated by the child slot's CMR on confirmation,
+                               or by a confirmation msg from tower.  Has no bearing on if the full slot is correct or not. */
   uint lowest_verified_fec; /* lowest fec index that has been verified so far, inclusive. complete_idx / 32UL,
                                if the last merkle root is verified, 5 if every merkle root after fec set 5*32 is verified */
 
-  uchar confirmed; /* 1 if all the FECs the slot has been confirmed, 0 otherwise */
-  uchar consumed;  /* 1 if the slot has been consumed (i.e., all shreds received, and parents are complete), 0 otherwise */
+  uchar chain_confirmed; /* 1 if all the FECs the slot have been confirmed via fec_chain_verify, 0 otherwise.  Note confirmed_bid
+                            can be populated before this is set to 1. */
+
+  uchar consumed;        /* 1 if the slot has been consumed (i.e., all shreds received, and parents are complete), 0 otherwise */
   /* only consumed slots may be fec_chain_verified */
 
   int est_buffered_tick_recv; /* tick of shred at buffered_idx.  Note since we don't track all the
--- a/src/discof/reasm/fd_reasm.c
+++ b/src/discof/reasm/fd_reasm.c
@@ -11,8 +11,9 @@ fd_reasm_align( void ) {
 
 FD_FN_CONST ulong
 fd_reasm_footprint( ulong fec_max ) {
-  ulong max_fec_per_slot = FD_SHRED_BLK_MAX / 32; /* untrue until fix-32. TODO probably replace with macro after fix-32 is active */
-  int lgf_max = fd_ulong_find_msb( fd_ulong_pow2_up( fec_max + ( fec_max / max_fec_per_slot ) ) ); /* capacity for fec_max fecs + (fec_max / 1024) more block ids */
+  ulong max_fec_per_slot = FD_SHRED_BLK_MAX / 32;                             /* untrue until fix-32. TODO probably replace with macro after fix-32 is active */
+  ulong max_slots        = fd_ulong_max(fec_max / max_fec_per_slot, 1UL);     /* add capacity for a block id per slot */
+  int lgf_max = fd_ulong_find_msb( fd_ulong_pow2_up( fec_max + max_slots ) ); /* capacity for fec_max fecs + (fec_max / 1024) more block ids */
   return FD_LAYOUT_FINI(
     FD_LAYOUT_APPEND(
     FD_LAYOUT_APPEND(
@@ -65,8 +66,9 @@ fd_reasm_new( void * shmem,
 
   fd_memset( shmem, 0, footprint );
 
-  ulong max_fec_per_slot = FD_SHRED_BLK_MAX / 32; /* untrue until fix-32. TODO probably replace with macro after fix-32 is active */
-  int lgf_max = fd_ulong_find_msb( fd_ulong_pow2_up( fec_max + ( fec_max / max_fec_per_slot ) ) ); /* capacity for fec_max fecs + (fec_max / 1024) more block ids */
+  ulong max_fec_per_slot = FD_SHRED_BLK_MAX / 32;                             /* untrue until fix-32. TODO probably replace with macro after fix-32 is active */
+  ulong max_slots        = fd_ulong_max(fec_max / max_fec_per_slot, 1UL);     /* add capacity for a block id per slot */
+  int   lgf_max = fd_ulong_find_msb( fd_ulong_pow2_up( fec_max + max_slots ) ); /* capacity for fec_max fecs + (fec_max / 1024) more block ids */
 
   fd_reasm_t * reasm;
   FD_SCRATCH_ALLOC_INIT( l, shmem );
@@ -169,8 +171,9 @@ fd_reasm_peek( fd_reasm_t * reasm ) {
   for( out_iter_t iter = out_iter_init( reasm->out       );
                         !out_iter_done( reasm->out, iter );
                   iter = out_iter_next( reasm->out, iter ) ) {
-    fd_reasm_fec_t * fec = pool_ele( reasm_pool( reasm ), *out_iter_ele( reasm->out, iter ) );
-    if( FD_LIKELY( !fec->popped && ( !fec->eqvoc || fec->confirmed ) ) ) return fec;
+    fd_hash_t * mr = out_iter_ele( reasm->out, iter );
+    fd_reasm_fec_t * fec = fd_reasm_query( reasm, mr );
+    if( FD_LIKELY( fec && !fec->popped && ( !fec->eqvoc || fec->confirmed ) ) ) return fec;
   }
   return NULL;
 }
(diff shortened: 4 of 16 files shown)


src/discof/repair/fd_repair_tile.c
```diff
@@ -447,6 +448,7 @@ before_frag( ctx_t * ctx,
     return sig!=FD_GOSSIP_UPDATE_TAG_CONTACT_INFO &&
            sig!=FD_GOSSIP_UPDATE_TAG_CONTACT_INFO_REMOVE;
   }
+  if( FD_UNLIKELY( in_kind==IN_KIND_REPLAY ) ) return 1; /* skip for now */
```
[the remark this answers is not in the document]
emwang-jump: ok 🤖

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +318,284 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+                      iter = dlist_iter_fwd_next( iter, subtreel, pool ) ) {
+      fd_reasm_fec_t * ele = dlist_iter_ele( iter, subtreel, pool );
+      UPDATE_BEST_CANDIDATE( confirmed_orphan, unconfrmd_orphan, ele, ele->child != ULONG_MAX || memcmp( &ele->key, parent_root, sizeof(fd_hash_t) ) == 0 );
+    }
+    for( orphaned_iter_t iter = orphaned_iter_init( orphaned, pool );
+                               !orphaned_iter_done( iter, orphaned, pool );
+                         iter = orphaned_iter_next( iter, orphaned, pool ) ) {
+      fd_reasm_fec_t *    ele = orphaned_iter_ele( iter, orphaned, pool );
+      UPDATE_BEST_CANDIDATE( confirmed_orphan, unconfrmd_orphan, ele, ele->child != ULONG_MAX || memcmp( &ele->key, parent_root, sizeof(fd_hash_t) ) == 0 );
+    }
```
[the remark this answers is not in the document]
emwang-jump: bro what

src/discof/reasm/fd_reasm.h
```diff
@@ -190,6 +195,15 @@ struct __attribute__((aligned(128UL))) fd_reasm_fec {
...
 typedef struct fd_reasm_fec fd_reasm_fec_t;
 
+struct evicted {
+  fd_hash_t mr;
+  ulong     slot;
+  uint      fec_set_idx;
+  ulong     bank_idx;
+};
+typedef struct evicted evicted_t;
+typedef struct evicted fd_reasm_evicted_t; /* duplicate typedef for convenience */
```
lidatong: lift this type into `fd_replay_tile.h` and make it a union member of `fd_replay_message` (following that whole one out per tile pattern)

src/app/firedancer/topology.c
```diff
@@ -450,6 +451,8 @@ fd_topo_initialize( config_t * config ) {
 
   fd_topob_wksp( topo, "shred_out"     );
   fd_topob_wksp( topo, "repair_shred"  );
+  fd_topob_wksp( topo, "repair_replay" );
```
lidatong: `repair_out`

src/app/firedancer/topology.c
```diff
@@ -450,6 +451,8 @@ fd_topo_initialize( config_t * config ) {
 
   fd_topob_wksp( topo, "shred_out"     );
   fd_topob_wksp( topo, "repair_shred"  );
```
lidatong: also remove

src/app/firedancer/topology.c
```diff
@@ -450,6 +451,8 @@ fd_topo_initialize( config_t * config ) {
...
+  fd_topob_wksp( topo, "repair_replay" );
+  fd_topob_wksp( topo, "replay_repair" );
```
lidatong: remove this link (see below `replay_out` comment)

src/app/firedancer/topology.c
```diff
@@ -614,6 +617,7 @@ fd_topo_initialize( config_t * config ) {
 
   FOR(quic_tile_cnt)   fd_topob_link( topo, "quic_verify",   "quic_verify",   config->tiles.verify.receive_buffer_size, FD_TPU_REASM_MTU,              config->tiles.quic.txn_reassembly_count );
   FOR(verify_tile_cnt) fd_topob_link( topo, "verify_dedup",  "verify_dedup",  config->tiles.verify.receive_buffer_size, FD_TPU_PARSED_MTU,             1UL );
+  /**/                 fd_topob_link( topo, "replay_repair", "replay_repair", 1024,                                     sizeof(fd_reasm_evicted_t),   1UL );
```
lidatong: remove

src/app/firedancer/topology.c
```diff
@@ -648,7 +652,7 @@ fd_topo_initialize( config_t * config ) {
   /**/                 fd_topob_link( topo, "sign_txsend",   "sign_txsend",   128UL,                                    sizeof(fd_ed25519_sig_t)*2UL,  1UL ); /* TODO: Depth probably doesn't need to be 128 */
 
   FOR(shred_tile_cnt)  fd_topob_link( topo, "shred_out",     "shred_out",     shred_depth,                              FD_SHRED_OUT_MTU,              3UL ); /* TODO: Pretty sure burst of 3 is incorrect here */
-  FOR(shred_tile_cnt)  fd_topob_link( topo, "repair_shred",  "repair_shred",  shred_depth,                              sizeof(fd_ed25519_sig_t),      1UL );
+  /**/                 fd_topob_link( topo, "repair_replay", "repair_replay", 32768UL,                                  FD_SHRED_OUT_MTU,              1UL );
```
lidatong: remove

src/app/firedancer/topology.c
```diff
@@ -917,17 +921,19 @@ fd_topo_initialize( config_t * config ) {
 
   /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "genesi_out",    0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "gossip_out",    0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
+  /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "replay_repair", 0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
```
lidatong: remove

src/app/firedancer/topology.c
```diff
@@ -917,17 +921,19 @@ fd_topo_initialize( config_t * config ) {
...
   FOR(shred_tile_cnt)  fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "shred_out",     i,            FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   if( snapshots_enabled ) {
                        fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "snapin_manif",  0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   }
   /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "tower_out",     0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
-  FOR(shred_tile_cnt)  fd_topob_tile_out(   topo, "repair",  0UL,                       "repair_shred",  i                                                  );
+  /**/                 fd_topob_tile_out(   topo, "repair",  0UL,                       "repair_replay", 0                                                  );
```
lidatong: repair_out

src/app/firedancer/topology.c
```diff
@@ -917,17 +921,19 @@ fd_topo_initialize( config_t * config ) {
...
 
-  FOR(shred_tile_cnt)  fd_topob_tile_in(    topo, "replay",  0UL,          "metric_in", "shred_out",     i,            FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
+  /**/                 fd_topob_tile_in (   topo, "replay",  0UL,          "metric_in", "repair_replay", 0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
```
lidatong: repair_out

src/app/firedancer/topology.c
```diff
@@ -917,17 +921,19 @@ fd_topo_initialize( config_t * config ) {
...
   /**/                 fd_topob_tile_in (   topo, "replay",  0UL,          "metric_in", "genesi_out",    0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   /**/                 fd_topob_tile_out(   topo, "replay",  0UL,                       "replay_out",    0UL                                                );
   /**/                 fd_topob_tile_out(   topo, "replay",  0UL,                       "replay_epoch",  0UL                                                );
+  /**/                 fd_topob_tile_out(   topo, "replay",  0UL,                       "replay_repair", 0UL                                                );
```
lidatong: remove

src/app/firedancer/topology.c
```diff
@@ -1348,16 +1353,29 @@ fd_topo_initialize( config_t * config ) {
...
-      reduced to 1024 once FECs are restricted to always be size 32,
-      given the current consensus limit of 32768 shreds per block). */
-
-  if( FD_UNLIKELY( !fd_ulong_is_pow2( config->firedancer.runtime.max_live_slots ) ) ) FD_LOG_ERR(( "max_live_slots must be a power of 2 for store" ));
-  fd_topo_obj_t * store_obj = setup_topo_store( topo, "store", config->firedancer.runtime.max_live_slots * 1024UL /* FIXME temporary hack to run on 512 gb boxes */, (uint)shred_tile_cnt );
+      be retained by store.
+
+      The base value is from multiplying max_live_slots by the maximum
+      number of FEC sets in a block (which is 1024, given the current
+      consensus limit of 32768 shreds per block). This is notably, the
```
lidatong:
```suggestion
      consensus limit of 32768 shreds per block).  This is notably, the
```

src/app/firedancer/topology.c
```diff
@@ -1348,16 +1353,29 @@ fd_topo_initialize( config_t * config ) {
...
+
+      Store needs to hold depth(shred_out) + depth(repair_replay) more
+      FEC sets than reasm. Any FEC inserted to store will also be
+      inserted to reasm,  but store can be up to depth(shred_out) +
+      depth(repair_replay) ahead of reasm. If store is exactly this
+      total depth ahead of reasm, then this means the links must be
+      full and shred is backpressured. No more FEC sets will be inserted
+      to store until reasm inserts the next FEC, at which point it will
+      be over capacity and will select a FEC to evict from itself and
+      from store. */
```
lidatong:
```suggestion
      Store needs to hold _at least_ depth(shred_out) + depth(repair_out) more
      FEC sets than reasm.  Shred inserts FECs into store (before publishing to shred_out) and Replay
      inserts FECs into reasm (after consuming from repair_out), so store can be up to depth(shred_out) +
      depth(repair_out) FECs ahead of reasm.  If store is exactly this
      total depth ahead of reasm, then this means the links must be
      full and shred is backpressured via repair_out -> shred_out.  Thus, no more FEC sets will be inserted
      to store until reasm inserts the next FEC, at which point it must be at capacity (given store is _at least_  total depth larger) and will evict a FEC set which is evicted together from store. */
```

src/app/firedancer/topology.c
```diff
@@ -1348,16 +1353,29 @@ fd_topo_initialize( config_t * config ) {
...
+
+  if( FD_UNLIKELY( !fd_ulong_is_pow2( config->firedancer.runtime.max_live_slots ) ) ) FD_LOG_ERR(( "max_live_slots must be a power of 2" ));
```
lidatong: think we should either check this one earlier in this functino

src/app/firedancer/topology.c
```diff
@@ -1348,16 +1353,29 @@ fd_topo_initialize( config_t * config ) {
...
+  fd_topo_link_t * repair_replay_link = &topo->links[ fd_topo_find_link( topo, "repair_replay", 0UL ) ];
```
lidatong: repair_out

src/app/firedancer/topology.c
```diff
@@ -1348,16 +1353,29 @@ fd_topo_initialize( config_t * config ) {
...
+  ulong store_fec_max = config->firedancer.runtime.max_live_slots * 1024 + (shred_depth * shred_tile_cnt) + repair_replay_link->depth + 1;
+        store_fec_max = fd_ulong_pow2_up( store_fec_max );
```
lidatong:
```suggestion
  ulong store_fec_max = fd_ulong_pow2_up( config->firedancer.runtime.max_live_slots * 1024 + (shred_depth * shred_tile_cnt) + repair_replay_link->depth + 1 );
```

src/discof/forest/fd_forest.c
```diff
@@ -1074,15 +1074,17 @@ fd_forest_data_shred_insert( fd_forest_t * forest,
                              int           slot_complete,
                              int           ref_tick,
                              int           src,
+                             int           is_dup, /* TODO document / rename */
```
lidatong: eqvoc?

src/discof/forest/fd_forest.h
```diff
@@ -742,6 +743,7 @@ fd_forest_data_shred_insert( fd_forest_t * forest,
                              int           slot_complete,
                              int           ref_tick,
                              int           src,
+                             int           is_dup,
```
lidatong:
```suggestion
                             int           eqvoc,
```

src/discof/forest/fd_forest.h
```diff
@@ -169,6 +167,9 @@ struct __attribute__((aligned(128UL))) fd_forest_blk {
     fd_hash_t mr;
     fd_hash_t cmr;
   } merkle_roots[ FD_FEC_BLK_MAX ]; /* */
+
+   /* The following fields are only valid if the slot is confirmed.  They
+      are used for chain verification and tracking the confirmed block id. */
   fd_hash_t confirmed_bid;  /* confirmed block id - can't be wrapped in the above struct because we can create sentinel blocks
                                on confirmation, and don't know the index of the last fec set until we repair the slot
                                hash_null if not confirmed. */
```
lidatong:
```suggestion
   /* The following fields are only valid if confirmed = 1.  They
      are used for chain verification and tracking the confirmed block id. */
      
fd_hash_t confirmed_bid;  /* confirmed block id - can't be wrapped in the above struct because we can create sentinel blocks
                              on confirmation, and don't know the index of the last fec set until we repair the slot
                              hash_null if not confirmed. */
```

src/discof/forest/fd_forest.h
lidatong: move up

src/ballet/shred/fd_shred.h
```diff
@@ -150,6 +150,10 @@ FD_STATIC_ASSERT( sizeof(fd_bmtree_node_t) == FD_SHRED_MERKLE_ROOT_SZ, update FD
 
 /* Maximum number of data shreds in a slot, also maximum number of parity shreds in a slot */
 #define FD_SHRED_BLK_MAX (1 << 15UL) /* 32,768 shreds */
+#define FD_SHRED_IDX_MAX (FD_SHRED_BLK_MAX - 1)
+
+/* Maximum number of FEC sets in a slot / 32 shreds per FEC */
+#define FD_FEC_BLK_MAX (FD_SHRED_BLK_MAX / 32)
```
lidatong: would move into `fd_fec_set.h`

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+static int
+evict( fd_reasm_t      * reasm,
+       fd_store_t      * opt_store,
+       fd_hash_t const * new_root FD_PARAM_UNUSED,
+       fd_hash_t const * parent_root ) {
+  fd_reasm_fec_t * pool = reasm_pool( reasm );
+  frontier_t * frontier = reasm->frontier;
+  orphaned_t * orphaned = reasm->orphaned;
+  subtrees_t * subtrees = reasm->subtrees;
+  dlist_t *    subtreel = reasm->subtreel;
```
lidatong: align variable names

src/discof/reasm/fd_reasm.h
```diff
@@ -311,7 +332,42 @@ fd_reasm_insert( fd_reasm_t *      reasm,
...
+                     fd_store_t     * opt_store );
+
+fd_reasm_evicted_t *
+fd_reasm_evicted_peek_head( fd_reasm_t * reasm );
+
+int
+fd_reasm_evicted_empty( fd_reasm_t * reasm );
+
+fd_reasm_evicted_t
+fd_reasm_evicted_pop_head( fd_reasm_t * reasm );
```
lidatong: instead of a 2-part caller-driven API based around an evicted queue, i think you can instead do it all internally as part of insert and take an out pointer fd_reasm_t ** evicted that would be point to the first FEC of the evicted bank (which is disconnected from the parent bank node)
then, the caller can just walk down .child, which still is valid even if its no longer in the tree and you know there's only one child given you stop at the first bank property
this way you don't need to introduce the evicted queue and the corresponding new APIs that is driven by replay tile

src/discof/replay/fd_replay_tile.c
```diff
@@ -489,7 +481,7 @@ scratch_footprint( fd_topo_tile_t const * tile ) {
   l = FD_LAYOUT_APPEND( l, alignof(fd_block_id_ele_t),   sizeof(fd_block_id_ele_t) * tile->replay.max_live_slots );
   l = FD_LAYOUT_APPEND( l, fd_block_id_map_align(),      fd_block_id_map_footprint( chain_cnt ) );
   l = FD_LAYOUT_APPEND( l, fd_txncache_align(),          fd_txncache_footprint( tile->replay.max_live_slots ) );
-  l = FD_LAYOUT_APPEND( l, fd_reasm_align(),             fd_reasm_footprint( tile->replay.fec_max ) );
+  l = FD_LAYOUT_APPEND( l, fd_reasm_align(),             fd_reasm_footprint( 512 ) );
```
lidatong: don't forget to remove

src/discof/reasm/fd_reasm.h
```diff
@@ -311,7 +332,42 @@ fd_reasm_insert( fd_reasm_t *      reasm,
...
+   two cases handle any equivocation that may have occurred.
+
+   The evicted FEC metadata will be pushed to the evicted queue.  It is
+   the caller's responsibility to clear the evicted queue before the
+   next fd_reasm_insert or fd_reasm_clear_leaf call. */
+
+int
+fd_reasm_clear_leaf( fd_reasm_t     * reasm,
+                     fd_reasm_fec_t * head,
+                     int              clear_chain,
```
lidatong: don't need granularity of `clear_chain`

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+    evicted_push_tail( reasm->evicted, (fd_reasm_evicted_t){ .mr = head->key, .slot = head->slot, .fec_set_idx = head->fec_set_idx, .bank_idx = head->bank_idx } );
+    pool_ele_release( pool, clear_slot_metadata( reasm, head, opt_store ) );
+  }
+
+  /* Remove elements from out queue if they are free */
+  ulong cnt = out_cnt( reasm->out );
+  for( ulong i = 0UL; i < cnt; i++ ) {
+    ulong idx = out_pop_head( reasm->out );
+    if( FD_LIKELY( pool_ele( pool, idx )->free==0 ) ) out_push_tail( reasm->out, idx );
+  }
```
lidatong: this should still be done lazily (and probably requires changing the out queue to contain the actual block ids in the map)

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+                   └──  64' (b=1) ── 96' (b=1) ── 128'(b=1)         (eqvoc)
+
+      Now we know we should evict until the max(parent bank idx changes, or parent has > 1 child, or fec set idx == 0) */
+
+    while( FD_LIKELY( head ) ) {
+      fd_reasm_fec_t * parent = fd_reasm_parent( reasm, head );
+      if( FD_UNLIKELY( head->fec_set_idx==0 ) )                    break;
+      if( FD_UNLIKELY( head->sibling != pool_idx_null( pool ) ) )  break; /* if the parent has more than 1 child, we know for sure the parent is a slot boundary, so we can stop here. */
+      /* should be impossible for parent to be null, assert or verify */
+      if( FD_UNLIKELY( parent->bank_idx != head->bank_idx ) )      break; /* also possible none of these have been executed */
```
lidatong: what case is here that hasn't been caught by the one above?

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+      head = parent;
```
lidatong, line 395: btw make sure to think through this still works if a block gets more FECs after `.slot_complete` (if you haven't already thought through)

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+      fd_reasm_fec_t * prev    = child;
+      while( FD_LIKELY( sibling ) ) {
+        if( FD_LIKELY( 0==memcmp( &sibling->key, &head->key, sizeof(fd_hash_t) ) ) ) {
+          prev->sibling = sibling->sibling;
+          break;
+        }
+        prev = sibling;
+        sibling = pool_ele( pool, sibling->sibling );
+      }
+    }
```
lidatong, line 418:
```suggestion
   /* Clean up the parent pointing to this head, and remove block from the maps
      remove the block from the parent's child list */

    fd_reasm_fec_t * child = pool_ele( pool, parent->child );
    if( FD_LIKELY( 0==memcmp( &child->key, &head->key, sizeof(fd_hash_t) ) ) ) { /* evicted is left-child (including only child) */
      parent->child = child->sibling;
    } else { /* evicted is right-sibling(s) */
      fd_reasm_fec_t * sibling = child;
      fd_reasm_fec_t * prev    = NULL;
      while( FD_LIKELY( sibling && 0==memcmp( &sibling->key, &head->key, sizeof(fd_hash_t) ) ) ) { /* go through the sibling list, and remove the block */
        prev = sibling;
        sibling = pool_ele( pool, sibling->sibling );
      }
      FD_TEST( prev ); /* invariant violation: reasm tree bug not linked correctly */
      prev->sibling = sibling->sibling;
    }
```

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+  fec->bank_idx = ULONG_MAX;
+  return fec;
+}
+
+/* returns 0 if an orphan or subtree was cleared, 1 if something in main
+   tree was cleared */
+int
+fd_reasm_clear_leaf( fd_reasm_t     * reasm,
+                     fd_reasm_fec_t * head,
+                     int              clear_chain,
```
lidatong: it seems like `clear_chain` is passed in by caller and only used to determine whether it needs to search up ancestry in the `if` statement below? it seems to figure out itself anyway whether its in orphaned or connected (as with the FD_TESTs)

if that's accurate, remove this parameter. it makes it harder to follow the code

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+        prev = sibling;
+        sibling = pool_ele( pool, sibling->sibling );
+      }
+    }
+
+    /* remove the chain itself from the maps */
+
+    fd_reasm_fec_t * removed = orphaned_ele_remove( orphaned, &head->key, NULL, pool );
+    if( FD_LIKELY  ( removed ) ) {
+      FD_TEST( !clear_chain ); /* if we selected to clear an orphan leaf, clear_chain must be false */
```
lidatong: see above comment about clear_chain, since it's re-determining this information anyway the caller shouldnt pass it in

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+      evicted_push_tail( reasm->evicted, (fd_reasm_evicted_t){ .mr = head->key, .slot = head->slot, .fec_set_idx = head->fec_set_idx, .bank_idx = head->bank_idx } );
+      pool_ele_release( pool, clear_slot_metadata( reasm, head, opt_store ) );
+    }
+
+    if( FD_UNLIKELY( !removed ) ) {
```
lidatong:
```suggestion
    if( FD_UNLIKELY( !removed_orphan ) ) {
```

src/discof/replay/fd_replay_tile.c
```diff
@@ -2201,6 +2170,19 @@ after_credit( fd_replay_tile_t *  ctx,
...
+
+  if( FD_UNLIKELY( !fd_reasm_evicted_empty( ctx->reasm ) ) ) {
+    fd_reasm_evicted_t evicted_fec = fd_reasm_evicted_pop_head( ctx->reasm );
+    fd_memcpy( fd_chunk_to_laddr( ctx->repair_out->mem, ctx->repair_out->chunk ), &evicted_fec, sizeof(fd_reasm_evicted_t) );
+    fd_stem_publish( stem, ctx->repair_out->idx, 0, ctx->repair_out->chunk,  sizeof(fd_reasm_evicted_t), 0UL, 0UL, fd_frag_meta_ts_comp( fd_tickcount() ) );
+    ctx->repair_out->chunk = fd_dcache_compact_next( ctx->repair_out->chunk, sizeof(fd_reasm_evicted_t), ctx->repair_out->chunk0, ctx->repair_out->wmark );
+    *charge_busy = 1;
+    *opt_poll_in = 0;
+    return;
+  }
```
lidatong: this should be queued up in a tile-specific deque (eg. tower tile's `publishes`) bc draining publishes one frag at a time is a replay tile not reasm concern

src/discof/reasm/fd_reasm.h
```diff
@@ -311,7 +332,42 @@ fd_reasm_insert( fd_reasm_t *      reasm,
...
+   we reach fec_set_idx 0, we reach a fec set with an equivocating
+   sibling, or we reach a fec set with a different bank_idx.  The last
+   two cases handle any equivocation that may have occurred.
+
+   The evicted FEC metadata will be pushed to the evicted queue.  It is
+   the caller's responsibility to clear the evicted queue before the
+   next fd_reasm_insert or fd_reasm_clear_leaf call. */
+
+int
+fd_reasm_clear_leaf( fd_reasm_t     * reasm,
```
lidatong:
```suggestion
fd_reasm_remove( fd_reasm_t     * reasm,
```

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+
+      The state after executing the wrong version first is:
+
+            0  ────  32 ────  64 ────  96 ──── 128
+      (bank_idx=1) (bank_idx=1) .... (all bank_idx=1)
+
+      After receiving and executing the confirmed version, the state looks like
+
+      0 (b=2) ── 32 (b=2) ──  64 (b=2) ──  96 (b=2)        (confirmed)
+                 └──  64' (b=1) ── 96' (b=1) ── 128' (b=1) (eqvoc)
```
lidatong: nice docs and diagrams. could you align these equivocating siblings in the diagram similar to above?

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+
+      Here we want to evict only until fec 64'. Or let's say we are
+      getting around to executing the confirmed version, but we haven't
+      executed it yet.
+
+      0 (b=1) ── 32 (b=1) ──  64 (b=ULONG_MAX) ──  96 (b=ULONG_MAX) (confirmed, but not executed yet)
+                   └──  64' (b=1) ── 96' (b=1) ── 128'(b=1)         (eqvoc)
```
lidatong: align sibs

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+      pool_ele_release( pool, clear_slot_metadata( reasm, head, opt_store ) );
+    }
+
+    if( FD_UNLIKELY( !removed ) ) {
+      /* remove from ancestry and frontier */
+      FD_TEST( clear_chain );
+      fd_reasm_fec_t * curr = head;
+      while( FD_LIKELY( curr ) ) {
+                       removed = ancestry_ele_remove( ancestry, &curr->key, NULL, pool );
+        if( !removed ) removed = frontier_ele_remove( frontier, &curr->key, NULL, pool );
```
lidatong:
```suggestion
        removed = ancestry_ele_remove( ancestry, &curr->key, NULL, pool );
        if( !removed ) removed = frontier_ele_remove( frontier, &curr->key, NULL, pool );
```

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
   return xid;
 }
 
+static fd_reasm_fec_t *
+clear_slot_metadata( fd_reasm_t     * reasm,
+                     fd_reasm_fec_t * fec,
+                     fd_store_t     * opt_store ) {
+  fec->free = 1;
```
lidatong: remove this free bit after addressing comment about just putting the merkle root in out

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+  if( FD_UNLIKELY( fec->slot_complete ) ) {
+    xid_t * bid = xid_query( reasm->xid, (fec->slot << 32)|UINT_MAX, NULL );
+    bid->cnt--;
+    if( FD_LIKELY( !bid->cnt ) ) xid_remove( reasm->xid, bid );
+  }
+  xid_t * xid = xid_query( reasm->xid, ( fec->slot << 32 ) | fec->fec_set_idx, NULL );
+  xid->cnt--;
+  if( FD_LIKELY( !xid->cnt ) ) xid_remove( reasm->xid, xid );
+
+  if( FD_LIKELY( opt_store ) ) fd_store_remove( opt_store, &fec->key );
```
lidatong: do this outside this function

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+  fec->bank_idx = ULONG_MAX;
```
lidatong: should be set / zerod on insert

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+      }
+
+      /* We removed from the main tree, so we might need to insert parent into the frontier.
+          Only need to add parent to the frontier if it doesn't have any other children. */
+
+      if( parent->child == pool_idx_null( pool ) ) {
+        parent = ancestry_ele_remove( ancestry, &parent->key, NULL, pool );
+        FD_TEST( parent );
+        frontier_ele_insert( frontier, parent, pool );
+      }
```
lidatong: you probably already have determined this information earlier and can do it there?

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+  orphaned_t *        orphaned = reasm->orphaned;
+  frontier_t *        frontier = reasm->frontier;
+  ancestry_t *        ancestry = reasm->ancestry;
+  subtrees_t *        subtrees = reasm->subtrees;
+  dlist_t    *        subtreel = reasm->subtreel;
+  int                 is_ancestry = FD_REASM_EVICT_ORPHAN;
+
+  FD_TEST( head && head->child == ULONG_MAX ); /* must be a leaf node */
+  fd_reasm_fec_t * prehead = head;
+  if( FD_UNLIKELY( clear_chain ) ) {
```
lidatong: overall i think this control flow can be simplified. it seems like it should just be removing from all the maps and not special-casing whether it's an orphan or not? (ie. you can just remove from the map, and if the remove doesn't return anything you know it's not an orphan)

so

```c
ele = remove_orphan...
 if( ele ) subtrees_remove....
```

also see above comment about 2-part API -- should all just be happening together in `reasm_insert`, and returning the head pool ele to bank

src/discof/reasm/fd_reasm.h
```diff
@@ -52,6 +52,11 @@
 #define FD_REASM_ERR_UNIQUE (-1) /* key uniqueness conflict */
 #define FD_REASM_ERR_MERKLE (-2) /* chained merkle root conflict */
 
+#define FD_REASM_EVICT_FAIL     (-1) /* eviction failed, caller should not insert new FEC */
+#define FD_REASM_EVICT_ORPHAN   ( 0) /* eviction succeeded, caller can insert new FEC */
+#define FD_REASM_EVICT_ANCESTOR ( 1) /* eviction succeeded, caller can insert new FEC, but replay should evict corresponding bank  */
+#define FD_REASM_EVICT_UNNEEDED ( 2) /* no eviction required to insert FEC */
```
lidatong: remove these with new API i'm suggesting

src/discof/reasm/fd_reasm.c
lidatong: also is `bank_idx` guaranteed to be set to a valid value during an evict? sched can not be ready and hold reasm back right while replay_tile continues to insert?

src/discof/repair/fd_repair_tile.c
```diff
@@ -1169,14 +1123,14 @@ unprivileged_init( fd_topo_t *      topo,
       ctx->net_out_wmark  = fd_dcache_compact_wmark( ctx->net_out_mem, link->dcache, link->mtu );
       ctx->net_out_chunk  = ctx->net_out_chunk0;
 
-    } else if( 0==strcmp( link->name, "repair_shred" ) ) {
+    } else if( 0==strcmp( link->name, "repair_replay" ) ) {
```
lidatong: repair_out

src/discof/repair/fd_repair_tile.c
```diff
@@ -1169,14 +1123,14 @@ unprivileged_init( fd_topo_t *      topo,
...
 
-      out_ctx_t * shred_out = &ctx->shred_out_ctx[ ctx->shred_tile_cnt++ ];
-      shred_out->idx        = out_idx;
-      shred_out->mem        = topo->workspaces[ topo->objs[ link->dcache_obj_id ].wksp_id ].wksp;
-      shred_out->chunk0     = fd_dcache_compact_chunk0( shred_out->mem, link->dcache );
-      shred_out->wmark      = fd_dcache_compact_wmark( shred_out->mem, link->dcache, link->mtu );
-      shred_out->chunk      = shred_out->chunk0;
+      out_ctx_t * replay_out = ctx->replay_out_ctx;
```
lidatong: repair_out

src/discof/replay/fd_replay_tile.c
```diff
@@ -2037,35 +2002,39 @@ process_fec_set( fd_replay_tile_t *  ctx,
 
   for( fd_reasm_fec_t * curr = reasm_fec;; ) {
     curr = fd_reasm_parent( ctx->reasm, curr );
-    if( FD_UNLIKELY( !curr ) ) return; /* If can't connect, drop the FEC. */
+    if( FD_UNLIKELY( !curr ) ) {
+      FD_LOG_WARNING(( "can't connect FEC, dropping it" ));
+      return; /* If can't connect, drop the FEC. */
+    }
```
lidatong: can this ever happen validly given current eviction? reasm evictions always result in a synchronized bank eviction (and vice versa)?

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+  if( FD_UNLIKELY( filter ) ) continue;                                                                                       \
+  do {                                                                                                                        \
+    if( FD_UNLIKELY( ele->confirmed ) ) {                                                                                     \
+      if( FD_LIKELY( !best_confrmd ) ) best_confrmd = ele;                                                                    \
+      else                             best_confrmd = fd_ptr_if( best_confrmd->slot < ele->slot, ele, best_confrmd );         \
+    } else {                                                                                                                  \
+      if( FD_LIKELY( !best_unconfrmd ) ) best_unconfrmd = ele;                                                                \
+      else                               best_unconfrmd = fd_ptr_if( best_unconfrmd->slot < ele->slot, ele, best_unconfrmd ); \
+    }                                                                                                                         \
+  } while(0)
```
lidatong, line 508: this looks like `.max_by` with a filter?

i would just inline it since it's just used in two places instead of making a closure macro

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+      return fd_reasm_clear_leaf( reasm, unconfrmd_orphan, 0, opt_store );
+    }
+    if( FD_UNLIKELY( unconfrmd_leaf )) {
+      return fd_reasm_clear_leaf( reasm, unconfrmd_leaf, 1, opt_store );
+    }
+    if( FD_UNLIKELY( confirmed_orphan )) {
+      fd_reasm_fec_t * parent = fd_reasm_query( reasm, parent_root );
+      if( !parent ) {
+        return fd_reasm_clear_leaf( reasm, confirmed_orphan, 0, opt_store );
+      }
```
lidatong: i think this will be easier to follow by restructuring to order the checks in priority and doing the `clear_leaf` stuff with each check, then return early, vs. saving variables and re-checking the order

src/discof/reasm/fd_reasm.c
```diff
@@ -310,6 +317,304 @@ xid_update( fd_reasm_t * reasm, ulong slot, uint fec_set_idx, ulong pool_idx ) {
...
+        subtree_root = pool_idx( pool, root );
+      }
+
+      fd_reasm_fec_t * latest_confirmed_leaf = latest_confirmed_fec( reasm, subtree_root );
+      if( !latest_confirmed_leaf || latest_confirmed_leaf == gca( reasm, latest_confirmed_leaf, parent )) {
+        return fd_reasm_clear_leaf( reasm, confirmed_orphan, 0, opt_store );
+      }
+      /* is a useless new fork. */
+      return FD_REASM_EVICT_FAIL;
+    }
```
lidatong: hm i'm not following why this is needed. don't we already know 7 is not confirmed and therefore REASM_CANT_EVICT?

src/discof/repair/fd_repair_tile.c
```diff
@@ -883,12 +886,13 @@ after_frag( ctx_t *             ctx,
     fd_hash_t  * mr    = (fd_hash_t *)(ctx->buffer + fd_shred_header_sz( shred->variant ));
     fd_hash_t  * cmr   = (fd_hash_t *)(ctx->buffer + fd_shred_header_sz( shred->variant ) + sizeof(fd_hash_t) );
     uint         nonce = FD_LOAD(uint, ctx->buffer + fd_shred_header_sz( shred->variant ) + sizeof(fd_hash_t) + sizeof(fd_hash_t) ); /* gibberish if not shred msg */
+    int          is_dup = FD_LOAD(int, ctx->buffer + fd_shred_header_sz( shred->variant ) + sizeof(fd_hash_t) + sizeof(fd_hash_t) + sizeof(uint) ); /* gibberish if not shred msg */
+                 is_dup = is_dup == FD_FEC_RESOLVER_SHRED_DUPLICATE;
```
lidatong:
```suggestion
    int          rv = FD_LOAD(int, ctx->buffer + fd_shred_header_sz( shred->variant ) + sizeof(fd_hash_t) + sizeof(fd_hash_t) + sizeof(uint) ); /* gibberish if not shred msg */
int  is_dup = rv == FD_FEC_RESOLVER_SHRED_DUPLICATE;
```

src/discof/replay/fd_replay_tile.c
```diff
@@ -2226,6 +2208,10 @@ after_credit( fd_replay_tile_t *  ctx,
       if( FD_UNLIKELY( ctx->is_leader && frontier_indices[i]==ctx->leader_bank->data->idx ) ) continue;
       mark_bank_dead( ctx, stem, bank->data->idx );
       fd_sched_block_abandon( ctx->sched, bank->data->idx );
+
+      /* evict it from reasm */
+      fd_block_id_ele_t * block_id_ele = &ctx->block_id_arr[ bank->data->idx ];
+      FD_TEST( fd_reasm_clear_leaf( ctx->reasm, fd_reasm_query( ctx->reasm, &block_id_ele->latest_mr ), 1, ctx->store ) );
```
lidatong: reasm_remove

src/discof/forest/fd_forest.c
emwang-jump: No this one is specifically FEC_RESOLVER_DUPLICATE return code 😢  unrelated to equivocation

src/app/firedancer/topology.c
emwang-jump: rnonce lives in this wksp

src/app/firedancer/topology.c
emwang-jump: turns out it can be non pow2, so removing it

src/discof/forest/fd_forest.h
emwang-jump: I think this whole comment is incorrect, so just deleting it and clarifying further in the variable comments

src/discof/reasm/fd_reasm.c
emwang-jump: changed to match the style of the above w/ adjustments

src/discof/reasm/fd_reasm.c
emwang-jump: saur true, removing, realized we can be mid replay of this slot so the bank_idxs would be !=

src/discof/reasm/fd_reasm.c
emwang-jump: for simplicity we can evict until the first slot complete

src/discof/reasm/fd_reasm.c
emwang-jump: It's that we aren't sure if 7 is confirmed or not, so anything that continues along the confirmed chain we know of has potential to be confirmed, so we should add it

src/discof/reasm/fd_reasm.c
emwang-jump: i agree, but then we would need to do repeated traversal of the tree that looks pretty similar to each other. I did move the frontier iteration after the first check though

src/discof/reasm/fd_reasm.c
emwang-jump: This is ancestry specific so I just grouped it with the rest of ancestry/frontier logic, whereas the parent unchaining logic doesn't care about which map the node belongs to

src/discof/replay/fd_replay_tile.c
```diff
@@ -2185,6 +2212,10 @@ after_credit( fd_replay_tile_t *  ctx,
       if( FD_UNLIKELY( ctx->is_leader && frontier_indices[i]==ctx->leader_bank->data->idx ) ) continue;
       mark_bank_dead( ctx, stem, bank->data->idx );
       fd_sched_block_abandon( ctx->sched, bank->data->idx );
+
+      /* evict it from reasm */
+      fd_block_id_ele_t * block_id_ele = &ctx->block_id_arr[ bank->data->idx ];
+      FD_TEST( fd_reasm_remove( ctx->reasm, fd_reasm_query( ctx->reasm, &block_id_ele->latest_mr ), ctx->store ) );
```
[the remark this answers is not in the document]
emwang-jump: nice

src/discof/replay/fd_replay_tile.c
emwang-jump: Not sure, this is an artifact of bank eviction before reasm eviction got solidified. will leave a TODO to investigate/clean up but I don't think this'll cause problems

src/app/firedancer/topology.c
```diff
@@ -648,7 +650,7 @@ fd_topo_initialize( config_t * config ) {
   /**/                 fd_topob_link( topo, "sign_txsend",   "sign_txsend",   128UL,                                    sizeof(fd_ed25519_sig_t)*2UL,  1UL ); /* TODO: Depth probably doesn't need to be 128 */
 
   FOR(shred_tile_cnt)  fd_topob_link( topo, "shred_out",     "shred_out",     shred_depth,                              FD_SHRED_OUT_MTU,              3UL ); /* TODO: Pretty sure burst of 3 is incorrect here */
-  FOR(shred_tile_cnt)  fd_topob_link( topo, "repair_shred",  "repair_shred",  shred_depth,                              sizeof(fd_ed25519_sig_t),      1UL );
+  /**/                 fd_topob_link( topo, "repair_out",    "repair_out",    32768UL,                                  FD_SHRED_OUT_MTU,              1UL );
```
lidatong: where does 32768 come from?

src/app/firedancer/topology.c
```diff
@@ -1348,16 +1350,29 @@ fd_topo_initialize( config_t * config ) {
...
+      up to depth(shred_out) + depth(repair_out) ahead of reasm.  If
+      store is exactly this total depth ahead of reasm, then this means
+      the links must be full and shred is backpressured via repair_out
+      -> shred_out.  Thus, no more FEC sets will be inserted to store
+      until reasm inserts the next FEC, at which point it must be at
+      capacity (given store is _at least_ total depth larger) and will
+      evict a FEC which is together evicted from store. */
+
+  fd_topo_link_t * repair_out_link = &topo->links[ fd_topo_find_link( topo, "repair_out", 0UL ) ];
+  ulong store_fec_max = fd_ulong_pow2_up( config->firedancer.runtime.max_live_slots * 1024 + (shred_depth * shred_tile_cnt) + repair_out_link->depth + 1 );
```
lidatong: what's `+1` for here, burst? think it all ends up rounding to next pow2 anyway

lidatong:
```suggestion
  ulong store_fec_max = fd_ulong_pow2_up( config->firedancer.runtime.max_live_slots * FD_FEC_BLK_MAX + (shred_depth * shred_tile_cnt) + repair_out_link->depth + 1 );
```

src/app/firedancer/topology.c
lidatong: ah ok rename to rnonce?

src/discof/forest/fd_forest.c
```diff
@@ -212,7 +212,7 @@ fd_forest_init( fd_forest_t * forest, ulong root_slot ) {
   root_ele->sibling          = null;
   root_ele->buffered_idx     = 0;
   root_ele->complete_idx     = 0;
-  root_ele->confirmed        = 1;
+  root_ele->chain_confirmed  = 1;
```
lidatong, line 215: hm, probably don't do this change in this PR, but thinking about it chained and confirmed seem like orthogonal concepts where we'd want to potentially track each individually and both set to `1` would be the confirmed chain

src/discof/replay/fd_replay_tile.c
```diff
@@ -486,7 +490,7 @@ scratch_footprint( fd_topo_tile_t const * tile ) {
   l = FD_LAYOUT_APPEND( l, alignof(fd_block_id_ele_t),   sizeof(fd_block_id_ele_t) * tile->replay.max_live_slots );
   l = FD_LAYOUT_APPEND( l, fd_block_id_map_align(),      fd_block_id_map_footprint( chain_cnt ) );
   l = FD_LAYOUT_APPEND( l, fd_txncache_align(),          fd_txncache_footprint( tile->replay.max_live_slots ) );
-  l = FD_LAYOUT_APPEND( l, fd_reasm_align(),             fd_reasm_footprint( tile->replay.fec_max ) );
+  l = FD_LAYOUT_APPEND( l, fd_reasm_align(),             fd_reasm_footprint( 512 ) );
```
lidatong: revert before merge

src/discof/repair/fd_repair_tile.c
```diff
@@ -1169,14 +1190,14 @@ unprivileged_init( fd_topo_t *      topo,
...
-      shred_out->mem        = topo->workspaces[ topo->objs[ link->dcache_obj_id ].wksp_id ].wksp;
-      shred_out->chunk0     = fd_dcache_compact_chunk0( shred_out->mem, link->dcache );
-      shred_out->wmark      = fd_dcache_compact_wmark( shred_out->mem, link->dcache, link->mtu );
-      shred_out->chunk      = shred_out->chunk0;
+      out_ctx_t * replay_out = ctx->repair_out_ctx;
+      replay_out->idx        = out_idx;
+      replay_out->mem        = topo->workspaces[ topo->objs[ link->dcache_obj_id ].wksp_id ].wksp;
+      replay_out->chunk0     = fd_dcache_compact_chunk0( replay_out->mem, link->dcache );
+      replay_out->wmark      = fd_dcache_compact_wmark( replay_out->mem, link->dcache, link->mtu );
+      replay_out->chunk      = replay_out->chunk0;
```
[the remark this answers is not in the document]
lidatong: fair, assert this later below

src/discof/repair/fd_repair_tile.c
lidatong: just FD_TEST the expected out links

src/discof/replay/fd_replay_tile.c
```diff
@@ -2034,35 +2040,36 @@ process_fec_set( fd_replay_tile_t *  ctx,
 
   for( fd_reasm_fec_t * curr = reasm_fec;; ) {
     curr = fd_reasm_parent( ctx->reasm, curr );
-    if( FD_UNLIKELY( !curr ) ) return; /* If can't connect, drop the FEC. */
+    if( FD_UNLIKELY( !curr ) ) return; /* If can't connect, drop the FEC. TODO: impossible? */
```
lidatong: yes, this should never happen based on current eviction unless the out queue is buggy -- would assert

src/discof/replay/fd_replay_tile.c
```diff
@@ -2034,35 +2040,36 @@ process_fec_set( fd_replay_tile_t *  ctx,
...
-    if( FD_UNLIKELY( fd_sched_can_ingest_cnt( ctx->sched )<leaf->fec_set_idx ) ) return;
+    if( FD_UNLIKELY( fd_sched_can_ingest_cnt( ctx->sched ) < (leaf->fec_set_idx/32 + 1) ) ) return;
     if( FD_UNLIKELY( fd_banks_is_full( ctx->banks ) ) ) return;
 
     /* Gather all FECs for this slot; */
+    fd_reasm_fec_t * slot_fecs[ FD_FEC_BLK_MAX ];
     fd_reasm_fec_t * curr = leaf;
     for(;;) {
-      slot_fecs[ curr->fec_set_idx ] = curr;
+      slot_fecs[ curr->fec_set_idx/32 ] = curr;
```
lidatong:
```suggestion
      slot_fecs[ curr->fec_set_idx/FD_FEC_SHRED_CNT ] = curr;
```

src/discof/replay/fd_replay_tile.c
```diff
@@ -2034,35 +2040,36 @@ process_fec_set( fd_replay_tile_t *  ctx,
...
       if( curr->fec_set_idx==0U ) break;
       curr = fd_reasm_parent( ctx->reasm, curr );
       FD_TEST( curr );
     }
+    FD_LOG_NOTICE(( "back-inserting FEC sets for slot %lu from fec_set_idx %u to fec_set_idx %u", leaf->slot, leaf->fec_set_idx, curr->fec_set_idx ));
```
lidatong:
```suggestion
    FD_LOG_WARNING(( "backfilling FEC sets for slot %lu from fec_set_idx %u to fec_set_idx %u", leaf->slot, leaf->fec_set_idx, curr->fec_set_idx ));
```

src/discof/replay/fd_replay_tile.c
```diff
@@ -2034,35 +2040,36 @@ process_fec_set( fd_replay_tile_t *  ctx,
...
 
-    for( ulong j=0UL; j<=leaf->fec_set_idx; j++ ) {
+    for( ulong j=0UL; j<=leaf->fec_set_idx/32; j++ ) {
```
lidatong:
```suggestion
    for( ulong j=0UL; j<=leaf->fec_set_idx/FD_FEC_SHRED_CNT; j++ ) {
```

src/discof/replay/fd_replay_tile.c
```diff
@@ -2198,6 +2205,25 @@ after_credit( fd_replay_tile_t *  ctx,
...
+  if( FD_UNLIKELY( ctx->reasm_evicted ) ) {
+    fd_reasm_evicted_t evicted = (fd_reasm_evicted_t){ .mr = ctx->reasm_evicted->key, .slot = ctx->reasm_evicted->slot, .fec_set_idx = ctx->reasm_evicted->fec_set_idx, .bank_idx = ctx->reasm_evicted->bank_idx };
+    fd_memcpy( fd_chunk_to_laddr( ctx->replay_out->mem, ctx->replay_out->chunk ), &evicted, sizeof(fd_reasm_evicted_t) );
+    fd_stem_publish( stem, ctx->replay_out->idx, REPLAY_SIG_REASM_EVICTED, ctx->replay_out->chunk,  sizeof(fd_reasm_evicted_t), 0UL, 0UL, fd_frag_meta_ts_comp( fd_tickcount() ) );
+    ctx->replay_out->chunk = fd_dcache_compact_next( ctx->replay_out->chunk, sizeof(fd_reasm_evicted_t), ctx->replay_out->chunk0, ctx->replay_out->wmark );
+
+    /* eviction policy only evicts chains of nodes until there is a
+       fork, so guaranteed that the evict path is always the left-child */
+    if( FD_LIKELY( ctx->reasm_evicted != ctx->evicted_sentinel ) ) fd_reasm_pool_release( ctx->reasm, ctx->reasm_evicted );
+    ctx->reasm_evicted = fd_reasm_child( ctx->reasm, ctx->reasm_evicted ); /* indexes into pool, safe to use */
```
lidatong, line 2219: hm thinking about this you can get rid of this `evicted_sentinel` concept -- if there's always room for 1 more reasm pool element, inside insert you can just acquire the element and immediately evict inside insert? then you don't have the special casing below too

src/discof/replay/fd_replay_tile.c
```diff
@@ -2223,6 +2249,16 @@ after_credit( fd_replay_tile_t *  ctx,
...
       mark_bank_dead( ctx, stem, bank->data->idx );
       fd_sched_block_abandon( ctx->sched, bank->data->idx );
+
+      /* evict it from reasm - we can guarantee this is a leaf because
+         no new bank is allocated until a slot boundary.  If a fec has
+         children that are of a different slot, then it would never be
+         evicted from banks because that means the bank finished
+         executing on that slot. */
+      fd_block_id_ele_t * block_id_ele = &ctx->block_id_arr[ bank->data->idx ];
+      fd_reasm_fec_t * fec = fd_reasm_query( ctx->reasm, &block_id_ele->latest_mr );
```
lidatong, line 2258: check: can a partially executed bank have a ref cnt and therefore exist after reasm has published past it?

src/discof/replay/fd_replay_tile.c
```diff
@@ -2461,11 +2488,28 @@ process_fec_complete( fd_replay_tile_t * ctx,
...
+       is to punt it and "go around."  We set the sentinel to this
+       failed FEC so that in after_credit we can let repair retry it. */
+    ctx->evicted_sentinel->key         = *merkle_root;
+    ctx->evicted_sentinel->slot        = shred->slot;
+    ctx->evicted_sentinel->fec_set_idx = shred->fec_set_idx;
+    ctx->evicted_sentinel->bank_idx    = ULONG_MAX;
+    ctx->evicted_sentinel->child       = ULONG_MAX;
+    ctx->reasm_evicted = ctx->evicted_sentinel;
+    return;
+  }
```
lidatong, line 2501: see above comment about special casing

src/discof/replay/fd_replay_tile.h
```diff
@@ -106,13 +108,23 @@ struct fd_replay_txn_executed {
 };
 typedef struct fd_replay_txn_executed fd_replay_txn_executed_t;
 
+struct fd_reasm_evicted {
+  fd_hash_t mr;
+  ulong     slot;
+  uint      fec_set_idx;
+  ulong     bank_idx;
+};
+typedef struct fd_reasm_evicted fd_reasm_evicted_t; /* duplicate typedef for convenience */
```
lidatong:
```suggestion
struct fd_replay_fec_evicted {
  fd_hash_t mr;
  ulong     slot;
  uint      fec_set_idx;
  ulong     bank_idx;
};
typedef struct fd_reasm_evicted fd_reasm_evicted_t; /* duplicate typedef for convenience */
```

lidatong: where is the duplicate typedef?

src/discof/replay/fd_replay_tile.c
emwang-jump: This happens because sometimes we can't evict anything in reasm by policy,  not necessarily because we don't have that 1 free element

src/app/firedancer/topology.c
emwang-jump: an artifact, changing it back to shred_depth

src/app/firedancer/topology.c
emwang-jump: yeah cause replay tile could be mid-process of a FEC but pre-insert which means store needs to hold +1. it does end up rounding up, will leave a comment

src/discof/replay/fd_replay_tile.h
emwang-jump: oops an artifact, removing

src/discof/replay/fd_replay_tile.c
emwang-jump: Not possible, seems like banks will only prune away a root if all the minority forks also have 0 refcnt, and this is root is also used for reasm publish

src/discof/replay/fd_replay_tile.c
```diff
@@ -2185,6 +2214,17 @@ after_credit( fd_replay_tile_t *  ctx,
...
       mark_bank_dead( ctx, stem, bank->data->idx );
       fd_sched_block_abandon( ctx->sched, bank->data->idx );
+
+      /* evict it from reasm - we can guarantee this is a leaf because
+         no new bank is allocated until a slot boundary.  If a fec has
+         children that are of a different slot, then it would never be
+         evicted from banks because that means the bank finished
+         executing on that slot. */
+      fd_block_id_ele_t * block_id_ele = &ctx->block_id_arr[ bank->data->idx ];
+      fd_reasm_fec_t * fec = fd_reasm_query( ctx->reasm, &block_id_ele->latest_mr );
```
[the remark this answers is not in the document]
lidatong: FD_TEST

PR #8300, opened 2026-02-13
Fix firedancer-dev bench

ripatel-fd:
Fixes bench-zen4-128core. Does 600k TPS on my box, needs further
tuning.

- Introduce HTTP code 503 for when RPC is not yet ready (no bank),
  fixes crash in bencho
- Enable sandbox when watch is disabled (seems to work fine)
- Add support for `firedancer-dev mem --topo bench`
- Drop bench genesis account configs to avoid hardcoded limit in
  fd_genesis_parse

### DIFF
--- a/src/app/firedancer/config/bench-zen3-32core.toml
+++ b/src/app/firedancer/config/bench-zen3-32core.toml
@@ -17,7 +17,7 @@
   sign_tile_count = 2
 
 [development.genesis]
-  fund_initial_accounts = 32768
+  fund_initial_accounts = 1000
 
 [development.bench]
   benchg_tile_count = 12
--- a/src/discof/rpc/fd_rpc_tile.c
+++ b/src/discof/rpc/fd_rpc_tile.c
@@ -529,7 +529,7 @@ fd_rpc_validate_config( fd_rpc_tile_t *             ctx,
     _bank_idx = ctx->finalized_idx;
   }
   if( FD_UNLIKELY( _bank_idx==ULONG_MAX ) ) {
-    *res = (fd_http_server_response_t){ .status = 500 };
+    *res = (fd_http_server_response_t){ .status = 503 };
     return 0;
   }
   *bank_idx = _bank_idx;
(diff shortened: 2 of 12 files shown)


ripatel-fd:
image

src/discof/rpc/fd_rpc_tile.c
```diff
@@ -1077,7 +1077,9 @@ static fd_http_server_response_t
 getLatestBlockhash( fd_rpc_tile_t * ctx,
                     cJSON const *   id,
                     cJSON const *   params ) {
-  if( FD_UNLIKELY( ctx->processed_idx==ULONG_MAX || ctx->banks[ ctx->processed_idx ].slot==ULONG_MAX ) ) return (fd_http_server_response_t){ .status = 500 };
+  if( FD_UNLIKELY( ctx->processed_idx==ULONG_MAX || ctx->banks[ ctx->processed_idx ].slot==ULONG_MAX ) ) {
```
mmcgee-jump, line 1080: @jherrera-jump is 500 for agave conformance?

mmcgee-jump: Don't love using a special code for this though ... we should match agave behavior if the snapshot isn't loaded yet or whatever (probably, the rpc doesn't accept connections until then?)

src/app/firedancer/config/bench-zen3-32core.toml
```diff
@@ -17,7 +17,7 @@
   sign_tile_count = 2
 
 [development.genesis]
-  fund_initial_accounts = 32768
+  fund_initial_accounts = 1000
```
mmcgee-jump, line 20: Why?

src/discof/rpc/fd_rpc_tile.c
jherrera-jump: >is 500 for agave conformance

no, I'll change the behavior here to match

src/app/firedancer/config/bench-zen3-32core.toml
ripatel-fd: See commit description

ripatel-fd: Any larger and fd_genesis_parse.c breaks

ripatel-fd:
@mmcgee-jump @jherrera-jump Are you ok with merging this as a quick fix? I just wanted a branch with bench working. The HTTP 503 hack is obviously ugly and not conformant, but at least it doesn't crash `firedancer-dev bench`.

Alternatively, maybe we can fix the underlying bugs properly?
- @jherrera-jump - fix RPC
- @ibhatt-jumptrading - make genesis_parse more scalable so we can boot of big genesis blobs
- @mmcgee-jump - fix watch crash with full client bench?

src/app/firedancer/config/bench-zen3-32core.toml
mmcgee-jump: Can you fix fd_genesis_parse ?

mmcgee-jump:
Sure, fine with it if we file an issue to track fix

src/app/firedancer/config/bench-zen3-32core.toml
ripatel-fd: @mmcgee-jump yes: https://github.com/firedancer-io/firedancer/pull/8401

PR #8306, opened 2026-02-13
Clean up account database config

ripatel-fd:


src/app/firedancer/config/default.toml
```diff
@@ -176,12 +176,18 @@ telemetry = true
     # Firedancer instance.
     genesis = ""
 
-    # Absolute path where to store the `accounts.db` file containing all
-    # account data from the chain.  If no path is provided, it defaults
```
mmcgee-jump, line 180: Why delete this info?

ripatel-fd: fixed

mmcgee-jump (dismissed):

mmcgee-jump (approved):

PR #8308, opened 2026-02-13
layout: skip HT pairs when CPU has enough physical cores

0x0ece:
Notes for Frankendancer:
1. We need cores also for Agave, so I set a 2x requirement (basically this is to avoid the default case with 21 tiles on a 32x2-core machine, that would leave Agave with just handful of cores)
2. Agave benefits from HT, so we always allocate it to HT pairs

### DIFF
--- a/src/disco/topo/fd_topob.c
+++ b/src/disco/topo/fd_topob.c
@@ -355,8 +355,9 @@ validate( fd_topo_t const * topo ) {
 }
 
 void
-fd_topob_auto_layout( fd_topo_t * topo,
-                      int         reserve_agave_cores ) {
+fd_topob_auto_layout_cpus( fd_topo_t *      topo,
+                           fd_topo_cpus_t * cpus,
+                           int              reserve_agave_cores ) {
   /* Incredibly simple automatic layout system for now ... just assign
      tiles to CPU cores in NUMA sequential order, except for a few tiles
      which should be floating. */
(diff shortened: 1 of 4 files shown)


src/disco/topo/fd_topob.c
```diff
@@ -495,7 +527,7 @@ fd_topob_auto_layout( fd_topo_t * topo,
             }
           }
 
-          if( FD_UNLIKELY( is_ht_critical ) ) {
+          if( FD_UNLIKELY( is_ht_critical && !skip_ht_pairs ) ) {
```
mmcgee-jump: I don't understand this logic. Shouldn't this be


```suggestion
          if( FD_UNLIKELY( is_ht_critical || skip_ht_pairs ) ) {
```

And you can remove the `cpu_assigned` loop above on line 498 - 504 ?

0x0ece: Yes, they're equivalent. I was focusing on minimizing changes in non skip_ht_pairs to avoid issues.
In fact I can also remove the change for Agave if I don't do 498-504.

I added tests (well, the AI added tests). I think the result is pretty nice, but lmk if you don't like I can remove it. Goal was to be able to concretely see the layout in various common config. For full Firedancer we can tweak the tests when topology is finalized.
I tested my code before and after your suggested change and they both pass all tests.

mmcgee-jump (dismissed):
Nice!

PR #8311, opened 2026-02-13
gui: fix block engine status icon

jherrera-jump:


src/app/firedancer/topology.c
```diff
@@ -1008,7 +1009,14 @@ fd_topo_initialize( config_t * config ) {
     /**/                 fd_topob_tile_in(  topo, "pack",   0UL,           "metric_in", "sign_pack",      0UL,        FD_TOPOB_UNRELIABLE, FD_TOPOB_UNPOLLED );
     /**/                 fd_topob_tile_out( topo, "sign",   0UL,                        "sign_pack",      0UL                                                );
 
-    /* TODO: bundle gui support needs to be integrated here */
+    if( config->tiles.gui.enabled ) { /* GUI is the only consumer of bundle_status */
```
mmcgee-jump, line 1012: && `bundle_enabled` ?

jherrera-jump: not visible in the diff, but this is inside an `if( FD_UNLIKELY( config->tiles.bundle.enabled ) ) {` block

Issue #8194, opened 2026-02-06
test_deshredder never runs


two-heart:
It is not the Local.mk and uses removed symbols

two-heart:
same for test_repair

PR #8319, opened 2026-02-17
repair: delete unused tests

Rexicon226:


two-heart:
Would fix half of https://github.com/firedancer-io/firedancer/issues/8194 but it is up to @lidatong and @emwang-jump to decide if deleting is the right thing to do here vs. fixing and building it.

PR #8328, opened 2026-02-17
gossip: add new parser

mmcgee-jump:
It is important for the gossip protocol that both Agave and Firedancer agree on what CRDS values are valid in the table.  A disagreement could cause repeated requests for the same "invalid" values, due to exclusion from the request bloom filter, which would waste bandwidth in a loop.

 * Add a new deserializer for gossip messages, with a fuzzer to ensure it agreees with the existing Agave parser perfectly, including the message sanitization pass.  The hand written deserializer is over 30x faster than the Agave one.

 * Also rewrote the serializer along with a fuzzer of its own to ensure messages that are serializable (kind of) round trip properly, and that the serializer is free of UB.

 * Consolidated the various gossip types behind a consolidated fd_gossip_message_t type, with fd_gossip_value_t for CRDS values within.  The performance benefit (none) of separating owned and view types didn't justify the complexity.

 * Fixed myraid bugs with both the serializer and deserializer, some found while rewriting, but many (5-10+) found by the fuzzers.  Of note, the transaction deserializer was using our fd_txn_parse which is not designed to match the agave vote transaction sanitizer, so this part was rewritten.

 * Updated the owned types to better align with the wire format for faster conversion.  As part of this, the socket addresses were updated to support IPv6 as given on the wire, and all the callers updated.


ptaffet-jump (approved):
Stamping to unblock. @riptl to review

PR #8330, opened 2026-02-17
set `--replay-transactions-threads` to value of `--unified-scheduler-handler-threads`

cavemanloverboy:
Since the introduction of `agave_unified_scheduler_handler_threads` argument in `fd_config_t`, agave has introduced yet another thread pool to handle entry verification (PoH) and sigverify in replay. This is `--replay-transactions-threads`.

The intention of `agave_unified_scheduler_handler_threads` is really to constrain the number of threads to assign to replay. However, the default value for `replay-transactions-threads` is `num_cpus::get()`. This PR intends to extend this constraint to `replay-transactions-threads`.

In a cleaner version of this PR, we would rename the config variable `agave_replay_threads` and use this for both the number of US handler threads and the number of replay entry/sigverify threads, but to avoid a breaking change near the end-of-life for `fdctl`, we simply do a little fix on the backend.

Why this PR?  I discovered i had 90 replay transactions threads...

### DIFF
--- a/src/app/fdctl/commands/run_agave.c
+++ b/src/app/fdctl/commands/run_agave.c
@@ -166,6 +166,7 @@ agave_boot( config_t const * config ) {
   ADDU( "--maximum-snapshot-download-abort", config->frankendancer.snapshots.maximum_snapshot_download_abort );
   ADDU( "--minimal-snapshot-download-speed", config->frankendancer.snapshots.minimum_snapshot_download_speed );
 
+  uint replay_threads;
   if( config->frankendancer.layout.agave_unified_scheduler_handler_threads ) {
     if( FD_UNLIKELY( config->frankendancer.layout.agave_unified_scheduler_handler_threads>config->topo.agave_affinity_cnt ) ) {
       FD_LOG_ERR(( "Trying to spawn %u handler threads but the agave subprocess has %lu cores. "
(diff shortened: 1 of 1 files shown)


src/app/fdctl/commands/run_agave.c
```diff
@@ -166,23 +166,25 @@ agave_boot( config_t const * config ) {
...
+    replay_threads = config->frankendancer.layout.agave_unified_scheduler_handler_threads;
   } else {
     //          agave_affinity_cnt >= 8  =>  agave_affinity_cnt - 4
     //     4 <= agave_affinity_cnt <  8  =>  4
     //          agave_affinity_cnt <  4  =>  agave_affinity_cnt
-    ulong num_threads = fd_ulong_if( config->topo.agave_affinity_cnt>=4UL,
-                                     fd_ulong_if( config->topo.agave_affinity_cnt>=8, config->topo.agave_affinity_cnt-4UL, 4UL ),
-                                     config->topo.agave_affinity_cnt );
-    ADDU( "--unified-scheduler-handler-threads", (uint)num_threads );
+    replay_threads = (uint)fd_ulong_if( config->topo.agave_affinity_cnt>=4UL,
```
mmcgee-jump, line 182: alignment

mmcgee-jump (dismissed):

src/app/fdctl/commands/run_agave.c
cavemanloverboy: what are we aligning?

cavemanloverboy: the subsequent two lines?

cavemanloverboy: done

PR #8338, opened 2026-02-18
gossip: filter crds without a contact info in pull requests

mmcgee-jump:
The CRDS in Agave stores values for staked nodes for two days.  It's not entirely clear why, but sometimes Agave happens to be storing many CRDS entries for nodes that have no contact info anymore.  The protocol requires us to filter these CRDS out on arrival (we don't know the contact info, and nobody has it to send to us, so we don't know the shred version, and therefore it can't enter the table).

This is a bit of a problem: if we filter all of these out naively, they don't make it into the bloom filter for our pull requests, and so we end up getting these values back repeatedly in a loop (this issue seems to impact Agave although they are not aware).

This change makes it so we "filter" the CRDS out, but also add them to a special structure so they get incorporated into our outgoing bloom filters.  If we do learn of the node later on, we have a fast index where we can remove all of their entries from the "filtered" structure.

### DIFF
--- a/src/discof/gossip/fd_gossvf_tile.c
+++ b/src/discof/gossip/fd_gossvf_tile.c
@@ -444,7 +444,8 @@ static void
 filter_shred_version_crds( fd_gossvf_tile_ctx_t * ctx,
                            uint                   tag,
                            fd_gossip_value_t *    values,
-                           ulong *                values_len ) {
+                           ulong *                values_len,
+                           uchar *                failed ) {
   ulong i = 0UL;
   while( i<*values_len ) {
     int keep      = 0;
@@ -475,9 +476,7 @@ filter_shred_version_crds( fd_gossvf_tile_ctx_t * ctx,
           ctx->metrics.crds_rx_bytes[ FD_METRICS_ENUM_GOSSVF_CRDS_OUTCOME_V_DROPPED_PUSH_ORIGIN_SHRED_VERSION_IDX ] += values[ i ].length;
         }
       }
-      values[ i ] = values[ *values_len-1UL ];
-      (*values_len)--;
-      continue;
+      failed[ i ] = 1;
     }
 
     i++;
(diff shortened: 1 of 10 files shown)


src/flamenco/gossip/crds/fd_crds.c
```diff
@@ -866,6 +981,99 @@ fd_crds_insert_failed_insert( fd_crds_t *   crds,
...
+  nci_origin_entry_t * origin_entry = nci_origin_map_ele_query( crds->purged.nci_origin_map, (fd_pubkey_t *)origin, NULL, crds->purged.nci_origin_pool );
+  if( FD_UNLIKELY( !origin_entry ) ) {
+    if( FD_LIKELY( nci_origin_pool_free( crds->purged.nci_origin_pool ) ) ) {
+      origin_entry = nci_origin_pool_ele_acquire( crds->purged.nci_origin_pool );
+      fd_memcpy( origin_entry->pubkey.uc, origin, 32UL );
+      origin_entry->head = ULONG_MAX;
+      nci_origin_map_ele_insert( crds->purged.nci_origin_map, origin_entry, crds->purged.nci_origin_pool );
+    }
+    /* If pool is full, entry still goes into treap+dlist but won't be
+       drainable by pubkey (will expire via 20s TTL instead). */
```
ptaffet-jump: Is this not `STAKED_EXPIRE_DURATION_NANOS`? (2 days?)

src/flamenco/gossip/crds/fd_crds.c
```diff
@@ -340,6 +350,55 @@ typedef struct fd_crds_purged fd_crds_purged_t;
...
+#define MAP_NAME               nci_origin_map
+#define MAP_KEY                pubkey
+#define MAP_ELE_T              nci_origin_entry_t
+#define MAP_KEY_T              fd_pubkey_t
+#define MAP_PREV               map.prev
+#define MAP_NEXT               map.next
+#define MAP_KEY_EQ(k0,k1)      fd_pubkey_eq( k0, k1 )
+#define MAP_KEY_HASH(key,seed) (seed^fd_ulong_load_8( (key)->uc ))
+#define MAP_OPTIMIZE_RANDOM_ACCESS_REMOVAL 1
+#include "../../../util/tmpl/fd_map_chain.c"
```
ptaffet-jump: I think you can use a single map_chain of fd_crds_purged_t with `MAP_MULTI 1`. That would simplify the code a bit. It might be a little hacky because you'll have to delete as you iterate, but it can be done safely.

src/flamenco/gossip/crds/fd_crds.c
mmcgee-jump: Ah this isn't right, we want it to drain faster

src/discof/gossip/fd_gossvf_tile.c
```diff
@@ -799,10 +790,12 @@ handle_net( fd_gossvf_tile_ctx_t * ctx,
     if( FD_UNLIKELY( !message->push->values_len ) ) return FD_METRICS_ENUM_GOSSVF_MESSAGE_OUTCOME_V_DROPPED_PUSH_NO_VALID_CRDS_IDX;
   }
 
-  int result = filter_shred_version( ctx, message );
+  uchar failed[ 17UL ] = {0};
```
ptaffet-jump: Where does 17 come from?

kbhargava-jump (approved):

PR #8355, opened 2026-02-19
gossip: align address checks with agave

mmcgee-jump:


src/discof/gossip/fd_gossvf_tile.c
```diff
@@ -555,53 +555,51 @@ check_duplicate_instance( fd_gossvf_tile_ctx_t *      ctx,
...
+    FD_LOG_NOTICE(( "pinging %s (" FD_IP4_ADDR_FMT ":%hu) (%lu)", base58, FD_IP4_ADDR_FMT_ARGS( addr.addr ), addr.port, ctx->ping_cnt ));
     ctx->ping_cnt++;
 #endif
     return 1;
   }
   return 0;
 }
 
+static int
+check_addr( fd_ip4_port_t addr,
```
ptaffet-jump, line 596: You should put a comment that return value of 1 means the address is bad. I was expecting the other way.

src/discof/gossip/fd_gossvf_tile.c
```diff
@@ -634,7 +635,13 @@ verify_addresses( fd_gossvf_tile_ctx_t * ctx,
       continue;
     }
 
-    if( FD_UNLIKELY( ping_if_unponged_contact_info( ctx, value, stem ) ) ) {
+    fd_ip4_port_t addr = {
+      .addr = value->contact_info->sockets[ FD_GOSSIP_CONTACT_INFO_SOCKET_GOSSIP ].is_ipv6 ? 0U : value->contact_info->sockets[ FD_GOSSIP_CONTACT_INFO_SOCKET_GOSSIP ].ip4,
```
ptaffet-jump, line 641:
```suggestion
      /* if it's IPv6, set addr to 0 so that check_addr drops it */
      .addr = value->contact_info->sockets[ FD_GOSSIP_CONTACT_INFO_SOCKET_GOSSIP ].is_ipv6 ? 0U : value->contact_info->sockets[ FD_GOSSIP_CONTACT_INFO_SOCKET_GOSSIP ].ip4,
```

ptaffet-jump (approved):

PR #8359, opened 2026-02-19
replace libsecp256k1 with custom secp256k1 recovery

Rexicon226:
Replaces our implementation of the secp256k1 recovery precompile/syscall with one based on s2n-bignum. The scalar implementation is only meant to serve as a reference, with a future improvement focusing on an AVX512 accelerated implementation.


config/cross/macos-arm-clang_x_linux-x86.mk
```diff
@@ -48,6 +48,4 @@ FD_HAS_ROCKSDB:=1
 CFLAGS+=-DFD_HAS_ROCKSDB=1
 ROCKSDB_LIBS:=-lrocksdb -lsnappy
 
-FD_HAS_SECP256K1:=1
-CFLAGS+=-DFD_HAS_SECP256K1=1
 LDFLAGS+=-lsecp256k1
```
ripatel-fd: remove this ldflags

src/ballet/fiat-crypto/secp256k1_montgomery_scalar_64.c
```diff
@@ -148,7 +148,7 @@ static FIAT_SECP256K1_MONTGOMERY_SCALAR_FIAT_INLINE void fiat_secp256k1_montgome
   uint64_t x2;
   uint64_t x3;
   x1 = (!(!arg1));
-  x2 = ((fiat_secp256k1_montgomery_scalar_int1)(0x0 - x1) & UINT64_C(0xffffffffffffffff));
+  x2 = ((unsigned long)(fiat_secp256k1_montgomery_scalar_int1)(0x0 - x1) & UINT64_C(0xffffffffffffffff));
```
ripatel-fd:
```suggestion
  x2 = ((uint64_t)(fiat_secp256k1_montgomery_scalar_int1)(0x0 - x1) & UINT64_C(0xffffffffffffffff));
```

ripatel-fd:
> Copilot encountered an error and was unable to review this pull request. You can try again by re-requesting a review.

 😂 🫵

src/app/firedancer-dev/commands/snapshot_load.c
```diff
@@ -23,6 +23,7 @@
 #include <sys/resource.h>
 #include <linux/capability.h>
 #include <unistd.h> /* close, sleep */
+#include <stdlib.h>
```
0x0ece, line 26: this?

src/ballet/secp256k1/fd_secp256k1.c
```diff
@@ -1,44 +1,100 @@
...
   }
 
-  if( FD_UNLIKELY( !secp256k1_ecdsa_recover( secp256k1_context_static, &internal_public_key,
-                                             &recoverable_sig, msg_hash ) ) ) {
+  fd_secp256k1_fp_t r[1];
+  bignum_tomont_p256k1( r->limbs, rs->limbs );
+
+  if( recovery_id & 2 ) {
+    /* If rs >= p - n, return NULL. Otherwise, add the n to r.
+       https://github.com/bitcoin-core/secp256k1/blob/c7a7f732bd17c71499341af77966c7ee2e75956c/src/modules/recovery/main_impl.h#L104-L109 */
```
0x0ece, line 53: nit: pref use explicit versions than commit, easier to know when there's an upgrade
https://github.com/bitcoin-core/secp256k1/blob/v0.7.1/src/modules/recovery/main_impl.h#L104-L109

src/ballet/secp256k1/fd_secp256k1.c
```diff
@@ -1,44 +1,100 @@
...
+  fd_secp256k1_scalar_tomont( msg, msg );
+
+  fd_secp256k1_scalar_t rn[1], u1[1], u2[1];
+  fd_secp256k1_point_t pubkey[1];
+
+  /* We delay converting rs into montgomery domain since we need to perform the compare outside of it. */
+  fd_secp256k1_scalar_tomont( s, s );
+
+  /* Unfortunately s2n-bignum has no API for performing in-montgomery inversion, so we invert and then convert. */
+  fd_secp256k1_scalar_invert( rn, rs );
```
0x0ece, line 78: on line 57 you may have added n to r. you should then add to rs?

src/ballet/secp256k1/fd_secp256k1.c
```diff
@@ -1,44 +1,100 @@
...
     return NULL;
   }
 
-  if( FD_UNLIKELY( !secp256k1_ec_pubkey_serialize( secp256k1_context_static, serialized_public_key, &public_key_len,
-                                                    &internal_public_key, SECP256K1_EC_UNCOMPRESSED ) ) ) {
-    /* COV: the output of secp256k1_ecdsa_recover should always serialize */
+  fd_uint256_t msg[1];
+  memcpy( msg, msg_hash, 32 );
+  fd_uint256_bswap( msg, msg );
+  fd_secp256k1_scalar_tomont( msg, msg );
```
0x0ece, line 69: check: we never checked that msg < n. is it ok to go to mont in msg >= n?

src/ballet/secp256k1/fd_secp256k1_s2n.c
```diff
@@ -0,0 +1,565 @@
...
+
+static inline fd_secp256k1_scalar_t *
+fd_secp256k1_scalar_mul( fd_secp256k1_scalar_t *       r,
+                         fd_secp256k1_scalar_t const * a,
+                         fd_secp256k1_scalar_t const * b ) {
+  /* bignum_montmul has an undocumented restriction
+     that the input and outputs may not alias. */
+  ulong t1[ 4 ];
+  ulong t2[ 4 ];
+  memcpy( t1, a->limbs, 32 );
```
0x0ece, line 50: One thing we could try is to add `restrict` to r? Just an idea.

src/ballet/secp256k1/fd_secp256k1_s2n.c
```diff
@@ -0,0 +1,565 @@
...
+  memcpy( t2, b->limbs, 32 );
+  bignum_montmul( 4, r->limbs, t1, t2, (ulong *)fd_secp256k1_const_n[0].limbs );
+  return r;
+}
+
+/* r = -a */
+static inline fd_secp256k1_scalar_t *
+fd_secp256k1_scalar_negate( fd_secp256k1_scalar_t *       r,
```
0x0ece, line 58: This is a bit confusing, maybe we should be explicit if the input should be mont or not.
You're using it on mont, but the const n is non-mont.

Personally, I'd find it more clear to just have sub non-mont and computed -u1 at the very end, after demont.

src/ballet/secp256k1/fd_secp256k1_s2n.c
```diff
@@ -0,0 +1,565 @@
...
+    } else {
+      fd_secp256k1_point_dbl( &r[i], &r[i / 2]    );
+    }
+  }
+  return r;
+}
+
+/* Computes s1*G + s2*P2, where G is the base point */
+static inline fd_secp256k1_point_t *
+fd_secp256k1_double_base_mul( fd_secp256k1_point_t *        r,
```
0x0ece, line 504: no endomorphism? 😆

src/app/firedancer-dev/commands/snapshot_load.c
Rexicon226: not included by default on arm

src/ballet/secp256k1/fd_secp256k1.c
Rexicon226: It seems to me this is what libsecp256k1 does. It copies sigr to fp form, then adds to it. But not back to sigr.
Copy: https://github.com/bitcoin-core/secp256k1/blob/v0.7.1/src/modules/recovery/main_impl.h#L101
Add: https://github.com/bitcoin-core/secp256k1/blob/v0.7.1/src/modules/recovery/main_impl.h#L108
Unmodified: https://github.com/bitcoin-core/secp256k1/blob/v0.7.1/src/modules/recovery/main_impl.h#L114

src/ballet/secp256k1/fd_secp256k1.c
Rexicon226: Seems like it, https://github.com/bitcoin-core/secp256k1/blob/v0.7.1/src/modules/recovery/main_impl.h#L149
I'm happy to improve docs here.

src/ballet/secp256k1/fd_secp256k1_s2n.c
0x0ece: Ok, resolved. -x is the same in mont and non-mont.

src/ballet/secp256k1/fd_secp256k1.c
0x0ece: I think they do reduce it though (ignoring overflow). Opportunity to use the infamous mod_4 :)
https://github.com/bitcoin-core/secp256k1/blob/master/src/scalar_4x64_impl.h#L151

`bignum_montmul` docs seem to recommend we reduce:
```
// Does z := (x * y / 2^{64k}) mod m, assuming x * y <= 2^{64k} * m, which is
// guaranteed in particular if x < m, y < m initially (the "intended" case).
```

src/ballet/secp256k1/fd_secp256k1.c
0x0ece: ok yes, confirmed.

src/ballet/secp256k1/fd_secp256k1_s2n.c
Rexicon226: I guess for the mul() case we do not need to copy out, as there are no aliasing callsites. I'll make this one restrict.

src/ballet/secp256k1/fd_secp256k1_s2n.c
Rexicon226: Not worth the complexity on a reference implementation imo :).

PR #8377, opened 2026-02-20
replay: ignore fecs after first slot complete

emwang-jump:


src/discof/replay/fd_replay_tile.c
```diff
@@ -1806,6 +1806,70 @@ process_fec_set( fd_replay_tile_t *  ctx,
...
+         If Agave receives the SLOT_COMPLETE C first, and then receives
+         shreds AFTER that, Agave will reject those shreds, mark
+         duplicate, but replay the slot.
+
+         We can't really do the same because if we mark a block dead
+         that others are going to dump and repair and vote on, our hfork
+         will mark it has a bank hash mismatch, and the validator will
+         crash.  Rather if we see this case, we need to make sure it
+         registers as an equivocation, and dump and repair this block
+         through regular equivocation pipeline.
```
lidatong: clarifying this -- agave either:
- receives, replays and votes for shorter before seeing longer
- sees longer before voting, and mark slot dead

since they're keyed by slot, they're implicitly dumping and repairing the shorter one for the longer one

since we're keyed by block id, we keep both shorter and longer block id regardless of receive order. we should not mark the shorter block id as dead, but the longer one should be marked dead

src/discof/replay/fd_replay_tile.c
```diff
@@ -1806,6 +1806,70 @@ process_fec_set( fd_replay_tile_t *  ctx,
...
+         and reset on it..... so likeeeee ???? the rest of our leader
+         rotation gets f-ed but consensus-wise is safe.
+
+         Another danger is let's say we replay A - B - C (slot cmpl) and
+         then stop there, but the correct chain is
+         A - B - C (slot cmpl) - D - E - F (slot cmpl).
+         LUCKILY..... the above is impossible with Agave.  If C comes
+         before anything else, DEF would be rejected. If D/E/F comes
+         before C, C would arrive and then make the slot be dead.  That
+         puts us in the same state as the case above. */
```
lidatong: think this explanation can be simplified:

the exact policy is similar to agave https://github.com/firedancer-io/firedancer/pull/8377/changes#r2840843432

- if we see all FEC sets of A,B,C (shorter) we will replay and vote for A,B,C
- if we see any FEC of D,E,F before seeing C, we should mark as duplicate (and therefore A,B,C will be withheld by reasm until confirmation and never delivered to be replayed and voted on)

src/discof/reasm/fd_reasm.c
```diff
@@ -425,6 +425,7 @@ fd_reasm_insert( fd_reasm_t *      reasm,
      coalesce connected orphans into the same tree.  This way we only
      need to search the orphan tree roots (vs. all orphaned nodes). */
 
+  ulong min_child_slot = ULONG_MAX; /* needed for eqvoc checks below */
```
lidatong: `deepest_descendant` 🧠

src/discof/reasm/fd_reasm.c
```diff
@@ -489,6 +491,73 @@ fd_reasm_insert( fd_reasm_t *      reasm,
...
+        1, where it isn't a true equivocation, it is impossible that
+        correct chain is to replay the full A - B - C (slot cmpl) - D -
+        E - F (slot cmpl). This is because in Agave, if C comes before
+        anything else, D/E/F would be rejected, and the slot would be
+        marked duplicate (and not voted on if it hadn't been voted on
+        already). If D/E/F came before C, C would arrive and Agave would
+        mark it dead(**).  So here we can just mark the slot
+        equivocating, and not replay the FEC.  If it just so happens
+        that version A-B-C gets confirmed, we can use fd_reasm_confirm
+        to redeliver those FEC sets. */
```
lidatong: I would re-phrase this to not really mention Agave and more just describe Firedancer's handling and perhaps just briefly mention at the end ("Agave's handling differs because they key by slot but this is compatible with the protocol as they implemented")

i think the two things to break down are:
1. what we the individual node do (mark it eqvoc, mark it dead, replay vote for it, repair it etc.)
2. what are the different scenarios if 52% swings one way or another, depending on what the majority of the cluster saw in what order

src/discof/reasm/fd_reasm.c
emwang-jump: unfortunately its `shallowest_descendant` which doesnt sound as nice 😓

PR #8378, opened 2026-02-20
rpc: add fuzz tests

jherrera-jump:
Add fuzz test for rpc RX POST path and fd_rpc_file_as_tarball, moving logic to a separate header file.

Applied the following small fix in two places, caught by fuzzer.
```
+ctx->banks[ ctx->processed_idx ]
+ctx->banks[ bank_idx ]
```


src/discof/rpc/fd_rpc.h
```diff
@@ -0,0 +1,1416 @@
+#ifndef HEADER_fd_src_discof_rpc_fd_rpc_h
```
mmcgee-jump: Not sure I'm following why all this implementation moved into the header?

jherrera-jump: Assumed including source file directly in fuzz test was a no-go but I see now this is done in a few places so will change it back

mmcgee-jump (dismissed):

PR #8391, opened 2026-02-23
http_server: fix bad overflow check

intrigus-lgtm:


src/waltz/http/fd_http_server.c
```diff
@@ -515,13 +515,13 @@ read_conn_http( fd_http_server_t * http,
         return;
       }
 
-      ulong next = content_len*10UL + (ulong)(content_length[ i ]-'0');
-      if( FD_UNLIKELY( next<content_len ) ) { /* Overflow */
+      ulong digit = (ulong)(content_length[ i ]-'0');
+      if( FD_UNLIKELY( content_len>(ULONG_MAX-digit)/10UL ) ) { /* Overflow */
```
ripatel-fd, line 519: Can we just use `__builtin_uaddl_overflow` here?

src/waltz/http/fd_http_server.c
```diff
@@ -515,13 +515,13 @@ read_conn_http( fd_http_server_t * http,
...
         close_conn( http, conn_idx, FD_HTTP_SERVER_CONNECTION_CLOSE_LARGE_REQUEST );
         return;
       }
 
-      content_len = next;
+      content_len = content_len*10UL + digit;
     }
 
     ulong total_len = (ulong)result+content_len;
```
intrigus-lgtm, line 527: @ripatel-fd something like this?
Doesn't look pretty, but would avoid a division although the compiler would have probably replaced the division by constant anyway with something smarter.

```suggestion
      ulong digit = (ulong)(content_length[ i ]-'0');
      ulong next_content_len;
      if( FD_UNLIKELY( __builtin_mul_overflow( content_len, 10UL, &next_content_len ) ||
                       __builtin_add_overflow( next_content_len, digit, &next_content_len ) ) ) { /* Overflow */
        close_conn( http, conn_idx, FD_HTTP_SERVER_CONNECTION_CLOSE_LARGE_REQUEST );
        return;
      }

      content_len = next_content_len;
}
ulong total_len = (ulong)result+content_len;
```

src/waltz/http/fd_http_server.c
```diff
@@ -515,13 +515,13 @@ read_conn_http( fd_http_server_t * http,
...
-      ulong next = content_len*10UL + (ulong)(content_length[ i ]-'0');
-      if( FD_UNLIKELY( next<content_len ) ) { /* Overflow */
+      ulong digit = (ulong)(content_length[ i ]-'0');
+      if( FD_UNLIKELY( content_len>(ULONG_MAX-digit)/10UL ) ) { /* Overflow */
         close_conn( http, conn_idx, FD_HTTP_SERVER_CONNECTION_CLOSE_LARGE_REQUEST );
         return;
       }
 
-      content_len = next;
+      content_len = content_len*10UL + digit;
```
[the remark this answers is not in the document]
ripatel-fd: @copilot ULONG_MAX is provided by fd_util_base.h i think

src/waltz/http/fd_http_server.c
```diff
@@ -515,13 +515,13 @@ read_conn_http( fd_http_server_t * http,
...
     }
```
ripatel-fd, line 525: @intrigus-lgtm How about we just copy out the integer string to `uchar x[]`, zero terminate it, and then call https://man7.org/linux/man-pages/man3/strtoumax.3.html for the conversion? IMO that's the cleanest option

ripatel-fd: @copilot please implement this suggestion by modifying PR 8391 and submit it as a separate PR. Maintain original attribution

PR #8407, opened 2026-02-23
runtime: make sure we update capitalization after refactor

nlgripto:
during refactor, break statement became a return.


nlgripto:
@ripatel-fd

nlgripto:
https://github.com/firedancer-io/auditor-internal/issues/373

PR #8411, opened 2026-02-23
snapshots: fix ssarchive latest pair

cali-jumptrading:


src/discof/restore/utils/fd_ssarchive.c
```diff
@@ -18,8 +20,7 @@ typedef struct fd_ssarchive_entry fd_ssarchive_entry_t;
 #define SORT_BEFORE(a,b) ( (a).slot>(b).slot )
 #include "../../../util/tmpl/fd_sort.c"
 
-#define FD_SSARCHIVE_MAX_FULL_ENTRIES        (512UL)
-#define FD_SSARCHIVE_MAX_INCREMENTAL_ENTRIES (512UL)
+#define FD_SSARCHIVE_MAX_ENTRIES (256UL)
```
jvarela-jump: `FD_SSARCHIVE_MAX_ENTRIES` is effectively half the previous quantity. Was this intentional?

cali-jumptrading: hmm what do you think? At 512, the two arrays would take 4MB, which is half the stack space.

jvarela-jump: if 256 for each is sufficient to hold enough entries, then the change is ok.

cali-jumptrading: hmm but maybe we should just go back to 512 since the stack can hold it

cali-jumptrading: I'll change it back

jvarela-jump (approved):

PR #8420, opened 2026-02-24
txsend: clear UDP ports on removal

two-heart:
Removal only cleared indices 0..1 (QUIC), while send paths read all 0..3 QUIC+UDP


mmcgee-jump:
I have a pending rewrite of this tile, just behind some gossip changes in the queue

ripatel-fd (approved):

PR #8421, opened 2026-02-24
genesis: fix stack memory leak over http

two-heart:


src/discof/genesis/fd_genesis_client.c
```diff
@@ -133,17 +133,22 @@ write_conn( fd_genesis_client_t * client,
...
+                                                                      "Cache-Control: no-cache\r\n"
+                                                                      "Connection: keep-alive\r\n"
+                                                                      "Pragma: no-cache\r\n"
+                                                                      "User-Agent: Firedancer\r\n"
+                                                                      "Host: " FD_IP4_ADDR_FMT ":%hu\r\n\r\n",
+                                                                      FD_IP4_ADDR_FMT_ARGS( peer->addr.addr ), fd_ushort_bswap( peer->addr.port ) ) );
+  if( FD_UNLIKELY( peer->request_bytes_sent>request_sz ) ) {
+    close_one( client, conn_idx );
+    return;
+  }
```
intrigus-lgtm: This condition as written right now should be impossible, right? So FD_TEST seems more appropriate.
```suggestion
  FD_TEST( request_sz > peer->request_bytes_sent );
```
`request_sz == peer->request_bytes_sent` should not be possible as in that case the peer should have transitioned to `writing = 0` and we'd return early.

two-heart: Fair point

PR #8435, opened 2026-02-24
poh: fix mixin race during skipped tick publishing

kbhargava-jump:


src/discof/poh/fd_poh.c
```diff
@@ -261,6 +261,11 @@ fd_poh_must_tick( fd_poh_t const * poh ) {
   return poh->state==STATE_LEADER && (poh->hashcnt%poh->hashcnt_per_tick)==(poh->hashcnt_per_tick-1UL);
 }
 
+int
+fd_poh_has_skipped_ticks( fd_poh_t const * poh ) {
+  return poh->state==STATE_LEADER && poh->last_slot<poh->slot;
```
mmcgee-jump, line 266: I think you need `(poh->last_slot<poh->slot || poh->last_hashcnt<poh->hashcnt)` ? Check with claude

Issue #8153, opened 2026-02-04
Correct program deployment checks to use feature set for the next slot


mjain-jump:
Agave's program deployment checks uses the environment for the next epoch. This causes a consensus break in the following scenario:

1. Feature gate for a new syscall gets activated at epoch 1000
2. User writes a program into a buffer account that references the new syscall
3. At the last slot of epoch 999, user deploys the program. 
- Agave transaction succeeds as verification uses upcoming feature set for epoch 1000. 
- FD transaction fails as verification uses the feature set for the current slot / epoch

Deployment checks need to be fixed to reference the feature set for the next slot. This would also require ensuring that feature accounts that get created during an epoch update the bank's active featureset with the correct activation slot.

PR #8446, opened 2026-02-24
runtime: use slot+1 feature set to verify deployments

mjain-jump:
Closes https://github.com/firedancer-io/firedancer/issues/8153


ripatel-fd (changes requested):
This approach is rather messy. Let's not modify global state, feature set should stay immutable per bank. I suggest instead modifying the "feature query" method used during deploy checks.

PR #8453, opened 2026-02-25
runtime: validate aligned realloc against original serialized length

two-heart:
The current length can be changed with CPI


two-heart:
Duplicate of https://github.com/firedancer-io/firedancer/pull/8453

PR #8462, opened 2026-02-25
snapshots: download progress check

cali-jumptrading:


src/discof/restore/fd_snapld_tile.c
```diff
@@ -228,6 +241,25 @@ transition_malformed( fd_snapld_tile_t *  ctx,
   fd_stem_publish( stem, 0UL, FD_SNAPSHOT_MSG_CTRL_ERROR, 0UL, 0UL, 0UL, 0UL, 0UL );
 }
 
+static void
+check_download_progress( fd_snapld_tile_t *  ctx,
+                         fd_stem_context_t * stem,
+                         long                now ) {
```
jvarela-jump: Should we return an `int` in order to break the `after_credit` case `FD_SSHTTP_ADVANCE_DATA` if the error was triggered and the `fd_sshttp_cancel` was invoked?

src/discof/restore/fd_snapld_tile.c
```diff
@@ -268,11 +300,15 @@ after_credit( fd_snapld_tile_t *  ctx,
...
-    int   result   = fd_sshttp_advance( ctx->sshttp, &data_len, out, fd_log_wallclock() );
+    long  now      = fd_log_wallclock();
+    int   result   = fd_sshttp_advance( ctx->sshttp, &data_len, out, now );
     switch( result ) {
       case FD_SSHTTP_ADVANCE_AGAIN:
+        check_download_progress( ctx, stem, now );
         break;
       case FD_SSHTTP_ADVANCE_DATA: {
+        ctx->bytes_in_window += data_len;
+        check_download_progress( ctx, stem, now );
```
jvarela-jump: should we break here on error?

PR #8464, opened 2026-02-25
Fix various cryptography link errors

ripatel-fd:
- Fixes build without ./deps.sh (useful for various security tools)
- Allows lots more compile units to build that don't depend on
  cryptography directly
- Resolve builtin programs at runtime in case s2n-bignum is missing
- Use a weak symbol as a fallback in case blst is missing
- Guard firedancer and firedancer-dev build targets behind
  FD_HAS_S2NBIGNUM


ripatel-fd:
This is the first time I've seen copilot get every single review message wrong

PR #8479, opened 2026-02-26
tower: fix oob accesses in tower deserialization

two-heart:


lidatong (commented):
thanks for fixing. this can prob be simplified to just be `CHECK( buf_sz, 2 )` or something in the caller. the changes to the private helpers aren't needed imo -- they're static to this file so can assume inputs satisfy the contract, as long as the callers in the file are

mmcgee-jump:
Good find  but code to be replaced by #8638

PR #8498, opened 2026-02-26
app: validate repair tile link count limits

two-heart:
Reject firedancer configs with shred_tile_count>16 or sign_tile_count>17 to prevent repair init out-of-bounds writes. No really a likely real world config, but we should not oob in any case.

### DIFF
--- a/src/app/shared/fd_config.c
+++ b/src/app/shared/fd_config.c
@@ -454,6 +454,9 @@ fd_config_fill( fd_config_t * config,
   }                                                    \
 } while(0)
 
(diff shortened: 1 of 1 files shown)


src/app/shared/fd_config.c
```diff
@@ -454,6 +454,9 @@ fd_config_fill( fd_config_t * config,
   }                                                    \
 } while(0)
 
+#define FD_REPAIR_MAX_SHRED_LINK_CNT (16U)
```
mmcgee-jump, line 457: Not a good place for this. Just `FD_TEST` them in the repair tile where it assigns to the arrays

PR #8500, opened 2026-02-26
config: enable LtHash verification by default

ripatel-fd:


cali-jumptrading:
> Wait for #8503 to be merged first! It fixes a recovery bug in snapls

oh oops I didn't realize you had auto merge on

PR #8502, opened 2026-02-26
repair: fix off-by-one in repair command vis

two-heart:


lidatong (approved):
good catch, thanks for fixing!

PR #8530, opened 2026-02-27
base64: make macro robust

intrigus-lgtm:


ripatel-fd (approved):

ripatel-fd:
well, base58 not base64

PR #8586, opened 2026-03-01
genesis: prefix genesis file path by network

ripatel-fd:
Create files such as mainnet-genesis.bin instead of genesis.bin.
This makes it easier to switch clusters during development.


jvarela-jump (approved):

ripatel-fd:
cc @mmcgee-jump

Issue #8598, opened 2026-03-01
Keyguard authorize ambiguity detected


ripatel-fd:
Debug log during snapshot loading

`sign:0 src/disco/keyguard/fd_keyguard_authorize.c(322): ambiguous payload type (role=0x1 mask=0x5)`

jvarela-jump:
To reproduce: this seems to happen with
```
build/native/gcc/bin/firedancer-dev --testnet
```

jvarela-jump:
It also happens with full client (e.g. with testnet config):
```
sudo build/native/gcc/bin/firedancer run --config config.toml 
```

jvarela-jump:
dissecting a35a3d6972f13c494172fcb06da30ab1f1cd16d2

PR #8611, opened 2026-03-01
keyguard: extend keyguard match txn checks

jvarela-jump:
Follow-up of https://github.com/firedancer-io/firedancer/pull/8575.

`fd_keyguard_payload_matches_txn_msg` requires additional checks to avoid
```
sign:0 src/disco/keyguard/fd_keyguard_authorize.c(322): ambiguous payload type (role=0x1 mask=0x5)
```

Closes https://github.com/firedancer-io/firedancer/issues/8598.

UPDATE: Superseded by https://github.com/firedancer-io/firedancer/pull/8630.

### DIFF
--- a/src/disco/keyguard/fd_keyguard_match.c
+++ b/src/disco/keyguard/fd_keyguard_match.c
@@ -154,6 +154,23 @@ fd_keyguard_payload_matches_txn_msg( uchar const * data,
 
   if( sig_cnt>addr_cnt ) return 0;
 
+  /* Verify that the remaining bytes can fit addr_cnt pubkeys and a
+     recent blockhash. */
+  ulong instr_off = addr_cnt * FD_PUBKEY_FOOTPRINT + FD_HASH_FOOTPRINT;
+  if( cursor + instr_off > end ) return 0;
+  cursor += instr_off;
+
+  /* Instruction count must be a valid compact_u16 and >= 1. */
+  if( cursor + 3UL > end ) return 0;
+  ulong instr_cnt_sz = fd_cu16_dec_sz( cursor, 3UL );
+  if( !instr_cnt_sz ) return 0;
+  ulong instr_cnt = fd_cu16_dec_fixed( cursor, instr_cnt_sz );
+  if( !instr_cnt ) return 0;
+
+  /* Each instruction is at least 3 bytes (program_id_index,
+     compact_u16 acct_idx_cnt, compact_u16 data_len) */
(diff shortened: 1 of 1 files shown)


src/disco/keyguard/fd_keyguard_match.c
```diff
@@ -154,6 +154,22 @@ fd_keyguard_payload_matches_txn_msg( uchar const * data,
 
   if( sig_cnt>addr_cnt ) return 0;
 
+  /* Verify that the remaining bytes can fit addr_cnt pubkeys and a
+     recent blockhash. */
+  cursor += addr_cnt * FD_PUBKEY_FOOTPRINT + FD_HASH_FOOTPRINT;
+  if( cursor > end ) return 0;
```
drubin-fd: Writing the check this way will trigger a CBMC error (as cursor could potentially go out of bounds of the provided `[data, data+sz)` memory region). Instead we could change it a bit, into something like:

```suggestion
  ulong offset = addr_cnt * FD_PUBKEY_FOOTPRINT + FD_HASH_FOOTPRINT;
  if( (cursor-data) + offset > sz ) return 0;
  cursor += offset;
```

Which CBMC agrees is equivalent to the version written here, and prevents it from complaining.

src/disco/keyguard/fd_keyguard_match.c
```diff
@@ -154,6 +154,22 @@ fd_keyguard_payload_matches_txn_msg( uchar const * data,
...
+  /* Instruction count must be a valid compact_u16 and >= 1. */
+  if( cursor + 3UL > end ) return 0;
+  ulong instr_cnt_sz = fd_cu16_dec_sz( cursor, 3UL );
+  if( !instr_cnt_sz ) return 0;
+  ulong instr_cnt = fd_cu16_dec_fixed( cursor, instr_cnt_sz );
+  if( !instr_cnt ) return 0;
+
+  /* Each instruction is at least 3 bytes (program_id_index,
+     compact_u16 acct_idx_cnt, compact_u16 data_len) */
+  if( cursor + instr_cnt * 3UL + instr_cnt_sz > end ) return 0;
```
drubin-fd: Same issue here. Could be rewritten to:

```suggestion
  if( (cursor-data) + (instr_cnt * 3UL + instr_cnt_sz) > sz ) return 0;
```

to appease CBMC (and also proven to be identical).

drubin-fd (changes requested):

jvarela-jump:
Superseded by https://github.com/firedancer-io/firedancer/pull/8630.

PR #8614, opened 2026-03-02
ci: add CBMC proofs

drubin-fd:
- Adds a workflow for checking CBMC proofs.
   
- Introduces a helper in the build system for adding
    CBMC proofs. To use, you specify the file containing
    said proofs, and a list of the function names within
    that are the entrypoints. CBMC will be invoked for
    each entrypoint listed.

### DIFF
--- a/config/everything.mk
+++ b/config/everything.mk
@@ -1,4 +1,4 @@
-.PHONY: all info check bin rust include lib unit-test integration-test fuzz-test help clean distclean asm ppp show-deps
+.PHONY: all info check bin rust include lib unit-test integration-test fuzz-test help clean distclean asm ppp show-deps proof
 .PHONY: run-unit-test run-integration-test run-script-test run-fuzz-test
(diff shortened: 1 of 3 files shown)


config/everything.mk
```diff
@@ -264,6 +264,23 @@ make-integration-test = $(eval $(call _make-exe,$(1),$(2),$(3),integration-test,
 run-integration-test  = $(eval $(call _run-integration-test,$(1)))
 make-fuzz-test = $(eval $(call _fuzz-test,$(1),$(2),$(3),$(4) $(LDFLAGS_EXE)))
 
+##############################
+# Usage: $(call make-proof,name,source_file)
+
+CBMC ?= cbmc
```
ripatel-fd: this should go into base.mk

ripatel-fd (dismissed):

ripatel-fd (approved):

PR #8621, opened 2026-03-02
sign, authorized_voter: use fd_memzero_explicit

intrigus-lgtm:
Found by https://github.com/firedancer-io/firedancer/pull/8533 :)


ripatel-fd:
What's wrong with explicit_bzero?

drubin-fd:
> What's wrong with explicit_bzero?

It's not easily available cross-target as it is non-standard, and the definition is so trivial we may as well just copy it in.
If we were to upgrade to C23, we could use `memset_explicit`.

PR #8629, opened 2026-03-02
tower: fix serdes bounds checks

nlgripto:
Add bounds-safe de_short_u16/de_var_int with src_sz parameter to prevent OOB reads on untrusted input. (fuzzer was dying immediately on startup)


nlgripto:
@ripatel-fd since you wrote the fuzzer

PR #8657, opened 2026-03-03
snapct: switch to tcp-based ping

### LINKED ISSUE #8255: snapct: unpriviliged ping sockets don't work on Ubuntu 24.04.3 LTS
I was trying to run the full client like this `sudo build/native/gcc/bin/firedancer run --config ~/config-no-gui.toml` when I encountered this error:
```bash
WARNING 02-11 22:42:35.028312 282516 f0   main src/app/shared/commands/configure/hyperthreads.c(69): pack cpu 5 has hyperthread pair cpu 29 which should be offline. Proceeding but performance may be reduced.
WARNING 02-11 22:42:35.028755 282516 f0   main src/app/shared/commands/configure/hyperthreads.c(73): poh cpu 7 has hyperthread pair cpu 31 which should be offline. Proceeding but performance may be reduced.
WARNING 02-11 22:42:35.944063 282516 f0   main src/disco/topo/fd_topo.c(157): fd_topo_wksp_new(funk) took 893 ms
WARNING 02-11 22:42:36.221430 282516 f0   main src/disco/topo/fd_topo.c(157): fd_topo_wksp_new(banks) took 275 ms
NOTICE  02-11 22:42:36.291575 282552 f0   metric:0 src/disco/metrics/fd_metric_tile.c(132): prometheus metrics endpoint listening at http://127.0.0.1:7999/metrics
ERR     02-11 22:42:36.294004 282565 15   snapct:0 src/discof/restore/utils/fd_ssping.c(182): socket(SOCK_DGRAM,IPPROTO_ICMP) failed (13-EACCES-permission denied)
```
Contrary to what is claimed here, not every user can use those sockets - in fact, not even root can depending on the configuration!
https://github.com/firedancer-io/firedancer/blob/69d8e8e93c8ffed4169b7327e7c843d98752f1f5/src/discof/restore/utils/fd_ssping.c#L177-L180
I had to run `sudo sysctl -w net.ipv4.ping_group_range="0 0"` to fix/enable it.

**Did I miss running some configuration step?**
(OS: Ubuntu 24.04.3 LTS)

<details>
<summary>~/config-no-gui.toml</summary>

```toml
[tiles.gui]
    enabled = false

[gossip]
    entrypoints = [
      "entrypoint.testnet.solana.com:8001",
      "entrypoint2.testnet.solana.com:8001",
      "entrypoint3.testnet.solana.com:8001",
    ]

[layout]
    verify_tile_count = 2
    execrp_tile_count = 2
    snapshot_hash_tile_count = 2
```
</details>

### LINKED ISSUE #8661: restore: ssping track seq number for repeated pings
```
/* FIXME: Properly set and track sequence numbers for repeated pings. */
```

### LINKED ISSUE #8662: restore: investigate ssping fd_ip4_port usage as key
```
/* FIXME: This code uses fd_ip4_port_t as the key for peers, but it
   should really just use uint (IPv4 address) as port has no meaning
   for ICMP pings.  Making this change however requires some significant
   changes in snapct as we are also effectively storing peer invalidation
   state in this data structure.  The number of distinct peers with
   the same IP address but different ports will be low, so this is fine
   for now. */
```

ptaffet-jump:
Closes #8255
Closes #8662 
Closes #8661 

Aslo addressed point 17 in https://github.com/firedancer-io/firedancer/issues/9176.

### DIFF
--- a/src/discof/restore/fd_snapct_tile.c
+++ b/src/discof/restore/fd_snapct_tile.c
@@ -456,6 +456,8 @@ init_load( fd_snapct_tile_t *  ctx,
       FD_TEST( fd_cstr_printf_check( ctx->http_incr_snapshot_name, PATH_MAX, NULL, "incremental-snapshot-%lu-%lu-%s.tar.zst", ctx->peer.full_slot, ctx->peer.incr_slot, encoded_hash ) );
     }
 
+    out->is_https = 0; /* if not found in the config list, it's not https */
+    out->hostname[0] = '\0'; /* .. and it doesn't have a hostname either. */
     for( ulong i=0UL; i<SERVER_PEERS_MAX; i++ ) {
       if( FD_UNLIKELY( ctx->peer.addr.l==ctx->config.sources.servers[ i ].addr.l ) ) {
         fd_cstr_ncpy( out->hostname, ctx->config.sources.servers[ i ].hostname, sizeof(out->hostname) );


ripatel-fd:
@copilot+claude-opus-4.6 code review[agent] Does this reliably clean up TCP connection state at the remote end? (RST/FIN sent to free up remote's state)

src/discof/restore/utils/fd_ssping.c
```diff
@@ -295,83 +289,125 @@ fd_ssping_invalidate( fd_ssping_t * ssping,
...
+  int fdesc = ssping->used_fds[ fdesc_idx ].fd;
+  /* Abort the connection attempt or close the connection by connecting
+     to AF_UNSPEC. */
+  struct sockaddr_in addr[1] = {{
+    .sin_family = AF_UNSPEC,
+    .sin_addr   = { .s_addr = 0U },
+    .sin_port   = 0
+  }};
+  if( FD_UNLIKELY( connect( fdesc, addr, sizeof(addr) ) ) ) FD_LOG_ERR(( "connect(AF_UNSPEC) failed (%d-%s)", errno, fd_io_strerror( errno ) ));
+
```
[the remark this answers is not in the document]
ptaffet-jump: > Some protocol sockets (e.g., TCP sockets as well as datagram
       sockets in the UNIX and Internet domains) may dissolve the
       association by connecting to an address with the sa_family member
       of sockaddr set to AF_UNSPEC; thereafter, the socket can be
       connected to another address.  (AF_UNSPEC is supported since Linux
       2.2.)

ptaffet-jump:
Based on a pcap, it seems that it does reliably clean up connection state. Scanning the linux source code confirms this:

- https://elixir.bootlin.com/linux/v6.8.2/source/net/ipv4/af_inet.c#L647 calls https://elixir.bootlin.com/linux/v6.8.2/source/net/ipv4/tcp.c#L2979
- https://elixir.bootlin.com/linux/v6.8.2/source/net/ipv4/tcp.c#L2939

It sends an RST when in the states the spec recommends doing so. In other states, it closes the socket, and when the host responds with a `SYN+ACK`, the kernel sends back an RST because there's no matching socket.

When you call close(), if there's no unread data, it initiates a TCP graceful shutdown, but if there is unread data, it initiates an abortive shutdown. Calling connect with AF_UNSPEC initiates an abortive shutdown in both cases.

src/discof/restore/utils/fd_ssping.c
```diff
@@ -295,83 +289,125 @@ fd_ssping_invalidate( fd_ssping_t * ssping,
...
+      if( FD_LIKELY( !is_err ) ) {
+        peer->latency_nanos  = (ulong)fd_long_max( now - (peer->deadline_nanos - PEER_DEADLINE_NANOS_PING), 1L );
+        peer->state          = PEER_STATE_VALID;
+        peer->deadline_nanos = now + PEER_DEADLINE_NANOS_VALID;
+        deadline_list_ele_push_tail( ssping->valid, peer, ssping->pool );
+
+        FD_LOG_INFO(( "pinged " FD_IP4_ADDR_FMT ":%hu in %lu nanos",
+              FD_IP4_ADDR_FMT_ARGS( peer->addr.addr ), fd_ushort_bswap( peer->addr.port ), peer->latency_nanos ));
+        ssping->on_ping_cb( ssping->cb_arg, peer->addr, peer->latency_nanos );
+      } else {
```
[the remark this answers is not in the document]
ptaffet-jump: I read the [tcp_poll](https://elixir.bootlin.com/linux/v6.8.2/C/ident/tcp_poll) source and don't see any case in which this can happen. Claude agrees.

src/discof/restore/fd_snapct_tile.c
```diff
@@ -351,9 +351,10 @@ static ulong
 rlimit_file_cnt( fd_topo_t const *      topo FD_PARAM_UNUSED,
                  fd_topo_tile_t const * tile ) {
   ulong cnt = 1UL +                             /* stderr */
-              1UL;                              /* logfile */
+              1UL +                             /* logfile */
+              1UL;                              /* boot control pipe */
```
ptaffet-jump, line 361: In a follow-up PR, I'll make `fd_http_resolver` allocate all its file descriptors in privileged_init so that we can set rlimit to 0.

src/discof/restore/fd_snapct_tile.seccomppolicy
```diff
@@ -74,28 +75,31 @@ close: (not (or (eq (arg 0) 2)
...
+                (and (>= (arg 0) min_ping_fd)
+                     (<= (arg 0) max_ping_fd))))
 
 # snapshot: we need to be able to poll existing connections that we want
 # to download snapshots and resolve snapshots from
 ppoll
 
+# snapshot: we need to be able to poll the ping connections to see if
+# any peers have responded
+poll
```
ripatel-fd: can we use ppoll everywhere? poll syscall doesn't exist on some architectures, e.g. arm64

ptaffet-jump:
TCP based ping times seem to be about 2% faster, but I'm not sure why. Otherwise, they are extremely comparable.

src/discof/restore/utils/fd_ssping.c
jvarela-jump:
```suggestion
        deadline_list_ele_remove( ssping->pinged, peer, ssping->pool );
        remove_fdesc_idx( ssping, peer->used_fd_idx );
```

jvarela-jump:
```suggestion
        deadline_list_ele_remove( ssping->refreshing, peer, ssping->pool );
        remove_fdesc_idx( ssping, peer->used_fd_idx );
```

src/discof/restore/fd_snapct_tile.c
```diff
@@ -443,6 +456,7 @@ init_load( fd_snapct_tile_t *  ctx,
       FD_TEST( fd_cstr_printf_check( ctx->http_incr_snapshot_name, PATH_MAX, NULL, "incremental-snapshot-%lu-%lu-%s.tar.zst", ctx->peer.full_slot, ctx->peer.incr_slot, encoded_hash ) );
     }
 
+    out->is_https = 0; /* if not found in the config list, it's not https */
```
jvarela-jump, line 459:
```suggestion
    out->is_https = 0; /* if not found in the config list, it's not https */
    out->hostname[0] = '\0';  /* gossip peers have no hostname */
```

src/discof/restore/utils/fd_ssping.c
jvarela-jump: This would require `remove_fdesc_idx()` to be defined before `fd_ssping_remove()`.

jvarela-jump: This would require `remove_fdesc_idx()` to be defined before `fd_ssping_remove()`.

src/discof/restore/fd_snapct_tile.c
ptaffet-jump: Good point. Fixed.

src/discof/restore/utils/fd_ssping.c
ptaffet-jump: Good catch.

PR #8715, opened 2026-03-04
claude: add CLAUDE.md

topointon-jump:
Adds minimal CLAUDE.md

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,54 @@
+# Firedancer
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,50 @@
+# Firedancer
```
ibhatt-jumptrading, line 1: you forgot to include "no mistakes, do a good job"

topointon-jump: ah crap

PR #8738, opened 2026-03-05
util: fix typo in bplus implementation

drubin-fd:
`pair_t` comes from the test_bplus.c file, but users define their own type through `BPLUS_PAIR_T`


drubin-fd:
> Shows that fd_bplus doesn't have much use in the rest of the project...

Yeah, it has no usages. I only found this when I was implementing replay server, but I've since moved away from the B+ tree in my design anyways, so back to no users 😄.

PR #8749, opened 2026-03-05
gui: incrementally index votes

jherrera-jump:
Eliminate GUI dependence on replay banks and get vote updates from votes on
- replay->execrp / execrp->replay
- pack->bank | bank->poh

Technically, votes on a non-canonical fork can show up in the GUI vote index but since we use this info for the purposes of monitoring delinquency, this is a non-issue.

closes https://github.com/firedancer-io/firedancer/issues/8692

### DIFF
--- a/src/disco/gui/fd_gui_peers.c
+++ b/src/disco/gui/fd_gui_peers.c
@@ -243,6 +243,11 @@ fd_gui_peers_new( void *             shmem,
 
     ctx->slot_voted = ULONG_MAX;
 
+    for( ulong i=0UL; i<2UL; i++ ) {
+      ctx->epochs[ i ].epoch      = ULONG_MAX;
+      ctx->epochs[ i ].stakes_cnt = 0UL;
+    }
+
     ctx->next_client_nanos              = now;
     ctx->next_metric_rate_update_nanos  = now;
     ctx->next_gossip_stats_update_nanos = now;
@@ -704,6 +709,7 @@ fd_gui_peers_handle_gossip_update( fd_gui_peers_ctx_t *               peers,
           memset( &peer->gossvf_rx_sum, 0, sizeof(peer->gossvf_rx_sum) );
           memset( &peer->gossip_tx_sum, 0, sizeof(peer->gossip_tx_sum) );
           peer->has_vote_info = 0;
+          peer->delinquent = 0;
           peer->stake = ULONG_MAX;
 
           fd_gui_config_parse_info_t * info =  fd_gui_peers_node_info_map_ele_query( peers->node_info_map, fd_type_pun_const(update->origin ), NULL, peers->node_info_pool );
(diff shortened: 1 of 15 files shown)


src/disco/gui/fd_gui_tile.c
```diff
@@ -549,20 +494,39 @@ after_frag( fd_gui_ctx_t *      ctx,
...
                                       trailer->txn_preload_end_pct,
                                       trailer->tips );
+
+      for( ulong i=0UL; i<txn_cnt; i++ ) {
+        fd_txn_p_t * txn_p = &((fd_txn_p_t *)src)[ i ];
+        if( FD_UNLIKELY( (txn_p->flags & FD_TXN_P_FLAGS_IS_SIMPLE_VOTE) && (txn_p->flags & FD_TXN_P_FLAGS_EXECUTE_SUCCESS) ) ) {
+          fd_pubkey_t identity[ 1 ];
+          fd_pubkey_t vote_acct[ 1 ];
+          ulong       vote_slot;
+          if( FD_LIKELY( fd_txn_parse_simple_vote( TXN( txn_p ), txn_p->payload, identity, vote_acct, &vote_slot ) ) ) {
```
mmcgee-jump: Why would parsing fail if it successfully executed? You can probably FD_TEST this?

mmcgee-jump: Also technically the vote isn't valid yet you need it to later sigverify too, but not sure if that matters for this UI use case?

src/discof/replay/fd_replay_tile.h
```diff
@@ -64,6 +63,17 @@ struct fd_replay_slot_completed {
   int is_leader; /* whether we were leader for this slot */
   ulong identity_balance;
 
+  /* since slot start, default ULONG_MAX */
+  ulong total_txn_cnt;
```
mmcgee-jump: No need to line these up to the bizarre agave / frontend taxonomy we have. Just define a mutually exclusive+exhaustive set instead.

(I presume it's vote_success, vote_fail, nonvote_success, nonvote_fail ?)

src/disco/gui/fd_gui_tile.c
mmcgee-jump: Also why do we reparse for execle, but the execrp tiles do this parse for the replay case?

src/disco/gui/fd_gui_peers.c
```diff
@@ -786,134 +792,189 @@ fd_gui_peers_handle_gossip_update( fd_gui_peers_ctx_t *               peers,
...
-  ulong total_stake = 0UL;
-  fd_pubkey_t prev_peer = { 0 };
-  for( ulong i=0UL; i<vote_cnt; i++ ) {
-    if( FD_UNLIKELY( !memcmp( prev_peer.uc, votes_sorted[ i ].node_account.uc, sizeof(fd_pubkey_t) ) ) ) {
-      votes_sorted[ i ].stake = ULONG_MAX; /* flag as duplicate */
-    } else {
-      total_stake += votes_sorted[ i ].stake;
-    }
-    prev_peer = votes_sorted[ i ].node_account;
+fd_gui_peers_handle_vote( fd_gui_peers_ctx_t * peers,
```
mmcgee-jump, line 806: let's discuss aligning these GUI structures to the fork tree the rest of the application uses

src/disco/gui/fd_gui_tile.c
jherrera-jump: > Also why do we reparse for execle, but the execrp tiles do this parse for the replay case?
oh duh

> vote isn't valid yet you need it to later sigverify too
yeah figured it doesn't matter so kept it simple

PR #8752, opened 2026-03-05
gossip: fix push_dlist livelock in active set

cmoyes-jump:


cmoyes-jump:
closing in favor of https://github.com/firedancer-io/firedancer/pull/8753

PR #8757, opened 2026-03-06
log: improve detection of color support

drubin-fd:


src/util/log/fd_log.c
```diff
@@ -1627,12 +1622,90 @@ fd_log_private_stack_discover( ulong   stack_sz,
...
+  return 0;
+}
+
 #elif FD_LOG_STYLE==1 /* generic embedded target */
 
 #include "fd_log.h"
 
 FD_FN_UNUSED static inline void _unused( void ) {} /* required due to -Wpedantic */
 
+FD_FN_UNUSED static inline int fd_log_should_colorize( void ) { return 0; }
```
ripatel-fd: Leave this undefined or make it a weak symbol. There are embedded targets with color support (e.g. the para-virtualized kernel that forwards messages to hypervisor)

src/util/log/fd_log.c
```diff
@@ -1186,12 +1186,7 @@ fd_log_private_boot( int  *   pargc,
     char const * cstr = fd_env_strip_cmdline_cstr( pargc, pargv, "--log-colorize", "FD_LOG_COLORIZE", NULL );
     if( cstr ) { colorize = fd_cstr_to_int( cstr ); break; }
 
-    cstr = fd_env_strip_cmdline_cstr( NULL, NULL, NULL, "COLORTERM", NULL );
-    if( cstr && !strcmp( cstr, "truecolor" ) ) { colorize = 1; break; }
-
-    cstr = fd_env_strip_cmdline_cstr( NULL, NULL, NULL, "TERM", NULL );
-    if( cstr && strstr( cstr, "256color" ) ) { colorize = 1; break; }
-
+    colorize = fd_log_should_colorize();
```
ripatel-fd, line 1189: This is breaks the `--fd-log-colorize` flag

ripatel-fd: Nvm, didn't see the inline break

src/util/log/fd_log.c
drubin-fd: Just made it undefined, good points.

ripatel-fd (approved):

Issue #8654, opened 2026-03-03
Fix bundle tile STEM_BURST


mmcgee-jump:
Single recvmsg can deliver multiple gRPC messages → multiple bundles → unbounded publishes to verify_out. Needs restructuring to cap publishes per after_credit, or correct/accurate upper bound based on max message size

ripatel-fd:
@mmcgee-jump Sure, we can make a stateful receive buffer and work off one bundle per after_credit.

PR #8763, opened 2026-03-06
bundle: defer publish to after_credit via pending deque

cmoyes-jump:
Split into `before_credit` (I/O, gated on deque empty and `cr_avail`) and `after_credit` (drain one bundle or up to `STEM_BURST` packets per call). Publish metadata is buffered in a zero-copy deque. dcache writes are capped at `cr_avail` to prevent overwriting data the downstream consumer hasn't read.

Fixes https://github.com/firedancer-io/firedancer/issues/8654

### DIFF
--- a/src/disco/bundle/fd_bundle_tile.c
+++ b/src/disco/bundle/fd_bundle_tile.c
@@ -37,7 +37,6 @@ scratch_align( void ) {
 
 FD_FN_CONST static ulong
 scratch_footprint( fd_topo_tile_t const * tile ) {
-  (void)tile;
   ulong pending_max = tile->bundle.out_depth;
   ulong l = FD_LAYOUT_INIT;
   l = FD_LAYOUT_APPEND( l, alignof(fd_bundle_tile_t), sizeof(fd_bundle_tile_t)                        );


src/disco/metrics/metrics.xml
```diff
@@ -378,6 +378,8 @@ metric introduced.
     <histogram name="MessageRxDelayNanos" min="100000" max="1000000000">
         <summary>Message receive delay in nanoseconds from bundle server to bundle client</summary>
     </histogram>
+    <gauge name="PendingTransactions" summary="Number of transactions buffered and waiting to be published" />
```
jherrera-jump, line 381: PendingTransactions here is probably less useful as a metric due to how slow consumers poll metrics compared to the rate of change of the metric itself.

Maybe instead this should be the number of times `pending_transaction_cnt` exceeds `pending_max/2`?

Also don't feel too strongly about this so fine to leave it as-is.

src/disco/metrics/metrics.xml
```diff
@@ -1040,6 +1042,8 @@ metric introduced.
     <gauge name="IncrementalBytesTotal" summary="Total size of the incremental snapshot file. Might change if snapshot load is aborted and restarted" />
     <gauge name="IncrementalDownloadRetries" summary="Number of times we retried the incremental snapshot download because the peer was too slow" />
     <gauge name="PredictedSlot" summary="The predicted slot from which replay starts after snapshot loading finishes. Might change if snapshot load is aborted and restarted" />
+    <gauge name="GossipFreshCount" summary="Number of fresh gossip peers seen when collecting gossip peers. " />
+    <gauge name="GossipTotalCount" summary="Number of total gossip peers seen when collecting gossip peers. " />
```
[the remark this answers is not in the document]
cmoyes-jump: I suspect this was a pending auto-generation.

src/disco/bundle/fd_bundle_client.c
```diff
@@ -483,13 +483,18 @@ fd_bundle_tile_publish_bundle_txn(
     return;
   }
 
+  if( FD_UNLIKELY( pending_pub_cnt( ctx->pending_pubs )>=ctx->stem->cr_avail[ ctx->verify_out.idx ] ) ) {
```
jherrera-jump: Instead of checking credits here, can we just always add to `ctx->pending_pubs` unless the queue is full (in which case we drop the bundle)?

I would generally prefer to keep credit management out of tile code if possible. Since `after_credit` will only be called (and drain the queue) when sufficient credits are available, there's no risk of backpressure.

src/disco/bundle/fd_bundle_client.c
```diff
@@ -499,20 +504,20 @@ fd_bundle_tile_publish_bundle_txn(
...
-  ulong tspub = (ulong)fd_frag_meta_ts_comp( fd_bundle_now() );
-  fd_stem_publish( ctx->stem, ctx->verify_out.idx, sig, ctx->verify_out.chunk, sz, 0UL, 0UL, tspub );
+  ulong sz = fd_txn_m_realized_footprint( txnm, 0, 0 );
+  fd_bundle_pending_pub_t pub = {
+    .chunk      = ctx->verify_out.chunk,
+    .sz         = sz,
+    .sig        = 1UL,
+    .bundle_seq = ctx->bundle_seq,
+  };
+  pending_pub_push_tail( ctx->pending_pubs, pub );
```
jherrera-jump: I think `push_tail` here should be all-or-nothing for a bundle. We don't want to partially push some of the transactions in a bundle but then leave out the last few because they were dropped.

src/disco/metrics/metrics.xml
```diff
@@ -1040,6 +1042,8 @@ metric introduced.
     <gauge name="IncrementalBytesTotal" summary="Total size of the incremental snapshot file. Might change if snapshot load is aborted and restarted" />
     <gauge name="IncrementalDownloadRetries" summary="Number of times we retried the incremental snapshot download because the peer was too slow" />
     <gauge name="PredictedSlot" summary="The predicted slot from which replay starts after snapshot loading finishes. Might change if snapshot load is aborted and restarted" />
+    <gauge name="GossipFreshCount" summary="Number of fresh gossip peers seen when collecting gossip peers. " />
```
mmcgee-jump: merge conflict, these are removed

src/disco/bundle/fd_bundle_tile.c
```diff
@@ -140,14 +148,42 @@ fd_bundle_tile_publish_block_engine_update(
...
 
+static void
+before_credit( fd_bundle_tile_t *  ctx,
+               fd_stem_context_t * stem,
+               int *               charge_busy ) {
+  if( FD_UNLIKELY( !ctx->stem ) ) {
+    ctx->stem = stem;
+  }
+
+  if( FD_LIKELY( pending_pub_empty( ctx->pending_pubs ) && ctx->stem->cr_avail[ ctx->verify_out.idx ]>=STEM_BURST ) ) {
```
jherrera-jump: I think instead of moving this to before_credit with an explicit credit check, keep it in after_credit and only drain up to STEM_BURST transactions at a time?

src/disco/bundle/fd_bundle_client.c
cmoyes-jump: @mmcgee-jump had the same feedback, redid.

src/disco/bundle/fd_bundle_tile.c
cmoyes-jump: this is now gone

src/disco/bundle/fd_bundle_client.c
cmoyes-jump: resolved

src/disco/metrics/metrics.xml
cmoyes-jump: if others also feel this is worth doing, I will change, how does that sound?

src/disco/bundle/fd_bundle_tile.c
```diff
@@ -26,18 +27,23 @@
...
 
 FD_FN_CONST static ulong
 scratch_footprint( fd_topo_tile_t const * tile ) {
   (void)tile;
+  ulong pending_max = tile->bundle.buf_sz / FD_BUNDLE_MIN_GRPC_WIRE_SZ;
   ulong l = FD_LAYOUT_INIT;
   l = FD_LAYOUT_APPEND( l, alignof(fd_bundle_tile_t), sizeof(fd_bundle_tile_t)                        );
   l = FD_LAYOUT_APPEND( l, fd_grpc_client_align(),    fd_grpc_client_footprint( tile->bundle.buf_sz ) );
   l = FD_LAYOUT_APPEND( l, fd_alloc_align(),          fd_alloc_footprint()                            );
+  l = FD_LAYOUT_APPEND( l, pending_txn_align(),       pending_txn_footprint( pending_max )            );
```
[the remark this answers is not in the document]
cmoyes-jump: need guidance on whether we can safely size down the theoretical upper bound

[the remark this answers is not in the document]
jherrera-jump: I think you can just size to match the bundle_verif link, i.e. `config->tiles.verify.receive_buffer_size / FD_TPU_PARSED_MTU`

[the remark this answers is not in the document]
cmoyes-jump: I agree, that makes sense, I made your suggested change

src/disco/bundle/fd_bundle_tile.c
```diff
@@ -260,10 +315,13 @@ privileged_init( fd_topo_t *      topo,
                  fd_topo_tile_t * tile ) {
   void * scratch = fd_topo_obj_laddr( topo, tile->tile_obj_id );
 
+  const ulong pending_max = tile->bundle.out_depth;
```
ripatel-fd:
```suggestion
  ulong const pending_max = tile->bundle.out_depth;
```

ripatel-fd (approved):

PR #8765, opened 2026-03-06
progcache: rewrite with cache replacement

ripatel-fd:


src/flamenco/progcache/fd_progcache_admin.c
```diff
@@ -25,502 +27,545 @@ fd_progcache_est_rec_max( ulong wksp_footprint,
...
-
-  rec->map_next = FD_FUNK_REC_IDX_NULL;
-  fd_funk_val_flush( rec, funk->alloc, funk->wksp );
-  fd_funk_rec_pool_release( funk->rec_pool, rec, 1 );
+  rec->map_next = UINT_MAX;
+  fd_progcache_val_free( rec, cache );
+  rec->exists = 0;
+  fd_rwlock_unwrite( &rec->lock );
+  fd_prog_recp_release( cache->rec.pool, rec, 1 );
+  fd_progcache_admin_metrics_g.gc_root_cnt++;
```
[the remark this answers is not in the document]
ripatel-fd: This is just wrong

src/flamenco/progcache/fd_progcache.h
```diff
@@ -0,0 +1,208 @@
...
+  struct {
+    fd_rwlock_t        lock;
+    fd_progcache_rec_t rec[ FD_MAX_INSTRUCTION_STACK_DEPTH ];
+    uint               rec_used;
+    uint               spad_used;
+    uint               spad_off[ FD_MAX_INSTRUCTION_STACK_DEPTH ];
+    uchar              spad[ FD_PROGCACHE_SPAD_MAX ] __attribute__((aligned(64UL)));
+  } spill;
+
+};
```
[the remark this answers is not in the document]
ripatel-fd: retard mode...

src/flamenco/progcache/fd_progcache_admin.h
```diff
@@ -55,7 +46,7 @@ fd_progcache_admin_leave( fd_progcache_admin_t * cache,
    this is called. */
 
 void
-fd_progcache_txn_attach_child( fd_progcache_admin_t *    cache,
+fd_progcache_txn_attach_child( fd_progcache_join_t *     cache,
                                fd_funk_txn_xid_t const * xid_parent,
```
mmcgee-jump, line 50: using `fd_funk_txn_xid_t` is dumb and adds an additional map traversal everywhere which isn't necessary. Just use bank_idx directly, and it's already the index is an 4096 wide array.

src/flamenco/progcache/fd_progcache_admin.c
```diff
@@ -25,501 +28,541 @@ fd_progcache_est_rec_max( ulong wksp_footprint,
...
   }
 
-  /* Phase 5: Free transaction object */
+  /* Free transaction object */
 
-  fd_rwlock_unwrite( &funk->txn_lock[ txn_idx ] );
-  FD_VOLATILE( txn->state ) = FD_FUNK_TXN_STATE_FREE;
-  fd_funk_txn_pool_release( funk->txn_pool, txn, 1 );
+  fd_rwlock_unwrite( &txn->lock );
+  fd_prog_txnp_ele_release( cache->txn.pool, txn );
```
mmcgee-jump, line 184: I don't like this pattern of release lock, then cleanup after, because if the whole application shares a bound on max_live_banks it creates a race window where we can crash here (replay evicts the bank, gets told its now reusable, another thread comes in and tries to grab the 4096th element, and now its not yet freed so app crashes).

mmcgee-jump: Also why repeatedly take write lock inside cancel_one, instead of just taking it once for cancel_tree ?

ripatel-fd: > I don't like this pattern of release lock, then cleanup after, because if the whole application shares a bound on max_live_banks it creates a race window where we can crash here (replay evicts the bank, gets told its now reusable, another thread comes in and tries to grab the 4096th element, and now its not yet freed so app crashes).

This is impossible to reach

ripatel-fd: > Also why repeatedly take write lock inside cancel_one, instead of just taking it once for cancel_tree ?

This is a per-element write lock. It has to be taken for each transaction getting cleaned up

src/flamenco/progcache/fd_progcache_admin.h
ripatel-fd: out of scope for this change, this is existing behavior. the point here was to replace the cache eviction logic. I can do this in another PR

mmcgee-jump (approved):

PR #8767, opened 2026-03-06
snapshots: verify capitalization

cali-jumptrading:


src/discof/restore/utils/fd_ssctrl.h
```diff
@@ -83,19 +84,20 @@
 /* snapin -> snapls */
 /* snapin -> snapwm -> snaplv */
 #define FD_SNAPSHOT_HASH_MSG_EXPECTED         (10UL) /* Hash result sent from snapin to snapls or from snapin to snapwm to snaplv */
+#define FD_SNAPSHOT_MSG_EXP_CAPITALIZATION    (11UL) /* Capitalization sent from snapin to snapwm in vinyl mode to verify capitalization */
```
jvarela-jump, line 87: Just a heads-up: this now requires an update on `fd_ssctrl_msg_ctrl_str`.
We can do so in this PR or a subsequent one.

src/discof/restore/fd_snaplv_tile.c
```diff
@@ -488,7 +521,12 @@ after_credit( fd_snaplv_t *        ctx,
     ctx->hash_accum.awaiting_results  = 0;
     ctx->hash_accum.received_lthashes = 0UL;
 
+    ctx->running_capitalization -= ctx->dup_capitalization;
+    FD_TEST( ctx->running_capitalization>=0L );
```
jvarela-jump: Could we instead do FD_LOG_WARNING here and treat as a failed test?

src/discof/restore/fd_snaplv_tile.c
```diff
@@ -504,13 +542,23 @@ after_credit( fd_snaplv_t *        ctx,
...
+
+    if( FD_UNLIKELY( !capitalization_match ) ) {
+      /* SnapshotError::MismatchedCapitalization
+         TODO: use a tagged release when it becomes available
+         https://github.com/anza-xyz/agave/blob/master/runtime/src/snapshot_bank_utils.rs#L217 */
+      FD_LOG_WARNING(( "snapshot manifest capitalization %lu does not match computed capitalization %lu",
+                       ctx->manifest_capitalization, computed_capitalization ));
+      transition_malformed( ctx, stem );
+      return;
+    }
```
jvarela-jump: I think we should only forward ack_sig to `snapct` on both `lthash` and `capitalization` matching.
Relocating the  `if( FD_UNLIKELY( !capitalization_match ) ) {` block before `if( FD_UNLIKELY( test ) ) {` would be an easy fix.

Issue #3618, opened 2024-12-04
Net tile v2: AF_XDP busy polling


ripatel-fd:
Tracking issue for the net tile rewrite.

Ships the following features:
- [ ] AF_XDP busy polling support (to reduce ksoftirqd) load
- [ ] Removal of xsk_aio
- [ ] Zero copy support

PR #8773, opened 2026-03-07
net: preferred busy polling

tristanx86:
Implements https://github.com/firedancer-io/firedancer/issues/3618

Softirq poll mode is identical to what Firedancer currently uses and is set to be the default poll mode in this PR since it is more well tested. Prefbusy is the new poll mode this PR adds.

Prefbusy poll mode gives Firedancer significantly more control over how and when NAPI is polled, with no networking related IRQs or ksoftirqds during main runtime when used on a well supported driver.

Even when used on a less well supported driver inwhich ksoftirqds still occur like bnxt_en (Broadcom), the poll scheduling code used with prefbusy in this PR have been shown to be beneficial to RX throughput robustness as demonstrated in the test results below.
(bnxt_en Copy Mode)
image
(note TX still degraded towards 0Mpps like in softirq mode)

Prefbusy when well supported, removes Firedancer’s reliance on softirqs and ksoftirqds to poll NAPI. ksoftirqds are separate threads which softirqs execute inside of instead of the net tile's thread once network load is medium-high.

ksoftirqds are bad for a few reasons:
1. They (generally but not always) run on a different core to the net tile so each net tile in softirq poll mode effectively occupies two cores (when under medium-high network load) or if multiple ksoftirqd threads are running on the same core (polling for separate net tiles), only one of those net tiles can poll NAPI at a time.
2. If Linux schedules a ksoftirqd onto a core corresponding to a Firedancer tile, the performance of that tile will significantly degrade as network load increases, this situation is analysed and tested in the mlx5 testing pdf with the net tile being the example choked Firedancer tile.
3. Linux decides how much the ksoftirqd thread runs vs other irq/house keeping threads, creating unwanted dependence on scheduling algorithms designed to prioritise fairness over performance.

[ixgbe PrefBusy Tests.pdf](https://github.com/user-attachments/files/29358809/ixgbe.PrefBusy.Tests.pdf)
[mlx5 PrefBusy Tests.pdf](https://github.com/user-attachments/files/29358810/mlx5.PrefBusy.Tests.pdf)
[i40e PrefBusy Tests.pdf](https://github.com/user-attachments/files/29358811/i40e.PrefBusy.Tests.pdf)

### DIFF
--- a/src/disco/net/fd_net_tile_topo.c
+++ b/src/disco/net/fd_net_tile_topo.c
@@ -45,6 +45,13 @@ setup_xdp_tile( fd_topo_t *             topo,
   tile->xdp.zero_copy           = net_cfg->xdp.xdp_zero_copy;
   fd_cstr_ncpy( tile->xdp.xdp_mode, net_cfg->xdp.xdp_mode, sizeof(tile->xdp.xdp_mode) );
 
+  fd_cstr_ncpy( tile->xdp.poll_mode, net_cfg->xdp.poll_mode, sizeof(tile->xdp.poll_mode) );
+
+  tile->xdp.prefbusy_timebudget_micros   = 80L;
+  tile->xdp.prefbusy_rx_budget           = 64L;
+  tile->xdp.prefbusy_min_interval_nanos  = 5000L;
+  tile->xdp.prefbusy_stall_timeout_nanos = 150000L;
+
   tile->xdp.net.umem_dcache_obj_id = umem_obj->id;
   tile->xdp.netdev_tbl_obj_id      = netlink_tile->netlink.netdev_tbl_obj_id;
   tile->xdp.fib4_main_obj_id       = netlink_tile->netlink.fib4_main_obj_id;
--- a/src/disco/topo/fd_topo.h
+++ b/src/disco/topo/fd_topo.h
@@ -190,6 +190,13 @@ struct fd_topo_tile {
       int    zero_copy;
 
       ulong netdev_tbl_obj_id;
+
+      char poll_mode[ 16 ]; /* "prefbusy" or "softirq" */
+      int prefbusy_timebudget_micros;
+      int prefbusy_rx_budget;
+      long prefbusy_min_interval_nanos;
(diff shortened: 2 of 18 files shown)


tristanx86:
@greptile-jt review

src/disco/topo/fd_topo.h
```diff
@@ -190,6 +190,13 @@ struct fd_topo_tile {
       int    zero_copy;
 
       ulong netdev_tbl_obj_id;
+
+      char poll_mode[ 16 ]; /* "prefbusy" or "softirq" */
+      int prefbusy_timebudget_micros;
+      int prefbusy_rx_budget;
+      long prefbusy_min_interval_nanos;
+      long prefbusy_stall_timeout_nanos;
```
ripatel-fd, line 198: Since this is hardcoded now, can we remove these from `fd_topo.h` and move the default constants to `fd_xdp_tile.c`?

src/disco/net/fd_net_tile_topo.c
```diff
@@ -45,6 +45,13 @@ setup_xdp_tile( fd_topo_t *             topo,
   tile->xdp.zero_copy           = net_cfg->xdp.xdp_zero_copy;
   fd_cstr_ncpy( tile->xdp.xdp_mode, net_cfg->xdp.xdp_mode, sizeof(tile->xdp.xdp_mode) );
 
+  fd_cstr_ncpy( tile->xdp.poll_mode, net_cfg->xdp.poll_mode, sizeof(tile->xdp.poll_mode) );
+
+  tile->xdp.prefbusy_timebudget_micros   = 80L;
+  tile->xdp.prefbusy_rx_budget           = 64L;
+  tile->xdp.prefbusy_min_interval_nanos  = 5000L;
+  tile->xdp.prefbusy_stall_timeout_nanos = 150000L;
```
ripatel-fd, line 53: Once these are moved to `fd_xdp_tile.c` can you add a small comment explaining how you've chosen these defaults? (e.g. trial and error for a box with a certain CPU and NIC)

PR #8778, opened 2026-03-09
feat: enable Frankendancer follower mode for multi-validator clusters

MythicRapha:
## Summary

This PR enables Frankendancer validators to operate in **follower (non-leader) mode** within multi-validator clusters. Currently, a Frankendancer node started as a follower (e.g., via `fdctl` connecting to an `fddev` leader) fails to replay blocks due to two issues in the Agave ledger layer.

### Problem

When running a 2+ node Frankendancer cluster:

1. **TooFewTicks rejection** — The fddev leader produces blocks with fewer ticks than the genesis-configured `hashes_per_tick` expects. The Agave blockstore processor rejects these blocks with `BlockError::TooFewTicks`, causing the follower to fall behind permanently.

2. **Snapshot-to-turbine gap** — When a follower boots from a snapshot, there is a slot gap between the snapshot root and the first slot received via turbine. The Agave replay stage uses `get_slots_since()` which walks the `SlotMeta` parent chain in the blockstore. If that chain is broken (no metadata for intermediate slots), replay never discovers new slots and the follower stalls with `bank_count=0`.

### Solution

Two targeted fixes in the Agave ledger crate (via submodule update):

1. **`blockstore_processor.rs`** — Replace the `TooFewTicks` error with a warning log. This allows followers to replay blocks from leaders that don't fill all tick slots, which is expected behavior for chains with relaxed or custom genesis configurations (L2s, app-chains, devnets).

2. **`blockstore.rs` (`fd_ext_blockstore_insert_shreds`)** — On the first batch of shreds received after startup, detect if there is a gap between the blockstore root and the first shred's slot. If so, create synthetic `SlotMeta` entries to bridge the gap, connecting the parent chain so `get_slots_since()` can walk from root to the new slots. This runs exactly once (guarded by an `AtomicBool`).

### Changes

- `agave/ledger/src/blockstore_processor.rs` — Relax `TooFewTicks` validation (warn instead of error)
- `agave/ledger/src/blockstore.rs` — Add snapshot gap bridging in `fd_ext_blockstore_insert_shreds`
- Agave submodule pointer updated to include both fixes

### Testing

Tested on a production 2-node Frankendancer cluster running v0.812.30108:
- **Leader**: fddev on node S1 (producing blocks at slot 1.8M+)
- **Follower**: fdctl on node S2 (booted from snapshot, successfully replaying via turbine)
- Cluster: [Mythic L2](https://mythic.sh) mainnet

Before these patches, the fdctl follower would:
- Insert shreds into the blockstore but never replay them
- Log `bank_count=0` with no progress
- Eventually stall completely

After the patches, the follower successfully:
- Bridges the snapshot gap on first shred insertion
- Replays all blocks from the leader
- Maintains sync with the leader node

### Notes

- These fixes are specifically for the Firedancer<->Agave interop layer (`fd_ext_blockstore_*` functions)
- The TooFewTicks relaxation only affects block replay, not block production
- The gap bridging is a one-time operation at startup, with negligible performance impact
- Both fixes are safe for standard Solana mainnet validators (the conditions only trigger for non-standard genesis configurations or snapshot boots with gaps)

## Test plan

- [ ] Verify follower node replays blocks from fddev leader
- [ ] Verify no regression on standard single-node fddev startup
- [ ] Verify gap bridging only triggers once (check logs for `fd_ext_blockstore: bridging snapshot gap`)
- [ ] Verify TooFewTicks warning appears in logs instead of error

---

Agave changes: [`MythicFoundation/agave@feat/follower-mode-fixes`](https://github.com/MythicFoundation/agave/tree/feat/follower-mode-fixes)


ptaffet-jump:
These changes seem a bit like hacks that are not fixing the actual problem.

- Changing an error to a warning that's going to be emitted for every block doesn't seem like the right fix.  `hashes_per_tick` does not affect the number of ticks per block, so I'm a bit confused.  The concept of "don't fill all tick slots" doesn't make sense to me either. I don't think we support anything other than 64 ticks per block, but we do support different values of hashes_per_tick (except for low-power mode, which is currently incompatible with how Agave implements it, and is very low on our priority list to fix). Is there a reason you can't use 64 ticks per block?

- Certainly when any Solana validator boots, there's often a gap between the first Turbine slot and the snapshot slot. Those slots are populated via repair. I'm not sure why you need the synthetic `SlotMeta`s.  Can you explain what is different about Mythic that requires this to be changed?

MythicRapha:
Thanks for the review @ptaffet-jump, good questions.

On TooFewTicks: You're right — the number of ticks per block should be 64 regardless of hashes_per_tick. The actual issue we're seeing is thatwhen the fddev leader's blocks are written into the Agave blockstore via fd_ext_blockstore_insert_shreds, the tick structure Agave reconstructs from the shreds doesn't match what verify_ticks() expects (specifically the tick count and PoH hash chain). This is likely a bug in how FD shreds encode tick entries vs how Agave decodes them. Converting to a warning was our workaround to keep the follower running, but you're right that the proper fix is upstream — either in FD's shred encoding or in the blockstore bridge layer. We'll investigate the actual tick mismatch and file a more targeted issue.

On the snapshot gap: After further testing, you're correct that repair does handle this. Our follower successfully received 30K+ shreds/3sec via repair once gossip was configured correctly. The synthetic SlotMeta approach was from an earlier debugging attempt before we had repair working properly. I'll remove that change from the PR.

Happy to close this and re-submit once we've identified the actual tick encoding mismatch. Or if there's a better channel for reporting FD/Agave interop issues, let me know.

PR #8783, opened 2026-03-09
tower: don't error out if tower vote acc dne

ibhatt-jumptrading:
this is the desired behavior because it is possible for a vote account with non-zero t-1 and t-2 stake to be deleted in the t-0 epoch. It is also possible for the same account to be revived in the same epoch that it was deleted in so it doesn't make sense to purge it from the tower_voters

### DIFF
--- a/src/discof/replay/fd_replay_tile.c
+++ b/src/discof/replay/fd_replay_tile.c
@@ -821,7 +821,6 @@ publish_slot_completed( fd_replay_tile_t *  ctx,
   slot_info->bank_hash             = *bank_hash;
   slot_info->block_hash            = *block_hash;
   slot_info->transaction_count     = fd_bank_txn_count_get( bank );
-  slot_info->is_epoch_boundary     = fd_slot_to_epoch( epoch_schedule, fd_bank_parent_slot_get( bank ), NULL )!=epoch;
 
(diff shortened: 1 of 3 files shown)


src/discof/replay/fd_replay_tile.c
```diff
@@ -821,6 +821,7 @@ publish_slot_completed( fd_replay_tile_t *  ctx,
   slot_info->bank_hash             = *bank_hash;
   slot_info->block_hash            = *block_hash;
   slot_info->transaction_count     = fd_bank_txn_count_get( bank );
+  slot_info->is_epoch_boundary     = fd_slot_to_epoch( epoch_schedule, fd_bank_parent_slot_get( bank ), NULL )!=epoch;
```
lidatong: does this handle replaying a minority fork before the boundary? it seems it would end up using the next epoch's stakes

src/discof/replay/fd_replay_tile.h
```diff
@@ -63,6 +63,8 @@ struct fd_replay_slot_completed {
   long last_transaction_finished_nanos;   /* timestamp when replay received the last execution completion */
   long completion_time_nanos;             /* timestamp when replay completed finalizing the slot and notified tower */
 
+  int is_epoch_boundary;
```
lidatong: remove if no longer used

PR #8784, opened 2026-03-09
pack: plumb through max alloc

ptaffet-jump:


src/disco/pack/fd_pack_cost.h
```diff
@@ -160,6 +160,18 @@ FD_STATIC_ASSERT( FD_PACK_MAX_TXN_COST < (ulong)UINT_MAX, fd_pack_max_cost );
...
+   account data that can be allocated in a block (via the creation of
+   new accounts and the growth of existing accounts).  Unfortunately,
+   the way this limit is specified by the protocol is buggy and means
+   that it is neither an upper bound nor a lower bound for what it is
+   supposed to measure.  If USE_TRUE_ALLOC_BOUND is set to 1, this code
+   will compute a true upper bound for the amount the transaction can
+   allocate.  If it is set to 0, it will compute the value in an
+   Agave-compatible way, that is, a value that is normally identical to
+   what Agave produces, but slightly higher in certain cases. */
+#define FD_PACK_COST_USE_TRUE_ALLOC_BOUND 0
```
topointon-jump, line 173: Just curious, what's the rationale for making this toggle-able? Seems like the code might be clearer/simpler if we didn't have this. Not a blocker tho, up to you :)

src/disco/pack/fd_pack_cost.h
```diff
@@ -184,6 +196,8 @@ FD_STATIC_ASSERT( FD_PACK_MAX_TXN_COST < (ulong)UINT_MAX, fd_pack_max_cost );
 #define FD_PACK_MAX_VOTE_COST_PER_BLOCK_UPPER_BOUND ( 36000000UL)
 #define FD_PACK_MAX_WRITE_COST_PER_ACCT_UPPER_BOUND ( FD_PACK_MAX_COST_PER_BLOCK_UPPER_BOUND * 4UL / 10UL ) /* simd 0306 */
 
+#define FD_PACK_MAX_ALLOCATED_DATA_PER_BLOCK        ( 100UL*1000UL*1000UL )
```
topointon-jump, line 200: We should probably add some documentation as to how this is derived, maybe a link to https://github.com/anza-xyz/agave/blob/v4.0.0-beta.1/cost-model/src/block_cost_limits.rs#L39-L41

src/disco/pack/fd_pack_cost.h
```diff
@@ -374,7 +402,114 @@ fd_pack_compute_cost( fd_txn_t const * txn,
...
+          break;
+      }
+    } else {
+      /* This could be the writable accounts only, but this bound is
+         much cheaper, and good enough. */
+      allocated_data += DEFAULT_ALLOC * txn->instr[i].acct_cnt;
+#endif /* FD_PACK_COST_USE_TRUE_ALLOC_BOUND */
+    }
   }
+  allocated_data = fd_ulong_min( allocated_data, 20UL*1024UL*1024UL ); /* MAX_PERMITTED_ACCOUNTS_DATA_ALLOCATIONS_PER_TRANSACTION */
```
topointon-jump, line 519: Might be worth explaining here that if a transaction exceeds `MAX_PERMITTED_ACCOUNTS_DATA_ALLOCATIONS_PER_TRANSACTION` it will be rejected by the runtime, so this is safe. Could be worth linking to https://github.com/anza-xyz/agave/blob/v4.0.0-beta.1/transaction-context/src/transaction_accounts.rs#L335-L341

src/disco/pack/fd_pack_cost.h
```diff
@@ -374,7 +402,114 @@ fd_pack_compute_cost( fd_txn_t const * txn,
...
+          base += 32UL+8UL;     data_sz -= 32UL+8UL;
+          if( FD_UNLIKELY( data_sz<seed_len ) ) break;
+          base += seed_len;     data_sz -= seed_len;
+          if( FD_UNLIKELY( data_sz<8UL ) ) break;
+          allocated_data += fd_ulong_min( MAX_ALLOC, FD_LOAD( ulong, base ) );
+          break;
+        case 13UL: /* create_account_allow_prefund */
+          if( FD_UNLIKELY( data_sz<8UL+8UL ) ) break;
+          allocated_data += fd_ulong_min( MAX_ALLOC, FD_LOAD( ulong, base+8UL ) );
+          break;
```
topointon-jump, line 472: Agave charges 0 for this until the `create_account_allow_prefund` feature which enables this instruction is activated, but this is fine because the conservative thing to do is to reserve extra. Even so, might be worth adding a comment.

src/disco/pack/fd_pack.c
```diff
@@ -2061,13 +2100,14 @@ fd_pack_schedule_impl( fd_pack_t          * pack,
     if( FD_UNLIKELY( (cu_limit<smallest_in_treap->cus) | (txn_limit==0UL) | (byte_limit<smallest_in_treap->bytes) ) ) break;
```
topointon-jump, line 2100: Should we also skip here if the `alloc_limit` is exceeded? We'd need to track it separately in `smallest_in_treap`

src/disco/pack/fd_pack_cost.h
ptaffet-jump: I just didn't want to throw away all our hard work to find all the alloc cases. I can also just save it in a branch somewhere though

src/disco/pack/fd_pack_cost.h
ptaffet-jump: The behavior is not quite what you might think, but I added a comment.

src/disco/pack/fd_pack_cost.h
ptaffet-jump: Good call.

src/disco/pack/fd_pack.c
ptaffet-jump: It just saves computation, but I think the case is so rare in practice that it's not worth tracking.  Good thought though.

src/disco/pack/fd_pack_cost.h
ptaffet-jump: Done. Thanks.

PR #8789, opened 2026-03-10
repair: implement rserve

drubin-fd:
TODO:
- [x] Handle and implement all necessary repair requests
- [ ] Consider the bursts more carefully
- [x] Add metrics
- [x] Clean up the code! Many TODOs left atm.


src/disco/store/fd_ledger.c
```diff
@@ -0,0 +1,386 @@
+#include "fd_ledger.h"
```
topointon-jump: Small nit (and ignore me if you've already discussed with someone else) - imo "ledger" is quite overloaded. I wonder if it might be better to call this something more specific?

We previously had an "fd_ledger" in the codebase which was our equivalent of the Agave ledger tool, and this might get confused with that. A "ledger" in Agave also usually includes the snapshots.

drubin-fd: I was talking about this with @lidatong, I don't have any strong opinions on the naming. It was the most obvious name that came to mind, we can call it something like "shred archive". Just need a satisfying way to compress that into 6 letters :).

src/discof/repair/fd_rserve_tile.seccomppolicy
```diff
@@ -0,0 +1,20 @@
+# logfile_fd: It can be disabled by configuration, but typically tiles
+#             will open a log file on boot and write all messages there.
+unsigned int logfile_fd
+
+# logging: all log messages are written to a file and/or pipe
+#
+# 'WARNING' and above are written to the STDERR pipe, while all messages
+# are always written to the log file.
+#
+# good_peer_cache_file: we periodically write out the good peers to the file
```
mmcgee-jump: Unused

src/disco/store/fd_ledger.c
mmcgee-jump: Why does it need to be 6 letters? I'd go with

`rprsrv` for the tile name and `shredstore` for this component? Then name the file by default `shreds.db` or similar?

src/app/firedancer/config/default.toml
```diff
@@ -199,6 +199,12 @@ telemetry = true
     # Firedancer instance.
     accounts = ""
 
+    # TODO: explain
+    #
+    # If no path is provided, defaults to the `ledger.db` path within
+    # the base directory above.
+    ledger = ""
```
mmcgee-jump: Absolutely do not call this ledger, lol. Something like `shreds` or `shredstore`

src/app/firedancer/config/default.toml
```diff
@@ -199,6 +199,12 @@ telemetry = true
...
+
```
mmcgee-jump, line 214: I think the natural configuration for this is something like:

```
[tiles.rprsrv]
    # The maximum size of the shred storage on disk.  Determines the
    # amount of history that the validator will be able to serve to
    # peers making repair requests.  Roughly, for mainnet, 1 GiB of
    # storage corresponds to ~ xxx slots, and the default is set to
    # match Agave's ability to serve around six hours of shreds.
    #
    # This is an upper bound on the disk space used, but when first
    # booted the validator will start with just 1 GiB, and incrementally
    # grow the storage file on disk up to the limit as needed.
    shred_storage_limit_gib = 50 # TODO: Correct value?
```

src/app/firedancer/config/default.toml
```diff
@@ -1459,6 +1464,10 @@ telemetry = true
         # Must be power of 2
         slot_max = 4096
 
+    [tiles.rserve]
+        repair_serve_listen_port = 8702
+        max_slots = 2000 # TODO: increase to epoch?
```
mmcgee-jump: See above comment, this should be specified in GiB imo, for operators to understand. The agave way of doing this (`DEFAULT_MAX_LEDGER_SHREDS`) is extremely bad, but slots isn't a whole lot better. Disk is the resource we are bounding use of here, not "slots".

src/app/firedancer/topology.c
```diff
@@ -924,6 +938,8 @@ fd_topo_initialize( config_t * config ) {
   /**/                 fd_topob_tile_in(    topo, "repair",  0UL,          "metric_in", "tower_out",     0UL,          FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
   FOR(shred_tile_cnt)  fd_topob_tile_out(   topo, "repair",  0UL,                       "repair_shred",  i                                                  );
 
+  FOR(shred_tile_cnt)  fd_topob_tile_in(    topo, "rserve",  0UL,          "metric_in", "shred_out",     i,            FD_TOPOB_RELIABLE,   FD_TOPOB_POLLED );
```
mmcgee-jump: I believe this doesn't include shreds that we produced when we were leader, which we definitely need to serve. Check with Philip if true, and how to fix.

mmcgee-jump (changes requested):
some high level notes

src/disco/store/fd_ledger.c
drubin-fd: > Why does it need to be 6 letters?

I was told it should be 6 characters :).

I'm happy with whatever names, just went with this to not get hung up on naming. 
I 100% agree that ledger is not a good name.

I kinda like rserve though. Lmk what names, I should not be the one to provide feedback on it, as it requires previous context.

mmcgee-jump: The tile name has to be 6 chars for compact display in GUI etc, but anything else you are free to name as long as you like.

src/app/firedancer/config/default.toml
drubin-fd: Yeah I agree with this. I had just made it slots because I wanted to directly be able to change the size of the maps for testing, but a space-based configuration makes sense.

I need to do some snooping around mainnet to see what the average shreds per slot is (was assuming ~1k).

Regarding that last comment, would you expect the shredstore to resize itself as it is filled? The approach I took was to just size it at the start with an ftruncate() to the maximum amount it would need. The usecase for shredstore in validator running for longer than like 6 hours becomes a FIFO ring buffer, so I think it makes sense to just have the full size right away.

I guess I could make it check if it is close to the maximum capacity and then superlinearly ftruncate it to larger sizes. Don't see why though.

src/app/firedancer/topology.c
drubin-fd: Great point, didn't consider that. I will investigate. Not sure how best to test it though, but I will cross that bridge when I get there.

src/app/firedancer/config/default.toml
mmcgee-jump: Starting small and growing lets the file naturally scale to the cluster, and support development, testing, security audit and other scenarios without needing to be reconfigured. Kinda just removes an additional thing to think about.

src/disco/store/fd_ledger.c
lidatong: there's already a fd_store, would call it `shredb`

src/discof/repair/fd_rserve_tile.seccomppolicy
```diff
@@ -0,0 +1,20 @@
+# logfile_fd: It can be disabled by configuration, but typically tiles
+#             will open a log file on boot and write all messages there.
+unsigned int logfile_fd
```
lidatong: you'll also need a file descriptor for the actual db file, as well as appropriate syscalls. `open` should be called in `privileged_init` so you won't need that syscall

lidatong (commented):
nice work getting a PoC of this up quickly!

src/discof/repair/fd_repair_tile.c
```diff
@@ -574,7 +571,7 @@ after_sign( ctx_t             * ctx,
   /* Thhis is a pong message */
   if( FD_UNLIKELY( pending->msg.kind == FD_REPAIR_KIND_PONG ) ) {
     fd_memcpy( pending->msg.pong.sig, ctx->buffer, 64UL );
-    send_packet( ctx, stem, 1, pending->pong_data.peer_addr.addr, pending->pong_data.peer_addr.port, pending->pong_data.daddr, pending->buf, fd_repair_sz( &pending->msg ), fd_frag_meta_ts_comp( fd_tickcount() ) );
+    send_packet( ctx, stem, pending->pong_data.peer_addr.addr, pending->pong_data.peer_addr.port, pending->pong_data.daddr, pending->buf, fd_repair_sz( &pending->msg ), fd_frag_meta_ts_comp( fd_tickcount() ) );
```
lidatong, line 697: thanks for cleaning this up

src/discof/repair/fd_rserve.h
```diff
@@ -0,0 +1,58 @@
...
+
+typedef struct {
+  fd_pubkey_t addr;      /* The public key of the validator which sent the ping. */
+  uint        hash;      /* reserved for use by fd_map */
+  ulong       timestamp; /* The time at which the ping was received. */
+} ping_cache_entry_t;
+
+#define MAP_NAME               ping_cache
+#define MAP_T                  ping_cache_entry_t
+#define MAP_LG_SLOT_CNT        (16UL) /* 1<<16 = 65k */
```
lidatong: probably should not be hardcoded like this and configurable by `new` (use fd_map_dynamic)

src/discof/repair/fd_rserve.h
```diff
@@ -0,0 +1,58 @@
...
+   the existing set of "good" nodes we'd need to know about from Gossip.
+   A ping-cache is a simple solution for the time being. */
+
+/* In seconds for now, TODO: change */
+#define FD_RSERVE_PING_CACHE_TTL 1280UL
+
+typedef struct {
+  fd_pubkey_t addr;      /* The public key of the validator which sent the ping. */
+  uint        hash;      /* reserved for use by fd_map */
+  ulong       timestamp; /* The time at which the ping was received. */
```
lidatong: add unit

src/discof/repair/fd_rserve_tile.seccomppolicy
drubin-fd: Yeah. Didn't need them for the static-size, but added them locally while implementing the growing file version :).

src/discof/repair/fd_rserve_tile.c
```diff
@@ -0,0 +1,481 @@
...
+//   hdr->udp->check = 0U;
+
+//   ulong tspub     = fd_frag_meta_ts_comp( fd_tickcount() );
+//   ulong sig       = fd_disco_netmux_sig( dst_ip_addr, dst_port, dst_ip_addr, DST_PROTO_OUTGOING, sizeof(fd_ip4_udp_hdrs_t) );
+//   ulong packet_sz = payload_sz + sizeof(fd_ip4_udp_hdrs_t);
+//   ulong chunk     = ctx->net_out_chunk;
+//   fd_stem_publish( stem, ctx->net_out_idx, sig, chunk, packet_sz, 0UL, tsorig, tspub );
+//   ctx->net_out_chunk = fd_dcache_compact_next( chunk, packet_sz, ctx->net_out_chunk0, ctx->net_out_wmark );
+// }
+
```
lidatong, line 162: you can probably use the newer `returnable_frag` pattern than `before / during / after` (those predate returnable, and only really used when requires special casing in before / during -- otherwise it's in some tiles as a legacy that hasn't been cleaned up)

src/discof/repair/fd_rserve_tile.c
```diff
@@ -0,0 +1,481 @@
...
+        fd_memcpy( offsets, set->block_offs, sizeof(uint)*32 );
+        fd_memcpy( data,    set->data,       set->data_sz    );
+      } FD_STORE_SLOCK_END;
+
+      FD_LOG_DEBUG(( "got final shred, slot=%lu, shred_idx=%u", slot, shred->idx ));
+
+      /* Insert the last shred header, which we have just received. */
+      fd_ledger_insert_header( ctx->ledger, ctx->buffer );
+
+      for( uint i=0; i<32; i++ ) {
```
lidatong:
```suggestion
      for( uint i=0; i<FD_FEC_SHRED_CNT; i++ ) {
```

src/app/firedancer/topology.c
drubin-fd: My current conclusion after some dicussions with Philip are that leader shreds are already being sent through shred_out. I am infering this as when I run a local cluster, the leader shreds that we produce are inserted into the ledger.

We may still refactor the shred_out link to include the full payload size, but that is not necassary for a v1.0 repair server. The main limitation of the current design is that we cannot serve shreds from a FEC set that is not fully recovered.

src/disco/store/fd_ledger.h
```diff
@@ -0,0 +1,169 @@
...
+}
+
+FD_FN_CONST static inline ulong
+fd_ledger_key_slot( ulong key ) {
+  return key >> 16;
+}
+
+FD_FN_CONST static inline uint
+fd_ledger_key_shred_idx( ulong key ) {
+  return (uint)(key & 0xFFFFUL);
```
lidatong: fyi `fd_bits` has helpers for these bitwise ops eg. see `fd_ulong_extract_bits`

src/disco/store/fd_ledger.h
```diff
@@ -0,0 +1,169 @@
...
+                          ulong         payload_sz,
+                          ulong         slot,
+                          uint          shred_idx );
+
+/* Inserts a FD_SHRED_DATA_HEADER_SZ sized header into a shred's entry.
+   Leaves the payload part of the entry untouched.
+   Will derive the (slot, shred_idx) from the header itself. */
+void
+fd_ledger_insert_header( fd_ledger_t * ledger,
+                         uchar const * header );
```
lidatong: can prob use `fd_shred_t *` type here?

src/disco/store/fd_ledger.h
```diff
@@ -0,0 +1,169 @@
...
+
+/* Given a (slot, shred_index), returns the corresponding entry.
+   If no entry was found, returns -1, otherwise returns the amount
+   of bytes written to out. */
+int
+fd_ledger_query( fd_ledger_t * ledger,
+                 uchar         out[ FD_SHRED_MAX_SZ ],
```
lidatong: usually out parameter is the last one by convention. also i think you only store data shreds which should all be FD_SHRED_MIN_SZ

src/discof/repair/fd_rserve_tile.c
drubin-fd: Understood. Will rework the loop to use that.

src/disco/store/fd_ledger.c
drubin-fd: I like shredb, going with that for now. Again, no opinion it honestly.

src/discof/repair/fd_rserve.h
drubin-fd: Specified it is in nanoseconds.

src/discof/repair/fd_rserve_tile.c
drubin-fd: Reworked it to use only returnable_frag.

src/discof/repair/fd_rserve.h
drubin-fd: Do you think this should be a user configurable option? Or do you just want it to be configurable from the internal API?

lidatong: good clarification: probably internal API, derived from an existing limit to avoid spamming operator configuration options

src/app/firedancer/config/default.toml
drubin-fd: The size is now bound by a configuration option provided in GiB, and before that it grows superlinearly (2x the previous size). The metadata hashmaps are sized to the worst-case of the maximum possible size of the ledger, given the option provided by the operator.

src/app/firedancer/config/default.toml
```diff
@@ -1486,6 +1491,22 @@ telemetry = true
         # Must be power of 2
         slot_max = 4096
 
+    [tiles.rserve]
+        enabled = true
```
mmcgee-jump, line 1454: Please document all options. What does it do / mean to disable the repair server? Should an operator ever do this? (probably no, unless RPC or something)

src/app/firedancer/config/default.toml
```diff
@@ -1486,6 +1491,22 @@ telemetry = true
...
+        repair_serve_listen_port = 8702
+        # The maximum size of the shred storage on disk.  Determines the
+        # amount of history that the validator will be able to serve to
+        # peers making repair requests.  Roughly, for mainnet, 1 GiB of
+        # storage corresponds to ~1000 slots, and the default is set to
+        # match Agave's ability to serve around six hours of shreds.
+        #
+        # This is an upper bound on the disk space used, but when first
+        # booted the validator will start with just ~40MiB, and incrementally
+        # grow the storage file on disk up to the limit as needed. Once the
```
```suggestion
        # grow the storage file on disk up to the limit as needed.  Once the
```

src/disco/store/test_shredb.c
```diff
@@ -0,0 +1,407 @@
...
+  }
+}
+
+static fd_shredb_t *
+setup_store( void ** out_mem, ulong max_size_gib ) {
+  ulong footprint = fd_shredb_footprint( max_size_gib );
+  FD_TEST( footprint );
+  void * mem = aligned_alloc( fd_shredb_align(), footprint );
+  FD_TEST( mem );
+  memset( mem, 0, footprint );
```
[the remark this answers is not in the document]
lidatong: copilot is right about this one, i think we agreed fd_memset for runtime sz, memset for compile-time sz

src/discof/repair/fd_rserve_tile.c
```diff
@@ -0,0 +1,654 @@
...
+  fd_stem_publish( stem, ctx->net_out_idx, sig, chunk, packet_sz, 0UL, tsorig, tspub );
+  ctx->net_out_chunk = fd_dcache_compact_next( chunk, packet_sz, ctx->net_out_chunk0, ctx->net_out_wmark );
+}
+
+static inline void
+handle_pong( ctx_t              * ctx,
+             uchar const        * payload,
+             ulong                payload_sz,
+             ulong                tsorig ) {
+  if( FD_UNLIKELY( payload_sz!=132UL ) ) return;
```
lidatong: `sizeof(fd_repair_pong_t)` (see also comment about subtracting discriminant)

src/discof/repair/fd_repair.h
```diff
@@ -77,11 +77,12 @@
...
-#define FD_REPAIR_KIND_PONG          (7U)
-#define FD_REPAIR_KIND_SHRED         (8U)
-#define FD_REPAIR_KIND_HIGHEST_SHRED (9U)
-#define FD_REPAIR_KIND_ORPHAN        (10U)
+#define FD_REPAIR_KIND_PING            (0U)
+#define FD_REPAIR_KIND_PONG            (7U)
+#define FD_REPAIR_KIND_SHRED           (8U)
+#define FD_REPAIR_KIND_HIGHEST_SHRED   (9U)
+#define FD_REPAIR_KIND_ORPHAN          (10U)
+#define FD_REPAIR_KIND_ANCESTOR_HASHES (11U)
```
lidatong, line 76: remove this constant if its not implemented and just add a TODO

src/discof/repair/fd_repair.c
```diff
@@ -76,8 +76,8 @@ fd_repair_delete( void * repair ) {
 fd_repair_msg_t *
 fd_repair_pong( fd_repair_t * repair, fd_hash_t * ping_token ) {
   uchar pre_image[FD_REPAIR_PONG_PREIMAGE_SZ];
-  memcpy( pre_image, "SOLANA_PING_PONG", 16UL );
-  memcpy( pre_image+16UL, ping_token->uc, 32UL);
+  fd_memcpy( pre_image,      "SOLANA_PING_PONG", 16UL );
+  fd_memcpy( pre_image+16UL, ping_token->uc,     32UL );
```
lidatong: intentionally uses `memcpy`, the agreement at some point was memcpy for compile-time sz  and fd_memcpy for runtime

src/discof/repair/fd_rserve_tile.c
```diff
@@ -0,0 +1,654 @@
...
+
+  /* The Window, HighestWindow, and Orphan requests all start with:
+     - tag         (4 bytes)
+     - signature   (64 bytes)
+     - from pubkey (32 bytes)
+     - to pubkey   (32 bytes)
+     - timestamp   (8 bytes)
+     - nonce       (4 bytes)
+     which adds up to at least 144 bytes. */
+  if( FD_UNLIKELY( payload_sz<144UL ) ) return;
```
lidatong: good comment and i would keep it, but recommend using `sizeof` on the structs since you are casting anyway

src/discof/repair/fd_rserve_tile.c
```diff
@@ -0,0 +1,654 @@
...
+  fd_repair_msg_t const * msg = (fd_repair_msg_t const *)payload;
```
lidatong: it's technically a strict aliasing violation per C std, though unlikely to be abused by a compiler we by convention always type pun these

```suggestion
  fd_repair_msg_t const * msg = (fd_repair_msg_t const *)fd_type_pun_const( payload );
```

src/discof/repair/fd_rserve_tile.c
```diff
@@ -0,0 +1,654 @@
...
+static inline void
+handle_net_request( ctx_t             * ctx,
+                    fd_stem_context_t * stem,
+                    uchar const       * payload,
+                    ulong               payload_sz,
+                    fd_udp_hdr_t      * udp,
+                    fd_ip4_hdr_t      * ip4,
+                    ulong               tsorig ) {
+  if( FD_UNLIKELY( payload_sz<4UL ) ) return;
+  uint tag = FD_LOAD( uint, payload );
```
lidatong, line 227: this is fine but also pointing out we have `fd_uint_load_4_fast`

src/discof/repair/fd_rserve_tile.c
lidatong: i'm also not sure this cast is safe -- i believe the compiler can assume pointer now points to at least sizeof the largest member of the union?

it might be better to just load the UINT, then `switch` on the kind similar to what you do below, but earlier where you return early if the sz is wrong (using `sizeof(specific_struct_member_for_kind)`)

src/discof/repair/fd_rserve_tile.c
```diff
@@ -0,0 +1,654 @@
...
+  if( FD_UNLIKELY( FD_ED25519_SUCCESS!=fd_ed25519_verify( signable, signable_sz, header->sig, header->from.uc, ctx->sha512 ) ) ) {
+    /* Invalid signature, ignore. */
+    ctx->metrics->fail_sigverify_request++;
+    return;
+  }
+
+  /* Check whether we've heard a pong response from them. */
+  ping_cache_entry_t * entry = ping_cache_query( ctx->rserve->ping_cache, header->from, NULL );
+  if( FD_LIKELY( entry ) ) {
+    switch( tag ) {
```
lidatong, line 299: pattern matching is the language feature i miss the most in C 🥲

lidatong (commented):
i think a good next step would be determining how much request traffic we can handle from a single repair client trying to DoS us from serving other clients and whether we even need some kind of QoS

if we can handle a more than reasonable amount of request traffic, then i don't think we need that

src/discof/repair/fd_rserve_tile.c
```diff
@@ -0,0 +1,654 @@
...
+     previous rotating token.  This prevents replayed or stale pongs. */
+  if( FD_UNLIKELY( !fd_rserve_pong_token_verify( ctx->rserve, request->hash.uc ) ) ) {
+    return;
+  }
+
+  ping_cache_entry_t * entry = ping_cache_query( ctx->rserve->ping_cache, request->from, NULL );
+  /* If it doesn't exist yet, create the entry. */
+  if( FD_LIKELY( !entry ) ) {
+    ctx->metrics->ping_cache_entries++;
+    entry = ping_cache_insert( ctx->rserve->ping_cache, request->from );
```
emwang-jump: haven't followed the full pinged -> ponged -> cache_insert flow but do we need to be concerned about ping_cache having no available slots? might need to check capacity / preemptive evict

drubin-fd: yeah, fixed by doing LRU on the ping cache

src/discof/repair/fd_rserve_tile.c
```diff
@@ -0,0 +1,593 @@
...
+static inline void
+handle_pong( ctx_t              * ctx,
+             uchar const        * payload,
+             ulong                payload_sz ) {
+  if( FD_UNLIKELY( payload_sz!=sizeof(uint)+sizeof(fd_repair_pong_t) ) ) return;
+  fd_repair_msg_t const * msg = (fd_repair_msg_t const *)fd_type_pun_const( payload );
+  fd_repair_pong_t const * request = &msg->pong;
+
+  if( FD_UNLIKELY( fd_pubkey_eq( &ctx->identity_public_key, &request->from ) ) ) {
+    /* We've received our own repair request, ignore. */
```
mmcgee-jump, line 172: For metrics here, I like to have a mutually exclusive and exhaustive set of outcomes for the packet, e.g.

<enum PongRxOutcome>
   <loopback> /* our own key */
  <sigverify_failure>
  <invalid_token>
  <success>
<enum>

This is very useful for tracking what happened to packets

src/discof/repair/fd_rserve_tile.c
```diff
@@ -0,0 +1,593 @@
...
+  if( FD_UNLIKELY( !fd_rserve_pong_token_verify( ctx->rserve, request->hash.uc ) ) ) {
+    return;
+  }
+
+  fd_rserve_t * rserve = ctx->rserve;
+  ping_cache_entry_t * entry = ping_map_ele_query( rserve->ping_map, &request->from, NULL, rserve->ping_pool );
+
+  if( FD_LIKELY( !entry ) ) {
+    /* New entry, evict LRU if pool is full. */
+    if( FD_UNLIKELY( !ping_pool_free( rserve->ping_pool ) ) ) {
```
mmcgee-jump, line 195: We always want a metric counter for evictions

src/disco/store/fd_shredb.c
```diff
@@ -0,0 +1,302 @@
...
+                  ulong              shred_sz ) {
+  ulong slot      = shred->slot;
+  uint  shred_idx = shred->idx;
+
+  FD_LOG_DEBUG(( "inserting shred into store (slot=%lu, shred_idx=%u)", slot, shred_idx ));
+
+  ulong key = fd_shredb_key_pack( slot, shred_idx );
+  if( fd_shredb_shred_map_query( store->shred_map, key, NULL ) ) return;
+
+  /* Grow the backing file if the write head has reached the current
```
mmcgee-jump, line 214: metrics for backing file size / utilization please, for admin to monitor

mmcgee-jump (dismissed):
Nice, pretty complete. Can you add more metrics though, and a `watch.c` line with some topline metrics (rps, bandwidth, iops, file size, ...)

src/disco/store/fd_shredb.c
```diff
@@ -0,0 +1,302 @@
...
+
+  fd_shredb_entry_t * ring  = fd_shredb_ring( store );
+  fd_shredb_entry_t * entry = &ring[ map_entry->ring_idx ];
+
+  fd_memcpy( out, entry->shred, entry->shred_sz );
+  return entry->shred_sz;
+}
+
+int fd_shredb_query_highest( fd_shredb_t * store,
+                             ulong         slot,
```
ripatel-fd, line 284: return val on different line

ripatel-fd (approved):

PR #8791, opened 2026-03-10
vote-stakes: fix idx assignment

two-heart:


ibhatt-jumptrading:
this gets fixed in #8740

PR #8792, opened 2026-03-10
codeql: detect self assignments

two-heart:
Finds bugs like b64e1d3b1 and 8c960ff77 exactly

### DIFF
--- a/contrib/codeql/nightly/SelfAssignment.ql
+++ b/contrib/codeql/nightly/SelfAssignment.ql
@@ -1,0 +1,21 @@
+/**
+ * @name Self Assignment
+ * @description Detects assignments of a value to itself
+ * @kind problem
+ * @problem.severity warning
+ * @precision high
+ * @id asymmetric-research/self-assign
+ */
+
+import cpp
+import semmle.code.cpp.valuenumbering.HashCons
+import filter
+
+from AssignExpr assign
+where hashCons(assign.getLValue()) = hashCons(assign.getRValue())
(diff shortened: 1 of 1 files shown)


contrib/codeql/nightly/SelfAssignment.ql
```diff
@@ -0,0 +1,21 @@
...
+import semmle.code.cpp.valuenumbering.HashCons
+import filter
+
+from AssignExpr assign
+where hashCons(assign.getLValue()) = hashCons(assign.getRValue())
+/* FD_TXN_ERR_FOR_LOG_INSTR is an example of how a self assignment can
+be a valid result of a macro expansion */
+and not assign.isInMacroExpansion()
+and included(assign.getLocation())
+select assign, "Self Assignment"
```
intrigus-lgtm, line 20:
```suggestion
where
  hashCons(assign.getLValue()) = hashCons(assign.getRValue()) and
  /* FD_TXN_ERR_FOR_LOG_INSTR is an example of how a self assignment can be a valid result of a macro expansion */
  not exists(Macro m |
    m.hasName("FD_TXN_ERR_FOR_LOG_INSTR") and m.getAnInvocation().getAnExpandedElement() = assign
  ) and
  included(assign.getLocation())
select assign, "Self Assignment"
```

Tiny precision improvement.

two-heart: ig there can be other macros in the future where this is intended

PR #8802, opened 2026-03-10
watch: add node info row

mmcgee-jump:


ripatel-fd:
Taking this over, thanks @mmcgee-jump 🙌

PR #8803, opened 2026-03-10
reasm: flexible eviction from mid-tree, fix out queue

emwang-jump:


src/discof/reasm/fd_reasm.c
```diff
@@ -11,9 +11,8 @@ fd_reasm_align( void ) {
 
 FD_FN_CONST ulong
 fd_reasm_footprint( ulong fec_max ) {
-  ulong max_fec_per_slot = FD_SHRED_BLK_MAX / 32;                             /* untrue until fix-32. TODO probably replace with macro after fix-32 is active */
-  ulong max_slots        = fd_ulong_max(fec_max / max_fec_per_slot, 1UL);     /* add capacity for a block id per slot */
-  int lgf_max = fd_ulong_find_msb( fd_ulong_pow2_up( fec_max + max_slots ) ); /* capacity for fec_max fecs + (fec_max / 1024) more block ids */
+  ulong max_slots = fd_ulong_max(fec_max / FD_FEC_BLK_MAX, 1UL);                 /* add capacity for a block id per slot */
+  int lgf_max     = fd_ulong_find_msb( fd_ulong_pow2_up( fec_max + max_slots ) ); /* capacity for fec_max fecs + (fec_max / 1024) more block ids */
```
lidatong, line 15: hm ... i think this is more complicated than needs -- just doing the standard `fd_ulong_find_msb( fd_ulong_pow2_up( fec_max ) ) + 1` should be enough enough

since `(n + n/1024) / 2n ~= 0.5 fill ratio`

src/discof/reasm/fd_reasm.c
```diff
@@ -66,9 +63,8 @@ fd_reasm_new( void * shmem,
 
   fd_memset( shmem, 0, footprint );
 
-  ulong max_fec_per_slot = FD_SHRED_BLK_MAX / 32;                             /* untrue until fix-32. TODO probably replace with macro after fix-32 is active */
-  ulong max_slots        = fd_ulong_max(fec_max / max_fec_per_slot, 1UL);     /* add capacity for a block id per slot */
-  int   lgf_max = fd_ulong_find_msb( fd_ulong_pow2_up( fec_max + max_slots ) ); /* capacity for fec_max fecs + (fec_max / 1024) more block ids */
+  ulong max_slots = fd_ulong_max(fec_max / FD_FEC_BLK_MAX, 1UL);                  /* add capacity for a block id per slot */
```
lidatong, line 66: see comment above

src/discof/reasm/test_reasm.c
```diff
@@ -553,6 +558,393 @@ test_evict( fd_wksp_t * wksp ) {
...
+       root -> A -> B -> C -> D -> E
+     A and B are inserted and popped (non-eqvoc initially).
+     Then equivocating versions A', B' arrive, marking the originals eqvoc.
+     C is inserted (connected to B, eqvoc propagated from B).
+     D is inserted (connected to C, eqvoc propagated).
+     E is inserted (connected to D, eqvoc propagated).
+     All of C, D, E are in out but can't be popped (eqvoc, unconfirmed).
+     Confirm E: walks E -> D -> C -> B -> A.  B and A are already popped.
+     C, D, E should become confirmed.
+     Pop should return C, D, E in that order. */
```
lidatong, line 741: 🤖

src/discof/reasm/fd_reasm.c
```diff
@@ -388,15 +399,53 @@ fd_reasm_remove( fd_reasm_t     * reasm,
...
+      FD_TEST( head->bank_idx == tail->bank_idx );
+
       if( FD_UNLIKELY( head->fec_set_idx==0  ) )                   break;
       if( FD_UNLIKELY( head->sibling != pool_idx_null( pool ) ) )  break; /* if the parent has more than 1 child, we know for sure the parent is a slot boundary or eqvoc point, so we can stop here. */
+      if( FD_UNLIKELY( parent->child != pool_idx( pool, head ) ) ) break; /* if the parent has more than 1 child, we know for sure the parent is a slot boundary or eqvoc point, so we can stop here. */
       if( FD_UNLIKELY( parent->slot_complete ) )                   break; /* specifically catches case where slot complete is in middle of slot. */
       head = parent;
     }
+  } else {
+    FD_TEST( head->child == ULONG_MAX ); /* must be a leaf node */
```
lidatong: this assertion is kinda lost by putting it in an `else` block after a really big `if` block

would be more clear with an assertion at the top that "in orphaned / ancestry => must be leaf"

src/discof/replay/fd_replay_tile.c
```diff
@@ -2302,15 +2249,19 @@ after_credit( fd_replay_tile_t *  ctx,
...
-      FD_TEST( fec && fec->child == ULONG_MAX );
-      ctx->reasm_evicted = fd_reasm_remove( ctx->reasm, fec, ctx->store );
+      FD_TEST( fec );
+      fd_reasm_fec_t * evicted_head = fd_reasm_remove( ctx->reasm, fec, ctx->store );
+      if( FD_UNLIKELY( ctx->reasm_evicted ) ) {
+        /* already have a chain we are evicting. Prepend the new chain to the existing chain */
+        fec->child = fd_reasm_pool_idx( ctx->reasm, ctx->reasm_evicted );
+        ctx->reasm_evicted = evicted_head;
+      } else {
+        ctx->reasm_evicted = evicted_head;
```
lidatong: i'm confused by this -- it's prepending but it's putting it on the `child` pointer of the existing thing, but then setting the `ctx` pointer to the new thing?

src/discof/reasm/fd_reasm.c
emwang-jump: The `n/1024` is specifically because we put the block ids in the xid map as well, so I'm making it clear here that we basically need 1025 fecs per slot, or basically `fec_max + blk_max` capacity.  It is the same effectively because the ulong_pow2_up should give us the same fill ratio as the standard

src/discof/replay/fd_replay_tile.c
```diff
@@ -2302,15 +2249,19 @@ after_credit( fd_replay_tile_t *  ctx,
...
       fd_reasm_fec_t * fec = fd_reasm_query( ctx->reasm, &block_id_ele->latest_mr );
-      FD_TEST( fec && fec->child == ULONG_MAX );
-      ctx->reasm_evicted = fd_reasm_remove( ctx->reasm, fec, ctx->store );
+      FD_TEST( fec );
+      fd_reasm_fec_t * evicted_head = fd_reasm_remove( ctx->reasm, fec, ctx->store );
+      if( FD_UNLIKELY( ctx->reasm_evicted ) ) {
+        /* already have a chain we are evicting. Prepend the new chain to the existing chain */
+        fec->child = fd_reasm_pool_idx( ctx->reasm, ctx->reasm_evicted );
+        ctx->reasm_evicted = evicted_head;
+      } else {
```
lidatong: btw code can also be cleaned up a bit eg.

```c
if... prepend /* comment */
ctx->reasm_evicted /* no else */

src/discof/reasm/fd_reasm.c
lidatong: right i remembered that. ok fair to make it explicit

src/discof/replay/fd_replay_tile.c
emwang-jump: the evicted chain is effectively a list with head `ctx->reasm_evicted`, but the next pointer is `child`.  If we loop and evict multiple banks in this after_credit, the `ctx->reasm_evicted` would get overwritten each time. So instead we can set the tail of what we just evicted (`fec`) to point to the head of the evict list, and then repoint `ctx->reasm_evicted` to the new head of the evict list, which is the returned  `evicted_head`. 
this way we can interleave the after_credit publishing to repair / popping things off the list with evicting stuff

lidatong: ah. i was confused by `evicted_head`, since that traverses up to the ancestor

PR #8804, opened 2026-03-10
pack: throw out transactions that violate limit_instruction_accounts

topointon-jump:
`limit_instruction_accounts` limits the number of accounts an instruction can reference to 255. Any transactions that violate this limit will be invalid and cannot be included in a block.

To avoid feature-gating Pack, we prematurely always throw out transactions that violate this limit. This check can be moved into the transaction parser when `limit_instruction_accounts` is activated everywhere. 

See https://github.com/firedancer-io/firedancer/pull/8723 for the runtime implementation of this feature.


ptaffet-jump (approved):
Great!

PR #8805, opened 2026-03-11
bls: reject compressed or parity encodings while deserializing

drubin-fd:
https://github.com/anza-xyz/agave/pull/11155

Maybe don't merge yet, Anza may rekey the bls syscall gate to keep the -beta.0 min version, TBD.


zz-sol:
LGTM

PR #8812, opened 2026-03-11
disco, flamenco, shred: implement discard_unexpected_data_complete_shreds

topointon-jump:
Frankendancer change (adding `discard_unexpected_data_complete_shreds` to the features of interest) will be PR'd into main later to avoid conflicts. Without the Frankendancer change this PR is safe, the feature will just always be deactivated on Frankendancer.

### DIFF
--- a/src/disco/shred/fd_fec_resolver.c
+++ b/src/disco/shred/fd_fec_resolver.c
@@ -250,6 +250,14 @@ struct __attribute__((aligned(FD_FEC_RESOLVER_ALIGN))) fd_fec_resolver {
      signature. */
   ulong seed;
 
+  /* discard_unexpected_data_complete_shreds: activation slot for
+     discard_unexpected_data_complete_shreds.
+
+     This feature rejects data shreds with the DATA_COMPLETE flag set
+     at the wrong position within an FEC set (i.e. not the last data
+     shred). */
+  ulong discard_unexpected_data_complete_shreds;
+
   /* sha512 and reedsol are used for calculations while adding a shred.
      Their state outside a call to add_shred is indeterminate. */
   fd_sha512_t   sha512[1];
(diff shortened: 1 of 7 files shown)


src/disco/shred/fd_fec_resolver.c
```diff
@@ -562,9 +577,19 @@ fd_fec_resolver_add_shred( fd_fec_resolver_t         * resolver,
   if( FD_UNLIKELY( ( shred->fec_set_idx % FD_FEC_SHRED_CNT ) != 0UL ) ) return FD_FEC_RESOLVER_SHRED_REJECTED;
   if( FD_UNLIKELY( in_type_idx >= FD_FEC_SHRED_CNT ) )                  return FD_FEC_RESOLVER_SHRED_REJECTED;
   if( is_data_shred ) {
-    /* if it has data complete, it must be the last one in the FEC. */
-    if( FD_UNLIKELY( (shred->data.flags & FD_SHRED_DATA_FLAG_SLOT_COMPLETE) && ((1UL+shred->idx) % FD_FEC_SHRED_CNT) ) )
+    /* if it has slot complete, it must be the last one in the FEC. */
+    if( FD_UNLIKELY( (shred->data.flags & FD_SHRED_DATA_FLAG_SLOT_COMPLETE) && ((1UL+shred->idx) % FD_FEC_SHRED_CNT) ) ) {
       return FD_FEC_RESOLVER_SHRED_REJECTED;
+    }
```
topointon-jump, line 584: @ptaffet-jump this check was pre-existing, but I couldn't see an equivalent in the Agave code. The closest I can see is https://github.com/anza-xyz/agave/blob/v4.0.0-beta.1/ledger/src/shred.rs#L827-L834, but that checks `LAST_SHRED_IN_SLOT` (bits 6 and 7) vs our check for `FD_SHRED_DATA_FLAG_SLOT_COMPLETE` (bit 7). 

This check isn't changed in this PR, just flagging :)

ptaffet-jump: fd_shred_parse will fail if bit 7 is set and bit 6 is not.

PR #8813, opened 2026-03-11
snapshots, genesis: fix creation time units

cali-jumptrading:


src/discof/restore/utils/fd_ssmsg.h
```diff
@@ -266,8 +266,8 @@ typedef struct fd_snapshot_manifest_blockhash fd_snapshot_manifest_blockhash_t;
 
 struct fd_snapshot_manifest {
   /* The UNIX timestamp when the genesis block was for this chain
-     was created, in nanoseconds.  */
-  ulong creation_time_millis;
+     was created, in seconds.  */
```
jvarela-jump:
```suggestion
     was created, in seconds.
     https://github.com/anza-xyz/agave/blob/v4.0.0-beta.1/runtime/src/bank.rs#L2108-L2114 */
```

jvarela-jump (dismissed):

PR #8828, opened 2026-03-12
replay, snapshots: mix hard forks into bank hash

### LINKED ISSUE #8815: replay, snapshots: correctly mix in hard forks hash into bank hash
Agave includes the hard forks into the bank hash here: https://github.com/anza-xyz/agave/blob/master/runtime/src/bank.rs#L4808

Hard forks should usually be empty though.

Rhovian:
## Summary of changes
- Mix hard-fork count into the bank hash during replay finalization, matching Agave's `parent_slot < fork_slot <= slot` logic.
- Apply the same mixing in snapshot bank-hash verification.

## Assumptions
- Hard forks are usually empty in practice; this fixes a latent consensus gap when they are not.
- Snapshot manifests store the already-mixed frozen bank hash, consistent with Agave's serialization chain.
- Genesis needs no special handling.

## Tests
Added test_fd_hashes_apply_hard_forks in test_hashes.c:

- Added `test_fd_hashes_apply_hard_forks` in `test_hashes.c`
- Verified Agave reference outputs for fork-count 1 and 2 cases.
- Full Firedancer test execution is pending CI.

closes #8815

cc @cali-jumptrading

### DIFF
--- a/src/discof/replay/fd_replay_tile.c
+++ b/src/discof/replay/fd_replay_tile.c
@@ -926,8 +926,7 @@ replay_block_finalize( fd_replay_tile_t *  ctx,
   fd_bank_shred_cnt_set( bank, fd_sched_get_shred_cnt( ctx->sched, bank->data->idx ) );
 
   /* Do hashing and other end-of-block processing. */
-  fd_runtime_block_execute_finalize( bank, ctx->accdb, ctx->capture_ctx,
-                                     ctx->hard_forks, ctx->hard_forks_cnts, ctx->hard_forks_cnt );
+  fd_runtime_block_execute_finalize( bank, ctx->accdb, ctx->capture_ctx );
 
   /* Copy out cost tracker fields before freezing */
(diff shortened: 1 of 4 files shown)


jherrera-jump (changes requested):
Thanks for your contribution!

I'm going to suggest that we leave the hard fork mixing out of replay for now (i.e. the fd_runtime_block_execute_finalize call stack)

Adding this functionality there would really require that hard_forks be made into a bank field instead of keeping it in replay's global context, since it implicitly assumes that it could change from one bank to another. We're actually okay to ignore the fix here for now since in any realistic restart scenario there shouldn't be a case where replay will need to mix in a hard fork slot into the bank hash after booting.

The mix-in at snapshot verification is different since this will actually be used for a snapshot where the latest hard fork slot is equal to the manifest slot, which is the case during every real cluster restart.

Rhovian:
@jherrera-jump Done — removed the hard fork plumbing from fd_runtime_block_execute_finalize
and its callers. Snapshot verification path is unchanged.

 appreciate you moving this along!

PR #8836, opened 2026-03-12
gossip: fix gossvf crash; fix use-after-release when evicting peers

nlgripto:
ping_tracker_change() in fd_gossip.c filters callbacks where peer_pubkey == identity_pubkey.  When identity changes from B to A, an ACTIVE for peer B was suppressed (B was self), but the later INACTIVE passes through (identity is now A), with the result that `gossvf` crashes on FD_TEST(ping) trying to remove a peer it never added.

Fix: add fd_ping_tracker_remove() and call it in fd_gossip_set_identity for both old and new identity before updating identity_pubkey.  The old identity's INACTIVE is still filtered (identity hasn't changed yet), and the peer is gone from the tracker so no future INACTIVE can fire.  The new identity is also removed to prevent a leak (it would stay in gossvf's ping map forever since future INACTIVEs would be filtered).


mmcgee-jump:
Is this the right fix? Why does ping tracker filter identity_pubkey?

PR #8851, opened 2026-03-12
flamenco: missing check on full_inflation_vote

anarcheuz:
Does not currently impact mainnet since both features are enabled. But for code correctness and regression.

### DIFF
--- a/src/flamenco/rewards/fd_rewards.c
+++ b/src/flamenco/rewards/fd_rewards.c
@@ -49,7 +49,9 @@ get_inflation_start_slot( fd_bank_t const * bank ) {
       ? fd_bank_features_query( bank )->devnet_and_testnet
       : ULONG_MAX;
 
-  ulong enable = fd_bank_features_query( bank )->full_inflation_enable;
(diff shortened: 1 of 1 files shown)


src/flamenco/rewards/fd_rewards.c
```diff
@@ -49,7 +49,9 @@ get_inflation_start_slot( fd_bank_t const * bank ) {
       ? fd_bank_features_query( bank )->devnet_and_testnet
       : ULONG_MAX;
 
-  ulong enable = fd_bank_features_query( bank )->full_inflation_enable;
+  ulong enable = FD_FEATURE_ACTIVE_BANK(bank, full_inflation_vote)
+      ? fd_bank_features_query( bank )->full_inflation_enable
+      : ULONG_MAX;
```
topointon-jump, line 54: This has been cleaned up in our codebase, so we don't want conditional logic for it here. See `feature_map.json` - any features with `"cleaned_up":1` should not have any conditional logic in Firedancer.

PR #8869, opened 2026-03-13
flamenco: use the right prio for consistency

anarcheuz:
For consistency use the pointer on `&ele->pubkey`. Probably negligible perf impact. Tree might have a different shape than expected.


mjain-jump:
fixed by https://github.com/firedancer-io/firedancer/pull/9010

PR #8880, opened 2026-03-14
Fix firedancer-dev watch deadlock

ripatel-fd:
Fixes a startup deadlock where an app thread overflows the stderr
pipe before watch_cmd_fn runs.  This results in the app thread
permanently spinning on write EAGAIN while holding the log lock,
while watch.c can never boot (because it's also trying to acquire
the log lock).

As a workaround, release the log lock in fd_log.c and back off for
10µs, whenever EAGAIN is detected while trying to write to the log
file.


ripatel-fd:
> Not a good fix, makes production worse for a test scenario. Honestly we should probably just remove the log lock, interleaved log lines aren't that big a deal.

@mmcgee-jump This is a good trade, right now I cannot run _any_ tests 😂

PR #8882, opened 2026-03-14
support geyser plugin config

seraph98:
support geyser plugin


mmcgee-jump:
Not in scope for Frankendancer, we do not test geyser and do not warrant its correctness, nor plan to support in the default build.

PR #8894, opened 2026-03-16
replay: fix bank leak when evicting fork

ripatel-fd:
Reasm eviction in the replay tile was leaking banks and sched entries.


ripatel-fd:
This is probably wrong on second thought, @yufeng-jump @lidatong @ibhatt-jumptrading @mmcgee-jump can you review?

PR #8900, opened 2026-03-16
http: enforce client request timeout

ripatel-fd:
Add a hardcoded 5 second request timeout for any client request to
http_server.  Fixes a design flaw where a malicious client could
indefinitely stall a HTTP connection.

Also removes WARNING log spam from test_http_server


mmcgee-jump:
The entire point and design of this HTTP server is that it does not need to ever time anybody out, as should have been the design for the QUIC server too.

ripatel-fd:
@mmcgee-jump missed the eviction logic, makes sense

ripatel-fd:
> as should have been the design for the QUIC server too.

@mmcgee-jump That's what the QUIC server does. Firedancer disables timeouts, but Agave clients enforce a timeout upon themselves.

PR #8909, opened 2026-03-16
gui: fix infinite loop hazard

ripatel-fd:
If slot 0 gets optimistically confirmed, GUI can loop ULONG_MAX
times.  (Obviously, an AI slop report)


src/disco/gui/fd_gui.c
```diff
@@ -2271,7 +2271,8 @@ fd_gui_handle_optimistically_confirmed_slot( fd_gui_t * gui,
   if( FD_UNLIKELY( gui->summary.slot_optimistically_confirmed!=ULONG_MAX && _slot<gui->summary.slot_optimistically_confirmed ) ) {
```
jherrera-jump, line 2271: no infinite loop since
```
_slot<gui->summary.slot_optimistically_confirmed
```
prevents the loop from running if `gui->summary.slot_optimistically_confirmed` is 0

ripatel-fd: @jherrera-jump the infinite loop happens if `_slot` is 0

PR #8910, opened 2026-03-16
event: initialize genesis_hash

ripatel-fd:
In some unlucky timing case, genesis_hash can be read before it is
initialized.


ripatel-fd:
This code is confusing, there are undocumented invariants strewn over 3 different files and callback APIs

PR #8925, opened 2026-03-16
poh: fix max_microblocks case

jherrera-jump:


ptaffet-jump (approved):
Looks great! Thank you!

ptaffet-jump:
For the record, the reason this was broken before is:

> When we're at the end of the slot, and we've run up against restricted_hashcnt, the only way we can do any hashes is if a microblock arrives, which definitely reduces max_remaining_microblocks, and increases restricted_hashcnt . That gets violated if we have 1 microblock left, and that one arrives, but we keep max_remaining_microblocks at 1 because we haven't gotten the done_packing message. Basically, we did a hash, but we didn't see the expected decrease in max_remaining_microblocks.

PR #8962, opened 2026-03-19
gossvf: zero ContactInfos with multicast auxiliary sockets

### LINKED ISSUE #8396: Gossip should reject ContactInfos with multicast IP address

Rhovian:
## Summary

Zero out ContactInfo fields from gossip where any auxiliary socket (TVU, TPU, repair, etc.) has a multicast IP address. The gossip socket was already validated against multicast; this extends the check to all other sockets.

- Adds a multicast check loop over all ContactInfo sockets in `verify_addresses()`
- Intentionally scoped to multicast only — does not enforce public/private policy on non-gossip sockets
- IPv6 sockets are skipped (not coerced to zero), preserving existing behavior for peers advertising IPv6

Resolves #8396 

cc @mmcgee-jump


mmcgee-jump:
This behavior diverges from Agave, so I need to review why that is and how they solve this

mmcgee-jump (changes requested):
.

Rhovian:
> This behavior diverges from Agave, so I need to review why that is and how they solve this

My read of Agave is that it already sanitizes all ContactInfo sockets during deserialization, not just gossip. In [TryFrom<ContactInfoLite>](https://github.com/anza-xyz/agave/blob/master/gossip/src/contact_info.rs#L540), the loop at [L569](https://github.com/anza-xyz/agave/blob/master/gossip/src/contact_info.rs#L569) iterates every SocketEntry and calls [sanitize_socket()](https://github.com/anza-xyz/agave/blob/master/gossip/src/contact_info.rs#L578), which rejects multicast via [is_multicast()](https://github.com/anza-xyz/agave/blob/master/gossip/src/contact_info.rs#L603). This covers all socket types — TPU, TVU, repair, etc.

The difference is in severity: Agave silently omits the bad socket from the cache ([L578-580](https://github.com/anza-xyz/agave/blob/master/gossip/src/contact_info.rs#L578)) but still accepts the ContactInfo into CRDS — only a bad gossip socket triggers full rejection via [verify_gossip_addr()](https://github.com/anza-xyz/agave/blob/master/gossip/src/cluster_info.rs#L2483). This PR is stricter: it drops the entire ContactInfo at [L662](https://github.com/firedancer-io/firedancer/pull/8962/files#diff-L662), which seems like the right call given there's no legitimate multicast use case — Agave's own codebase [explicitly states](https://github.com/anza-xyz/agave/blob/master/gossip/src/node.rs#L62) "Agave does not currently support multicast," and turbine/repair/TPU are all unicast.

I'll leave the PR open as is lmk what you think @mmcgee-jump

ripatel-fd:
> This PR is stricter: it drops the entire ContactInfo

Firedancer aims to match Agave's CRDS filtering rules exactly. If either side filters more aggressively it would just cause persistent unnecessary traffic, because the protocol gets into a loop where the more lenient side keeps retransmitting the dropped value to the more aggressively-filtering side over and over.

Rhovian:
> > This PR is stricter: it drops the entire ContactInfo
> 
> Firedancer aims to match Agave's CRDS filtering rules exactly. If either side filters more aggressively it would just cause persistent unnecessary traffic, because the protocol gets into a loop where the more lenient side keeps retransmitting the dropped value to the more aggressively-filtering side over and over.

Given this, the divergence is entirely Multicast non-gossip socket (TVU, repair, etc.):

- agave: accept ContactInfo into CRDS, silently omit bad socket from cache
- firedancer: drop the entire ContactInfo

So I think to reach parity firedancer should also accept ContactInfo into CRDS but sanitize at deserialization time. I'd implement it by adjusting the current pr's drop to in place zeroing, something like:

```c
if( fd_ip4_addr_is_mcast( sock->ip4 ) ) {
  sock->ip4 = 0;
  sock->port = 0;
}
```

if yall think that makes sense, i'll adjust the PR.

edit, parity caveats:

- txsend and the repair path currently treat a zeroed socket as “keep the last cached endpoint” rather than “socket absent”
- multicast-only auxiliary-socket sanitize still wouldn’t fully match Agave, because Agave treats auxiliary IPv4 sockets with ip4 == 0 and nonzero port as absent, while Firedancer would still retain them

Rhovian:
Updated to match Agave's CRDS filtering: non-gossip multicast sockets are now zeroed in place (ip4=0, port=0) instead of dropping the entire ContactInfo.

Deffered for follow up work: txsend and repair treat zeroed sockets as "keep old cached value" rather than "socket absent"

PR #8965, opened 2026-03-19
repair: fix ping_cnt accounting for queued pongs

two-heart:


emwang-jump:
just noticed this! I ended up merging a fix in another PR but thanks for the catch 😁

PR #8976, opened 2026-03-19
bank: implement ABI changes for Agave 4.0

ptaffet-jump:
Compare with e.g. https://github.com/firedancer-io/firedancer/blob/ae885fde80068a6b54d9395a0edefc1e4989dda7/src/discoh/bank/fd_bank_abi.c

### DIFF
--- a/src/discoh/bank/fd_bank_abi.c
+++ b/src/discoh/bank/fd_bank_abi.c
@@ -253,6 +253,7 @@ struct ABI_ALIGN(8UL) fd_bank_abi_txn_private {
       ushort num_non_compute_budget_instructions;
       ushort num_non_migratable_builtin_instructions;
       ushort num_non_builtin_instructions;
+      ushort migrating_builtin[1]; /* the vote program */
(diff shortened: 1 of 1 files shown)


src/discoh/bank/fd_bank_abi.c
```diff
@@ -253,6 +253,7 @@ struct ABI_ALIGN(8UL) fd_bank_abi_txn_private {
       ushort num_non_compute_budget_instructions;
       ushort num_non_migratable_builtin_instructions;
       ushort num_non_builtin_instructions;
+      ushort migrating_builtin[1]; /* the vote program */
```
topointon-jump, line 256: nit: should we add a static_assert for this like the other fields?

```c
FD_STATIC_ASSERT( offsetof( struct fd_bank_abi_txn_private, compute_budget_instruction_details.migrating_builtin) == ...
```

topointon-jump: lol copilot had the same review comment 😂

PR #8980, opened 2026-03-19
gossip: reject LegacyContactInfo in pull requests

cmoyes-jump:
Match Agave's removal of `LegacyContactInfo` support (https://github.com/anza-xyz/agave/pull/9009).

Agave now rejects pull requests containing `LegacyContactInfo` during sanitization, only accepting `ContactInfo`. Firedancer was still accepting `LegacyContactInfo` in pull requests, causing a mismatch: pull requests that agave would reject were unexpectedly still being processed by Firedancer.

Also removes the now-dead `deser_legacy_contact_info()` function.

### DIFF
--- a/src/flamenco/gossip/fd_gossip_message.c
+++ b/src/flamenco/gossip/fd_gossip_message.c
@@ -6,11 +6,11 @@
 #include "../runtime/fd_system_ids.h"
 #include "../types/fd_types.h"
 
-/* https://github.com/anza-xyz/agave/blob/bff4df9cf6f41520a26c9838ee3d4d8c024a96a1/gossip/src/crds_data.rs#L22-L23 */
+/* https://github.com/anza-xyz/agave/blob/v4.0.0-alpha.0/gossip/src/crds_data.rs#L22-L23 */
 #define WALLCLOCK_MAX_MILLIS (1000000000000000UL)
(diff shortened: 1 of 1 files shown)


ripatel-fd (dismissed):

src/flamenco/gossip/fd_gossip_message.c
```diff
@@ -581,8 +566,8 @@ deser_pull_request( fd_gossip_message_t * message,
   message->pull_request->contact_info->offset = original_sz-*payload_sz;
   CHECK( deser_value( message->pull_request->contact_info, payload, payload_sz ) );
   message->pull_request->contact_info->length = original_sz-*payload_sz-message->pull_request->contact_info->offset;
-  CHECK( message->pull_request->contact_info->tag==FD_GOSSIP_VALUE_LEGACY_CONTACT_INFO ||
-         message->pull_request->contact_info->tag==FD_GOSSIP_VALUE_CONTACT_INFO );
+  /* https://github.com/anza-xyz/agave/blob/c13d79400af11752a8d6b61103971dffa3004741/gossip/src/protocol.rs#L160 */
```
ripatel-fd: can you use a version tag ref link here instead of a commit id?

mmcgee-jump (approved):

PR #8982, opened 2026-03-19
zksdk: fix range proofs for non-power of 2 bits

0x0ece:
This was a TODO, finally found a repro. Will add test vectors in a separated PR.


drubin-fd (approved):
LGTM. Very neat trick for computing the sumOfPowers, I did not see that one.

PR #8999, opened 2026-03-23
abi: fix die refaddr and sort

alpeng-jump:


contrib/tool/extract_abi.py
```diff
@@ -131,37 +131,47 @@ def c_escape(t, istype):
...
     if not offset in offset_union_info:
-        print(f"\n/* Skipping {t} */\n")
+        blocks.append(f"/* Skipping {t} */")
         continue
 
+    lines = []
     if len(offset_union_info[offset])>1:
-        print(f"/* t.decode('ascii') */")
-        print(f"typedef union {{")
+        lines.append(f"/* t.decode('ascii') */")
```
ptaffet-jump:
```suggestion
        lines.append(f"/* {t.decode('ascii')} */")
```

contrib/tool/extract_abi.py
```diff
@@ -131,37 +131,47 @@ def c_escape(t, istype):
...
+            if off>0: lines.append("\t};")
+        lines.append(f"}} {c_escape(t,1)};")
+        lines.append(f"FD_STATIC_ASSERT( sizeof({c_escape(t,1)})=={offset_to_type_sz[offset][0]}UL, bank_abi );")
+        lines.append(f"FD_STATIC_ASSERT( alignof({c_escape(t,1)})=={offset_to_type_sz[offset][1]}UL, bank_abi );")
+        blocks.append('\n'.join(lines))
         continue
 
-    print(f"/* t.decode('ascii') */")
-    print(f"typedef struct {{")
+    lines.append(f"/* t.decode('ascii') */")
```
ptaffet-jump:
```suggestion
    lines.append(f"/* {t.decode('ascii')} */")
```

contrib/tool/extract_abi.py
```diff
@@ -131,37 +131,47 @@ def c_escape(t, istype):
...
+        lines.append(f"FD_STATIC_ASSERT( offsetof({c_escape(t,1)}, {c_escape(name,0)})=={off}, bank_abi );")
+    lines.append(f"FD_STATIC_ASSERT( sizeof({c_escape(t,1)})=={offset_to_type_sz[offset][0]}UL, bank_abi );")
+    lines.append(f"FD_STATIC_ASSERT( alignof({c_escape(t,1)})=={offset_to_type_sz[offset][1]}UL, bank_abi );")
+    blocks.append('\n'.join(lines))
+
+blocks.sort(key=lambda b:
+    m.group(1) if (m := re.search(r'}\s+(\w+)\s*;', b))
+    else f"zzz_{m.group(1)}" if (m := re.search(r'Skipping\s+(.*)', b))
+    else b
+)
```
ptaffet-jump: I think it would be easier to make `blocks` a list of tuples `(key, string)`, with the key created when the element is created. Then you can just call sort without needing the regex to extract data we just created.

PR #9006, opened 2026-03-23
sched: smaller reorder buffer and more metrics

yufeng-jump:
sched footprint: 18 GB -> 1.6 GB


src/disco/metrics/metrics.xml
```diff
@@ -789,8 +789,14 @@ metric introduced.
   <gauge name="ReasmLatestFecIdx" summary="FEC set index of the latest FEC set in the reassembly queue that can be replayed" />
 
   <gauge name="SchedActiveBankIdx" summary="The bank that the replay scheduler is currently dispatching work for" />
+  <gauge name="SchedLastDispatchBankIdx" summary="The bank that the replay scheduler last dispatched work for" />
+  <gauge name="SchedLastDispatchNanos" converter="nanoseconds" summary="Nanoseconds since the replay scheduler last dispatched work" />
```
mmcgee-jump: Do not make metrics like this "duration since x", it is not sampleable. Just make it a timestamp like last_dispatch_time_nanos

PR #9028, opened 2026-03-25
implement enable_sbpf_v3_deployment_and_execution

topointon-jump:
- [x] restructure jump table to use v0 as a base with overrides, so it's obvious what is in each sbpf version
- [x] update stricter elf verification with latest changes from https://github.com/solana-foundation/solana-improvement-documents/pull/189
- [x] static syscalls https://github.com/solana-foundation/solana-improvement-documents/pull/178
- [x] sbpf v3 isa changes https://github.com/solana-foundation/solana-improvement-documents/pull/377
- [x] unit tests
- [x] ledgers

### DIFF
--- a/src/flamenco/vm/fd_vm.c
+++ b/src/flamenco/vm/fd_vm.c
@@ -156,8 +156,6 @@ fd_vm_validate( fd_vm_t const * vm ) {
 # define FD_CHECK_CALL_REG_SRC  ((uchar)9)  /* Validation should check that callx has valid register number in src */
 # define FD_CHECK_CALL_REG_DST  ((uchar)13) /* Validation should check that callx has a valid register number in dst */
 # define FD_CHECK_CALL_REG_IMM  ((uchar)10) /* Validation should check that callx has a valid register number in imm */
-# define FD_CHECK_CALL_IMM      ((uchar)11) /* Check call against functions registry */
-# define FD_CHECK_SYSCALL       ((uchar)12) /* Check call against syscalls registry */
 
   uchar validation_map[ 256 ] = {
     /* 0x00 */ FD_INVALID,    /* 0x01 */ FD_INVALID,    /* 0x02 */ FD_INVALID,    /* 0x03 */ FD_INVALID,
@@ -436,27 +434,6 @@ fd_vm_validate( fd_vm_t const * vm ) {
     case FD_CHECK_CALL_REG_DST: {
       if( FD_UNLIKELY( instr.dst_reg > 9 ) ) {
         return FD_VM_ERR_INVALID_REG;
-      }
-      break;
-    }
-
-    /* https://github.com/anza-xyz/sbpf/blob/v0.14.4/src/verifier.rs#L403-L409 */
-    case FD_CHECK_CALL_IMM: {
-      ulong target_pc = (ulong)( fd_long_sat_add( (long)i, fd_long_sat_add( (long)(int)instr.imm, 1 ) ) );
-      if( FD_UNLIKELY( !(
-        target_pc<text_cnt && fd_sbpf_is_function_start( fd_sbpf_instr( text[target_pc] ) )
-      ) ) ) {
-        return FD_VM_INVALID_FUNCTION;
-      }
-      break;
-    }
-
-    /* https://github.com/anza-xyz/sbpf/blob/v0.12.2/src/verifier.rs#L414-L423 */
-    case FD_CHECK_SYSCALL: {
-      /* check active syscall */
-      fd_sbpf_syscalls_t const * syscall = fd_sbpf_syscalls_query_const( vm->syscalls, (ulong)instr.imm, NULL );
-      if( FD_UNLIKELY( !syscall ) ) {
-        return FD_VM_INVALID_SYSCALL;
       }
       break;
     }


src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -739,6 +739,31 @@
...
+      /* External syscall
+         https://github.com/anza-xyz/sbpf/blob/v0.14.4/src/interpreter.rs#L545-L553 */
+      fd_sbpf_syscalls_t const * syscall = imm!=fd_sbpf_syscalls_key_null() ? fd_sbpf_syscalls_query_const( syscalls, (ulong)imm, NULL ) : NULL;
+      if( FD_UNLIKELY( !syscall ) ) goto sigillbr;
+      FD_VM_INTERP_SYSCALL_EXEC;
+    } else if( src == 1 ) {
+      /* Internal call
+         https://github.com/anza-xyz/sbpf/blob/v0.14.4/src/interpreter.rs#L555-L563
+         https://github.com/anza-xyz/sbpf/blob/v0.14.4/src/program.rs#L97-L103 */
+      long target_pc_l = fd_long_sat_add( (long)pc, fd_long_sat_add( (long)(int)imm, 1L ) );
```
0x0ece, line 754: we MAY need to special case entrypoint, like we do in 0x85?
(I guess best way is to just diff test what happens with that value...)

src/flamenco/vm/fd_vm_private.h
```diff
@@ -83,23 +83,30 @@ FD_STATIC_ASSERT( sizeof(fd_vm_vec_t)==FD_VM_VEC_SIZE, fd_vm_vec size mismatch )
    and create a huge mess.
    We define both, so hopefully it's foolproof. */
 
-#define FD_VM_SBPF_REJECT_RODATA_STACK_OVERLAP(v)  ( v != FD_SBPF_V0 )
-#define FD_VM_SBPF_ENABLE_ELF_VADDR(v)             ( v != FD_SBPF_V0 )
+/* https://github.com/anza-xyz/sbpf/blob/v0.14.4/src/program.rs#L28-L93 */
+#define FD_VM_SBPF_REJECT_RODATA_STACK_OVERLAP(v)          ( v != FD_SBPF_V0 )
```
0x0ece: delete this and next one? they seem unused (and they're still written as != V0 which seems risky)

src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -778,7 +801,10 @@
 
     FD_VM_INTERP_STACK_PUSH;
```
0x0ece, line 802: This should go AFTER the new code, because reg_dst could be 10.

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -757,71 +755,93 @@ fd_sbpf_elf_peek_strict( fd_sbpf_elf_info_t * info,
...
-  // if( FD_UNLIKELY( !fd_ulong_is_aligned( sizeof(fd_elf64_ehdr), 8UL ) ) ) {
-  //   return FD_SBPF_ELF_PARSER_ERR_INVALID_ALIGNMENT;
-  // }
 
-  /* Parse program headers (expecting 4 segments) */
+  /* Parse program headers (expecting up to 2 segments: rodata + bytecode)
+     https://github.com/anza-xyz/sbpf/blob/v0.14.4/src/elf.rs#L448-L484 */
+
+  #define STRICT_EXPECTED_PHDR_CNT (2U)
+  ulong expected_p_vaddr[ STRICT_EXPECTED_PHDR_CNT ] = { FD_SBPF_MM_BYTECODE_ADDR, FD_SBPF_MM_RODATA_ADDR }; /* { MM_RODATA_START(0), MM_BYTECODE_START(0x100000000) } */
```
0x0ece: Naming issue (not a bug, but maybe worth fixing?)

In fd we have:
```
#define FD_SBPF_MM_BYTECODE_ADDR (0x0UL)         /* bytecode */
#define FD_SBPF_MM_RODATA_ADDR   (0x100000000UL) /* readonly program data */
```

In agave:
```
MM_RODATA_START = 0 (rodata lives at 0 in V3)
MM_BYTECODE_START = 0x100000000 (bytecode lives at 0x100000000)
```

The names are kind of swapped...

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -757,71 +755,93 @@ fd_sbpf_elf_peek_strict( fd_sbpf_elf_info_t * info,
...
-  ulong expected_p_vaddr[ EXPECTED_PHDR_CNT ] = { FD_SBPF_MM_BYTECODE_ADDR, FD_SBPF_MM_RODATA_ADDR, FD_SBPF_MM_STACK_ADDR, FD_SBPF_MM_HEAP_ADDR };
-  uint  expected_p_flags[ EXPECTED_PHDR_CNT ] = { FD_SBPF_PF_X,             FD_SBPF_PF_R,           FD_SBPF_PF_RW,         FD_SBPF_PF_RW        };
-  fd_elf64_phdr     phdr[ EXPECTED_PHDR_CNT ];
-  for( uint i=0; i<EXPECTED_PHDR_CNT; i++ ) {
-    ulong phdr_off = sizeof(fd_elf64_ehdr) + i*sizeof(fd_elf64_phdr);
-    phdr[ i ] = FD_LOAD( fd_elf64_phdr, bin+phdr_off );
+  /* https://github.com/anza-xyz/sbpf/blob/v0.14.4/src/elf.rs#L455-L463
+     If the first PH is not marked as readonly, expect the rodata
+     segment to be skipped. */
+  fd_elf64_phdr phdr0 = FD_LOAD( fd_elf64_phdr, bin + sizeof(fd_elf64_ehdr) );
```
0x0ece, line 794: We need to be careful with alignment.

Alignment check: Agave's parse_file_header checks pointer alignment, but always succeeds because AlignedMemory::from_slice pre-aligns. 

Firedancer uses FD_LOAD for unaligned access.

The risk I can see is that we'd load an unaligned program but agave may fail to load it (or another side effect due to re-aligning it...).

We should make sure we fuzz this cc @mjain-jump

0x0ece: FYI, I had AI check and it should be good, but I still think we should cover this with the fuzzer just to be sure (esp. about future changes).

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -1887,6 +1901,34 @@ fd_sbpf_program_load_lenient( fd_sbpf_program_t *             prog,
...
+  /* https://github.com/anza-xyz/sbpf/blob/v0.14.4/src/elf.rs#L486-L496 */
+  fd_elf64_phdr bytecode_phdr;
+  if( FD_UNLIKELY( skip_rodata ) ) {
+    prog->rodata_sz = 0UL;
+    bytecode_phdr   = phdr_0;
+  } else {
+    prog->rodata_sz = phdr_0.p_memsz;
+    bytecode_phdr   = FD_LOAD( fd_elf64_phdr, bin+sizeof(fd_elf64_ehdr)+sizeof(fd_elf64_phdr) );
+  }
+
```
0x0ece, line 1929: We prob have to memcpy( prog->rodata, ... ) ?

src/ballet/sbpf/fd_sbpf_loader.c
topointon-jump: I _think_ this is fine, but you're right we definitely need to fuzz this

src/flamenco/vm/fd_vm_interp_core.c
topointon-jump: I don't think we do - I have this as a test program in the `enable_sbpf_v3_deployment_and_execution` ledger and the behaviour is the same. Agave also doesn't special-case this either.

src/flamenco/progcache/fd_prog_load.c
```diff
@@ -96,16 +96,16 @@ fd_prog_versions( fd_features_t const * features,
...
-  if( enable_v2 ) {
-    v.max_sbpf_version = FD_SBPF_V2;
-  } else if( enable_v1 ) {
-    v.max_sbpf_version = FD_SBPF_V1;
-  } else {
-    v.max_sbpf_version = FD_SBPF_V0;
-  }
-  v.min_sbpf_version = enable_v0 ? FD_SBPF_V0 : fd_uint_min( FD_SBPF_V2, v.max_sbpf_version );
+  /* https://github.com/anza-xyz/agave/blob/v4.0.0-beta.4/syscalls/src/lib.rs#L314-L319 */
+  v.min_sbpf_version = enable_v0 ? FD_SBPF_V0 : FD_SBPF_V3;
```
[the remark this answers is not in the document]
topointon-jump: I do agree but this is what agave does

src/flamenco/vm/fd_vm_interp_core.c
topointon-jump: I don't think this is necessary, the verifier will catch reg_dst > 9

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -778,16 +780,13 @@ fd_sbpf_elf_peek_strict( fd_sbpf_elf_info_t * info,
   if( FD_UNLIKELY( (program_header_table_end-sizeof(fd_elf64_ehdr))%sizeof(fd_elf64_phdr) ) ) {
     return FD_SBPF_ELF_PARSER_ERR_INVALID_SIZE;
   }
-  if( FD_UNLIKELY( program_header_table_end>bin_sz ) ) {
```
0x0ece, line 783: I think we need to keep this?

I feel it corresponds to this? https://github.com/anza-xyz/sbpf/blob/main/src/elf_parser/mod.rs#L592-L594

topointon-jump: checked in the block above

0x0ece: Redundant

0x0ece (approved):

src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -687,25 +692,48 @@
...
+      }
+    }
+
+    /* Internal call
+       https://github.com/anza-xyz/sbpf/blob/v0.14.4/src/interpreter.rs#L555-L573 */
+    if( FD_VM_SBPF_STATIC_SYSCALLS( sbpf_version ) ) {
+      /* https://github.com/anza-xyz/sbpf/blob/v0.14.4/src/interpreter.rs#L556-L563 */
+
+      /* https://github.com/anza-xyz/sbpf/blob/v0.14.4/src/program.rs#L97-L103 */
+      long target_pc_l = fd_long_sat_add( (long)pc, fd_long_sat_add( (long)(int)imm, 1L ) );
```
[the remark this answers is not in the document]
topointon-jump: it uses saturating

PR #9043, opened 2026-03-25
progcache: reclaim after release

anarcheuz:
`fd_progcache_txn_publish_release()` can enqueue to the reclaim queue.


ripatel-fd (approved):

ripatel-fd:
Good catch

PR #9046, opened 2026-03-26
progcache: reset potentially stale records

anarcheuz:
a new progcache can begin life with stale records when a workspace is reused (eg. stale `exists=1`, old `(xid,key)`, etc..). This can happen following a node restart or if an operator uses `fd_wksp_ctl reset` (only frees but does not reset).


ripatel-fd:
Good find, will fix this but slightly differently

ripatel-fd:
https://github.com/firedancer-io/firedancer/pull/9195

PR #9055, opened 2026-03-26
fix(tower,votes): key with (slot, block_id) and remove FD_VOTER_MAX

lidatong:


src/discof/tower/fd_tower_tile.c
```diff
@@ -77,6 +77,22 @@ FD_STATIC_ASSERT( 1<<AUTH_VTR_LG_MAX==32, AUTH_VTR_LG_MAX );
...
+
+#define VTR_MAX (2000) /* the maximum # of unique voters ie. node pubkeys. */
+
+/* PER_VTR_MAX controls how many "entries" a validator is allowed to
+   occupy in various vote-tracking structures.  This is set somewhat
+   arbitrarily based on expected worst-case usage by an honest validator
+   and is set to guard against a malicious spamming validator attempting
+   to fill up Firedancer structures. */
+
+#define PER_VTR_MAX (512) /* the maximum amount of slot history the sysvar retains */
```
emwang-jump, line 94: are we using this anywhere?

PR #9057, opened 2026-03-26
firedancer: clean up vote_keyed_lsched and chacha20

kbhargava-jump:


0x0ece:
For chacha, rust is still passing down chacha8 feature, but it's no longer used.
Therefore the file `src/disco/shred/fd_shredder.h` still contains a ref to the feature name.

When we touch the rust features we can remove it completely, but not worth making changes to agave in this PR.

Issue #8989, opened 2026-03-20
block duration always late by ~10ms


mmcgee-jump:
We target 350ms in Firedancer, but somehow even the local view always shows 360ms

jherrera-jump:
The extra 10ms comes from
- ~7ms tower fork choice
- ~1ms prepare_leader_bank
- ~4ms pack slot_done timing + poh tail-hashes

The plan is to fix the ~4ms by dynamically lowering the microblock upper bound throughout the slot. The remaining delay we will leave as is and optimize it down as a separate issue.

PR #9065, opened 2026-03-27
pack, poh: use a dynamic microblock bound

jherrera-jump:
closes https://github.com/firedancer-io/firedancer/issues/8989

Problem
--------
pack's slot_done message arrives at POH at 350ms after slot start. At this point POH still has to do ~130K hashes, which takes 4-5ms.

Solution
--------
By reducing the microblock bound over the course of the slot (and eventually to 1/3 of the remaining budget at the end of the slot), we should be able to cut this down to ~1-2ms. 1/3 is probably a bit conservative and we could go lower but I think it's a good starting point.

### DIFF
--- a/src/disco/pack/fd_pack_tile.c
+++ b/src/disco/pack/fd_pack_tile.c
@@ -206,6 +206,15 @@ typedef struct {
      for, if we are currently packing for a slot.*/
   long slot_end_ns;
 
+  /* The current dynamic upper bound on total microblocks for this slot.
+     Monotonically decreasing over the slot lifetime. */
+  ulong slot_dynamic_max_microblocks;
+
+  /* Set by during_housekeeping when the dynamic bound drops below
+     slot_max_microblocks.  Consumed by after_credit which publishes
+     the updated bound to POH over the pack_poh link. */
+  int pending_reduce_mb_bound;
+
   /* pacer and ticks_per_ns are used for pacing CUs through the slot,
      i.e. deciding when to schedule a microblock given the number of CUs
      that have been consumed so far.  pacer is an opaque pacing object,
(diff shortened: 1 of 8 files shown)


mmcgee-jump:
Reasonable but I don't think the math here is great, you probably want like a y = - x^1.5 or something curve on the pacing that hits 0 microblocks remaining right at end of slot? But you get the rapid decay closer to the end. Check with Philip might have good ideas.

mmcgee-jump (changes requested):
.

src/disco/pack/fd_pack_tile.c
```diff
@@ -451,6 +494,31 @@ during_housekeeping( fd_pack_ctx_t * ctx ) {
...
+  if( FD_LIKELY( ctx->leader_slot!=ULONG_MAX ) ) {
+    ulong raw = compute_dynamic_max_microblocks( ctx );
+    ulong prev = ctx->slot_dynamic_max_microblocks;
+
+    /* Enforce monotonically decreasing. This ensures pack's bound is
+       always smaller than POH's.  */
+    ctx->slot_dynamic_max_microblocks = fd_ulong_min( raw, prev );
+
+    /* If the bound decreased, update pack's internal scheduling limit
+       so it won't even attempt to schedule past the new bound. */
```
ptaffet-jump: Make this comment stronger. We have to update pack's bound (unfortunately). Otherwise, we might have only 4 microblocks left from the tile and poh's perspectives, and pack could schedule a 5 transaction microblock.

src/disco/pack/fd_pack_tile.c
ptaffet-jump, line 672: I think the `pending_reduce_mb_bound` should go here, but I'm not sure.

src/disco/pack/fd_pack_tile.c
```diff
@@ -520,6 +588,14 @@ after_credit( fd_pack_ctx_t *     ctx,
               int *               charge_busy ) {
   (void)opt_poll_in;
 
+  if( FD_UNLIKELY( ctx->pending_reduce_mb_bound ) ) {
+    ctx->pending_reduce_mb_bound = 0;
+    ulong * dst = fd_chunk_to_laddr( ctx->poh_out_mem, ctx->poh_out_chunk );
+    *dst = ctx->slot_dynamic_max_microblocks;
+    fd_stem_publish( stem, 1UL, FD_PACK_MSG_REDUCE_MB_BOUND, ctx->poh_out_chunk, sizeof(ulong), 0UL, 0UL, fd_frag_meta_ts_comp( fd_tickcount() ) );
+    ctx->poh_out_chunk = fd_dcache_compact_next( ctx->poh_out_chunk, sizeof(ulong), ctx->poh_out_chunk0, ctx->poh_out_wmark );
+  }
```
ptaffet-jump: This code should go later in the function. Also, doesn't the bust need to be updated?

src/discof/execle/fd_execle_tile.c
```diff
@@ -146,10 +146,10 @@ during_frag( fd_execle_tile_t * ctx,
   }
 
   fd_microblock_execle_trailer_t * trailer = (fd_microblock_execle_trailer_t *)( src+sz-sizeof(fd_microblock_execle_trailer_t) );
-  ctx->_bank_idx  = trailer->bank_idx;
-  ctx->_pack_idx  = trailer->pack_idx;
-  ctx->_txn_idx   = trailer->pack_txn_idx;
-  ctx->_is_bundle = trailer->is_bundle;
+  ctx->_bank_idx                = trailer->bank_idx;
```
ptaffet-jump: You can just revert these whitespace changes to clean up the PR

src/disco/pack/fd_pack_tile.c
jherrera-jump: Are you saying we should publish the bound decrease alongside done_packing?

I suppose we could but there's no real benefit right? Since done_packing is functionally telling POH to set the bound to exactly microblock_cnt

src/disco/pack/fd_pack_tile.c
jherrera-jump: Ah true. Will set opt_poll_in so that we can keep burst at 1

src/disco/pack/fd_pack_tile.c
ptaffet-jump: The place you put it looks good to me.

PR #9070, opened 2026-03-30
Extend macOS portability layer for Apple Silicon

yeabat8tcap:
This PR merges the Apple Silicon (M-series) portability layer from the firedancer-apple fork into the macos-integration branch. It includes:
- NUMA shims for macOS.
- Thread QoS/Priority handling via Mach APIs.
- Randomized shared memory mapping (fixed for Mach).
- accept4 shims and other socket handling fixes.


ripatel-fd:
Thank you for the contribution. This is a bit too big too review or merge (600K lines of changes), but I'd be happy to accept a series of PRs with smaller changes. For example, the first PR could just add `#if defined(__linux__)` guards.
Several components can be skipped entirely: It doesn't make sense to make the seccomp-generated files portable because seccomp itself is Linux specific.
Similarly, there's no point in adding NUMA support for macOS because there are no modern macOS systems that support multiple sockets.

ripatel-fd:
For this kind of contribution specifically, we cannot accept LLM-generated patches, because of the asymmetric human review cost.

yeabat8tcap:
Ahh I see, okay. Is it better to create a separate branch and cherry pick from there? I can be able to do that. Let me know, thanks. @ripatel-fd

ripatel-fd:
> Ahh I see, okay. Is it better to create a separate branch and cherry pick from there? I can be able to do that. Let me know, thanks. @ripatel-fd

Yes that would be great, thanks!

PR #9081, opened 2026-03-30
firedancer: mainnet and testnet toml update

jvarela-jump:
Updating `mainnet.toml` and `testnet.toml` to only fetch snapshots from configured servers under `[snapshot.sources]`, excluding gossip peers. This is particularly useful when doing `firedancer-dev --mainnet` and `firedancer-dev --testnet`.
Increasing `max_accounts` to 1.3B in `mainnet.toml`.

### DIFF
--- a/src/app/firedancer/config/mainnet.toml
+++ b/src/app/firedancer/config/mainnet.toml
@@ -10,7 +10,9 @@
     expected_genesis_hash = "5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d"
 [accounts]
     file_size_gib = 600
-    max_accounts = 1_100_000_000
+    max_accounts = 1_300_000_000
 [snapshots]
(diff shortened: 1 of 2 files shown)


mmcgee-jump:
Curious here what's the reason to disable arbitrary snapshot sources? The gossip sources should never get selected unless it's better?

jvarela-jump:
> Curious here what's the reason to disable arbitrary snapshot sources? The gossip sources should never get selected unless it's better?

During `firedancer-dev` tests, in a few occasions the download from rpc servers failed due to min bandwidth threshold (the threshold calculation was correct - the min bandwidth has been analyzed separately), causing the node to switch to gossip peers (which as not intended). But this PR has now become obsolete, considering that we have https://github.com/firedancer-io/firedancer/pull/9050 and https://github.com/firedancer-io/firedancer/pull/9088 in review.
Closing this PR.

PR #9090, opened 2026-03-31
resolv: fix Clang 21 build fail

riptl:


ptaffet-jump (approved):
It's actually GCC 15 I think on Ubuntu. You can do it without a cast by making another variable, but I think this is fine.

cmoyes-jump:
it's also clang 21 on Ubuntu

PR #9094, opened 2026-03-31
http_server: fix NULL terminator oob

anarcheuz:


src/waltz/http/fd_http_server.c
```diff
@@ -1422,13 +1422,15 @@ fd_http_server_printf( fd_http_server_t * http,
...
-             INT_MAX, /* We already proved it's going to fit above */
-             fmt,
-             ap );
+  vsnprintf( tmp, printed_len+1UL, fmt, ap );
   va_end( ap );
 
+  fd_memcpy( (char *)http->oring+(http->stage_off%http->oring_sz)+http->stage_len,
+              tmp,
+              printed_len );
+
```
ripatel-fd, line 1432: Would it not be way simpler to just do `fd_http_serevr_reserve( http, printed_len+1UL )` ???

anarcheuz: I was contemplating that version earlier but I think the API semantic is that messages are stored in the ring buffer without the null terminator, it accounts for the length of the message only. So if we do `fd_http_server_reserve( http, printed_len+1UL )` it might break API usage (messages that used to be stored could fail)

anarcheuz: Notably it broke `test_oring()`

anarcheuz: discussed with @jherrera-jump offline

PR #9096, opened 2026-03-31
Remove log print mutex

ripatel-fd:
Mutex is not signal-reentrant safe


mmcgee-jump:
I think you can now stop counting these normal pages in fd_topo_mlock etc

PR #9097, opened 2026-03-31
refactor(choreo): properly encapsulate API and improve diagnostics

lidatong:
Closes https://github.com/firedancer-io/auditor-internal/issues/476
Closes https://github.com/firedancer-io/auditor-internal/issues/467
Closes https://github.com/firedancer-io/auditor-internal/issues/406


src/discof/tower/fd_tower_tile.c
```diff
@@ -240,19 +250,23 @@ struct fd_tower_tile {
 
   /* static structures */
 
+  fd_pubkey_t                   id_keys  [VTR_MAX]; /* identity keys */
+  fd_pubkey_t                   vote_accs[VTR_MAX]; /* vote account addresses */
+  ulong                         stakes   [VTR_MAX]; /* stake[i] for vote_accs[i] */
```
emwang-jump, line 254: not using the vtr_t you defined above for this? If so delete the struct definition above (or use it here)

src/discof/tower/fd_tower_tile.c
```diff
@@ -819,97 +968,43 @@ replay_slot_completed( fd_tower_tile_t *            ctx,
...
-    found = 1;
-    ctx->our_vote_acct_sz = fd_ulong_min( fd_accdb_ref_data_sz( ro ), FD_VOTE_STATE_DATA_MAX );
-    our_vote_acct_bal = fd_accdb_ref_lamports( ro );
-    fd_memcpy( ctx->our_vote_acct, fd_accdb_ref_data_const( ro ), ctx->our_vote_acct_sz );
-    fd_accdb_close_ro( ctx->accdb, ro );
-    fd_tower_reconcile( ctx->tower, ctx->root_slot, ctx->our_vote_acct, ctx->tower_blocks );
-  }
+  fd_ghost_blk_t * ghost_blk;
+  if( FD_UNLIKELY( !ctx->init ) ) ghost_blk = fd_ghost_init  ( ctx->ghost, slot_completed->slot, &slot_completed->block_id );
+  else                            ghost_blk = fd_ghost_insert( ctx->ghost, slot_completed->slot, &slot_completed->block_id, &slot_completed->parent_block_id );
```
emwang-jump, line 974: We used to have a check that covers "it's possible this block from replay_slot_completed is on a minority fork Tower already pruned after publishing a new root," i.e. double checks if the ghost parent exists in the tree before calling ghost_insert. I think we are missing it now? ghost_insert would FD_TEST fail in that case

emwang-jump: I see you have it below, I think the SLOT_IGNORED case just needs to be moved before the fd_ghost_insert

src/choreo/tower/fd_tower.c
```diff
@@ -289,30 +484,31 @@ switch_check( fd_tower_t const  * tower,
...
 
-      for( fd_tower_lockos_slot_t const * slot = fd_tower_lockos_slot_map_ele_query_const( lockos->slot_map, &candidate_slot, NULL, lockos->slot_pool );
-                                          slot;
-                                          slot = fd_tower_lockos_slot_map_ele_next_const ( slot, NULL, lockos->slot_pool ) ) {
-        ulong interval_end = slot->interval_end;
-        ulong key = fd_tower_lockos_interval_key( candidate_slot, interval_end );
+      ulong sentinel_key = lockout_interval_key( candidate_slot, 0 );
+      for( lockout_interval_t const * sentinel = lockout_interval_map_ele_query_const( lockos->interval_map, &sentinel_key, NULL, lockos->interval_pool );
+                                                  sentinel;
+                                                  sentinel = lockout_interval_map_ele_next_const( sentinel, NULL, lockos->interval_pool ) ) {
```
emwang-jump: align

src/choreo/tower/fd_tower.c
```diff
@@ -4,11 +4,208 @@
...
+   Since a validator can have up to 31 entries in the tower, and we have
+   a max_vote_accounts, we can pool the interval objects to be
+   31*max_vote_accounts entries PER bank / executed slot. We can also
+   string all the intervals of the same bank together as a linkedlist. */
+
+struct lockout_interval {
+  ulong     key;   /* vote_slot (32 bits) | expiration_slot (32 bits) ie. vote_slot + (1 << confirmation count) */
+  ulong     next;  /* reserved for fd_map_chain and fd_pool */
+  fd_hash_t addr;  /* vote account address */
+  ulong     start; /* start of interval, also vote slot */
```
emwang-jump, line 57: Would be nice to also upd documentation / note that we will have multiple sentinel entries per slot, and in that case the start field holds different end interval values.

src/choreo/tower/fd_tower.c
```diff
@@ -1080,8 +1273,274 @@ to_cstr( fd_tower_t const * tower, ulong root, char * s, ulong len ) {
...
+fd_tower_lockos_remove( fd_tower_lockos_t * lockos,
+                        ulong               slot ) {
+
+  ulong sentinel_key = lockout_interval_key( slot, 0 );
+  for( lockout_interval_t * sentinel = lockout_interval_map_ele_remove( lockos->interval_map, &sentinel_key, NULL, lockos->interval_pool );
+                                        sentinel;
+                                        sentinel = lockout_interval_map_ele_remove( lockos->interval_map, &sentinel_key, NULL, lockos->interval_pool ) ) {
+    ulong interval_end = sentinel->start;
+    lockout_interval_pool_ele_release( lockos->interval_pool, sentinel );
+
```
emwang-jump, line 1465: align iterators

src/discof/tower/fd_tower_tile.c
lidatong: good catch, fixed

PR #9102, opened 2026-03-31
zksdk: fix double transcript_init

0x0ece:


drubin-fd (approved):
Just noting, no behaviour changes, the second transcript init call was just overriding what the first one did.

PR #9123, opened 2026-04-01
sysvar: remove clock from fd_types

riptl:


src/flamenco/runtime/sysvar/fd_sysvar_base.h
```diff
@@ -42,4 +42,13 @@
...
 #define FD_SYSVAR_STAKE_HISTORY_FOOTPRINT  ( 16408UL)
 
+struct fd_sol_sysvar_clock {
+  ulong slot;
+  long  epoch_start_timestamp;
+  ulong epoch;
+  ulong leader_schedule_epoch;
+  long  unix_timestamp;
+};
+typedef struct fd_sol_sysvar_clock fd_sol_sysvar_clock_t;
```
mjain-jump, line 50: may be good to add a static assert for the size, since theres assumptions on the wire size / layout being equivalent to the in-memory layout

mjain-jump (approved):

PR #9133, opened 2026-04-02
gossip: larger client vote history

### LINKED ISSUE #8444: gossip: all votes pushed with index 0

jherrera-jump:
closes https://github.com/firedancer-io/firedancer/issues/8444

### DIFF
--- a/src/discof/gossip/fd_gossip_tile.c
+++ b/src/discof/gossip/fd_gossip_tile.c
@@ -300,10 +300,12 @@ handle_shred_version( fd_gossip_tile_ctx_t * ctx,
 
 static void
 handle_local_vote( fd_gossip_tile_ctx_t * ctx,
+                   ulong                  vote_slot,
                    fd_txn_m_t const *     txn_m,
                    fd_stem_context_t *    stem ) {
   long now = ctx->last_wallclock + (long)((double)(fd_tickcount()-ctx->last_tickcount)/ctx->ticks_per_ns);
-  fd_gossip_push_vote( ctx->gossip, fd_txn_m_payload_const( txn_m ), txn_m->payload_sz, stem, now );
+  int err = fd_gossip_push_vote( ctx->gossip, vote_slot, fd_txn_m_payload_const( txn_m ), txn_m->payload_sz, stem, now );
+  if( FD_UNLIKELY( err ) ) FD_LOG_WARNING(( "fd_gossip_push_vote failed (%d): CRDS insert failed (stale or duplicate)", err ));
 }
 
 static void
@@ -426,7 +428,7 @@ returnable_frag( fd_gossip_tile_ctx_t * ctx,
 
   switch( ctx->in[ in_idx ].kind ) {
     case IN_KIND_SHRED_VERSION: handle_shred_version( ctx, sig ); break;
-    case IN_KIND_TXSEND:        handle_local_vote( ctx, fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk ), stem ); break;
+    case IN_KIND_TXSEND:        handle_local_vote( ctx, sig, fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk ), stem ); break;
     case IN_KIND_EPOCH:         handle_epoch( ctx, fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk ) ); break;
     case IN_KIND_TOWER:         handle_local_duplicate_shred( ctx, sig, fd_chunk_to_laddr_const( ctx->in[ in_idx ].mem, chunk ), stem ); break;
     case IN_KIND_SNAPIN_MANIF: {
--- a/src/flamenco/gossip/fd_gossip.c
+++ b/src/flamenco/gossip/fd_gossip.c
@@ -19,6 +19,14 @@ FD_STATIC_ASSERT( FD_METRICS_ENUM_CRDS_VALUE_CNT==FD_GOSSIP_VALUE_CNT,
 #define BLOOM_FALSE_POSITIVE_RATE (0.1)
 #define BLOOM_NUM_KEYS            (8.0)
 
+struct vote_entry {
+  int   is_occupied;
+  ulong vote_slot;
+  long  wallclock_ns;
+};
+
+typedef struct vote_entry vote_entry_t;
+
 struct stake {
   fd_pubkey_t pubkey;
   ulong       stake;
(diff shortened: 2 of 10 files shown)


src/flamenco/gossip/fd_gossip.c
```diff
@@ -802,23 +834,76 @@ fd_gossip_push( fd_gossip_t *             gossip,
...
+      /* client exit matches Agave's behavior */
+      FD_LOG_ERR(( "Existing vote (%lu) higher than newly cast vote (%lu). This might occur on a non-malicious node if "
+                   "there is already another node casting votes for this identity joined to the cluster.",
+                   e->vote_slot, vote_slot ));
+    } else if( FD_UNLIKELY( e->wallclock_ns<min_wallclock_ns || (e->wallclock_ns==min_wallclock_ns && e->index<min_index) ) ) {
+      min_wallclock_ns = e->wallclock_ns;
+      min_index     = e->index;
+      evict_pos     = i;
+    }
+  }
```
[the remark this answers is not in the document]
jherrera-jump: FD_LOG_ERR kills the client

src/flamenco/gossip/fd_gossip_message.h
```diff
@@ -23,6 +23,8 @@
 
 #define FD_CONTACT_INFO_TABLE_SIZE (32768UL)
 
+#define FD_GOSSIP_VOTE_IDX_MAX (12UL)
```
mmcgee-jump: Why is this 12 rather than 256 though?

jherrera-jump: To match Agave, which rejects indicies larger than 31. Otherwise Agave would drop CRDS values for larger indicies and we would keep trying to resend them in pull reponses.

The reason I'm using 12 instead of 32 is because agave's limit of 32 is based on their `OLD_MAX_VOTES` which is the bound on received (not produced) indicies -- their nodes only produce indicies 0..11, so we don't expect to see indicies 12..31 anyways. I could keep a separate bound for recieved/produced to match Agave exactly but didn't seem necessary.

jherrera-jump: Added 
#define FD_GOSSIP_VOTE_IDX_MAX     (12UL)
#define FD_GOSSIP_VOTE_IDX_OLD_MAX (32UL)
to differentiate

src/discof/gossip/fd_gossip_tile.c
```diff
@@ -291,10 +291,12 @@ handle_shred_version( fd_gossip_tile_ctx_t * ctx,
 
 static void
 handle_local_vote( fd_gossip_tile_ctx_t * ctx,
+                   ulong                  vote_slot,
                    fd_txn_m_t const *     txn_m,
                    fd_stem_context_t *    stem ) {
   long now = ctx->last_wallclock + (long)((double)(fd_tickcount()-ctx->last_tickcount)/ctx->ticks_per_ns);
-  fd_gossip_push_vote( ctx->gossip, fd_txn_m_payload_const( txn_m ), txn_m->payload_sz, stem, now );
+  int err = fd_gossip_push_vote( ctx->gossip, vote_slot, fd_txn_m_payload_const( txn_m ), txn_m->payload_sz, stem, now );
+  if( FD_UNLIKELY( err ) ) FD_LOG_WARNING(( "fd_gossip_push_vote failed (%d): CRDS insert failed (stale)", err ));
```
mmcgee-jump: this could be `stale` or `duplicate`

src/flamenco/gossip/fd_gossip.c
```diff
@@ -116,9 +124,44 @@ struct fd_gossip_private {
     fd_gossip_value_t ci[1];
   } my_contact_info;
 
+  vote_entry_t my_votes[ FD_GOSSIP_VOTE_IDX_MAX ]; /* Indexed by CRDS vote index */
```
mmcgee-jump, line 133: Do you want to clear this on identity key changed?

src/flamenco/gossip/fd_gossip.c
```diff
@@ -802,23 +849,61 @@ fd_gossip_push( fd_gossip_t *             gossip,
...
+    if( FD_LIKELY( !gossip->my_votes[ i ].is_occupied ) ) continue;
+    if( FD_UNLIKELY( gossip->my_votes[ i ].vote_slot==vote_slot ) ) {
+      /* Vote Refresh: same slot, new transaction (e.g. updated
+         blockhash).  Reuse the same CRDS vote index. */
+      vote_idx = i;
+      break;
+    }
+    if( FD_UNLIKELY( gossip->my_votes[ i ].vote_slot>vote_slot ) ) {
+      FD_LOG_ERR(( "Existing vote (%lu) higher than newly cast vote (%lu). This might occur on a non-malicious node if "
+                   "there is already another node casting votes for this identity joined to the cluster.",
```
mmcgee-jump, line 969: I don't think this can happen? push_vote is only for our own machine votes?

src/flamenco/gossip/fd_gossip.c
```diff
@@ -802,23 +849,61 @@ fd_gossip_push( fd_gossip_t *             gossip,
...
       .transaction_len = txn_sz,
     }}
   };
-  fd_memcpy( value.origin, gossip->identity_pubkey, 32UL );
+  fd_memcpy( value.origin, gossip->identity_pubkey, sizeof(fd_pubkey_t) );
   FD_TEST( txn_sz<=sizeof(value.vote->transaction) );
   fd_memcpy( value.vote->transaction, txn, txn_sz );
 
-  return fd_gossip_push( gossip, &value, stem, now );
+  if( FD_UNLIKELY( fd_gossip_push( gossip, &value, stem, now ) ) ) {
```
mmcgee-jump, line 992: You push it here  but it's also pushed in `handle_local_vote` now ?

src/flamenco/gossip/fd_gossip.c
```diff
@@ -802,23 +849,61 @@ fd_gossip_push( fd_gossip_t *             gossip,
...
+    /* CRDS insert failed (stale).  Do not record in tracker. */
```
mmcgee-jump: Stale or duplicate

mmcgee-jump (changes requested):
I think needs comprehensive `identity_key_changed` handling, otherwise good

src/flamenco/gossip/fd_gossip.c
jherrera-jump: Yeah it's more to fail fast if this invariant is ever broken.

src/discof/gossip/fd_gossip_tile.c
jherrera-jump: changing log message

src/flamenco/gossip/fd_gossip.c
jherrera-jump: changing comment

src/flamenco/gossip/fd_gossip.c
jherrera-jump: This invariant actually could be broken during identity changes if the new node starts on an older slot.

src/flamenco/gossip/fd_gossip.c
```diff
@@ -381,6 +426,22 @@ fd_gossip_set_identity( fd_gossip_t * gossip,
   fd_active_set_set_identity( gossip->active_set, gossip->identity_pubkey, gossip->identity_stake );
   fd_prune_finder_set_identity( gossip->prune_finder, gossip->identity_pubkey, gossip->identity_stake );
   refresh_contact_info( gossip, now );
+
+  /* Clear the vote tracker and repopulate it with any votes for the
+     new identity that already exist in CRDS. */
+
+  for( ulong i=0; i<FD_GOSSIP_VOTE_IDX_MAX; i++ ) {
```
jherrera-jump, line 444: fixed set-identity here

PR #9194, opened 2026-04-03
vm: optimizations

0x0ece:
These changes were inspired by conversations with the team and analysis of a trace of a p-token transfer.

On a global scale, the commits reduce the GCC interpreter .text size by ~8% (56.6 KB to 51.9 KB) and improve end-to-end mainnet ledger replay by 2.3% (134.0s to 130.9s mean over 3 runs of 1000 slots).

Moreover, p-token transfer doesn't access stack or heap. However our `fd_vm_new` zeros ~512 KB of stack and heap memory to account for the worst-case scenario. Inspired by [dancing cells](https://www.youtube.com/watch?v=PUJ_XdmSDZw), we delay zeroing to when the memory is actually accessed. This saves ~5k ns on start, but slows down each memory access. Fortunately, this combines well with the soft TLB change that reduces the number of memory accesses needing translation and zeroing checks. Overall, a transfer only does 2 of these checks, adding just 4-6 x86 instructions.

More details below.

**vm: soft TLB for memory translation**

Add a single-slot soft TLB that caches the most recent successful virtual-to-host address translation. On hit, translation reduces to ~8 instructions. Separate TLB instances are maintained for loads and stores (since `region_ld_sz` and `region_st_sz` can differ). The TLB is invalidated after syscalls and CPI to ensure correctness when memory mappings change.

**vm: lazy page zeroing (dancing pages)**

Defer stack/heap memory zeroing until first access using a 2 KB page bitmap. On `fd_vm_new`, only the config region (before `stack[]`) is zeroed. On `fd_vm_init`, bitmaps are reset. When `fd_vm_mem_haddr` translates a stack/heap address, untouched pages are zeroed on demand.

Note: I chose 2 KB instead of 4 KB because 4 KB is "too good" with the current max size of stack and heap — however if we ever need to change them, changing the impl would be much harder. 2 KB doesn't make the code noticeably worse, has no impact on perf (if anything, any zeroing is faster), and is much simpler to tweak for larger stack/heap.

Test harness updated to pre-zero 32 KB of stack/heap for syscall compatibility, with full cleanup via `fd_vm_zero_uninitialized_pages` after execution.


ripatel-fd:
based ... could you separate each optimization into a different commit and post A/B bench numbers?

ripatel-fd (dismissed):
Looks like backtest performance regressed. I would not take the risk of adding complexity to the VM (at least the soft TLB part) unless it improves performance

0x0ece:
Yeah, not merging until we have a clear picture (but I need the PR open for CI). There was an issue with gcc not inlining as requested, I fixed it and the improvement is more visible now. Keep in mind that the ledger "only" has 450k BPF ixs, so the gains are limited to a couple seconds (basically skipping the memset to zero).

topointon-jump (changes requested):
Just blocking this until I can take a proper look later today 😄

src/flamenco/vm/fd_vm_private.h
```diff
@@ -539,6 +761,32 @@ static inline void fd_vm_mem_st_8( ulong haddr,
...
+
+static inline void
+fd_vm_instr_decode_pext( ulong   instr,
+                         ulong * p_opcode,
+                         ulong * p_dst,
+                         ulong * p_src,
+                         ulong * p_offset,
+                         uint  * p_imm ) {
+  *p_opcode = instr & 0xFFUL;
+  ulong dst_src = _pext_u64( instr, 0x000000000000FF00ULL );
```
topointon-jump: Just curious, do we actually need the PEXT here? From what I can tell PEXT is useful when the bits you want to extract aren't contiguous (extracting non-adjacent bits and packing them into an adjacent area, so you don't have to do shift + mask twice), but here they _are_ contiguous. So it's the same as doing `dst_src = (instr >> 8) & 0xFF` - there is only one shift + mask operation.

Sorry if I'm misunderstanding, I'm not that familiar with PEXT so there might be some other benefit I'm missing.

src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -199,63 +247,39 @@
...
+     adopt this as pc_max.  At the start of a segment, pc_max = pc + cu.
+     Multiword instructions (LDDW) increment pc_max to compensate for
+     consuming an extra text word without costing an extra CU.
+
+     At a branch at pc:
+       - Check: pc >= pc_max implies CU exhaustion
+       - Remaining CU: cu_new = pc_max - pc - 1
+       - Segment instruction count: cu_old - cu_new
+
+     Both tracing and non-tracing paths use pc_max. */
```
topointon-jump: The CU accounting took us _ages_ to get right, so I'm most nervous about this change. Do you have benchmarks against mainnet for the CU change specifically?

src/flamenco/vm/fd_vm_private.h
0x0ece: Removed from this PR

src/flamenco/vm/fd_vm_interp_core.c
0x0ece: Remove from this PR

src/flamenco/vm/fd_vm_interp_core.c
```diff
@@ -180,9 +180,77 @@
...
+     In non-tracing mode, use fd_vm_mem_haddr_with_tlb for fast-path
+     translation (~5 instructions on TLB hit). */
+
+# undef  FD_VM_INTERP_MEM_LD_TRANSLATE
+# undef  FD_VM_INTERP_MEM_ST_TRANSLATE
+# undef  FD_VM_INTERP_MEM_LD_1_TRANSLATE
+# undef  FD_VM_INTERP_MEM_ST_1_TRANSLATE
+# undef  FD_VM_INTERP_TLB_INVALIDATE
+# undef  FD_VM_INTERP_TLB_LD_ARGS
+# undef  FD_VM_INTERP_TLB_ST_ARGS
```
ptaffet-jump: Why do we need these?

src/flamenco/vm/fd_vm_private.h
```diff
@@ -480,12 +536,210 @@ fd_vm_mem_haddr( fd_vm_t const * vm,
...
+  }
+
+  return haddr;
+}
+
+/* fd_vm_mem_haddr_with_tlb is a TLB-accelerated wrapper around
+   fd_vm_mem_haddr.  It caches the most recent successful translation
+   in a single-slot "soft TLB" using vaddr-space bounds (3 ulongs:
+   haddr_base, vaddr_lo, vaddr_hi).  On hit, translation costs ~8
+   x86 instructions (always inlined): two range comparisons on vaddr
```
ptaffet-jump: Instructions are not a good measure of time. Cycles are better. Especially since this introduces a branch, instruction count can be deceptive.

src/flamenco/vm/fd_vm_private.h
```diff
@@ -480,12 +536,210 @@ fd_vm_mem_haddr( fd_vm_t const * vm,
...
+fd_vm_mem_haddr_with_tlb( fd_vm_t const * vm,
+                          ulong           vaddr,
+                          ulong           sz,
+                          ulong const *   vm_region_haddr,
+                          uint  const *   vm_region_sz,
+                          uchar           write,
+                          ulong           sentinel,
+                          ulong *         p_tlb_haddr_base,
+                          ulong *         p_tlb_vaddr_lo,
+                          ulong *         p_tlb_vaddr_hi,
```
ptaffet-jump: I'd make a struct for these three values and pass the struct. When it's inlined, the compiler should be able to figure it out.  This function already has a lot of parameters.

src/flamenco/vm/fd_vm_private.h
```diff
@@ -480,12 +536,210 @@ fd_vm_mem_haddr( fd_vm_t const * vm,
...
+   region_ld_sz and region_st_sz can differ. */
+
+static inline __attribute__((always_inline)) ulong
+fd_vm_mem_haddr_with_tlb( fd_vm_t const * vm,
+                          ulong           vaddr,
+                          ulong           sz,
+                          ulong const *   vm_region_haddr,
+                          uint  const *   vm_region_sz,
+                          uchar           write,
+                          ulong           sentinel,
```
ptaffet-jump: This is separate, but since you're looking here: I previously had the insight that this `sentinel` approach is not a good idea. There are two operations: checking if the address is valid, and computing the `haddr` if so. If we separate these out (i.e. by returning two values), then the CPU should be able to start speculatively loading `haddr` even if it ends up not being valid, which should help hide memory latency.

src/flamenco/vm/fd_vm_interp.c
```diff
@@ -72,3 +104,93 @@ fd_vm_exec_trace( fd_vm_t * vm ) {
...
+  if( FD_UNLIKELY( region == FD_VM_STACK_REGION && stack_gaps_enabled ) ) {
+    ulong frame_base = offset & ~0x1FFFUL;
+    vaddr_lo = region_bits | frame_base;
+    slot_sz  = 0x1000UL;
+    fd_vm_t * vm_mut = (fd_vm_t *)vm;
+    fd_vm_lazy_zero_pages( vm_mut->stack_zero_bitmap, vm_mut->stack, frame_base >> 1, 0x1000UL );
+  } else if( FD_UNLIKELY( region == FD_VM_STACK_REGION || region == FD_VM_HEAP_REGION ) ) {
+    ulong page_base = offset & ~(FD_VM_LAZY_PAGE_SZ - 1UL);
+    vaddr_lo = region_bits | page_base;
+    slot_sz  = FD_VM_LAZY_PAGE_SZ;
```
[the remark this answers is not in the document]
0x0ece: Fixed

src/flamenco/vm/fd_vm_private.h
0x0ece: I AI-tried it but no measurable difference, we can prob test it in isolation later?

src/flamenco/vm/fd_vm_private.h
0x0ece: fd_vm_tlb_slot_t

src/flamenco/vm/fd_vm_interp_core.c
0x0ece: They're there just because _core.c is imported twice in fd_vm_exec_trace and notrace. I can probably remove them if you don't like them...

src/flamenco/vm/bench_vm_interp.c
```diff
@@ -0,0 +1,869 @@
...
+/* Hardware performance counters (Linux only) */
+
+#if defined(__linux__)
+#include <linux/perf_event.h>
+#include <sys/ioctl.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+extern long syscall( long, ... );
+
```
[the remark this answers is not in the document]
ripatel-fd: @0x0ece I agree with the box, maybe put `#define _GNU_SOURCE` at the beginning?

0x0ece:
Good bye TLB https://github.com/firedancer-io/firedancer/pull/11367

PR #9201, opened 2026-04-06
sbpf: compute calldests_max correctly

drubin-fd:
In the instance that a function is registered on the last few pcs of the text section, we will compute the calldests_max too small, resulting in it not being added to the calldests map while registering.


topointon-jump (dismissed):
you deleted the agave submodule?

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -1441,7 +1441,7 @@ fd_sbpf_lenient_elf_validate( fd_sbpf_elf_info_t * info,
   info->text_sz       = text_section_range.hi-text_section_range.lo;
   info->text_cnt      = (uint)( info->text_sz/8UL );
   info->shndx_text    = shndx_text;
-  info->calldests_max = fd_ulong_min( text_shdr->sh_size, bin_sz )/8UL;
+  info->calldests_max = (fd_ulong_min( text_shdr->sh_size, bin_sz ) + 7) / 8UL;
```
mjain-jump:
```suggestion
  info->calldests_max = (fd_ulong_min( text_shdr->sh_size, bin_sz )+7UL)/8UL;
```

PR #9217, opened 2026-04-06
stakes: include vote accounts with zero delegations while refreshing

drubin-fd:
In the Agave version of refresh_vote_accounts(), they compute the delegated_stakes from the stake_delegations, and then iterate over the vote_accounts, keeping entries but setting them to the Default if they aren't found in the delegated_stakes map. We would drop those entries in the delegation loop, which would lead to an inconsistent leader schedule state.


src/flamenco/stakes/fd_stakes.c
```diff
@@ -427,6 +427,36 @@ fd_refresh_vote_accounts( fd_bank_t *                    bank,
...
   ulong total_deactivating = 0UL;
   ulong staked_accounts    = 0UL;
+
+  /* Seed stake_accum_map with all vote accounts from the parent fork
+     with zero stake. The delegation loop below will update the stake
+     for any account that has active delegations.
+
+     In Agave, refresh_vote_accounts() builds delegated_stakes from
+     stake_delegations, but then iterates vote_accounts.iter(), instead
+     of the delegated_stakes map, so they never actually remove any entries
```
ibhatt-jumptrading: nit: it looks like this comment spills 72 chars

src/flamenco/stakes/fd_stakes.c
```diff
@@ -427,6 +427,36 @@ fd_refresh_vote_accounts( fd_bank_t *                    bank,
...
+     from the vote_accounts while refreshing. */
+  {
+    fd_vote_stakes_t * vs   = fd_bank_vote_stakes( bank );
+    ushort             pidx = bank->vote_stakes_fork_id;
```
ibhatt-jumptrading: nit: no need to define parent_idx twice (it's done below as well)

src/flamenco/stakes/fd_stakes.c
```diff
@@ -421,12 +421,43 @@ fd_refresh_vote_accounts( fd_bank_t *                    bank,
...
   ulong staked_accounts    = 0UL;
+
+  /* Seed stake_accum_map with all vote accounts from the parent fork
+     with zero stake. The delegation loop below will update the stake
+     for any account that has active delegations.
+
+     In Agave, refresh_vote_accounts() builds delegated_stakes from
+     stake_delegations, but then iterates vote_accounts.iter(), instead
+     of the delegated_stakes map, so they never actually remove any
+     entries from the vote_accounts while refreshing. */
```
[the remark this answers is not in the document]
drubin-fd: literally not true

PR #9222, opened 2026-04-06
remove FD_HAS_DOUBLE

drubin-fd:
Every reasonable compiler has double support, and for targets without hardware 64-bit floats, there are compiler-rt libraries built into the compilers which can emulate through soft-floats.

This changeset leaves src/util still guarded by FD_HAS_DOUBLE, but removes the guards from the rest of the codebase. It is impossible to implement Solana without doubles (in some form), as the protocol itself depends on it. FD_HAS_DOUBLE is then hardcoded to true within base.mk.


mmcgee-jump (dismissed):

ripatel-fd (changes requested):
Blocked by kbowers

ptaffet-jump:
Yes, you should talk to Kevin about the reason this exists.

ripatel-fd:
> Yes, you should talk to Kevin about the reason this exists.

We did before sending this PR. The conflict is that FD_HAS_DOUBLE solely exists to work around badly written compilers. This is not needed for any modern Clang or GCC build for any of the major architectures (x86, BPF, RISC-V, Arm, MIPS, etc) which all have proper software floating support.

This then opened a wider discussion about the build configuration slop that exists in config and fd_util_base. CMake does everything our build system does better and more correct at a fraction of the LOC, but it's less configurable and relies more on compilers being sane.

ripatel-fd:
Philip brought up a good point that the second motivation is to inform code whether there is fast double support, even if there is software emulation.

This is useful for embedded/exotic users of the util package but it's useless for Firedancer.

PR #9236, opened 2026-04-07
replay: publish correct execution, priority fees

cavemanloverboy:
GUI slot fees (`transaction_fee`, `priority_fee`) were always showing 0 because `publish_slot_completed` read them from the bank after `fd_runtime_settle_fees` had already zeroed them during freeze. This captures `execution_fees` and `priority_fees` before `fd_runtime_block_execute_finalize` and passes them through to the publish. Tips were unaffected since freeze only applies commission without zeroing.


cavemanloverboy:
we pass in 0 for snapshots and genesis banks btw

Issue #9177, opened 2026-04-02
util: known issues in utilities and infrastructure


hg8:
Known issues identified by security auditors for the **Utilities / Infrastructure** subsystem.
Listed for the Immunefi bug bounty contest to prevent duplicate reports.

> **Note:** Detailed reproduction steps and exploit specifics are intentionally omitted.

### 1. fd_cast: double_to_ulong incorrectly doesn't wrap to 0 on -INF

- **Impact**: Conformance
- **Code**: [`src/flamenco/types/fd_cast.h#L21`](https://github.com/firedancer-io/firedancer/blob/7749015ae1651dd71c60dceee9223ef23f46a6ba/src/flamenco/types/fd_cast.h#L21)
- **Summary**: Edge-case handling of negative infinity in float-to-unsigned-long cast does not match expected wrapping behavior.
- **Ref**: [auditor-internal#475](https://github.com/firedancer-io/auditor-internal/issues/475)

### 2. Potential alignment issues with FD_LAYOUT

- **Impact**: Memory Safety
- **Code**: [`src/util/bits/fd_bits.h#L961`](https://github.com/firedancer-io/firedancer/blob/7749015ae1651dd71c60dceee9223ef23f46a6ba/src/util/bits/fd_bits.h#L961)
- **Code**: [`src/discof/rpc/fd_rpc_tile.c#L203`](https://github.com/firedancer-io/firedancer/blob/7749015ae1651dd71c60dceee9223ef23f46a6ba/src/discof/rpc/fd_rpc_tile.c#L203)
- **Code**: [`src/disco/bundle/fd_bundle_tile.c#L34`](https://github.com/firedancer-io/firedancer/blob/7749015ae1651dd71c60dceee9223ef23f46a6ba/src/disco/bundle/fd_bundle_tile.c#L34)
- **Summary**: Several FD_LAYOUT macro usages violate the documented page_sz >= max(align) requirement.
- **Ref**: [auditor-internal#374](https://github.com/firedancer-io/auditor-internal/issues/374)

### 3. SCRATCH_ALLOC redzones tracking

- **Impact**: Enhancement (fuzzing)
- **Summary**: Tracking issue for adding deepasan redzones to SCRATCH_ALLOC for better OOB detection.
- **Ref**: [auditor-internal#56](https://github.com/firedancer-io/auditor-internal/issues/56)

### 4. Missing validation in MAP_(remove)

- **Impact**: Memory Safety
[...]

PR #9272, opened 2026-04-08
util: extra checks / correct misusages

jherrera-jump:
addresses some issues in https://github.com/firedancer-io/firedancer/issues/9177

the deepsan redzones is a more involved change, and more of a suggestion than an active vulnerability. We'll handle it in a future PR.


jherrera-jump:
Going to break this up

PR #9299, opened 2026-04-10
topo: allow startup and post-start tiles to overlap

riptl:
Some tiles are only active before snapshot load finishes, some
tiles only after load.  Tiles that cooperate in the following way
can be scheduled to the same CPU core to support lower core count
machines:
- startup tiles must shut down after snapshot load
- post-start tiles must sleep before snapshot load


src/disco/metrics/metrics.xml
```diff
@@ -883,6 +882,8 @@ metric introduced.
   <gauge name="AccdbCacheFreePartMaxBytes" summary="Largest free heap partition in account database cache" />
   <gauge name="AccdbCacheUsedPartMedianBytes" summary="Median used heap partition size in account database cache" />
   <gauge name="AccdbCacheUsedPartMeanBytes" summary="Mean used heap partition size in account database cache" />
+
+  <gauge name="Status" summary="Current replay tile status. 0 is booting, 1 is running" />
```
mmcgee-jump: Maybe not needed? There is a <tile> level metric for all tiles that tracks this. Or if it's different, clarify those semantics in the doc here

ripatel-fd: Needed, will clarify

ripatel-fd (approved):

PR #9314, opened 2026-04-11
repair, forest: construct startup repair requests from root to turbine_slot0

cavemanloverboy:
at startup, firedancer constructs repair requests by iterating backwards from the first slot it observes via turbine. this is because there could be skipped slots (e.g. root+1 could be skipped), which would result in unnecessary repair requests being sent out.

this is a bit more efficient in terms of the requests sent, but it means that replay is totally stalled until the root's child is discovered and repaired.

with the skip rate at ≈17 bps, it's better to just request everything in root..turbine_slot0 and build the tree from root -> turbine_slot0 instead, which unblocks replay instantly.

this improves firedancer restart times drastically overall, but especially so on low core count machines that can only afford a few execrp tiles.

### DIFF
--- a/src/discof/repair/fd_repair_tile.c
+++ b/src/discof/repair/fd_repair_tile.c
@@ -1000,6 +1000,31 @@ after_frag( ctx_t *             ctx,
         ctx->turbine_slot0 = shred->slot;
         fd_repair_metrics_set_turbine_slot0( ctx->slot_metrics, shred->slot );
         fd_policy_set_turbine_slot0( ctx->policy, shred->slot );
+
+        /* On first turbine shred, seed repair by queuing highest_shred
+           requests for slots between snapshot and turbine_slot0. This
+           bypasses forest entirely and dispatches directly via the sign
+           queue. Cap at half queue capacity to leave room for pongs. */
+        ulong root = fd_forest_root_slot( ctx->forest );
+        if( FD_LIKELY( root != ULONG_MAX && shred->slot > root ) ) {
+          ulong capacity = fd_signs_queue_max( ctx->pong_queue ) - fd_signs_queue_cnt( ctx->pong_queue );
+          ulong seed_cnt = fd_ulong_min( shred->slot-root, capacity/2 );
+          long  now_ms   = fd_log_wallclock()/(long)1e6;
+          ulong seeded   = 0UL;
+          for( ulong i=1; i<=seed_cnt; i++ ) {
+            if( fd_signs_queue_full( ctx->pong_queue ) ) break;
+            ulong slot = root + i;
+            fd_pubkey_t const * peer = fd_policy_peer_select( ctx->policy );
+            if( FD_UNLIKELY( !peer ) ) break;
+            fd_repair_msg_t * msg = fd_repair_highest_shred( ctx->protocol, peer, (ulong)now_ms, 0, slot, 0 );
+            if( FD_LIKELY( msg ) ) {
+              fd_signs_queue_push( ctx->pong_queue, (sign_pending_t){ .msg = *msg } );
+              seeded++;
+            }
+          }
+          FD_LOG_NOTICE(( "seeded %lu highest_shred requests for slots [%lu, %lu]",
+                          seeded, root + 1, root + seeded ));
+        }
       }
 
 


cavemanloverboy:
image

this works. i booted 3 times. each time i got this first red slot though not really sure if i got un/lucky with 3 skips or if there's something else going.

cavemanloverboy:
this should NOT be merged until we figure out why the gui shows one skipped slot always

cavemanloverboy:
gui issue fixed in https://github.com/firedancer-io/firedancer-frontend/pull/296

src/discof/forest/fd_forest.h
```diff
@@ -307,6 +307,21 @@ typedef struct fd_forest_ref fd_forest_ref_t;
...
+  ulong slot;
+  ulong idx; /* forest pool idx */
+};
+typedef struct fd_forest_req_pq_ele fd_forest_req_pq_ele_t;
+
+#define PRQ_NAME      fd_forest_req_pq
+#define PRQ_T         fd_forest_req_pq_ele_t
+#define PRQ_TIMEOUT_T ulong
+#define PRQ_TIMEOUT   slot
+#include "../../util/tmpl/fd_prq.c"
```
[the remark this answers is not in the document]
cavemanloverboy: this is fake, adding .c tmpl is commonplace in fd

src/discof/forest/fd_forest.c
```diff
@@ -151,6 +156,48 @@ requests_insert( fd_forest_t * forest,
...
+  fd_forest_req_pq_ele_t * prq = (reqsmap == fd_forest_requests( forest ))
+    ? fd_forest_reqsprq( forest )
+    : fd_forest_orphprq( forest );
+  fd_forest_req_pq_ele_t pq_ele = {
+    .prio = forest->req_seq++,
+    .slot = fd_forest_pool_ele_const( blk_pool, pool_idx )->slot,
+    .idx  = pool_idx
+  };
+  fd_forest_req_pq_insert( prq, &pq_ele );
+}
```
lidatong: could you potentially use a dlist instead of pq for arbitrary position insertion?

lidatong (commented):
nice, this is super cool thanks for exploring it

we had floated with this idea of repairing forwards using the slot + 1 thing instead of prioritizing backwards, but decided to stick with backwards orphaning

we did that partly for simplicity, partly because there's a security argument which is you want to repair & replay the canonical merkle chain backwards from the first duplicate confirmed slot you observe via gossip (vs. assuming slot + 1 is canonical). it's a somewhat unconvincing argument though, because there would have to be (an) unconfirmed competing fork(s) while we're catching up for that to matter

seeing the visually faster catchup for replay makes this convincing though

src/discof/forest/fd_forest.c
cavemanloverboy: the original struct was a dlist. i switched to a pq so we can send earlier slot requests first (via pop min)

cavemanloverboy:
> nice, this is super cool thanks for exploring it
> 
> we had floated with this idea of repairing forwards using the slot + 1 thing instead of prioritizing backwards, but decided to stick with backwards orphaning
> 
> we did that partly for simplicity, partly because there's a security argument which is you want to repair & replay the canonical merkle chain backwards from the first duplicate confirmed slot you observe via gossip (vs. assuming slot + 1 is canonical). it's a somewhat unconvincing argument though, because there would have to be (an) unconfirmed competing fork(s) while we're catching up for that to matter
> 
> seeing the visually faster catchup for replay makes this convincing though

there's a couple of ways we could go about this. the other way we can go about this is to record `turbine_slot0` right on startup and then fill in shreds from turbine while loading snapshot. then, the backwards traversal is much shorter.

emwang-jump (commented):
Nice! we floated the idea of this before -- nice to see that it's making a noticeable difference.  I think there's a cleaner way to do this than manipulating forest though (its already very stateful and heavy to work with, clearly).  Rather on `turbine_slot0` arrival it's probably better to push highest_window_index requests for `[snapshot_slot + 1, snapshot_slot + x]` into some queue/dstruct and dispatch with send_sign_request directly in repair_tile. (currently you have `x=pool_free / 2;`).  Should end up resulting in something similar?

lidatong:
> there's a couple of ways we could go about this. the other way we can go about this is to record `turbine_slot0` right on startup and then fill in shreds from turbine while loading snapshot. then, the backwards traversal is much shorter.

we need the leader schedule from snapshot to validate shreds though

src/discof/forest/fd_forest.c
lidatong: oh sorry i misremembered it as a deque. same comment though, i think pq is a little overkill because you basically have two insertion priorities and can configure the insertion priority / position by instead walking the dlist (or incrementally maintaining it, but prob fine to just walk at insertion time)

lidatong (approved):

PR #9321, opened 2026-04-13
flamenco: fix uninitialized stack variable usage

anarcheuz:
Harmless bug but zero-lamport accounts could pollute the acctdb


anarcheuz:
fixed by #9282

PR #9323, opened 2026-04-13
flamenco, features: re-key discard_unexpected_data_complete_shreds

topointon-jump:


ptaffet-jump (commented):
Looks like you forgot to update one place which is why CI is failing, but this LGTM otherwise

ptaffet-jump (approved):

PR #9325, opened 2026-04-13
flamenco: mark stake_minimum_delegation_for_rewards as reverted

drubin-fd:
It is unclear what Agave's plan is with this feature gate, and we explicitly do not implement it:
https://github.com/firedancer-io/firedancer/blob/881335e95bcf2987042a28a7106a76cb11b7e8f9/src/flamenco/rewards/fd_rewards.c#L360-L364
so we just remove it from the map, to indicate that we do not support it.


topointon-jump (dismissed):

topointon-jump:
we probably want to take it out of the supported features list in solfuzz-agave too

ripatel-fd (dismissed):

topointon-jump (dismissed):

topointon-jump (approved):

PR #9351, opened 2026-04-14
gossip: do not include restart CRDS in pull responses

ripatel-fd:
Requested by @alexpyattaev


mmcgee-jump:
Not a good change, this constant tinkering with gossip and what gets sent or not sent is unscientific and silly. I have communicated this to Anza as well in the past and they continue to do it, but we will not.

alexpyattaev:
@mmcgee-jump this is reflective of the fact that Agave removed support for restart. We no longer originate those messages. So parsers for them are dead weight in both clients. If you choose to not remove them we'll have to keep them around as well of course.

alexpyattaev:
A bit of context for this PR (as it lacks description at the moment): https://discord.com/channels/428295358100013066/1016733708562202674/1493700042547007703

PR #9365, opened 2026-04-15
deps: improve upgrade of commit-based deps

0x0ece:


kbhargava-jump:
this will increase CI jobs by 2 minutes for every CI job and will cause a new deps rerun every commit to every PR. this doesn't seem viable

kbhargava-jump (approved):

PR #9373, opened 2026-04-16
secp: add edge cases for k256 and align all cases in p256

0x0ece:
**E** = Explicit guard in our code, **I** = Implicit (handled correctly by s2n-bignum assembly)

### `point_add` (Jacobian + Jacobian)

| Case | k256 | p256 |
|---|---|---|
| `0 + 0 = 0` | I | I |
| `P + 0 = P` | I | I |
| `0 + P = P` | I | I |
| `P + P = 2P` (doubling) | **E** | **E** |
| `P + (-P) = 0` (negation) | I | I |

Both curves: `jadd` / `montjadd` handle identity and negation internally. Only the **doubling case** needs an explicit guard.

### `point_add_mixed` (Jacobian + affine)

| Case | k256 | p256 |
|---|---|---|
| `0 + 0 = 0` | **E** (b=0 check) | **E** (b=0 check) |
| `P + 0 = P` | **E** (b=0 check) | **E** (b=0 check) |
| `0 + P = P` | I | I |
| `P + P = 2P` (doubling) | **E** | **E** |
| `P + (-P) = 0` (negation) | I | I |

Both curves: `jmixadd` / `montjmixadd` handle `a.z==0` (identity first operand) and negation internally. The **b=0 guard** is needed because `(0,0)` is not a valid curve point — passing it to the assembly would give undefined behavior. The **doubling guard** is also needed.

### DIFF
--- a/src/ballet/secp256k1/fd_secp256k1.c
+++ b/src/ballet/secp256k1/fd_secp256k1.c
@@ -1,5 +1,12 @@
 #include "fd_secp256k1_private.h"
 
+/* Public wrapper for fp_sqrt (internal impl is static inline). */
+fd_uint256_t *
+fd_secp256k1_fp_sqrt( fd_uint256_t *       restrict r,
+                      fd_uint256_t const * restrict a ) {
+  return (fd_uint256_t *)fd_secp256k1_fp_sqrt_impl( (fd_secp256k1_fp_t *)r,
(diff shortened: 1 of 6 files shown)


0x0ece:
Closing in favor of https://github.com/firedancer-io/firedancer/pull/9384

PR #9402, opened 2026-04-17
sbpf: minor tweak to elf loader to remove code that does nothing

0x0ece:


drubin-fd (approved):

ripatel-fd (changes requested):
Seems silly

0x0ece:
what? code that does nothing, or reports that don't show coverage? this fixes both ;)

ripatel-fd:
> what? code that does nothing, or reports that don't show coverage? this fixes both ;)

The goal is not full coverage. We use various tools to gain high confidence, and there is no meaningful difference between 98% coverage and 100% coverage. You can have 100% coverage and still have buggy code paths, e.g. arithmetic overflows, or bugs that only appear in a particular sequence of branches. So, at the tail end, the metric is no longer meaningful. It is merely a decent hint which areas are clearly uncovered or which areas lack testing. 

In this case, I think we have a bad trade in this PR: This is unreachable code today, and the coverage system should accurately reflect that it's unreachable. Instead we theoretically diverge from Agave to game that metric. And if we do that in 40 places, we will eventually have 1 place where a feature gate turns an unreachable code block into an actual reachable one. The only sustainable way to do this is let the compiler or CBMC dead-code eliminate this block and extract metrics for that. But not hand inserting exceptions that are not automatically verified

0x0ece:
I disagree on pretty much every sentence, but this is a waste of time.

ripatel-fd (approved):

ripatel-fd:
OK

PR #9434, opened 2026-04-21
forest: fix buffered idx read past set boundary

emwang-jump:
closes https://github.com/firedancer-io/auditor-internal/issues/490

### DIFF
--- a/src/discof/forest/fd_forest.c
+++ b/src/discof/forest/fd_forest.c
@@ -1238,7 +1238,7 @@ fd_forest_data_shred_insert( fd_forest_t * forest,
   if( FD_UNLIKELY( ele->first_shred_ts == 0 ) ) ele->first_shred_ts = fd_tickcount();
 
   fd_forest_blk_idxs_insert( ele->idxs, shred_idx );
-  while( fd_forest_blk_idxs_test( ele->idxs, ele->buffered_idx + 1U ) ) {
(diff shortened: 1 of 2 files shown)


src/discof/forest/fd_forest.c
```diff
@@ -1242,7 +1238,7 @@ fd_forest_data_shred_insert( fd_forest_t * forest,
   if( FD_UNLIKELY( ele->first_shred_ts == 0 ) ) ele->first_shred_ts = fd_tickcount();
 
   fd_forest_blk_idxs_insert( ele->idxs, shred_idx );
-  while( fd_forest_blk_idxs_test( ele->idxs, ele->buffered_idx + 1U ) ) {
+  while( ele->buffered_idx + 1 < FD_SHRED_BLK_MAX && fd_forest_blk_idxs_test( ele->idxs, ele->buffered_idx + 1U ) ) {
```
lidatong, line 1241:
```suggestion
  while( fd_forest_blk_idxs_test( ele->idxs, fd_uint_min( ele->buffered_idx + 1U, ... ) ) ) {
```

emwang-jump: I think these aren't equivalent statements, if buffered is FD_SHRED_BLK_MAX - 1, then the blk_idx_test would test true which would cause buffered_idx > complete_idx

lidatong (approved):

PR #9443, opened 2026-04-22
monitor: fix busy% calculation

ripatel-fd:
Fixes a monitor bug displaying tiles as idle when they aren't


mmcgee-jump (changes requested):
This is almost surely not correct, we have had this argument about 100 times. At minimum if you think this is correct, despite all the history, write a long comment in the code here explaining why.

ripatel-fd:
> This is almost surely not correct, we have had this argument about 100 times. At minimum if you think this is correct, despite all the history, write a long comment in the code here explaining why.

@mmcgee-jump This is a line-by-line port of the logic you maintain in the GUI

ripatel-fd:
@mmcgee-jump https://github.com/firedancer-io/firedancer-frontend/blob/7c384af7c21a8e3bbd2aa4777a0bff2fe73b3faa/src/features/Overview/LiveTileMetrics/index.tsx#L251-L254

If this is wrong, please fix the GUI too. You maintain this logic, but the console monitor is presently broken

mmcgee-jump (approved):

Issue #7906, opened 2026-01-20
firedancer-dev stderr log has no color with --no-sandbox --no-clone


ripatel-fd:
Seems to be a bug in Firedancer's boot procedure that treats stderr incorrectly -

Results in output being stuck/not line buffered (if watch is enabled), or have no color (if `--no-watch` is set)

Issue #7629, opened 2025-12-24
solcap: fflush on shutdown


ripatel-fd:
The solcap tile logs crucial debug information to disk.

Some forms of crashes (e.g. FD_LOG_CRIT) result in buffered debug info getting lost because the solcap tile doesn't have sufficient time to flush its buffers on shutdown.

PR #9454, opened 2026-04-23
fix: resolve solcap flush and dev stderr color issues

shubhamessier:
### Updates

Addresses two issues:

1. **solcap**
   - Added `fsync()` calls to ensure critical debug data (such as bank preimages) and periodic buffers are properly flushed to disk  
   - Prevents data loss during sudden termination  
   - Reference: #7629  

2. **dev**
   - Fixed stderr colorization when using `--no-sandbox` and `--no-clone`  
   - Now detects TTY and forces correct color state before redirecting the stream to the watch pipe  
   - Reference: #7906

### DIFF
--- a/src/app/shared_dev/commands/dev.c
+++ b/src/app/shared_dev/commands/dev.c
@@ -148,6 +148,13 @@ dev_cmd_fn( args_t *   args,
   }
 
   update_config_for_dev( config );
+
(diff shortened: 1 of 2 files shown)


shubhamessier:
Hey @SEJeff can you kindly review the changes, thanks!

PR #9457, opened 2026-04-23
refactor(tower): polish fd_tower_reconcile

lidatong:


src/choreo/tower/fd_tower.c
```diff
@@ -935,153 +935,160 @@ fd_tower_vote_and_reset( fd_tower_t * tower,
...
-  if( FD_LIKELY( on_chain_root == ULONG_MAX || tower_root > on_chain_root ) ) {
-    while( FD_LIKELY( !fd_tower_vote_empty( tower->votes ) ) ) {
-      fd_tower_vote_t const * vote = fd_tower_vote_peek_head_const( tower->votes );
-      if( FD_LIKELY( vote->slot > tower_root ) ) break;
-      fd_tower_vote_pop_head( tower->votes );
+  /* on-chain has been reconciled with local, so overwrite local with
+     updated on-chain values. */
+
+  tower->root = onchain_root;
+  fd_tower_vote_remove_all( tower->votes );
```
emwang-jump, line 1092: I think at this point we need to clear `tower_blk->voted` for the slots that got removed from the tower?

src/choreo/tower/fd_tower.c
```diff
@@ -935,153 +935,160 @@ fd_tower_vote_and_reset( fd_tower_t * tower,
...
-
-  if( FD_LIKELY( on_chain_root == ULONG_MAX || tower_root > on_chain_root ) ) {
-    while( FD_LIKELY( !fd_tower_vote_empty( tower->votes ) ) ) {
-      fd_tower_vote_t const * vote = fd_tower_vote_peek_head_const( tower->votes );
-      if( FD_LIKELY( vote->slot > tower_root ) ) break;
-      fd_tower_vote_pop_head( tower->votes );
+  /* on-chain has been reconciled with local, so overwrite local with
+     updated on-chain values. */
+
+  tower->root = onchain_root;
```
emwang-jump: not sure about updating `tower->root` directly here (might need to double check), but since we prune state by walking from the `tower->root` to the new root we might miss pruning some stuff and leak memory.  iirc originally it was fine to not update `tower->root` because the next vote would update the `tower->root` to the onchain_root

lidatong: good catch

src/choreo/tower/fd_tower.c
lidatong: good catch

src/choreo/tower/fd_tower.c
```diff
@@ -935,153 +935,175 @@ fd_tower_vote_and_reset( fd_tower_t * tower,
...
+    fd_tower_lockos_remove( tower, slot );
+    fd_tower_stakes_remove( tower, slot );
+  }
+
+  tower->root = onchain_root;
+  for( fd_tower_vote_iter_t iter = fd_tower_vote_iter_init( tower->votes );
+                                  !fd_tower_vote_iter_done( tower->votes, iter );
+                            iter = fd_tower_vote_iter_next( tower->votes, iter ) ) {
+    fd_tower_vote_t const * vote = fd_tower_vote_iter_ele_const( tower->votes, iter );
+    fd_tower_vote_push_tail( tower->votes, *vote );
```
emwang-jump: does this `push_tail` need to be removed?

lidatong: rip shouldnt have removed run unit test

emwang-jump (approved):

PR #9485, opened 2026-04-23
solfuzz: add extra ContactInfo fields to gossip decoding harness

cmoyes-jump:
Anza requested that we check these extra ContactInfo fields in our conformance testing: https://github.com/anza-xyz/agave/pull/11856

- Add extra ContactInfo fields to harness (I do not want production spad space getting polluted by them, but open to moving the deserialization logic from the harness to the production code, even if they are unused... please feel free to weigh in)
- ~Address mismatches found in local fuzzing~ Mismatches went away after constraining fuzzer max input to MTU size

### DIFF
--- a/src/flamenco/gossip/fd_gossip_message.c
+++ b/src/flamenco/gossip/fd_gossip_message.c
@@ -162,7 +162,7 @@ deser_vote_txn( fd_gossip_vote_t * vote,
     }
     ushort data_len;
     READ_U16_VARINT( data_len, payload, payload_sz );
-    uchar data[ FD_TPU_MTU ];
+    uchar data[ 1232UL ];
     READ_BYTES( data, data_len, payload, payload_sz );
     if( FD_LIKELY( i==0UL ) ) {
       CHECK( accounts_len );
@@ -234,12 +234,10 @@ deser_bitvec_u8_epoch_slots( uchar const ** payload,
   ulong bits_cap;
   READ_U64( bits_cap, payload, payload_sz );
   CHECK( bits_cap );
-  ulong bits_cap_x8;
-  CHECK( !__builtin_mul_overflow( bits_cap, 8UL, &bits_cap_x8 ) );
   SKIP_BYTES( bits_cap, payload, payload_sz );
   ulong bits_cnt;
   READ_U64( bits_cnt, payload, payload_sz );
-  CHECK( bits_cnt==bits_cap_x8 );
+  CHECK( bits_cnt==bits_cap*8UL );
   return 1;
 }
 
@@ -454,12 +452,10 @@ deser_bitvec_u8_restart_last_voted_fork_slots( uchar const ** payload,
   ulong bits_cap;
   READ_U64( bits_cap, payload, payload_sz );
   CHECK( bits_cap );
-  ulong bits_cap_x8;
-  CHECK( !__builtin_mul_overflow( bits_cap, 8UL, &bits_cap_x8 ) );
   SKIP_BYTES( bits_cap, payload, payload_sz );
   ulong bits_cnt;
   READ_U64( bits_cnt, payload, payload_sz );
-  CHECK( bits_cnt<=bits_cap_x8 );
+  CHECK( bits_cnt<=bits_cap*8UL );
   return 1;
 }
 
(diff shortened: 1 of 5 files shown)


src/flamenco/gossip/fd_gossip_message.c
```diff
@@ -298,6 +300,7 @@ deser_snapshot_hashes( fd_gossip_value_t * value,
   CHECK( value->snapshot_hashes->full_slot<MAX_SLOT );
   READ_BYTES( value->snapshot_hashes->full_hash, 32UL, payload, payload_sz );
   READ_U64( value->snapshot_hashes->incremental_len, payload, payload_sz );
+  CHECK( value->snapshot_hashes->incremental_len<=FD_GOSSIP_SNAPSHOT_HASHES_MAX_INCREMENTAL );
```
mmcgee-jump: Where did these mismatches come from? Were these latent and just undiscovered, or Anza made some changes?

src/flamenco/gossip/fd_gossip_message.h
```diff
@@ -52,6 +52,12 @@
 
 #define FD_GOSSIP_MESSAGE_MAX_CRDS (17UL)
 
+#define FD_GOSSIP_DUPLICATE_SHRED_MAX_CHUNK_SZ  (1055UL)
```
mmcgee-jump: I don't really like moving these away from the extremely long and thorough explanations of how they are derived. Why do this?

src/flamenco/gossip/fd_gossip_message.c
```diff
@@ -572,6 +579,7 @@ deser_pull_response( fd_gossip_message_t * message,
                      ulong                 original_sz ) {
   READ_BYTES( message->pull_response->from, 32UL, payload, payload_sz );
   READ_U64( message->pull_response->values_len, payload, payload_sz );
+  CHECK( message->pull_response->values_len<=FD_GOSSIP_MESSAGE_MAX_CRDS );
```
mmcgee-jump: Why? the parser is structured so that these checks aren't necessary, it will fail automatically due to not having enough bytes in this case anyway

src/flamenco/gossip/fd_gossip_message.c
cmoyes-jump: Discovered during live fuzzing as crashes. Anza didn't make any changes. I suspect if I constrain the max buffer length to the MTU, these checks won't be needed. Let me check.

src/flamenco/gossip/fd_gossip_message.c
```diff
@@ -532,6 +537,7 @@ deser_bitvec_u64( fd_gossip_bloom_t * bloom,
 
   READ_U64( bloom->bits_cap, payload, payload_sz );
   CHECK( bloom->bits_cap );
+  CHECK( bloom->bits_cap<=FD_GOSSIP_BLOOM_MAX_BITS );
```
mjain-jump: can we condense this with the check above, i.e. 
```suggestion
  CHECK( bloom->bits_cap && bloom->bits_cap<=FD_GOSSIP_BLOOM_MAX_BITS );
```

src/flamenco/gossip/fd_gossip_message.h
cmoyes-jump: They're magic numbers, I don't mind putting them back, but then they're in other places, too, without any context. If you would like me to undo this, I can?

src/flamenco/gossip/fd_gossip_message.c
cmoyes-jump: Previous teams I've worked on went out of their way to not do this: When one combined checks fails, you don't know whether it was too big or too small using the line number in a stack trace.

cmoyes-jump: Removed it anyway

src/flamenco/runtime/tests/fd_gossip_harness.c
```diff
@@ -223,6 +429,8 @@ fd_solfuzz_gossip_decode( fd_solfuzz_runner_t * runner,
                           ulong *               out_sz,
                           uchar const *         in,
                           ulong                 in_sz ) {
+  if( FD_UNLIKELY( in_sz>FD_TPU_MTU ) ) FD_LOG_CRIT(( "gossip input %lu bytes exceeds MTU %lu, check fuzzer configuration", in_sz, FD_TPU_MTU ));
```
mjain-jump:
```suggestion
  if( FD_UNLIKELY( in_sz>FD_TPU_MTU ) ) FD_LOG_CRIT(( "invariant violation: gossip input %lu bytes exceeds MTU %lu, check fuzzer configuration", in_sz, FD_TPU_MTU ));
```

not a big deal just matches convention of the other crits

mjain-jump (dismissed):

mjain-jump (approved):

src/flamenco/gossip/fd_gossip_message.h
mmcgee-jump: You can move them here if you like but I would then move the long descriptions of the limit derivation too, it doesn't make sense to separate them.

PR #9486, opened 2026-04-24
Fix feature account decoding

ripatel-fd:


ripatel-fd:
NVM, Agave changed the minimum feature account size to 9 bytes at one point

PR #9492, opened 2026-04-24
xdp: advance stem sequence after RX publish

RekCuy63:
Fixes XDP RX publish accounting so the stem output sequence stays in sync with packets written directly to downstream mcaches.

The XDP RX path publishes frames with `fd_mcache_publish` instead of `fd_stem_publish`, but it was advancing only the local net output sequence. Stem housekeeping uses its own output sequence to update mcache `seq[0]`, so monitoring could observe stale progress for net output links.

This changes XDP RX publishing to use the stem output sequence for the selected output link and calls `fd_stem_advance` after a successful publish, matching the manual-publish pattern used by QUIC. The XDP unit test now checks that the stem
sequence advances with the published mline.


tristanx86-jump:
Resolved as side effect of https://github.com/firedancer-io/firedancer/pull/11640

PR #9493, opened 2026-04-24
gossip: compute prune threshold without overflow                                                                                                                                           

lmorett1:
Compute min_ingress_stake with quotient/remainder decomposition, preserving floor(min_base * pct / 100) semantics without overflow.


jherrera-jump:
I'd suggest something like
```
ulong min_ingress_stake = (ulong)((double)min_base * (double)FD_PRUNE_FINDER_STAKE_THRESHOLD_PCT / 100.0);
```

lmorett1:
> I'd suggest something like
> 
> ```
> ulong min_ingress_stake = (ulong)((double)min_base * (double)FD_PRUNE_FINDER_STAKE_THRESHOLD_PCT / 100.0);
> ```

Updated. Dropped the __uint128_t path — now using quotient/remainder to compute floor(min_base * pct / 100), avoids overflow and behaves the same.

src/flamenco/gossip/fd_prune_finder.c
```diff
@@ -235,7 +236,9 @@ do_prune( fd_prune_finder_t * pf,
      Relayers covering the top min_ingress_nodes and enough cumulative
      stake to meet min_ingress_stake are kept; the rest are pruned. */
   ulong min_base = fd_ulong_min( pf->identity_stake, origin->origin_stake );
-  ulong min_ingress_stake = min_base * FD_PRUNE_FINDER_STAKE_THRESHOLD_PCT / 100UL;
+  /* Keep floor(min_base * pct / 100) semantics while avoiding overflow. */
+  ulong min_ingress_stake = fd_ulong_sat_mul( min_base / 100UL, FD_PRUNE_FINDER_STAKE_THRESHOLD_PCT )
```
jherrera-jump: We shouldn't need fd_ulong_sat_mul here, since `FD_PRUNE_FINDER_STAKE_THRESHOLD_PCT<=100` then

(min_base/100)*FD_PRUNE_FINDER_STAKE_THRESHOLD_PCT cannot overflow.

jherrera-jump (changes requested):

src/flamenco/gossip/fd_prune_finder.c
lmorett1: Good point — updated. I removed fd_ulong_sat_mul and switched to exact quotient/remainder math:                                                                                                     
  q = min_base/100, r = min_base%100, then min_ingress_stake = q*pct + floor(r*pct/100).                                                                                                              
  This preserves floor(min_base*pct/100) semantics and avoids overflow with pct<=100.

jherrera-jump (approved):

PR #9510, opened 2026-04-27
poh: fix pacing, don't limit hashing as follower

jherrera-jump:
Fixes
>3. PoH Tile Crash at Leader Slot Boundary

from https://github.com/firedancer-io/firedancer/issues/9162

Also includes a fix for a pacing bug which I thought I had already fixed but it seems like I didn't.

### DIFF
--- a/src/discof/poh/fd_poh.c
+++ b/src/discof/poh/fd_poh.c
@@ -225,7 +225,8 @@ fd_poh_begin_leader( fd_poh_t * poh,
                      ulong      hashcnt_per_tick,
                      ulong      ticks_per_slot,
                      ulong      tick_duration_ns,
-                     ulong      max_microblocks_in_slot ) {
+                     ulong      max_microblocks_in_slot,
+                     long       slot_start_ns ) {
   FD_TEST( poh->state==STATE_FOLLOWER || poh->state==STATE_WAITING_FOR_BANK );
   FD_TEST( slot==poh->next_leader_slot );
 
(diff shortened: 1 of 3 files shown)


src/discof/poh/fd_poh.c
```diff
@@ -388,21 +390,32 @@ fd_poh_advance( fd_poh_t *          poh,
...
+       max_remaining_microblocks/(hashcnt_per_tick-1) rounded up.
+
+       However, if hashcnt_per_tick is 1 because we're in low power mode,
+       this should probably just be max_remaining_microblocks. */
+    ulong max_remaining_ticks_or_microblocks = max_remaining_microblocks;
+    if( FD_LIKELY( !low_power_mode ) ) max_remaining_ticks_or_microblocks += (max_remaining_microblocks+poh->hashcnt_per_tick-2UL)/(poh->hashcnt_per_tick-1UL);
+
+    restricted_hashcnt = fd_ulong_if( poh->hashcnt_per_slot>=max_remaining_ticks_or_microblocks, poh->hashcnt_per_slot-max_remaining_ticks_or_microblocks, 0UL );
+  } else {
+    max_remaining_microblocks = 0UL;
```
mmcgee-jump, line 416: Wouldn't this be `max_microblocks_per_slot` here though? Else you could hash all the way through and not have space?

mmcgee-jump (changes requested):
.

src/discof/poh/fd_poh.c
jherrera-jump: Are you suggesting
```suggestion
    max_remaining_microblocks  = poh->max_microblocks_per_slot;
```

That branch is when hashing for non-leader slots, so there's nothing to reserve space for.

mmcgee-jump (approved):

PR #9527, opened 2026-04-28
fuzz: fix sched_rdisp fuzzer

two-heart:


ptaffet-jump (approved):
If this is performance sensitive, I think there are better ways, but this is okay

two-heart:
Fixed in recent d7994fd18acb8a3f5b5d136f0e73cd00ae6dd5de

PR #9528, opened 2026-04-28
bn254: compare full 128-byte G2 decompress output

AaronAjose:
This patch updates three BN254 G2 decompress assertions to compare all 128 bytes, restoring full regression coverage for decompress output correctness.


0x0ece (approved):
Thank you for the fix!

PR #9548, opened 2026-04-29
flamenco, runtime: harness changes for direct mapping

topointon-jump:


src/flamenco/runtime/tests/fd_vm_harness.c
```diff
@@ -253,11 +253,9 @@ fd_solfuzz_pb_syscall_run( fd_solfuzz_runner_t * runner,
...
   /* Actually invoke the syscall */
   int syscall_err = syscall->func( vm, vm->reg[1], vm->reg[2], vm->reg[3], vm->reg[4], vm->reg[5], &vm->reg[0] );
-  int stack_pop_err = fd_instr_stack_pop( ctx->runtime, ctx->txn_out, ctx->instr );
-  if( FD_UNLIKELY( stack_pop_err ) ) {
-      FD_LOG_WARNING(( "instr stack pop err" ));
-      goto error;
-  }
+
+  /* Call fd_execute_instr_end to pop the stack and correctly handle stack pop errors. */
+  syscall_err = fd_execute_instr_end( vm->instr_ctx, ctx->instr, syscall_err );
```
mjain-jump: why are you calling `fd_execute_instr_end` and not `fd_instr_stack_pop`?

topointon-jump: `fd_execute_instr_end` has the exact semantics we want - we only want to override the error code with the stack pop error if the result of the execution is a success. otherwise we want to return the `syscall_err`.

this is also what `fd_execute_instr` uses to correctly propagate stack pop errors, so it seems more correct to use it here. happy to use `fd_instr_stack_pop` instead but i'd basically be copying the logic from `fd_execute_instr_end`

src/flamenco/runtime/tests/fd_vm_harness.c
```diff
@@ -343,17 +341,27 @@ fd_solfuzz_pb_syscall_run( fd_solfuzz_runner_t * runner,
...
+        vm->input_mem_regions_cnt,
+        &effects->input_data_regions,
+        &effects->input_data_regions_count,
+        (void *)tmp_end,
+        fd_ulong_sat_sub( output_end, tmp_end )
+    );
+
+    if( !!vm->input_mem_regions_cnt && !effects->input_data_regions ) {
+      goto error;
+    }
```
mjain-jump: this looks redundant

topointon-jump: this was already there (see diff above) I just moved it into this conditional. I think this is needed to return an error if the `output_bufsz < input_regions_total_sz` check in `fd_solfuzz_vm_load_from_input_regions` fails:

```rust
  if( FD_UNLIKELY(   input_regions_total_sz == 0
                  || output_bufsz < input_regions_total_sz ) ) {
    *output = NULL;
    *output_count = 0;
    return 0;
  }
```

in which case we would want to goto error instead of returning success

src/flamenco/runtime/tests/fd_vm_harness.c
topointon-jump: parking this change for now

src/flamenco/runtime/tests/fd_vm_harness.c
mjain-jump: this should kill the harness then, not silently fail. these cases should be fixed up against. i wont block you on this though

mjain-jump (approved):

PR #9579, opened 2026-04-30
gossvf: harden ping check

jherrera-jump:


src/discof/gossip/fd_gossvf_tile.c
```diff
@@ -580,7 +580,7 @@ is_ping_active( fd_gossvf_tile_ctx_t *  ctx,
 
   /* 3. If the node has actively ponged a ping, it is active */
   ping_t * ping = ping_map_ele_query( ctx->ping_map, pubkey, NULL, ctx->pings );
```
mmcgee-jump, line 582: is there a matching check in gossip tile itself IIRC?

mmcgee-jump (approved):

PR #9596, opened 2026-04-30
disco/stem: centralize dcache bounds check (closes #826)

### LINKED ISSUE #826: standardize dcache bounds checking for links
Since all links are now specified in the topology, we can automatically bounds check messages received on links (mtu, chunk, wmark etc) to make sure they are valid, rather than every tile needing to remember to do this in `during_frag` etc

Jotanune:
## Summary

Issue #826 asks for a single source of truth for dcache fragment bounds checking. Today every `during_frag` callback hand-rolls the same `chunk<chunk0 || chunk>wmark || sz>mtu` check, which is error-prone (multiple historical bugs and inconsistencies — some tiles forget the `sz` bound, others use a hard-coded constant instead of the link mtu) and bloats every callsite with topology plumbing that does not belong in tile business logic.

This PR moves the topology-level dcache bounds check into `fd_stem` and removes the redundant copies in 21 tiles, while preserving every application-level invariant that the old per-tile checks happened to enforce.

### `fd_stem` changes

- `fd_stem_tile_in_t` is extended with three new fields appended after the hot polling state: `chunk0`, `wmark`, `mtu`. The first cache line (mcache/depth/idx/seq/mline/fseq/accum) is unchanged so the steady-state poll loop is bit-for-bit identical; the new fields land in a second cache line that is only touched on the `FD_UNLIKELY` corruption path.
- `STEM_(run1)` gains three new arrays (`in_chunk0[]`, `in_wmark[]`, `in_mtu[]`). When a caller passes `NULL` the bounds check is disabled for that input (sentinel: `mtu==0`), so out-of-tree consumers (dump, quic_trace) keep working unchanged.
- `STEM_(run)` builds these arrays on the stack from `topo->links[]` for every input, looking up the dcache via `fd_dcache_compact_chunk0/wmark`. Links with `mtu==0` or `dcache==NULL` (sign-tile sensitive frags, etc.) are kept opted out via the sentinel.
- The check itself fires before `STEM_CALLBACK_DURING_FRAG` runs and aborts via `FD_LOG_ERR` (same severity as before — no warning regression).

### Per-tile cleanup

For each tile that previously performed the same check:

- The `chunk<chunk0 || chunk>wmark` portion is removed everywhere.
- The `sz>link->mtu` portion is removed where it was numerically equivalent to the link's declared mtu (e.g. `sz>USHORT_MAX` against a `pack_bank`/`pack_execle`/`poh_shred` link declared with `USHORT_MAX`).
- All **application-level** `sz` invariants are kept verbatim:
  - strict-equality size matches against fixed-layout messages (`sz!=sizeof(fd_became_leader_t)`, `sz!=sizeof(fd_shred_features_activation_t)`),
  - lower bounds protecting header parses (`sz<FD_PACK_REBATE_MIN_SZ`, `sz<sizeof(fd_entry_batch_meta_t)+sizeof(fd_entry_batch_header_t)`, `sz<sizeof(fd_eth_hdr_t)+sizeof(fd_ip4_hdr_t)` in xdp, `sz<eth+ip4+udp` in sock),
  - protocol MTUs that are conceptually distinct from the link mtu even when they currently match it (`sz>FD_TPU_RAW_MTU`, `sz>FD_TPU_RESOLVED_MTU`, `sz>FD_POH_SHRED_MTU`, `sz>FD_NET_MTU` in shred-from-repair, `sz>FD_ETH_PAYLOAD_MAX` in xdp).

  These are kept as defense-in-depth and clearly commented as such.

## Out of scope

Tiles that use `sz==0` as a sentinel (replay, gossip, solcap) are intentionally untouched in this PR to keep the diff focused and reviewable. Happy to follow up if reviewers want them included.

## Testing

- `make -j` (gcc 13) — clean.
- `make -j CC=clang firedancer-dev` (clang 18) — clean, no new warnings.
- Smoke tests in `unit-test/`: `test_disco_base`, `test_tango_base`, `test_tile`, `test_verify_tile`, `test_rpc_tile`, `test_xdp_tile` all pass.
- `test_repair_tile` and `test_tower_tile` fail with `mmap ENOMEM` due to insufficient hugepages in the local sandbox; this is preexisting and unrelated.

## Closes

Closes #826


Jotanune:
Addressed the automated review feedback (layout asserts, plugin bounds, and stem init) in the latest commit.

ripatel-fd:
AI slop

Jotanune:
I used an LLM to help me draft the PR description and format the block comments because English isn't my first language and I wanted to make a good impression on a 25-file core refactor. Lesson learned. However, the actual C logic and the architecture change in fd_stem are real. I spent days tracking the redundant chunk/wmark checks across all 21 tiles to consolidate them as requested in the issue.

ripatel-fd:
This PR was hardly a consolidation because it added back twice as many lines of code as it removed. We shouldn't have added the "good first issue" label on this issue, sorry about that, we do not accept contributions for this at the moment.

Jotanune:
> This PR was hardly a consolidation because it added back twice as many lines of code as it removed. We shouldn't have added the "good first issue" label on this issue, sorry about that, we do not accept contributions for this at the moment.

Okay, so should I keep looking at other issues? because I have not found other good first issues that I can solve. The other one is already assigned.

Issue #9587, opened 2026-04-30
features: properly handle revoked feature gates


topointon-jump:
Need to correctly handle revoked feature gates in all places. Including when the feature is revoked in the last slot of the epoch.

PR #9600, opened 2026-04-30
flamenco, features: handle revoke features in the same slot

topointon-jump:
https://github.com/firedancer-io/firedancer/issues/9587


topointon-jump:
https://github.com/firedancer-io/solfuzz-agave/pull/508

mjain-jump (approved):

PR #9626, opened 2026-05-02
perf: specialize blake3 xof2048 expansion

parasol-aser:
## What

Specialize BLAKE3 2048-byte XOF expansion on AVX2.

This adds an internal `fd_blake3_avx_xof2048` helper and routes
`fd_blake3_fini_2048` through it when `FD_HAS_AVX` is enabled. The
non-AVX and AVX512 paths keep their existing behavior. The PR also adds
a focused `test_blake3_xof2048` regression test.

## Why it is faster

`fd_blake3_fini_2048` expands one root block into 32 consecutive
64-byte XOF output blocks. The previous AVX2 path used the generic
`fd_blake3_avx_compress8` backend four times, rebuilding batch arrays
and paying for generic pointer, tail, custom-CV, lane, and store
handling each time.

The new helper uses the fixed XOF shape directly: one root message, one
root CV, consecutive counters, fixed flags, and 64-byte outputs. It
broadcasts the CV and message words, reuses the existing AVX2 round
function, and stores the 32 output blocks directly.

## Measurements

Benchmark command, unchanged from baseline:

```sh
make -j BASEDIR=build test_blake3 >/dev/null && build/native/gcc/unit-test/test_blake3
```

Each result used five repetitions under a 180 second timeout with
`/usr/bin/time -v`.

Runtime:

| Metric | Baseline mean +/- std | Candidate mean +/- std | Delta |
| --- | ---: | ---: | ---: |
| elapsed seconds | 67.360 +/- 20.800 | 42.732 +/- 2.097 | -36.6% |
| user seconds | 47.208 +/- 2.783 | 40.630 +/- 0.837 | -13.9% |
| system seconds | 1.474 +/- 0.177 | 1.242 +/- 0.053 | -15.7% |

Targeted XOF rows:

| Row | Baseline mean +/- std | Candidate mean +/- std | Delta |
| --- | ---: | ---: | ---: |
| incremental_xof_2048 sz=64 ns/byte | 27.7757 +/- 11.1000 | 12.2765 +/- 1.2600 | -55.8% |
| incremental_xof_2048 sz=128 ns/byte | 16.4043 +/- 7.9000 | 6.62081 +/- 0.2330 | -59.6% |
| incremental_xof_2048 sz=256 ns/byte | 8.23971 +/- 4.3300 | 3.90800 +/- 0.1310 | -52.6% |
| incremental_xof_2048 sz=512 ns/byte | 4.78275 +/- 2.4700 | 2.54016 +/- 0.0885 | -46.9% |
| incremental_xof_2048 sz=1024 ns/byte | 2.81312 +/- 1.1000 | 1.83266 +/- 0.0355 | -34.9% |
| incremental_xof_2048 sz=524288 ns/byte | 0.551824 +/- 0.2730 | 0.389709 +/- 0.0304 | -29.4% |

## Correctness

Passed:

```sh
make -j BASEDIR=build test_blake3
make -j BASEDIR=build check
make -j BASEDIR=build test_blake3 >/dev/null && build/native/gcc/unit-test/test_blake3
```

All five benchmark repetitions exited 0 and reached the final `pass`
line.

The full repo unit-test command was attempted but is blocked on this
host by missing gigantic pages:

```sh
make -j BASEDIR=build unit-test && make BASEDIR=build run-unit-test
```

The runner reported 0 free gigantic pages on NUMA 0 and exited with
`Not enough memory!`. This also happened on the baseline and other
candidates.

## Regression test

Added:

```sh
make -j BASEDIR=build test_blake3_xof2048 && build/native/gcc/unit-test/test_blake3_xof2048
```

The test runs `fd_blake3_fini_2048` for input sizes 0 through 65536 and
checks the expected accumulator `0x79836ea1df1a342aUL`.

## Branch and commits

Branch: `perf/B3-HOT-002`

Commits:

- `8fcedcfd9431d0becf5d12c80db1dd1060cb25c3` `perf: specialize blake3 xof2048 expansion`
- `3db419f29863c4450d5ad59f86c81cd564eabe1c` `tests: regression for B3-HOT-002`

## Caveats

This touches AVX2 cryptographic compression logic. Review should pay
particular attention to counter carry handling, flags, block length,
root CV loading, lane ordering, and storing both halves of each 64-byte
XOF output.

The benchmark baseline is noisy, so the claim here is limited to the
measured BLAKE3 XOF(2048) workload and the exact command above.

## Reproduction

Baseline:

```sh
git checkout c1417286c712de2227776459d406e2316782a781
make -j BASEDIR=build test_blake3
for rep in 1 2 3 4 5; do
  /usr/bin/time -v -o "baseline/benchmark_rep_${rep}.time.log" \
    timeout --foreground 180s \
    bash -lc 'make -j BASEDIR=build test_blake3 >/dev/null && build/native/gcc/unit-test/test_blake3' \
    > "baseline/benchmark_rep_${rep}.log" 2>&1
done
```

Candidate:

```sh
git checkout perf/B3-HOT-002
make -j BASEDIR=build test_blake3
make -j BASEDIR=build check
make -j BASEDIR=build test_blake3_xof2048 && build/native/gcc/unit-test/test_blake3_xof2048
for rep in 1 2 3 4 5; do
  /usr/bin/time -v -o "logs/B3-HOT-002/benchmark_rep_${rep}.time.log" \
    timeout --foreground 180s \
    bash -lc 'make -j BASEDIR=build test_blake3 >/dev/null && build/native/gcc/unit-test/test_blake3' \
    > "runs/B3-HOT-002/benchmark_rep_${rep}.log" 2>&1
done
```

### DIFF
--- a/src/ballet/blake3/Local.mk
+++ b/src/ballet/blake3/Local.mk
@@ -12,6 +12,8 @@ endif
 
 $(call make-unit-test,test_blake3,test_blake3,fd_ballet fd_util)
 $(call run-unit-test,test_blake3)
+$(call make-unit-test,test_blake3_xof2048,test_blake3_xof2048,fd_ballet fd_util)
+$(call run-unit-test,test_blake3_xof2048)
 ifdef FD_HAS_HOSTED
 $(call make-fuzz-test,fuzz_blake3,fuzz_blake3,fd_ballet fd_util)
 endif
--- a/src/ballet/blake3/fd_blake3.c
+++ b/src/ballet/blake3/fd_blake3.c
@@ -723,12 +723,9 @@ fd_blake3_fini_2048( fd_blake3_t * sha,
   FD_BLAKE3_TRACE(( "fd_blake3_fini_2048: sz=%lu ctr0=%lu flags=%x",
                     sha->pos.input_sz, ctr0, last_block_flags ));
 
-  /* Expand LtHash
-     For now, this uses the generic AVX2/AVX512 compress backend.
-     Could write a more optimized version in the future saving some of
-     the matrix transpose work. */
-  for( ulong i=0UL; i<32UL; i+=FD_BLAKE3_PARA_MAX ) {
+  /* Expand LtHash */
 #if FD_HAS_AVX512
+  for( ulong i=0UL; i<32UL; i+=FD_BLAKE3_PARA_MAX ) {
     ulong  batch_data [ 16 ] __attribute__((aligned(64)));
     /*                     */ for( ulong j=0; j<16; j++ ) batch_data [ j ] = (ulong)root_msg;
     uint   batch_sz   [ 16 ]; for( ulong j=0; j<16; j++ ) batch_sz   [ j ] = last_block_sz;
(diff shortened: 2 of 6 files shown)


ripatel-fd:
Thank you for the contribution. The numbers do look attractive but I'd like to do some further research whether this is worth doing. fd_blake3 is not a general-purpose library, so we only optimize towards actual uses in Firedancer. 

Currently, most XOF hashing is done via the LtHash multi-message hashing path during snapshot loading. fd_blake3_fini_2048 serves as a slower fallback when hashing large accounts only. If we start optimizing these fallback cases at the expense of more code, we'd expect to see more instruction cache misses, which could balance out raw throughput wins.

I will run this through some snapshot loading and replay scenarios to see. My guess is it's worth doing.

ripatel-fd:
The AVX512 code is analogous if I remember, so the same optimization could work there. We could also parametrize the compress{8,16}_fast functions to make them more flexible with how the input is advanced to make it work for XOF to reduce code footprint.

parasol-aser:
Thanks, that framing makes sense. I intentionally kept this PR scoped to the AVX2 XOF path because that was the path exercised on my test host, and I did not want to add an unmeasured AVX512 variant or a broader compress-fast refactor before knowing whether the fallback path matters in snapshot/replay workloads.

I agree the AVX512 code looks analogous, and a parameterized `compress{8,16}_fast`-style shape could be the better direction if this is worth carrying long-term. If your snapshot loading / replay runs show this is worthwhile but code footprint is the main objection, I can rework this toward a smaller shared fast path and include AVX512 measurements on appropriate hardware rather than keeping the current standalone AVX2 helper.

ripatel-fd:
> I agree the AVX512 code looks analogous

Would you mind adding the AVX512 optimization too then? I can run the benchmarks for you if you don't have an AVX512 box.  You can compile for AVX512 even if your host doesn't support it with `make -j MACHINE=linux_clang_zen5`. 

> a parameterized `compress{8,16}_fast`-style shape could be the better direction if this is worth carrying long-term

I experimented with this yesterday and it's a lot trickier than I thought it would be. Mainly because a compress step produces 32 bytes of output, and an expand step 64 bytes. Meaning more register pressure and the final transpose has a different shape (8x8 vs 16x8). So, I think your PR was exactly the right approach. 

When I get a bit more time I'll post a thorough review, but this looks structurally very good on first glance.

parasol-aser:
Added the AVX512 XOF2048 fast path in a4e06da67 and routed fd_blake3_fini_2048 to it under FD_HAS_AVX512.

Verified:
- native test_blake3_xof2048 build/run passes
- MACHINE=linux_clang_zen5 test_blake3_xof2048 compile passes
- local AVX512 execution check on this host's skylake-avx512 subset passes
- byte-for-byte comparison against the prior generic compress16 XOF path passes

I do not have Zen5 benchmark numbers locally, so I'll leave those to your AVX512 box.

PR #9640, opened 2026-05-06
gossip: parser hardening

jherrera-jump:


mmcgee-jump:
This doesn't harden anything, the CHECKs are redundant with the parser itself, which immediately fails if the arrays are too big anyway.

mmcgee-jump (changes requested):
.

PR #9644, opened 2026-05-07
restore: make AppendVec size fields optional

ripatel-fd:
Agave snapshots compress accounts in a messy format.  Some files
containing accounts ("AppendVec") have arbitrary trailing garbage
at the end.  The offset at which this garbage starts is stored in
length fields, separately in the snapshot manifest.

Currently, Firedancer truncates each of these files according to
these length fields.  It aborts the snapshot load if an AppendVec
file is found whose trailing garbage offset was not set in the
manifest.

This machinery is nonsensical and redundant.  This patch skips it
if there are no length fields in the snapshot (instead treating the
entire account file as valid).


drubin-fd (dismissed):
lgtm, forgot to update ele_cnt name in verify proto, and `mymap_verify` docstring

drubin-fd (approved):

PR #9712, opened 2026-05-10
restore: reject out-of-bounds txnhash_offset

ripatel-fd:
Fixes a snapshot DoS issue caused by OOB txnhash_offset values

#9705


ripatel-fd:
@mmcgee-jump Agave's slot parser silently clamps txnhash_offset to 11 to prevent OOB, maybe we should do the same?

mmcgee-jump (approved):

PR #9724, opened 2026-05-11
Revert "ci: remove fc-ng workflow"

intrigus-lgtm:
Reverts firedancer-io/firedancer#9722

This is run inside the [on_main_push](https://github.com/firedancer-io/firedancer/blob/8cad7287701203f3b1851e0211b54e0bbb083451/.github/workflows/on_main_push.yml#L14-L16) workflow.
E.g.: https://github.com/firedancer-io/firedancer/actions/runs/25673061513/job/75363239553


ripatel-fd:
#9725

PR #9736, opened 2026-05-11
pack: fix burst

jherrera-jump:


mmcgee-jump (dismissed):
Does this fix it? There's another publish above that might need same thing? And then my plain reading of this file should have STEM_BURST as 2?

ptaffet-jump:
I think burst of 2 is correct, but I think some of the "don't polls" are unnecessary, and the explanation is not quite right.

There are 6 places the tile publishes:
```
after_credit:
	slot times out. Sets ctx->leader_slot=ULONG_MAX. returns.
	done_draining. returns.
	reduce_mb_bound. returns.
	publish a transaction. *doesn't* return.
	run out of microblocks. Sets ctx->leader_slot=ULONG_MAX. returns.
after_frag:
	done packing. Requires ctx->leader_slot!=ULONG_MAX
```
There are a few paths that publish 2, but it's not possible to publish 3, even without setting `*opt_poll_in = 0;`
I think `/* 1 frag from after_credit + 1 frag from after_frag */` is not quite right though.

ptaffet-jump (approved):

PR #9743, opened 2026-05-11
pack: add extra filtering

jherrera-jump:
Not a security issue, just a slight mismatch in filtering which lets txns through that would later get caught by the runtime

### DIFF
--- a/src/disco/pack/test_compute_budget_program.c
+++ b/src/disco/pack/test_compute_budget_program.c
@@ -103,10 +103,52 @@ main( int     argc,
   FD_TEST( test_duplicate( 0, 1, 0, 1, 1 ) == 1 );
   FD_TEST( test_duplicate( 0, 0, 1, 1, 1 ) == 1 );
 
+  /* Test heap frame size bounds check.  RequestHeapFrame is instruction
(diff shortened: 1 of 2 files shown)


src/disco/pack/test_compute_budget_program.c
```diff
@@ -103,10 +103,47 @@ main( int     argc,
...
+    TEST_HEAP_FRAME( 128UL * 1024UL, 1 ); /* 128 KiB            */
+    TEST_HEAP_FRAME( 255UL * 1024UL, 1 ); /* 255 KiB            */
+
+    /* At maximum */
+    TEST_HEAP_FRAME( 256UL * 1024UL, 1 ); /* 256 KiB, exact max */
+
+    /* Above maximum (256 KiB) */
+    TEST_HEAP_FRAME( 257UL * 1024UL, 0 ); /* 257 KiB, just over */
+    TEST_HEAP_FRAME( 512UL * 1024UL, 0 ); /* 512 KiB            */
+
```
ptaffet-jump, line 145: Could you add a test for something that's not a multiple of 1kB?

ptaffet-jump (approved):

PR #9791, opened 2026-05-13
IBT and shadow stack support

ripatel-fd:
Update all compile units in Firedancer to support IBT and shadow stack.
IBT attempts to prevent hijacking of indirect call / jmp, and shadow stack hardens against ROP attacks.

Add a `fd_shstk` API for enabling shadow stacks even on old libc versions.

IBT is only available on Intel CPUs. 
Shadow stack is available on Intel Skylake-SP or AMD Zen 3 and newer, but requires at least Linux 6.4. 
Linux 6.6 makes shadow stack protections stronger. 

-------------

- **ballet: mark asm code as IBT/SHSTK compliant**
- **deps.sh: IBT/SHSTK support**
- **config: enable x86 IBT, shadow stack**
- **sandbox: add shadow stack support**
- **app: enable shadow stack for full Firedancer**

### DIFF
--- a/src/ballet/aes/fd_aes_gcm_aesni.S
+++ b/src/ballet/aes/fd_aes_gcm_aesni.S
@@ -1143,3 +1143,21 @@ SYM_FUNC_END(aes_gcm_enc_final_aesni_avx)
 SYM_FUNC_START(aes_gcm_dec_final_aesni_avx)
 	_aes_gcm_final	0
 SYM_FUNC_END(aes_gcm_dec_final_aesni_avx)
+
+# Advertise IBT, SHSTK compliance
+.section ".note.gnu.property", "a"
+.p2align 3
+.long 1f - 0f
+.long 4f - 1f
+.long 5
+0:
+.byte 0x47, 0x4e, 0x55, 0
+1:
+.p2align 3
+.long 0xc0000002
+.long 3f - 2f
+2:
+.long 3
+3:
+.p2align 3
+4:
--- a/src/ballet/reedsol/fd_reedsol_gfni_32.S
+++ b/src/ballet/reedsol/fd_reedsol_gfni_32.S
@@ -17,6 +17,7 @@ gfni_const_tbl:
 fd_reedsol_private_encode_32_32:
 .globl fd_reedsol_private_encode_32_32
 .cfi_startproc
+endbr64
 
 # This file implements the FFT-like O(n log n) algorithm for computing Reed
 # Solomon parity as described in
@@ -412,3 +413,21 @@ ret
 .align 16
 
 .cfi_endproc
+
+# Advertise IBT, SHSTK compliance
+.section ".note.gnu.property", "a"
+.p2align 3
+.long 1f - 0f
+.long 4f - 1f
+.long 5
+0:
+.byte 0x47, 0x4e, 0x55, 0
+1:
+.p2align 3
+.long 0xc0000002
+.long 3f - 2f
+2:
+.long 3
+3:
+.p2align 3
+4:
--- a/src/util/sandbox/fd_shstk.c
+++ b/src/util/sandbox/fd_shstk.c
@@ -1,0 +1,88 @@
+#define _GNU_SOURCE
+#include "fd_shstk.h"
+#include "../fd_util.h"
+#include <errno.h>
+#include <stdlib.h>
+/* GCC 11 complains about "unused parameters" on naked functions */
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+
+#if FD_HAS_HOSTED && defined(__linux__) && defined(__x86_64__) && !FD_HAS_ASAN && !FD_HAS_MSAN && !FD_HAS_UBSAN
+
+#ifndef ARCH_SHSTK_ENABLE
+#define ARCH_SHSTK_ENABLE 0x5001
+#endif
+
+#ifndef ARCH_SHSTK_LOCK
+#define ARCH_SHSTK_LOCK 0x5003
+#endif
+
+#ifndef ARCH_SHSTK_STATUS
+#define ARCH_SHSTK_STATUS 0x5005
+#endif
+
+#ifndef ARCH_SHSTK_SHSTK
+#define ARCH_SHSTK_SHSTK 1
+#endif
+
+#include <asm/prctl.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+__attribute__((noreturn))
+void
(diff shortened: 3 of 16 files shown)


ripatel-fd:
Some older versions of lld don't like `-z ibt` 

Someone also needs to audit firedancer tile processes run with shadow stack enabled

Also audit cryptographic code for possible missing endbr64 (needs to be run on an Intel CPU)

src/util/sandbox/fd_shstk.c
```diff
@@ -0,0 +1,102 @@
...
+    "mov %rax, %rdi;\n"
+    /* additional config (stack%16==0 here) */
+    "call fd_shstk_secure;\n"
+    /* call main_fn */
+    "pop %rsi;\n" /* argv */
+    "pop %rdi;\n" /* argc */
+    "pop %rax;\n" /* main_fn */
+    "sub $8, %rsp;\n" /* align stack */
+    "call *%rax;\n"
+    "add $8, %rsp;\n"
```
two-heart: This is wrong here and leads to calling into exit misaligned. The stack already had the correct alignment att the call to rax.

src/ballet/aes/fd_aes_gcm_aesni.S
```diff
@@ -1143,3 +1143,21 @@ SYM_FUNC_END(aes_gcm_enc_final_aesni_avx)
 SYM_FUNC_START(aes_gcm_dec_final_aesni_avx)
 	_aes_gcm_final	0
 SYM_FUNC_END(aes_gcm_dec_final_aesni_avx)
+
```
two-heart, line 1146: consider defining this at a central point and then including

src/util/sandbox/fd_shstk.c
```diff
@@ -0,0 +1,102 @@
...
+void
+fd_shstk_secure( long enable_res ) {
+  if( enable_res != 0 ) {
+    FD_LOG_WARNING(( "sandbox: failed to enable shadow stack (%ld-%s), running with weakened sandbox", -enable_res, fd_io_strerror( -(int)enable_res ) ));
+    return;
+  }
+
+  int rc = (int)syscall( SYS_arch_prctl, ARCH_SHSTK_LOCK, ARCH_SHSTK_SHSTK );
+  if( FD_UNLIKELY( rc ) ) FD_LOG_WARNING(( "sandbox: failed to enforce shadow stack (arch_prctl(ARCH_SHSTK_LOCK,ARCH_SHSTK_SHSTK) failed (%i-%s))", errno, fd_io_strerror( errno ) ));
+  FD_LOG_INFO(( "sandbox: shadow stack enabled" ));
```
two-heart: A bit spammy to do this per tile. More than half of the log output at startup now is repeated:
```
INFO    05-13 12:55:04.600539 134    16   0    src/util/sandbox/fd_shstk.c(85): sandbox: shadow stack not enabled by libc, enabling
INFO    05-13 12:55:04.600587 134    16   0    src/util/sandbox/fd_shstk.c(36): sandbox: shadow stack enabled
```

src/util/sandbox/fd_shstk.c
```diff
@@ -0,0 +1,102 @@
...
+    "pop %rdi;\n" /* argc */
+    "pop %rax;\n" /* main_fn */
+    "sub $8, %rsp;\n" /* align stack */
+    "call *%rax;\n"
+    "add $8, %rsp;\n"
+    /* call exit(3) with return value */
+    "mov %rax, %rdi;\n"
+    "call exit;\n"
+  );
+}
```
intrigus-lgtm: Wouldn't something like this also work?
It would fix the misaligned call to exit and allow us to get rid of the naked function and reduce the amount of asm we need to use and also better for future extensions.

```
__attribute__((noreturn))
void
fd_shstk_gate( int (* main_fn)( int     argc,
                                char ** argv ),
               int     argc,
               char ** argv ) {
  /* cannot use `syscall` call as that would crash on return due to missing shadow stack entry */
  long enable_res;
  __asm__ volatile ( "syscall"
                     : "=a"(enable_res)
                     : "a"((long)SYS_arch_prctl),
                       "D"((long)ARCH_SHSTK_ENABLE),
                       "S"((long)ARCH_SHSTK_SHSTK)
                     : "rcx", "r11", "memory" );
  /* a=rax, d=rdx, D=rdi, S=rsi */
  fd_shstk_secure(enable_res);
  exit( main_fn( argc, argv ) );
}
```

src/util/sandbox/fd_shstk.c
```diff
@@ -0,0 +1,102 @@
+#define _GNU_SOURCE
+#include "fd_shstk.h"
+#include "../fd_util.h"
+#include <errno.h>
+#include <stdlib.h>
+/* GCC 11 complains about "unused parameters" on naked functions */
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+
+#if FD_HAS_HOSTED && defined(__linux__) && defined(__x86_64__) && !FD_HAS_ASAN && !FD_HAS_MSAN && !FD_HAS_UBSAN
+
```
intrigus-lgtm, line 10: Add this if you consider applying https://github.com/firedancer-io/firedancer/pull/9791/changes#r3235720196
```
#ifndef ARCH_SHSTK_ENABLE
#define ARCH_SHSTK_ENABLE 0x5001
#endif
```

src/ballet/reedsol/fd_reedsol_gfni_32.S
```diff
@@ -17,6 +17,7 @@ gfni_const_tbl:
 fd_reedsol_private_encode_32_32:
 .globl fd_reedsol_private_encode_32_32
 .cfi_startproc
+endbr64
```
intrigus-lgtm, line 20: Not really that important, but I don't think this endbr64 is actually needed, as as far as I've checked, no one calls fd_reedsol_private_encode_32_32 using an indirect branch.
(But also not a big problem that people can call this function using an indirect branch)
Otherwise, we'd also have to add a few missing endbr64s to src/ballet/aes/fd_aes_base_aesni.S (e.g. fd_aesni_set_decrypt_key)

two-heart:
Ran a check with https://github.com/torvalds/linux/blob/master/tools/objtool/Documentation/objtool.txt, no findings

src/util/sandbox/fd_shstk.c
ripatel-fd: You're right that would indeed work

ripatel-fd:
Moving to #9982, which are the same patches but without IBT, which is not yet supported by Linux.

PR #9802, opened 2026-05-13
feat: add grpc actor-based fuzzer

intrigus-lgtm:
Would have found https://github.com/firedancer-io/auditor-internal/issues/510


ripatel-fd (approved):
Sorry, I missed this

PR #9805, opened 2026-05-13
runtime: remove capitalization underflow check

mjain-jump:


ibhatt-jumptrading (approved):
stamping to not block

PR #9807, opened 2026-05-13
tower: never vote on snapshot slot

### LINKED ISSUE #9620: firedancer shouldn't vote on the snapshot slot

kbhargava-jump:


src/choreo/tower/fd_tower.c
```diff
@@ -694,6 +713,20 @@ fd_tower_vote_and_reset( fd_tower_t * tower,
 
   if( FD_UNLIKELY( fd_tower_vote_empty( tower->votes ) ) ) {
     FD_BASE58_ENCODE_32_BYTES( best_blk->id.uc, best_blk_id );
+
+    /* is_recent: on snapshot boot, tower->root is set to the snapshot slot
+       before any votes are recorded. Refuse to vote for slot <= root. */
+    if( FD_UNLIKELY( !is_recent( tower, best_blk->slot ) ) ) {
```
lidatong: this is not correct way to fix. check `ctx->init` instead, and set it after

src/discof/tower/fd_tower_tile.c
```diff
@@ -620,7 +620,7 @@ publish_slot_done( fd_tower_tile_t *            ctx,
   ulong       authority_idx = ULONG_MAX;
   fd_pubkey_t authority[1];
   int         found_authority = deser_auth_vtr( ctx, slot_completed->epoch, found, authority, &authority_idx );
-  if( FD_LIKELY( found_authority ) ) {
+  if( FD_LIKELY( found_authority && out->vote_slot!=ULONG_MAX ) ) {
```
lidatong: also not correct. see the block comment

```
  /* This always contains a vote transaction with our current tower,
     regardless of whether there is a new vote slot or not (ie. vote
     slot can be ULONG_MAX and vote_txn will contain a txn of our
     current tower).  The vote is not yet signed.  This is necessary to
     support refreshing our last vote, ie. we retransmit our vote even
     when we are locked out / can't switch vote forks.
```

lidatong (changes requested):
see comments. thanks for fixing!

src/choreo/tower/fd_tower.c
```diff
@@ -401,8 +421,7 @@ static int
 lockout_check( fd_tower_t * tower,
                ulong        slot ) {
 
-  if( FD_UNLIKELY( fd_tower_vote_empty( tower->votes )                         ) ) return 1; /* always not locked out if we haven't voted. */
-  if( FD_UNLIKELY( slot <= fd_tower_vote_peek_tail_const( tower->votes )->slot ) ) return 0; /* always locked out from voting for slot <= last vote slot */
+  if( FD_UNLIKELY( !is_recent( tower, slot ) ) ) return 0; /* locked out: slot is not recent */
```
lidatong: remove the helper and just inline this check

src/choreo/tower/fd_tower.c
lidatong: actually dont even think you need the init thing. it will already determine it's locked out given the above

kbhargava-jump: split into case 0a and 0b

lidatong (approved):

Issue #9827, opened 2026-05-15
Build failure on Arch Linux due to unused variable warning treated as error


Unique-Usman:
Environment:

OS: Arch Linux
GCC version: gcc (GCC) 16.1.1 20260430
Firedancer branch/commit:

Error:
src/disco/gui/fd_gui_printf.c:1414:204: error:
variable ‘j’ set but not used [-Werror=unused-but-set-variable=]

https://github.com/firedancer-io/firedancer/blob/main/src/disco/gui/fd_gui_printf.c#L1414

Cause:
The variable j is initialized but never used, and the build uses -Werror, causing compilation to fail on newer GCC versions.

HGuillemet:
Even GCC 15 is not usable currently (at least on zen5).
You should stick with GCC 14.
That said, GCC 16 error seems valid. `j` is unused and even not declared.

Unique-Usman:
Noted, thanks.

PR #9828, opened 2026-05-15
gui: fix unused variable 'j' in fd_gui_peers_printf_node_all

Unique-Usman:
Remove the variable 'j' from the peer iteration loop. It was initialized and incremented but never used, causing a build failure when compiled with -Werror=unused-but-set-variable.


Unique-Usman:
Fix: #9827

ripatel-fd (approved):
Thank you

PR #9832, opened 2026-05-15
book: switch from bun to node

ripatel-fd:


mmcgee-jump:
Why churning this? Seems to work fine?

ripatel-fd:
> Why churning this? Seems to work fine?

@mmcgee-jump We already have a JS runtime for the frontend which is going to stay, always good to reduce amount of deps

PR #9837, opened 2026-05-15
Switch to yyjson

ripatel-fd:
- **ballet: vendor yyjson**
- **Migrate cJSON users to yyjson**
- **Remove cJSON**

The main goal is to remove a libc alloc dependency


mmcgee-jump:
Nice PR! But please hold this for next release branch

two-heart (changes requested):
Request parsing correctly uses a 128-KiB stack-backed yyjson pool. Response and
error formatting does not. The new macro at
[`fd_rpc_tile.c`](https://github.com/firedancer-io/firedancer/blob/90b9f2a32e14d6a4a19ddca9ddc7cc01ede86529/src/discof/rpc/fd_rpc_tile.c#L144-L154)
passes a null allocator:

```c
static void fd_rpc_cstr_free( char ** p ) { free( *p ); }
#define CSTR_JSON(__json, __out) \
  __attribute__((cleanup(fd_rpc_cstr_free))) char * __out = \
    yyjson_val_write( (__json), YYJSON_WRITE_NOFLAG, NULL )
```

For yyjson, a null writer allocator selects the default `malloc`/`realloc`/
`free` allocator. The PR therefore removed the explicit cJSON `fd_alloc` pool
but added an implicit libc allocation on almost every response and validation
error that echoes an ID or offending value.

The RPC tile's generated
[seccomp filter](https://github.com/firedancer-io/firedancer/blob/90b9f2a32e14d6a4a19ddca9ddc7cc01ede86529/src/discof/rpc/generated/fd_rpc_tile_seccomp.h#L29-L125)
allows selected `write`, `fsync`, `accept4`, `read`, `sendto`, `close`, and
`ppoll` calls. It does not allow `brk`, `mmap`, `mremap`, or `munmap`; any
unmatched syscall reaches `SECCOMP_RET_KILL_PROCESS`.

Consequently, the writer succeeds only while libc can satisfy the request from
already-mapped arena memory. If allocator initialization or arena growth
requires a syscall, the sandbox kills the RPC process. This is state-dependent,
so the source audit does not claim that every request deterministically kills a
default deployment.

PR #9849, opened 2026-05-18
feature: commission_rate_in_basis_points

mjain-jump:


src/flamenco/stakes/fd_top_votes.c
```diff
@@ -21,7 +21,7 @@ struct vote_ele {
   ulong       stake;
```
ibhatt-jumptrading, line 21: is it possible to reorder members to shrunk struct size?

mjain-jump: unfortunately not by just reordering

ibhatt-jumptrading (approved):

PR #9874, opened 2026-05-19
events: fix circq correctness issue

ripatel-fd:
Fixes a bug where circq did not correctly recover from loss due
to cache eviction or manual removals.


ripatel-fd:
@mmcgee-jump API documentation is not very clear what happens if the "internal cursor" (the logical consumer?) is so slow that push_back (producer) overruns it due to tail eviction

mmcgee-jump (approved):

PR #9881, opened 2026-05-20
sysvar: better epoch rewards validation

ripatel-fd:


ripatel-fd:
@mjain-jump I think this test vector has an invalid epoch rewards sysvar. Can we fix up solfuzz and remove the invalid test vectors?

mjain-jump:
> @mjain-jump I think this test vector has an invalid epoch rewards sysvar. Can we fix up solfuzz and remove the invalid test vectors?

if the account state is invalid (unable to be decoded) then you can remove it from test-vectors and update the commit sha

ripatel-fd:
> if the account state is invalid (unable to be decoded) then you can remove it from test-vectors and update the commit sha

@mjain-jump Yeah, but how do I verify this? How do I decode an entry in test-vectors

PR #9882, opened 2026-05-20
rewards: validate stake_rewards max_fork_width

ripatel-fd:
fd_stake_rewards_new did not validate the max_fork_width param
despite low internal limits. Fixes a bug where a config file with
max_fork_width > 128 could cause data corruption.

Reported-by: https://v12.sh/

### DIFF
--- a/src/flamenco/rewards/fd_stake_rewards.c
+++ b/src/flamenco/rewards/fd_stake_rewards.c
@@ -2,8 +2,6 @@
 #include "fd_rewards_base.h"
 #include "../../ballet/siphash13/fd_siphash13.h"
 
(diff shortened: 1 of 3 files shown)


src/flamenco/rewards/fd_stake_rewards.c
```diff
@@ -2,8 +2,6 @@
 #include "fd_rewards_base.h"
 #include "../../ballet/siphash13/fd_siphash13.h"
 
-#define MAX_SUPPORTED_FORKS (128UL)
```
mmcgee-jump, line 5: Should this just take the parameter from config?

PR #9894, opened 2026-05-21
bank: update capitalization without overflow checks

mjain-jump:


src/flamenco/runtime/fd_accdb_svm.c
```diff
@@ -113,12 +95,7 @@ fd_accdb_svm_credit( fd_accdb_user_t *         accdb,
   }
   fd_accdb_ref_lamports_set( rw, lamports );
 
-  ulong * cap = &bank->f.capitalization;
```
mmcgee-jump, line 116: Is this really the best way to fix the fuzzer? This is making production code worse and less safe, we would rather crash than overflow capitalization.

mjain-jump: other option is to constrain the mutator to cap the capitalization values. this fd change is technically more "correct" as its exactly what agave does, but we can change the fuzzer if this isnt preferred

mjain-jump: i can also upstream something to agave to force them to use saturating adds here for safeness

mjain-jump:
closing this for now in favor of a mutator change to constrain capitalization

PR #9899, opened 2026-05-21
gossip-cmd: output cluster info to json

jherrera-jump:
This will allow us to do a one-shot gossip scrape for our fd.validators indexer


jherrera-jump:
not planned

PR #9902, opened 2026-05-21
gossvf: stake-weighted prioritization

### LINKED ISSUE #9569: gossvf: move verify_signatures before verify_addresses

jherrera-jump:
closes https://github.com/firedancer-io/firedancer/issues/9569


mmcgee-jump:
Seems fine, but curious what was the bug?

jherrera-jump:
it was the report we saw where an attacker can include a victims pubkey/addr in the payload (which can trigger a ping pre-address-validation) and then can cause it to ping itself with loopback amplification.

PR #9906, opened 2026-05-22
bundle: minor crank fixes

ripatel-fd:
- fix TOCTOU issue in bundle_crank_generate
- fix typo in test_bundle_crank


ptaffet-jump (changes requested):
The bug in the test seems real, but the map removal is actually fine. If you look carefully at the map insertion code, it never moves elements. Only deletion moves elements. That means that if you delete the most recently inserted element, it is guaranteed not to move any elements, so all indices other than the deleted one stay valid. You can extend that as long as you are careful to delete in the opposite order you inserted. I'm pretty sure this code does that properly.

mmcgee-jump:
Can we add a comment about this so future AI tools don't flag?

PR #9909, opened 2026-05-22
txsend: fix conn mgmt bugs

ripatel-fd:


src/discof/txsend/fd_txsend_tile.c
```diff
@@ -403,31 +447,37 @@ handle_contact_info_update( fd_txsend_tile_t *                 ctx,
...
 }
 
 static inline void
 handle_contact_info_remove( fd_txsend_tile_t *                 ctx,
                             fd_gossip_update_message_t const * msg ) {
+  if( FD_UNLIKELY( msg->contact_info_remove->idx>=FD_CONTACT_INFO_TABLE_SIZE ) )
+    FD_LOG_ERR(( "unexpected remove_contact_info_idx %lu >= %lu", msg->contact_info_remove->idx, FD_CONTACT_INFO_TABLE_SIZE ));
+
   peer_entry_t * entry = &ctx->peers[ msg->contact_info_remove->idx ];
-  entry->tombstoned = 1;
```
mmcgee-jump, line 430: I intentionally did this tombstoning logic so that if someone drops from gossip temporarily we still send them transactions. This does happen on occasion.

src/discof/txsend/fd_txsend_tile.c
```diff
@@ -403,31 +447,37 @@ handle_contact_info_update( fd_txsend_tile_t *                 ctx,
...
+  if( FD_LIKELY( entry->in_map && !memcmp( entry->pubkey.uc, msg->origin, 32UL ) ) ) {
```
mmcgee-jump, line 474: You don't need to check pubkey here, it's guaranteed the same as the added one by gossip tile

PR #9932, opened 2026-05-22
restore: remove snapshot lthash verification pipeline

mmcgee-jump:
Deletes the snapla (lthash accumulator) and snapls (lthash sum) tiles along with their config flags (development.snapshots.disable_lthash_verification, layout.snapshot_hash_tile_count), CLI flags, seccomp policies, generated metrics, and the FD_SNAPSHOT_HASH_MSG_* control protocol used to feed account hashes between snapin and the verifier tiles.

The runtime lthash code under src/ballet/lthash and the live bank-hash computation are unaffected; only the snapshot-load-time verification is removed.


ripatel-fd (changes requested):
Please remove SNAPLA and SNAPLS from src/disco/metrics/generate/types.py

src/discof/restore/fd_snapin_tile.c
jvarela-jump:
```suggestion
#define STEM_BURST 1UL
```
Removing `snapls` means that the worst-case for the `snapin_ct` link is 1 output message (instead of 2 for `snapin_ls`).

ripatel-fd (approved):

Issue #9588, opened 2026-04-30
restore: review recoverable errors in snapshot load pipeline


jvarela-jump:
Review all tiles and utils in the snapshot load pipeline regarding recoverable errors that may still be treated as FD_TEST, FD_LOG_ERR, FD_LOG_CRIT.

jvarela-jump:
Up-to-date with current pipeline architecture - reopen at a later day if needed.

PR #9939, opened 2026-05-26
restore: snapin data_len bounds check for non-batch account headers

jvarela-jump:
Adding `data_len` bounds check for non-batch account headers in `snapin` tile.

Partially addresses https://github.com/firedancer-io/firedancer/issues/9588.


ripatel-fd (approved):

jvarela-jump:
Already checked in ssparse (lines 624 and 547)

PR #9944, opened 2026-05-26
keyguard: make keypage readonly

ptaffet-jump:


mmcgee-jump (dismissed):

ptaffet-jump:
I think this is going to break set-identity... It's a bit more to plumb through

ptaffet-jump:
Dup of https://github.com/firedancer-io/firedancer/pull/9887

PR #9945, opened 2026-05-26
feature: loader_v3_minimum_extend_program_size

mjain-jump:


src/flamenco/runtime/program/fd_bpf_loader_program.c
```diff
@@ -750,6 +750,20 @@ common_extend_program( fd_exec_instr_ctx_t * instr_ctx,
     return FD_EXECUTOR_INSTR_ERR_INVALID_REALLOC;
   }
 
+  /* https://github.com/anza-xyz/agave/blob/v4.1.0-beta.1/programs/bpf_loader/src/lib.rs#L861-L883 */
+  if( FD_FEATURE_ACTIVE_BANK( instr_ctx->bank, loader_v3_minimum_extend_program_size ) ) {
+    ulong headroom = fd_ulong_sat_sub( MAX_PERMITTED_DATA_LENGTH, old_len );
+    if( FD_UNLIKELY( additional_bytes<MINIMUM_EXTEND_PROGRAM_BYTES && additional_bytes!=headroom ) ) {
+      fd_log_collector_printf_dangerous_max_127(
+        instr_ctx,
+        "ExtendProgram requires a minimum of %lu additional bytes or to extend to maximum size, but only %u were requested",
```
topointon-jump, line 760: nit but we should probably add a comment justifying if the length is safe, like we do for all the other `fd_log_collector_printf_dangerous_max_127` calls

topointon-jump, line 760: I think we should use `fd_log_collector_printf_dangerous_128_to_2k` here, as this can overflow 127 with the worse-case size of `%lu` (`ULONG_MAX`, 20 chars) and `%u `(`UINT_MAX`, 10 chars).

len("ExtendProgram requires a minimum of %lu additional bytes or to extend to maximum size, but only %u were requested") - len("%lu") - len("%u") + len(ULONG_MAX) + len(UINT_MAX)
= 113 - 3 - 2 + 20 + 10
= 138

mjain-jump: fixed

mjain-jump: no, this is fine. `MINIMUM_EXTEND_PROGRAM_BYTES` is hardcoded to 10kb (5 digits) and `additional_bytes` is capped to UINT_MAX (10 digits) so its 113 - 3 - 2 + 5 + 10 = 123 < 127

topointon-jump (approved):

PR #9950, opened 2026-05-27
neigh: fix prober rate calculation

two-heart:
rate is tokens/tick, previously we probed to much by a factor of tick_per_ns**2

### DIFF
--- a/src/waltz/neigh/fd_neigh4_probe.c
+++ b/src/waltz/neigh/fd_neigh4_probe.c
@@ -42,7 +42,7 @@ fd_neigh4_prober_init( fd_neigh4_prober_t * prober,
     .probe_delay = (long)( tick_per_ns * probe_delay_seconds * 1e9f ),
     .rate_limit  = (fd_token_bucket_t) {
       .ts      = fd_tickcount(),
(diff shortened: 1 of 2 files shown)


ripatel-fd (approved):

src/waltz/neigh/fd_neigh4_probe.c
```diff
@@ -42,7 +42,7 @@ fd_neigh4_prober_init( fd_neigh4_prober_t * prober,
     .probe_delay = (long)( tick_per_ns * probe_delay_seconds * 1e9f ),
     .rate_limit  = (fd_token_bucket_t) {
       .ts      = fd_tickcount(),
-      .rate    = tick_per_ns * (max_probes_per_second / 1e9f),
+      .rate    = max_probes_per_second * (tick_per_ns / 1e9f),
```
ripatel-fd:
```suggestion
      .rate    = max_probes_per_second / (tick_per_ns * 1e9f),
```

ripatel-fd (changes requested):

ripatel-fd (approved):

PR #9973, opened 2026-05-28
ARM-optimized cryptography using NEON SIMD intrinsics

cmoyes-jump:
Adds ARM NEON backends for blake3, chacha, sha256, and sha512 using ARMv8 crypto extensions where available. No x86 code paths are affected.

Follow-ups:
- blake3: add compress4 NEON batching for lthash
- chacha: shuffle/permute transpose instead of using scalar lanes

This is alpha, we shouldn't call ARM first-class support yet until further tested. Plenty of cycle-count shaving to go.

### DIFF
--- a/config/extra/with-arm.mk
+++ b/config/extra/with-arm.mk
@@ -44,3 +44,27 @@ FD_ARCH_SUPPORTS_SANDBOX:=1
 
 CPPFLAGS+=-DFD_HAS_ARM=1
 FD_HAS_ARM:=1
+
+CPPFLAGS+=-DFD_HAS_NEON=1
+FD_HAS_NEON:=1
+
+define _arm_map_define
+  ifeq ($(shell echo | $(CC) $(CPPFLAGS) -E -dM - | grep -c $(2)),1)
+    CPPFLAGS+=-D$(1)=1
+    $(1):=1
+  endif
+endef
+
+arm-map-define = $(eval $(call _arm_map_define,$(1),$(2)))
+
+$(call arm-map-define,FD_HAS_ARM_AES,__ARM_FEATURE_AES)
+$(call arm-map-define,FD_HAS_ARM_CRYPTO,__ARM_FEATURE_CRYPTO)
+$(call arm-map-define,FD_HAS_ARM_SHA256,__ARM_FEATURE_SHA2)
+$(call arm-map-define,FD_HAS_ARM_SHA512,__ARM_FEATURE_SHA512)
+
+ifdef FD_HAS_ARM_AES
+ifndef FD_HAS_ARM_CRYPTO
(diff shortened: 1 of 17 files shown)


ripatel-fd:
@cmoyes-jump Looks great 🎉  I also opened #9976 to make multi-arch feature support easier

config/extra/with-arm.mk
```diff
@@ -44,3 +44,27 @@ FD_ARCH_SUPPORTS_SANDBOX:=1
...
+    $(1):=1
+  endif
+endef
+
+arm-map-define = $(eval $(call _arm_map_define,$(1),$(2)))
+
+$(call arm-map-define,FD_HAS_ARM_AES,__ARM_FEATURE_AES)
+$(call arm-map-define,FD_HAS_ARM_CRYPTO,__ARM_FEATURE_CRYPTO)
+$(call arm-map-define,FD_HAS_ARM_SHA256,__ARM_FEATURE_SHA2)
+$(call arm-map-define,FD_HAS_ARM_SHA512,__ARM_FEATURE_SHA512)
```
ripatel-fd, line 63: Can we remove this feature detection code for now? I'm not sure if this is safe to run in `with-arm.mk`, it's better to run this in the machine target (like `native.mk`) instead.

However, we don't support bare `make` (MACHINE=native) for Arm either, since changes to that file risk breaking x86 users. 

Therefore, the easiest thing to do is let the machine target hardcode feature support in `config/machine/linux_gcc_neoverse_n2.mk`.

config/extra/with-arm.mk
```diff
@@ -44,3 +44,27 @@ FD_ARCH_SUPPORTS_SANDBOX:=1
...
+
+ifdef FD_HAS_ARM_AES
+ifndef FD_HAS_ARM_CRYPTO
+CPPFLAGS+=-DFD_HAS_ARM_CRYPTO=1
+FD_HAS_ARM_CRYPTO:=1
+endif
+endif
```
ripatel-fd, line 70: would remove this too

config/machine/native.mk
ripatel-fd, line 1: Could you revert changes to native.mk for now?
I suggest keeping Arm out of default build configuration until things are more stable.
(And just use `make -j MACHINE=linux_gcc_neoverse_n2` for now

ripatel-fd:
@cmoyes-jump Looks great 🎉  I also opened #9976 to make multi-arch feature support easier.
Could you split the changes into separate commits so it's easier to review/bisect also? (I'm happy to do this for you and preserve Git blame if you're busy). Something like this:
- config: arm crypto feature detection
- ballet: armv8 accelerated SHA-2
- ballet: armv8 accelerated ChaCha
- ballet: armv8 accelerated BLAKE3

ripatel-fd:
I am going to rebase this while retaining commit authorship

PR #9976, opened 2026-05-28
util: separate arch-specific code

ripatel-fd:
- Also remove fd_memeq x86 arch implementation
- Also fix FD_ATOMIC_XCHG generic version


src/util/fd_util_base.h
```diff
@@ -1229,32 +1089,7 @@ fd_memzero_explicit( void * d,
...
-  __asm__( "test %3, %3;"
-           "repe cmpsb"
-         : "=@cce" (r), "+S" (s0), "+D" (s1), "+c" (sz)
-         : "m" (*(char const (*)[sz]) s0), "m" (*(char const (*)[sz]) s1)
-         : "cc" );
-  return r;
-}
-
-#else
+   is zero (pointers may be invalid).. */
```
ripatel-fd, line 1092:
```suggestion
   is zero (pointers may be invalid). */
```

PR #9871, opened 2026-05-19
repair: cycle bad peers

emwang-jump:
fixes https://github.com/firedancer-io/firedancer/issues/9689

PR #9979, opened 2026-05-28
Fix repair dedup key packing

intrigus-lgtm:
Mask the packed kind, slot, and shred index fields so sentinel shred indexes cannot spill into the slot bits.
Add a regression for orphan and highest-shred requests on adjacent slots.


ripatel-fd (approved):

intrigus-lgtm:
It appears that this fixes https://github.com/firedancer-io/firedancer/issues/9689 but this PR is a duplicate of https://github.com/firedancer-io/firedancer/pull/9871 which I didn't know at the time.
Closing.

Issue #9799, opened 2026-05-13
far repair stall


ripatel-fd:
after repairing far for a while, replay gets stuck,
shred is backpressured on shred->repair link 
repair tile is 100% busy 

<img width="983" height="786" alt="Image" src="https://github.com/user-attachments/assets/99d81a48-b1b3-46c6-bc6d-0386b09ec8c1" />

will try to grab a stack trace

ripatel-fd:
```
(gdb) bt
#0  0x0000555557e50d2d in fd_forest_orphaned_iter_ele (pool=0x4bc2dd502880, join=0x4bc2ef61a8c8, iter=...) at src/discof/forest/../../util/tmpl/fd_map_chain.c:605
#1  evict (new_slot=408055377, parent_slot=408055376, forest=0x4bc2dd502680) at src/discof/forest/fd_forest.c:792
#2  acquire (evicted=0x5d30229d5088, parent_slot=408055376, slot=408055377, forest=0x4bc2dd502680) at src/discof/forest/fd_forest.c:887
#3  fd_forest_blk_insert (forest=0x4bc2dd502680, slot=408055377, parent_slot=408055376, evicted=evicted@entry=0x5d30229d5088) at src/discof/forest/fd_forest.c:961
```

seems to be eviction loop

Issue #9786, opened 2026-05-12
repair logs 32k lines per second when catching up far


ripatel-fd:
```
💬 GOSSIP...... RX  23.0 Mbit TX 118.1 Kbit CRDS      8.6 K PEERS        691 BUSY  18% BACKP   0%
🧱 REPAIR...... RX     0 bits TX   1.3 Mbit REPAIR SLOT 407586307 (-375489) TURBINE SLOT 407961796
💥 REPLAY...... SLOT 407583899 (-377897) CU/s     17.1 M TPS          - SPS       13.0 LEADER IN never ROOT DIST 31 BANKS 147
📡 EVENT....... STATE disconnected QUEUE          0 SENT        0.0 /s ACKED        0.0 /s BW     0 bits in     0 bits out DROPS          0 FULL   0%
```

```
INFO    2026-05-12 22:39:04.603246419 GMT+00 3361355:3361403 ripatel:tsfra2-ossdev-firedancer50:21   fd1:[group]:repair:0 src/discof/forest/fd_forest.c(1444)[fd_forest_fec_clear]: [fd_forest_fec_clear] cleared slot 407949633 fec set 128
INFO    2026-05-12 22:39:04.603247929 GMT+00 3361353:3361406 ripatel:tsfra2-ossdev-firedancer50:16   fd1:[group]:shred:0 src/disco/shred/fd_fec_resolver.c(624)[fd_fec_resolver_add_shred]: Spilled from fec_resolver in-progress map 407949634 0, data_shreds_rcvd 20, parity_shreds_rcvd 0
```

```
$ tail -f /tmp/validator.log | pv -l >/dev/null
 971k 0:00:26 [32.6k/s] 
```

Issue #9672, opened 2026-05-07
repair: restrict pings per pubkey identity further


emwang-jump:
may need (unclear) to stricter way instead of peer->ping to throttle pings, because as long as something is in the ping_queue no new repair request is sent. need to benchmark because I'm not sure with stem polling + peer->ping if this attack is even possible

Issue #9553, opened 2026-04-29
reasm: missing validation for fec set indexes and parent_off


emwang-jump:
no validation for first FEC index in slot == 0, or even that FECs follow each other by 32
causes sched crit

emwang-jump:
added 5/1 5:19 PM local 
Also does not verify that parent_off matches CMR

This is now considered a protocol-issue -- there's no way to determine what the proper parent block_id should be by only using parent_off, and CMR _should_ be the correct identifier. This is fixed by an incoming feature activation

emwang-jump:
5/8 3:18 PM local 
Should also validate all the fecs have the same parent_off

emwang-jump:
FEC chain for slot needs to terminate in a slot_complete / FEC 0 must follow a slot complete ....

ibhatt-jumptrading:
Per-Shred / Resolver Checks
Reject invalid parent_off.

For data shreds, require parent_off <= slot.
Require computed parent slot is valid: parent_slot < slot.
Require parent_slot >= root / slot_old as appropriate.
Reject shreds too far in the future.

Agave has max_slot; Firedancer only has slot_old.
Add an upper slot bound if Firedancer has equivalent context available.
Require SLOT_COMPLETE => DATA_COMPLETE.

Agave rejects LAST_SHRED_IN_SLOT without DATA_COMPLETE.
Firedancer currently checks slot-complete alignment, but not the flag implication.
Cross-FEC / Reasm Link Checks
First FEC in a slot must be fec_set_idx == 0.

If child->slot != parent->slot, then child->fec_set_idx == 0.
Same-slot FECs must be adjacent by exactly 32.

If child->slot == parent->slot, require:
child->fec_set_idx == parent->fec_set_idx + FD_FEC_SHRED_CNT
Same-slot FECs must share the same parent_off.

If child->slot == parent->slot, require:
child->parent_off == parent->parent_off
Cross-slot parent_off must match the actual linked parent.

If child->slot != parent->slot, require:
child->slot - child->parent_off == parent->slot
FEC 0 must follow a completed parent slot.

If child->slot != parent->slot, require:
parent->slot_complete
A same-slot child must not follow a slot-complete parent.

If child->slot == parent->slot, require:
!parent->slot_complete
Do not overwrite bad slot-boundary CMRs as valid.

Current overwrite_invalid_cmr() can convert “CMR doesn’t match parent” into “use parent from parent_off”.
[...]

PR #10001, opened 2026-05-29
fix(shred): filter out shreds with parents < root

lidatong:


emwang-jump (approved):

MattHintz:
For the public record — this PR touches the code path identified in Immunefi Report **#74093 — "FEC Resolver Eviction-Pinning DoS — Firedancer / Frankendancer"**, filed 2026-04-19 by `@Venator` (Matthew S. Hintz). The report was closed on 2026-04-30 and Immunefi mediation was concluded on 2026-05-29 as *"duplicate of #67117"*, with no substantive evidence produced for that duplicate determination.

This comment is filed so that reviewers of this PR have access to the disclosure history on the same code path. It is not an attempt to reopen mediation on the Immunefi channel.

---

## Mapping — this PR implements Option A of Report #74093's Recommended Fixes, verbatim

**What this PR does** (from the PR title): *"shred: filter out shreds with parents < root"* — adds a parent/root freshness rejection directly in `fd_fec_resolver_add_shred`.

**Verbatim from Report #74093, "Recommended Fixes" section, Option A:**

> **Option A — Freshness check on incoming shreds (simplest, recommended)**
> ```c
> // In fd_fec_resolver_add_shred, after slot_old check:
> if( shred->slot < current_replay_head - FRESHNESS_WINDOW ) {
>     return FD_FEC_RESOLVER_SHRED_REJECTED;
> }
> ```
> Reject first-shreds where the slot is more than `FRESHNESS_WINDOW` (e.g., 16 slots) behind the current replay head. This closes the replay-based amplification while still allowing legitimate repair of recent past.

Same file (`fd_fec_resolver.c`), same function (`fd_fec_resolver_add_shred`), same effect (reject shreds outside the freshness window).

---

## Root primitive this PR closes

**Verbatim from Report #74093, §1 "Root Cause":**

> The FEC resolver's eviction policy at `fd_fec_resolver.c:604-634` targets the highest `(slot, fec_set_idx)` context when the free list drops to `≤ partial_depth` (hardcoded to `1` at `fd_shred_tile.c:63`). This design assumes high-slot contexts are expendable and can be re-requested via repair.
>
> However, an attacker can fill the resolver with valid past-slot shreds that occupy contexts with lower slot values. When a legitimate current-slot shred arrives, it becomes the highest-slot entry and is immediately targeted for eviction by the next incoming shred. This inverts the intended eviction priority.

The parent/root filter shipped by this PR closes the admission side of that primitive: past-slot shreds with `parent < root` are rejected before they can allocate a resolver context, so the resolver can no longer be filled with the past-slot replay pool the report weaponizes.

---

## Closure rationale the report was given

**Verbatim from `int (Frankendancer)`, 2026-04-30, in the #74093 report thread:**

> "the behaviour you describe might be real for frankendancer, but if this ever happens, frankendancer would fall back on repair in that case and continue operating normally. So there is no actual DoS. Even if this were a valid bug (it isn't), it would be a duplicate of #67117 and #69417"

The "fall back on repair" premise was falsified at the resolver level by `poc_fec_eviction_repair_survives.c` in the PoC gist below (bit-identical outcome under `is_repair=1`), and again at the coverage level by `poc_fec_eviction_fleet_wide.c` (32 of 32 repair requests find no source when the flood is fleet-wide). The full falsification detail is in `MEDIATION_ADDENDUM_2026_05_08.md` and the sweep output `POC_DEPTH_ORDER_SWEEP_OUTPUT.txt` in the PoC gist.

That the freshness fix this PR now ships was specifically listed as Option A in the report is consistent with the report having identified a real primitive.

---

**Full evidence corpus (all public, all verifiable):**

- **Receipts + timeline gist** (this PR mapped to Report #74093 §1 + Recommended Fixes): https://gist.github.com/MattHintz/bfc2d1499acc2edf3e9a11007017cf10
- **PoC gist** (6 deterministic reproducers, built against upstream firedancer at `c1417286` and later): https://gist.github.com/MattHintz/3f884da0308f4e3610b9faa7cbfee7ba
- **Immunefi Report #74093**: https://bugs.immunefi.com/dashboard/submission/74093 (access-limited; the two gists above contain the full public record)

**Firedancer-authored issues on the same primitive** (all public):

- https://github.com/firedancer-io/firedancer/issues/7839 — `lidatong` 2026-01-13: *"track shreds by slot (currently: just uses FEC sig) to prevent spam"*
- https://github.com/firedancer-io/firedancer/issues/8103 — `emwang-jump` 2026-01-30, closed `not_planned` Priority: Low: *"fec_resolver is not going to let it through! ... if reasm has evicted it (unlikely, but possible), then we could stall"*
- https://github.com/firedancer-io/firedancer/issues/9553 — `ibhatt-jumptrading` 2026-05-12: *"Agave has `max_slot`; Firedancer only has `slot_old`. Add an upper slot bound if Firedancer has equivalent context available."*
- https://github.com/firedancer-io/firedancer/issues/9672 — `emwang-jump` 2026-05-07: *"I'm not sure with stem polling + `peer->ping` if this attack is even possible"*
- https://github.com/firedancer-io/firedancer/issues/9786 — `ripatel-fd` 2026-05-12: *`fd_fec_resolver.c(624)... Spilled from fec_resolver`* firing at ~32k lines/sec on FD-team validator, benign load
- https://github.com/firedancer-io/firedancer/issues/9799 — `ripatel-fd` 2026-05-13: *"repair tile is 100% busy"* — diagnosed eviction loop on production validator, benign load

PR #10002, opened 2026-05-29
gossip: no loopback

jherrera-jump:


src/app/firedancer/topology.c
```diff
@@ -1172,6 +1172,17 @@ fd_topo_configure_tile( fd_topo_tile_t * tile,
     tile->gossvf.entrypoints_cnt = config->gossip.entrypoints_cnt;
     fd_memcpy( tile->gossvf.entrypoints, config->gossip.resolved_entrypoints, tile->gossvf.entrypoints_cnt * sizeof(fd_ip4_port_t) );
 
+    if( FD_UNLIKELY( strcmp( config->firedancer.gossip.host, "" ) ) ) {
```
mmcgee-jump, line 1186: Don't love doing a DNS resolve here (although we already do for entrypoints IIRC?) as it slows boot of all tiles for gossip specific work.

src/discof/gossip/fd_gossvf_tile.c
```diff
@@ -605,6 +608,8 @@ verify_addresses( fd_gossvf_tile_ctx_t * ctx,
                   fd_gossip_message_t *  view,
                   uchar *                failed,
                   fd_stem_context_t *    stem ) {
+  int is_loopback_peer = ctx->peer.l==ctx->src_addr.l || ctx->peer.l==ctx->gossip_addr.l || (fd_ip4_addr_is_loopback( ctx->peer.addr ) && ctx->peer.port==ctx->src_addr.port);
```
mmcgee-jump: This `.l` pattern should be removed everywhere, it compares the padding too which can be arbitrary garbage. Also don't you just need an `addr` match for loopback?

src/app/firedancer/topology.c
jherrera-jump: changing resolve_addresses to first try to parse as an ip4 address, then fall back to dns lookup.

src/discof/gossip/fd_gossvf_tile.c
jherrera-jump: `ulong l : 48;` prevents this, but I'll switch to compare addr + port.

fd_ip4_addr_is_loopback prevents the case where we bind to our public address (e.g. 203.0.113.254) but a malicious peer advertises a loopback address in the payload (e.g. 127.0.0.1) which could trigger a self-ping

mmcgee-jump (approved):

PR #10008, opened 2026-05-30
Add irq-affinity configure stage

ripatel-fd:


ripatel-fd:
@greptile-jt review

src/app/firedancer/config/default.toml
```diff
@@ -901,6 +901,25 @@ telemetry = true
     # tiles.
     sign_tile_count = 2
 
+    # Firedancer can reconfigure hardware interrupt affinity, preventing
+    # IRQs from being handled by CPUs running fixed Firedancer tiles.
+    # This guards against performance problems under heavy network load.
+    [layout.interrupts]
```
mmcgee-jump: I don't really see the point of these config options. The whole point of the configure stages is they are optional, the operator just doesn't have to run them.

kbhargava-jump (approved):

PR #10009, opened 2026-05-30
keyguard: improve tests

ripatel-fd:
- **keyguard: separate keyguard and keyload test**
- **keyguard: make keyload pages readonly**


two-heart (approved):

ptaffet-jump:
I think there are changes needed to the keyload process so it doesn't segfault. Josh also has a PR for this. We should combine them, since you have some nice test additions. https://github.com/firedancer-io/firedancer/pull/9887/

ripatel-fd:
@ptaffet-jump @jherrera-jump good catch, will rebase when Joshua's patch is in.

jherrera-jump:
> @ptaffet-jump @jherrera-jump good catch, will rebase when Joshua's patch is in.

@ripatel-fd My patch landed

ripatel-fd:
@jherrera-jump rebased

jherrera-jump (approved):

PR #10012, opened 2026-06-01
diag: add IRQs metrics

ripatel-fd:
Counts IRQs and soft-IRQs reported by procfs.
Detects when IRQs steal fixed tile CPU.


ripatel-fd:
@greptile-jt re-review

ripatel-fd:
@greptile-jt review

PR #10013, opened 2026-06-01
events: mTLS authentication

### LINKED ISSUE #9968: Use mTLS for events
Use mutual TLS auth (ideally Raw Public Keys) instead of custom signing scheme to authenticate event clients for telemetry

ripatel-fd:
Replace custom challenge-response auth with TLS 1.3 client authentication

- **events: fix lifecycle bugs**
- **events: remove custom signing scheme**
- **events: make OpenSSL mandatory**
- **events: mutual TLS authentication**
- **events: add option to skip server cert verify**

Close #9968


src/app/firedancer/main.c
```diff
@@ -107,7 +107,9 @@ fd_topo_run_tile_t * TILES[] = {
   &fd_tile_shred,
   &fd_tile_sign,
   &fd_tile_metric,
+# if FD_HAS_OPENSSL
   &fd_tile_event,
+# endif
```
ripatel-fd, line 112: This leaves a nasty error message if someone compiles without openssl and tries to use event client, should be fixed

ripatel-fd:
@greptile-jt review

ripatel-fd:
@greptile-jt review

src/app/firedancer/topology.c
```diff
@@ -772,6 +773,9 @@ fd_topo_initialize( config_t * config ) {
     fd_topob_tile_in(  topo, "event",  0UL, "metric_in", "sign_event", 0UL, FD_TOPOB_UNRELIABLE, FD_TOPOB_UNPOLLED );
     fd_topob_tile_out( topo, "sign",   0UL,              "sign_event", 0UL                                         );
   }
+# else /* no OpenSSL */
+  if( telemetry_enabled ) FD_LOG_WARNING(( "ignoring [telemetry] = true: this build of Firedancer is missing OpenSSL" ));
```
mmcgee-jump, line 777: This should be an error, user has explicit intent for telemetry, tell them to turn it off.

ripatel-fd:
Cloudflare does not support EdDSA SPKI, so this is not worth working on

PR #10022, opened 2026-06-01
perf(eqvoc): leverage new fec resolver / shred tile API

lidatong:


src/choreo/eqvoc/fd_eqvoc.c
```diff
@@ -640,190 +640,118 @@ construct_proof( fd_shred_t const *          shred1,
...
 int
 fd_eqvoc_shred_insert( fd_eqvoc_t *                eqvoc,
-                       ushort                      shred_version,
-                       ulong                       root,
-                       fd_epoch_leaders_t const *  leader_schedule,
+                       int                         shred_hint,
                        fd_shred_t const *          shred,
                        fd_gossip_duplicate_shred_t chunks_out[static FD_EQVOC_CHUNK_CNT] ) {
 
-  if( FD_UNLIKELY( shred->slot <= root             ) ) return FD_EQVOC_ERR_SHRED_SLOT;
```
emwang-jump, line 776: not sure if we want to leave this in (probably doesn't matter), but shred and tower tile root  could be out of sync

emwang-jump (approved):

PR #10026, opened 2026-06-01
topo: fix >1 resolv tile

jherrera-jump:


mmcgee-jump (changes requested):
I'm not sure it's this easy due to bundles, which may have to arrive to pack in order? Round robining bundle txns between resolvs could re-order them into pack

jherrera-jump:
Will follow execle pattern and send all bundles to resolv:0

mmcgee-jump (approved):

PR #10047, opened 2026-06-03
docs: simplify tense in Frankendancer description

x4v13r64:
Tiny phrasing tweak: 'is using the Agave' → 'uses the Agave' (matches the surrounding present-tense voice).


x4v13r64:
Whoops, dropping this — opened by accident while exploring.

PR #10048, opened 2026-06-03
Unit test cleanup + partial makefile linting

intrigus-lgtm:
I was recently PRing some [fixes](https://github.com/firedancer-io/firedancer/pull/10027) for forest and realized that the tests were broken even without my fix.
Which made me realize that the test wasn't run in CI at all...

There have been quite a few times we missed tests in the past (just a sample):
https://github.com/firedancer-io/firedancer/pull/8319
https://github.com/firedancer-io/firedancer/issues/8194
https://github.com/firedancer-io/firedancer/pull/5640
https://github.com/firedancer-io/firedancer/pull/1139

This PR adds linting for makefiles:
1. make doesn't fail when github.com/firedancer-io/firedancer/commit/27a047b88378bdacc1fe3c56a665d4889b14ea2f `$(call add-hdr,fd_uint256.h fd_uint256_mul.h)` (should be add-hrd**s**) or when there are [typos](https://github.com/firedancer-io/firedancer/commit/b7c440b6294901259dfc4c4a37c98927a2820aae). I didn't find any other good way to prevent this without custom linting.
2. We often add make-unit-test calls, but don't add the corresponding run-unit-test calls. Linting now fails if there is a make-unit-test calls, but no corresponding run-unit-test calls.

Linting for 2. can be disabled using `# lint-no-run-unit-test` (e.g. if the test required CLI args).

In total, 21 tests were re-enabled. I disabled four of those again as they are broken and will be fixed in follow-up PRs.


src/util/wksp/Local.mk
```diff
@@ -10,15 +10,15 @@ $(call make-unit-test,test_wksp_used_treap,test_wksp_used_treap,fd_util)
...
+$(call make-unit-test,test_wksp_helper,test_wksp_helper,fd_util) # lint-no-run-unit-test, appears broken?
 $(call make-unit-test,test_wksp_tpool,test_wksp_tpool,fd_util)
 $(call make-unit-test,test_wksp,test_wksp,fd_util)
 
 $(call run-unit-test,test_wksp_used_treap)
 $(call run-unit-test,test_wksp_free_treap)
 $(call run-unit-test,test_wksp_admin)
 $(call run-unit-test,test_wksp_user)
-#$(call run-unit-test,test_wksp_helper) # FIXME: why was this not enabled?
+$(call run-unit-test,test_wksp_tpool)
```
ripatel-fd, line 21: This unit test wastes too much CPU on one tile, should decrease --iter-max or use multiple tiles (in script-tests)

PR #10052, opened 2026-06-03
alut: zero rest of buffer to match agave wire encoding 

intrigus-lgtm:


ripatel-fd (changes requested):
Could you split up the CI changes and the ALUT fix please?
For CI, a few of those tests are included in "script tests" instead, btw.

intrigus-lgtm:
Rebased.

ripatel-fd (approved):

PR #10060, opened 2026-06-03
config: auto-regenerate features and metrics

ripatel-fd:


mmcgee-jump (dismissed):

ripatel-fd:
Too annoying to deal with

PR #10066, opened 2026-06-03
refactor(reasm): remove overwrite invalid cmr

lidatong:


parasol-aser:
**`fd_reasm.c:896-909` — `fd_reasm_insert` orphan-linking loop never advances on a non-match.**
`subtreel_ele_peek_head` returns the head without removing it, but the non-match branch at `:898`
(`!fd_hash_eq1(orphan_root->cmr, fec->key)`) does `continue` without popping or advancing — so it
re-peeks the same head forever. Only the validate-fail and match paths pop the head. `subtreel`
holds every disconnected orphan-subtree root and a new FEC chains to at most a few of them, so any
non-matching root at the head hangs the loop. This is the common case (orphan FECs arriving before
their parent is normal), not just an attacker case: the existing `test_reasm.c`
("turbine orphan FECs added") hangs at this commit. The pre-PR `for` loop captured
`orphan_root->subtreel.next` before its `continue`, which is the progress this rewrite dropped.

Fix: advance past non-matching heads — restore the cursor capture before any `continue`/removal.

```suggestion
  for( ulong orphan_root_idx = subtreel_is_empty( subtreel, pool ) ? null : subtreel_idx_peek_head( subtreel, pool );
             orphan_root_idx != null; ) {
    fd_reasm_fec_t * orphan_root = pool_ele( pool, orphan_root_idx );
    orphan_root_idx = orphan_root->subtreel.next; /* advance before any continue/removal */
    if( FD_UNLIKELY( !fd_hash_eq1( orphan_root->cmr, fec->key ) ) ) continue; /* not parent */

    if( FD_UNLIKELY( !validate_chained_block_id( fec /* parent */, orphan_root /* child*/ ) ) ) {
      subtrees_remove( reasm, orphan_root, opt_store );
      continue;
    }
    link( reasm, fec, orphan_root );
    subtrees_ele_remove( subtrees, &orphan_root->key, NULL, pool );
    subtreel_ele_remove( subtreel, orphan_root, pool );
    orphaned_ele_insert( orphaned, orphan_root, pool );
    min_descendant = fd_ulong_min( min_descendant, orphan_root->slot );
  }
```

---
<sub>Automated review — may contain mistakes; please disregard if not useful.</sub>

src/discof/reasm/fd_reasm.c
```diff
@@ -899,6 +837,18 @@ fd_reasm_insert( fd_reasm_t *      reasm,
...
 
+  /* If the new FEC set is a child of a FEC that already exists,
+     validate the new FEC set against the parent.  If it fails
+     validation, remove the FEC from store. */
+
+  fd_reasm_fec_t * parent = fd_reasm_query( reasm, chained_merkle_root );
+  if( FD_UNLIKELY( parent && !validate_chained_block_id( parent, fec ) ) ) {
+    FD_LOG_NOTICE(( "failed validation of %lu %u", slot, fec_set_idx ));
+    pool_ele_release( pool, fec );
+    if( FD_LIKELY( opt_store ) ) fd_store_remove( opt_store, merkle_root );
```
emwang-jump: `replay_tile.c` calls `fd_store_remove` for anything that fails insertion so either we can update all the `return NULL` sites in reasm insert to do this or delete this

emwang-jump (approved):

PR #10071, opened 2026-06-03
contrib: Add GCC 12 container

esemeniuc:
## Summary

- add a Rocky 8 GCC Toolset 12 container image
- document an AVX512 Ice Lake build flow using the GCC 12 image


ripatel-fd (approved):
Thank you 🎉

PR #10078, opened 2026-06-04
refactor(tower): clean up voter querying out of banks / accdb

lidatong:


parasol-aser:
### Security review — PR #10078 (`fix(tower): confirmation and vtr querying`)

This PR is ~90% a behavior-preserving rename refactor of the tower tile's voter-querying functions; the one substantive logic change is in `replay_slot_completed` (hard-fork stake selection), where both findings below live.

**F1 — Unguarded NULL dereference of `lsched` (Medium, likely introduced)**
`src/discof/tower/fd_tower_tile.c:1232-1233`

The PR removes the prior `if( FD_LIKELY( hfork_lsched ) )` guard and now dereferences `lsched->epoch` directly. `fd_multi_epoch_leaders_get_lsched_for_slot` is genuinely nullable — it returns NULL when the slot's epoch isn't installed in `ctx->mleaders` yet. If a slot completes before its leader schedule lands (an epoch-update vs. replay ordering window), the consensus-critical tower tile segfaults instead of taking the old stake-0 path. Every other call site in this file guards this pointer, including the sibling slot-dead handler right below (`FD_TEST( lsched )` at line 1620) which does the identical `fd_hfork_record_our_bank_hash` computation. Not attacker-controlled.

**F2 — Stake silently non-zero for epochs ∉ {root, root+1} (Low, introduced; currently gated unreachable)**
`src/discof/tower/fd_tower_tile.c:1233`

The old 3-way selection defaulted out-of-range epochs to stake `0`; the new `fd_ulong_if( lsched->epoch==ctx->root_epoch, root_total, next_total )` hands any non-root epoch `next_epoch_total_stake`, perturbing a consensus-threshold input. Reachability is gated out by the file's existing epoch-window invariants (root advances ≤1 epoch; votes ≤1 epoch ahead), so this is defense-in-depth only — but the maintainers' own slot-dead handler asserts this exact invariant at line 1621.

**Suggested fix** — make this path identical to the slot-dead sibling:

```c
  fd_epoch_leaders_t const * lsched = fd_multi_epoch_leaders_get_lsched_for_slot( ctx->mleaders, slot_completed->slot );
  FD_TEST( lsched );
  FD_TEST( lsched->epoch==ctx->root_epoch || lsched->epoch==ctx->root_epoch + 1 );
```

Both fixes are trivial and match existing in-file idiom (line 1620-1621).

emwang-jump (dismissed):

mmcgee-jump (approved):

PR #10111, opened 2026-06-05
sched: rework eager pruning

yufeng-jump:
bug reported by @ibhatt-jumptrading

### DIFF
--- a/src/discof/replay/fd_replay_tile.c
+++ b/src/discof/replay/fd_replay_tile.c
@@ -338,7 +338,7 @@ publish_slot_completed( fd_replay_tile_t *  ctx,
   /* HACKY: hacky way of checking if we should send a null parent block
      id */
   fd_hash_t parent_block_id = {0};
-  if( FD_UNLIKELY( !is_initial ) ) {
+  if( FD_LIKELY( !is_initial ) ) {
     parent_block_id = ctx->block_id_arr[ bank->parent_idx ].latest_mr;
   }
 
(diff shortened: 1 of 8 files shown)


parasol-aser:
**`fd_sched.c` — dead-OOD branch of `fd_sched_fec_ingest` (new `subtree_release_refcnt` at ~L846)**

This branch now pushes to `ref_q`, but `fd_sched_fec_ingest` asserts `ref_q` is empty on entry (`FD_TEST( ref_q_empty(...) )`). `backfill_fec_sets` calls `insert_fec_set` → `fec_ingest` in a loop with no `ref_q` drain between iterations (the only drain runs once per `after_credit`, before FEC processing). So a dead-OOD ingest that pushes one entry, followed by any further ingest in the same backfill loop, trips the entry assert and aborts the replay tile — a remote, equivocating-leader-influenced liveness crash. Before this PR the dead-OOD branch only bumped counters and never touched `ref_q`, so the loop was safe. The one piece I couldn't confirm without a live node is whether backfill can stop at an ancestor that's `dying` in sched while its bank is still valid — but that's exactly the lifetime-discrepancy window this PR targets, so it looks reachable.

Fix: keep `fec_ingest` `ref_q`-neutral — either drain `ref_q` between iterations in `backfill_fec_sets`, or defer the dead-OOD release so it doesn't push from inside `fec_ingest`.

Worth a quick unit check either way: mark a block `dying`, then drive two first-in-block child FEC ingests back-to-back without an intervening `fd_sched_pruned_block_next` — the second aborts at this head but passes at the base SHA.

src/discof/replay/fd_replay_tile.c
```diff
@@ -1821,8 +1820,17 @@ try_prune_bank( fd_replay_tile_t * ctx ) {
       fd_accdb_cancel( ctx->accdb_admin, &xid );
       __attribute__((fallthrough));
     }
-    case 1: /* pruning bank + no cancellation is needed */
+    case 1: { /* pruning bank + no cancellation is needed */
+      /* A sched block exists, and can be marked dead, for a bank as
```
ibhatt-jumptrading, line 1824: nit: this comment is phrased kind of confusingly

ibhatt-jumptrading (approved):

PR #10117, opened 2026-06-08
restore: ignore append vec map

riptl:
Safe according to Agave

### DIFF
--- a/src/discof/restore/utils/fd_ssparse.c
+++ b/src/discof/restore/utils/fd_ssparse.c
@@ -19,186 +19,11 @@
 #define FD_SSPARSE_STATE_SCROLL_ACCOUNT_GARBAGE (8)
 #define FD_SSPARSE_STATE_ACCOUNT_BATCH          (9)
 
-struct fd_ssparse_private {
-  int state;
-  uint batch_enabled : 1;
-
-  struct {
-    int seen_zero_tar_frame;
-    int seen_manifest;
-    int seen_status_cache;
-    int seen_version;
-  } flags;
-
-  uchar version[ 5UL ];
-
-  struct {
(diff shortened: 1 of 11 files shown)


src/discof/restore/utils/fd_ssparse.c
```diff
@@ -19,93 +19,12 @@
...
-  void *         _acc_vec_map  = FD_SCRATCH_ALLOC_APPEND( l, acc_vec_map_align(),  acc_vec_map_footprint( max_acc_vecs )  );
-
-  ssparse->manifest.acc_vec_pool = acc_vec_pool_join( acc_vec_pool_new( _acc_vec_pool, max_acc_vecs ) );
-  FD_TEST( ssparse->manifest.acc_vec_pool );
-
-  ssparse->manifest.acc_vec_map = acc_vec_map_join( acc_vec_map_new( _acc_vec_map, max_acc_vecs, seed ) );
-  FD_TEST( ssparse->manifest.acc_vec_map );
-
+fd_ssparse_t *
+fd_ssparse_init( fd_ssparse_t * ssparse ) {
```
topointon-jump, line 23: we should also set `batch_enabled` here so to avoid an uninitialized read in `fuzz_snapshot_parser.c` (or call `fd_ssparse_batch_enable` in `fuzz_snapshot_parser.c`)

topointon-jump (dismissed):
Looks good! Just one minor comment

src/discof/restore/utils/fd_ssparse.c
ripatel-fd: Thanks, fixing. I misread the memset

mmcgee-jump (approved):

jherrera-jump (approved):

PR #10120, opened 2026-06-08
Remove usages of libc assert

riptl:
libc assert has a few problems
- whether it runs depends on compiler settings (e.g. 'zig cc'
  behaves differently than 'gcc' on a modern Ubuntu/RHEL box)
- if NDEBUG is defined, the expression inside assert is not
  evaluated, which breaks various Firedancer code

Replaced with FD_DCHECK / FD_LOG_CRIT / FD_TEST as appropriate

### DIFF
--- a/src/flamenco/vm/test_vm_instr.c
+++ b/src/flamenco/vm/test_vm_instr.c
@@ -5,7 +5,6 @@
 #include "fd_vm_base.h"
 #include "fd_vm_private.h"
 #include "../../ballet/sbpf/fd_sbpf_opcodes.h"
-#include <assert.h>
 #include <ctype.h>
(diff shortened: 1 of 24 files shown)


src/flamenco/vm/test_vm_instr.c
```diff
@@ -234,7 +233,7 @@ parse_token( test_parser_t *  p,
     return NULL;
   }
 
-  case ':': {  /* assertion */
+  case ':': {  /* FD_TESTion */
```
intrigus-lgtm, line 236: :)
```suggestion
  case ':': {  /* assertion */
```

topointon-jump (approved):

PR #10131, opened 2026-06-08
txsend: don't assert on leader missing

ibhatt-jumptrading:
have to do this since we only publish leader schedule when root advances so this can cause problems when catching up on a local cluster while epochs are still warming up. This could lead to the client replaying slots 2 epochs ahead of where the root is. 

thanks @anwayde for flagging


kbhargava-jump (dismissed):

intrigus-lgtm:
Good to see that this finally fixes https://github.com/firedancer-io/firedancer/issues/9281

ripatel-fd (dismissed):

anwayde (approved):

PR #10142, opened 2026-06-09
http: fix Sec-WebSocket-Key decode stack overflow

publicqi:
## Summary

`decoded_key` in the WebSocket upgrade path was sized 16 bytes, but
`fd_base64_decode` writes up to `FD_BASE64_DEC_SZ(24) == 18` bytes
*before* returning the length we check. A 24-char `Sec-WebSocket-Key`
without padding decodes to 18 bytes, overflowing the stack buffer by 2
attacker-controlled bytes on every WS upgrade request with such a key.

Fix: size the buffer to the decoder's max output. The `!= 16` check
(RFC 6455 s4.2.1 conformance) is unchanged.

## Test

`test_ws_bad_key_close` previously used a 20-char key, which the
`value_len != 24` check rejected before ever reaching the decode — so
the overflow path was untested. Updated it to a 24-char unpadded key
that actually reaches `fd_base64_decode`; it fails under ASAN without
the fix.


drubin-fd (approved):
nice find, thank you

PR #10145, opened 2026-06-09
Fix skipped slot eating our time window

HGuillemet:
When we become leader, and without `lagged_consecutive_leader_start`,  `slot_start_ns` is set to `reset_slot_start_ns` + `slot_duration_ns * (slot-reset_slot)`.
But then, when we skip a slot because it took too much time to complete, all the extra time we waited eats our time window.
Example:
image


ptaffet-jump (approved):
Looks good. We tested this internally and it works correctly.

ptaffet-jump:
CI is failing, but it's just because the acctdb PR changed the backtest flags. I'm going to merge.

PR #10170, opened 2026-06-10
fix(secp256r1): add one more point

publicqi:
`fd_secp256r1_base_point_table` was one point short, causing a 64-byte
out-of-bounds read on every secp256r1 verify.

the OOB does not affect correctness because it's not used. the OOB is reading
.rodata and it's unlikely to crash a validator

added points can be cross-verified at [pycryptodome:p256_table.c](https://github.com/Legrandin/pycryptodome/blob/d12264940c40417915fb85acb8c787ad9da58969/src/p256_table.c#L6656-L6659) and it's generated using the following script:

```python
import re

# ---- P-256 parameters ----
p  = 0xffffffff00000001000000000000000000000000ffffffffffffffffffffffff
a  = p - 3
b  = 0x5ac635d8aa3a93e7b3ebbd55769886bc651d06b0cc53b0f63bce3c3e27d2604b
Gx = 0x6b17d1f2e12c4247f8bce6e563a440f277037d812deb33a0f4a13945d898c296
Gy = 0x4fe342e2fe1a7f9b8ee7eb4a7c0f9e162bce33576b315ececbb6406837bf51f5
n  = 0xffffffff00000000ffffffffffffffffbce6faada7179e84f3b9cac2fc632551
R  = 1 << 256

def inv(x, m): return pow(x % m, m-2, m)

# affine point ops, identity = None
def padd(P, Q):
    if P is None: return Q
    if Q is None: return P
    x1,y1 = P; x2,y2 = Q
    if x1 == x2 and (y1 + y2) % p == 0: return None
    if P == Q:
        lam = (3*x1*x1 + a) * inv(2*y1, p) % p
    else:
        lam = (y2 - y1) * inv(x2 - x1, p) % p
    x3 = (lam*lam - x1 - x2) % p
    y3 = (lam*(x1 - x3) - y1) % p
    return (x3, y3)

def smul(k, P):
    R0 = None
    while k:
        if k & 1: R0 = padd(R0, P)
        P = padd(P, P)
        k >>= 1
    return R0

G = (Gx, Gy)

# sanity: G on curve
assert (Gy*Gy - (Gx*Gx*Gx + a*Gx + b)) % p == 0, "G not on curve"

def to_mont(x): return (x * R) % p
def limbs_le(v): return [(v >> (64*i)) & ((1<<64)-1) for i in range(4)]
def point_to_row(P):
    x,y = P
    xm, ym = to_mont(x), to_mont(y)
    return limbs_le(xm) + limbs_le(ym)

def fmt_row(row): return ", ".join("0x%016x" % w for w in row)

# ---- parse existing table from file ----
txt = open("src/ballet/secp256r1/fd_secp256r1_table.c").read()
hexes = [int(h,16) for h in re.findall(r'0x[0-9a-fA-F]+', txt)]
assert len(hexes) % 8 == 0, f"not multiple of 8: {len(hexes)}"
npts = len(hexes)//8
points = [hexes[8*i:8*i+8] for i in range(npts)]
print(f"file has {len(hexes)} ulongs = {npts} points")
print(f"blocksize=6 needs 43*32 = {43*32} points; missing {43*32 - npts}")
print()

# table index (B*i + (j-1)) = 2^(6i)*j*G ; B=32
def expected_point(idx):
    B = 32
    i, jm1 = divmod(idx, B)
    j = jm1 + 1
    k = (1 << (6*i)) * j
    return smul(k % n, G)

# ---- VERIFY method against existing rows ----
ok = True
for idx in [0, 1, 31, 32, 1374]:
    want = points[idx]
    got  = point_to_row(expected_point(idx))
    match = (want == got)
    ok &= match
    print(f"verify point[{idx}] (2^{6*(idx//32)}*{idx%32+1}*G): {'MATCH' if match else 'MISMATCH'}")
    if not match:
        print("  want:", fmt_row(want))
        print("  got :", fmt_row(got))
print()
print("ALL VERIFICATIONS PASSED" if ok else "VERIFICATION FAILED -- do not trust output")
print()

# ---- compute the MISSING point 1375 = i=42,j=32 = 2^252 * 32 * G = 2^257 * G ----
if ok:
    miss = expected_point(1375)
    assert miss == smul((1<<257) % n, G)
    row = point_to_row(miss)
    print("MISSING point[1375] = 2^252 * 32 * G  (append this row):")
    print("  " + fmt_row(row) + ",")
```


0x0ece:
Good catch! 🎉

src/ballet/secp256r1/fd_secp256r1_s2n.c
```diff
@@ -19,6 +19,12 @@
 
 #include "fd_secp256r1_table.c"
 
+/* p256_scalarmulbase() with blocksize=6 reads, in constant time, all 32
+   affine points (8 ulongs each) of every window i with 6*i<=256, i.e.
+   43 windows (i=0..42).  A short table is an out-of-bounds read, so pin
+   the size here. */
```
drubin-fd: No need to mention the short table in the comment, it is for the future, not the past :)

Something like,
```suggestion
/* p256_base_table_size is used for p256_scalarmulbase with a blocksize=6.
   B=2^(6-1)=32, there are 43 blocks i with 6*i<=256 (6*42=252<=256), and
   43 * 32 * 8 = 11008 entries.
   https://github.com/awslabs/s2n-bignum/blob/9061e8b76522beafa5ca020f3c8d99b23eba4fbc/x86/p256/p256_scalarmulbase.S#L14-L23 */
```

drubin-fd (changes requested):
This is a good catch.
Looking through how our usage of `p256_scalarmulbase` works, the `u1` input should always be reduced to the scalar field, so the most signficant bits will be zeroed, leading to the garbage OOB read never being added to the result as it does the table scan. So in practice, not a visible issue, but of course we never want to be doing OOB reads in the first place.

If you don't mind just clarifying that comment a little, I'm happy to merge it.

publicqi:
@drubin-fd just applied your comments

drubin-fd (approved):
Thank you! (idrc about the comments; will probably be cleaning this all up in the nearish future anyways)

PR #10174, opened 2026-06-10
fuzz(instr): copy mock_compile_message semantics for instruction harness

mjain-jump:


mjain-jump:
superceded by https://github.com/firedancer-io/firedancer/pull/10384

PR #10188, opened 2026-06-11
quic: fix wrong bound checks

chino-osec:
`fd_tpu_reasm_publish_fast` and `fd_tpu_reasm_frag` do not account for the `fd_txn_m_t` header size. This was not a triggerable bug, since callers are already doing more strict bound checks, but it is nonetheless a wrong check.


ripatel-fd:
Thank you for the contribution. I found a couple more issues and fixed them in #10206.

PR #10198, opened 2026-06-11
tower: refuse to vote if identity and vote account mismatch

ripatel-fd:
Increases protection for hot-spare operation, e.g. in the
following scenario:

  Main:    vote_acc=9Diao, identity=DNVZ,  auth_vtr=DNVZ
  Standby: vote_acc=9Diao, identity=Junk1, auth_vtr=DNVZ

On chain state:

  vote_acc=9Diao, node_pubkey=DNVZ, auth_vtr=DNVZ

Previously, this would cause both main and standy to generate
conflicting votes (standby signs twice, main signs once).

Credit to meyerbro / PigsInBlankets validator for reporting this
bug through Votalizer (https://votalizer.vercel.app/)


ripatel-fd:
@greptile-jt re-review

lidatong (approved):

PR #10207, opened 2026-06-12
util: implement wwb_{shl,shr} with one instruction

drubin-fd:


ptaffet-jump (approved):
Nice! One instruction, but a very CISC one...

PR #10215, opened 2026-06-12
events: add signed_vote

ripatel-fd:
Report signed vote transactions to event server


ripatel-fd:
@greptile-jt re-review

mmcgee-jump (approved):

PR #10222, opened 2026-06-12
events: simplify signing domain

ripatel-fd:
Use a static signing domain instead of a random key.  This makes it easier to prove that there are no signing confusion bugs.


mmcgee-jump (dismissed):

ripatel-fd:
@greptile-jt re-review

drubin-fd (approved):

PR #10224, opened 2026-06-13
bundle: build crank with p-memo

cavemanloverboy:
p-memo uses way less CUs than spl-memo

when building bundle crank, invoke new memo program and reduce budget by 2200 CUs (2500 to 300 CUs for memo. [Harmonic Major used 274 CUs](https://solscan.io/tx/3M3yuEyjktfAFaeCRmVioYUzreGLQrTvDmx8uP6S3DkQcohbwHjJ3B9E3w635FbPsMCT7GgHoyJXW9WVQJP7bxTq)).

@febo

this should not be merged until the memo program upgrade authority is removed, imo


ptaffet-jump:
I support this, but I agree we need to wait for p-memo to get a bit more widely supported. e.g. the upgrade authority as you mentioned, and it would be nice if the explorers decoded the instruction data as UTF-8 instead of hex.

Woody4618:
> I support this, but I agree we need to wait for p-memo to get a bit more widely supported. e.g. the upgrade authority as you mentioned, and it would be nice if the explorers decoded the instruction data as UTF-8 instead of hex.

This is fixed now in explorer: 
image

PR #10226, opened 2026-06-14
rpc: add WebSocket support (voteSubscribe)

ripatel-fd:
Add support for subscribing to votes via the RPC tile.
https://solana.com/docs/rpc/websocket/votesubscribe


ripatel-fd:
@greptile-jt re-review

jherrera-jump (approved):

PR #10254, opened 2026-06-15
events: add boot event with net device info

ripatel-fd:


ripatel-fd:
@greptile-jt re-review

PR #10257, opened 2026-06-16
tower: report confirmation event telemetry

mmcgee-jump:


src/disco/events/schema/slot_confirmed.json
```diff
@@ -0,0 +1,62 @@
...
+    "fields": {
+        "bank_seq": {
+            "type": "UInt64",
+            "description": "App-wide monotonic bank sequence number of the confirmed slot; the join key to block_completed.bank_seq. Obtained by querying the bank for this block_id at emission (fd_bank.bank_seq). Unique within a boot/run (resets to 0 on snapshot load), so pair with the stream boot_id/instance_id. May be unavailable for forward confirmations (forward=true), where we have not replayed the block; set to UINT64_MAX then."
```
ripatel-fd: IMO 0 as default is more clean for Protobuf, it's also easier to write queries for this. We can ensure that bank_seq starts as 1

src/disco/events/schema/slot_confirmed.json
```diff
@@ -0,0 +1,62 @@
...
+            "type": "UInt64",
+            "description": "The slot being confirmed. fd_tower_slot_confirmed.slot"
+        },
+        "block_id": {
+            "type": "Hash",
+            "description": "Block id being confirmed, guaranteed unique (join key to block_completed.block_id). 32-byte merkle root (fd_hash_t). fd_tower_slot_confirmed.block_id"
+        },
+        "stake": {
+            "type": "UInt64",
+            "description": "Stake (in lamports) supporting this confirmation at emission time. For the vote-stake levels (propagated/duplicate/optimistic/super) this is the vote stake observed for this exact (slot, block_id) -- divide by total_stake for the realized confirmation ratio that crossed the level threshold. For processed/rooted it is the ghost subtree stake (sum of stake voting for this block or any descendant). UINT64_MAX when unavailable (e.g. ignored, where the block has no ghost node)."
```
ripatel-fd: again, 0 as default preferable

src/discof/tower/fd_tower_tile.c
```diff
@@ -1344,6 +1382,7 @@ replay_slot_completed( fd_tower_tile_t *            ctx,
   /* Publish a slot_done frag to tower_out. */
 
   publish_slot_done( ctx, slot_completed, &out, found, our_vote_acct_bal, tsorig, stem );
+  report_slot_confirmed( slot_completed->bank_seq, slot_completed->slot, &slot_completed->block_id, ghost_blk->stake, ghost_blk->total_stake, ghost_blk->valid, FD_EVENT_SLOT_CONFIRMED_LEVEL_PROCESSED, 0 /* not forward */ );
```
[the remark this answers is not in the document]
mmcgee-jump: Good catch — fixed. `ghost_blk` can indeed be pruned by `fd_ghost_publish` during root advancement when the replayed block is on a minority fork, so `stake`/`total_stake`/`valid` are now captured into locals right after `QUERY_TOWERS` (before the root-advancement block) and the processed event reports from those. The rooted emission in the same function reads `intr->*` before the publish, so it's unaffected.

ripatel-fd (approved):

PR #10271, opened 2026-06-16
tower: fix impossible pool null check

two-heart:
this pool API never returns NULL, and all users of fd_tower_blocks_insert immediately dereference the return value. This defensive check for capacity should therefore use the pool_free API.


two-heart:
Fixes https://github.com/firedancer-io/firedancer/security/code-scanning/486

lidatong (approved):

PR #10273, opened 2026-06-16
txncache: stronger hashing

yufeng-jump:


src/flamenco/runtime/fd_txncache_shmem.c
```diff
@@ -171,6 +171,11 @@ fd_txncache_shmem_new( void * shmem,
   tc->txnpages_per_blockhash_max = _max_txnpages_per_blockhash;
   tc->max_txnpages               = _max_txnpages;
 
+  if( FD_UNLIKELY( !fd_rng_secure( &tc->seed, sizeof(tc->seed) ) ) ) {
```
mmcgee-jump: Not here, take seed as argument

src/flamenco/runtime/fd_txncache.c
```diff
@@ -155,6 +155,12 @@ fd_txncache_reset( fd_txncache_t * tc ) {
   fd_rwlock_unwrite( tc->shmem->lock );
 }
 
+FD_FN_PURE static inline ulong
+fd_txncache_bucket( fd_txncache_t const * tc,
+                    uchar const *         txnhash ) {
+  return fd_hash( tc->shmem->seed, txnhash, 20UL )%tc->shmem->txn_per_slot_max;
```
mmcgee-jump: Think this is too slow? Use faster hash here

src/flamenco/runtime/fd_txncache.c
```diff
@@ -155,6 +155,12 @@ fd_txncache_reset( fd_txncache_t * tc ) {
   fd_rwlock_unwrite( tc->shmem->lock );
 }
 
+FD_FN_PURE static inline ulong
+fd_txncache_bucket( fd_txncache_t const * tc,
+                    uchar const *         txnhash ) {
+  return fd_ulong_hash( FD_LOAD( ulong, txnhash )^tc->shmem->seed )%tc->shmem->txn_per_slot_max;
```
mmcgee-jump, line 161: Shouldn't you remove the fd_ulong_hash now?

mmcgee-jump (changes requested):

src/flamenco/runtime/fd_txncache.c
yufeng-jump: Wouldn’t that be the same bit strength as today?  Just need the low bits to collide.

mmcgee-jump (approved):

PR #10274, opened 2026-06-16
hfork: fix use-after-release

two-heart:
not a huge issue since this runs single threaded and release only clobbers next, still against the API and worth fixing.


lidatong:
hm i intentionally write it this way to keep all the structure release ops on the same line

it's a pretty common pattern to combine pool, map, dlist, and as a result:

list_pop...
map_remove...
pool_release...

the idea is to maximize readability of missing a potential release / avoiding a leak when skimming through code

it is technically UAF for the pool, but i don't think the template would ever change (at least without a careful comb through all usages). does that seem reasonable?

lidatong (approved):

PR #10281, opened 2026-06-17
disco: use clock API for tick-to-ns conversion

ripatel-fd:


src/disco/pack/fd_pack_tile.c
```diff
@@ -650,7 +640,7 @@ after_credit( fd_pack_ctx_t *     ctx,
 
   /* If we time out on our slot, then stop being leader.  This can only
      happen in the first after_credit after a housekeeping. */
```
ptaffet-jump: The second sentence is no longer true. That's not a problem, but just don't want the comment to be misleading.

ptaffet-jump:
Thanks! I only looked at pack, but other than my one comment, it looks good.

mmcgee-jump (dismissed):

mmcgee-jump (approved):

PR #10285, opened 2026-06-17
gui: add fd_gui_store

jherrera-jump:
The plan is to eventually migrate to the following variable-size element storages (backed by disk)

per-slot data
per-epoch data
per-min/hour/day data (maybe?)


jherrera-jump:
Going to iterate on the design

PR #10299, opened 2026-06-19
keys: bound CLI path copy

fallintoplace:
## Summary

- Reject `keys` file paths that do not fit in the fixed 256-byte action buffer.
- Replace fixed-size `fd_memcpy` from `argv` with `fd_cstr_ncpy` after a bounded length check.
- Add parser tests for short paths, 255-byte paths, and paths of 256 bytes or more for both `new` and `pubkey`.

## Root cause

`keys_cmd_args` copied 256 bytes from the CLI argument into `args->keys.file_path` regardless of the actual argument length. For shorter strings this could read past the argument allocation, and for 256-byte-or-longer strings the destination was not guaranteed to be NUL-terminated.

## Validation

- `git diff --check`
- `clang -std=c17 -fsyntax-only -I. -D_XOPEN_SOURCE=700 -DFD_HAS_HOSTED=1 -DFD_HAS_LINUX=1 -DFD_HAS_INT128=1 -DFD_HAS_DOUBLE=1 -DFD_HAS_ALLOCA=1 src/app/shared/test_keys_args.c`
- `clang -std=c17 -fsyntax-only -I. -D_XOPEN_SOURCE=700 -DFD_HAS_HOSTED=1 -DFD_HAS_LINUX=1 -DFD_HAS_INT128=1 -DFD_HAS_DOUBLE=1 -DFD_HAS_ALLOCA=1 -DGRND_RANDOM=2 -Wno-implicit-function-declaration src/app/shared/commands/keys.c`

### DIFF
--- a/src/app/shared/commands/keys.c
+++ b/src/app/shared/commands/keys.c
@@ -15,6 +15,16 @@ typedef enum {
   CMD_PUBKEY,
 } cmd_type_t;
 
+static int
+cstr_ncpy_check( char *       dst,
+                 char const * src,
+                 ulong        dst_sz ) {
+  if( FD_UNLIKELY( fd_cstr_nlen( src, dst_sz )>=dst_sz ) ) return 0;
+
+  fd_cstr_ncpy( dst, src, dst_sz );
+  return 1;
+}
+
 void
 keys_cmd_args( int *    pargc,
                char *** pargv,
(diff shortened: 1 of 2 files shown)


ripatel-fd (changes requested):
Thanks for the contribution.

Could you fix `fdctl_args::keys.file_path` to be `PATH_MAX` size? I'd call the helper `cstr_ncpy_check` or something, and let the error message simply say: `key file path too long`. 

Something like this

```c
if( !cstr_ncpy_check( args->keys.file_path, *pargv[ 0 ], sizeof(args->keys.file_path) ) ) {
  FD_LOG_ERR(( "key file path too long" ));
}
```

I don't think the test is necessary since it's simple to read high-level logic, could you remove it?

src/app/shared/commands/keys.c
```diff
@@ -26,14 +43,14 @@ keys_cmd_args( int *    pargc,
     (*pargv)++;
     if( FD_UNLIKELY( *pargc < 1 ) ) goto err;
     args->keys.cmd = CMD_NEW_KEY;
-    fd_memcpy( args->keys.file_path, *pargv[ 0 ], sizeof( args->keys.file_path ) );
+    keys_cmd_copy_file_path( args, *pargv[ 0 ] );
```
ripatel-fd: sorry one last nit, can you remove this helper function and just have the FD_LOG_ERR check duplicated (with braces) other than that it looks great


```suggestion
  if( FD_UNLIKELY( !cstr_ncpy_check( args->keys.file_path, file_path, sizeof( args->keys.file_path ) ) ) ) {
    FD_LOG_ERR(( "key file path too long" ));
  }
```

fallintoplace: Thank you for your attention, I will have a look as soon as possible. @ripatel-fd

ripatel-fd (approved):
Looks good, thanks!

PR #10301, opened 2026-06-19
vm: bound trace data printing

fallintoplace:
## Summary

- replace unbounded VM trace string appends with checked `vsnprintf`-backed helpers
- cap printed memory event data at the existing 2048-byte stack buffer contract
- add a large trace-data print regression case with `event_data_max` greater than 2048

## Root cause

`fd_vm_trace_printf()` sized its stack buffer for 2048 printable data bytes, but `data_sz` came from `trace->event_data_max`. A larger trace data cap let the memory dump loop write past the buffer through unbounded `fd_cstr_append_printf()` / `vsprintf` calls.

## Validation

- `git diff --check`
- `clang -std=c17 -fsyntax-only -I. -D_XOPEN_SOURCE=700 -DFD_HAS_HOSTED=1 -DFD_HAS_INT128=1 -DFD_HAS_DOUBLE=1 -DFD_HAS_ALLOCA=1 src/flamenco/vm/fd_vm_trace.c`
- `clang -std=c17 -fsyntax-only -I. -D_XOPEN_SOURCE=700 -DFD_HAS_HOSTED=1 -DFD_HAS_INT128=1 -DFD_HAS_DOUBLE=1 -DFD_HAS_ALLOCA=1 src/flamenco/vm/test_vm_base.c`

Could not run `gmake -j test_vm_base` locally on macOS: the native make path hits BSD `grep -P` and does not expose a `test_vm_base` target in this environment.

### DIFF
--- a/src/flamenco/vm/fd_vm_trace.c
+++ b/src/flamenco/vm/fd_vm_trace.c
@@ -2,6 +2,43 @@
 #include "../../ballet/sbpf/fd_sbpf_instr.h"
 #include "../../ballet/sbpf/fd_sbpf_opcodes.h"
 
+#include <stdarg.h>
+#include <stdio.h>
+
+#define FD_VM_TRACE_PRINTF_DATA_MAX (2048UL)
+#define FD_VM_TRACE_PRINTF_BUF_SZ   (1024UL + 6UL*FD_VM_TRACE_PRINTF_DATA_MAX)
+
+static int
+fd_vm_trace_printf_append( char **      p,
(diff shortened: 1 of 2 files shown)


ripatel-fd (changes requested):
Thanks for the fix. I'd rather rewrite fd_vm_trace to stream and not use printf at all. There's no reason to use printf formatting, and given the high output volume, should probably be optimized

fallintoplace:
Let me know if anything is needed from my side.

PR #10302, opened 2026-06-19
codeql: query to check decoding output buffer size

two-heart:


two-heart:
Would have caught a4795c54fd0794cb4ae6274b1bbf111246f243b1

ripatel-fd (approved):

PR #10321, opened 2026-06-24
progcache: minimize alloc

0x0ece:
Minimize alloc required to load a sbpf program.

Result: all sbpf load (in mainnet) require 1 single memcpy and no scratch.
- in mainnet there's currently 19.4k accounts, 9.83GB total
   - before: 98.5 memcpy, 23.27GB
   - after: 19.4k memcpy, 6.74GB
- on backtest:
   - before ~3.8k memcpy, 1.43GB
   - after: 757 calls, 0.4GB

Work is split in 2 commits:
- v3 (trivial) - just shows that v3 needs 1 single memcpy and no scratch. In v3 we have 2 adjacent mem areas: rodata (consts, strings) followed by text (bpf code).
- v0-v2 - same optimal result, 1 single memcpy and no scratch. In v0-v2 there's a large region [0..rodata_sz) and text is within that at [text_off..text_off+text_sz).

The idea for v0-v2 is pretty simple:
- Let's assume that the elf is "good enough", for some definition of "good enough" that's easy to test during peek.
- Then, we can just copy rodata_sz bytes and do relocation in place
- If not "good enough", keep doing the 3-5 memcpy with scratch

Next, I analyzed all 19.4k programs in mainnet and -surprise!- they're all "good enough". So fallback is never triggered.

Note: for now I tried to be extra careful and added all the required checks to peek (an alternative that I tested was to do only some quick checks, and asset the others during load... but it's easy to create adversarial inputs). Regardless, when disable_sbpf_v0_v1_v2_deployment is active, we can simply verify all 20k programs in mainnet again, and can remove most (all?) overhead from peek (the v0-v2 programs in mainnet will only be a known/fixed set, and they'll likely all pass the fast-lenient requirements).


src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -1926,13 +2120,21 @@ fd_sbpf_program_load_strict( fd_sbpf_program_t * prog,
...
+     Note: memcpy merged below */
   prog->text = (ulong *)( (uchar *)prog->rodata + prog->rodata_sz );
-  fd_memcpy( (uchar *)prog->text, (uchar const *)bin + bytecode_phdr.p_offset, bytecode_phdr.p_filesz );
+  // fd_memcpy( (uchar *)prog->text, (uchar const *)bin + bytecode_phdr.p_offset, bytecode_phdr.p_filesz );
+
+  /* Copy the rodata and bytecode (text) segments into the destination buffer.
+     rodata and text are contiguous, so we can copy them in a single memcpy. */
+  memcpy( prog->rodata,
+          (uchar const *)bin + phdr_0.p_offset,
+          prog->rodata_sz + (ulong)prog->info.text_sz );
```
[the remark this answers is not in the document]
0x0ece: text_sz >= 8.

src/ballet/sbpf/fd_sbpf_loader.h
```diff
@@ -170,6 +170,18 @@ struct fd_sbpf_elf_info {
   uint  text_cnt;       /* Instruction count */
   ulong text_sz;        /* size of text segment. Guaranteed to be <= bin_sz. */
 
+  /* Size of the buffer the loader assembles the program into, computed at
+     peek.  This is exactly the buffer the program cache must allocate:
+       - strict (v3+):       rodata + text segments (text_off + text_sz)
+       - lenient fast path:  the assembled rodata image (highest ro section
```
mjain-jump, line 176: its probably worth mentioning in the comment that the fast path triggers when the rodata has no overlapping sections

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -1711,6 +1785,44 @@ fd_sbpf_parse_ro_sections( fd_sbpf_program_t *             prog,
...
+    /* Sort the ro slice section indices by file offset (small N). */
+    for( ulong i=1UL; i<ro_slices_cnt; i++ ) {
+      ulong v = ro_slices_shidxs[ i ];
+      ulong j = i;
+      while( j>0UL && shdrs[ ro_slices_shidxs[ j-1UL ] ].sh_offset > shdrs[ v ].sh_offset ) {
+        ro_slices_shidxs[ j ] = ro_slices_shidxs[ j-1UL ];
+        j--;
+      }
+      ro_slices_shidxs[ j ] = v;
+    }
```
mjain-jump: cleaner to use a sorting template for this

src/ballet/sbpf/fd_sbpf_loader.c
```diff
@@ -1449,6 +1458,133 @@ fd_sbpf_lenient_elf_validate( fd_sbpf_elf_info_t * info,
...
+    ro_fill_length = fd_ulong_sat_add( ro_fill_length, section_data_len );
+    if( slices ) slices[ slice_cnt ] = i;
+    slice_cnt++;
+  }
+
+  /* The read-only sections must not overlap (incomplete: ignores gaps). */
+  if( FD_UNLIKELY( config->reject_broken_elfs &&
+                   fd_ulong_sat_add( lowest_addr, ro_fill_length )>highest_addr ) ) {
+    return FD_SBPF_ELF_ERR_VALUE_OUT_OF_BOUNDS;
+  }
```
mjain-jump: is this sufficient for checking if ANY two sections overlap? my approach would have been to aggregate the sections (vaddr_l, vaddr_h), sort in ascending order, and do a line sweep to check for any overlaps

0x0ece: this is existing code that matches agave's behavior, just moved around. Adding a comment but can't change it.

mjain-jump (approved):

PR #10336, opened 2026-06-26
firedancer: add remove-all-authorized-voters command

mindrunner:
Adds the inverse of add-authorized-voter, matching agave-validator's `authorized-voter remove-all` (removeAllAuthorizedVoters admin RPC). Removes every authorized voter from a running validator, including config-seeded and runtime-added ones, leaving only the identity key.

Mechanism mirrors add-authorized-voter but in reverse order: a new FD_KEYSWITCH_STATE_CLEAR_PENDING keyswitch signal clears the tower tile first (so it stops referencing any authorized voter index), then the sign tiles (which securely zero key material). Vote signing is synchronous and keyswitch handling runs between frags, so no in-flight vote-sign request can reference a key after the tower stops emitting them. Removal is unconditional and idempotent.

### DIFF
--- a/src/disco/keyguard/fd_keyswitch.h
+++ b/src/disco/keyguard/fd_keyswitch.h
@@ -17,6 +17,7 @@
 #define FD_KEYSWITCH_STATE_UNHALT_PENDING (3UL)
 #define FD_KEYSWITCH_STATE_FAILED         (4UL)
 #define FD_KEYSWITCH_STATE_COMPLETED      (5UL)
+#define FD_KEYSWITCH_STATE_CLEAR_PENDING  (6UL)
 
 struct __attribute__((aligned(FD_KEYSWITCH_ALIGN))) fd_keyswitch_private {
   ulong magic;     /* ==FD_KEYSWITCH_MAGIC */
--- a/book/api/firedancer-cli.md
+++ b/book/api/firedancer-cli.md
@@ -75,3 +75,50 @@ than 16).
 | `--config <path>` | Path to a configuration TOML file of the validator to add an authorized voter for. This must be the same configuration file the validator was started with |
 
 <<< @/snippets/commands/add-authorized-voter.ansi
+
+## `remove-all-authorized-voters`
+Removes all authorized voters from the running validator, including any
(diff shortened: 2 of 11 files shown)


mmcgee-jump:
Thanks! Could you add documentation to the book along with the command?

mindrunner:
> Thanks! Could you add documentation to the book along with the command?

done

mmcgee-jump:
Thanks again! Was holding this for the new admin tile, if you would like to try rebasing / porting it over :) Otherwise we can get this merged for you.

mindrunner:
> Thanks again! Was holding this for the new admin tile, if you would like to try rebasing / porting it over :) Otherwise we can get this merged for you.

done! `greptile` had some concerns, but i don't think they are valid. at least they would require changing the other commands as well to stay consistent. happy to hear your opinion.

src/disco/keyguard/fd_keyswitch.h
```diff
@@ -17,6 +17,7 @@
 #define FD_KEYSWITCH_STATE_UNHALT_PENDING (3UL)
 #define FD_KEYSWITCH_STATE_FAILED         (4UL)
 #define FD_KEYSWITCH_STATE_COMPLETED      (5UL)
+#define FD_KEYSWITCH_STATE_CLEAR_PENDING  (6UL)
```
ibhatt-jumptrading, line 20: we can probably just reuse SWITCH_PENDING since we are "switching" the key(s) to the empty set. for auth voters we reuse switch pending when we are adding to the set.

ibhatt-jumptrading (commented):
I think there's a race here:
1. tower tile sends out a vote with some authority idx n to txsend tile
2. keyswitches tower and sign (so the set of authority accounts is empty)
3. txsend sends to sign tile for a signature and will crash

probably want to flush txsend similar to what we do for set identity

ibhatt-jumptrading:
We made some more config tweaks and didn't want to make you keep having to rebase your PR so I cut another one with you #10458 as the author. Thank you for your contributions, they are very much appreciated!

PR #10344, opened 2026-06-29
keep track of cumulative transaction count

anwayde:


ibhatt-jumptrading (approved):

src/discof/replay/fd_replay_tile.c
```diff
@@ -332,7 +332,7 @@ publish_slot_completed( fd_replay_tile_t *  ctx,
   ulong epoch = fd_slot_to_epoch( epoch_schedule, slot, &slot_idx );
 
   ctx->metrics.slots_total++;
-  ctx->metrics.transactions_total = bank->f.txn_count;
```
mmcgee-jump, line 335: Think this breaks the GUI?

jherrera-jump (approved):

src/discof/replay/fd_replay_tile.c
jherrera-jump: GUI doesn't use this field, so should be fine.

PR #10346, opened 2026-06-30
reasm: reject self chaining FECs

emwang-jump:


lidatong (commented):
i don't think this can be externally triggered by an attacker via shreds because this would imply they found a merkle root collision (cryptography is broken, would be out of scope)

however, for the sake of defense-in-depth (compromised shred or repair tile), it would be good to document fd_reasm_insert "assumes CMR /= MR" and as good defense-in-depth practice have an assertion check in the caller

src/discof/reasm/test_reasm.c
```diff
@@ -1258,6 +1258,51 @@ test_eqvoc_xid_evict( fd_wksp_t * wksp ) {
...
+     the in-flight FEC itself -- matches the FEC against itself
+     (fd_hash_eq(parent->key, child->cmr) is true because key==cmr),
+     fails validate_chained_block_id (a FEC cannot be its own parent),
+     and calls subtrees_remove() on the in-flight FEC.  That happens
+     *before* xid_update() registered the FEC in the xid map, so
+     clear_xid_metadata()'s xid_query() returns NULL and the xid->cnt--
+     dereferences NULL.
+
+     Before the fix this test segfaults; after it the self-referential
+     FEC is rejected cleanly. */
```
ripatel-fd, line 1279: Could we remove this AI generated guilt admission

emwang-jump: 🤣

PR #10347, opened 2026-06-30
progcache: refresh CLOCK reference bit on cache hit

0x0ece:


src/flamenco/progcache/fd_progcache_user.c
```diff
@@ -213,6 +213,9 @@ fd_progcache_peek( fd_progcache_t *       cache,
   fd_progcache_load_fork( cache, fork_id );
   fd_progcache_rec_t * rec = fd_progcache_query( cache, prog_addr, feature_slot, deploy_slot );
   if( FD_UNLIKELY( !rec ) ) return NULL;
+  /* Mark the record as recently accessed */
+  fd_prog_clock_touch( cache->join->clock.bits,
+                       (ulong)( rec - cache->join->rec.pool->ele ) );
```
ripatel-fd: peek is pure / test function can you move this to pull

0x0ece: fixed

ripatel-fd (approved):

PR #10355, opened 2026-07-01
set identity: update gossip id outset

ibhatt-jumptrading:


src/flamenco/gossip/fd_gossip.c
```diff
@@ -391,15 +394,17 @@ fd_gossip_set_identity( fd_gossip_t * gossip,
   fd_active_set_set_identity( gossip->active_set, gossip->identity_pubkey, gossip->identity_stake );
   fd_prune_finder_set_identity( gossip->prune_finder, gossip->identity_pubkey, gossip->identity_stake );
 
-  refresh_contact_info( gossip, now );
+  /* For identity swaps, refresh the contact info outset so this
+     instance can override older contact info for the same identity. */
+  refresh_contact_info( gossip, now, identity_outset );
```
mmcgee-jump: Refresh here is kind of pure, you should just set identity_outset here in the caller, not add an arg

mmcgee-jump (changes requested):
.

mmcgee-jump (approved):

PR #10371, opened 2026-07-02
net: preferred busy polling

tristanx86-jump:
## Cover letter
Relates to https://github.com/firedancer-io/firedancer/issues/3618

Softirq poll mode is identical to what Firedancer currently uses and is set to be the default poll mode in this PR since it is more well tested. Prefbusy is the new poll mode this PR adds.

Prefbusy poll mode gives Firedancer significantly more control over how and when NAPI is polled, with no networking related IRQs or ksoftirqds during main runtime when used on a well supported driver.

Even when used on a less well supported driver inwhich ksoftirqds still occur like bnxt_en (Broadcom), the poll scheduling code used with prefbusy in this PR have been shown to be beneficial to RX throughput robustness as demonstrated in the test results below.
(bnxt_en Copy Mode)
image
(note TX still degraded towards 0Mpps like in softirq mode)

Prefbusy when well supported, removes Firedancer’s reliance on softirqs and ksoftirqds to poll NAPI. ksoftirqds are separate threads which softirqs execute inside of instead of the net tile's thread once network load is medium-high.

ksoftirqds are bad for a few reasons:
1. They (generally but not always) run on a different core to the net tile so each net tile in softirq poll mode effectively occupies two cores (when under medium-high network load) or if multiple ksoftirqd threads are running on the same core (polling for separate net tiles), only one of those net tiles can poll NAPI at a time.
2. If Linux schedules a ksoftirqd onto a core corresponding to a Firedancer tile, the performance of that tile will significantly degrade as network load increases, this situation is analyzed and tested in the mlx5 testing pdf with the net tile being the example choked Firedancer tile.
3. Linux decides how much the ksoftirqd thread runs vs other irq/house keeping threads, creating unwanted dependence on scheduling algorithms designed to prioritize fairness over performance.

## Test results
Update: IRQs/softirqs should no longer land on Firedancer tiles with the new IRQ affinity config stage implemented.

[ixgbe PrefBusy Tests.pdf](https://github.com/user-attachments/files/29358809/ixgbe.PrefBusy.Tests.pdf)
[mlx5 PrefBusy Tests.pdf](https://github.com/user-attachments/files/29358810/mlx5.PrefBusy.Tests.pdf)
[i40e PrefBusy Tests.pdf](https://github.com/user-attachments/files/29358811/i40e.PrefBusy.Tests.pdf)

### DIFF
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -978,36 +978,31 @@ dynamic_port_range = "8900-9000"
         xdp_zero_copy = false
 
         # The poll_mode option determines how Firedancer polls NAPI (Linux's
-        # high perf networking mechanisms). Its possible values are:
+        # high perf networking mechanisms).  Its possible values are:
         #
         #  "softirq" (default)
         #   Relies significantly more on Linux to manage NAPI through wakeups,
         #   softirqs, and under higher network load, a separate ksoftirqd
-        #   thread which Linux schedules onto a separate core. If a ksoftirqd
-        #   is scheduled by Linux onto a core corresponding to a different
-        #   tile, the performance of that tile is significantly degraded. softirq
-        #   poll mode is the default since it is more well tested.
+        #   thread which Linux schedules onto a separate core.
         #
         #  "prefbusy"
         #   Experimental option which gives Firedancer significantly more
-        #   control over how and when NAPI is polled, leading to much fewer
-        #   IRQs and ksoftirqds.
-        #
-        #   Its main benefit is significantly increased IRQ/ksoftirqd livelock
-        #   protection, which is important during DoS attacks and high net load.
+        #   control over how and when NAPI is polled.  Reduces networking
+        #   related IRQs and ksoftirqds to near zero on well supported NIC
+        #   drivers, increasing determinism and stability during DoS attacks
+        #   and high net load.
         #
         #   Hardware Support: Currently best supported on mlx5 (Mellanox).
-        #   Depending on the driver, XDP Copy Mode (SKB) can see massive
-        #   throughput gains (up to 2x on mlx5, 1.5x on i40e). Zero-copy (AF_XDP)
-        #   throughput generally remains the same across both poll modes.
+        #   Depending on the driver, XDP Copy Mode can see noticeable throughput
+        #   gains (up to 2.4x on mlx5, 1.5x on i40e).  Zero-copy RX throughput
+        #   generally remains the same across both poll modes.
         #
         # Extra info
         #
-        #  Older kernels: If your kernel is too old (< v5.11), Firedancer will
-        #  automatically fall back to "softirq" mode and emit a warning.
+        #  Requires a kernel version of at least v5.11 (Feb 2021).
         #
-        #  SSH minor latency: You may notice a small-medium increase in SSH
-        #  latency if using the same interface(s) as Firedancer for SSH.
+        #  Prefbusy mode increases tail latency for other applications on
+        #  the system sharing the same network interface.
         poll_mode = "softirq"
 
         # XDP uses metadata queues shared across the kernel and
@@ -1015,7 +1010,7 @@ dynamic_port_range = "8900-9000"
         # This setting defines the number of entries in these metadata
         # queues, which has to be a power of two.
         #
-        # Smaller values use less memory. Larger values reduce packet
+        # Smaller values use less memory.  Larger values reduce packet
         # loss caused by excessive scheduling or inter-CPU latency.
         xdp_rx_queue_size = 32768
         xdp_tx_queue_size = 32768
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -977,12 +977,45 @@ dynamic_port_range = "8900-9000"
         # "operation not supported".
         xdp_zero_copy = false
 
+        # The poll_mode option determines how Firedancer polls NAPI (Linux's
+        # high perf networking mechanisms). Its possible values are:
+        #
+        #  "softirq" (default)
+        #   Relies significantly more on Linux to manage NAPI through wakeups,
+        #   softirqs, and under higher network load, a separate ksoftirqd
+        #   thread which Linux schedules onto a separate core. If a ksoftirqd
+        #   is scheduled by Linux onto a core corresponding to a different
+        #   tile, the performance of that tile is significantly degraded. softirq
+        #   poll mode is the default since it is more well tested.
+        #
+        #  "prefbusy"
+        #   Experimental option which gives Firedancer significantly more
+        #   control over how and when NAPI is polled, leading to much fewer
+        #   IRQs and ksoftirqds.
+        #
+        #   Its main benefit is significantly increased IRQ/ksoftirqd livelock
+        #   protection, which is important during DoS attacks and high net load.
+        #
+        #   Hardware Support: Currently best supported on mlx5 (Mellanox).
+        #   Depending on the driver, XDP Copy Mode (SKB) can see massive
+        #   throughput gains (up to 2x on mlx5, 1.5x on i40e). Zero-copy (AF_XDP)
+        #   throughput generally remains the same across both poll modes.
+        #
+        # Extra info
+        #
+        #  Older kernels: If your kernel is too old (< v5.11), Firedancer will
+        #  automatically fall back to "softirq" mode and emit a warning.
+        #
+        #  SSH minor latency: You may notice a small-medium increase in SSH
+        #  latency if using the same interface(s) as Firedancer for SSH.
+        poll_mode = "softirq"
+
         # XDP uses metadata queues shared across the kernel and
         # userspace to relay events about incoming and outgoing packets.
         # This setting defines the number of entries in these metadata
         # queues, which has to be a power of two.
         #
-        # Smaller values use less memory.  Larger values reduce packet
+        # Smaller values use less memory. Larger values reduce packet
         # loss caused by excessive scheduling or inter-CPU latency.
         xdp_rx_queue_size = 32768
         xdp_tx_queue_size = 32768
--- a/src/app/shared/commands/configure/sysfs-poll.c
+++ b/src/app/shared/commands/configure/sysfs-poll.c
@@ -1,0 +1,94 @@
+/* This stage configures the OS to support effective preferred busy
+   polling, allowing for significantly improved network stack (XDP)
+   reliability under high load if enabled on a well supported driver. */
+
+#include "configure.h"
+
+#define NAME "sysfs-poll"
+
+#include "../../../platform/fd_file_util.h"
+
+#include <string.h> /* strcmp */
+#include <unistd.h> /* access */
+#include <linux/capability.h>
+
+/* Values for NAPI_DEFER_HARD_IRQS and GRO_FLUSH_TIMEOUT
+   chosen based on napibusy configuration tested and shown
+   in https://lwn.net/Articles/997491/ Linux patch cover letter.
+   Chosen over fullbusy since for Firedancer the values of
+   fullbusy are unnecessarily high and cause significant SSH
+   latency. */
+#define NAPI_DEFER_HARD_IRQS 100U
+#define GRO_FLUSH_TIMEOUT    200000U
+
+static char const setting_napi_defer_hard_irqs[] = "napi_defer_hard_irqs";
+static char const setting_gro_flush_timeout[]    = "gro_flush_timeout";
+
+static int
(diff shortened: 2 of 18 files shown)


src/app/fdctl/config/default.toml
```diff
@@ -977,12 +977,45 @@ dynamic_port_range = "8900-9000"
...
+        #
+        #   Hardware Support: Currently best supported on mlx5 (Mellanox).
+        #   Depending on the driver, XDP Copy Mode (SKB) can see massive
+        #   throughput gains (up to 2x on mlx5, 1.5x on i40e). Zero-copy (AF_XDP)
+        #   throughput generally remains the same across both poll modes.
+        #
+        # Extra info
+        #
+        #  Older kernels: If your kernel is too old (< v5.11), Firedancer will
+        #  automatically fall back to "softirq" mode and emit a warning.
```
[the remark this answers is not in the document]
ripatel-fd: This is probably fine, I don't think we need to support very old kernels

src/app/fdctl/config/default.toml
```diff
@@ -977,12 +977,45 @@ dynamic_port_range = "8900-9000"
...
 
+        # The poll_mode option determines how Firedancer polls NAPI (Linux's
+        # high perf networking mechanisms). Its possible values are:
+        #
+        #  "softirq" (default)
+        #   Relies significantly more on Linux to manage NAPI through wakeups,
+        #   softirqs, and under higher network load, a separate ksoftirqd
+        #   thread which Linux schedules onto a separate core. If a ksoftirqd
+        #   is scheduled by Linux onto a core corresponding to a different
+        #   tile, the performance of that tile is significantly degraded. softirq
```
ripatel-fd: I don't think we need to mention the tile overlap hazard in the config file here because the `irq-affinity` stage is supposed to take care of this

src/app/fdctl/config/default.toml
```diff
@@ -977,12 +977,45 @@ dynamic_port_range = "8900-9000"
...
+        #   throughput gains (up to 2x on mlx5, 1.5x on i40e). Zero-copy (AF_XDP)
+        #   throughput generally remains the same across both poll modes.
+        #
+        # Extra info
+        #
+        #  Older kernels: If your kernel is too old (< v5.11), Firedancer will
+        #  automatically fall back to "softirq" mode and emit a warning.
+        #
+        #  SSH minor latency: You may notice a small-medium increase in SSH
+        #  latency if using the same interface(s) as Firedancer for SSH.
```
ripatel-fd: I would rephrase this as: prefbusy mode increases tail latency for other applications on the system

src/app/fdctl/config/default.toml
```diff
@@ -977,12 +977,45 @@ dynamic_port_range = "8900-9000"
...
+        poll_mode = "softirq"
+
         # XDP uses metadata queues shared across the kernel and
         # userspace to relay events about incoming and outgoing packets.
         # This setting defines the number of entries in these metadata
         # queues, which has to be a power of two.
         #
-        # Smaller values use less memory.  Larger values reduce packet
+        # Smaller values use less memory. Larger values reduce packet
```
ripatel-fd: We put two spaces between sentences to make things more readable, could you do that for your XDP docs too?

src/app/shared/commands/configure/sysfs-poll.c
```diff
@@ -0,0 +1,94 @@
...
+#include <string.h> /* strcmp */
+#include <unistd.h> /* access */
+#include <linux/capability.h>
+
+/* Values for NAPI_DEFER_HARD_IRQS and GRO_FLUSH_TIMEOUT
+   chosen based on napibusy configuration tested and shown
+   in https://lwn.net/Articles/997491/ Linux patch cover letter.
+   Chosen over fullbusy since for Firedancer the values of
+   fullbusy are unnecessarily high and cause significant SSH
+   latency. */
```
ripatel-fd: SSH latency can be refusing to readers - I would avoid mentioning it and then just say "regular traffic" or "other sockets"

src/app/shared/commands/configure/sysfs-poll.c
```diff
@@ -0,0 +1,94 @@
...
+
+static void
+sysfs_net_set( char const * device,
+               char const * setting,
+               ulong        value ) {
+  char path[ PATH_MAX ];
+  fd_cstr_printf_check( path, PATH_MAX, NULL, "/sys/class/net/%s/%s", device, setting );
+  FD_LOG_NOTICE(( "RUN: `echo \"%lu\" > %s`", value, path ));
+  if( FD_UNLIKELY( -1==fd_file_util_write_uint( path, (uint)value ) ) ) {
+    FD_LOG_ERR(( "could not write to `%s`, 'prefbusy' poll mode may not be supported for this network configuration, consider switching back 'poll_mode' within your config to 'softirq' mode.", path));
```
ripatel-fd:
```suggestion
    FD_LOG_ERR(( "could not write to `%s` (%i-%s), 'prefbusy' poll mode may not be supported for this network configuration, consider switching back 'poll_mode' within your config to 'softirq' mode.", path, errno, fd_io_strerror( errno ) ));
```

src/app/shared/commands/configure/sysfs-poll.c
```diff
@@ -0,0 +1,94 @@
...
+}
+
+static configure_result_t
+check( config_t const * config,
+       int              check_type FD_PARAM_UNUSED ) {
+  char path[ PATH_MAX ];
+  uint value;
+  fd_cstr_printf_check( path, PATH_MAX, NULL, "/sys/class/net/%s/%s", config->net.interface, setting_napi_defer_hard_irqs );
+  if( fd_file_util_read_uint( path, &value )
+      || value != ( NAPI_DEFER_HARD_IRQS ) ) {
```
ripatel-fd:
```suggestion
      || value != NAPI_DEFER_HARD_IRQS ) {
```

src/app/shared/commands/configure/sysfs-poll.c
```diff
@@ -0,0 +1,94 @@
...
+    NOT_CONFIGURED("Setting napi_defer_hard_irqs failed.");
+  }
+
+  fd_cstr_printf_check( path, PATH_MAX, NULL, "/sys/class/net/%s/%s", config->net.interface, setting_gro_flush_timeout );
+  if( fd_file_util_read_uint( path, &value )
+      || value != ( GRO_FLUSH_TIMEOUT ) ) {
```
ripatel-fd:
```suggestion
      || value != GRO_FLUSH_TIMEOUT ) {
```
brackets usually go in the macro definition

src/app/shared/commands/configure/sysfs-poll.c
```diff
@@ -0,0 +1,94 @@
...
+    NOT_CONFIGURED("Setting gro_flush_timeout failed.");
```
ripatel-fd:
```suggestion
    NOT_CONFIGURED( "Setting gro_flush_timeout failed." );
```

src/app/shared/commands/configure/sysfs-poll.c
```diff
@@ -0,0 +1,94 @@
...
+
+static configure_result_t
+check( config_t const * config,
+       int              check_type FD_PARAM_UNUSED ) {
+  char path[ PATH_MAX ];
+  uint value;
+  fd_cstr_printf_check( path, PATH_MAX, NULL, "/sys/class/net/%s/%s", config->net.interface, setting_napi_defer_hard_irqs );
+  if( fd_file_util_read_uint( path, &value )
+      || value != ( NAPI_DEFER_HARD_IRQS ) ) {
+    NOT_CONFIGURED("Setting napi_defer_hard_irqs failed.");
```
ripatel-fd:
```suggestion
    NOT_CONFIGURED( "Setting napi_defer_hard_irqs failed." );
```

src/disco/net/xdp/fd_xdp_tile.c
```diff
@@ -51,6 +51,45 @@
...
+   packets that the NIC driver can move from the hw rings into the
+   XSK rings per poll).
+
+   Default RX budget used by NIC drivers is 64, therefore
+   it is safest to use 64 in prefbusy polling. Also reduces
+   TX starvation risks as the TX budget set by the NIC driver
+   is also generally 64. */
+#define PREFBUSY_RX_BUDGET (64L)
+
+/* Min time between each prefbusy poll (further protects against livelocks).
```
ripatel-fd: I see two mentions of "further protects against livelocks" but these are the only mention of livelock as far as I can tell. Can you add a comment explaining the livelock hazard?

src/disco/net/xdp/fd_xdp_tile.c
```diff
@@ -1122,6 +1168,99 @@ net_rx_event( fd_net_ctx_t * ctx,
...
+
+static void
+net_prefbusy_poll_flush( fd_net_flusher_t * flusher,
+                         long               now ) {
+  flusher->prefbusy_last_poll_ticks = now;
+}
+
+/* before_credit_prefbusy is called every loop iteration if net
+   tile is in preferred busy (often referred to as "prefbusy" in
+   Firedancer) polling mode. */
```
ripatel-fd: This comment is a bit redundant

src/disco/net/xdp/fd_xdp_tile.c
```diff
@@ -1122,6 +1168,99 @@ net_rx_event( fd_net_ctx_t * ctx,
   fill_ring->cached_prod = fill_prod+1U;
 }
 
+/* before_credit_softirq is called every loop iteration if net tile
+   is in softirq polling mode (fallback if prefbusy polling mode
+   is not enabled). */
```
ripatel-fd: This comment is a bit redundant

src/waltz/xdp/fd_xsk.c
```diff
@@ -187,6 +203,57 @@ fd_xsk_setup_umem( fd_xsk_t *              xsk,
   return 0;
 }
 
+/* fd_xsk_setup_poll: Setup preferred busy polling if the user has
+   set that to be their preferred polling method */
```
ripatel-fd:
```suggestion
/* fd_xsk_setup_poll: Setup preferred busy polling */
```

src/waltz/xdp/fd_xsk.c
```diff
@@ -187,6 +203,57 @@ fd_xsk_setup_umem( fd_xsk_t *              xsk,
...
+
+static void
+fd_xsk_setup_poll( fd_xsk_t *              xsk,
+                   fd_xsk_params_t const * params ) {
```
ripatel-fd, line 210: This function logs quite technical warnings that are unintelligible to operators.
These "function call failed" logs are very useful for debugging so they should stay.
But could you also add a human readable description of the error?

e.g. `failed to enable XDP busy polling: setsockopt(...`

src/waltz/xdp/fd_xsk.c
```diff
@@ -187,6 +203,57 @@ fd_xsk_setup_umem( fd_xsk_t *              xsk,
...
+  if( 0!=strcmp( params->poll_mode, "prefbusy" ) ) return;
+
+  /* Configure socket options for preferred busy polling */
+
+  int prefbusy_poll = 1;
+  if( FD_UNLIKELY( 0!=setsockopt( xsk->xsk_fd, SOL_SOCKET, SO_PREFER_BUSY_POLL, &prefbusy_poll, sizeof(int) ) ) ) {
+    int err = errno;
+    FD_LOG_WARNING(( "setsockopt(xsk_fd,SOL_SOCKET,SO_PREFER_BUSY_POLL,1) failed (%i-%s)", err, fd_io_strerror( err ) ));
+    if( err==EINVAL ) {
+      FD_LOG_WARNING(( "Hint: Does your kernel support preferred busy polling? SO_PREFER_BUSY_POLL is available from Linux 5.11 onwards" ));
```
ripatel-fd:
```suggestion
      FD_LOG_NOTICE(( "Hint: Does your kernel support preferred busy polling? SO_PREFER_BUSY_POLL is available from Linux 5.11 onwards" ));
```

src/disco/net/xdp/fd_xdp_tile.c
tristanx86-jump: Since i have made it poll whenever the RX queue is empty, without a min time between each poll TX would be livelocked if there are no packets being received.

But I believe this min time between each poll is also why bnxt_en worked better with prefbusy than softirq which is why I added 'further'.

tristanx86-jump: I will just mention the RX queue empty explanation though in the code for readability.

ripatel-fd (approved):

PR #10375, opened 2026-07-03
frank, gui: fix accounts scan bottleneck

jherrera-jump:
agave [diff](https://github.com/firedancer-io/agave/compare/4e72d35c2a8a122eca272614f391e084371fed97..f7db9dee441713162a70086cff23e84788f05386)

Fixes an issue where the ConfigProgram scan at startup prevents the database root from advancing during the scan.


ptaffet-jump (approved):
Don't fully understand the invariants, but I think it looks good.

PR #10376, opened 2026-07-03
xdp: minor prefbusy configuration changes

tristanx86-jump:


src/app/shared/commands/configure/sysfs-poll.c
```diff
@@ -70,15 +70,27 @@ check( config_t const * config,
...
-    NOT_CONFIGURED( "Setting napi_defer_hard_irqs failed." );
+
+  /* Check NAPI_DEFER_HARD_IRQS value set correctly. */
+
+  if( FD_UNLIKELY( -1==fd_file_util_read_uint( path, &value ) ) ) {
+    FD_LOG_ERR(( "could not read `%s` (%i-%s)", path, errno, fd_io_strerror( errno ) ));
+  }
+
+  if( FD_UNLIKELY( NAPI_DEFER_HARD_IRQS != value ) ) {
+    NOT_CONFIGURED( "kernel parameter `%s` is set to %u, not the expected value of %u. Proceeding but prefbusy polling may not work correctly.", path, value, NAPI_DEFER_HARD_IRQS );
```
mmcgee-jump: The comment here is not right, if you return `NOT_CONFIGURED` it's going to run `init` and configure it

tristanx86-jump: Believed check to only run post fini/init, my mistake.

ripatel-fd (approved):

PR #10379, opened 2026-07-03
configure: fix ethtool-channels with offline CPUs

ripatel-fd:
Fixes a bug where ethtool-channels would fail to initialize on
systems with offline CPUs

### DIFF
--- a/src/app/shared/commands/configure/fd_ethtool_ioctl.c
+++ b/src/app/shared/commands/configure/fd_ethtool_ioctl.c
@@ -1,5 +1,6 @@
 #include <errno.h>
 #include <unistd.h>
+#include <sys/sysinfo.h>
 #include <linux/ethtool.h>
 #include <linux/sockios.h>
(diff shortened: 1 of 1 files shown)


src/app/shared/commands/configure/fd_ethtool_ioctl.c
```diff
@@ -17,6 +18,16 @@
 
 #define ETHTOOL_CMD_SIZE( base_t, data_t, data_len ) ( sizeof(base_t) + (sizeof(data_t)*(data_len)) )
 
+/* cpu_present_cnt returns the number of installed CPUs in the system,
+   including offline CPUs */
+
+static uint
+cpu_present_cnt( void ) {
```
mmcgee-jump, line 25: I'd use `fd_topo_cpus_init` here it's a bit cleaner

mmcgee-jump (changes requested):
.

PR #10382, opened 2026-07-06
xdp: allow up to 4 GRE tunnels

ptaffet-jump:


src/disco/net/xdp/test_xdp_tile.c
tristanx86-jump, line 478:
```suggestion
  FD_TEST( net_gre_tunnel_ip( ctx )==0UL );
```

Return type changed from U to UL. I believe the purpose of this test still works fine in its current form but should use the correct return type.

tristanx86-jump (approved):
Looks good! Just the small implicit type conversion I think and that's it :)

ripatel-fd (approved):

PR #10384, opened 2026-07-06
runtime, gossip, shred: changes for agave v4.2

mjain-jump:


src/flamenco/runtime/fd_borrowed_account.c
```diff
@@ -210,15 +210,6 @@ int
...
                                          ulong                         new_length,
                                          int *                         err ) {
-
-  /* Only the owner can change the length of the data
-     https://github.com/anza-xyz/agave/blob/v2.1.14/sdk/src/transaction_context.rs#L1095 */
-  if( FD_UNLIKELY( (borrowed_acct->acc->data_len!=new_length) &
-                   (!fd_borrowed_account_is_owned_by_current_program( borrowed_acct )) ) ) {
-    *err = FD_EXECUTOR_INSTR_ERR_ACC_DATA_SIZE_CHANGED;
-    return 0;
-  }
```
topointon-jump, line 220: Let me check this

src/flamenco/runtime/fd_borrowed_account.c
```diff
@@ -210,15 +210,6 @@ int
...
-
   /* The new length can not exceed the maximum permitted length
```
topointon-jump, line 203: I think this check is still needed for 1-1 parity with Agave's borrowed accounts API - Agave's `BorrowedInstructionAccount::can_data_be_resized` calls `self.can_data_be_changed()` which has the `!self.is_owned_by_current_program()` check: https://github.com/anza-xyz/agave/blob/master/transaction-context/src/instruction_accounts.rs#L350-L357

topointon-jump: It looks like our `fd_borrowed_account_can_data_be_resized` is actually incorrect wrt Agave - so the check you removed was redundant in Agave, but not in ours.

mjain-jump: yikes, good find. fixed

topointon-jump (approved):

PR #10398, opened 2026-07-07
net: pktgen multiple net tiles.

tristanx86-jump:
Pktgen currently only works with a single net tile. This complicates testing on servers using bonds (since native bonding needs at least one net tile per underlying slave interface to work).

This PR updates the topo setup within pktgen to allow for multiple net tiles, and makes per net tile stats visible to the user.


tristanx86-jump:
Moved to https://github.com/firedancer-io/firedancer/pull/10577

PR #10411, opened 2026-07-08
shred: prevent rare repair re-requests

two-heart:


ptaffet-jump:
Thanks for catching this. I think you should be able to get away with just one `& 0x7FFFFFFFUL` operation after the subtraction,  The version I had with two is incorrect, like you noticed.  Your version with three is okay, but unnecessary.

two-heart:
You are right. I changed it to just use a single `& 0x7FFFFFFFUL`

ptaffet-jump (approved):

PR #10419, opened 2026-07-09
accdb: fix concurrent read/write edge case due to implied programdata

mmcgee-jump:


yufeng-jump (approved):
Stamping to unblock.  Will review.

PR #10430, opened 2026-07-10
Fix offline cpu handling

HGuillemet:
This is a replacement for PR #8065 that was incomplete and created an additional unneeded dependency to disco library.

Problems fixed:
* Automatic tile layout can pin tiles to offline CPUs, and cause `sched_setaffinity` to fail later on.
* `fd_shmem`'s CPU topology sizes the cpu<>numa mapping using a call to `get_nprocs()` which returns the number of online CPUs, while the cpu indices are from `/sys/devices/system/cpu/cpuXX` and can be up to the total number of CPU -1.  This results in workspaces assigned to wrong numa nodes, or errors when trying to set an affinity for cpu 63, for instance, on a machine with 64 logical cpu with 2 cpu offline.
* In `irq-affinity.c `, `topo_banned_cpus` only considers cpu with an index < available count, so IRQs on high index CPUs are possibly not moved off.
* In `fd_topob.c`, `available_physical` includes offline CPUs, possibly  enabling wrongly  the 1-tile-per-physical-cpu setup.

### DIFF
--- a/src/app/shared/commands/configure/fd_ethtool_ioctl.c
+++ b/src/app/shared/commands/configure/fd_ethtool_ioctl.c
@@ -86,7 +86,7 @@ fd_ethtool_ioctl_channels_set_num( fd_ethtool_ioctl_t * ioc,
   ech.cmd = ETHTOOL_SCHANNELS;
   if( num == 0 ) {
     uint max_queue_count = ech.max_combined ? ech.max_combined : ech.max_rx;
-    num = fd_uint_min( max_queue_count, (uint)fd_shmem_cpu_cnt() );
+    num = fd_uint_min( max_queue_count, (uint)fd_shmem_available_cpu_cnt() );
   }
   if( ech.max_combined ) {
     ech.combined_count = num;
@@ -123,7 +123,7 @@ fd_ethtool_ioctl_channels_get_num( fd_ethtool_ioctl_t * ioc,
 
   if( FD_LIKELY( ech.combined_count ) ) {
     channels->current = ech.combined_count;
-    channels->max = fd_uint_min( ech.max_combined, (uint)fd_shmem_cpu_cnt() );
+    channels->max = fd_uint_min( ech.max_combined, (uint)fd_shmem_available_cpu_cnt() );
     return 0;
   }
   if( ech.rx_count || ech.tx_count ) {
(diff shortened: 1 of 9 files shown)


mmcgee-jump:
Thanks for the change, I'll review shortly. Just FYI the CPU offline pieces are likely going away soon, as it's been replaced by cpusets (see the new stage cpusets.c), which effectively isolated the pair core from kernel without turning it offline (along with some other configure stages to prevent random work ending up on there).

heronimus:
Thanks!, I'm not trying to give a review for these changes,
but I tested it both on main branch version & also cherry-picked onto the `v1.0.1` tag we deploy on testnet, this fixes our offline-cpu startup issue on our testnet box (48-core, bonded xdp, we manually offline the pack/poh HT siblings), works cleanly both ways :pray:

Also, as mentioned above, I saw the new cpuset stage with the main branch version and it looks like it'll handle isolating the HT sibling too, so I get that it'll likely replace core-offlining. 
But since v1.0.1 still hits the offlining issue today, might be worth landing this ahead of the cpuset stage 🤔

mmcgee-jump:
Not needed I believe after cpusets change?

HGuillemet:
@mmcgee-jump I'm not sure. 
cpusets replaces the isolation that could have been made manually before.
But it seems unrelated to CPU put offline when we want, for instance, to disable SMT for some cores where the tiles running on won't benefit from SMT.

mmcgee-jump:
We isolate the SMT pair so nothing should ever schedule on it, I believe that's functionally equivalent to turning it off.

HGuillemet:
* I'm not sure it's completely equivalent. I think some overhead still exist if SMT is enabled, even if nothing run on the sibbling. But that's probably negligible.
* cpusets is still incompatible with cpu offlines. For instance you use `fd_shmem_cpu_cnt()` as an upper bound for cpu indices, although it returns the number of online CPU. Some indices may be larger if some CPU are offlines.
* That may be ok if putting CPUs offline is now useless. But including the sibling is only performed for `smt_sensitive_tiles` (pack, poh). In one of my setup for instance, I decided to use SMT for verify tiles,  and for plugin+gui, but not for most other tiles, so I disabled the sibblings for these other tiles. I cannot do it, unless I missed sth.

PR #10432, opened 2026-07-10
tower: skip unreadable vote accounts in epoch voter cache

kbhargava-jump:


mjain-jump (commented):
lgtm, will defer to charlie

lidatong (approved):

PR #10439, opened 2026-07-11
Fix non-temporal copy in shred tile

HGuillemet:
The end condition of this loop is obviously wrong.
It does not affect functionality since the memcpy afterwards copies what is not copied in the loop (trashing the cache).
Since this loop probably never run in production, it could be worthy to now check how it affects latency or throughput.
Simple tests pass.

### DIFF
--- a/src/disco/shred/fd_shred_tile.c
+++ b/src/disco/shred/fd_shred_tile.c
@@ -911,7 +911,7 @@ send_shred( fd_shred_ctx_t                 * ctx,
 
   ulong end_offset = shred_sz + sizeof(fd_ip4_udp_hdrs_t);
   ulong i;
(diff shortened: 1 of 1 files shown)


HGuillemet:
Is bypassing the cache a good idea anyway ? The net tile probably shares an L3 cache with the shred tile in an optimized setup. Or even more with SMT if on same physical core.

ptaffet-jump:
🤦 Thanks for catching this. I think `i<=end_offset-64UL` is the cleanest, but we'll run some experiments internally to see if it helps or hurts, as you pointed out. Really the best solution would be to use only one dcache entry for all the copies of the same shred and encode the outgoing IP in the mcache somehow. We've talked about this but never gotten around to it.

PR #10468, opened 2026-07-14
eqvoc: detect same-index equivocation across FEC sets

pcrane:
## eqvoc: detect same-index equivocation across FEC sets

   ### Problem
   `verify_proof` (`src/choreo/eqvoc/fd_eqvoc.c`) has no general same-index check. Its only index comparison
 is gated to data shreds (last-shred conflict), and its other content check requires equal `fec_set_idx`
 (same-FEC merkle conflict). So two coding shreds for one slot with the same coding index but
 different `fec_set_idx` and different payloads — both leader-signed — match neither path and fall through
 to `FD_EQVOC_IGNORED`.

   ### Fix
   Add a general same-index check to `verify_proof`, before the FEC-set-scoped checks: two same-type
 shreds at the same type-appropriate index (coding index for coding shreds, data index for data shreds) with
 differing merkle roots are an equivocation regardless of `fec_set_idx`, returning `FD_EQVOC_SUCCESS`.
 Additive — the existing last-shred and same-FEC-set merkle paths are unchanged.

   Merkle-root inequality is the differing-content proxy (equivalent to Agave's payload compare modulo hash
 collision, and reuses the same idiom as the same-FEC-set check just below).

   ### Commits (red → green)
   1. `eqvoc: add test for same-index equivocation across FEC sets` — adds
 `test_same_coding_index_different_fec_sets`, asserting `verify_proof` detects the pair (`FD_EQVOC_SUCCESS`).
 Fails on current code, documenting the bug test-first.
   2. `eqvoc: detect same-index equivocation across FEC sets` — adds the check; turns the test green.

   ### Verification
   `build/native/gcc/unit-test/test_eqvoc`:
   - With fix: RC=0 — `test_same_coding_index_different_fec_sets` passes; `test_slot_zero_proof_rejected`
 / `test_version_max_rejected` unaffected.
   - Commit 1 alone (test without fix): RC=134 — `FAIL: verify_proof(...)` (SIGABRT), confirming the test
 catches the regression.

### DIFF
--- a/src/choreo/eqvoc/fd_eqvoc.c
+++ b/src/choreo/eqvoc/fd_eqvoc.c
@@ -683,6 +683,24 @@ verify_proof( fd_eqvoc_t *               eqvoc,
     }
   }
 
+  /* Two same-type shreds with the same index (shred->idx) but conflicting
+     merkle roots are an equivocation regardless of FEC set.  Mirrors Agave
+     check_shreds; index() is the common-header index (shred->idx), not the
+     per-FEC-set position shred->code.idx:
+       https://github.com/anza-xyz/agave/blob/v4.2/gossip/src/duplicate_shred.rs#L212-L220
+       https://github.com/anza-xyz/agave/blob/v4.2/ledger/src/shred.rs#L470-L471 */
+
+  int both_data = fd_shred_is_data( fd_shred_type( shred1->variant ) ) &&
+                  fd_shred_is_data( fd_shred_type( shred2->variant ) );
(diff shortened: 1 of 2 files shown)


lidatong (commented):
thanks for pointing this out. this is actually addressed by SIMD-0504 (see `Family 1, 3.`), and currently addressed by the fec resolver here

https://github.com/firedancer-io/firedancer/blob/main/src/disco/shred/fd_fec_resolver.c#L528

lmk if you think there's an additional vulnerability?

pcrane:
Thanks for the review @lidatong - I can't see any further issues, so I'll close this PR.

PR #10469, opened 2026-07-14
Reject Flate2 EpochSlots at wire boundary

TGWDB:
### Summary

  Firedancer's gossip parser accepts EpochSlots messages whose Flate2 (DEFLATE-compressed) slot range carries an invalid payload. `deser_epoch_slots` (`src/flamenco/gossip/fd_gossip_message.c:244-272`)
  `SKIP_BYTES`-es the compressed payload without validating it, so `fd_gossip_message_deserialize` returns success on garbage input and the gossip relay path (`fd_gossip.c:647` / `fd_gossip.c:659`) forwards
  the unchanged frame to peers.

  The Uncompressed and Flate2 branches of the same message type are handled asymmetrically — Uncompressed validates inline, Flate2 does not. This PR makes the two branches symmetric by refusing to deserialize
  a Flate2 slot range at the wire boundary.

  ### Root cause

  `deser_epoch_slots` handles the CRDS `EpochSlots` message's two slot-range variants inconsistently:

  - **Uncompressed** — validates the structure inline; malformed payloads fail parse.
  - **Flate2** — reads `is_uncompressed=0`, then `SKIP_BYTES` past the length-prefixed compressed payload with no DEFLATE-side check. Malformed payloads parse successfully.

  A `SKIP_BYTES` fallthrough on the Flate2 branch means any bytes at all — including non-DEFLATE data — pass the deserializer and reach the relay path.

  ### Fix

  `src/flamenco/gossip/fd_gossip_message.c` — Return failure from `deser_epoch_slots` on the Flate2 variant. The relay path already drops on parse failure, so no other change is required.

  **Choice made.** Two options were considered:

  1. Vendor a DEFLATE decoder (miniz or libz) and inflate-validate the payload.
  2. Refuse to deserialize the Flate2 variant.

  This PR takes option (2) for three reasons:

  - **No consumer state forfeited.** Firedancer has no downstream consumer for EpochSlots. `fd_gossip_message.h:227-229` retains only `{ uchar index }`, and a full-tree grep confirms no code reads slot data
  from either variant.
  - **Symmetric branches.** The Uncompressed branch already validates. This change makes the two branches symmetric without adding a new dependency.
  - **Cost is minimal.** One relay hop's worth of redundancy for valid Flate2 EpochSlots. The originator's other gossip neighbours still propagate the message, and Agave nodes receive the data via direct peers
  or non-Firedancer relays.

  If a future consumer needs the payload, the `return 0` should be replaced with a DEFLATE inflater (miniz vendoring, libz, or equivalent).

  ### Commits (test-first)

  Each commit's tests demonstrably fail on unfixed code and pass with the fix:

  - **`gossip: add tests for EpochSlots Flate2 validation`** — Adds a unit test and an integration test, both parameterized on a malformed DEFLATE payload:
    - `test_epoch_slots_flate2` (unit) — asserts `fd_gossip_message_deserialize` rejects a Push containing an EpochSlots Flate2 slot range whose compressed payload is not a valid DEFLATE stream.
    - `test_epoch_slots_flate2_wire` (integration) — drives the same malformed payload through the real gossvf tile's `handle_net` entry point (the wire-arrival path). Asserts `handle_net` does not return
  `SUCCESS_PUSH` and does not publish a relay-eligible frag.
    Both error messages point at the specific function to fix. **Fail on unfixed code**: unit returns `SUCCESS`, integration returns `SUCCESS_PUSH=2`.

  - **`gossip: reject Flate2 EpochSlots at wire boundary`** — Applies the fix (`deser_epoch_slots` returns failure on the Flate2 variant). Both tests pass. Integration test observes `handle_net` returning
  `DROPPED_UNPARSEABLE=6` and no relay-eligible frag published.

  ### Test plan

  - [ ] `make -j test_epoch_slots_flate2 && ./build/native/gcc/unit-test/test_epoch_slots_flate2` → PASS
  - [ ] `make -j test_epoch_slots_flate2_wire && ./build/native/gcc/integration-test/test_epoch_slots_flate2_wire` → PASS
  - [ ] Existing gossip test suite → no regressions
  - [ ] `make -j firedancer-dev` → builds clean

  ### Files

  | File | Change |
  |---|---|
  | `src/flamenco/gossip/fd_gossip_message.c` | +8 / −3 (reject Flate2 variant in `deser_epoch_slots`) |
  | `src/flamenco/gossip/test_epoch_slots_flate2.c` | +95 (new unit test) |
  | `src/flamenco/gossip/Local.mk` | +3 (register unit test) |
  | `src/discof/gossip/test_epoch_slots_flate2_wire.c` | +418 (new integration test) |
  | `src/discof/gossip/Local.mk` | +3 (register integration test) |

  ### Scope and limitations

  - Firedancer-only. Agave's EpochSlots handling is unchanged.
  - Valid Flate2 EpochSlots are dropped by Firedancer relays. This is acceptable because no in-tree consumer reads the payload (see rationale above) and gossip has other propagation paths.
  - Does not remove the Flate2 code path — the variant is still recognized at parse; only the payload is refused. If a future consumer needs EpochSlots data, an inflater can be added at the same site.

### DIFF
--- a/src/discof/gossip/Local.mk
+++ b/src/discof/gossip/Local.mk
@@ -3,4 +3,7 @@ $(call add-objs,fd_gossip_tile,fd_discof)
 $(call add-objs,fd_gossvf_tile,fd_discof)
 $(call make-fuzz-test,fuzz_gossvf_tile,fuzz_gossvf_tile,fd_disco fd_flamenco fd_ballet fd_tango fd_util)
 $(call make-fuzz-test,fuzz_gossvf_gossip_pair,fuzz_gossvf_gossip_pair,fd_disco fd_flamenco fd_ballet fd_tango fd_util)
+
+$(call make-integration-test,test_epoch_slots_flate2_wire,test_epoch_slots_flate2_wire,fd_disco fd_flamenco fd_ballet fd_tango fd_util)
+$(call run-integration-test,test_epoch_slots_flate2_wire)
 endif
--- a/src/discof/gossip/test_epoch_slots_flate2_wire.c
+++ b/src/discof/gossip/test_epoch_slots_flate2_wire.c
@@ -1,0 +1,418 @@
+#if !FD_HAS_HOSTED
+#error "This target requires FD_HAS_HOSTED"
+#endif
+
+/* BP707 — flate2: Integration test enforcing that a malformed
+   Flate2-compressed EpochSlots payload arriving from the network is
+   rejected at the wire boundary by the real firedancer gossvf tile.
+
+   This test FAILS while the bug is present and PASSES once the fix
+   lands.  Today's behaviour (the bug):
+
+     - deser_epoch_slots (fd_gossip_message.c:244-272) skips the
+       compressed bytes without validating them.
+     - handle_net (fd_gossvf_tile.c:773) therefore returns SUCCESS_PUSH
+       and the gossvf tile publishes the payload to the gossip tile,
+       which forwards `payload + value->offset` for `value->length`
+       bytes verbatim via fd_active_set_push (fd_gossip.c:659).
+
+   Expected behaviour after the fix (audit §6.1, "validate but don't
+   consume"): the malformed DEFLATE stream is detected and rejected
+   before the payload is forwarded — handle_net returns a DROPPED_*
+   outcome and no relay-eligible frag is published.
+
+   What this exercises:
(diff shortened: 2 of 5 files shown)


ripatel-fd:
Breaking parsing of individual CRDS values will cause wire incompatibilities with Agave* and drop other CRDS packaged in the same message.

PR #10470, opened 2026-07-14
Fix false-positive equivocation on same-merkle-root FEC set

TGWDB:
### Summary

  Firedancer's FEC resolver returns `FD_FEC_RESOLVER_SHRED_EQUIVOC` when two shreds in the same FEC set carry different-but-valid ed25519 signatures over the **same** reconstructed merkle root. This is a false
  positive: the two shreds are consistent (same merkle root, same reconstructable content), so the FEC set should complete normally. Instead, the incorrect EQUIVOC verdict propagates through the shred/tower
  tiles into `fd_ghost_eqvoc`, marks the block invalid, and causes FD to abstain from voting on a block Agave votes on — a fork-choice divergence between the two clients.

  The fix aligns FD's semantics with Agave's `check_merkle_root_consistency` (`agave/ledger/src/blockstore.rs:3010`): only flag equivocation when merkle roots actually differ.

  ### Root cause

  `fd_fec_resolver` keys its in-progress FEC-set map by ed25519 signature bytes (`MAP_KEY = sig`). The `(slot, fec_set_idx)` treap is a secondary index whose only purpose is catching "same FEC set, different
  signature". Any hit sets `equivoc_or_invalid=1` and — after per-shred sigverify succeeds — the resolver returns `SHRED_EQUIVOC` without checking whether the merkle roots actually differ.

  This produces a false positive whenever a leader emits two valid signatures over the same merkle root. RFC-8032 deterministic-`k` is a signer-side property, not a verifier requirement — signers that use
  randomized `k` (permitted by RFC 8032) or that lose deterministic-nonce state across a restart can produce byte-different signatures for the same `(sk, M)` pair that both pass verify.

  The buggy propagation chain:

  ```
  fd_fec_resolver EQUIVOC (false positive)
    -> fd_shred_tile.c:997        (SHRED_SIG_RESULT_EQVOC)
    -> fd_tower_tile.c:1732       (fd_eqvoc_shred_insert, shred_hint=1)
    -> fd_eqvoc.c:717-728         (dup_insert)
    -> fd_tower_tile.c:1268-1275  (fd_ghost_eqvoc on SLOT_COMPLETED)
    -> fd_ghost.c:638             (mark_invalid zeros valid on subtree)
    -> FD refuses to vote for the slot; Agave votes normally.
  ```

  ### Fix

  `src/disco/shred/fd_fec_resolver.c` — Before returning `SHRED_EQUIVOC` on the treap-hit path:

  1. Look up the existing FEC set's merkle root (via `ctx_treap` for in-progress or `done_map` for completed).
  2. If it matches the newly reconstructed root → return `SHRED_IGNORED` (Agave-equivalent behavior).
  3. If it differs → keep the existing `SHRED_EQUIVOC` path with `SIG_HASH_EQUIVOC` stamp (genuine equivocation).

  To support (1), `done_ele_t` gains a 32-byte `merkle_root` field (struct grows 32→64 bytes; `FD_STATIC_ASSERT` updated). Populated at both the FEC-set completion path and the equivoc-stamp path.

  ### Commits (test-first)

  Each commit's tests demonstrably fail on unfixed code and pass with the fix:

  - **`divergence(BP627)`** — Adds `src/disco/shred/test_fec_resolver_bp627.c`. Crafts a valid 32-data/32-parity chained-merkle FEC set at `slot=100, fec_set_idx=0`, then produces a second valid ed25519
  signature `Sig_B` over the same merkle root via an RFC-8032-compatible signer with a distinct nonce. Asserts `shred_A` returns `OKAY`, `shred_B` returns `IGNORED`, resent `shred_B` returns `IGNORED`. Also
  adds a `test_bp627_real_equivocation` subtest: two distinct FEC sets at the same `(slot, fec_set_idx)` with different merkle roots must still return `EQUIVOC` (regression guard). **Fails on unfixed 
  resolver** (returns `EQUIVOC=-4`, test expects `IGNORED=-2`).

  - **`integration(BHP-627)`** — Adds `src/discof/tower/test_bhp_627_integration/`. Drives the resolver end-to-end with the same crafted shred pair, builds a minimal ghost tree (`root → eqvoc → child`), and
  simulates the tower-tile propagation: iff the resolver reports `EQUIVOC`, invokes `fd_ghost_eqvoc` on the eqvoc slot's block-id. Asserts `eqvoc_blk->valid == 1`, `fd_ghost_best(root) == child`, and
  `fd_ghost_invalid_ancestor(child) == NULL`. Companion `RUN.md` documents the test and specifies the two-validator co-run harness that would confirm the fork-choice divergence across the full tile stack.
  **Fails on unfixed resolver** (`EQUIVOC` → propagation → `mark_invalid` → `valid == 0`, assertion trips).

  - **`fix(BP627)`** — Applies the resolver fix. Both tests pass.

  ### Test plan

  - [ ] `make -j test_fec_resolver_bp627 && ./build/native/gcc/unit-test/test_fec_resolver_bp627` → PASS (same-root: `OKAY → IGNORED → IGNORED`; different-root: `OKAY → EQUIVOC`).
  - [ ] `make -j test_bhp_627_integration && ./build/native/gcc/unit-test/test_bhp_627_integration --page-sz normal --page-cnt 512` → PASS (resolver returns `IGNORED`, ghost block stays valid).
  - [ ] `make -j test_fec_resolver && ./build/native/gcc/unit-test/test_fec_resolver` → PASS (neighbor suite, no regression).
  - [ ] `make -j firedancer-dev` → builds clean.

  ### Files

  | File | Change |
  |---|---|
  | `src/disco/shred/fd_fec_resolver.c` | +64 / −15 (merkle-root compare, `done_ele_t` merkle_root field) |
  | `src/disco/shred/test_fec_resolver_bp627.c` | +422 (new unit test) |
  | `src/disco/shred/Local.mk` | +2 (register `test_fec_resolver_bp627`) | 
  | `src/discof/tower/test_bhp_627_integration/test_bhp_627_integration.c` | +360 (new integration test) |
  | `src/discof/tower/test_bhp_627_integration/RUN.md` | +96 (run notes + harness spec) |
  | `src/discof/tower/Local.mk` | +1 (register `test_bhp_627_integration`) |

  ### Scope and limitations

  - Firedancer-only; targets the C-side FEC resolver. No Agave-side changes needed.
  - The integration test compresses the tile-layer propagation into a single conditional call to `fd_ghost_eqvoc` — it does not run the full shred → tower → vote-emit tile stack. `RUN.md` specifies the
  two-validator co-run harness that would confirm the fork-choice divergence in production shape.
  - No repair-path change: `is_repair` continues to bypass the treap check (the fix acts on the same Turbine-ingest path where the false positive originally manifested).

### DIFF
--- a/src/disco/shred/Local.mk
+++ b/src/disco/shred/Local.mk
@@ -6,6 +6,8 @@ $(call add-objs,fd_fec_resolver,fd_disco)
 $(call add-objs,fd_shred_tile,fd_disco)
 $(call make-unit-test,test_fec_resolver,test_fec_resolver,fd_flamenco fd_disco fd_ballet fd_util fd_tango fd_reedsol)
 $(call run-unit-test,test_fec_resolver)
+$(call make-unit-test,test_fec_resolver_bp627,test_fec_resolver_bp627,fd_flamenco fd_disco fd_ballet fd_util fd_tango fd_reedsol)
+$(call run-unit-test,test_fec_resolver_bp627)
 endif
 $(call make-unit-test,test_shred_dest,test_shred_dest,fd_disco fd_flamenco fd_ballet fd_util)
 $(call make-unit-test,test_shred_dest_conformance,test_shred_dest_conformance,fd_disco fd_flamenco fd_ballet fd_util)
--- a/src/disco/shred/fd_fec_resolver.c
+++ b/src/disco/shred/fd_fec_resolver.c
@@ -117,21 +117,27 @@ struct done_ele {
   uint            heap_right;
   uint            map_next;
   uint            map_prev;
-  /* In order to save space in the done_map and make this struct 32
-     bytes, we store a 32 bit validator-specific hash of the shred
-     signature.  If a malicious leader equivocates and produces two FEC
-     sets which have the same hash for us, a task which takes a decent
-     but doable amount of effort, the only impact is that we would
-     reject the shreds with SHRED_IGNORED instead of SHRED_EQUIVOC,
-     which is not a big deal.  It's documented that SHRED_EQUIVOC
-     detection is on a best-effort basis.  If we detect an equivocation
-     for this (slot, FEC set idx), sig_hash gets set to
-     SIG_HASH_EQUIVOC, and we start returning SHRED_IGNORED for any
-     non-repair shred for that (slot, FEC set idx). */
+  /* We store a 32 bit validator-specific hash of the shred signature.
+     If a malicious leader equivocates and produces two FEC sets which
+     have the same hash for us, a task which takes a decent but doable
+     amount of effort, the only impact is that we would reject the
+     shreds with SHRED_IGNORED instead of SHRED_EQUIVOC, which is not a
+     big deal.  It's documented that SHRED_EQUIVOC detection is on a
+     best-effort basis.  If we detect an equivocation for this (slot,
+     FEC set idx), sig_hash gets set to SIG_HASH_EQUIVOC, and we start
+     returning SHRED_IGNORED for any non-repair shred for that
+     (slot, FEC set idx). */
   uint           sig_hash;
+  /* Merkle root of the completed FEC set.  Used to distinguish a
+     signature-only mismatch (BP627: two valid signatures over the same
+     merkle root -- not equivocation, matches Agave semantics) from a
+     genuine equivocation (different merkle roots for the same
+     (slot, fec_set_idx)).  Populated on FEC set completion and on the
+     equivoc-detection path. */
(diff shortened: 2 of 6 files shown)


ripatel-fd (changes requested):
Please remove the random AI slop from this commit (RUN.md, references to "BP627"). This is a huge amount of changes for a relatively simple problem

lidatong:
this is a good find ty. we will look into making a fix separately

lidatong:
closing this PR to reopen as issue https://github.com/firedancer-io/firedancer/issues/10490

PR #10474, opened 2026-07-14
repair: fix OOB read in `check_confirmed` on partial last FEC set

peterschrammel:
Fixes an out-of-bounds read in the repair tile's chained-merkle verification path (`check_confirmed`, `src/discof/repair/fd_repair_tile.c`).

  When selecting the expected merkle root for a FEC set that failed chain verification, the guard compared **shred** indices:

  ```c
  fd_hash_t const * expected =
      (bad_fec_idx == bad_blk->complete_idx - (FD_FEC_SHRED_CNT - 1))
          ? &bad_blk->confirmed_bid
          : &bad_blk->merkle_roots[(bad_fec_idx / 32) + 1].cmr;
  ```

  This guard only matches the last FEC set when it is **full** (i.e. `complete_idx` is of the form `N*32 - 1`). If the last FEC set is **partial** (`complete_idx % 32 != 31`), the   guard is false and the else-branch indexes `merkle_roots[(bad_fec_idx / 32) + 1]`. When the bad FEC is the highest possible set (index 1023, `complete_idx` in `[32736, 32766]`), this reads `merkle_roots[1024]` — one element past the end of the 1024-element (`FD_FEC_BLK_MAX`) array.

  ### Impact

  `merkle_roots` sits in `fd_forest_blk_t`, so the read lands in adjacent struct memory (`confirmed_bid` / `lowest_verified_fec`). Consequences:

  - An incorrect "expected" merkle root is logged and fed into the subsequent `fd_forest_fec_clear` decision.
  - Under sanitizer builds (`-fsanitize=bounds` / ASan) the validator aborts.

 A leader may set `FD_SHRED_DATA_FLAG_SLOT_COMPLETE` on a shred whose index is not of the form `N*32 - 1` (nothing in the protocol forbids this), then a   tower confirmation arrives with a block_id that does not match the last FEC set. The repair tile processes shreds from other validators, so this is externally reachable.

  ### Fix

  Compare **FEC-set** indices instead of shred indices:

  ```c
  uint bad_fec_set = bad_fec_idx / 32U;
  fd_hash_t const * expected =
      (bad_fec_set == bad_blk->complete_idx / 32U)
          ? &bad_blk->confirmed_bid
          : &bad_blk->merkle_roots[bad_fec_set + 1].cmr;
  ```

  This selects `confirmed_bid` whenever the bad FEC set is the last one in the slot — full or partial — so the `+1` access never exceeds `FD_FEC_BLK_MAX - 1`.

  ### Testing

  Added `test_check_confirmed_partial_last_fec_oob` to `test_repair_tile.c`. It drives the real `after_tower` → `check_confirmed` path with a partial last FEC set (`complete_idx = 32766`, FEC index 1023) and a mismatching tower confirmation.

  - **Before the fix**, under `EXTRAS=ubsan`:

    ```
    fd_repair_tile.c: runtime error: index 1024 out of bounds for type 'struct (...)[1024]'
        #0 check_confirmed  fd_repair_tile.c
        #2 test_check_confirmed_partial_last_fec_oob  test_repair_tile.c
    ```

  - **After the fix**: no OOB diagnostic; all repair tile tests pass. Also confirmed a clean default `gcc`/`native` build.

  Build/run:

  ```bash
  make -j BUILDDIR=clang-ubsan CC=clang EXTRAS=ubsan test_repair_tile
  ./build/clang-ubsan/unit-test/test_repair_tile
  ```

  ### Commits

  - `repair: add test exposing OOB read in check_confirmed` — test-first, confirms the OOB is present.
  - `repair: fix OOB read in check_confirmed on partial last FEC set` — the fix.

### DIFF
--- a/src/discof/repair/fd_repair_tile.c
+++ b/src/discof/repair/fd_repair_tile.c
@@ -836,7 +836,15 @@ check_confirmed( ctx_t           * ctx,
        early exit above. */
     FD_TEST( bad_fec_idx != UINT_MAX );
 
-    fd_hash_t const * expected = (bad_fec_idx == bad_blk->complete_idx - (FD_FEC_SHRED_CNT - 1)) ? &bad_blk->confirmed_bid : &bad_blk->merkle_roots[(bad_fec_idx / 32) + 1].cmr;
+    /* Compare in FEC-set-index space rather than shred-index space.
+       The shred-index guard (bad_fec_idx == complete_idx - 31) only
+       matches when the last FEC set is full; for a partial last FEC set
+       it is false and the else-branch would index merkle_roots[1024]
+       (one past the end) when the bad FEC is the last one (index 1023).
+       Selecting confirmed_bid whenever the bad FEC set is the last one
+       in the slot keeps the +1 access within bounds. */
+    uint bad_fec_set = bad_fec_idx / 32U;
+    fd_hash_t const * expected = (bad_fec_set == bad_blk->complete_idx / 32U) ? &bad_blk->confirmed_bid : &bad_blk->merkle_roots[bad_fec_set + 1].cmr;
 
     FD_BASE58_ENCODE_32_BYTES( confirmed_bid->uc,                             confirmed_bid_b58 );
     FD_BASE58_ENCODE_32_BYTES( expected->uc,                                  expected_mr );
--- a/src/discof/repair/test_repair_tile.c
+++ b/src/discof/repair/test_repair_tile.c
@@ -636,6 +636,96 @@ test_after_tower_confirmed_eviction( fd_wksp_t * wksp ) {
   FD_LOG_NOTICE(( "pass: test_after_tower_confirmed_eviction" ));
 }
 
+/* test_check_confirmed_partial_last_fec_oob
+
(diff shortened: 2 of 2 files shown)


intrigus-lgtm:
I looked at this code a while ago as well and I don't see how this could ever be reachable in practice?
> A leader may set FD_SHRED_DATA_FLAG_SLOT_COMPLETE on a shred whose index is not of the form N*32 - 1 (nothing in the protocol forbids this)

fd_shred/fd_fec_resolver should very much prevent any such shred from reaching this part. Did you find a way to bypass the existing checks?

peterschrammel:
Thanks for your review, @intrigus-lgtm 

> fd_shred/fd_fec_resolver should very much prevent any such shred from reaching this part. 

I confirm that you are right.
I reran the test generation including the `fd_fec_resolver` layer and all cases that would trigger this issue are actually caught by https://github.com/firedancer-io/firedancer/blob/c288cec43ac76508521196f64accee14c22f2ac1/src/disco/shred/fd_fec_resolver.c#L546
If desired, I can add a comment/assertion in `repair/fd_repair_tile.c` that documents this assumption.

Closing this PR.

PR #10475, opened 2026-07-14
agave: improve handling of equivocating shreds

ptaffet-jump:


topointon-jump (approved):
Checked that this logic is equivalent to the handling in Agave's `window_service`

PR #10502, opened 2026-07-15
boot: remove landlock warning

mmcgee-jump:


ripatel-fd (approved):

ripatel-fd:
Fine with it , but why? it's nice to remind users to upgrade the kernel

mmcgee-jump:
Doesn't rise to the level of operator warning, it's an extreme defense in depth that does basically nothing.

PR #10520, opened 2026-07-16
runtime: support custom_commission_collector

ripatel-fd:


src/flamenco/stakes/fd_vote_stakes.h
```diff
@@ -174,6 +174,16 @@ fd_vote_stakes_root_insert_key( fd_vote_stakes_t *  vote_stakes,
...
+void
+fd_vote_stakes_root_insert_key1( fd_vote_stakes_t *  vote_stakes,
+                                 fd_pubkey_t const * pubkey,
+                                 fd_pubkey_t const * node_account_t_1,
+                                 fd_pubkey_t const * inflation_rewards_collector_t_1,
+                                 fd_pubkey_t const * block_revenue_collector_t_1,
+                                 ulong               stake_t_1,
+                                 ushort              commission_t_1,
+                                 ulong               epoch );
+
```
topointon-jump: I know it increases the scope of this change, but I don't really love having these `key`/`key1` and `meta`/`meta1` apis - can we not just have `key`/`meta`?

ripatel-fd: @topointon-jump will fix

ripatel-fd:
+10GB memory usage :/

ripatel-fd:
This implementation is not good I'm redoing this.

PR #10542, opened 2026-07-20
snapshot: remove feature snooping

mjain-jump:


ripatel-fd:
Why is snapshot loading aware of feature accounts? There is a tiny number of feature accounts, we can ditch this complexity by just querying them after load.

mjain-jump:
> Why is snapshot loading aware of feature accounts? There is a tiny number of feature accounts, we can ditch this complexity by just querying them after load.

Are you saying we should ditch this function and instead just restore the feature set after snapshot load?

ripatel-fd:
@mjain-jump Why close this? the feature snooping code is nonsense

ripatel-fd:
https://github.com/firedancer-io/firedancer/pull/11488

PR #10553, opened 2026-07-21
flamenco: stricter fp rounding rules

0x0ece:


topointon-jump (approved):
Personally I think we should also add `-ffp-contract=off` to the build flags to make sure there are no more of these.

PR #10554, opened 2026-07-22
build: remove -ffast-math

0x0ece:
gcc / clang options to match rust
```
# required — these differ from the compiler defaults
-ffp-contract=off            # gcc defaults to `fast`, clang to `on`; Rust never fuses a*b+c
-fno-math-errno              # Rust has no errno for math; both default to -fmath-errno

# required — never pass these (or -Ofast, which implies them)
# -ffast-math -funsafe-math-optimizations -ffinite-math-only
# -fno-signed-zeros -fassociative-math -freciprocal-math

# already correct at default, safe to state explicitly
-fno-associative-math -fno-reciprocal-math -fno-finite-math-only
-fsigned-zeros -fno-rounding-math -fno-signaling-nans
-fexcess-precision=standard  # implied by -std=c17; needed explicitly under -std=gnu*
-mfpmath=sse                 # default on x86-64; no x87 excess precision```

### CODE REVIEW

File: config/extra/with-optimization.mk [hunk 1]
```diff
@@ -1,9 +1,15 @@
...
+CPPFLAGS+=-O3 -ffast-math -fno-associative-math -fno-reciprocal-math -ffp-contract=off
 CPPFLAGS+=-DFD_HAS_OPTIMIZATION=1
 FD_HAS_OPTIMIZATION:=1
 else
-CPPFLAGS+=-O0 -ffast-math -fno-associative-math -fno-reciprocal-math
+CPPFLAGS+=-O0 -ffast-math -fno-associative-math -fno-reciprocal-math -ffp-contract=off
+endif
+
+# -ffp-contract=off overrides -ffast-math
+# clang errors on ovverride, but since this is intentional we relax it
```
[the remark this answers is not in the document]
0x0ece: fixed

0x0ece:
Closing because it doesn't really fix anything. We'll revisit in future.

config/extra/with-optimization.mk
```diff
@@ -1,9 +1,18 @@
 ifeq ($(FD_DISABLE_OPTIMIZATION),)
-CPPFLAGS+=-O3 -ffast-math -fno-associative-math -fno-reciprocal-math
+CPPFLAGS+=-O3 -ffast-math -fno-associative-math -fno-reciprocal-math -ffp-contract=off
 CPPFLAGS+=-DFD_HAS_OPTIMIZATION=1
 FD_HAS_OPTIMIZATION:=1
 else
-CPPFLAGS+=-O0 -ffast-math -fno-associative-math -fno-reciprocal-math
+CPPFLAGS+=-O0 -ffast-math -fno-associative-math -fno-reciprocal-math -ffp-contract=off
```
ripatel-fd: We should remove ffast-math instead

0x0ece: done

drubin-fd (approved):

ripatel-fd (approved):

PR #10555, opened 2026-07-22
accdb: fix fork id off by 1 on boot

ibhatt-jumptrading:
currently leak 1 fork id at boot, creates off-by-1 capacity for bank capacity

### DIFF
--- a/src/discof/replay/test_replay_tile.c
+++ b/src/discof/replay/test_replay_tile.c
@@ -18,8 +18,12 @@
    will be a no-op. ---- */
 
 #include "../../flamenco/runtime/fd_bank.h"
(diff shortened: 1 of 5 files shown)


src/discof/replay/test_replay_tile.c
```diff
@@ -23,6 +23,7 @@
 #include "../../flamenco/runtime/fd_txncache.h"
 #include "../../flamenco/rewards/fd_stake_rewards.h"
 #include "fd_sched.h"
+#include <setjmp.h>
```
mmcgee-jump: wtf?

ibhatt-jumptrading: 🤷

yufeng-jump (approved):

PR #10558, opened 2026-07-22
replay: defer shred version check with wait-for-supermajority

ripatel-fd:
Fixes a bug where wait-for-supermajority fails when loading a
full snapshot before the restart slot, and an incremental at the
restart slot.

So, instead, have the client defer shred version checks until the
wait-for-supermajority stage is reached.


jvarela-jump (approved):
Following review from jherrera-jump and testing by ripatel-fd.

PR #10559, opened 2026-07-22
[v1.1 backport] replay: defer shred version check with wait-for-supermajority

ripatel-fd:
Backport of #10558


jvarela-jump (approved):
Following review from jherrera-jump and testing by ripatel-fd.

PR #10567, opened 2026-07-24
runtime: clean up deplete_cu_meter_on_vm_failure

mjain-jump:


mjain-jump:
Waiting on https://github.com/anza-xyz/agave/pull/14116 to be cut into a release before merging

PR #10568, opened 2026-07-26
quic: implement PATH_CHALLENGE response (RFC 9000 Section 8.2)

### LINKED ISSUE #3791: Respond to QUIC path challenges
fd_quic currently ignores PATH_CHALLENGE frames.
This is a potential source of conn failures since TPU clients might require a path challenge before sending stream data.

sabijuraa:
Right now fd_quic receives PATH_CHALLENGE frames and does nothing with them.
RFC 9000 §8.2 says the recipient MUST send back a PATH_RESPONSE with the
same 8 bytes — if it doesn't, path validation fails on the sender's side
and some clients won't proceed with sending stream data.

I ran into this while looking at the QUIC networking layer after my
RESET_STREAM finding. Noticed #3791 had been sitting open for a while
so wrote the fix.

The implementation follows the same pattern as PING frames — store the
8-byte payload on the connection struct, set a flag, generate the response
in the next packet's frame assembly pass. No retransmission on my end
since §8.2.4 says PATH_RESPONSE is sent once and the sender is responsible
for retrying with a new PATH_CHALLENGE if it doesn't hear back.

Changes are in two files:

fd_quic_conn.h gets a new flag (FD_QUIC_CONN_FLAGS_PATH_RESPONSE, bit 6)
and an 8-byte field to hold the challenge data between receiving the
challenge and generating the response.

fd_quic.c: the existing handler now stores the data and sets the flag
instead of ignoring the frame, and there's a new 26-line generator
(fd_quic_gen_path_response_frame) that writes the response frame when
the flag is set.

The test_quic_proto.c already has test_path_response covering the
encoding side. Happy to add handler integration tests if you want them
in a specific location.

Fixes #3791

### DIFF
--- a/src/waltz/quic/fd_quic.c
+++ b/src/waltz/quic/fd_quic.c
@@ -3414,6 +3414,32 @@ fd_quic_gen_ping_frame( fd_quic_conn_t             * conn,
   return frame_sz;
 }
 
+static ulong
+fd_quic_gen_path_response_frame( fd_quic_conn_t * conn,
+                                 uchar          * payload_ptr,
+                                 uchar          * payload_end ) {
+
+  if( ~conn->flags & FD_QUIC_CONN_FLAGS_PATH_RESPONSE ) return 0UL;
+
+  fd_quic_path_response_frame_t response = {
+    .data = conn->path_response_data
+  };
+
+  ulong frame_sz = fd_quic_encode_path_response_frame( payload_ptr,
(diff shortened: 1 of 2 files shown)


ripatel-fd:
Thank you for the contribution. fd_quic hardcodes disable_active_migration, so any PATH_CHALLENGE frames should be treated as a protocol violation.

PR #10570, opened 2026-07-27
xdp: improve frame recycling

ripatel-fd:
The XDP frame recycling code was written assuming the 'chunk'
(64 byte aligned pointer) of a packet does not change.  With very
large XDP headroom and GRE decapsulation (which shifts the start
of a packet), this can technically lead to frames overlapping and
drifting, although this is not reachable today.

Any freed frame pointers are now rebased down to FD_NET_MTU buffer
alignment.

### DIFF
--- a/src/disco/net/xdp/fd_xdp_tile.c
+++ b/src/disco/net/xdp/fd_xdp_tile.c
@@ -232,8 +232,7 @@ typedef struct {
   ulong    umem_sz; /* Size  of UMEM */
 
   /* UMEM chunk region within workspace */
-  uint     umem_chunk0; /* Lowest allowed chunk number */
-  uint     umem_wmark;  /* Highest allowed chunk number */
+  uint     umem_chunk0; /* Chunk number of the first byte of UMEM */
 
   /* All net tiles are subscribed to the same TX links.  (These are
      incoming links from app tiles asking the net tile to send out packets)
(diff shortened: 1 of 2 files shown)


src/disco/net/xdp/fd_xdp_tile.c
```diff
@@ -1234,18 +1233,17 @@ net_rx_event( fd_net_ctx_t * ctx,
...
-  }
-
   FD_STATIC_ASSERT( FD_ULONG_IS_POW2( FD_NET_MTU ), "FD_NET_MTU must be a power of two" );
-  uint  fill_prod  = fill_ring->cached_prod;
-  uint  fill_mask  = (fill_ring->depth)-1U;
   ulong frame_mask = FD_NET_MTU - 1UL;
-  ulong freed_off  = (freed_chunk - ctx->umem_chunk0)<<FD_CHUNK_LG_SZ;
-  fill_ring->frame_ring[ fill_prod&fill_mask ] = freed_off & (~frame_mask);
+  ulong freed_off  = ( (ulong)( freed_chunk - ctx->umem_chunk0 )<<FD_CHUNK_LG_SZ ) & (~frame_mask);
+  if( FD_UNLIKELY( freed_off+FD_NET_MTU > ctx->umem_sz ) ) {
```
tristanx86-jump, line 1254: This is nice, maybe a comment saying this checks both the lower and upper bound through unsigned underflow, but not necessary

tristanx86-jump (approved):
Looks good!

PR #10575, opened 2026-07-27
util: bench fd_hash ns/byte

0xfka:
Added the bench because of the   /* TODO: This takes >100 cycles! */
at file src/disco/pack_fd_pack.c:361 requires a feedback loop first. 
 I got the --extra-bench name from another test, which has performance tests ( src/disco/pack/test_pack.c ) but I'm not sure it explains what the code is doing well. 
The thing is, the cycles/op value is affected so much by the buf variable. When I add a longer text to buf, the cycles/op is even increasing to 170+ cycles/op, so I'm open for ideas. I think a longer text will be better so we can actually see small improvements easier, think current one as a dummy. 
Tried to provide more info at the benchmark, such as p99, jitter etc. with histf, but it requires to call functions one by one, which means 2 fd_tickcount() per function call, so high noise. It even makes it impossible to understand how much cycles/op the function is. 
For making sure the function is really calculated, I removed all the code but this test, run perf report, and saw realistic cycles/op values at instructions:
```perf
  18.89 │       cmp    r10,rcx                                                                                             ▒
        │     ↑ jne    2a0                                                                                                 ▒
        │       mov    rax,rdx                                                                                             ▒
        │       nop                                                                                                        ▒
        │2c0:   mov    rdx,rax                                                                                             ▒
        │       shr    rdx,0x21                                                                                            ▒
   4.53 │       xor    rax,rdx                                                                                             ▒
        │       movabs rdx,0xc2b2ae3d27d4eb4f                                                                              ▒
        │       imul   rax,rdx                                                                                             ▒
   5.68 │       mov    rbx,QWORD PTR [rbp-0x28]                                                                            ▒
        │       leave                                                                                                      ▒
        │       mov    rdx,rax                                                                                             ▒
        │       shr    rdx,0x1d                                                                                            ▒
   6.22 │       xor    rdx,rax                                                                                             ▒
        │       movabs rax,0x165667b19e3779f9                                                                              ▒
        │       imul   rdx,rax                                                                                             ▒
  14.99 │       mov    rax,rdx                                                                                             ▒
        │       shr    rax,0x20                                                                                            ▒
   7.08 │       xor    rax,rdx                                                                                             ▒
        │     ← ret
```

### DIFF
--- a/src/util/test_util_base.c
+++ b/src/util/test_util_base.c
@@ -121,6 +121,7 @@ main( int     argc,
       char ** argv ) {
   fd_boot( &argc, &argv );
 
+  int extra_benchmark = fd_env_strip_cmdline_contains( &argc, &argv, "--extra-bench" );
   fd_rng_t _rng[1]; fd_rng_t * rng = fd_rng_join( fd_rng_new( _rng, 0U, 0UL ) );
 
   /* Test signed integer overflow is wrapping.  Needs to be at run time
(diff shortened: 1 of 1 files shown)


ripatel-fd:
A hash function's runtime is roughly proportional to its input size.
Therefore hash functions usually measure cycles per byte (normalized over input size).
Your code is currently not measuring CPU cycles because the invariant TSC ticks at a different rate than the CPU.

0xfka:
working on it, thanks

ripatel-fd:
I suggest measuring nanoseconds/byte, you can see this in the benchmark code of some of the other hash functions.

0xfka:
@ripatel-fd 
NOTICE  07-27 21:49:17.767217 0         test_util_base.c(465): fd_hash: 0.075 ns/byte (sz 8192)
NOTICE  07-27 21:46:07.958762 0         test_util_base.c(465): fd_hash: 0.080 ns/byte (sz 1024)
NOTICE  07-27 21:48:23.608619 0         test_util_base.c(465): fd_hash: 0.256 ns/byte (sz 32)
also tried gbps, I think that works. 
edit: I misunderstood what you are trying to say at first so I'm updating this comment to prevent confusion.

0xfka:
will fix in a minute, don't know how its happened.

ripatel-fd (approved):

ripatel-fd:
Shouldn't have merged, the calculation is nonsense

PR #10580, opened 2026-07-27
firedancer: txn v1 + larger txn sizes

topointon-jump:
Implements `enable_tx_v1` (https://github.com/solana-foundation/solana-improvement-documents/pull/385 and https://github.com/solana-foundation/solana-improvement-documents/pull/296)


src/ballet/txn/fd_txn.h
```diff
@@ -86,34 +85,61 @@
...
+
+/* FD_TXN_MTU_V0: The maximum size (in bytes, inclusive) of a serialized
+   legacy or V0 transaction.
+
+   Any transaction that has an MTU larger than this is required to
+   use the transaction V1 format. */
+#define FD_TXN_MTU_V0               (1232UL)
 
 /* FD_TXN_MIN_SERIALIZED_SZ: The minimum size (in bytes) of a serialized
    transaction, using fd_txn_parse() verification rules. */
```
ptaffet-jump: Just add a comment along the lines of "For V1, this is 138"

src/ballet/txn/fd_txn.h
```diff
@@ -273,8 +309,32 @@ struct fd_txn {
...
+     Bit 2:   compute unit limit
+     Bit 3:   loaded accounts data size
+     Bit 4:   program heap size
+
+     For V1 transactions, only bits 0-4 can be non-zero. Compute
+     budget program instructions are silently ignored and the values
+     from the config values region are used instead.
+     For legacy/V0 transactions, all bits are zeroed as these values
+     cannot be specified. */
+  uchar      v1_txn_config_mask;
```
ptaffet-jump: Let me see how you use this field, but my instinct is that we don't want it. It's at a constant offset in the payload, and it might be bigger than 8 bits.

src/ballet/txn/fd_txn.h
```diff
@@ -714,6 +775,77 @@ fd_txn_parse( uchar const * payload, ulong payload_sz, void * out_buf, fd_txn_pa
...
+
+   The results of the parsing are written to the
+   out_priority_fee_lamports, out_compute_unit_limit,
+   out_loaded_accounts_data_sz_limit and out_heap_size fields.  If a
+   field is not specified by the config mask the defaults as per
+   SIMD 385 are used.
+
+   https://github.com/anza-xyz/agave/blob/v4.2.0-beta.1/runtime-transaction/src/runtime_transaction/transaction_view.rs#L98-L107 */
+static inline void
+fd_txn_parse_v1_config( uchar          config_mask,
```
ptaffet-jump: This feels like it doesn't belong here, but again, I'll have more to say after I see how it is used.

src/ballet/txn/fd_txn_parse.c
```diff
@@ -76,8 +76,141 @@ fd_txn_parse_core( uchar const             * payload,
...
+    parsed->v1_txn_config_mask           = (uchar)config_mask;
+    parsed->v1_txn_config_values_off     = (ushort)config_values_off;
+    parsed->instr_cnt                    = (ushort)instr_cnt;
+
+    /* Check for leftover bytes if payload_sz_opt not specified. */
+    CHECK( (payload_sz_opt!=NULL) | (i==payload_sz) );
+
+    if( FD_LIKELY( counters_opt   ) ) counters_opt->success_cnt++;
+    if( FD_LIKELY( payload_sz_opt ) ) *payload_sz_opt = i;
+    return fd_txn_footprint( instr_cnt, parsed->addr_table_lookup_cnt );
```
ptaffet-jump: addr_table_lookup_cnt can just be 0

src/ballet/txn/fd_txn_parse.c
```diff
@@ -76,8 +76,141 @@ fd_txn_parse_core( uchar const             * payload,
...
+    }
+
+    /* Pass 2: instruction payloads */
+    uchar max_acct = 0;
+    for( ulong j=0UL; j<instr_cnt; j++ ) {
+      ushort acct_cnt = parsed->instr[ j ].acct_cnt;
+      ushort data_sz  = parsed->instr[ j ].data_sz;
+
+      CHECK_LEFT( acct_cnt                      );   ulong acct_off       =          i  ;
+      for( ulong end=i+acct_cnt; i<end; i++ ) max_acct = fd_uchar_max( max_acct, payload[ i ] );
```
ptaffet-jump: This kind of for loop looks very strange to me. I think it would be better to just have the declaration of `end` on the line before.

ptaffet-jump: The reason for using `max_acct` in the `v0` parser is because you don't know the number of accounts from ALTs at that point. This way might be faster, but something like `CHECK( payload[ i ]<acct_addr_cnt )` would be more in line with the rest of the code here.

src/disco/keyguard/fd_keyguard_match.c
ptaffet-jump, line 128: I think this needs more significant changes. It's supposed to return 1 for anything that could be any sort of valid transaction IIRC.  Maybe talk to @riptl about it.

src/disco/pack/fd_pack.c
```diff
@@ -2102,33 +2095,69 @@ fd_pack_schedule_impl( fd_pack_t          * pack,
...
-      FD_STATIC_ASSERT( offsetof(fd_txn_p_t, flags          )+sizeof(((fd_txn_p_t*)NULL)->flags         )<=1280UL, nt_memcpy );
-      FD_STATIC_ASSERT( offsetof(fd_txn_p_t, _              )                                            <=1280UL, nt_memcpy );
-      const ulong offset_into_txn = 1280UL - offsetof(fd_txn_p_t, _ );
-      fd_memcpy( offset_into_txn+(uchar *)TXN(out_txnp), offset_into_txn+(uchar const *)txn,
-          fd_ulong_max( offset_into_txn, fd_txn_footprint( txn->instr_cnt, txn->addr_table_lookup_cnt ) )-offset_into_txn );
+
+      /* For V1 transactions, the payload can be up to 4096 bytes so we copy an additional 2816 bytes. */
+      if( FD_UNLIKELY( txn->transaction_version==FD_TXN_V1 ) ) {
+        _mm512_stream_si512( (void*)(out_txnp->payload+1280UL), _mm512_load_epi64( cur->txn->payload+1280UL ) );
+        _mm512_stream_si512( (void*)(out_txnp->payload+1344UL), _mm512_load_epi64( cur->txn->payload+1344UL ) );
```
ptaffet-jump: The right way to do this is for me to finish the non-temporal memcpy PR that I started, but for now, just replace this part with a loop. This is too many copy-pasted lines for me.

src/disco/pack/fd_pack_cost.h
```diff
@@ -525,7 +527,36 @@ fd_pack_compute_cost( fd_txn_t const * txn,
...
+    ulong v1_priority_fee, v1_cu_limit, v1_loaded, v1_heap;
+    fd_txn_parse_v1_config( txn->v1_txn_config_mask, payload+txn->v1_txn_config_values_off,
+                            &v1_priority_fee, &v1_cu_limit, &v1_loaded, &v1_heap );
+    *fee            = v1_priority_fee;
+    *execution_cost = (uint)fd_ulong_min( v1_cu_limit, FD_COMPUTE_BUDGET_MAX_CU_LIMIT );
+    v1_loaded       = fd_ulong_min( v1_loaded, FD_COMPUTE_BUDGET_MAX_LOADED_DATA_SZ );
+
+    /* V1 transactions which request a loaded accounts data size limit
+       of 0, or do not explicitly set this limit, are thrown out by
+       Firedancer's block packer. Agave reserves 1 page for these,
```
ptaffet-jump: Is this behavior okay?

src/disco/pack/fd_pack.c
```diff
@@ -1291,13 +1291,6 @@ validate_transaction( fd_pack_t               * pack,
   if( FD_UNLIKELY( bundle_blacklist & !!check_bundle_blacklist             ) ) return FD_PACK_INSERT_REJECT_BUNDLE_BLACKLIST;
```
ptaffet-jump, line 1287: Pack depends on the signature being at offset 1 in the transaction. We need to talk about how to fix it.

src/discof/replay/fd_sched.c
```diff
@@ -2214,13 +2214,16 @@ FD_WARN_UNUSED static int
...
-  ulong txn_sz = fd_txn_parse_core( payload,
-                                    fd_ulong_min( FD_TXN_MTU, block->fec_buf_sz-block->fec_buf_soff ),
-                                    txn,
-                                    NULL,
-                                    &pay_sz );
+  uchar * payload   = block->fec_buf+block->fec_buf_soff;
+  ulong   remaining = block->fec_buf_sz-block->fec_buf_soff;
+  ulong   pay_sz    = 0UL;
+  int   is_v1       = remaining && ( payload[ 0 ] & 0x80U );
+  ulong ver_mtu     = is_v1 ? FD_TXN_MTU : FD_TXN_MTU_V0;
```
ptaffet-jump: Is there a reason we need to pass the smaller MTU if we can? Doesn't `fd_txn_parse_core` take care of it?

src/flamenco/runtime/context/fd_exec_instr_ctx.c
```diff
@@ -113,9 +113,10 @@ fd_exec_instr_ctx_get_signers( fd_exec_instr_ctx_t const * ctx,
         continue;
       }
 
-      /* This should never be possible */
-      if( FD_UNLIKELY( j>=FD_TXN_SIG_MAX ) ) {
-        FD_LOG_CRIT(( "invariant violation: too many signers (cnt=%lu, FD_TXN_SIG_MAX=%lu)", j, (ulong)FD_TXN_SIG_MAX ));
+      /* This should never be possible: distinct signers are bounded by the
+         transaction account count (FD_INSTR_SIGNERS_MAX). */
```
ptaffet-jump: This comment seems a little confusing to me. What do you mean by `transaction account count`?

src/flamenco/runtime/program/vote/fd_vote_codec.h
```diff
@@ -698,6 +698,9 @@ fd_vote_account_epoch_credits( uchar const * data,
    data (deques, arrays) is placed in memory embedded within the
    sub-structs of instruction.
 
+   Only bytes up to FD_TXN_MTU_V0 are parsed, the rest are discarded,
```
ptaffet-jump, line 701: Good find.  Did we follow up with Anza about whether this is intentional or not though?

src/disco/pack/fd_pack_cost.h
```diff
@@ -385,7 +385,9 @@ fd_pack_compute_cost( fd_txn_t const * txn,
     fd_pack_builtin_prog_cost_t const * in_tbl = fd_pack_builtin_query( prog_id, null_row );
```
ptaffet-jump, line 379: Does `FD_PACK_MAX_TXN_COST` need to be updated? If so, we also need to update the bundle constants.

src/flamenco/runtime/program/vote/fd_vote_codec.h
topointon-jump: this is almost definitely an accident. i'm following up with them at the moment

src/ballet/txn/fd_txn_parse.c
```diff
@@ -76,8 +76,141 @@ fd_txn_parse_core( uchar const             * payload,
...
+    CHECK_LEFT( 1UL                             );   uchar acct_addr_cnt  = payload[ i ];     i++;
+    CHECK( (signature_cnt<=acct_addr_cnt) & (acct_addr_cnt<=FD_TXN_ACCT_ADDR_MAX) );
+    CHECK( (ulong)signature_cnt+(ulong)ro_unsigned_cnt<=(ulong)acct_addr_cnt );
+
+    CHECK_LEFT( FD_TXN_ACCT_ADDR_SZ*acct_addr_cnt ); ulong acct_addr_off  =          i  ;     i+=FD_TXN_ACCT_ADDR_SZ*acct_addr_cnt;
+
+    /* Config values: 4 bytes per set mask bit */
+    ulong config_values_off = i;
+    ulong num_config_values = (ulong)fd_uint_popcnt( config_mask );
+    CHECK_LEFT( 4UL*num_config_values           );                                            i+=4UL*num_config_values;
```
ptaffet-jump, line 143: > - If bit 4 is set, the requested heap size MUST be a multiple of 1KiB and in
  the inclusive range [32KiB, 256KiB]. Values outside of these bounds are
  sanitization failures.

This is probably the right place to check that property in the spec

src/discof/replay/fd_sched.c
topointon-jump: `fd_txn_parse_core` enforces that the payload size is `FD_TXN_MTU` for V1 transactions, and `FD_TXN_MTU_V0` for V0/legacy transactions. this means that if we pass FD_TXN_MTU in here for _all_ transactions, V0/legacy transactions will error out

topointon-jump: this is the only place in the codebase where we have this issue, so I opted to fix it here instead of changing the api

src/disco/pack/fd_pack_cost.h
topointon-jump: good catch thanks! will check this

src/disco/pack/fd_pack.c
topointon-jump: Great catch, thank you!

src/disco/pack/fd_pack_cost.h
topointon-jump: talk to josh

src/ballet/fd_ballet_base.h
```diff
@@ -6,7 +6,7 @@
 /* FD_TPU_MTU: The maximum size of a Solana transaction in serialized
    wire-protocol form.  This does not count any network-level (e.g. UDP
    or QUIC) headers. */
-#define FD_TPU_MTU (1232UL)
+#define FD_TPU_MTU (4096UL)
```
ptaffet-jump, line 9: In a later PR, we should change fd_txn_p and fd_txn_e so that the space for resolved ALTs overlaps with the payload for v1 txns, but this seems like a painful change, and it's fine for now.

ptaffet-jump (dismissed):
PR looks good. Once CI is green, it's good to merge!

yufeng-jump (approved):

PR #10596, opened 2026-07-30
vote: fix commission rounding

0x0ece:


0x0ece:
Both are already active in mainnet, instead of fixing this edge case we should just clean them up

PR #10602, opened 2026-07-31
Fix typos in code comments

peterschrammel:


src/ballet/lthash/fd_lthash_adder.h
```diff
@@ -129,7 +129,7 @@ fd_lthash_adder_push_solana_account(
     void const *        owner
 ) {
   fd_lthash_value_t value[1];
-  /* FIXME opportunities for memcpy hax here */
+  /* FIXME opportunities for memcpy hacks here */
```
ripatel-fd, line 132: that one was deliberate :)

src/ballet/txn/fd_txn_build.h
```diff
@@ -82,7 +82,7 @@ struct fd_txn_builder {
     uchar               map_mem[ FD_TXN_B_ADDR_MAP_SZ ];
   };
 
-  /* Bump allocator for instruction datas */
+  /* Bump allocator for instruction data */
```
ripatel-fd, line 85: "datas" is more technically correct (if less so grammatically) because a transaction can contain multiple instructions

ripatel-fd (changes requested):
Thank you for the contribution. Could you revert this one "instruction datas" change?

ripatel-fd (approved):
Eh, actually, it's fine. Thanks again!

PR #10610, opened 2026-07-31
net: networking stack auto configuration

tristanx86-jump:
## Overview
Many (possibly most) validators use the default configuration provided by the default.toml for the networking subsystem. In these cases large amounts of performance are being left on the table since the defaults are designed to be as safe as possible.

Many validators enable features like zero copy on drivers/kernels that are supposed to support it but have a broken driver implementation. This creates unnecessary support burden in Discord when they report issues that are not Firedancer's fault, as well as making the operator have a less smooth experience using Firedancer.

## Info used
This PR lets Firedancer decide on behalf of the user which networking features to enable based on their system's:
- Linux kernel version
- NIC driver
- Whether the networking interface is a bond
- Whether the networking interface is a virtual interface
- How many slaves there are of the bond (if it is a bond)

## Testing
What is deemed safe was mostly decided using the results of fd-linux-version-testbed, (results cannot be attached at this time due to an infosec issue on my mac).

This is not a 100% foolproof way to test as it does not take into account every edge case e.g. each individual Linux minor release revision, however I believe the risk of these edge cases being the reason for a failure to be low (although not 0).

### DIFF
--- a/src/app/fdctl/config/default.toml
+++ b/src/app/fdctl/config/default.toml
@@ -902,24 +902,47 @@ dynamic_port_range = "8900-9000"
 # fanning in from these various network senders to transmit on the
 # queues we have available.
 [net]
+    # Firedancer can automatically configure most networking
+    # options. The auto_level determines how the network stack
+    # is optimized. Options set to an explicit value are never
+    # modified regardless of auto_level's value.
+    #
+    #  "standard"
+    #   Enables networking features which auto configuration
+    #   determines are safe and supported on this system.
+    #
+    #  "minimal" (default)
+    #   Resolves all "auto" options to their most conservative
+    #   values.  This will reduce networking performance and is
+    #   a worst case backup.
+    #
+    #   If you suspect an error is caused by Firedancer's auto
+    #   configuration, try setting this to "minimal" to see if
+    #   the issue resolves. If you need "minimal" to make
+    #   your validator work, please report your issue with "standard"
+    #   as a bug at https://github.com/firedancer-io/firedancer/issues
+    #   with the prefix "auto config: ", alongside your log file.
+    auto_level = "minimal"
+
     # The provider determines which Linux networking stack is used by
     # Firedancer.  Possible values for the option are:
     #
     #  "xdp" (default)
-    #   Use Linux express data path APIs for high performance networking.
-    #   These APIs map memory from the network device directly into
-    #   Firedancer to bypass parts of the kernel when reading and writing
-    #   data.
+    #   Use Linux express data path APIs for high performance
+    #   networking. These APIs map memory from the network device
+    #   directly into Firedancer to bypass parts of the kernel when
+    #   reading and writing data.
     #
     #  "socket"
     #   Use UDP sockets and system calls like `send(2)` for networking.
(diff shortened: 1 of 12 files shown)


src/app/fdctl/config/default.toml
```diff
@@ -981,25 +1001,35 @@ dynamic_port_range = "8900-9000"
         # to 100 Gbps per net tile.
         #
         # Only works when XDP is provided by the network driver (see
-        # the xdp_mode option above).  If the kernel/hardware does not
-        # support zero copy, `fdctl run` will fail to start up with
-        # "operation not supported".
-        xdp_zero_copy = false
+        # the xdp_mode option above).  With true chosen on kernel/hardware
+        # which does not support zero copy, `firedancer run` will fail to
+        # start up with "operation not supported".
```
ripatel-fd, line 1006: Maybe worth mentioning that xdp_zero_copy can be broken even if the driver supposedly supports it. (e.g. ice memory corruption bug)

src/app/platform/fd_config_extract.h
```diff
@@ -111,6 +111,20 @@ fdctl_cfg_get_bool( int *                 out,
...
+fdctl_cfg_get_boolau( int *                 out,
+                      ulong                 out_sz FD_PARAM_UNUSED,
+                      fd_pod_info_t const * info,
+                      char const *          path ) {
+  if( info->val_type==FD_POD_VAL_TYPE_CSTR
+   && !strcmp( info->val, "auto" ) ) {
+    *out = 2;
+    return 1;
+  }
+  return fdctl_cfg_get_bool( out, out_sz, info, path );
```
ripatel-fd, line 136: Can you make this accept `"true"` and `"false"` as strings in addition to direct bools so we have a more consistent schema?

src/app/shared/fd_config.c
```diff
@@ -530,11 +533,25 @@ fd_config_validate( fd_config_t const * config ) {
   CFG_HAS_NON_EMPTY( hugetlbfs.max_page_size );
 
   CFG_HAS_NON_ZERO( net.ingress_buffer_size );
+  if( 0!=strcmp( config->net.auto_level, "standard" )
+   && 0!=strcmp( config->net.auto_level, "minimal"  ) ) {
```
ripatel-fd: nit usually && usually goes on the end of the previous line

src/app/shared/fd_config.h
```diff
@@ -203,16 +205,16 @@ struct fd_config_net {
...
+    int  xdp_zero_copy; /* true/false or "auto" */
+    char poll_mode[ 16 ]; /* "prefbusy", "softirq" or "auto" */
 
     uint xdp_rx_queue_size;
     uint xdp_tx_queue_size;
     uint flush_timeout_micros;
     char rss_queue_mode[ 16 ]; /* "simple", "dedicated", or "auto" */
     int  listen_gre;
-    int  native_bond;
+    int  native_bond; /* true/fase or "auto" */
```
ripatel-fd:
```suggestion
    int  native_bond; /* true/false or "auto" */
```

ripatel-fd (dismissed):

ripatel-fd (approved):

PR #10620, opened 2026-08-03
util: using SIMD acceleration on hashing

0xfka:
# Abstract 
This pr introduces an AVX-512VL/DQ implementation for fd_hash, but I have several questions and need feedback for several ideas.
This code is tested on the ns/byte tests I sent as pr, at https://github.com/firedancer-io/firedancer/pull/10575.
## Performance impact:
avx512
fd_hash is often called with 32, 64, rarely with values like 36, which is all compile time known, and with rarely with variable strlen. Not a multiple of 32(1232) workload shows the affect of executing the unchanged paths as well, even if it's not practically used.
Static check shows that the fd_hash is often called with compile time known values at 32-64 :  

```cli
rg 'fd_hash\s*\([^,]+,[^,]+,\s*([^)]+)' -o -r '$1' src/ --no-filename --no-line-number --glob '!*test*' --glob '!*bench*' | sed 's/^[ \t]*//;s/[ \t]*$//' | sort | uniq -c | sort -nr
     13 sizeof(fd_pubkey_t
      7 sizeof(fd_hash_t
      6 32UL
      3 64UL
 ``` 
Not included values lower than 3 calls. 

# Benchmark env.
To get variance-reduced performance results, I used 
```grub
processor.max_cstate=1 isolcpus=managed_irq,domain,0 nohz_full=0 rcu_nocbs=0 transparent_hugepage=never mitigations=off no_timer_check
```
on AWS  m8azn.metal-12xl instance, AMD EPYC 9R05 64-Core Processor, used taskset to 0. Tests were never run concurrently because they can affect each other at shared resources, such as l3 cache. 
Also the profiling itself is warming up inside test to make it variance-reduced, which can  be checked at the pr 10575. 
# Questions  
1) Needed to use some Intrinsics directly from immintrin, for example, 
"src/util/simd/fd_avx_wv.h Likewise, there is no _mm256_mullo_epi64 pre AVX-512." 
this is not added to avx because it is surely not avx, but also not added to avx 512. So,what to do? we may add it to avx512 library or, I don't know. To be more clear;
### Where does firedancer want wrappers for 256 bit instructions which requires avx 512? 
2) This code works on zen 4 and newer, but not sure about intel part. with the e-p core thing, they are removing avx512 support from cpus that is newer than the ones which supporting avx512.
I think that code may return illegal instruction if that situation is not handled properly, but if config/machine is making sure we can execute all the instructions here if FD_HAS_AVX512, it is not a problem.
### edit: 3. is deprecated after checking calling patterns more in depth and found out it doesn't have much practical usage. Everything after avx512 acceleration at power of 32 path at fd_hash is diminishing returns for maintaining a hand-written Intrinsics code.
3) Since we are processing data at while( p<=stop32 ) line 256 bit at once, I cannot %100 leveraged avx512, and this code is somewhere between avx and avx512. If I add a function that uses the same formula, but calculates 2 hash provided with 2 different seeds, I can fully leverage it. Thinking of something like;
foo(seed_1, seed_2, buf_1, buf_2, size_1, size_2) returning hash_1,and hash_2, and wanted to get a feedback at this pr.

### DIFF
--- a/src/util/fd_hash.c
+++ b/src/util/fd_hash.c
@@ -24,7 +24,6 @@ fd_hash( ulong        seed,
   uchar const * stop = p + sz;
   /* Sharing this values between 2 conditional
   compilation blocks */
-  __m256i state_vec; (void)state_vec;
   __m256i c1_vec;       (void)c1_vec;
   __m256i c2_vec;       (void)c2_vec;
   ulong h;
@@ -47,11 +46,11 @@ fd_hash( ulong        seed,
        Provides the intention clearly to the compiler.
        Same goes for other arrays used in #IF FD_HAS_AVX512 blocks. */
       FD_ALIGNED ulong arr[4] = { w, x, y, z };
-      state_vec =  _mm256_loadu_si256(( const wv_t*)arr   );
-      input_vec = _mm256_mullo_epi64( input_vec, c2_vec   );
-      state_vec =  wl_add( state_vec, input_vec           );
-      state_vec =  wv_rol( state_vec, 31                  );
-      state_vec = _mm256_mullo_epi64( state_vec, c1_vec   );
+      wv_t state_vec =  _mm256_loadu_si256(( const wv_t*)arr   );
+      input_vec =      _mm256_mullo_epi64( input_vec, c2_vec   );
+      state_vec =       wl_add( state_vec, input_vec           );
+      state_vec =       wv_rol( state_vec, 31                  );
+      state_vec =      _mm256_mullo_epi64( state_vec, c1_vec   );
       FD_ALIGNED ulong results[4];
       _mm256_storeu_si256(( wv_t*)results, state_vec );
       w = results[0]; x = results[1]; y = results[2]; z = results[3];
@@ -66,11 +65,12 @@ fd_hash( ulong        seed,
 
     h = ROTATE_LEFT( w, 1 ) + ROTATE_LEFT( x, 7 ) + ROTATE_LEFT( y, 12 ) + ROTATE_LEFT( z, 18 );
     #if FD_HAS_AVX512
+    /* state_vec shouldn't kept between upper conditional compilation block and this one because it creates much register preassure, since it puts both w x y z and state_vec at registers at the same time. */
     FD_ALIGNED ulong arr[4] = { w, x, y, z };
-    state_vec =  _mm256_loadu_si256(( const wv_t*)arr );
-    state_vec = _mm256_mullo_epi64( state_vec, c2_vec );
-    state_vec =                 wv_rol( state_vec, 31 );
-    state_vec = _mm256_mullo_epi64( state_vec, c1_vec );
+    wv_t state_vec =  _mm256_loadu_si256(( const wv_t*)arr );
+    state_vec =      _mm256_mullo_epi64( state_vec, c2_vec );
+    state_vec =                      wv_rol( state_vec, 31 );
+    state_vec =      _mm256_mullo_epi64( state_vec, c1_vec );
     FD_ALIGNED ulong results[4];
        _mm256_storeu_si256(( wv_t*)results, state_vec );
     w = results[0]; x = results[1]; y = results[2]; z = results[3];
--- a/src/util/fd_hash.c
+++ b/src/util/fd_hash.c
@@ -118,7 +118,7 @@ fd_hash_avx512dq( ulong        seed,
     h ^= z; h = h*C1 + C4;
   }
 
-  h += ((ulong)sz);
+  h += sz;
 
   while( (p+8)<=stop ) { /* Last 1 to 3 complete ulong's */
     ulong w = FD_LOAD( ulong, p );
--- a/src/util/fd_hash.c
+++ b/src/util/fd_hash.c
@@ -66,8 +66,6 @@ fd_hash( ulong        seed,
 
     h = ROTATE_LEFT( w, 1 ) + ROTATE_LEFT( x, 7 ) + ROTATE_LEFT( y, 12 ) + ROTATE_LEFT( z, 18 );
     #if FD_HAS_AVX512
-    FD_ALIGNED ulong arr[4] = { w, x, y, z };
-    state_vec =  _mm256_loadu_si256(( const wv_t*)arr );
     state_vec = _mm256_mullo_epi64( state_vec, c2_vec );
     state_vec =                 wv_rol( state_vec, 31 );
     state_vec = _mm256_mullo_epi64( state_vec, c1_vec );
--- a/src/util/fd_hash.c
+++ b/src/util/fd_hash.c
@@ -2,6 +2,11 @@
 #include "bits/fd_bits.h"
 
 /* A cleaner implementation of xxhash-r39 (Open Source BSD licensed). */
+#if FD_HAS_AVX512
+#include "simd/fd_avx.h"
+#else
+#include <immintrin.h>
+#endif
 
 #define ROTATE_LEFT(x,r) (((x)<<(r)) | ((x)>>(64-(r))))
 #define C1 (11400714785074694791UL)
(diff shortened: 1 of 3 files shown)


0xfka:
@ripatel-fd

src/util/fd_hash.c
```diff
@@ -28,19 +37,53 @@ fd_hash( ulong        seed,
     ulong z = seed - C1;
 
     do { /* All complete blocks of 32 */
+
+      #if FD_HAS_AVX512
```
ripatel-fd: this should be gated on `__AVX512DQ__ && __AVX512VL__`

src/util/fd_hash.c
```diff
@@ -28,19 +37,53 @@ fd_hash( ulong        seed,
...
+      c1_vec =         _mm256_set1_epi64x(( long long)C1  );
+      c2_vec =         _mm256_set1_epi64x(( long long)C2  );
+      wv_t input_vec = _mm256_loadu_si256(( const wv_t*)p );
+      /* Temporary array, eliminated by compiler already.
+       Provides the intention clearly to the compiler.
+       Same goes for other arrays used in #IF FD_HAS_AVX512 blocks. */
+      FD_ALIGNED ulong arr[4] = { w, x, y, z };
```
ripatel-fd: please prefer the fd_avx.h APIs where possible. 
There should be no reason to use a local ulong array here, `wv_t state_vec = wv( w, x, y, z )` should do the same

src/util/fd_hash.c
```diff
@@ -28,19 +37,53 @@ fd_hash( ulong        seed,
...
+      wv_t state_vec =  _mm256_loadu_si256(( const wv_t*)arr   );
+      input_vec =      _mm256_mullo_epi64( input_vec, c2_vec   );
+      state_vec =       wl_add( state_vec, input_vec           );
+      state_vec =       wv_rol( state_vec, 31                  );
+      state_vec =      _mm256_mullo_epi64( state_vec, c1_vec   );
```
ripatel-fd: the vertical alignment is weird here, suggest just using single spaces

src/util/fd_hash.c
```diff
@@ -28,19 +37,53 @@ fd_hash( ulong        seed,
...
       y += FD_LOAD( ulong, p+16 )*C2; y = ROTATE_LEFT( y, 31 ); y *= C1;
       z += FD_LOAD( ulong, p+24 )*C2; z = ROTATE_LEFT( z, 31 ); z *= C1;
+      #endif
       p += 32;
     } while( p<=stop32 );
 
     h = ROTATE_LEFT( w, 1 ) + ROTATE_LEFT( x, 7 ) + ROTATE_LEFT( y, 12 ) + ROTATE_LEFT( z, 18 );
-
+    #if FD_HAS_AVX512
+    /* state_vec shouldn't kept between upper conditional compilation block and this one because it creates much register preassure, since it puts both w x y z and state_vec at registers at the same time. */
```
ripatel-fd: not worth documenting compiler register pressure here because there are large regalloc differences between gcc and clang

src/util/fd_hash.c
```diff
@@ -16,7 +22,10 @@ fd_hash( ulong        seed,
          ulong        sz ) {
   uchar const * p    = ((uchar const *)buf);
   uchar const * stop = p + sz;
-
+  /* Sharing this values between 2 conditional
+  compilation blocks */
+  __m256i c1_vec;       (void)c1_vec;
+  __m256i c2_vec;       (void)c2_vec;
```
ripatel-fd: instead of `(void)c1_vec` gate this AVX512

ripatel-fd (changes requested):
This needs a bit of code style work. Perhaps let an LLM do the formatting. 

Can you move the AVX512 core to a separate function, e.g. `fd_hash_avx512dq`?
And rename the scalar version to `fd_hash_generic`. These should be static functions so they can get inlined in the main function.

Then have a small shim as the main function:

```
fd_hash( ... ) {
  #if __AVX512DQ__ && __AVX512VL__
  return fd_hash_avx512dq( ... );
  #else
  return fd_hash_generic( ... );
  #endif
}
```

src/util/fd_hash.c
```diff
@@ -28,19 +37,53 @@ fd_hash( ulong        seed,
...
+      #if FD_HAS_AVX512
+      c1_vec =         _mm256_set1_epi64x(( long long)C1  );
+      c2_vec =         _mm256_set1_epi64x(( long long)C2  );
+      wv_t input_vec = _mm256_loadu_si256(( const wv_t*)p );
+      /* Temporary array, eliminated by compiler already.
+       Provides the intention clearly to the compiler.
+       Same goes for other arrays used in #IF FD_HAS_AVX512 blocks. */
+      FD_ALIGNED ulong arr[4] = { w, x, y, z };
+      wv_t state_vec =  _mm256_loadu_si256(( const wv_t*)arr   );
+      input_vec =      _mm256_mullo_epi64( input_vec, c2_vec   );
```
ripatel-fd: please modify `fd_avx_wv.h` to provide `wv_mul`, but gate it on `__AVX512DQ__ && __AVX512VL__`.
Then you can have the entire algo in readable FD vector style instead of Intel intrinsics

ripatel-fd:
> but not sure about intel part. with the e-p core thing, they are removing avx512 support from cpus that is newer than the ones which supporting avx512.

@0xfka You don't need to worry about this. On AVX2-era CPUs your algorithm won't run (due to lack of wide multiplies). AVX512 is fine and so is AVX10-256 (the Intel downgraded version). On AVX10-256 CPUs you still get all the AVX512 extensions (like `_mm256_mullo_epi64`), but using 256-bit registers, not 512-bit.

src/util/fd_hash.c
0xfka: It is generating much instructions, putting all 4 data to registers, and then writing.
current code generates vmovdqu
_mm256_setr_epi64x ( wv ) is already defined as "Instruction: Sequence" at intel intrinsics guide, 
so this is why I use a local ulong array.

0xfka:
@ripatel-fd thank you for all your feedback, I think I solved them.

src/util/fd_hash.c
```diff
@@ -73,6 +76,90 @@ fd_hash( ulong        seed,
...
+    x = wv_extract( state_vec, 1 );
+    y = wv_extract( state_vec, 2 );
+    z = wv_extract( state_vec, 3 );
+    h ^= w; h = h*C1 + C4;
+    h ^= x; h = h*C1 + C4;
+    h ^= y; h = h*C1 + C4;
+    h ^= z; h = h*C1 + C4;
+  }
+
+  h += ((ulong)sz);
```
ripatel-fd:
```suggestion
  h += sz;
```

unnecessary cast

ripatel-fd (changes requested):
looks great! just one more concern

src/util/simd/fd_avx_wv.h
```diff
@@ -173,6 +174,10 @@ static inline wv_t wv_ror( wv_t a, int imm ) { return wv_or( wv_shr( a, imm & 63
 static inline wv_t wv_rol_variable( wv_t a, int n ) { return wv_or( wv_shl_variable( a, n&63 ), wv_shr_variable( a, (-n)&63 ) ); }
 static inline wv_t wv_ror_variable( wv_t a, int n ) { return wv_or( wv_shr_variable( a, n&63 ), wv_shl_variable( a, (-n)&63 ) ); }
 
+#if defined(__AVX512VL__)
```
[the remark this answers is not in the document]
0xfka: cannot have avx512vl without avx512f, f stands for foundation already

ripatel-fd (dismissed):
Thank you! Nice change

ripatel-fd:
@0xfka could you fix the clang compile fail?

ripatel-fd:
`MACHINE=linux_clang_sse CC=clang make -j check`

0xfka:
@ripatel-fd this should fix it, thank you for all your reviews

ripatel-fd (approved):

PR #10652, opened 2026-08-04
[WIP] runtime: skip writing untouched accounts

kbhargava-jump:


ripatel-fd (approved):

ripatel-fd:
rebased on latest main so we get an updated bench comment

PR #10655, opened 2026-08-04
net: initial ibverbs implementation

tristanx86-jump:
Implements a new ibeth networking tile which drives mlx5 hardware queues directly, without relying on libibverbs or other networking libraries.

The ibeth tile only supports mlx5 due to the nature of the technology. However, we're seeing more and more bare metal providers move onto mlx5 NICs due to how good they are, and from the small surveys done on validators, they also seem keen to adopt mlx5 based NICs.

This initial implementation processed a maximum of 23Mpps RX (min sized packets) during my testing on a Mellanox CX6 NIC.


tristanx86-jump:
Moved to production mlx5 tile PR, separated from this early PoC

PR #10681, opened 2026-08-06
gossip: optimize gossip handling

ripatel-fd:
The GUI tile calls the full gossip packet parser for every
incoming packet even though it only looks at the first 4 bytes.


jherrera-jump (approved):
Introduces false negatives annoyingly but worth the tradeoff I guess

PR #10710, opened 2026-08-07
backup: CBMC proofs for custom HTTP parsers

ripatel-fd:


ripatel-fd:
Closing this because it adds 10 minutes to CBMC CI times

PR #10718, opened 2026-08-07
backup: serialize stake delegations

ripatel-fd:
Technically, serializing stake delegations into the snapshot
manifest is pointless.  All state can be recovered from accounts
instead.  However, this is what Agave expects, so Firedancer has
to output delegations or Agave will crash.


ripatel-fd:
On second thought, I don't want to do this

PR #10767, opened 2026-08-11
fec sets: shave memory for full client

ibhatt-jumptrading:


ibhatt-jumptrading:
duplicates @drubin-fd store changes

PR #10777, opened 2026-08-12
gui: fix slot creation before epoch schedule

two-heart:
prevents crash when starting local cluster


mmcgee-jump:
cc @jherrera-jump does not seem like a great fix? Why don't we just wait for the schedule?

jherrera-jump:
Yeah. The simplest fix here would probably to defer replay processing until after we've gotten the epoch message. We do something similar to handle the tower/replay ordering here:

https://github.com/firedancer-io/firedancer/blob/be2456fa477fce465d3aa6f31950c6b677f14301/src/disco/gui/fd_gui_tile.c#L575

two-heart:
Implemented now

jherrera-jump (approved):

PR #10782, opened 2026-08-12
accdb: reduce cacheline object footprint

ibhatt-jumptrading:


ibhatt-jumptrading:
needs to get reworked with better bit packing

PR #10784, opened 2026-08-12
toml: fix bug in floating point parsing

dfedorin-jump:
There is a pretty severe bug in floating point parsing. It's very easy to get an incorrect number out of fractional digits. Specifically, we skip underscores but count them as digits, and we miss the last digit (because `break` doesn't trigger the post-loop part of the `for` loop). So, instead of returning the number of digits we parsed, we return digits+underscores-1. The result:

| Key | Parsed As |
|-----|-----------|
| `0.250` | 2.5           |
| `0.2_5_0` | 0.025 |
| `3.14` | 4.4 |

It doesn't look like we use any floating point keys (?), so this hasn't actually surfaced.


ripatel-fd:
Thanks @dfedorin-jump do you mind if we remove floating point support from fd_toml entirely or do you use it for something?

ripatel-fd (approved):

dfedorin-jump:
I was about to use it for something. However, I could make do with integers.

PR #10787, opened 2026-08-12
runtime: fix heap overflow in instr harness

cmoyes-jump:
txn->payload is a fixed FD_TPU_MTU buffer. test_ctx->data->size is fixture-controlled and can exceed it, so bound the copy to avoid overflowing the buffer (a real transaction can never exceed the MTU anyway). solfuzz fix ups should make this not reachable but nothing stops us from replaying any arbitrary input fixture that can still hit this.

### DIFF
--- a/src/flamenco/runtime/tests/fd_instr_harness.c
+++ b/src/flamenco/runtime/tests/fd_instr_harness.c
@@ -66,8 +66,9 @@ fd_solfuzz_pb_instr_ctx_create( fd_solfuzz_runner_t *                runner,
   fd_txn_p_t * txn            = fd_spad_alloc_check( runner->spad, alignof(fd_txn_p_t), sizeof(fd_txn_p_t) );
   fd_txn_t *   txn_descriptor = TXN( txn );
   if( test_ctx->data ) {
-    memcpy( txn->payload, test_ctx->data->bytes, test_ctx->data->size );
-    txn->payload_sz = test_ctx->data->size;
+    ulong payload_sz = fd_ulong_min( test_ctx->data->size, sizeof(txn->payload) );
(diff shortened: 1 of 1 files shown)


ripatel-fd (approved):

src/flamenco/runtime/tests/fd_instr_harness.c
```diff
@@ -66,8 +66,9 @@ fd_solfuzz_pb_instr_ctx_create( fd_solfuzz_runner_t *                runner,
   fd_txn_p_t * txn            = fd_spad_alloc_check( runner->spad, alignof(fd_txn_p_t), sizeof(fd_txn_p_t) );
   fd_txn_t *   txn_descriptor = TXN( txn );
   if( test_ctx->data ) {
-    memcpy( txn->payload, test_ctx->data->bytes, test_ctx->data->size );
-    txn->payload_sz = test_ctx->data->size;
+    ulong payload_sz = fd_ulong_min( test_ctx->data->size, sizeof(txn->payload) );
+    memcpy( txn->payload, test_ctx->data->bytes, payload_sz );
+    txn->payload_sz = payload_sz;
```
mjain-jump, line 71: this should be unreachable given the instruction fuzzer constrains instruction data size to 1232. was this because of a test vector failure?

ripatel-fd: not disagreeing with this, but it's worthwhile to make the entire fuzzing interface free of undefined behavior given arbitrary protobuf blobs

cmoyes-jump: closing because @mjain-jump has this same change in another PR now

PR #10790, opened 2026-08-13
votor: implement the alpenglow voting receive loop

lidatong:


ripatel-fd (commented):
I'm not a big fan of a new `ag_` prefix for this module but I suppose it's too late to change ...

emwang-jump (approved):

PR #10791, opened 2026-08-13
pack: raise byte bound now that shreds close batches at two FECs

### LINKED ISSUE #10747: Shred space over-provisioned for padding
Currently, the available data space in a block (`max_data_bytes_per_block`, `slot_max_data`) is computed very conservatively from some limits.
Particularly, a provision of half of total space is taken for padding, that is `FD_SHREDDER_CHAINED_FEC_SET_PAYLOAD_SZ` per batch.
However, the max space lost from padding is only the max size of a microblock, that is 1 single transaction, or 5 transactions if bundles are enabled.

I can draft a PR to lower this provision accordingly if you agree with this. This should nearly double the available space.

cavemanloverboy:
pack’s leader byte budget assumed some weird worst-case ~50% FEC padding. the shred tile does not do that. it closes a chained batch when the next microblock would not fit in two FEC sets, so padding is at worst case one (worst-case) microblock (minus one byte).

pack now takes the min of that shred-fit entry-byte bound and SIMD-0525 `max_entry_bytes_per_slot`, then subtracts ticks. with one-txn microblocks the protocol cap binds (15.73 MiB at 300 ms, up from 11.75 MiB).

shoutout @KirillLykov for finding this

### DIFF
--- a/src/disco/shred/fd_shred_batch.h
+++ b/src/disco/shred/fd_shred_batch.h
@@ -101,35 +101,59 @@ FD_STATIC_ASSERT( ( FD_SHRED_BATCH_RAW_BUF_SZ ) >= ( FD_SHRED_BATCH_WMARK_NORMAL
 FD_STATIC_ASSERT( ( FD_SHRED_BATCH_RAW_BUF_SZ ) >= ( FD_SHRED_BATCH_WMARK_CHAINED  + FD_POH_SHRED_MTU ), FD_SHRED_BATCH_RAW_BUF_SZ );
 FD_STATIC_ASSERT( ( FD_SHRED_BATCH_RAW_BUF_SZ ) >= ( FD_SHRED_BATCH_WMARK_RESIGNED + FD_POH_SHRED_MTU + FD_SHRED_BATCH_RESIGNED_WMARK_REGRESSION ), FD_SHRED_BATCH_RAW_BUF_SZ );
 
-/* Each block is limited to 32k parity shreds.  Since each FEC set
-   is now guaranteed to contain 32 data shreds and 32 parity shreds,
-   the maximum number of FEC sets is FEC_SETS_MAX = 1024 (32k/32).
-   We consider the payload capacity of chained FEC sets as the raw
-   capacity (which is smaller than normal FEC sets), that means
-   FEC_SETS_MAX * FD_SHREDDER_CHAINED_FEC_SET_PAYLOAD_SZ, and then
-   we subtract from there the worst-case overheads from: padding,
-   watermark regression and batch header.
-   - OHEAD_PAD (padding overhead): except for the first and last batch
-   in a block, each batch typically has FD_SHRED_BATCH_FEC_SETS_WMARK
-   FEC sets, but can contain as little as one FEC set.  However, the
-   worst case occurs when each batch has two FEC sets, of which the
-   second one contains a single byte of data and the rest is padding.
-   In that case, OHEAD_PAD is approximately 1/2 of the maximum raw
-   capacity, i.e.
-   (FEC_SETS_MAX / 2) * FD_SHREDDER_CHAINED_FEC_SET_PAYLOAD_SZ.
-   - OHEAD_REG (watermark regression overhead): this is basically
-   FD_SHRED_BATCH_FEC_SETS_MAX * 2048 bytes (the difference in
-   payload size between chained and (chained+)resigned FEC sets).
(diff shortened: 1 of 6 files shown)


HGuillemet:
Fix #10747 ?

cavemanloverboy:
> Fix #10747 ?

ah yes

src/disco/shred/fd_shred_batch.h
```diff
@@ -101,35 +101,56 @@ FD_STATIC_ASSERT( ( FD_SHRED_BATCH_RAW_BUF_SZ ) >= ( FD_SHRED_BATCH_WMARK_NORMAL
...
+                              ulong max_microblock_sz ) {
+  ulong const C     = FD_SHREDDER_CHAINED_FEC_SET_PAYLOAD_SZ;
+  ulong const wmark = FD_SHRED_BATCH_WMARK_CHAINED; /* 2*C - 8 */
+
+  if( FD_UNLIKELY( max_microblock_sz>=wmark ) ) return 0UL;
+
+  ulong fec_set_cnt = slot_max_data_shreds / 32UL;
+  ulong first_fec   = ( 8UL + max_microblock_sz + C - 1UL ) / C;
+  if( FD_UNLIKELY( first_fec<1UL ) ) first_fec = 1UL;
+  ulong const last_fec = 3UL; /* resigned: chained-wmark pending + empty tick */
```
ptaffet-jump: I'd make this `(wmark-1UL + 8UL + max_microblock_sz)/FD_SHREDDER_RESIGNED_FEC_SET_PAYLOAD_SZ`

src/disco/shred/fd_shred_batch.h
```diff
@@ -101,35 +101,56 @@ FD_STATIC_ASSERT( ( FD_SHRED_BATCH_RAW_BUF_SZ ) >= ( FD_SHRED_BATCH_WMARK_NORMAL
...
+fd_shred_batch_pack_data_max( ulong slot_max_data_shreds,
+                              ulong max_microblock_sz ) {
+  ulong const C     = FD_SHREDDER_CHAINED_FEC_SET_PAYLOAD_SZ;
+  ulong const wmark = FD_SHRED_BATCH_WMARK_CHAINED; /* 2*C - 8 */
+
+  if( FD_UNLIKELY( max_microblock_sz>=wmark ) ) return 0UL;
+
+  ulong fec_set_cnt = slot_max_data_shreds / 32UL;
+  ulong first_fec   = ( 8UL + max_microblock_sz + C - 1UL ) / C;
+  if( FD_UNLIKELY( first_fec<1UL ) ) first_fec = 1UL;
```
ptaffet-jump: The only way `first_fec==0` is possible is if the addition overflows, but you've already ensured `max_microblock_sz<wmark`, so I'd just delete the check.

ptaffet-jump (commented):
Thanks for taking a crack at this. A few minor things to fix, but could you also get rid of the references to `SIMD-0525` `max_entry_bytes_per_slot`, since that's not a real spec thing. If you want to mention it as a side note related to Agave, that's fine.

src/disco/shred/fd_shred_batch.h
```diff
@@ -101,35 +101,59 @@ FD_STATIC_ASSERT( ( FD_SHRED_BATCH_RAW_BUF_SZ ) >= ( FD_SHRED_BATCH_WMARK_NORMAL
...
+                              ulong max_microblock_sz ) {
+  ulong const C     = FD_SHREDDER_CHAINED_FEC_SET_PAYLOAD_SZ;
+  ulong const R     = FD_SHREDDER_RESIGNED_FEC_SET_PAYLOAD_SZ;
+  ulong const wmark = FD_SHRED_BATCH_WMARK_CHAINED; /* 2*C - 8 */
+
+  if( FD_UNLIKELY( max_microblock_sz>=wmark ) ) return 0UL;
+
+  ulong fec_set_cnt = slot_max_data_shreds / 32UL;
+  ulong first_fec   = ( 8UL + max_microblock_sz + C - 1UL ) / C;
+  ulong last_fec    = ( wmark + 8UL + max_microblock_sz + R - 1UL ) / R;
```
ptaffet-jump: Sorry I didn't catch these before, but these should probably be first_batch and last_batch, right?

ptaffet-jump: Or maybe just plural `first_fecs`, `last_fecs`, `middle_fec`? `first_fec_cnt`?

ptaffet-jump (dismissed):
One nit, but looks good. Thanks!

ptaffet-jump (approved):

PR #10797, opened 2026-08-13
gossip: trim memory

ibhatt-jumptrading:
1. uint pool indices
2. lower default max entries

### DIFF
--- a/src/app/firedancer/config/default.toml
+++ b/src/app/firedancer/config/default.toml
@@ -1339,7 +1339,7 @@ telemetry = true
         # table, and this number should be set so that the table never
         # gets full, and all entries expire before they would need to be
         # evicted.
-        max_entries = 2_097_152
+        max_entries = 524_288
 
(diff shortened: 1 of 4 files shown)


src/app/firedancer/config/default.toml
```diff
@@ -1339,7 +1339,7 @@ telemetry = true
         # table, and this number should be set so that the table never
         # gets full, and all entries expire before they would need to be
         # evicted.
-        max_entries = 2_097_152
+        max_entries = 1_048_576
```
mmcgee-jump: Can probably go lower, just look at live mainnet/testnet/devnet to decide, with 30% headroom

ibhatt-jumptrading: lowered by another factor of 2, 512k gives us around 40% headroom on mainnet

mmcgee-jump (approved):

PR #10800, opened 2026-08-13
snapshot: parallel decompression

mjain-jump:


src/app/firedancer/config/default.toml
```diff
@@ -975,6 +975,10 @@ telemetry = true
     # this parallelism supported by the accounts database.
     execrp_tile_count = 10
 
+    # How many snapshot decompressor tiles to run.  These tiles do not
```
mmcgee-jump: I'd remove the idle comment? Operators don't care, and I'm not sure why they would ever be idle?

src/app/shared/commands/watch/watch.c
```diff
@@ -573,14 +573,24 @@ write_snapshots( config_t const * config,
       if( FD_UNLIKELY( incremental ) ) {
         consumed   = fd_ulong_min( cur_tile[ snapin_idx*FD_METRICS_TOTAL_SZ+MIDX( GAUGE, SNAPIN, INCREMENTAL_BYTES_READ ) ],
                                    cur_tile[ snapwr_idx*FD_METRICS_TOTAL_SZ+MIDX( GAUGE, SNAPWR, INCREMENTAL_BYTES_READ ) ] );
-        dc_in      = cur_tile[ snapdc_idx*FD_METRICS_TOTAL_SZ+MIDX( GAUGE, SNAPDC, INCREMENTAL_COMPRESSED_BYTES_READ ) ];
-        dc_out     = cur_tile[ snapdc_idx*FD_METRICS_TOTAL_SZ+MIDX( GAUGE, SNAPDC, INCREMENTAL_DECOMPRESSED_BYTES_WRITTEN ) ];
+        dc_in      = 0UL;
+        dc_out     = 0UL;
+        for( ulong i=0UL; i<snapdc_tile_cnt; i++ ) {
+          ulong snapdc_idx = fd_topo_find_tile( &config->topo, "snapdc", i );
+          dc_in  += cur_tile[ snapdc_idx*FD_METRICS_TOTAL_SZ+MIDX( GAUGE, SNAPDC, INCREMENTAL_COMPRESSED_BYTES_READ ) ];
```
mmcgee-jump, line 590: This doesn't feel right. Wouldn't you take the min(across_tiles) as the dc_in position?

src/discof/restore/utils/fd_ssctrl.h
```diff
@@ -96,6 +96,7 @@
    exceed the control structs. */
 #define FD_SNAPSHOT_DATA_DEPTH                 (256UL)
 #define FD_SNAPSHOT_DATA_MTU                   (65408UL)
+#define FD_SNAPDC_TILE_MAX                     (4UL)
```
mmcgee-jump: Why? Seems pretty arbitrary, and I could see us needing many more?

src/discof/restore/utils/fuzz_zstd_frame.c
```diff
@@ -0,0 +1,149 @@
+#include "fd_zstd_frame.h"
```
mmcgee-jump, line 1: Whats coverage of the fuzzer?

src/discof/restore/fd_snapwr_tile.c
```diff
@@ -358,21 +372,119 @@ handle_control_frag( fd_snapwr_tile_t *  ctx,
...
+    if( sig==FD_SNAPSHOT_MSG_CTRL_INIT_FULL || sig==FD_SNAPSHOT_MSG_CTRL_INIT_INCR ) {
+      ctx->full = sig==FD_SNAPSHOT_MSG_CTRL_INIT_FULL;
+    }
+
+    ctx->pending_control = sig;
+  }
+
+  /* Only process the control frag when all upstream tiles have sent
+     the same control message. */
+  ctx->control_seen[ in_idx ] = 1U;
```
mmcgee-jump, line 464: `FD_TEST( !control_seen[ in_idx ] )` ?

mmcgee-jump (changes requested):

src/discof/restore/utils/fuzz_zstd_frame.c
mjain-jump: it covers frame boundary parsing correctness between the custom parser i made and zstd itself

src/app/shared/commands/watch/watch.c
mjain-jump: dc_in represents the number of compressed bytes read, not the actual position. so it makes sense here to sum over every tile

mmcgee-jump: This whole thing doesn't make sense. You can't average compression ratio across multiple dc tiles.

mmcgee-jump (approved):

mjain-jump:
Testing on mainnet using an RPC running Agave 4.3, will merge after

ripatel-fd (approved):

PR #10803, opened 2026-08-14
events: support variable sized reporting

mmcgee-jump:


ripatel-fd (changes requested):
<s>The iov infrastructure is ugly. 

Just publish every event as a single frag and backpressure on available byte space in addition to frag count. A much cleaner solution is to upgrade stem infrastructure and tango conventions to support links with highly variable message sizes. This is useful for the snapshot loader and producer too.

Happy to unblock if we need this as a quick hack.</s>

EDIT: I misunderstodod the purpose of iov, it's not for event batcing

ripatel-fd (approved):

PR #10815, opened 2026-08-14
net: mlx5 driver

tristanx86-jump:
Implements a new mlx5 tile, this drives mlx5 hardware queues directly, without libibverbs, libmlx5 or other networking libraries, therefore the hot path has no syscalls.

The new mlx5 networking provider is supported on Linux 5.8+ due to some UAPI definitions the mlx5 tile uses but weren’t added until then.

This initial version showed a max RX pps of 28Mpps in firedancer-dev pktgen, and a max TX pps of 11Mpps. This is similar max TX Mpps to XDP Zero Copy, but about double the max RX Mpps of XDP ZC :). Testing has shown there are a good few other optimizations which can take these numbers even higher, but for now the goal is getting this version, with all the optimisation it already has, stable.


src/app/firedancer/config/default.toml
```diff
@@ -1124,6 +1128,13 @@ telemetry = true
     # excessive latency.
     ingress_buffer_size = 16384
 
+    # mlx5 networking provider configuration.  Only active if
+    # [net.provider] is set to "mlx5".
+    [net.mlx5]
+        rx_queue_size = 16384
+        tx_queue_size = 16384
+        batch_size = 64
```
ripatel-fd: Needs documentation

src/disco/net/mlx5/fd_mlx5_tile.seccomppolicy
```diff
@@ -0,0 +1,17 @@
...
+sendto: (and (eq (arg 0) qp_stats_fd)
+             (eq (arg 3) 0))
+
+recvfrom: (and (eq (arg 0) qp_stats_fd)
+               (eq (arg 3) MSG_TRUNC))
+
+write: (or (eq (arg 0) 2)
+           (eq (arg 0) logfile_fd))
+
+fsync: (eq (arg 0) logfile_fd)
```
ripatel-fd: each of these should have a description why it's necessary, see other seccomp policies

src/disco/net/mlx5/fd_mlx5_private.h
ripatel-fd, line 1: could we move this module to src/disco/net/mlx5 since this code is not really reusable outside of the tile context

src/disco/metrics/metrics.xml
```diff
@@ -459,6 +460,32 @@ EXAMPLES
...
+    <counter name="RxMalformedCnt" summary="Number of packets dropped because the Ethernet, IPv4, or UDP header was malformed." />
+    <counter name="RxRouteFailCnt" summary="Number of packets dropped because the destination UDP port could not be routed to an output link." />
+    <counter name="RxGreCnt" summary="Number of GRE packets successfully decapsulated and published." />
+    <counter name="RxGreInvalidCnt" summary="Number of GRE packets dropped because the tunnel peer or headers were invalid." />
+    <counter name="RxGreIgnoredCnt" summary="Number of GRE packets ignored because no GRE tunnel is configured." />
+    <gauge name="RxBufferBusy" summary="Number of RX buffers not currently posted to the NIC." />
+    <gauge name="RxBufferIdle" summary="Number of RX buffers currently posted to the NIC." />
+    <counter name="TxPktCnt" summary="Number of packets completed by the NIC for transmission." />
+    <counter name="TxBytesTotal" summary="Total number of bytes transmitted (including Ethernet header)." />
+    <counter name="TxCqeErrCnt" summary="Number of TX CQEs reporting an error." />
```
ripatel-fd: is this not a fatal condition?

src/disco/net/fd_net_tile_topo.c
```diff
@@ -171,28 +208,40 @@ fd_topos_net_tiles( fd_topo_t *             topo,
     for( ulong i=0UL; i<net_tile_cnt; i++ ) {
       setup_sock_tile( topo, tile_to_cpu, net_cfg );
     }
+  } else if( 0==strcmp( net_cfg->provider, "mlx5" ) ) {
```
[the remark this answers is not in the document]
tristanx86-jump: Only affects a couple developer testing cmds, but will still fix

src/disco/net/fd_net_router.h
```diff
@@ -0,0 +1,218 @@
...
+  return 1;
+}
+
+static int
+fd_net_tx_validate_ip4( fd_ip4_hdr_t const * ip4_hdr,
+                        ulong                ip4_packet_sz ) {
+  if( FD_UNLIKELY( !ip4_hdr || ip4_packet_sz<sizeof(fd_ip4_hdr_t) ) ) return 0;
+  ulong const ip4_hdr_sz = FD_IP4_GET_LEN( *ip4_hdr );
+  return FD_IP4_GET_VERSION( *ip4_hdr )==4U &&
+         ip4_hdr_sz>=sizeof(fd_ip4_hdr_t) && ip4_hdr_sz<=ip4_packet_sz;
```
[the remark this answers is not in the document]
tristanx86-jump: This check would conflict with pktgen as pktgen doesn't set net_tot_len in order for packets to not get accidentally routed.

src/disco/metrics/metrics.xml
tristanx86-jump: This counter tracked the number of CQEs with the MLX5_CQE_REQ_ERR opcode. The [Linux mlx5 driver](https://elixir.bootlin.com/linux/v7.1.8/source/drivers/infiniband/hw/mlx5/cq.c#L524) treats this as a failed work completion but not fatal. However based on [this](https://mails.dpdk.org/archives/dev/2019-May/133218.html) issue raised with DPDK's PMD, you are right that we should treat it as fatal.

src/disco/net/fd_net_tile_topo.c
```diff
@@ -13,6 +13,53 @@
...
+  tile->mlx5.rx_queue_size    = net_cfg->mlx5.mlx5_rx_queue_size;
+  tile->mlx5.tx_queue_size    = net_cfg->mlx5.mlx5_tx_queue_size;
+  tile->mlx5.batch_size       = FD_MLX5_BATCH_SIZE;
+
+  tile->net.umem_dcache_obj_id = umem_obj->id;
+  tile->mlx5.netdev_tbl_obj_id = netlink_tile->netlink.netdev_tbl_obj_id;
+  tile->mlx5.route_max         = route_max;
+  tile->mlx5.route_peer_max    = route_peer_max;
+  tile->mlx5.route_peer_seed   = 1UL + tile_kind_id;
+  tile->mlx5.neigh4_obj_id     = netlink_tile->netlink.neigh4_obj_id;
```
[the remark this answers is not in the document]
tristanx86-jump: tile->mlx5.net.umem_dcache_obj_id and tile->net.umem_dcache_obj_id are the same, so this comment is false

src/disco/metrics/metrics.xml
```diff
@@ -417,6 +417,7 @@ EXAMPLES
     <int value="1" name="RouteType" label="Unsupported route type" />
     <int value="2" name="Interface" label="Interface not available" />
     <int value="3" name="SourceIp" label="No source IP address chosen" />
+    <int value="4" name="UnsupportedInterface" label="Interface type not supported" />
```
ripatel-fd, line 420: How is this different from 2/Interface?

src/app/firedancer/commands/monitor_gossip/gossip_diag.h
```diff
@@ -43,6 +43,7 @@ typedef struct {
   ulong                       prev_net0_rx_bytes;
 
   int                         is_xdp;
+  int                         is_mlx5;
```
ripatel-fd: enum seems more appropriate here, but no big deal

src/app/firedancer/config/default.toml
```diff
@@ -1081,8 +1081,16 @@ telemetry = true
...
+    # The provider determines how Firedancer sends and receives
+    # packets.  Possible values for the option are:
+    #
+    #  "mlx5"
+    #   Use Linux uverbs to create mlx5 hardware resources, then drive
+    #   the hardware queues directly through a kernel-bypass data path.
+    #   This does not depend on libibverbs or libmlx5. Testing has shown
+    #   significantly higher packet rates than XDP. The mlx5 provider
+    #   requires Linux 5.8 or newer and a Mellanox ConnectX NIC of
+    #   version 4 or later, as these are mlx5 based.
```
ripatel-fd: put this below xdp and socket because it's the least generic

ripatel-fd: This should be a bit more operator focused, something like:

Also in general, it makes more sense to document what it does, rather than what it doesn't ("does not depend on libibverbs").

```suggestion
    #   Fast userland driver for Mellanox ConnectX NICs.
    #   Requires Linux 5.8 or newer.  Uses a true kernel bypass data
    #   path and is faster than XDP in all aspects.  Requires a Mellanox
    #   ConnectX-4 card or newer. Bonding is supported, but all cables
    #   must be attached to the same card.
```

src/app/firedancer/config/default.toml
```diff
@@ -1081,8 +1081,16 @@ telemetry = true
 # fanning in from these various network senders to transmit on the
 # queues we have available.
 [net]
-    # The provider determines which Linux networking stack is used by
-    # Firedancer.  Possible values for the option are:
+    # The provider determines how Firedancer sends and receives
+    # packets.  Possible values for the option are:
```
ripatel-fd:
```suggestion
    # network packets.  Possible values for the option are:
```

src/app/firedancer/config/default.toml
```diff
@@ -1124,6 +1128,20 @@ telemetry = true
     # excessive latency.
     ingress_buffer_size = 16384
 
+    # mlx5 networking provider configuration.  Only active if
+    # [net.provider] is set to "mlx5".
```
ripatel-fd: Follow existing precedent:


```suggestion
    # This section contains Mellanox specific network configuration.
    # Only active if [net.provider] is set to "mlx5".
```

src/app/firedancer/config/default.toml
```diff
@@ -1124,6 +1128,20 @@ telemetry = true
...
+        # mlx5 uses metadata queues shared between the NIC
+        # hardware and Firedancer userspace to relay events about
+        # incoming and outgoing packets.  This setting defines the
+        # number of entries in these metadata queues, which has to
+        # be a power of two.
+        #
+        # Smaller values use less memory.  Larger values reduce packet
+        # loss caused by excessive scheduling or inter-CPU latency.
+        mlx5_rx_queue_size = 32768
+        mlx5_tx_queue_size = 32768
```
ripatel-fd: remove the `mlx5_` prefix here this is redundant per section

src/app/firedancer/config/default.toml
```diff
@@ -1124,6 +1128,20 @@ telemetry = true
...
+    # [net.provider] is set to "mlx5".
+    [net.mlx5]
+        # mlx5 uses metadata queues shared between the NIC
+        # hardware and Firedancer userspace to relay events about
+        # incoming and outgoing packets.  This setting defines the
+        # number of entries in these metadata queues, which has to
+        # be a power of two.
+        #
+        # Smaller values use less memory.  Larger values reduce packet
+        # loss caused by excessive scheduling or inter-CPU latency.
```
ripatel-fd:
```suggestion
        # Controls the depths of receive (RX) and transmit (TX) queues. 
        # Must be a power of two.
        #
        # Smaller values use less memory.  Larger values reduce packet
        # loss caused by excessive scheduling or inter-CPU latency.
```

src/app/firedancer/config/default.toml
ripatel-fd, line 1218: (just noting this should have also dropped the "xdp_" prefix, but probably too late to fix now...)

src/app/shared/commands/configure/sysctl.c
```diff
@@ -183,7 +183,7 @@ check( config_t const * config,
     sock_params[ 0 ].value = config->net.socket.receive_buffer_size;
     sock_params[ 1 ].value = config->net.socket.send_buffer_size;
     r = check_param_list( sock_params );
-  } else {
+  } else if( strcmp( config->net.provider, "mlx5" ) ) {
```
ripatel-fd:
```suggestion
  } else if( 0==strcmp( config->net.provider, "mlx5" ) ) {
  } else {
```

src/app/shared/fd_config.c
```diff
@@ -584,12 +585,28 @@ fd_config_validate( fd_config_t const * config ) {
       FD_LOG_ERR(( "invalid `net.xdp.rss_queue_mode`: \"%s\"; must be \"simple\", \"dedicated\", or \"auto\"",
                    config->net.xdp.rss_queue_mode  ));
     }
+  } else if( 0==strcmp( config->net.provider, "mlx5" ) ) {
+    if( FD_UNLIKELY( !config->is_firedancer ) ) {
+      FD_LOG_ERR(( "invalid `net.provider`: \"mlx5\" is only supported by Firedancer" ));
+    }
```
ripatel-fd: there's no need to ban it on Frankendancer, but we should just not enable it by default

src/app/shared/fd_config.c
```diff
@@ -584,12 +585,28 @@ fd_config_validate( fd_config_t const * config ) {
...
+                   FD_MLX5_BATCH_SIZE, FD_MLX5_BATCH_SIZE ));
+    }
   } else if( 0==strcmp( config->net.provider, "socket" ) ) {
     CFG_HAS_NON_ZERO( net.socket.receive_buffer_size );
     CFG_HAS_NON_ZERO( net.socket.send_buffer_size );
   } else {
-    FD_LOG_ERR(( "invalid `net.provider`: \"%s\"; must be \"xdp\" or \"socket\"",
-                 config->net.provider ));
+    if( FD_UNLIKELY( config->is_firedancer ) ) {
+      FD_LOG_ERR(( "invalid `net.provider`: \"%s\"; must be \"mlx5\", \"xdp\" or \"socket\"",
```
ripatel-fd: put mlx5 at the end

src/app/shared/fd_config.c
```diff
@@ -584,12 +585,28 @@ fd_config_validate( fd_config_t const * config ) {
...
+      FD_LOG_ERR(( "invalid mlx5 queue depth: RX must exceed %u and TX must be at least %u",
+                   FD_MLX5_BATCH_SIZE, FD_MLX5_BATCH_SIZE ));
+    }
   } else if( 0==strcmp( config->net.provider, "socket" ) ) {
     CFG_HAS_NON_ZERO( net.socket.receive_buffer_size );
     CFG_HAS_NON_ZERO( net.socket.send_buffer_size );
   } else {
-    FD_LOG_ERR(( "invalid `net.provider`: \"%s\"; must be \"xdp\" or \"socket\"",
-                 config->net.provider ));
+    if( FD_UNLIKELY( config->is_firedancer ) ) {
```
ripatel-fd: remove frank/fire split

src/app/shared/fd_config.h
```diff
@@ -217,7 +211,7 @@ struct fd_configf {
 typedef struct fd_configf fd_configf_t;
 
 struct fd_config_net {
-  char provider[ 8 ]; /* "xdp" or "socket" */
+  char provider[ 8 ]; /* "mlx5", "xdp" or "socket" */
```
ripatel-fd: put mlx5 at the end

src/disco/metrics/metrics.xml
tristanx86-jump: XDP routing code checks if the interface type is ethernet (drops if it isn't) and whether the interface for routing to the dest ip is the same as the firedancer configured interface(drops if it isnt), but doesn't record these drops in a published metric.

The new routing code does keep track of these drops via this new metric, I do think this should stay a separate metric since 2/Interface drops mean the interface wasn't even in Firedancer's netdev table, whereas 4/UnsupportedInterface means it was in fd's netdev table but isn't correct, "IncorrectInterface" could be a better name.

src/disco/net/mlx5/fd_mlx5_tile.c
```diff
@@ -0,0 +1,1637 @@
...
+    FD_LOG_ERR(( "packet too big %lu (in_idx=%lu)", frame_sz, in_idx ));
+  }
+
+  ctx->tx_frame_valid = 1;
+  uchar const * src = fd_chunk_to_laddr_const( input_ctx->wksp_base, chunk );
+  fd_eth_hdr_t const * eth_hdr = (fd_eth_hdr_t const *)src;
+  if( FD_UNLIKELY( eth_hdr->net_type!=fd_ushort_bswap( FD_ETH_HDR_TYPE_IP ) ) ) {
+    FD_LOG_CRIT(( "in link %lu attempted to send packet with invalid ethertype %04x",
+                  in_idx, fd_ushort_bswap( eth_hdr->net_type ) ));
+  }
```
ripatel-fd: This does a fatal assertion before the overrun check. Probably fine for now, but in the future a producer could first zero a buffer, and then publish a valid Ethernet header. This could read to a torn read on an unreliable TX link where this tile seems zeros in place of net_type

src/disco/net/mlx5/fd_mlx5_tile.c
```diff
@@ -0,0 +1,1637 @@
...
+    if( FD_UNLIKELY( chunk<ctx->umem_chunk0 || chunk>ctx->umem_wmark ) ) {
+      FD_LOG_CRIT(( "RX completion chunk %lu out of bounds [%u,%u]", chunk, ctx->umem_chunk0, ctx->umem_wmark ));
+    }
+    __builtin_prefetch( fd_chunk_to_laddr_const( ctx->umem_base, chunk ), 0, 3 );
+  }
+
+  for( uint i=0U; i<(uint)comp_cnt; i++ ) {
+    ulong freed_chunk;
+    if( FD_UNLIKELY( comp[ i ].opcode==FD_MLX5_CQE_OP_RX_ERR ) ) {
+      freed_chunk = comp[ i ].chunk;
```
ripatel-fd, line 723: Do we export a counter metric anywhere for these kind of RX errors?
Every drop reason should have at least one counter, otherwise packet loss is hard to diagnose.

src/disco/net/mlx5/fd_mlx5_tile.c
```diff
@@ -0,0 +1,1637 @@
...
+    ctx->metrics.tx_invalid_cnt += (ulong)(fill_result==FD_NET_TX_FILL_INVALID);
+    return;
+  }
+
+  if( FD_UNLIKELY( ctx->tx_route.use_loopback ) ) {
+    ulong const rx_pkt_cnt  = ctx->metrics.rx_pkt_cnt;
+    ulong const freed_chunk = fd_mlx5_tile_rx_pkt( ctx, stem, chunk, frame_sz, (ulong)fd_frag_meta_ts_comp( fd_tickcount() ) );
+
+    if( ctx->metrics.rx_pkt_cnt!=rx_pkt_cnt ) {
+      ctx->sq_wqe_buf_chunk[ ctx->qp.sq_prod & (ctx->qp.tx_depth-1U) ] = (uint)freed_chunk;
```
ripatel-fd, line 971: We should not use metric counters for memory management

src/disco/net/mlx5/fd_mlx5_tile.c
```diff
@@ -0,0 +1,1637 @@
...
+    }
+    ctx->metrics.tx_pkt_cnt++;
+    ctx->metrics.tx_bytes_total += frame_sz;
+    return;
```
ripatel-fd, line 975: Are you sure this branch correctly increments error counters on failed loopback transmission

src/disco/net/mlx5/fd_mlx5_tile.c
tristanx86-jump: I'll just put it under PktRxMalformed given an oversized packet over the MTU is basically malformed

src/disco/net/mlx5/fd_mlx5_tile.c
tristanx86-jump: We do yes, just in the rx drop counters not tx counters

src/app/shared_dev/commands/pktgen/pktgen.c
```diff
@@ -161,15 +159,44 @@ get_net_stats( ulong volatile const * net_metrics[ FD_TOPO_MAX_TILES ],
...
+      stats->tx_bufs_busy = metrics[ MIDX( GAUGE, MLX5, TX_BUFFER_BUSY ) ];
+
+      rx_ok_now     = metrics[ MIDX( COUNTER, MLX5, PKT_RX       ) ];
+      rx_byte_now   = metrics[ MIDX( COUNTER, MLX5, PKT_RX_BYTES ) ];
+      rx_drop_now   = metrics[ MIDX( COUNTER, MLX5, RX_OUT_OF_BUFFER  ) ];
+      /* */ rx_drop_now += metrics[ MIDX( COUNTER, MLX5, PKT_RX_MALFORMED  ) ];
+      /* */ rx_drop_now += metrics[ MIDX( COUNTER, MLX5, PKT_RX_ROUTE_FAIL ) ];
+      tx_ok_now     = metrics[ MIDX( COUNTER, MLX5, PKT_TX_COMPLETED ) ];
+      tx_byte_now   = metrics[ MIDX( COUNTER, MLX5, PKT_TX_BYTES     ) ];
+    }
```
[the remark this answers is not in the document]
tristanx86-jump: true but not relevant to mlx5 tile PR and not caused by it

src/app/firedancer/config/default.toml
```diff
@@ -1097,9 +1097,11 @@ telemetry = true
...
     #   features that are not yet implemented in the XDP stack.
     #
-    # Using the XDP networking stack is strongly preferred where
-    # possible, as it is faster and better tested by the development
-    # team.
+    #  "mlx5"
+    #   Very fast userspace networking for Mellanox ConnectX-4 and newer
+    #   NICs.  Requires Linux 5.8 or newer and uses a kernel-bypass
+    #   packet data path.  Bonding is supported when every bond member
+    #   is a port on the same NIC.
```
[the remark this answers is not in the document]
tristanx86-jump: false

ripatel-fd (approved):

PR #10816, opened 2026-08-14
frankendancer, pohh: correctly handle parameter changes

topointon-jump:
https://github.com/firedancer-io/auditor-internal/issues/590


topointon-jump:
Closing as we decided not to fix this

PR #10825, opened 2026-08-15
shred: make FEC signature keys unaligned-safe

esemeniuc:
Represents FEC signature keys as alignment-1 byte storage and hashes the first word with `FD_LOAD`.

### Production change

Previously, `wrapped_sig_t` was a union:

```c
typedef union {
  fd_ed25519_sig_t u;  /* 64 signature bytes */
  ulong            l;  /* first 8 bytes for hashing */
} wrapped_sig_t;
```

Including `ulong` gave the union an eight-byte alignment requirement. The resolver then cast `shred->signature` to `wrapped_sig_t const *`, even though `fd_shred_t` is packed and its address is not guaranteed to be eight-byte aligned. Accessing `key->l` could therefore trigger undefined behavior.

The PR changes it to byte-aligned storage:

```c
typedef struct {
  fd_ed25519_sig_t u;
} wrapped_sig_t;

FD_STATIC_ASSERT( alignof(wrapped_sig_t)==1UL, wrapped_sig_t_align );
```

The static assertion ensures that a future field cannot accidentally restore a stronger alignment requirement.

The hash changes from:

```c
fd_ulong_hash( key->l ^ seed )
```

to:

```c
fd_ulong_hash( FD_LOAD( ulong, key->u ) ^ seed )
```

### DIFF
--- a/src/disco/shred/fd_fec_resolver.c
+++ b/src/disco/shred/fd_fec_resolver.c
@@ -5,10 +5,10 @@
 #include "../metrics/fd_metrics.h"
 #include "fd_fec_resolver.h"
 
-typedef union {
+typedef struct {
   fd_ed25519_sig_t u;
-  ulong            l;
 } wrapped_sig_t;
+FD_STATIC_ASSERT( alignof(wrapped_sig_t)==1UL, wrapped_sig_t_align );
 
 typedef struct __attribute__((packed)) {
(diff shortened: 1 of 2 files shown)


ptaffet-jump:
This is fixed as a side effect of https://github.com/firedancer-io/firedancer/pull/10607.

PR #10826, opened 2026-08-15
pack: avoid invalid estimator arithmetic

esemeniuc:
Skips variance division until the weighted denominator is positive and squares observations after conversion to `double`.

A single observation previously evaluated `0.0/0.0`, while large `uint` observations were squared with 32-bit wraparound before conversion.

Test: `test_est_tbl` with halt-on-error UBSan, including single-sample and `UINT_MAX` coverage.


ptaffet-jump (approved):

ptaffet-jump:
I haven't thought about this code in ages!

PR #10827, opened 2026-08-15
util: validate clock inputs before arithmetic

esemeniuc:
Short-circuits clock range validation before subtraction and validates `init_w` before taking its reciprocal.

Invalid inputs previously reached signed overflow or floating division by zero before being rejected.

Test: `test_clock` with halt-on-error UBSan, including the `LONG_MAX` boundary.

### DIFF
--- a/src/util/clock/fd_clock.c
+++ b/src/util/clock/fd_clock.c
@@ -165,7 +165,7 @@ fd_clock_new( void * shmem,
     return NULL;
   }
 
-  if( FD_UNLIKELY( !((1L<=recal_jit) & (recal_jit<=recal_avg) & (recal_avg<=(LONG_MAX-recal_jit))) ) ) {
+  if( FD_UNLIKELY( !((1L<=recal_jit) && (recal_jit<=recal_avg) && (recal_avg<=(LONG_MAX-recal_jit))) ) ) {
     FD_LOG_WARNING(( "bad recal_avg / recal_jit" ));
     return NULL;
   }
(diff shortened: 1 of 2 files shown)


src/util/clock/fd_clock.c
```diff
@@ -180,8 +180,12 @@ fd_clock_new( void * shmem,
     return NULL;
   }
 
+  if( FD_UNLIKELY( !(init_w>0.) ) ) {
+    FD_LOG_WARNING(( "bad w" ));
+    return NULL;
+  }
   double init_m = 1./init_w;
-  if( FD_UNLIKELY( !((init_w>0.) & (init_m>0.)) ) ) {
+  if( FD_UNLIKELY( !(init_m>0.) ) ) {
```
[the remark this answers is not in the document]
esemeniuc: updated

src/util/clock/fd_clock.c
```diff
@@ -165,7 +165,7 @@ fd_clock_new( void * shmem,
     return NULL;
   }
 
-  if( FD_UNLIKELY( !((1L<=recal_jit) & (recal_jit<=recal_avg) & (recal_avg<=(LONG_MAX-recal_jit))) ) ) {
+  if( FD_UNLIKELY( !((1L<=recal_jit) && (recal_jit<=recal_avg) && (recal_avg<=(LONG_MAX-recal_jit))) ) ) {
```
ptaffet-jump, line 168: We deliberately prefer to use `&` instead of `&&`, unless the second operand is really expensive to evaluate or unsafe to evaluate. It tends to produce fewer branches, and the compiler has more flexibility to optimize it.

esemeniuc: clanker suggests the we shouldnt use `&`

```
but the original subtraction cannot be restored under &: all operands would evaluate, and LONG_MAX-recal_jit can overflow for invalid negative inputs
```

esemeniuc: it suggests the following, seems more complex overall
```
    168 -  if( FD_UNLIKELY( !((1L<=recal_jit) && (recal_jit<=recal_avg) && (recal_avg<=(LONG_MAX-
         recal_jit))) ) ) {
    168 +  if( FD_UNLIKELY( !((1L<=recal_jit) & (recal_jit<=recal_avg) & (((ulong)recal_avg+(ulon
         g)recal_jit)<=(ulong)LONG_MAX)) ) ) {
```

ripatel-fd (approved):

src/util/clock/fd_clock.c
ripatel-fd: it's a new() function it doesn't matter @ptaffet-jump

PR #10828, opened 2026-08-15
util: avoid ctz zero in uint128 LSB helper

esemeniuc:
Selects the caller-provided default before evaluating the 64-bit least-significant-bit primitive, while preserving branchless code generation on the tested compiler/target combinations.

The zero case previously reached `__builtin_ctzl(0)` on BMI targets before selecting the caller-provided default. The conditional expression now evaluates `fd_ulong_find_lsb` only when the selected half is nonzero, and preserves arbitrary `int` defaults without signed-overflow arithmetic.

In optimized code-generation checks, Clang 22 and GCC 12/16 on generic x86-64 and Icelake emitted `bsf`/`tzcnt` plus conditional moves, with no conditional jump.

Tests:
- `test_bits` with halt-on-error UBSan on Icelake and generic x86-64 builds
- `-O3` assembly inspection with Clang 22 and GCC 12/16 on generic x86-64 and Icelake

### DIFF
--- a/src/util/bits/test_bits.c
+++ b/src/util/bits/test_bits.c
@@ -675,7 +675,11 @@ main( int     argc,
     }
     FD_TEST( fd_uint128_popcnt  ( zeros )==0                ); FD_TEST( fd_uint128_popcnt  ( ones  )==w                     );
     FD_TEST( fd_uint128_find_lsb( ones  )==0                ); FD_TEST( fd_uint128_find_msb( ones  )==(w-1)                 );
-    FD_TEST( fd_uint128_find_lsb_w_default( zeros, -1 )==-1 ); FD_TEST( fd_uint128_find_lsb_w_default( ones, -1 )==0        );
+    FD_TEST( fd_uint128_find_lsb_w_default( zeros,      -1 )==     -1 ); FD_TEST( fd_uint128_find_lsb_w_default( ones,  -1 )== 0       );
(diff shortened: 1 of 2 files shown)


src/util/bits/test_bits.c
```diff
@@ -675,7 +675,11 @@ main( int     argc,
     }
     FD_TEST( fd_uint128_popcnt  ( zeros )==0                ); FD_TEST( fd_uint128_popcnt  ( ones  )==w                     );
     FD_TEST( fd_uint128_find_lsb( ones  )==0                ); FD_TEST( fd_uint128_find_msb( ones  )==(w-1)                 );
-    FD_TEST( fd_uint128_find_lsb_w_default( zeros, -1 )==-1 ); FD_TEST( fd_uint128_find_lsb_w_default( ones, -1 )==0        );
+    FD_TEST( fd_uint128_find_lsb_w_default( zeros,  -1 )==-1 ); FD_TEST( fd_uint128_find_lsb_w_default( ones, -1 )==0        );
+    FD_TEST( fd_uint128_find_lsb_w_default( zeros,   0 )== 0 ); FD_TEST( fd_uint128_find_lsb_w_default( zeros, 63 )==63       );
```
ptaffet-jump: Could you fix the alignment here please?

esemeniuc: updated

ptaffet-jump (approved):

PR #10829, opened 2026-08-15
util: make zero-size memcpy null-safe

esemeniuc:
Returns before the generic/libc and MSan `fd_memcpy` backends when the size is zero, and routes the nullable CPI instruction-data copy through the wrapper. The x86 `rep movsb` fast path remains branch-free because a zero count performs no access.

`fd_memcpy` now explicitly guarantees that when `sz` is zero, the source and destination may be null, no memory is accessed, and the destination is returned. This makes the null/zero contract consistent across generic, icelake, MSan, ASan, and UBSan configurations.

The CPI semantics fixture now emits address/length zero instruction-data descriptors for both C and Rust ABIs, so the runtime path is covered under halt-on-error UBSan.

Tests:
- `test_util` under icelake UBSan, generic UBSan, and generic ASan+UBSan
- `test_vm_syscall_cpi` under halt-on-error UBSan
- `test_cpi_semantics` under Clang 22 generic x86-64 UBSan with `halt_on_error=1`

### DIFF
--- a/src/flamenco/vm/syscall/test_cpi_semantics.c
+++ b/src/flamenco/vm/syscall/test_cpi_semantics.c
@@ -103,6 +103,8 @@ struct cpi_test_cfg {
   uchar             cpi_meta_writable[ MAX_CFG_ACCTS ];
   uchar             cpi_meta_signer  [ MAX_CFG_ACCTS ];
 
+  int zero_instruction_data;
+
   /* Optional hook: mutate input_buf after env_build but before the CPI
      syscall.  NULL = no mutation. */
   pre_cpi_hook_t    pre_cpi_hook;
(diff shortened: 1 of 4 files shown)


src/flamenco/vm/syscall/test_cpi_semantics.c
```diff
@@ -103,6 +103,8 @@ struct cpi_test_cfg {
   uchar             cpi_meta_writable[ MAX_CFG_ACCTS ];
   uchar             cpi_meta_signer  [ MAX_CFG_ACCTS ];
 
+  _Bool zero_instruction_data;
```
ptaffet-jump: We don't really use stdbool. This should be an int.

esemeniuc: updated

ripatel-fd:
I would rather fix up the call sites than have an fd_memcpy that behaves slightly differently from memcpy

PR #10830, opened 2026-08-15
discof: initialize resolv tile test context

esemeniuc:
Fix `test_resolv_tile` after the resolver footprint and startup-gate changes.

`scratch_footprint` now reads `tile->resolv.max_live_slots`, so the standalone test must provide a minimally configured tile rather than passing `NULL`. The configured live-slot limit is shared with the SVM mini limits so the allocated footprint matches the test environment.

The resolver startup gate also begins closed. The standalone fixture marks it ready so the fragment-processing paths under test do not stop at the startup gate.

Tests:
- GCC `test_resolv_tile`
- Clang 22 UBSan `test_resolv_tile` with `halt_on_error=1`

### DIFF
--- a/src/discof/resolv/test_resolv_tile.c
+++ b/src/discof/resolv/test_resolv_tile.c
@@ -4,6 +4,7 @@
 #include "../../util/tmpl/fd_unit_test.c"
 
 #define TOPO_TAG 2UL
+#define TEST_MAX_LIVE_SLOTS 32UL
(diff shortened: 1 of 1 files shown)


mmcgee-jump (dismissed):

esemeniuc:
merge conflict dismissed review

mmcgee-jump (approved):

ibhatt-jumptrading (approved):

ptaffet-jump (approved):

PR #10833, opened 2026-08-15
util: make SIMD tests UBSan-clean

esemeniuc:
Make the SIMD unit-test oracles clean under Firedancer's explicit UBSan configuration without changing production SIMD helpers.

Firedancer compiles with `-fwrapv`, so signed addition, subtraction, and multiplication wrap rather than being language-level undefined behavior. Clang nevertheless documents that an explicitly enabled `-fsanitize=signed-integer-overflow` check still diagnoses these operations under `-fwrapv`. The tests use full-range random lanes and intentionally exercise modulo-2^N SIMD arithmetic, so their scalar reference expressions are now calculated through unsigned arithmetic and compared as unsigned lanes. This preserves the exact wrapping result while keeping the explicit sanitizer useful for unintended signed overflow elsewhere.

Separately, converting negative floating-point values to unsigned integer types is outside the representable input range and is not affected by `-fwrapv`. As the existing float-to-`uint` tests already do, the float-to-`ulong` cases now use nonnegative inputs.

Clang behavior: https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#available-checks

Tests:
- All ten SSE, AVX, and AVX-512 unit tests with Clang 22 UBSan and `halt_on_error=1`

### DIFF
--- a/src/util/simd/test_avx512_16x32.c
+++ b/src/util/simd/test_avx512_16x32.c
@@ -100,12 +100,18 @@ main( int     argc,
                             fd_int_max(x4,y4), fd_int_max(x5,y5), fd_int_max(x6,y6), fd_int_max(x7,y7),
                             fd_int_max(x8,y8), fd_int_max(x9,y9), fd_int_max(xa,ya), fd_int_max(xb,yb),
                             fd_int_max(xc,yc), fd_int_max(xd,yd), fd_int_max(xe,ye), fd_int_max(xf,yf) );
-    WWI_TEST( wwi_add(x,y), x0+y0, x1+y1, x2+y2, x3+y3, x4+y4, x5+y5, x6+y6, x7+y7,
-                            x8+y8, x9+y9, xa+ya, xb+yb, xc+yc, xd+yd, xe+ye, xf+yf );
-    WWI_TEST( wwi_sub(x,y), x0-y0, x1-y1, x2-y2, x3-y3, x4-y4, x5-y5, x6-y6, x7-y7,
-                            x8-y8, x9-y9, xa-ya, xb-yb, xc-yc, xd-yd, xe-ye, xf-yf );
-    WWI_TEST( wwi_mul(x,y), x0*y0, x1*y1, x2*y2, x3*y3, x4*y4, x5*y5, x6*y6, x7*y7,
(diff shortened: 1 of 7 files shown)


src/util/simd/test_avx512_16x32.c
```diff
@@ -100,12 +100,18 @@ main( int     argc,
...
                             fd_int_max(xc,yc), fd_int_max(xd,yd), fd_int_max(xe,ye), fd_int_max(xf,yf) );
-    WWI_TEST( wwi_add(x,y), x0+y0, x1+y1, x2+y2, x3+y3, x4+y4, x5+y5, x6+y6, x7+y7,
-                            x8+y8, x9+y9, xa+ya, xb+yb, xc+yc, xd+yd, xe+ye, xf+yf );
-    WWI_TEST( wwi_sub(x,y), x0-y0, x1-y1, x2-y2, x3-y3, x4-y4, x5-y5, x6-y6, x7-y7,
-                            x8-y8, x9-y9, xa-ya, xb-yb, xc-yc, xd-yd, xe-ye, xf-yf );
-    WWI_TEST( wwi_mul(x,y), x0*y0, x1*y1, x2*y2, x3*y3, x4*y4, x5*y5, x6*y6, x7*y7,
-                            x8*y8, x9*y9, xa*ya, xb*yb, xc*yc, xd*yd, xe*ye, xf*yf );
+    WWU_TEST( wwi_to_wwu( wwi_add(x,y) ), (uint)x0+(uint)y0, (uint)x1+(uint)y1, (uint)x2+(uint)y2, (uint)x3+(uint)y3,
+                                             (uint)x4+(uint)y4, (uint)x5+(uint)y5, (uint)x6+(uint)y6, (uint)x7+(uint)y7,
+                                             (uint)x8+(uint)y8, (uint)x9+(uint)y9, (uint)xa+(uint)ya, (uint)xb+(uint)yb,
```
ptaffet-jump: The alignment here is not right. Perhaps your tab settings aren't FD-standard?

esemeniuc: should be fixed now

ripatel-fd (approved):

PR #10834, opened 2026-08-15
ci: separate UBSan coverage

esemeniuc:
Split the combined sanitizer job into independent ASan and Clang 21 UBSan jobs. UBSan now halts on the first report, and a pointer-overflow probe verifies that `-fwrapv` has not disabled coverage.

Local checks:
- Workflow YAML parsed successfully
- Clang 21 probe exits nonzero with the expected pointer-overflow report


esemeniuc:
think temp dir is borked for clang

image

esemeniuc:
just tried rebasing thru github

ripatel-fd (approved):

ripatel-fd (dismissed):

PR #10849, opened 2026-08-17
tower: return frags on replay_out until epoch schedule is loaded

jvarela-jump:
`fd_multi_epoch_leaders_get_lsched_for_slot` returns NULL when the slot's epoch is not yet tracked by mleaders. On boot, replay publishes the epoch schedule frag (replay_epoch link) first, but the consumer may see it after the slot frag (replay_out link) because they travel on different links with no cross-link ordering guarantee. Temporary backpressure lets the stem process the replay_epoch frag first, guaranteeing lsched is non-NULL by the time replay_slot_completed runs.

### DIFF
--- a/src/discof/tower/fd_tower_tile.c
+++ b/src/discof/tower/fd_tower_tile.c
@@ -1873,6 +1873,11 @@ returnable_frag( fd_tower_tile_t *   ctx,
     case REPLAY_SIG_SLOT_COMPLETED:;
       if( FD_UNLIKELY( ctx->halt_signing ) ) return 1; /* backpressure replay_slot_completed during halt_signing. */
       fd_replay_slot_completed_t * slot_completed = (fd_replay_slot_completed_t *)fd_type_pun( fd_chunk_to_laddr( ctx->in[ in_idx ].mem, chunk ) );
+      /* Return frags until the epoch schedule is loaded.  The epoch
+         schedule frag (replay_epoch link) can arrive after this
+         slot frag (replay_out link) because they travel on different
+         links with no cross-link ordering guarantee. */
+      if( FD_UNLIKELY( !fd_multi_epoch_leaders_get_lsched_for_slot( ctx->mleaders, slot_completed->slot ) ) ) return 1;
       replay_slot_completed( ctx, slot_completed, tsorig, stem );
       break;
     case REPLAY_SIG_SLOT_DEAD:;


src/discof/tower/fd_tower_tile.c
```diff
@@ -1459,8 +1459,14 @@ replay_slot_completed( fd_tower_tile_t *            ctx,
   /* Insert into hard fork detector. */
 
   fd_epoch_leaders_t const * lsched = fd_multi_epoch_leaders_get_lsched_for_slot( ctx->mleaders, slot_completed->slot );
-  int hfork_flag = fd_hfork_record_our_bank_hash( ctx->hfork, &slot_completed->block_id, &slot_completed->bank_hash, fd_ulong_if( lsched->epoch==ctx->root_epoch, ctx->root_epoch_total_stake, ctx->next_epoch_total_stake ) );
-  update_metrics_hfork( ctx, hfork_flag, slot_completed->slot, &slot_completed->block_id );
+  /* lsched may be NULL on boot: the epoch schedule frag (replay_epoch
```
mmcgee-jump: You can just return the replay_out frag from returnable_frag until you get the epoch schedule?

mmcgee-jump (changes requested):
.

src/discof/tower/fd_tower_tile.c
```diff
@@ -1873,6 +1873,11 @@ returnable_frag( fd_tower_tile_t *   ctx,
     case REPLAY_SIG_SLOT_COMPLETED:;
       if( FD_UNLIKELY( ctx->halt_signing ) ) return 1; /* backpressure replay_slot_completed during halt_signing. */
       fd_replay_slot_completed_t * slot_completed = (fd_replay_slot_completed_t *)fd_type_pun( fd_chunk_to_laddr( ctx->in[ in_idx ].mem, chunk ) );
+      /* Return frags until the epoch schedule is loaded.  On boot, the
+         epoch schedule frag (replay_epoch link) can arrive after this
+         slot frag (replay_out link) because they travel on different
+         links with no cross-link ordering guarantee. */
+      if( FD_UNLIKELY( !fd_multi_epoch_leaders_get_lsched_for_slot( ctx->mleaders, slot_completed->slot ) ) ) return 1;
```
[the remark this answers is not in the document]
jvarela-jump: This is just a transient ordering issue (see replay tile), not an unbounded backpressure point.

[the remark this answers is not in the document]
jvarela-jump: After boot, `sched` can't be NULL for a slot_completed slot.
The `permanent backpressure` scenario can't happen (see replay tile).
Adding `ctx->init` gate.

src/discof/tower/fd_tower_tile.c
jvarela-jump: Ok - I have implemented the changes. Could you recheck?

src/discof/tower/fd_tower_tile.c
```diff
@@ -1873,6 +1873,11 @@ returnable_frag( fd_tower_tile_t *   ctx,
     case REPLAY_SIG_SLOT_COMPLETED:;
       if( FD_UNLIKELY( ctx->halt_signing ) ) return 1; /* backpressure replay_slot_completed during halt_signing. */
       fd_replay_slot_completed_t * slot_completed = (fd_replay_slot_completed_t *)fd_type_pun( fd_chunk_to_laddr( ctx->in[ in_idx ].mem, chunk ) );
+      /* On boot, return frags until the epoch schedule is loaded.  The
+         epoch schedule frag (replay_epoch link) can arrive after this
+         slot frag (replay_out link) because they travel on different
+         links with no cross-link ordering guarantee. */
+      if( FD_UNLIKELY( !ctx->init && !fd_multi_epoch_leaders_get_lsched_for_slot( ctx->mleaders, slot_completed->slot ) ) ) return 1;
```
lidatong: do you mean ||?

lidatong: ctx->init is set by the first replay_slot_completed so `!ctx->init` should probably jsut be removed

jvarela-jump: `&&` was intentional. Removing the `ctx->init` gate then.

src/discof/tower/fd_tower_tile.c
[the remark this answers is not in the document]
jvarela-jump: Removing `ctx->init` gate following later feedback.

lidatong (approved):

src/discof/tower/fd_tower_tile.c
```diff
@@ -1873,6 +1873,11 @@ returnable_frag( fd_tower_tile_t *   ctx,
     case REPLAY_SIG_SLOT_COMPLETED:;
       if( FD_UNLIKELY( ctx->halt_signing ) ) return 1; /* backpressure replay_slot_completed during halt_signing. */
       fd_replay_slot_completed_t * slot_completed = (fd_replay_slot_completed_t *)fd_type_pun( fd_chunk_to_laddr( ctx->in[ in_idx ].mem, chunk ) );
+      /* Return frags until the epoch schedule is loaded.  The epoch
+         schedule frag (replay_epoch link) can arrive after this
+         slot frag (replay_out link) because they travel on different
+         links with no cross-link ordering guarantee. */
+      if( FD_UNLIKELY( !fd_multi_epoch_leaders_get_lsched_for_slot( ctx->mleaders, slot_completed->slot ) ) ) return 1;
```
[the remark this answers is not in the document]
jvarela-jump: This does not seem right. After the epoch schedule is loaded, `get_lsched_for_slot` cannot return NULL for a `slot_completed` slot. The replay tile only produces `slot_completed` for slots in the current or next epoch, and `mleaders` always tracks both.

mmcgee-jump (approved):

PR #10854, opened 2026-08-17
configure: improve error message

jherrera-jump:


src/app/shared/commands/configure/hugetlbfs.c
```diff
@@ -291,8 +291,15 @@ fini( config_t const * config,
   }
 
   FD_LOG_NOTICE(( "%sRUN: `rmdir %s`%s", fd_log_style_dim(), config->hugetlbfs.mount_path , fd_log_style_normal() ));
-  if( FD_UNLIKELY( rmdir( config->hugetlbfs.mount_path ) && errno!=ENOENT ) )
-    FD_LOG_ERR(( "error removing hugetlbfs directory at `%s` (%i-%s)", config->hugetlbfs.mount_path, errno, fd_io_strerror( errno ) ));
+  if( FD_UNLIKELY( rmdir( config->hugetlbfs.mount_path )==-1 && errno!=ENOENT ) ) {
```
mmcgee-jump, line 294: Hmm, I haven't seen this but I can see why it would happen. We can probably just remove the lock artifacts here?

jherrera-jump: If you run a second validator with a different configured `name`. Classic example is running something like running bootstrap validator or bench command with default config (name = `fd1`) and then launching another validator later (e.g. name = 'firedancer'). Operators may see this if they ever decide to change `name` on a whim.

PR #10866, opened 2026-08-18
frankendancer, discoh: add txn v1 support to bank abi

topointon-jump:


src/discoh/bank/fd_bank_abi.c
```diff
@@ -671,15 +811,88 @@ fd_bank_abi_txn_init( fd_bank_abi_txn_t * out_txn,
...
+                           rejects these transactions before they make it to the bank.
+
+                           !fd_pack_unwritable_contains( (const fd_acct_addr_t*)(payload + txn->acct_addr_off + i*32UL) ) */
+                        (!is_key_called_as_program( txn, i ) || _is_upgradeable_loader_present);
+      v1->is_writable_account_cache[ i ] = !!is_writable;
+    }
+    out_sidecar += txn->acct_addr_cnt;
+    out_sidecar = (void*)fd_ulong_align_up( (ulong)out_sidecar, 8UL );
+
+    message->priority_fee.discr                    = (ulong)( v1_config_mask     &1U);
```
ptaffet-jump: I found this reparsing confusing. Could you add a comment? Something like:
`The compute budget information is stored twice in the struct, once in RuntimeTransaction<SanitizedTransaction>::meta::versioned_transaction_config::V1::* which has the default values applied, and one in 
RuntimeTransaction<SanitizedTransaction>::transaction::message::V1::message::config::* that stores options (None if the option is not specified in the transaction). 
To simplify the code, even when the option is not specified, we set the default value, and then use the tag field to note that it is a none value.

src/discoh/bank/fd_bank_abi.c
```diff
@@ -671,15 +811,88 @@ fd_bank_abi_txn_init( fd_bank_abi_txn_t * out_txn,
...
+    message->priority_fee._0                       = v1_priority_fee;
```
ptaffet-jump: I'm not positive the lines will be long enough, but this will probably be prettier in two columns, one for the discriminators, and one for the values.

src/discoh/bank/fd_bank_abi.h
ptaffet-jump, line 28: Can you add a comment that V1 transactions obviously don't have (3) and (4)?

src/discoh/bank/fd_bank_abi.c
```diff
@@ -671,15 +811,99 @@ fd_bank_abi_txn_init( fd_bank_abi_txn_t * out_txn,
...
+    /* The compute budget information is stored twice in the struct,
+       once in
+       RuntimeTransaction::meta::versioned_transaction_config::V1::*
+       which has the default values applied, and once in
+       RuntimeTransaction::transaction::message::V1::message::config::*
+       that stores options (None if the option is not specified in the
+      transaction).
+      To simplify the code, even when the option is not specified, we
+      set the default value, and then use the tag field to note that it
+      is a none value. */
```
ptaffet-jump: nit: the spacing here got messed up

ptaffet-jump (dismissed):

src/discoh/bank/fd_bank_abi.c
```diff
@@ -671,15 +811,99 @@ fd_bank_abi_txn_init( fd_bank_abi_txn_t * out_txn,
...
+    message->priority_fee.discr                    = (ulong)( v1_config_mask    &1U);
```
ptaffet-jump, line 870: And while you're at it, why not add a comment: 

```this sets discr to 0 or 1. That means `v1->message.discr != ABI_V1_COW_BORROWED_DISCR`, so we know it is owned.```

ptaffet-jump (approved):

PR #10868, opened 2026-08-18
gossip: stronger hash functions in maps

ibhatt-jumptrading:


ptaffet-jump (approved):
It would probably be better if the gossvf ones didn't all use the same seed, but I can't think of any actual issues with the way they are seeded now.

PR #10894, opened 2026-08-19
discof, flamenco: add support for no-op transactions (relax_fee_payer_constraint)

topointon-jump:
`relax_fee_payer_constraint` (https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0290-relax-fee-payer-constraint.md) introduces the concept of a "no-op" transaction.

These transactions:
- Have no effect on the account state
- Charge no fees
- Consume the requested CUs and the loaded accounts data size limit
- Are not packed by Firedancer or Agave by default

As of `relax_fee_payer_constraint`, transactions with invalid fee payers are allowed inside a block, but are treated as no-op transactions. The exception is durable nonce transactions with invalid fee payers, which are still not allowed in the block.

This is to facilitate bankless leader/async execution: allowing the block packer to produce blocks without needing to keep up to the head of the chain.

As these charge no fees, there is no incentive for the block producer to include them.

Tested with ledgers and local clusters.

### DIFF
--- a/src/discof/replay/fd_replay_tile.h
+++ b/src/discof/replay/fd_replay_tile.h
@@ -190,6 +190,7 @@ struct fd_replay_txn_executed {
   fd_txn_p_t txn[ 1 ];
   int is_committable;
   int is_fees_only;
+  int is_noop;
   int txn_err;
   int is_simple_vote;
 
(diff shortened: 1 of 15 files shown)


src/discof/replay/fd_replay_tile.h
```diff
@@ -184,6 +184,7 @@ struct fd_replay_txn_executed {
   fd_txn_p_t txn[ 1 ];
   int is_committable;
   int is_fees_only;
+  int is_noop;
```
topointon-jump, line 193: we technically don't need this, but I thought it was good for completeness/to avoid confusion (to distinguish between a normal txn and a no-op txn, for debugging etc). it's also good to force people to think about no-op transactions if they consume this struct. happy to remove it though

ibhatt-jumptrading (commented):
how are these going to be treated with bundles? if they invalidate a bundle, we probably need to add handling for that. or would these get filtered out upstream?

ibhatt-jumptrading (approved):

PR #10906, opened 2026-08-20
snap: reduce link depths

ibhatt-jumptrading:


ripatel-fd:
Might not be trivial and affect performance, @mjain-jump can you research shallower links without performance loss

mjain-jump (commented):
lets hold off on this for now, currently doing some things with parallel snapshot dc and from there we can size this value out better

ibhatt-jumptrading:
based on mainnet tests, there seems to not be performance regressions here. if the PR is ready and it doesn't impact snapshot load performance, i don't see why it would be problematic, but maybe i'm missing something

mjain-jump:
> based on mainnet tests, there seems to not be performance regressions here. if the PR is ready and it doesn't impact snapshot load performance, i don't see why it would be problematic, but maybe i'm missing something

can confirm, performance impact is negligible so this should be fine

mjain-jump (approved):

PR #10920, opened 2026-08-21
update agave submodule to v4.3.0-beta.2

anwayde:


src/discoh/bank/fd_bank_abi.c
```diff
@@ -298,7 +298,7 @@ FD_STATIC_ASSERT( offsetof(sanitized_txn_abi_v1_cached_msg_t, is_writable_accoun
 typedef union ABI_ALIGN(8UL) {
   ulong discr;
 
-  /* when discr==ABI_HIGH_BIT */
+  /* when discr==ULONG_MAX */
```
ptaffet-jump: This one seems to stay ABI_HIGH_BIT according to my run of extract_abi.py. The rest are correct.

ptaffet-jump (commented):
Looks good. I agree with the `drop_noop_transactions` change. Fix the one ABI comment, and we should be good to merge.

ptaffet-jump (approved):

ripatel-fd (approved):

PR #10934, opened 2026-08-23
config: reject arrays whose elements fail validation

latent-9:
## What

`CFG_POP_ARRAY` and `CFG_POP1_ARRAY` ignored the return value of the element getter, while the scalar `CFG_POP` path already rejects the whole key when its getter fails.

## Why

When one element of an array-valued key fails validation, for example a non-string entry in `gossip.entrypoints` or a value too long for its fixed destination slot, the getter writes nothing and returns 0. The array macros discarded that result and incremented the index anyway, so extraction succeeded with the bad element present as an empty string and counted in `*_cnt`. Because `fd_pod_remove` still runs afterward, the leftover-key check cannot catch it either. The invalid value then surfaces later as a confusing failure somewhere else, such as an empty entrypoint reaching DNS resolution.

## Change

Check the getter's return value in both array macros and reject the config with a warning naming the key, matching what the scalar path does. Regression tests in `test_config_parse.c` cover a wrong-type element in both macros and an over-long entrypoint string.

### DIFF
--- a/src/app/platform/fd_config_macros.c
+++ b/src/app/platform/fd_config_macros.c
@@ -39,7 +39,9 @@
         return NULL;                                                   \
       }                                                                \
       fd_pod_info_t sub_info = fd_pod_iter_info( iter );               \
-      fdctl_cfg_get_##type( &config->cfg_path[j], sizeof(config->cfg_path[j]), &sub_info, key ); \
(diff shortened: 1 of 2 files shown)


esemeniuc:
might be already merged in https://github.com/firedancer-io/firedancer/pull/10932

PR #10936, opened 2026-08-24
http: accept requests that exactly fill the read buffer

latent-9:
## What

`read_conn_http` closed the connection as `LARGE_REQUEST` as soon as the accumulated request reached `max_request_len` bytes, before attempting to parse it.

## Why

The POST path explicitly admits a total size of `max_request_len` (`total_len > max_request_len` is what gets rejected), so the two bounds disagree at the boundary. A request whose head and body together occupy exactly `max_request_len` bytes can never complete: the final read fills the buffer to capacity and trips the early close first, even though the request is valid under the body bound. The same applies to a GET whose head is exactly `max_request_len` bytes.

Since reads are clamped to the remaining buffer space, the accumulated length can never exceed `max_request_len`, so every maximum-size request hit this path.

## Change

Remove the pre-parse close. Treat a full buffer as oversized only when the head still does not parse (`phr_parse_request` returns -2); requests that do parse proceed through the existing content-length bounds, which already admit an exact fill. An incomplete head in a full buffer is genuinely oversized and still closes with `LARGE_REQUEST`.

Added `test_exact_max_request_len_accepted`: it sends a POST whose head and body total exactly 1024 bytes against a server with `max_request_len` 1024, and asserts the request callback fires with no connection close.

### DIFF
--- a/src/waltz/http/fd_http_server.c
+++ b/src/waltz/http/fd_http_server.c
@@ -464,6 +464,7 @@ accept_conns( fd_http_server_t * http ) {
     http->conns[ conn_id ].state                  = FD_HTTP_SERVER_CONNECTION_STATE_READING;
     http->conns[ conn_id ].request_bytes_read     = 0UL;
     http->conns[ conn_id ].request_head_len       = 0UL;
+    http->conns[ conn_id ].request_total_len      = 0UL;
     http->conns[ conn_id ].response_bytes_written = 0UL;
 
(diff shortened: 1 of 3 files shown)


ripatel-fd (changes requested):
300 LOC for what looks like an off-by-one error is excessive. I hate this AI slop

ripatel-fd:
At least type the PR description by hand. If you just send us AI-generated code and comments maybe you're just a robot and then I don't feel like I need to treat you as a human, closing ...

PR #10971, opened 2026-08-25
mlx5: annotate printf function

two-heart:
this is necessary for compiler format string checking


ripatel-fd:
@tristanx86-jump can we get rid of this printf wrapper here?

tristanx86-jump:
> @tristanx86-jump can we get rid of this printf wrapper here?

We can yeah

tristanx86-jump:
fd_rdma_path() removed in https://github.com/firedancer-io/firedancer/pull/10978

PR #10993, opened 2026-08-25
ballet: unpoison asm and ffi outputs

mjain-jump:


cmoyes-jump (approved):

ripatel-fd:
@drubin-fd @0x0ece can you fix the msan poison spam at the bignum api level? we can probably have static inline wrappers for calls to asm routines

PR #10998, opened 2026-08-26
net: add listen_gre and rss_queue_mode to auto-mode

tristanx86-jump:
Some Firedancer users who use DoubleZero were experiencing issues due to Firedancer enabling "dedicated" rss_queue_mode on a system which supports "dedicated" mode but can't support DoubleZero while also in "dedicated" mode.

This PR brings rss_queue_mode and listen_gre under auto-mode by default, which will take into account the nuance of this subset so it's handled correctly.


src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -172,10 +172,11 @@ init_device( char const *        device,
...
     if( FD_UNLIKELY( gre_ntuple_error ) ) {
-      FD_LOG_ERR(( "error configuring network device (%s), failed to install ntuple rule "
-                   "to route GRE packets to Firedancer.  If you require GRE-wrapped "
-                   "traffic (e.g. with DoubleZero), set `net.xdp.rss_queue_mode=\"simple\"`.  "
-                   "Otherwise, set `net.xdp.listen_gre=false`.", device ));
+      if( strict ) FD_LOG_ERR(( "error configuring network device (%s), failed to install ntuple rule "
+                               "to route GRE packets to Firedancer.  If you require GRE-wrapped "
+                               "traffic (e.g. with DoubleZero), set `net.xdp.rss_queue_mode=\"simple\"`.  "
+                               "Otherwise, set `net.xdp.listen_gre=false`.", device ));
+      else         return 1;
```
ripatel-fd: just use curly braces here

ripatel-fd (dismissed):

src/app/fdctl/config/default.toml
```diff
@@ -1063,27 +1063,29 @@ dynamic_port_range = "8900-9000"
...
-        # ignored, and Firedancer automatically listens to GRE-wrapped
-        # traffic when a GRE tunnel interface is configured.
-        listen_gre = false
+        # deliver packets to the validator in a GRE tunnel.
+        #
+        # If rss_queue_mode is "simple", this option is ignored, and
+        # Firedancer automatically listens to GRE-wrapped traffic when
+        # a GRE tunnel interface is configured.
+        #
+        # If you set this to "true" we will ensure you get GRE support,
```
mmcgee-jump: No "we" terminology in here, just prescribe what will will happen.

> Setting this option to "true" will do x to your kernel and cause y.

mmcgee-jump (changes requested):
.

tristanx86-jump:
Tests done with DoubleZero were successful, should be ready to merge.

src/app/shared/commands/configure/ethtool-channels.c
```diff
@@ -172,10 +172,13 @@ init_device( char const *        device,
...
-                   "to route GRE packets to Firedancer.  If you require GRE-wrapped "
-                   "traffic (e.g. with DoubleZero), set `net.xdp.rss_queue_mode=\"simple\"`.  "
-                   "Otherwise, set `net.xdp.listen_gre=false`.", device ));
+      if( strict ) FD_LOG_ERR(( "error configuring network device (%s), failed to install ntuple rule "
+                               "to route GRE packets to Firedancer.  If you require GRE-wrapped "
+                               "traffic (e.g. with DoubleZero), set `net.xdp.rss_queue_mode=\"simple\"`.  "
+                               "Otherwise, set `net.xdp.listen_gre=false`.", device ));
+      else {
+        return 1;
+      }
```
ripatel-fd:
```suggestion
      if( strict ) {
        FD_LOG_ERR(( "error configuring network device (%s), failed to install ntuple rule "
                               "to route GRE packets to Firedancer.  If you require GRE-wrapped "
                               "traffic (e.g. with DoubleZero), set `net.xdp.rss_queue_mode=\"simple\"`.  "
                               "Otherwise, set `net.xdp.listen_gre=false`.", device ));
      } else {
        return 1;
      }
```

src/app/firedancer/config/default.toml
```diff
@@ -1246,29 +1243,28 @@ telemetry = true
...
-        # GRE tunnel interface is configured.
-        listen_gre = false
+        # deliver packets to the validator in a GRE tunnel.
+        #
+        # If rss_queue_mode is "simple", this option is ignored, and
+        # Firedancer automatically listens to GRE-wrapped traffic when
+        # a GRE tunnel interface is configured.
+        #
+        # If this option is set to "auto", GRE support will only be
+        # ensured if DoubleZero is detected.
```
ripatel-fd: can you more precisely specify how it detects whether DZ is running?

src/app/firedancer/config/default.toml
```diff
@@ -1246,29 +1243,28 @@ telemetry = true
...
-        # use GRE tunnels, set rss_queue_mode = "simple".  If
-        # rss_queue_mode is simple, this option is ignored, and
-        # Firedancer automatically listens to GRE-wrapped traffic when a
-        # GRE tunnel interface is configured.
-        listen_gre = false
+        # deliver packets to the validator in a GRE tunnel.
+        #
+        # If rss_queue_mode is "simple", this option is ignored, and
+        # Firedancer automatically listens to GRE-wrapped traffic when
+        # a GRE tunnel interface is configured.
```
ripatel-fd, line 1260: Maybe `If rss_queue_mode is "simple", Firedancer always binds to GRE tunnels`.

mmcgee-jump (approved):

PR #11007, opened 2026-08-26
gui: fix peers sock map key hash and duplicate tiebreak

mmcgee-jump:


src/disco/gui/fd_gui_peers.h
```diff
@@ -281,7 +281,8 @@ typedef struct fd_gui_peers_gossip_stats fd_gui_peers_gossip_stats_t;
 #define MAP_IDX_T ulong
 #define MAP_NEXT  sock_map.next
 #define MAP_PREV  sock_map.prev
-#define MAP_KEY_HASH(k,s) ( fd_hash( (s), (k), sizeof(uint) + sizeof(ushort) ) )
+#define MAP_KEY_HASH(k,s) ( !(k)->is_ipv6 ? fd_ulong_hash( (s) ^ ((ulong)(k)->ip4<<16) ^ (ulong)(k)->port ) \
```
ripatel-fd: int he ipv6 case port is not properly mixed in. move this into a separate static inline function

ripatel-fd (changes requested):

ripatel-fd (approved):

PR #11048, opened 2026-08-27
frontend: development updates

jherrera-jump:
adds
https://github.com/firedancer-io/firedancer-frontend/compare/8d892ed...9ad48b6


mmcgee-jump (dismissed):

ripatel-fd (dismissed):
can we get the mlx5 change in too

mmcgee-jump (approved):

PR #11076, opened 2026-08-28
rewards: fix ad-hoc epoch slot count calculation

ripatel-fd:


ripatel-fd:
@copilot fix the error push a new commit onto this branch

mmcgee-jump (approved):

PR #11084, opened 2026-08-28
runtime: fix UB with invalid epoch schedule

ripatel-fd:


src/flamenco/runtime/fd_bank.h
```diff
@@ -440,6 +440,15 @@ typedef struct fd_banks fd_banks_t;
 /* Bank accessors and mutators.  Different accessors are emitted for
    different types depending on if the field has a lock or not. */
 
+/* fd_bank_banks{,_const} return the fd_banks_t that owns the given
+   bank.  Must only be used on pointers returned by fd_banks_bank_query. */
+
+fd_banks_t *
+fd_bank_banks( fd_bank_t * bank );
```
ibhatt-jumptrading: you don't really want this to be publicly exposed

ripatel-fd: Why not?

ripatel-fd:
@mjain-jump could you fixup the invalid instruction vectors?

PR #11086, opened 2026-08-28
test: fuzz sol_compat txn/instr

cmoyes-jump:


cmoyes-jump:
Going to try to put this in solfuzz instead.

PR #11112, opened 2026-08-29
mlx5: support odd tile counts

### LINKED ISSUE #11108: mlx5: relaxed tile counts
Permit mlx5 tile counts that are not powers of two by using an imperfect indirection table distribution

ripatel-fd:
- always use an indir table 256 or larger
- remove pow2 assumption on tile counts

Closes #11108


ripatel-fd:
@tristanx86-jump could you review and test this on real hw before merging?

PR #11113, opened 2026-08-29
pack: prevent non-bundle txns from touching tip account

ripatel-fd:
the bundle code seemingly wanted to prevent regular txns from touching bundle tip/config accounts via blacklist, but this was not applied correctly. not sure if this is correct, please verify


ptaffet-jump:
That was the intention originally, but normal transactions sometimes pay tips too, and it was silly to drop these.

PR #11116, opened 2026-08-30
util: make signed saturating arithmetic UBSan-clean

esemeniuc:
## Problem
The signed saturating add, subtract, and multiply helpers constructed LONG_MIN by evaluating signed 1+LONG_MAX. Function arguments are evaluated before fd_long_if selects a result, so the overflowing expression ran even when saturation was not selected. Signed overflow is undefined and defeats the purpose of these safety helpers.

## Fix
Select LONG_MIN or LONG_MAX directly from the operand signs after the compiler overflow builtins report overflow. This preserves branchless saturation without evaluating an overflowing boundary expression.


ripatel-fd:
We build with `-fwrapv`, so signed overflow is defined

PR #11117, opened 2026-08-30
shred: make FEC signature keys unaligned-safe

esemeniuc:
## Problem
The FEC signature key wrapper was a union containing ulong, which required eight-byte alignment. Resolver keys can point into byte-aligned shred storage, so map lookup formed and accessed a misaligned typed key. This is undefined even on CPUs that tolerate unaligned loads and may fault on stricter architectures.

## Fix
Make the wrapper byte-aligned and load the hash word with FD_LOAD. Key comparison remains byte-based, and a static assertion protects the required alignment contract.


esemeniuc:
duplicate of https://github.com/firedancer-io/firedancer/pull/10825

PR #11125, opened 2026-08-30
ballet: make r43x6 wrapping references UBSan-clean

esemeniuc:
## Problem
The AVX-512 r43x6 test implemented reference add and subtract with signed long lanes. Random inputs can overflow those scalar operations even though the SIMD instructions intentionally wrap modulo 2^64. Signed overflow makes the reference undefined and allows it to diverge from the operation it checks.

## Fix
Use ulong lanes in the reference unions. Unsigned arithmetic exactly models the SIMD wraparound while preserving every lane bit.


ripatel-fd:
Firedancer uses fwrapv, signed overflow not UB

PR #11171, opened 2026-08-31
build: fingerprint build flavor per OBJDIR

mmcgee-jump:
Every object and link now depends on a cmp-guarded $(OBJDIR)/.flags stamp of the effective compiler+flags, so an OBJDIR reused under a different config rebuilds correctly instead of shipping mixed-flag binaries. Default BUILDDIRs gain a <version>[-extras] suffix so flavors stop sharing paths; bins also hardlink to build/<name> for convenience.


ripatel-fd (dismissed):
`make: *** No rule to make target 'build/linux/clang/icelake/cov/cov.lcov'.  Stop.`

kbhargava-jump (approved):

PR #11194, opened 2026-09-01
bundle: defense-in-depth against zero tip receiver

ripatel-fd:
turn unreachable memory corruption into explicit reject


ptaffet-jump (approved):
The Jito tip program ensures this can't happen, but this defense in depth is okay.

PR #11245, opened 2026-09-01
net: do not enable zero copy on ice

ripatel-fd:
causes remote kernel memory corruption
https://lore.kernel.org/netdev/20260512152658.2818805-1-lorenz@monogon.tech/


mmcgee-jump:
It's fixed in some version? Can't we enable after that version?

ripatel-fd:
reverting #11244 instead

PR #11247, opened 2026-09-01
scheduler: add waker tile

mmcgee-jump:


ripatel-fd:
LETS GOOOOO 🐐 🔥

ripatel-fd (dismissed):
we up waking

kbhargava-jump (approved):

PR #11249, opened 2026-09-01
ed25519: arm support

ripatel-fd:
- **third_party: add s2n-bignum arm curve25519 routines**
- **ed25519: Arm support**


src/ballet/ed25519/fd_curve25519.c
```diff
@@ -13,7 +13,27 @@
 #endif
 
 #define WNAF_BIT_SZ 4
-#define WNAF_TBL_SZ (2*WNAF_BIT_SZ)
+#define WNAF_TBL_SZ 8
```
ripatel-fd, line 16: accident

ripatel-fd:
@Rexicon226 take it over

PR #11277, opened 2026-09-02
flamenco: vectorize fd_borrowed_account_is_zeroed

carlo-osec:
`fd_borrowed_account_is_zeroed` was using a byte-by-byte scan.

This PR implements AVX512 and AVX2 vectorization for that function.


ripatel-fd (approved):
Thanks LGTM. Would have been nice to see perf numbers but I believe this is real for users that do a separate system program Allocate then Assign.

ripatel-fd:
@carlo-osec perhaps worth sending a follow-up if you're interested that relies on the auto-vectorizer on GCC and Clang that way it's portable (godbolt can help figure out ...). would be better than hand-adding for cpu architectures.

PR #11290, opened 2026-09-04
failover: initial MVP with tower file steaming

cmoyes-jump:
https://jumptrading.slack.com/archives/C0B5SNLSA12/p1787855643080949


src/discof/tower/fd_tower_tile.c
```diff
@@ -502,6 +508,165 @@ publish_slot_confirmed( fd_tower_tile_t * ctx,
...
+tower_file_names( fd_pubkey_t const * identity,
+                  char                tower_file_name[ static 64 ],
+                  char                tower_file_name_new[ static 64 ] ) {
+  FD_BASE58_ENCODE_32_BYTES( identity->uc, identity_b58 );
+  FD_TEST( fd_cstr_printf_check( tower_file_name,     64UL, NULL, "tower-1_9-%s.bin",     identity_b58 ) );
+  FD_TEST( fd_cstr_printf_check( tower_file_name_new, 64UL, NULL, "tower-1_9-%s.bin.new", identity_b58 ) );
+}
+
+static int
+tower_file_dir_open( char const * base_path ) {
```
mmcgee-jump, line 521: What is all the directory management stuff for? I can't imagine why we need any of this?

src/discof/admin/fd_admin_tile_failover.seccomppolicy
```diff
@@ -0,0 +1,15 @@
+# The admin tile with failover reads a key from disk only for the moment
+# of a switch, so it never holds a private key resident.  The two key
```
mmcgee-jump: Not sure why it would need this? It can go to the signing tile if anything needs signing? What other reasons it's accessing key?

src/disco/events/schema/admin_command.json
```diff
@@ -11,7 +11,9 @@
                 "get_identity":                 { "description": "Read the running validator identity." },
                 "add_authorized_voter":         { "description": "Add an authorized voter key." },
                 "remove_all_authorized_voters": { "description": "Remove every authorized voter key." },
-                "snapshot_create":              { "description": "Request creation of a snapshot." }
+                "snapshot_create":              { "description": "Request creation of a snapshot." },
+                "failover_status":              { "description": "Read failover pool status." },
+                "failover_control":             { "description": "Apply a failover operator command." }
             }
         },
         "result": {
```
mmcgee-jump, line 19: You added one here?

src/discof/tower/fd_tower_tile.c
cmoyes-jump: This just opens a directory with some error handling, so we have the file descriptor to do a proper atomic sync of the tower file later.

https://man7.org/linux/man-pages/man2/fsync.2.html

>        Calling fsync() does not necessarily ensure that the entry in the
>        directory containing the file has also reached disk.  For that an
>        explicit fsync() on a file descriptor for the directory is also
>        needed.

We don't _need_ it. I'm fine removing it if we don't care about necessarily persisting the tower/votor journal during power outages.

There's a lot of prior art in HA systems for doing it this way though:

https://github.com/postgres/postgres/blob/REL_17_STABLE/src/common/file_utils.c#L455
https://github.com/sqlite/sqlite/blob/version-3.46.1/src/os_unix.c#L3756

FWIW, Agave doesn't do this. So maybe this isn't important? I'm personally indifferent.

src/disco/events/schema/admin_command.json
cmoyes-jump: Sorry?

src/discof/admin/fd_admin_tile_failover.seccomppolicy
cmoyes-jump: Right, it can just be loaded there at startup instead. Let me move.

PR #11291, opened 2026-09-04
[WIP] offline-replay: resolve OBJDIR from the built binary

kbhargava-jump:


ripatel-fd (changes requested):
This is nondeterministic, should resolve it from the build config, not grab a random firedancer-dev bin

PR #11320, opened 2026-09-07
mlx5: detect if ib_uverbs is missing

ripatel-fd:
our CI host had a Mellanox NIC but the ib_uverbs module was not
loaded.  This resulted in firedancer(-dev) startup issues because
auto mode picked mlx5, but mlx5 couldn't init.

- net: do not auto-pick mlx5 if /dev/infiniband/uverbs* missing
- mlx5: human-readable error message if uverbs socket missing

### DIFF
--- a/src/disco/net/mlx5/fd_mlx5_tile.c
+++ b/src/disco/net/mlx5/fd_mlx5_tile.c
@@ -1292,6 +1292,10 @@ fd_topo_install_mlx5( fd_topo_t *     topo,
   }
 
   fd_mlx5_tile_t * first = ctxs[ 0 ];
+  if( FD_UNLIKELY( !fd_mlx5_uverbs_avail() ) ) {
+    FD_LOG_ERR(( "cannot run mlx5 tile: kernel does not provide uverbs API "
+                 "(ib_uverbs kernel module missing or NIC not Mellanox?)" ));
+  }
   FD_LOG_INFO(( "Opening direct mlx5 device `%s` port %u for %lu tiles",
                 rdma_device_name, rdma_port_num, tile_cnt ));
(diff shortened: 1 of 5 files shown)


ripatel-fd:
@tristanx86

ripatel-fd:
@tristanx86 @mmcgee-jump It seems `ib_uverbs` module is loaded by udev rules installed by `rdma-core`. That CI host didn't have that package, so when it reboots, nothing loaded the kernel module. Should probably load `ib_uverbs` in a configure stage if the bus has a PCIe device matching Mellanox.

src/disco/net/mlx5/fd_mlx5_tile.c
```diff
@@ -1292,6 +1292,10 @@ fd_topo_install_mlx5( fd_topo_t *     topo,
   }
 
   fd_mlx5_tile_t * first = ctxs[ 0 ];
+  if( FD_UNLIKELY( !fd_mlx5_uverbs_avail() ) ) {
+    FD_LOG_ERR(( "cannot run mlx5 tile: kernel does not provide uverbs API "
+                 "(ib_uverbs kernel module missing or NIC not Mellanox?)" ));
```
tristanx86, line 1297: imo we should split "or NIC not Mellanox?" into an mlx5 driver check just before

mmcgee-jump (approved):

src/disco/net/mlx5/fd_mlx5_tile.c
ripatel-fd: @tristanx86 ty good point ... would you mind sending a followup?

tristanx86: Yep!

PR #11321, opened 2026-09-07
gossip: fix 3 byte OOB write

ripatel-fd:
benign OOB write into padding

Octane report:

  Out-of-bounds write in deser_duplicate_shred:
  a DUPLICATE_SHRED-typed CRDS value embedded in a PULL_REQUEST
  reaches the parser with a 3-byte-smaller preamble, letting
  unvalidated chunk_len (up to 1058) exceed chunk[1055] and
  overflow the array.

### DIFF
--- a/src/flamenco/gossip/fd_gossip_message.c
+++ b/src/flamenco/gossip/fd_gossip_message.c
@@ -291,6 +291,7 @@ deser_duplicate_shred( fd_gossip_value_t * value,
   READ_U8( value->duplicate_shred->chunk_index, payload, payload_sz );
   CHECK( value->duplicate_shred->chunk_index<value->duplicate_shred->num_chunks );
   READ_U64( value->duplicate_shred->chunk_len, payload, payload_sz );
+  CHECK( value->duplicate_shred->chunk_len<=sizeof(value->duplicate_shred->chunk) );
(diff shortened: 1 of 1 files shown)


src/flamenco/gossip/fd_gossip_message.c
```diff
@@ -291,6 +291,7 @@ deser_duplicate_shred( fd_gossip_value_t * value,
   READ_U8( value->duplicate_shred->chunk_index, payload, payload_sz );
   CHECK( value->duplicate_shred->chunk_index<value->duplicate_shred->num_chunks );
   READ_U64( value->duplicate_shred->chunk_len, payload, payload_sz );
+  CHECK( value->duplicate_shred->chunk_len<=sizeof(value->duplicate_shred->chunk) );
```
mmcgee-jump, line 294: Prefer just make the buffer bigger, and update the documentation (the calculation docs) to allow this rather than reject?

mmcgee-jump: Also there's a fuzzer for this, why doesn't trigger? Are you sure the report is valid?

ripatel-fd: @mmcgee-jump ASan cannot find OOB that small. I don't think making the buffer bigger would fix it it's some weird edge case. Will look into it more

PR #11334, opened 2026-09-07
topo: let tiles float within the layout

mmcgee-jump:
A tile with floats set keeps its cpu_idx for memory placement and isolation but runs with the affinity of every floating tile's CPU on its NUMA node, never a pinned tile's.  The cpuset stage then creates a root partition (exclusive CPUs with their own scheduler domain) instead of an isolated one, run refuses floaters on isolcpus= CPUs and raises the waker's priority (it delivers their fd readiness), and the SMT and per-CPU diagnostics stop treating a floater's CPU as its own.  No topology sets floats yet.

### DIFF
--- a/src/app/shared/commands/configure/cpuset.c
+++ b/src/app/shared/commands/configure/cpuset.c
@@ -11,6 +11,14 @@
    `isolcpus=` boot parameter, and the strongest tool available for
    keeping foreign tasks off tile CPUs.
 
+   The partition type follows the topology.  When every tile is pinned
+   to its own CPU the partition is "isolated".  When some tiles float
+   (the kernel schedules them across a shared set of tile CPUs) it is
+   "root" instead: still exclusive to Firedancer, but the CPUs keep a
+   scheduler domain, which load balancing needs.  An isolated partition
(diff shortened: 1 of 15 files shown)


src/app/shared/commands/configure/cpuset.c
```diff
@@ -11,6 +11,13 @@
    `isolcpus=` boot parameter, and the strongest tool available for
    keeping foreign tasks off tile CPUs.
 
+   When tiles float within the tile CPU set (efficient scheduler mode)
+   the partition is created as "root" instead: the same exclusivity,
+   but the CPUs keep a scheduler domain of their own so the floating
+   tiles are load balanced across them.  An isolated partition has no
+   domain, and floaters inside it would never leave the CPU they were
+   forked on.
```
ripatel-fd: This comment is unreadable

ripatel-fd (dismissed):

ripatel-fd:
> No topology sets floats yet.

Isn't this just wrong?

ripatel-fd (dismissed):

ripatel-fd (dismissed):

ripatel-fd (approved):

PR #11338, opened 2026-09-08
rotor: never overflow the alpenglow request queue

mmcgee-jump:
ag_req_queue holds FD_FEC_BLK_MAX messages and drains one per credit, but a verified ParentAndFecSetCount response pushed up to FD_FEC_BLK_MAX FecSetRoot requests unchecked, so two such responses in flight overflowed the ring: fd_queue_push has no capacity check, entries were overwritten and the count went inconsistent.  Guard every push; when the queue cannot take a response's requests the inflight is kept and the aged-out redispatch retries it, since a FEC root that was never requested is never re-requested on its own.


lidatong:
@emwang-jump

emwang-jump:
im working on a fix for this, i think this PR is missing some inflights tracking

PR #11340, opened 2026-09-08
runtime: cost tracker polishing

ripatel-fd:
~~security fix, needs v26.09 backport~~


topointon-jump:
See discussion on Slack, I don't think we need to merge this

PR #11361, opened 2026-09-08
config: automatic net tile count selections

drubin-fd:
@tristanx86

### DIFF
--- a/src/app/shared/fd_config_auto.c
+++ b/src/app/shared/fd_config_auto.c
@@ -257,13 +257,19 @@ mlx5_check( fd_config_t    const * config,
   if( strcmp( config->net.provider, "auto" ) ) return 0;
   if( !info->has_mlx5_rdma_port ) return 0;
   if( !info->has_uverbs ) return 0;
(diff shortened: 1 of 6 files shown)


src/app/shared/fd_config_auto.c
```diff
@@ -264,6 +264,7 @@ mlx5_check( fd_config_t    const * config,
 static void
 mlx5_apply( fd_config_t * config ) {
   fd_memcpy( config->net.provider, "mlx5", 5 );
+  config->layout.net_tile_count = 1U;
```
mmcgee-jump: Not the right way to do this, decoupled from the option in the config file.

Better to add a 0 option to `net_tile_count` which represents auto.

mmcgee-jump (changes requested):
.

src/app/shared/fd_config_auto.c
drubin-fd: Understood, I'll do that.

drubin-fd:
I'll just leave this for @tristanx86 to finish, duplicating work.

PR #11374, opened 2026-09-09
keyguard: bls signer

emwang-jump:
stilll waiting to test do not click-happy merge

### DIFF
--- a/src/disco/keyguard/fd_sign_tile.c
+++ b/src/disco/keyguard/fd_sign_tile.c
@@ -9,6 +9,7 @@
 #include "../../discof/admin/fd_adminctl.h"
 #include "../../ballet/base58/fd_base58.h"
 #include "../metrics/fd_metrics.h"
+#include "../../choreo/votor/ag_bls.h"
 
 #include "../../util/hist/fd_histf.h"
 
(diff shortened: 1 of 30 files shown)


src/disco/keyguard/fd_sign_tile.c
```diff
@@ -89,6 +94,17 @@ derive_fields( fd_sign_ctx_t * ctx ) {
 
   fd_base58_encode_32( ctx->public_key, &ctx->public_key_base58_sz, (char *)ctx->concat );
   ctx->concat[ ctx->public_key_base58_sz ] = '-';
+
+  /* Alpenglow BLS key derivation, matching
+     solana_bls_signatures::SecretKey::derive_from_signer: the BLS IKM
+     is the identity's ed25519 signature over a fixed message. */
+#if FD_HAS_BLST
```
drubin-fd: Let's avoid the `FD_HAS_BLST`, it was originally stubs, but I do not see any reason to do this anymore.

emwang-jump: 👌

ripatel-fd (approved):
Looks good. It would be nice to run the Alpenglow state machine in the signing tile and refuse to sign for equivocating or otherwise invalid blocks as defense-in-depth.

lidatong (approved):

PR #11399, opened 2026-09-10
backt: read shred version from genesis

emwang-jump:


kbhargava-jump (approved):

emwang-jump:
Closing -- better fix is to always add the genesis tile to backtest topology. Then we don't need to special case anything in replay; hard forks always read from snapshot and genesis hash always read from genesi tile

PR #11423, opened 2026-09-11
blake3: specialize the 2048 byte XOF for a single root block

mmcgee-jump:
Expanding the root block into 32 output blocks went through the generic compress16, which stages sixteen copies of the same message and chaining value and transposes them, when only the counter differs between lanes.  Broadcast them instead.  Per byte of input, 8.98 -> 4.80 ns at 64 bytes and 4.92 -> 2.88 ns at 128; execle runs two of these per transaction for the account lthash.


ripatel-fd (dismissed):
Nice but we have an lthash_adder path that significantly outperforms this strat by streamlining XOF with input hashing. Philip is integrating

ripatel-fd (approved):

PR #11436, opened 2026-09-12
ballet, txn: reject v1 transactions with duplicate addresses

topointon-jump:
technically this is unnecessary, these are caught by a runtime/replay check. but it is more consistent with agave's behaviour, which throws these out at sanitization time, and makes it easier to fuzz. it's also more in line with the SIMD.

@ptaffet-jump we should discuss on monday if we actually want to merge this

### DIFF
--- a/src/ballet/txn/fd_txn_parse.c
+++ b/src/ballet/txn/fd_txn_parse.c
@@ -137,6 +137,14 @@ fd_txn_parse_core( uchar const             * payload,
 
     CHECK_LEFT( FD_TXN_ACCT_ADDR_SZ*acct_addr_cnt ); ulong acct_addr_off  =          i  ;     i+=FD_TXN_ACCT_ADDR_SZ*acct_addr_cnt;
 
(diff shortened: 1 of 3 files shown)


src/ballet/txn/fd_txn_parse.c
```diff
@@ -137,6 +137,14 @@ fd_txn_parse_core( uchar const             * payload,
 
     CHECK_LEFT( FD_TXN_ACCT_ADDR_SZ*acct_addr_cnt ); ulong acct_addr_off  =          i  ;     i+=FD_TXN_ACCT_ADDR_SZ*acct_addr_cnt;
 
+    /* Throw out transactions with duplicate addresses at parse time.
+       Agave rejects these at sanitisation time. */
```
mjain-jump, line 141:
```suggestion
       Agave rejects these at sanitization time. */
```
🇺🇸

topointon-jump:
Closing in favour of https://github.com/firedancer-io/firedancer/pull/11464

PR #11439, opened 2026-09-13
topology: keep verify_dedup shallow enough to stay in cache

mmcgee-jump:
Each verify tile wrote its transactions into a 16384 x FD_TPU_PARSED_MTU ring per dedup, 160 MiB per tile that cycles through DRAM, so every copy out was a read for ownership from memory.  1024 slots is 25 ms of slack at 40k transactions per second per tile.


ripatel-fd (approved):
very nice

PR #11441, opened 2026-09-13
bmtree: copy proof nodes with one masked store

mmcgee-jump:
fd_bmtree_get_proof copied each sibling with a runtime-sized fd_memcpy, which GCC lowers to a memcpy PLT call: 384 calls per leader FEC set (64 shreds x 6 layers).  Nodes are 32 bytes, so on AVX-512 load the node and store hash_sz bytes with one masked store.


ripatel-fd (approved):
LGTM, but why not just 
```
switch( hash_sz ) {
  ...
  default: FD_LOG_CRIT( ... )
}
```

PR #11442, opened 2026-09-13
sha256: fold the batch kernel's booleans into vpternlogd

mmcgee-jump:
GCC does not fuse the xor/and/andnot trees in the AVX-512 batch kernel; each round spent 12 logic ops where 4 vpternlogd suffice (416 -> 160 per 32 rounds, about a quarter fewer uops per block). The 32-repeated PoH kernel in the same file already did this.  On Zen 4 the leaf hashing of a 64-shred FEC set goes from 2.92 to 3.53 GB/s.  Output is unchanged; test_sha256 now also checks the batch path against fd_sha256_hash at every size up to 4096 in every lane.


ripatel-fd (approved):
vpternlog has some interesting performance issues that @ptaffet-jump and @drubin-fd investigated, but LGTM

PR #11449, opened 2026-09-13
bmtree: hash a bulk append layer by layer with the sha256 batch

mmcgee-jump:
fd_bmtree_commit_append merged nodes one at a time as each leaf arrived, so the shredder's 64-leaf tree cost 63 dependent single-stream SHA-256 calls of two blocks each.  When an empty commit is given 8..64 leaves at once, build the tree layer by layer instead and hash each layer's merges as one fd_sha256_batch.  The resulting state is the same: same inclusion_proofs slots, same node_buf, odd nodes left for fini exactly as before; test_bmtree checks that against the leaf-at-a-time path for every leaf count 1..64 and three (hash_sz, prefix_sz) shapes.  A 64-leaf tree with its 64 proofs goes from 7.4 to 4.4 us on one Zen 4 core.


ripatel-fd:
@mmcgee-jump These changes seem dubious? Zero bench improvement, why should we eat the new code footprint and code complexity?

ripatel-fd:
I see, it's for 1M TPS!

ripatel-fd (approved):

PR #11451, opened 2026-09-13
shred: stream the copies that nothing in the tile reads back

mmcgee-jump:
New fd_memcpy_nt_pad_nofence streams the partial last line whole for a destination that owns the slack, so the tail of a copy is no longer a read for ownership of a cold line.  The tile stages packets with it and publishes after one fence per fan-out, streams the shred_out and store copies with one fence each, and pulls the entry batch out of the poh frag with fd_memcpy_tn.


src/disco/shred/fd_shred_tile.c
```diff
@@ -1195,6 +1220,7 @@ after_frag( fd_shred_ctx_t *    ctx,
         for( ulong j=0UL; j<*max_dest_cnt; j++ ) send_shred( ctx, stem, new_shreds[ i ], fd_shred_dest_idx_to_dest( sdest, dests[ j*out_stride+i ]), ctx->tsorig );
       }
     } while( 0 );
+    net_flush( ctx, stem );
```
ripatel-fd: lgtm but doesn't have much to do with the change?

src/disco/shred/fd_shred_tile.c
```diff
@@ -912,6 +922,19 @@ during_frag( fd_shred_ctx_t * ctx,
   }
 }
 
+static inline void
+net_flush( fd_shred_ctx_t *    ctx,
+           fd_stem_context_t * stem ) {
+  if( FD_LIKELY( !ctx->net_pend_cnt ) ) return;
+
+  _mm_sfence();
```
ripatel-fd: sfence might be too expensive to run per-packet, have you profiled?

src/disco/shred/fd_shred_tile.c
ripatel-fd: nvm i see now, stem publish is deferred

src/disco/shred/fd_shred_tile.c
mmcgee-jump: Yes, this is about -25% on shred tile CPU
> memmove was 30% of the leader shred tile — normal stores into cold dcache lines, not the copies themselves. Shred 97% → 75%.

ptaffet-jump (approved):

PR #11458, opened 2026-09-14
zstd: remove unused code

two-heart:


mmcgee-jump (approved):

mmcgee-jump (dismissed):
It looks like this code is unused? Can you just delete these files?

ripatel-fd (approved):

PR #11461, opened 2026-09-14
runtime: support gaps in block hash queue

emwang-jump:


src/flamenco/runtime/fd_blockhashes.h
```diff
@@ -10,7 +10,27 @@
...
+
+   The block hash queue only purges once the map holds at least 300
+   entries, but duplicate registrations eject entries from the map and
+   not the deque (for purposes of maintaining the mapping from blockhash
+   to age/position).  Each hole therefore lets the span grow by one past
+   301 until the "holes" fall off the end.  512 covers more than 200
+   holes; anything beyond is rejected at snapshot load. */
+
+#define FD_BLOCKHASHES_MAX      (301UL)
+#define FD_BLOCKHASHES_SPAN_MAX (512UL)
```
mmcgee-jump, line 33: This seems arbitrary? What is the real bound?

emwang-jump: Theres no real bound if we keep this as a deq, because the way this can exceed FD_BLOCKHASHES_MAX is using the agave-ledger-tool.  Not sure if its worth refactoring to use a hash_index for this edge case?

PR #11463, opened 2026-09-14
backup: add slow-peer throughput guard

### LINKED ISSUE #10952: Snapshot server vulnerable to slowloris / conn squatting
reported by @hrvoje-asym https://github.com/firedancer-io/auditor-internal/issues/600

By default the snapshot server has a timeout of 1 second where it waits for a valid http requests on a connection slot, however there is no absolute timeout for how long a connection can stay open, and/or maximum number of requests a single connection can make.

This can result in a DoS in case the attacker opens 128 connections to the server and approximately every half a second sends a small and valid HTTP request, for example:
```
HEAD / HTTP/1.1
Host: h
```

After this request is consumed, the server will reset `request_start_nanos` to 0 [here](https://github.com/firedancer-io/firedancer/blob/6640f59511b7637d99414b4d8d88427378d6eca9/src/discof/backup/fd_snapsv_tile.c#L1451), and then when it starts waiting for the next request `prep_peek` sees its value is 0 and starts a fresh timeout window [here](https://github.com/firedancer-io/firedancer/blob/6640f59511b7637d99414b4d8d88427378d6eca9/src/discof/backup/fd_snapsv_tile.c#L579). This results in the attacker holding all 128 connection slots occupied indefinitely, and a real client trying to connect [will not be accepted](https://github.com/firedancer-io/firedancer/blob/6640f59511b7637d99414b4d8d88427378d6eca9/src/discof/backup/fd_snapsv_tile.c#L754-L756).

Low severity because the snapshot server is not publicly exposed by default.
Probably a blocker for making the snapshot server publicly exposed.

0xfka:
This PR introduces minimum throughput requirement for snapsv's peers to increase the cost of DoS significantly. 
To halt snapsv on current conditions, sending few bytes per second to reset the timer is enough. 
Scaling conn_max is not increasing the cost of DoS too much because without any throughput requirement it's not that hard to keep connections alive. If we add a minimum threshold for throughput, such as the dummy 3 MiB per second, with 10 second window  I coded, and scale the conn_max to 100k as Richard said,  the l7 traffic required is 2.52 terabits per sec, which is quite hard. I think the 10 second window is preventing false positives caused by jitter but probably can  be lower, like 7-8 etc. 
aims to fix https://github.com/firedancer-io/firedancer/issues/10952 and many snapsv DoS posibilities


0xfka:
@ripatel-fd  
I think this PR should be open with some label on it so when snapshot server is going to be exposed this can be reviewed. I know it is currently low priority but this code + making conn_max 100k as you said, increases the cost of attack to **2.52 terabits per second,** which I believe is required on this snapsv tile.

mmcgee-jump:
I don't think this should exist? Needing to guard throughput indicates other issues with the server, e.g. a bad eviction policy.

0xfka:
@mmcgee-jump I agree on eviction policy part, you can check the linked issue for that one
Do you have any policy in mind ?

ripatel-fd:
As discussed in Discord, the snapshot server should not have per-conn throughput limits

PR #11487, opened 2026-09-15
third_party: remove zlib

### LINKED ISSUE #11486: remove zlib/gzip
all major browsers support zstandard (including safari)

gzip support saves us 78 bytes in total (in cases where zstd packs larger). 

not worth having a whole dependency

ripatel-fd:
Closes #11486 

All major browsers support Zstandard. Removes 17.8K lines of code that save 71 bytes worth of GUI assets


ripatel-fd:
image


cooked 
image

PR #11488, opened 2026-09-15
snapshot: remove feature snooping

ripatel-fd:
Fixes bug where features are sometimes not restored correctly


ripatel-fd:
I know this regresses snap load time (which seems to be incorrectly attributed to replay), but this is a correctness fix

mjain-jump (approved):

PR #11502, opened 2026-09-15
ed25519: add fd_ed25519_sign_batch8

0x0ece:
Sign in batches of n (1..8).

This is a "safe" implementation where:
- sha512 is batched (eg using avx512)
- const time scalar mul is UNTOUCHED
- point compression is batched using batch inversion

For messages of 1024 bytes, batch n=8 is 1.85x faster on zen4, 2x faster on zen5.

Future work: const time batch scalar mul. cc @drubin-fd 

---

(not in this PR)
When batch is applied to sign tile, repair gets an improvement on avg time to sign a msg (from the time the msg is queued to sign, to the time it gets signed).
```
  ┌───────────────┬────────┬─────────┐
  │      run      │  p50   │   p99   │
  ├───────────────┼────────┼─────────┤
  │ A (unbatched) │ 525 µs │ 1883 µs │
  ├───────────────┼────────┼─────────┤
  │ B (batch8)    │ 292 µs │ 1829 µs │
  └───────────────┴────────┴─────────┘
```


ripatel-fd:
can you post A/B test results for tiles that depend on synchronous signing such as shred and repair? curious what the p99 sign latency impact is of this, and how much more ticks they are blocked on waiting for sign responses

mmcgee-jump (approved):

PR #11529, opened 2026-09-15
rotor: v2 APIs, scheduler, requester

emwang-jump:


src/discof/chainer/fd_chainer.h
```diff
@@ -246,6 +247,45 @@ typedef struct out_ele out_ele_t;
...
+                   every turbine version renamed at finalization
+   - FEC_ADDED     a FEC entry attached to a version: first turbine
+                   shred, getFecRoot sentinel, or repoint on rekey
+   - FEC_EVICTED   one of a version's FEC sets lost its shreds
+   - SLOT_COMPLETE every FEC set of the version became reconstructable,
+                   once per version, before any rename at finalization
+   - SLOT_RETIRED  the version needs no more repair: rooted at init,
+                   abandoned, renamed at finalization, slot-complete FEC
+                   delivered, parent is a dead fork, or pruned by
+                   publish.  May repeat for one version.
```
lidatong: BLK_ADDED / BLK_COMPLETE is a lot clearer than this "version" concept imo

lidatong (commented):
i think you can pull all the event handlers and their definitions into the tile itself

the modules ideally are "pure data structures", just aggregations of the structures you need with a simple `insert` `query` `remove` `iter` ... API

the actual event state machine is entirely driven by the tile (see tiles like gossip, snap)

votor got trapped by the ref impl, and also adhering to the pool / votor separation from the paper. so there's this awkward translation layer between internal (talking to other alpenglow components like pool) vs. external (talking to other tiles)

but i think here that doesn't apply. rotor reacts exclusively to events from shred_out or votor_out (based on fd_rotor_event) or events it generates itself

so essentially schedulor and requestor can just be the tile code

src/discof/rotor2/fd_schedulor.c
```diff
@@ -0,0 +1,317 @@
...
+   pool on schedule and released on pop or cancel. */
+
+struct task_key {
+  ulong     slot;
+  fd_hash_t block_id;
+};
+typedef struct task_key task_key_t;
+FD_STATIC_ASSERT( sizeof(task_key_t)==40UL, task_key_sz );
+
+struct task {
```
lidatong, line 16: req

lidatong (commented):
requestor / schedulor separation still needed?

lidatong (approved):

PR #11531, opened 2026-09-15
ballet: fix undefined ag_bls_agg refs with GNU ld

mjain-jump:


ripatel-fd (changes requested):
we should not move ag_ prefixed APIs into ballet

mjain-jump:
discussed offline, will be fixed properly later

lidatong:
> we should not move ag_ prefixed APIs into ballet

this is due to an unfortunate circular dep between choreo / flamenco right now. legacy tower depends on flamenco, in AG dep flows other way

i think the best solution is to move alpenglow into its own library, and then replace choreo when we no longer have to maintain both

PR #11538, opened 2026-09-16
votor: reduce mem by only retaining vote sigs

lidatong:


ripatel-fd (commented):
can you rebase on latest main so we can test benchmark CI print the AG memory requirements?

ripatel-fd (approved):

PR #11551, opened 2026-09-16
build: thin archives

mmcgee-jump:
ar rcsT records member paths instead of copying the objects into lib/*.a; the archives now require obj/ beside them.  The archiver and its flags (minus VERBOSE's v) join each lib's member-list stamp so a format change re-archives.  macos/freebsd (BSD ar, where T truncates names) keep rcs.


ripatel-fd (dismissed):
Lol we can just delete FreeBSD support no one uses that

ripatel-fd (approved):

PR #11568, opened 2026-09-16
config: use static binaries with pie

ripatel-fd:
statically link libc.  This is fine even for glibc because we do not use NSS, dlopen, iconv, locales, or time zones.  Important for reproducible runtime behavior, especially for libm.so.


drubin-fd (approved):
megabased

ripatel-fd:
@drubin-fd can you take this over thanks

PR #11576, opened 2026-09-17
build: no bash or GNU find needed to parse the makefiles

mmcgee-jump:
native_config.sh was #!/bin/bash but uses nothing beyond POSIX sh (its logic is awk), and `find -L` is not understood by busybox find.  On a box without bash the script failed inside $(shell) and the build went on with an empty compiler config; on Alpine `find -L` made LOCAL_MKS empty.  Command streams are unchanged.


config/everything.mk
```diff
@@ -51,11 +51,11 @@ endif
 
 # Grab all the Local.mk files in the source tree, save to a variable so that
 # other rules can depend on this list. We will include these files later on.
-# Don't use "-L" if source code directory structure has symlink loops.
+# No -L: busybox find lacks it and src/ has no symlinked directories.
```
ripatel-fd, line 54: ?

ripatel-fd (approved):

PR #11577, opened 2026-09-17
musl build fixes

ripatel-fd:


src/disco/events/fd_boot_report.c
```diff
@@ -206,8 +209,13 @@ collect_os( fd_boot_report_t * r ) {
     }
   }
 
-  r->libc_kind = 1; /* glibc: the only supported libc */
+#if __has_include(<gnu/libc-version.h>)
+  r->libc_kind = 1; /* glibc */
   fd_cstr_printf( r->libc_version, sizeof(r->libc_version), NULL, "%s", gnu_get_libc_version() );
+#else
+  r->libc_kind = 2; /* musl */
+  r->libc_version[ 0 ] = '\0';
```
mmcgee-jump, line 217: musl doesn't have a version?

ripatel-fd: no unfortunately not

drubin-fd (approved):

mmcgee-jump (changes requested):
just fix the version reporting, should take 30 seconds with claue

PR #11590, opened 2026-09-17
txncache, backup: lock-free serialization of the txncache

yufeng-jump:
The snapshot production pipeline is not pinned by default.  Without a way for the tiles involved to judiciously block and unblock preemption, these tiles cannot be allowed to hold on to critical spin locks.  The serialization of the txncache currently holds on to a spin rwlock that the replay tile also contends for.  This has been observed to cause replay to be stalled for upwards of a second at a time.

The txncache now tracks a mutation generation for transactions and a root generation for rooted ancestry.  This allows the snapshot producer to walk the txncache and detect concurrent changes and retry, all without taking any locks.


src/flamenco/runtime/fd_txncache.c
```diff
@@ -149,6 +149,8 @@ fd_txncache_join( void * ljoin ) {
 void
 fd_txncache_reset( fd_txncache_t * tc ) {
   fd_rwlock_write( tc->shmem->lock );
+  fd_txncache_mutation_begin( tc->shmem );
+  FD_VOLATILE( tc->shmem->root_gen ) = FD_VOLATILE_CONST( tc->shmem->root_gen )+1UL;
```
mmcgee-jump: you need store-load barriers after this, same as accdb

mmcgee-jump: Ah nvm, looking closer

mmcgee-jump (dismissed):

mmcgee-jump (approved):

PR #11593, opened 2026-09-18
poh: fix replay input starvation while following

mmcgee-jump:
1008b4921c moved unhandled replay messages into before_frag, bypassing the idle counter reset in returnable_frag.  PoH could then hash while replay input was still queued, delaying RESET and BECAME_LEADER behind transaction notifications and backpressuring replay.

Reset the idle counter when filtering replay messages without a leader bank.  Preserve hashing progress while holding a leader bank, including when hashing through skipped slots.

Thank you to Matthias Schmitz from Staking Facilities for help debugging.


ptaffet-jump (approved):
I see other frags cause ctx->idle_cnt to be reset in `returnable_frag`

